PR #10, opened 2016-06-03
Update docs with Git + encryption changes

brikis98:
Do Readme Driven Development to flush out the ideas for:
1. Using Git for locking
2. Automatic state management with Git
3. Using KMS for encryption

### DIFF
--- a/README.md
+++ b/README.md
@@ -484,8 +484,8 @@ or Consul? Should you enable encryption or not?
 
 We recommend using one of the following two configurations:
 
-1. Git locking, Git state management, and KMS encryption
-1. DynamoDB locking and S3 state management
+1. [Git locking, Git state management, and KMS encryption](#git-locking-git-state-management-and-kms-encryption)
+1. [DynamoDB locking and S3 state management](#dynamodb-locking-and-s3-state-management)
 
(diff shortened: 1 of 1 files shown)


brikis98:
Update: this PR is now very out of date, so it can't be merged as-is. However, the general ideas and discussion over whether Git is a good locking mechanism are still worth considering.

mechastorm:
Any chance this feature can be merged in yet to master after dealing with the conflicts?

brikis98:
This PR was just put up for discussion. It has not actually been implemented. Comments and feedback welcome :)

robkinyon:
I'm confused - why would KMS encryption only be usable with the Git backend? There's no discussion why I wouldn't be able to use KMS encryption with the S3 backend.

brikis98:
> There's no discussion why I wouldn't be able to use KMS encryption with the S3 backend.

The S3 backend already has encryption if you set the `encrypt` config to true.

brikis98:
Since Terraform now natively supports locking, no longer keeps a local cache of the tfstate file, and has more plans related to security on the horizon, going forward with the plan in this PR makes no sense, so I'm going to close it.

PR #18, opened 2016-08-05
Add Terragrunt integration test for latest Terraform.

josh-padnick:
This PR adds a simple integration test to validate that Terragrunt works with the latest version of Terraform. It executes a trivial Terraform template that just renders a `data.template_file` resource.

Do NOT merge this PR until https://github.com/gruntwork-io/module-ci/pull/6 is merged and a new release is tagged in https://github.com/gruntwork-io/module-ci since running this integration test successfully depended on an update to `configure-environment-for-gruntwork-module` in github.com/gruntwork-io/module-ci.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -5,8 +5,8 @@ machine:
 dependencies:
   override:
     # Install the gruntwork-module-circleci-helpers and use it to configure the build environment and run tests.
-    - curl -Ls https://raw.githubusercontent.com/gruntwork-io/gruntwork-installer/master/bootstrap-gruntwork-installer.sh | bash /dev/stdin --version 0.0.9
-    - gruntwork-install --module-name "gruntwork-module-circleci-helpers" --repo "https://github.com/gruntwork-io/module-ci" --tag "v0.0.10"
+    - curl -Ls https://raw.githubusercontent.com/gruntwork-io/gruntwork-installer/master/bootstrap-gruntwork-installer.sh | bash /dev/stdin --version 0.0.13
+    - gruntwork-install --module-name "gruntwork-module-circleci-helpers" --repo "https://github.com/gruntwork-io/module-ci" --branch "place-terraform-in-path"
     - configure-environment-for-gruntwork-module --packer-version NONE --go-src-path .
 
   cache_directories:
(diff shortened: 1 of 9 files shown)


circle.yml
```diff
@@ -5,9 +5,9 @@ machine:
 dependencies:
   override:
     # Install the gruntwork-module-circleci-helpers and use it to configure the build environment and run tests.
-    - curl -Ls https://raw.githubusercontent.com/gruntwork-io/gruntwork-installer/master/bootstrap-gruntwork-installer.sh | bash /dev/stdin --version 0.0.9
-    - gruntwork-install --module-name "gruntwork-module-circleci-helpers" --repo "https://github.com/gruntwork-io/module-ci" --tag "v0.0.10"
-    - configure-environment-for-gruntwork-module --terraform-version NONE --packer-version NONE --go-src-path .
+    - curl -Ls https://raw.githubusercontent.com/gruntwork-io/gruntwork-installer/master/bootstrap-gruntwork-installer.sh | bash /dev/stdin --version 0.0.13
+    - gruntwork-install --module-name "gruntwork-module-circleci-helpers" --repo "https://github.com/gruntwork-io/module-ci" --tag "v0.0.18"
+    - configure-environment-for-gruntwork-module --packer-version NONE --go-src-path .
```
brikis98: Won't this install terragrunt now? Shouldn't the test use its local version instead?

test/integration_test.go
```diff
@@ -0,0 +1,120 @@
...
+		t.Fatalf("Failed to run terragrunt: %s", err)
+	}
+}
+
+// Run Terragrunt Apply directory in the test fixture path
+func runTerragruntApply() error {
+	os.Chdir(TEST_FIXTURE_PATH)
+
+	app := cli.CreateTerragruntCli("TEST")
+	err := app.Run(strings.Split("terragrunt apply", " "))
```
brikis98: You can just return directly from this line

brikis98:
Changes lgtm. Just some minor comments.

test/integration_test.go
josh-padnick: Good call...I fixed those types of situations earlier and then had to revert them under the use of `assert` since the assert print to stdout gets swallowed when it's in a return, but in this case, that didn't apply. Fixed.

circle.yml
josh-padnick: The current test runs Terragrunt by [directly invoking the CLI app](https://github.com/gruntwork-io/terragrunt/pull/18/files#diff-e38680b417b49add3697d02eb94cf52fR39), but the point is well-taken. To avoid any confusion, I'll disable terragrunt explicitly.

josh-padnick:
Now that https://github.com/gruntwork-io/module-ci/pull/6 is merged, tests are passing, so merging.

Issue #21, opened 2016-08-11
AWS Profiles


pixie79:
Hi,

How does terragrunt interact with different AWS account credentials? Is there an option to set the profile in the config or will it respect the AWS_DEFAULT_PROFILE variable to get the correct API keys for use.

Thanks

Mark

josh-padnick:
Hi @pixie79, thanks for the question!

Since terragrunt is written in Go and uses the [AWS SDK for Go](https://docs.aws.amazon.com/sdk-for-go/latest/v1/developerguide/welcome.title.html), the AWS authentication is handled completely by the AWS SDK. 

You can read about the [specific authentication hierarchy used by the AWS SDK](https://docs.aws.amazon.com/sdk-for-go/latest/v1/developerguide/configuring-sdk.title.html), which clarifies that terragrunt will first look for the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`. If those are undefined, it will use the shared credentials file, and I'm guessing it will also respect the `AWS_DEFAULT_PROFILE` env var, though this would ultimately be up to the AWS SDK.

If this answered your question, feel free to close this issue. Otherwise, I'll leave it open for a bit in case you have any follow-up questions.

mmorianos:
I notice you don't support the use of temporary credentials which require a 3rd environment variable, AWS_SESSION_TOKEN.  We don't use IAM:users or issue long term access keys, instead we built a service that pulls temporary credentials via STS to provide access.  It looks like the SDK supports STS but not sure the level of effort to include that with terragrunt.

brikis98:
@mmorianos We are using the [default credentials providers](https://github.com/gruntwork-io/terragrunt/blob/master/dynamodb/dynamo_lock.go#L96) from the AWS Go SDK. Following that code SDK code, it uses the [CredChain function](https://github.com/aws/aws-sdk-go/blob/master/aws/defaults/defaults.go#L36), which includes an [EnvProvider](https://github.com/aws/aws-sdk-go/blob/master/aws/defaults/defaults.go#L91), which, at least in theory, includes the [AWS_SESSION_TOKEN environment variable](https://github.com/aws/aws-sdk-go/blob/master/aws/credentials/env_provider.go#L69).

[...]

PR #28, opened 2016-09-13
Use STS to get user information to support using session tokens

ryangardner:
This is in reference to issue #21  

I've tested a build of this locally and it does acquire locks with session tokens

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,15 +121,13 @@ using AWS. We take no responsibility for any charges you may incur.
 To use DynamoDB for locking, you must:
 
 1. Set your AWS credentials in the environment using one of the following options:
-    1. Set your credentials as the environment variables `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`.
+    1. Set your credentials as the environment variables `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` (and also `AWS_SESSION_TOKEN` if using [STS temporary credentials](http://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp.html))
     1. Run `aws configure` and fill in the details it asks for.
     1. Run Terragrunt on an EC2 instance with an IAM Role.
 1. Your AWS user must have an [IAM 
    policy](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/access-control-identity-based.html) 
    granting all DynamoDB actions (`dynamodb:*`) on the table `terragrunt_locks` (see the
-   [DynamoDB locking configuration](#dynamodb-locking-configuration) for how to configure this table name). In 
-   addition, IAM users will need the `iam:GetUser` permission on themselves so that DynamoDB can record which IAM
-   User wrote the most recent lock.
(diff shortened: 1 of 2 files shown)


ryangardner:
If the STS UserId from the `getCallerIdentity` call is not a 1:1 match for the username that came back from `iam.getUser`, some variables that are named "username" should probably be renamed for clarity purposes

brikis98:
Fantastic, thanks! A few questions:

1. Do all the test pass? You can run them using `go test -v $(glide novendor)`. We've unfortunately had to disable the CircleCI job for PRs as our tests involve AWS creds (see [here for more info](https://circleci.com/docs/fork-pr-builds/)).
1. Is an IAM permission necessary for the `sts:GetCallerIdentity`?

ryangardner:
An iam isn't necessary for `getCallerIdentity` as far as I can tell (there
is no checkbox to enable it in their policy builder tool, but they have one
for everything else. Others who were more involved with the changes to
terraform have said it doesn't require any iams)

I'll run all the tests in a little bit and confirm the work.

ryangardner:
is the s3 bucket used by the tests supposed to be accessible to other accounts?  It is failing with a 403 on the s3 bucket - is there a way for me to set up an s3 bucket on my account that I can use to run the tests?

```
go test -v $(glide novendor)
=== RUN   TestTerragruntWorksWithLocalTerraformVersion
--- FAIL: TestTerragruntWorksWithLocalTerraformVersion (1.00s)
	integration_test.go:30: The S3 Bucket in the .terragrunt file does not exist. S3 Bucket Name = 'gruntwork-terragrunt-tests'. S3 Bucket Region = 'us-west-2'. Full Error Message = AccessDenied: Access Denied
			status code: 403, request id: 3D2637FBC18C2117
FAIL
exit status 1
FAIL	github.com/gruntwork-io/terragrunt/test	1.014s
```

brikis98:
Ah, I think that may be a bad assumption in the tests. Instead, they should probably create the S3 bucket themselves. @josh-padnick Could you take a look?

josh-padnick:
Yeah, to keep things simple, I just assumed the existence of an S3 bucket (naturally in our Gruntwork account), but @brikis98 is correct that it'd be better if the test creates its own S3 bucket. 

That update will involve generating a random name for the S3 bucket, creating it, interpolating the values into the `.terragrunt` file, and deleting the S3 bucket, so if you just want to get passing tests for now, you might change the values in the `const` and the `.terragrunt` file locally only for now, and I'll update these shortly, hopefully tonight or tomorrow.

ryangardner:
I created a bucket and edited the .terragrunt file and and the test to point to the bucket and then the integration test passed.

For these tests I was using my STS credentials - I don't have any non-STS based credentials to test with at the moment. Earlier today I did have a set that I used to test that it still worked. 

One thing that does change from `iam getUser` vs `sts getCallerIdentiy` is that the caller identity returns the userId and not a username - it seems like this should be fine I think for what terragrunt uses it for (an identifier of which user has the lock) but it's worth noting that it is different

brikis98:
Excellent, thanks! Merging now.

ryangardner:
Thanks for the fast turnaround getting this merged!

PR #31, opened 2016-09-26
all: enable simpler lock provider support with more generic config

pmcatominey:
- dynamodb package moved under locks directory
- hcl lock config now in snake_case
- hcl lock config uses same generic style as remote state
- go fmt has fixed formatting on some edited files


config/builtin_locks.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"fmt"
+
+	"github.com/gruntwork-io/terragrunt/locks"
+	"github.com/gruntwork-io/terragrunt/locks/dynamodb"
+)
+
+// lockFactory provides an implementation of Lock with the provided configuration map
+type lockFactory func(map[string]string) (locks.Lock, error)
```
brikis98: Any reason all the lock factory stuff is private?

config/builtin_locks.go
```diff
@@ -0,0 +1,26 @@
...
+
+// lookupLock returns the implementation for the named lock or returns an error if
+// it is not found
+func lookupLock(name string, conf map[string]string) (locks.Lock, error) {
+	f, ok := builtinLocks[name]
```
brikis98: I generally prefer more descriptive variable names than `f` or `ok`. Perhaps `factory` and `containsFactory`?

pmcatominey:
Remaining:

- [ ] make documentation / help text more generic
- [ ] split config tests into separate files

Remote config will also need changing slightly to make it snake_case, probably a job for another PR.

config/builtin_locks.go
```diff
@@ -0,0 +1,26 @@
...
+	if !ok {
+		return nil, fmt.Errorf("no Lock implementation found for %s", name)
```
brikis98: To make it easier to check error cases with automated tests, and to allow clients of the code to make better decisions on how to handle errors, it's a good idea to define custom error types. In this case, a `LockNotFound` type (just a type alias for a `string` with an `Error` method) would make this error easier to handle programmatically. Also, we try to wrap each error with a call to `errors.WithStacktrace` so we can trace down the cause more easily. See https://github.com/gruntwork-io/terragrunt#error-handling for more details.

config/config.go
```diff
@@ -1,34 +1,48 @@
...
+	"github.com/gruntwork-io/terragrunt/locks"
+	"github.com/gruntwork-io/terragrunt/remote"
+	"github.com/hashicorp/hcl"
 )
 
-const TERRAGRUNT_CONFIG_FILE = ".terragrunt"
+const configFilePath = ".terragrunt"
+
+// Config represents a parsed and expanded configuration
+type Config struct {
```
brikis98: Here too I prefer the longer name. There are undoubtedly thousands of types called `Config` in the Go world. Seeing the type elsewhere in the code would tell me nothing about its intent without having to dig into its definition and documentation. The package name (also `config`) isn't much help here either. That's why I used `TerragruntConfig` originally.

config/config.go
```diff
@@ -1,34 +1,48 @@
...
+	Lock        locks.Lock
+	RemoteState *remote.RemoteState
+}
 
-// A common interface with all fields that could be in the .terragrunt config file.
-type TerragruntConfig struct {
-	DynamoDbLock *dynamodb.DynamoDbLock
-	RemoteState  *remote.RemoteState
+// fileConfig represents the configuration supported in the .terragrunt file
+type fileConfig struct {
```
brikis98: `terragruntFileConfig` or even `terragruntConfigFile` may be a slightly clearer name here.

config/config.go
```diff
@@ -1,34 +1,48 @@
...
+// LockConfig represents generic configuration for Lock providers
+type LockConfig struct {
+	Backend string            `json:"backend"`
+	Config  map[string]string `json:"config"`
 }
 
 // Read the Terragrunt config file from its default location
-func ReadTerragruntConfig() (*TerragruntConfig, error) {
-	return parseTerragruntConfigFile(TERRAGRUNT_CONFIG_FILE)
+func Read() (*Config, error) {
```
brikis98: Same comment here about longer names. To a random user, `config.Read()` says very little about what this code does and forces you to dig into the code & docs to understand; `config.ReadTerragruntConfig()` tells you a whole lot more. With auto-complete in modern IDEs, the extra characters don't cost much :)

config/config.go
```diff
@@ -37,26 +51,31 @@ func parseTerragruntConfigFile(configPath string) (*TerragruntConfig, error) {
 }
 
 // Parse the Terragrunt config contained in the given string
-func parseTerragruntConfig(config string) (*TerragruntConfig, error) {
-	terragruntConfig := &TerragruntConfig{}
-
-	if err := hcl.Decode(terragruntConfig, config); err != nil {
+func parseConfigString(configSrc string) (*Config, error) {
+	f := &fileConfig{}
```
brikis98: Same comment on naming here too. You're using `f` here to mean something completely different than the `f` in the `lookupLock` function above. To understand what this is, I'm forced to scroll and read a lot of code. With a clearer name like `terragruntConfigFromFile`, I can guess it from just the name. This makes it not only easier to read the code, but also prevents errors (see [Making Wrong Code Look Wrong](http://www.joelonsoftware.com/articles/Wrong.html)).

config/config.go
```diff
@@ -37,26 +51,31 @@ func parseTerragruntConfigFile(configPath string) (*TerragruntConfig, error) {
...
+	if err := hcl.Decode(f, configSrc); err != nil {
 		return nil, errors.WithStackTrace(err)
 	}
 
-	if terragruntConfig.DynamoDbLock != nil {
-		terragruntConfig.DynamoDbLock.FillDefaults()
-		if err := terragruntConfig.DynamoDbLock.Validate(); err != nil {
-			return nil, err
+	c := &Config{}
```
brikis98: s/c/terragruntConfig/

config/config.go
```diff
@@ -37,26 +51,31 @@ func parseTerragruntConfigFile(configPath string) (*TerragruntConfig, error) {
...
+
+	if lconf := f.Lock; lconf != nil {
```
brikis98: `lconf` made me wonder what a `left config` could be. `lockConfig` would be a clearer name.

config/config.go
```diff
@@ -37,26 +51,31 @@ func parseTerragruntConfigFile(configPath string) (*TerragruntConfig, error) {
...
+		lock, err := lookupLock(lconf.Backend, lconf.Config)
+		if err != nil {
+			return nil, fmt.Errorf("unable to configure lock %s: %s", lconf.Backend, err)
```
brikis98: As mentioned above, please use `errors.WithStackTrace` and a custom error type.

locks/dynamodb/dynamo_lock.go
```diff
@@ -2,24 +2,52 @@ package dynamodb
...
-	MaxLockRetries	int
+	StateFileId    string
+	AwsRegion      string
+	TableName      string
+	MaxLockRetries int
+}
+
+// New is the factory function for DynamoDbLock
+func New(conf map[string]string) (locks.Lock, error) {
+	if conf["state_file_id"] == "" {
```
brikis98: It's probably a good idea to extract all these string constants into `const` variables defined at the top of this file to keep things DRY.

locks/dynamodb/dynamo_lock.go
```diff
@@ -2,24 +2,52 @@ package dynamodb
...
+		StateFileId:    conf["state_file_id"],
+		AwsRegion:      conf["aws_region"],
+		TableName:      conf["table_name"],
+		MaxLockRetries: 0,
+	}
+
+	if confMaxRetries := conf["max_lock_retries"]; confMaxRetries != "" {
+		maxRetries, err := strconv.Atoi(confMaxRetries)
+		if err != nil {
+			return nil, errors.WithStackTrace(fmt.Errorf("unable to parse config value max_lock_retries: %s", err))
```
brikis98: Custom error type

config/builtin_locks.go
pmcatominey: There wasn't a reason to export it, it's only used within this file and package.

config/config.go
pmcatominey: I'm not sure how well a custom error type would work here since the detail is provided as format arguments.

locks/dynamodb/dynamo_lock.go
pmcatominey: Not sure if thats necessary, theres only 2 usages of state_file_id which I'll bring down to one by moving the validation check to after the mapping to the lock type.

config/config.go
brikis98: Actually, in this case, you can probably directly return the underlying `err` object.

locks/dynamodb/dynamo_lock.go
brikis98: You can use an alias:

```go
type InvalidMaxLockRetriesValue error
func (err InvalidMaxLockRetriesValue) Error() string {
  return fmt.Sprintf("unable to parse config value max_lock_retries: %s", err)
}
```

config/config.go
brikis98: That said, if you needed to have a custom type with several formal arguments, you could use a struct:

```go
type CustomErrorExample {
  Foo string
  Bar int
}
func (err CustomErrorExample) Error() string {
  return fmt.Sprintf("foo = %s, bar = %d", err.Foo, err.Bar)
}
```

pmcatominey:
Thanks, I've amended most and left comments on others.

1. Yes this is mostly a refactor to separate some of the code, the aim is to make adding a provider as simple as writing a package and adding an entry to the builtin map.

2. I'll be updating the README once the code is finished.

3. I assumed that the variables were just hidden from the CircleCI output, I'll look into running the tests once I have an AWS setup.

pmcatominey:
@brikis98 I've updated the README examples to the new syntax and ran the full test suite with AWS credentials, everything is now passing.

brikis98:
Excellent, thanks. Merging now.

PR #38, opened 2016-10-03
Fixed var names in README

antonbabenko:
Minor corrections after recent PR was merged.


brikis98:
Ah, good catch. Thank you!

Issue #36, opened 2016-09-28
fatal error: systemstack called from unexpected goroutine


antonbabenko:
I get this error once in a while:

```
fatal error: systemstack called from unexpected goroutine
...
### same line repeated 500+ times ###
...
fatal error: systemstack called from unexpected goroutine
./terraform.sh: line 210: 23108 Segmentation fault: 11  terragrunt get $LAYER_WORK_DIR
```

Rerun of the same command helps.

Here is nice stack trace to review - https://gist.github.com/antonbabenko/87bfdd7c070146d49d1021be424458d4

brikis98:
Hm, strange. Please post the stack trace when you have it so we can track this down!

antonbabenko:
Added link to stack trace to description

brikis98:
Wow, bizarre. If I'm reading your stacktrace correctly, it points to [this line in dynamo_lock_item.go](https://github.com/gruntwork-io/terragrunt/blob/master/locks/dynamodb/dynamo_lock_item.go#L114):

```go
output, err := stsconn.GetCallerIdentity(&sts.GetCallerIdentityInput{})
```

Browsing the stacktrace, it looks like a potential bug in the AWS Go SDK or even the underlying Go HTTP libraries it's using to make API calls.

antonbabenko:
New day, new [crash log](https://gist.github.com/antonbabenko/87bfdd7c070146d49d1021be424458d4#file-crash-log-3-10-2016). Maybe you will see anything strange in that one. Rerun works fine.

josh-padnick:
I independently verified it's the same `GetCallerIdentity()` function that caused the panic and it seems to be happening when the HTTP request itself is sent, so I'm in agreement with Jim that this is most likely an issue witht he AWS Go SDK.

That being said, I notice that we are on a [relatively old release](https://github.com/aws/aws-sdk-go/commit/665c623d7f3e0ee276596b006655ba4dbe0565b0) of the AWS SDK, so I'll update to the latest commit and issue a new release. Once that happens, can you let us know if this happens again?

Thanks for reporting this!

antonbabenko:
@josh-padnick I will be glad to try it out and report as soon as I hit it again.

brikis98:
[...]

PR #39, opened 2016-10-03
Bump all dependency versions, and AWS SDK in particular.

josh-padnick:
One user was reporting random panics when `STS.GetCallerIdentity()` was being called, so hopefully updating to the latest AWS SDK for Golang will resolve the issue. 

BTW, note the new explicit specification of a tag in `glide.yaml`. We should probably do this for all dependencies going forward. 

'cc @brikis98


brikis98:
lgtm, thanks!

josh-padnick:
For record keeping, this PR was in response to #36.

Issue #25, opened 2016-09-05
Add ability to choose .terragrunt file path or specify options on command line


mildred:
I am working on a special environment, and I would like the ability to choose a different `.terragrunt` file than the one on the current directory, or, even better, the ability to specify the options on the command line or environment variables. Currently, I'm automatically generating the `.terragrunt` file.

brikis98:
It should be pretty straightforward to add a `--config` flag to specify a custom location for the `.terragrunt` file. That said, one of the main benefits of Terragrunt is that it provides fairly *transparent* remote state management. That is, you don't need to remember to run a specific command to connect to the right remote data store or to use a lock; you just run `terragrunt plan`, `terragrunt apply`, etc. and things "just work". If you have to remember to provide a `--config` flag every time pointing to the right location, it significantly increases the chances of making a mistake.

Could you describe your use case a bit more? Perhaps there is some alternative solution that does what you want while remaining fairly transparent.

brikis98:
@mildred I haven't heard back in a while, so I'm going to close this for now. But let me know if this is something you want to discuss further and I can re-open the bug.

mildred:
Sorry, I skipped the notification.

The use case is that we use terragrunt to automate infrastructure deployment. Previously we ran terragrunt in an AWS lambda, in which the filesystem is read only. The terraform files are bundled in the lambda and are in a read only directory. The terragrunt configuration must be generated on the fly because we don't want to hardcode the S3 bucket in the lambda.

The problem is that we cannot generate the terragrunt configuration because it is in a read only directory.

Currently, I work around this by going to a temporary directory where I write the `.terragrunt` file, and tell terraform to provision another directory.

mildred:
I also implemented this : https://github.com/squarescale/terragrunt/commit/cbb4a4b74a067186d05e69446e40a2a068c468cc

brikis98:
> The terraform files are bundled in the lambda and are in a read only directory. The terragrunt configuration must be generated on the fly because we don't want to hardcode the S3 bucket in the lambda.

If the terraform files are bundled in the lambda job, why isn't the `.terragrunt` file included too? After all, it should be in the same folder as the terraform files.

mildred:
> If the terraform files are bundled in the lambda job, why isn't the .terragrunt file included too? After all, it should be in the same folder as the terraform files.

[...]

PR #42, opened 2016-10-13
Allow to specify a different .terragrunt config file

mildred:
`--config` on command line or `TERRAGRUNT_CONFIG` environment variable

For issue #25

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"fmt"
+	"os"
 	"regexp"
 
 	"github.com/gruntwork-io/terragrunt/config"
@@ -61,6 +62,19 @@ func CreateTerragruntCli(version string) *cli.App {
    Moreover, for the apply and destroy commands, Terragrunt will first try to acquire a lock using DynamoDB. For
    documentation, see https://github.com/gruntwork-io/terragrunt/.`
 
+	var defaultConfigFilePath = config.ConfigFilePath
+	if os.Getenv("TERRAGRUNT_CONFIG") != "" {
+		defaultConfigFilePath = os.Getenv("TERRAGRUNT_CONFIG")
+	}
+
+	app.Flags = []cli.Flag{
+		cli.StringFlag{
+			Name:  "terragrunt-config",
+			Value: defaultConfigFilePath,
+			Usage: ".terragrunt file to use",
+		},
+	}
+
 	return app
 }
 
@@ -75,7 +89,7 @@ func runApp(cliContext *cli.Context) (finalErr error) {
 		return nil
 	}
 
-	conf, err := config.ReadTerragruntConfig()
+	conf, err := config.ReadTerragruntConfig(cliContext.String("terragrunt-config"))
 	if err != nil {
 		return err
 	}
--- a/config/config.go
+++ b/config/config.go
@@ -9,7 +9,7 @@ import (
 	"github.com/hashicorp/hcl"
 )
 
-const configFilePath = ".terragrunt"
+const ConfigFilePath = ".terragrunt"
 
 // TerragruntConfig represents a parsed and expanded configuration
 type TerragruntConfig struct {
@@ -30,8 +30,8 @@ type LockConfig struct {
 }
 
 // ReadTerragruntConfig the Terragrunt config file from its default location
-func ReadTerragruntConfig() (*TerragruntConfig, error) {
-	return parseConfigFile(configFilePath)
+func ReadTerragruntConfig(filePath string) (*TerragruntConfig, error) {
+	return parseConfigFile(filePath)
 }
 
 // Parse the Terragrunt config file at the given path


cli/cli_app.go
```diff
@@ -61,6 +62,14 @@ func CreateTerragruntCli(version string) *cli.App {
    Moreover, for the apply and destroy commands, Terragrunt will first try to acquire a lock using DynamoDB. For
    documentation, see https://github.com/gruntwork-io/terragrunt/.`
 
+	app.Flags = []cli.Flag{
+		cli.StringFlag{
+			Name:  "config",
+			Value: os.Getenv("TERRAGRUNT_CONFIG"),
```
brikis98: The default value should fall back to `".terragrunt"` if that env var is not set.

config/config.go
```diff
@@ -30,8 +30,11 @@ type LockConfig struct {
 }
 
 // ReadTerragruntConfig the Terragrunt config file from its default location
-func ReadTerragruntConfig() (*TerragruntConfig, error) {
-	return parseConfigFile(configFilePath)
+func ReadTerragruntConfig(filePath string) (*TerragruntConfig, error) {
+	if filePath == "" {
+		filePath = configFilePath
+	}
```
brikis98: The CLI portion of the app can be responsible for setting the `filePath` to the default value (`.terragrunt`). If you do that, then you can merge the `parseConfigFile` method and `ReadTerragruntConfig` into one, since they would do the same thing.

cli/cli_app.go
```diff
@@ -61,6 +62,14 @@ func CreateTerragruntCli(version string) *cli.App {
    Moreover, for the apply and destroy commands, Terragrunt will first try to acquire a lock using DynamoDB. For
    documentation, see https://github.com/gruntwork-io/terragrunt/.`
 
+	app.Flags = []cli.Flag{
+		cli.StringFlag{
+			Name:  "config",
```
brikis98: Do any of the existing Terraform commands expect a `--config` flag? Even if it doesn't now, it may in the future, so to avoid a conflict, we should probably pick a less generic name (e.g. `--terragrunt-config` or `--tc` for short).

cli/cli_app.go
```diff
@@ -61,6 +62,14 @@ func CreateTerragruntCli(version string) *cli.App {
    Moreover, for the apply and destroy commands, Terragrunt will first try to acquire a lock using DynamoDB. For
    documentation, see https://github.com/gruntwork-io/terragrunt/.`
 
+	app.Flags = []cli.Flag{
```
brikis98: Note: the `runTerraformCommand` method passes through all arguments directly to Terraform. You probably will need to strip this one out to avoid errors.

cli/cli_app.go
mgagliardo: Hi there, been following this thread for a while. My personal opinion, I think it'd be much better if instead of setting an ENV var, you could do something like

`terragrunt apply my/path/to/terragrunt/`

Which would be more or less the same as:
`terraform apply my/dir`

brikis98: I agree that would be nice from a consistency standpoint, but that sort of command would conflict with the underlying terraform command. The `terraform apply` command accepts a path to a saved plan file, so if someone calls `terragrunt apply foo`, we wouldn't know if foo is a Terragrunt config file or a Terraform plan file.

brikis98: Not sure I follow. I was thinking of something like:

```go
defaultConfigPath := ".terragrunt"
if os.Getenv(TERRAGRUNT_CONFIG_ENV_VAR) != "" {
  defaultConfigPath = os.Getenv(TERRAGRUNT_CONFIG_ENV_VAR)
}

app.Flags = []cli.Flag{
		cli.StringFlag{
			Name:  "config",
			Value: defaultConfigPath,
		}
}
```

mgagliardo: My project needs support for many regions, also working directly on the cli executing commands like 
`
terraform apply
terraform plan
`

Makes nosense when your project starts getting multi-regional, so you can have something like:
`
$ tree -L 1
|-- Makefile
|-- README.md
|-- Rakefile
|-- Terrafile
|-- aws_region (each folder with its own environment_vars.tf file)
   |_ ireland
   |_ virginia
   |_ oregon
   |_ ........
|-- envs
   |_ bastion.tf
   |_ puppetmaster.tf
   |_ elasticsearch.tf
|-- global_variables.tf
`

And you can have a makefile that builds your infra depending on your ENV:

Makefile:
`
TF=terraform
# TF=terragrunt

TF_ENV ?= ireland
AWS_REGION ?= eu-west-1
TF_VAR_aws_account_id=`aws sts get-caller-identity --output text | cut -f1`

TF_VARS=TF_VAR_aws_account_id=$(TF_VAR_aws_account_id)
TF_ARGS=aws_region/$(TF_ENV)

apply:
	$(TF_VARS) $(TF) apply $(TF_ARGS)
`

brikis98: The code snippet I have above will find the `.terragrunt` file as follows:

1. A path explicitly specified via the `--terragrunt-config` option.
1. A path specified via the `TERRAGRUNT_CONFIG` environment variable.
1. Finally, fallback to the default of `.terragrunt` in the current directory.

Doesn't that support your use case?

mgagliardo: That is more than enough for me, thanks! :)

mildred: > The default value should fall back to ".terragrunt" if that env var is not set.

Actually, the empty string is replaced by `.terragrunt` later on.

config/config.go
mildred: ok, will do that

cli/cli_app.go
mildred: I think the arguments have been eaten by the flag parsing library, I saw no errors in terraform.

mildred: I also saw logs like `[terragrunt] 2016/10/14 11:45:07 Running command:` without the `--config` flag

mildred:
I just ran the tests (`go test ./...`), and the only error I see is:

```
?   	github.com/gruntwork-io/terragrunt/cli	[no test files]
ok  	github.com/gruntwork-io/terragrunt/config	0.005s
?   	github.com/gruntwork-io/terragrunt/errors	[no test files]
ok  	github.com/gruntwork-io/terragrunt/locks	0.007s
ok  	github.com/gruntwork-io/terragrunt/locks/dynamodb	71.690s
ok  	github.com/gruntwork-io/terragrunt/remote	0.004s
?   	github.com/gruntwork-io/terragrunt/shell	[no test files]
--- FAIL: TestTerragruntWorksWithLocalTerraformVersion (2.16s)
	integration_test.go:30: The S3 Bucket in the .terragrunt file does not exist. S3 Bucket Name = 'gruntwork-terragrunt-tests'. S3 Bucket Region = 'us-west-2'. Full Error Message = AccessDenied: Access Denied
			status code: 403, request id: A54616C59DE83041
FAIL
FAIL	github.com/gruntwork-io/terragrunt/test	2.166s
?   	github.com/gruntwork-io/terragrunt/util	[no test files]
```

I suppose I did miss some environment to run the tests.

brikis98:
@mildred The test failure is not your fault. 

@josh-padnick It looks like the integration test hard-codes the name of the S3 bucket. I thought we saw this before and fixed it, but perhaps I'm imagining things. What's the best way to update this code so other users can run the tests?

josh-padnick:
@brikis98 We did see this before but agreed to defer it for now. The problem is that we'd have to:

- Create a randomly named S3 Bucket
- Generate a new `.terragrunt` file with the new S3 Bucket Name
- Run the test
- Delete all files in the S3 Bucket
- Delete the S3 Bucket

Not hard, just laborious. Should we create an issue for this?

@mildred In the meantime, here's how you can manually validate the tests. I would make these changes locally but do *not* commit them:
- Update the [.terragrunt file](https://github.com/gruntwork-io/terragrunt/blob/master/test/fixture/.terragrunt#L17-L19) to specify an S3 Bucket you do have access to.
- Update the [const values](https://github.com/gruntwork-io/terragrunt/blob/master/test/integration_test.go#L19-L20) in `integration_test.go`.

If the tests pass, submit your PR, our CI job's tests should pass with the S3 bucket name as listed, and then we can merge.

Sorry you have to deal with this, we should definitely fix this so contributors don't run into these kinds of issues.

brikis98:
@josh-padnick Thanks for the context. I created #43 to track this. We should probably fix it fairly soon.

@mildred Please re-run the test as @josh-padnick indicated. If it passes, I think this is good to go!

mildred:
I ran the tests again on commit 8739c3c74b02dc4facc898348f0b911c0d26fc07 and they passed.

brikis98:
Excellent. Merging now. Thank you!

brikis98:
OK, just cut a new release, and once the build completes (assuming no tests fail), the new binaries will show up here: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.1.2

brikis98:
And the build failed :)

The cause is due to our CI scripts. I'll get a fix out shortly.

brikis98:
OK, I fixed the build failure and just recreated the new release, so the binaries *should* show up in a few minutes: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.1.2

PR #52, opened 2016-11-15
lock when using refresh command

### LINKED ISSUE #51: Locking on terraform refresh
According to the [documentation](https://www.terraform.io/docs/commands/refresh.html):

> The terraform refresh command is used to reconcile the state Terraform knows about (via its state file) with the real-world infrastructure. This can be used to detect any drift from the last-known state, and to update the state file.
> 
> This does not modify infrastructure, but does modify the state file. If the state is changed, this may cause changes to occur during the next plan or apply.

If someone is running an apply with terragrunt, another user could run refresh and potentially write a broken state in s3.

marcoamorales:
fixes #51

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ terragrunt destroy
 
 Terragrunt forwards almost all commands, arguments, and options directly to Terraform, using whatever version of
 Terraform you already have installed. However, before running Terraform, Terragrunt will ensure your remote state is
-configured according to the settings in the `.terragrunt` file. Moreover, for the `apply` and `destroy` commands,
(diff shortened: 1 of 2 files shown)


brikis98:
LGTM, thank you! Merging now.

brikis98:
I just created a new release: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.1.4. Assuming the tests pass, the new binaries will show up at that URL in a few minutes.

brikis98:
Haha, the build, of course, failed. The cause was that Terraform has been changing their version numbers back and forth, so the build failed to download the latest Terraform binary. I just committed a fix and recreated the v0.1.4 release. Hopefully, in a few more minutes, the new binaries will be there :)

PR #54, opened 2016-11-22
Ensure Terragrunt picks up changes to remote config

### LINKED ISSUE #15: Changes to .terragrunt file aren't picked up after initial terragrunt run.
Currently, terragrunt configures your localdev's terraform remote state with [this code](https://github.com/gruntwork-io/terragrunt/blob/master/remote/remote_state.go#L48-L63).

The problem is that this doesn't guarantee that the contents of the `.terragrunt` file match the current remote state configuration. This particularly becomes an issue when I've made a mistake in the `.terragrunt` file, correct it, and re-run terragrunt expecting terragrunt to pick up the change.

We should consider creating a hash of the `.terragrunt` file and writing it to the `.terraform` folder so that terragrunt can check whether any value in the `.terragrunt` file has changed, and if so, re-execute `terraform remote config`.

brikis98:
This PR fixes #15. Terragrunt now compares all settings of the remote configuration between what’s stored in the `.tfstate` file and in `.terragrunt`. If any of those settings don’t match up, it prompts the user whether it should re-run `terraform remote config`.

To make this code testable, I’ve also added a new `--terragrunt-non-interactive` flag that can be used to disable user prompts when running Terragrunt in an automated setting (e.g. in an automated test or script).

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,16 @@ remote_state = {
   different key/value pairs, so consult the [Terraform remote state docs](https://www.terraform.io/docs/state/remote/)
   for details.
 
+## CLI Options
+
+Terragrunt forwards all arguments and options to Terraform. The only exceptions are the options that start with the
+prefix `--terragrunt-`. The currently available options are:
+
+* `--terragrunt-config`: A custom path to the `.terragrunt` file. May also be specified via the `TERRAGRUNT_CONFIG`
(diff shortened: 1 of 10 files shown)


brikis98:
@josh-padnick I'm going to merge this one in to unblock myself for a few more PRs. Feedback welcome when you have time.

Issue #43, opened 2016-10-18
integration_test.go should generate the S3 bucket at runtime


brikis98:
Currently, the bucket is hard-coded to an S3 bucket in the Gruntwork AWS account. This means no contributors can run the test without making (non-obvious) changes locally.

To fix this issue, the test should be updated to:
- Create a randomly named S3 Bucket
- Generate a new .terragrunt file with the new S3 Bucket Name
- Run the test
- Delete all files in the S3 Bucket
- Delete the S3 Bucket

josh-padnick:
In the meantime, here's how you can manually validate the tests. I would make these changes locally but do *not* commit them:
- Update the [.terragrunt file](https://github.com/gruntwork-io/terragrunt/blob/master/test/fixture/.terragrunt#L17-L19) to specify an S3 Bucket you do have access to.
- Update the [const values](https://github.com/gruntwork-io/terragrunt/blob/master/test/integration_test.go#L19-L20) in `integration_test.go` with your S3 Bucket name and region.

If the tests pass, submit your PR, our CI job's tests should pass with the S3 bucket name as listed, and then we can merge.

brikis98:
Fixed via #55.

PR #55, opened 2016-11-22
Terragrunt will now create remote state S3 bucket if it doesn't exist

### LINKED ISSUE #35: terragrunt should create s3 bucket, if it is not present
Similar to how it verifies existence of dynamodb lock table and create it if it does not exist.

I think it should create versioned S3 bucket (default, `acl=private`) when `remote_state` section is defined.

brikis98:
I’ve also updated the integration test to use this functionality so the
S3 bucket we use at test time is not hard-coded.

Fixes #35 and #43.

I also worked around a bug/limitation in urfave CLI where command-line flags passed in after the name of a command (e.g. `terragrunt apply --foo`) are not parsed, at all.

### DIFF
--- a/aws_helper/config.go
+++ b/aws_helper/config.go
@@ -1,0 +1,19 @@
+package aws_helper
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/defaults"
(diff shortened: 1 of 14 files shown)


brikis98:
@josh-padnick Going to merge this so I can keep chugging along. Feedback welcome!

PR #59, opened 2016-11-25
Support inheritance and interpolation in .terragrunt files

### LINKED ISSUE #26: Interpolation inside .terragrunt file
We have a bunch of different Terraform state files, each handling .tf files in different directories. There's a `.terragrunt` file in each directory that looks something like:

```
dynamoDbLock = { stateFileId = "my_project_terraform_db_lock" }

remoteState = {
  backend = "s3"
  backendConfigs = {
    encrypt = "true"
    bucket = "my-bucket"
    key = "terragrunt/<dirname>.tfstate"
    region = "us-east-1"
  }
}
```

I'd really like to be able to use a single .terragrunt file for all the directories and be able do something like:

```
# Skip a bunch of common code shown above
remoteState = {
  backendConfigs = {
    key = "terragrunt/${dirname()}.tfstate"
  }
}
```

That way, I can symlink to a common `.terragrunt` in the parent directory and be assured I'm not screwing anything up when I copy the `.terragrunt` file into a new directory.

brikis98:
This PR adds support for two new features to Terragrunt:

1. Inheritance. A `.terragrunt` file can specify a `parent` configuration from which to inherit all settings.
1. Interpolation. Terragrunt now supports two basic interpolation functions: `find_in_parent_folders`, which can be used to automatically find a `.terragrunt` file in a parent folder rather than manually specifying the path, and `path_relative_to_parent`, which returns the relative path between a child `.terragrunt` file and a parent, which is useful for automatically populating `state_file_id` and `key` based on folder structure.

Check out the README for full details. Here is a [direct link to the branch](https://github.com/gruntwork-io/terragrunt/tree/inherit#managing-multiple-terragrunt-files), as that's easier to read than a diff.

This fixes #26.

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,15 +319,15 @@ similar-looking `.terragrunt` files?
 
 The solution is to use the following features of Terragrunt:
 
-* Inheritance
+* Includes
 * Find parent helper
 * Relative path helper
-* Overriding settings from the parent
+* Overriding included settings
 
-### Inheritance
+### Includes
 
-One `.terragrunt` file can "inherit" from another `.terragrunt` file. For example, imagine you have the following file
-layout:
+One `.terragrunt` file can automatically "include" the contents of another `.terragrunt` file using the `include` 
+block. For example, imagine you have the following file layout:
 
 ```
 my-terraform-repo
@@ -347,37 +347,40 @@ my-terraform-repo
 ```
 
 The `.terragrunt` file in the root folder defines the typical `lock` and `remote_state` settings. The `.terragrunt` 
-files in all the subfolders (e.g. `qa/my-app/.terragrunt`) can automatically reuse ("inherit") all the settings from
-a parent file using the `parent` block:
+files in all the subfolders (e.g. `qa/my-app/.terragrunt`) can automatically include all the settings from a parent 
+file using the `include` block:
 
 ```hcl
-parent = {
+include = {
   path = "../../.terragrunt"
 }
 ```
 
+When you run Terragrunt, it will see the `include` block and load the contents of the root `.terragrunt` file, almost
+as if you had copy/pasted the contents of the root `.terragrunt` file into `qa/my-app/.terragrunt`.
+
 There are a few problems with the simple approach above, so read on before using it!
 
-1. Having to manually manage the file paths to the parent folder is tedious. To solve this problem, you can use the 
-   `find_in_parent_folders` helper. 
-1. If the parent `.terragrunt` file hard-codes the `state_file_id` and `key` settings, then every child would rely on
-   the same lock and write state to the same location. To avoid this problem, you can use the `path_relative_to_parent`
-   helper.
-1. Some of the child `.terragrunt` files may want to override settings from the parent. To do this, see the section
-   on overriding settings from the parent.
+1. Having to manually manage the file paths to the included `.terragrunt` file is tedious and error prone. To solve 
+   this problem, you can use the `find_in_parent_folders()` helper. 
+1. If the included `.terragrunt` file hard-codes the `state_file_id` and `key` settings, then every child that includes
+   it would end up using the same lock and write state to the same location. To avoid this problem, you can use the 
+   `path_relative_to_include()` helper.
+1. Some of the child `.terragrunt` files may want to override the settings they include. To do this, see the section
+   on overriding included settings.
 
 Each of these items is discussed next.
 
 ### find_in_parent_folders helper
 
 Terragrunt supports the use of a few helper functions using the same syntax as Terraform: `${some_function()}`. One of
-the supported helper functions is `find_in_parent_folders`, which returns the path to the first `.terragrunt` file it
+the supported helper functions is `find_in_parent_folders()`, which returns the path to the first `.terragrunt` file it
 finds in the parent folders above the current `.terragrunt` file. 
 
 Example:
 
 ```hcl
-parent = {
+include = {
   path = "${find_in_parent_folders()}"
 }
 ```
@@ -386,10 +389,17 @@ If you ran this in `qa/my-app/.terragrunt`, this would automatically set `path` 
 always want to use this function, as it allows you to copy/paste the same `.terragrunt` file to all child folders with
 no changes.
 
-### path_relative_to_parent helper
+`find_in_parent_folders()` will search up the directory tree until it hits the root folder of your file system, and if
+no `.terragrunt` file is found, Terragrunt will exit with an error.
 
-Another helper function supported by Terragrunt is `path_relative_to_parent`, which returns the relative path between 
-the current `.terragrunt` file and the path specified in its `parent` block. For example, in the root `.terragrunt` 
+**Note**: only one level of includes is allowed. If  `foo/bar/.terragrunt` includes `foo/.terragrunt`, then 
+`foo/.terragrunt` may NOT itself specify an `include` block. Otherwise, reasoning about helper functions such as 
+`path_relative_to_include()` gets too complicated.
+
+### path_relative_to_include helper
+
+Another helper function supported by Terragrunt is `path_relative_to_include()`, which returns the relative path between 
+the current `.terragrunt` file and the path specified in its `include` block. For example, in the root `.terragrunt` 
 file, you could do the following:
  
 ```hcl
@@ -397,7 +407,7 @@ file, you could do the following:
 lock = {
   backend = "dynamodb"
   config {
-    state_file_id = "${path_relative_to_parent()}"
+    state_file_id = "${path_relative_to_include()}"
   }
 }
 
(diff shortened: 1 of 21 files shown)


README.md
```diff
@@ -242,6 +242,214 @@ remote_state = {
...
+  }
+}
+``` 
+
+The result is that when you run `terragrunt` commands in the `qa/my-app` folder, you get the `lock` settings from the 
+parent, but the `remote_state` settings of the child. 
+
+**Note**: only one level of inheritance is allowed. If `foo/.terragrunt` is specified as the `parent` of 
+`foo/bar/.terragrunt`, then `foo/.terragrunt` may NOT itself specify a `parent`. Otherwise, reasoning about helper
+functions such as `path_relative_to_parent` gets too complicated.
```
josh-padnick: Ok, thanks for clarifying the "chaining" question.

josh-padnick: This looks great overall, but I am left a little confused on the relationship between each of the interpolation functions and whether they are in a parent or child `.terragrunt` file. If I was reading your example correctly, the interpolation function itself, not its value, was inherited by child `.terragrunt` files, which is an awkward concept. Wouldn't it make more sense for the parent to simply omit the property you want each child to specify? In fact, you could even specify a value like `path = "${get_from_child()}"` to make explicit that this value should be specified by a child (though this probably isn't necessary).

Intuitively, I expect all interpolation functions to be resolved in the template they're in, and only the static value to be inherited. 

Also, given that we only support one level of inheritance (a good decision IMO), should we replace the concept of a "parent" folder with a "root" folder to make this more explicit?

config/config.go
```diff
@@ -34,40 +40,119 @@ type LockConfig struct {
...
 	util.Logger.Printf("Reading Terragrunt config file at %s", terragruntOptions.TerragruntConfigPath)
-	return parseConfigFile(terragruntOptions.TerragruntConfigPath)
+	return parseConfigFile(terragruntOptions.TerragruntConfigPath, terragruntOptions, nil)
 }
 
-// Parse the Terragrunt config file at the given path
-func parseConfigFile(configPath string) (*TerragruntConfig, error) {
-	bytes, err := ioutil.ReadFile(configPath)
+// Parse the Terragrunt config file at the given path. If parent is specified, then treat this config is a parent
+// config of some other file when resolving relative paths.
```
josh-padnick: - If **no**(?) parent is specified
- treat this config **as** a parent

config/config_helpers.go
```diff
@@ -0,0 +1,115 @@
...
+func findInParentFolders(terragruntOptions *options.TerragruntOptions) (string, error) {
+	previousDir := filepath.Dir(terragruntOptions.TerragruntConfigPath)
+
+	// To avoid getting into an accidental infinite loop (e.g. do to cyclical symlinks), set a max on the number of
+	// parent folders we'll check
+	for i := 0; i < MAX_PARENT_FOLDERS_TO_CHECK; i++ {
+		currentDir := filepath.Dir(previousDir)
+		if currentDir == previousDir {
+			return "", errors.WithStackTrace(ParentTerragruntConfigNotFound(terragruntOptions.TerragruntConfigPath))
+		}
```
josh-padnick: This cycle check will miss any symlink that isn't in the immediate parent folder. Should you instead maintain a slice (or map) of paths and check against that?

README.md
```diff
@@ -242,6 +242,214 @@ remote_state = {
...
+
+The `.terragrunt` file in the root folder defines the typical `lock` and `remote_state` settings. The `.terragrunt` 
+files in all the subfolders (e.g. `qa/my-app/.terragrunt`) can automatically reuse ("inherit") all the settings from
+a parent file using the `parent` block:
+
+```hcl
+parent = {
+  path = "../../.terragrunt"
+}
+```
```
brikis98: It will say "Error reading file at path XXX"

README.md
```diff
@@ -242,6 +242,214 @@ remote_state = {
...
+
+```hcl
+parent = {
+  path = "${find_in_parent_folders()}"
+}
+```
+
+If you ran this in `qa/my-app/.terragrunt`, this would automatically set `path` to `../../.terragrunt`. You will almost
+always want to use this function, as it allows you to copy/paste the same `.terragrunt` file to all child folders with
+no changes.
```
brikis98: > At what point does find_in_parent_folders() stop searching parent folders?

`find_in_parent_folders()` will search up the directory tree until it hits the root folder. If no `.terragrunt` file is found in a root folder, Terragrunt will exit with an error. I added this to the docs.

> Is "chaining" an envisioned use case? 

No. I think users would find it hard to reason about chaining. This is discussed further down in the docs. 

> Is there a use case where it practically makes sense to nest .terragrunt files?

For now, a `.terragrunt` file can include one other `.terragrunt` file. That's it. No chaining, no multiple levels of nesting, nothing like that. The goal is to find a balance between "super simple and purely declarative" and "avoid too much copy and paste".

> In your example folder structure above whose root is my-terraform-repo, there's a .terragrunt file at the top that doesn't correspond to a specific terraform template and can't be used without a child .terragrunt file.

Well, we have two options. One is to move to a `.terragrunt` format that explicitly identifies itself as a root, but that feels like more syntax to manage and learn. The other is to make it so that a root `.terragrunt` file *could* be used with its own Terraform templates, just like any other `.terragrunt` file. The `path_relative_to_parent` helper would just need to return something reasonable (e.g. "." or "root")  when running `terragrunt` against the parent itself. I think the latter approach would be more sound. It also leaves the door open for chaining or other fancier behavior if we ever needed it.

> parent_terragrunt_file_path()
> root_terragrunt_file_path()

What's the difference between these?

> Finally, would the function name parent_terragrunt_file_path() be more clear than find_in_parent_folders()? The latter doesn't really tell you what will be found in parent folders.

They both have pros and cons. The `parent_terragrunt_file_path` doesn't make it clear where that parent comes from. The `find_in_parent_folders` makes it clear this will search, but not what it's searching for. I suppose `find_terragrunt_file_in_parent_folders` is the most accurate name, but it feels a bit too long...

brikis98: Actually, `path_relative_to_parent` already returns "." if you're in the parent folder. So, as long as you are aware of that, you can already use the root `.terragrunt` file with its own Terraform templates, if you were so inclined.

README.md
```diff
@@ -242,6 +242,214 @@ remote_state = {
...
+Another helper function supported by Terragrunt is `path_relative_to_parent`, which returns the relative path between 
+the current `.terragrunt` file and the path specified in its `parent` block. For example, in the root `.terragrunt` 
+file, you could do the following:
+ 
+```hcl
+# Configure Terragrunt to use DynamoDB for locking
+lock = {
+  backend = "dynamodb"
+  config {
+    state_file_id = "${path_relative_to_parent()}"
```
brikis98: You know, now that I read your confusion, I'm realizing that "inheritance" is probably not a great term. I think a better term is "include". All that I'm doing is I'm including all the content of a parent `.terragrunt` file in the current `.terragrunt` file, just as if that content was copied & pasted. I'll update the docs to try to explain this more clearly.

README.md
```diff
@@ -242,6 +242,214 @@ remote_state = {
...
+``` 
+
+Using the configuration above, each child `.terragrunt` file will get a unique path for its `state_file_id` and `key` 
+settings. For example, in `qa/my-app/.terragrunt`, the `state_file_id` will resolve to `qa/my-app` and the `key` will
+resolve to `qa/my-app/terraform.tfstate`.  
+
+You will almost always want to use this helper too. The only time you may want to specify the `state_file_id` or `key` 
+manually is if you moved a child folder. In that case, to ensure it can reuse its old state and lock, you may want to 
+hard-code the `state_file_id` and `key` to the old file path. However, a safer approach would be to move the state 
+files themselves to match the new location of the child folder, as that makes things more consistent!
```
brikis98: Seems like a decent tool. File a bug so we can keep track of this and do it in a follow-up PR.

README.md
brikis98: > Intuitively, I expect all interpolation functions to be resolved in the template they're in, and only the static value to be inherited.

This is exactly what's happening. I've renamed the `parent` block to an `include` block to try to make it clearer. You are just included the contents of a single other `.terragrunt` file, and that's it. After that, it's exactly like you had copy/pasted those static values into your own `.terragrunt` file, and that's where they are evaluated. That's why the calls to `path_relative_to_include` (what used to be called `path_relative_to_parent`) can live in the parent file.

config/config.go
brikis98: No, I had it right :)

I updated the text to try to clarify things.

config/config_helpers.go
brikis98: Not sure I follow. Can you give an example?

brikis98:
OK, merging.

brikis98:
OK, merging

PR #60, opened 2016-11-30
Add spin-up and tear-down commands

### LINKED ISSUE #45: Add the ability to automatically walk the dependency tree of modules
When following [Terraform state management best practices](https://blog.gruntwork.io/how-to-manage-terraform-state-28f5697e68fa), you often end up with the all the components of a single environment broken into many folders. For example, here is a typical production environment:

```
prod
  └ vpc
  └ services
      └ frontend-app
      └ backend-app
          └ vars.tf
          └ outputs.tf
          └ main.tf
          └ .terragrunt
  └ data-storage
      └ mysql
      └ redis
```

The reason we break out these components into separate folders rather than defining them all in one set of templates provides is to ensure *isolation*. That way, when you're making changes to one component (e.g. deploying the frontend-app), you aren't risking breaking all the others (e.g. breaking the VPC, data storage, etc). 

The question is, how do you "spin up" an entire environment without having to manually go into each folder and run `terragrunt apply`? 

One option is to add a command like `terragrunt spin-up` which could find all the Terraform modules in the folder structure (based on the presence of a `.tf` files), parse the code in those modules to figure out the dependencies between them (that is, those dependencies explicitly specified via the usage of the [terraform_remote_state resource](https://www.terraform.io/docs/providers/terraform/d/remote_state.html)), and automatically call `terragrunt apply` on that dependency tree in the right order. There could be an analogous `terragrunt spin-down` command too. If all the modules contain `.terragrunt` files (or there is a root `.terragrunt` file if we implement https://github.com/gruntwork-io/terragrunt/issues/26#issuecomment-260665542), then Terragrunt would know how they manage remote state, and could therefore match the remote state configuration against the usage of the `terraform_remote_state` resource within each module.

This would work as long as all dependencies between modules are specified via the `terraform_remote_state` resource. If there are an implicit dependencies—for example, if the `iam` module creates an IAM role and the `kms` module has the name of that IAM role copy/pasted and just assumes it'll exist—then you may get errors. The good news is that fixing those errors would be very simply: just add a `terraform_remote_state` resource to the `kms` module that points to the remote state of the `iam` module. In other words, adding a `terraform_remote_state` resource would be like adding a `depends_on` parameter.

There is also the tricky question of what to do if Terragrunt hits an error part of the way through spinning up an environment. Probably the right thing to do is to exit, just as Terraform does. If you fix the issue and re-run `terragrunt spin-up`, you'd probably lose some time running `terraform apply` on modules that have already been applied, but since there won't be any actual changes to apply, that shouldn't take too long.

brikis98:
This PR adds `spin-up` and `tear-down` commands to Terragrunt that can be used to run `terragrunt apply` or `terragrunt destroy`, respectively, in a bunch of subfolders in parallel. This way, even if you have your Terraform code defined in separate modules to get isolation, you can still deploy or undeploy an entire environment in a single command.

To manage dependencies between modules (e.g. VPC must be deployed before an app), the `.terragrunt` config now lets you specify a `dependencies` block with a list of paths that must be deployed before the current module.

This was quite a bit more work than I originally expected. In particular, it took a while to write tests for this, and it’s good that I did, as they flushed out a huge number of very subtle bugs (and more may be lurking!).

This fixes #45. It’s an incremental solution for it, anyway.


spin/graph.go
```diff
@@ -0,0 +1,52 @@
+package spin
```
brikis98: All the interesting code in this PR is in the `spin` package. Since this is a massive PR due to all the logging refactoring, I'd recommend starting here!

spin/test_helpers.go
```diff
@@ -0,0 +1,190 @@
...
+	for expectedPath, expectedModule := range expectedModules {
+		actualModule, containsModule := actualModules[expectedPath]
+		if assert.True(t, containsModule, messageAndArgs...) {
+			assertRunningModulesEqual(t, expectedModule, actualModule, doDeepCheck, messageAndArgs...)
+		}
+	}
+}
+
+// We can't use assert.Equals on TerraformModule or any data structure that contains it (e.g. runningModule) because it
+// contains some fields (e.g. TerragruntOptions) that cannot be compared directly
```
brikis98: I know I say it a lot, but I spent *hours* in this PR working around limitations in Go, so I'll say it again: it's a shitty language.

brikis98:
@josh-padnick Please review this when you get a chance. Mainly focus on the `spin` package and the `README`, as much of the rest of the diff is less interesting refactoring (and a TON of test code).

README.md
```diff
@@ -315,14 +315,17 @@ remote_state = {
...
+similar-looking `.terragrunt` files? Also, if you want to spin up an entire environment (e.g. `stage`, `prod`), how do
+you do it without having to manually run `terragrunt apply` in each of the Terraform folders within that environment?
 
 The solution is to use the following features of Terragrunt:
 
 * Includes
 * Find parent helper
 * Relative path helper
 * Overriding included settings
+* The spin-up and tear-down commands
```
josh-padnick: Surround commands with `.

README.md
```diff
@@ -456,6 +459,129 @@ remote_state = {
...
+```hcl
+include = {
+  path = "${find_in_parent_folders()}"
+}
+
+dependencies = {
+  paths = ["../vpc", "../backend-app"]
+}
+```
+
```
josh-padnick: Do we have a way to detect dependency cycles? I think so, but double-checking.

README.md
```diff
@@ -456,6 +459,129 @@ remote_state = {
...
+Once you've specified the depenedencies in each `.terragrunt` file, when you run the `terragrunt spin-up` and 
+`terragrunt tear-down`, Terragrunt will ensure that the dependencies are applied or destroyed, respectively, in the
+correct order. For the example at the start of this section, the order for the `spin-up` command would be:
+
+1. Deploy the VPC
+1. Deploy MySQL and Redis in parallel
+1. Deploy the backend-app
+1. Deploy the frontend-app and search-app in parallel
+
+If any of the modules fail to deploy, then Terragrunt will not attempt to deploy the modules that depend on them.
```
josh-padnick: We should explicitly give advice on what to do if `spin-up` fails midway through running. That advice is probably just "fix the problem, and run again."

README.md
```diff
@@ -465,6 +591,8 @@ prefix `--terragrunt-`. The currently available options are:
   environment variable. The default path is `.terragrunt` in the current directory.
 * `--terragrunt-non-interactive`: Don't show interactive user prompts. This will default the answer for all prompts to 
   'yes'. Useful if you need to run Terragrunt in an automated setting (e.g. from a script).  
+* `--terragrunt-working-dir`: Set the directory where Terragrunt should execute the `terraform` command. Default is the
+  current working directory.
```
josh-padnick: This description is now a little ambiguous given the `spin-up` functionality. Is "where Terragrunt should execute the `terraform` command" the "root" folder, or each individual folder. I know it's root, but consider clarifying or rewording.

README.md
josh-padnick: Did we ultimately rule out directly parsing `terraform_remote_state` because it required that we resolve the Terraform template in order to discover the dependency?

cli/args.go
```diff
@@ -0,0 +1,136 @@
...
+func ParseTerragruntOptions(cliContext *cli.Context) (*options.TerragruntOptions, error) {
+	return parseTerragruntOptionsFromArgs(cliContext.Args())
+}
+
+// TODO: replace the urfave CLI library with something else.
+//
+// EXPLANATION: The normal way to parse flags with the urfave CLI library would be to define the flags in the
+// CreateTerragruntCLI method and to read the values of those flags using cliContext.String(...),
+// cliContext.Bool(...), etc. Unfortunately, this does not work here due to a limitation in the urfave
+// CLI library: if the user passes in any "command" whatsever, (e.g. the "apply" in "terragrunt apply"), then
```
josh-padnick: s/whatsever/whatsoever

cli/args.go
```diff
@@ -0,0 +1,136 @@
...
+// any flags that come after it are not parsed (e.g. the "--foo" is not parsed in "terragrunt apply --foo").
+// Therefore, we have to parse options ourselves, which is infuriating. For more details on this limitation,
+// see: https://github.com/urfave/cli/issues/533. For now, our workaround is to dumbly loop over the arguments
+// and look for the ones we need, but in the future, we should change to a different CLI library to avoid this
+// limitation.
```
josh-padnick: +1 for https://github.com/mitchellh/cli. Although it's a little clunkier to use than urfave/cli, I believe all the HashiCorp tools are based on this library and it does a nice job of creating a pleasant end-user experience (including colors, nice formatting/tabs, etc.)

cli/args.go
```diff
@@ -0,0 +1,136 @@
...
+	for _, arg := range args {
+		if arg == fmt.Sprintf("--%s", argName) {
+			return true
+		}
+	}
+	return defaultValue
+}
+
+// Find a string argument (e.g. --foo "VALUE") of the given name in the given list of arguments. If it's present,
+// return its value. If it is present, but has not value, return an error. If it isn't present, return defaultValue.
```
josh-padnick: s/not value/no value

cli/args.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	return ""
+}
+
+// Custom error types
+
+type ArgMissingValue string
+func (err ArgMissingValue) Error() string {
+	return fmt.Sprintf("You must specify a value for the --%s option", string(err))
+}
```
josh-padnick: I'll pause my review here so I can switch to a timely email.

spin/module.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+
+// Look through the dependencies of the modules in the given map and resolve the "external" dependency paths listed in
+// each modules config (i.e. those dependencies not in the given list of .terragrunt config canonical file paths).
+// These external dependencies are outside of the current working directory, which means they may not be part of the
+// environment the user is trying to spin-up or tear down. Therefore, this method also confirms whether the user wants
+// to actually apply those dependencies or just assume they are already applied. Note that this method will NOT fill in
+// the Dependencies field of the TerraformModule struct (see the crosslinkDependencies method for that).
+func resolveExternalDependenciesForModules(canonicalTerragruntConfigPaths []string, moduleMap map[string]*TerraformModule, terragruntOptions *options.TerragruntOptions) (map[string]*TerraformModule, error) {
```
josh-padnick: I'm curious. Did you intuit this use case, or did you encounter an issue during your tests where specifying modules outside the CWD led to strange behavior? Also, while it's an edge case, do you think this is practically something users will encounter often?

Either way, great that you caught this. Non-obvious.

spin/test_helpers.go
josh-padnick: Thanks, I was getting worried you might have broken your streak seeing this so far down the PR. :)

README.md
brikis98: Yep

README.md
brikis98: You mean the `spin-up` and `tear-down` thing?

README.md
brikis98: Yes

README.md
brikis98: Good point. Added.

README.md
brikis98: Good point. Updated.

cli/args.go
brikis98: Fixed

cli/args.go
brikis98: I found it fairly clunky to use. But having tried 3 Go CLI libs now, I've found them all clunky to use, and am beginning to think the problem is Go itself...

cli/args.go
brikis98: Fixed

cli/cli_app.go
```diff
@@ -30,8 +42,10 @@ COMMANDS:
    import               Acquire a lock and run 'terraform import'
    refresh              Acquire a lock and run 'terraform refresh'
    remote push          Acquire a lock and run 'terraform remote push'
-   acquire-lock         Acquire a long-term long for these templates
-   release-lock         Release a long-term lock or a lock that failed to clean up
+   %s                   Acquire a long-term long for these templates
```
brikis98: Fixed

cli/cli_app.go
```diff
@@ -43,15 +57,12 @@ VERSION:
 AUTHOR(S):
    {{range .Authors}}{{.}}{{end}}
    {{end}}
-`
+`, CMD_ACQUIRE_LOCK, CMD_RELEASE_LOCK, CMD_SPIN_UP, CMD_TEAR_DOWN)
```
brikis98: Depends how many we have, but for now, it's a nice way to make the purpose of those vars crystal clears.

options/options.go
```diff
@@ -1,9 +1,78 @@
...
+		TerraformCliArgs: terragruntOptions.TerraformCliArgs,
+		WorkingDir: workingDir,
+		Logger: util.CreateLogger(workingDir),
+		RunTerragrunt: terragruntOptions.RunTerragrunt,
+	}
+}
+
+// Custom error types
+
+var RunTerragruntCommandNotSet = fmt.Errorf("The RunTerragrunt option has not been set on this TerragruntOptions object")
```
brikis98: For totally static errors, I just use a var, since they are singletons. I'd use a `const`, but Go doesn't allow that if you have any sort of function call. This singleton allows you to check for the error using a plain old equality comparison rather than a type check, which is handy.

brikis98: Also, since this error takes no params, I'd have to either make it an alias of some random type or an empty struct, neither of which makes much sense to me.

spin/module.go
```diff
@@ -0,0 +1,229 @@
...
+func (module *TerraformModule) String() string {
+	dependencies := []string{}
+	for _, dependency := range module.Dependencies {
+		dependencies = append(dependencies, dependency.Path)
+	}
+	return fmt.Sprintf("Module %s (dependencies: [%s])", module.Path, strings.Join(dependencies, ", "))
+}
+
+// Go through each of the given .terragrunt config files and resolve the module that .terragrunt file represents it
+// into a TerraformModule struct. Return the list of these TerraformModule structs.
```
brikis98: Fixed

spin/module.go
brikis98: I realized this was a real case early on, but I initially just returned an error for it. Then I started to try to use Terragrunt with our real code and realized we had a very common example of this: if you try to spin-up, say, the stage VPC, it has dependencies on the mgmt VPC (e.g. bastion host security group, mgmt VPC CIDR block, etc) and the global environment (e.g. route 53 zones and S3 buckets). So this isn't really a corner case; it's quite normal.

brikis98:
Merging now.

README.md
josh-padnick: Yeah.

spin/module.go
josh-padnick: Ah, that's a good point. And just to clarify even though I know it's in the code, does that mean every "external" dependency will always prompt the user on whether they want to apply it unless the user runs in `--non-interactive` mode?

README.md
brikis98: Done in a commit to master

spin/module.go
brikis98: As it is now, yes, it'll prompt for each one. If that turns out to be problematic, we can explore improvements.

Issue #71, opened 2016-12-13
Feedback on the new spin-up and and tear-down commands.


josh-padnick:
I'm going to leave this as an open issue to report and collect feedback on `terragrunt spin-up` and `terragrunt tear-down`.

Overall, it's a joy to use! But was it Thoreau who warned that with each new technology we both gain and lose? With `spin-up` the gains are the ability to launch a lot of infra _fast_ while preserving terraform state isolation, and by extension the ability to launch a whole environment fast. These are wonderful wins. The loss is that I'm less aware of the details of each individual `terragrunt apply`.

For example, I had an issue where a git tag of one of my Terraform modules (module-security) was downloaded, but it was out of date. I'd needed to re-issue a git tag to fix a minor issue, but terraform won't automatically check for new versions of a git tag of an already downloaded module (which is reasonable behavior). If I hadn't manually inspected the results, I wouldn't have seen that in fact terraform was using an outdated module version.

It made me wonder if the `spin-up` paradigm demands a new level of guardrails. For example, should terragrunt do a `terraform get -update` on every module to be safe? Perhaps that's an option since there will be a performance hit with that.

Not sure, so just recording initial thoughts.

josh-padnick:
Ran into another issue around `terraform get`. I expected terragrunt to automatically run `terraform get -update` when it detected no `.terraform` folder, but it failed to do so. While using `terragrunt spin-up` on a fresh folder (i.e. no `.terraform` folder), I got this:

```
[terragrunt] [/repos/gruntwork-clients/<redacted>/infrastructure-live/mgmt/vpc] 2016/12/12 18:57:11 Module /repos/gruntwork-clients/<redacted>/infrastructure-live/mgmt/vpc must wait for 0 dependencies to finish
[terragrunt] [/repos/gruntwork-clients/<redacted>/infrastructure-live/mgmt/vpc] 2016/12/12 18:57:11 Running module /repos/gruntwork-clients/<redacted>/infrastructure-live/mgmt/vpc now
[terragrunt] [/repos/gruntwork-clients/<redacted>/infrastructure-live/mgmt/vpc] 2016/12/12 18:57:11 Reading Terragrunt config file at /repos/gruntwork-clients/<redacted>/infrastructure-live/mgmt/vpc/.terragrunt
[terragrunt] [/repos/gruntwork-clients/<redacted>/infrastructure-live/mgmt/vpc] 2016/12/12 18:57:11 Initializing remote state for the s3 backend
[...]

PR #73, opened 2016-12-14
Make sure terragrunt calls "get" in proper folder during spin-up and tear-down

brikis98:
This is a fix for #71. The `shouldDownloadModules` method in Terragrunt was using the (implicit) current working directory rather than the `WorkingDir` set in `TerragruntOptions`. When running the `spin-up` and `tear-down` commands, you must use `WorkingDir`, or you’ll be looking in the wrong folder.

I added a failing test case, implemented the fix, and now the test case passes.

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -13,6 +13,7 @@ import (
 	"github.com/urfave/cli"
 	"github.com/gruntwork-io/terragrunt/options"
 	"github.com/gruntwork-io/terragrunt/spin"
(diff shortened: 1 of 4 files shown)


brikis98:
@josh-padnick Going to merge this one in, as it's a fairly simple fix. Feedback welcome. I'll cut a new release too.

josh-padnick:
Awesome! I'll try this out tomorrow and report back any issues.

PR #76, opened 2016-12-17
TERRAGRUNT_TFPATH

dmrzzz:
Let user specify full path to terraform binary (so it doesn't need to be on PATH)

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,16 @@ a simple, free locking mechanism, and enforcing best practices around CLI usage 
 
 ## Install
 
-1. Install [Terraform](https://www.terraform.io/) and make sure it is in your PATH.
+1. Install [Terraform](https://www.terraform.io/), and let Terragrunt know where to find it using one of the following options:
+
+    * Place `terraform` in a directory on your PATH.
+
+       **Caution**: this makes it easy to accidentally invoke Terraform directly from the command line (thus bypassing the protections offered by Terragrunt).
+
+    * Specify the full path to the Terraform binary in the environment variable `TERRAGRUNT_TFPATH`.
+
+    * Specify the full path to the Terraform binary in `--terragrunt-tfpath` each time you run Terragrunt (see [CLI Options](#cli-options)).
+
 1. Install Terragrunt by going to the [Releases Page](https://github.com/gruntwork-io/terragrunt/releases), downloading
    the binary for your OS, renaming it to `terragrunt`, and adding it to your PATH.
 
@@ -599,6 +608,8 @@ prefix `--terragrunt-`. The currently available options are:
 
 * `--terragrunt-config`: A custom path to the `.terragrunt` file. May also be specified via the `TERRAGRUNT_CONFIG`
   environment variable. The default path is `.terragrunt` in the current directory.
+* `--terragrunt-tfpath`: A custom path to the Terraform binary. May also be specified via the `TERRAGRUNT_TFPATH`
+  environment variable. The default is `terraform` in a directory on your PATH.
 * `--terragrunt-non-interactive`: Don't show interactive user prompts. This will default the answer for all prompts to 
   'yes'. Useful if you need to run Terragrunt in an automated setting (e.g. from a script).  
 * `--terragrunt-working-dir`: Set the directory where Terragrunt should execute the `terraform` command. Default is the
--- a/cli/args.go
+++ b/cli/args.go
@@ -36,6 +36,14 @@ func parseTerragruntOptionsFromArgs(args []string) (*options.TerragruntOptions, 
 		terragruntConfigPath = config.DefaultTerragruntConfigPath
 	}
 
+	terraformPath, err := parseStringArg(args, OPT_TERRAGRUNT_TFPATH, os.Getenv("TERRAGRUNT_TFPATH"))
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -45,7 +45,7 @@ a simple, free locking mechanism, and enforcing best practices around CLI usage
 
 ## Install
 
-1. Install [Terraform](https://www.terraform.io/) and make sure it is in your PATH.
+1. Install [Terraform](https://www.terraform.io/) and make sure it is in your PATH (or set `TERRAGRUNT_TFPATH`).
```
brikis98: I'd update the docs to be a bit more explicit:

Install [Terraform](https://www.terraform.io/) and make sure it is in your PATH. Alternatively, if you want to ensure you never accidentally run Terraform instead of Terragrunt, instead of putting Terraform in your PATH, specify the path to the Terraform binary via the `--terragrunt-tfpath` command line flag or the  `TERRAGRUNT_TFPATH` environment variable.

dmrzzz:
Full output in [terragrunt_tests.txt](https://github.com/gruntwork-io/terragrunt/files/662110/terragrunt_tests.txt).  My takeaways:

1. With terraform on the PATH: everything seems fine except for

[terragrunt] 2016/12/19 12:29:08 Someone already has a lock on state file terragrunt-test-fixture-lock! arn:aws:iam::378517677616:user/dmrz-terraform@192.17.25.102 acquired the lock on 2016-12-19 18:29:08.023756381 +0000 UTC.
Unable to acquire lock for item terragrunt-test-fixture-lock after 1 retries.
FAIL	github.com/gruntwork-io/terragrunt/test	17.175s

Note: at the end of the test run, I also have a bunch of DynamoDB tables left behind which I have to manually delete from AWS, including "terragrunt_locks" as well as several "terragrunt_test_*"  Good thing I'm not running any real stuff in us-east-1!

2. I see the same problem testing the master branch (without my changes):

Unable to acquire lock for item terragrunt-test-fixture-lock after 1 retries.
FAIL    github.com/gruntwork-io/terragrunt/test 33.559s

3. With terraform not on the PATH (and TERRAGRUNT_TFPATH set correctly): the call to `validateCommandInstalled(t, "terraform")` in integration_test.go quite predictably fails with

        integration_test.go:244: Command 'terraform' not found in PATH

4. With terraform not on the PATH (and TERRAGRUNT_TFPATH set correctly), if we comment out the call to validateCommandInstalled: we're back to everything seems fine except for

Unable to acquire lock for item terragrunt-test-fixture-lock after 1 retries.
[...]
FAIL    github.com/gruntwork-io/terragrunt/test 24.056s

I note that this time # 4 happened to produce some positive output ("Apply complete!") that # 1 didn't produce; I presume this is arbitrary and occurred because a different test won the race condition for the lock.


I'm not sure what you want to do about # 3.  It seems to me that there shouldn't be any need for validateCommandInstalled in the first place (if terragrunt can't find terraform in the test environment, then the test will fail anyway, right?) so one option is to remove it (# 4), but I assume it was put there for some reason that I'm not seeing.  If so, another option is that runTerragruntCommand could be likewise supplemented to check for TERRAGRUNT_TFPATH and, if found, validate that rather than "terraform".  Of course this still won't help if you prefer using --terragrunt-tfpath instead of the env var, but I can't think of any remotely reasonable way to get the tests to handle that case.

I'm surprised by the remaining test failure in # 1/2/4, but I think # 2 proves that it's not my fault (and logically it has nothing to do with my changes anyway).

brikis98:
> With terraform on the PATH: everything seems fine except for

It's possible the (seeming unrelated) test failure you are seeing will be fixed by #82. 

> Note: at the end of the test run, I also have a bunch of DynamoDB tables left behind which I have to manually delete from AWS, including "terragrunt_locks" as well as several "terragrunt_test_*" Good thing I'm not running any real stuff in us-east-1!

Sorry about that. There must be a bug where the tests aren't properly cleaning up after themselves. I opened #83 to track this.

> (if terragrunt can't find terraform in the test environment, then the test will fail anyway, right?) so one option is to remove it (# 4), but I assume it was put there for some reason that I'm not seeing.

It was there to make the error message clearer, especially when running the tests in a CI job where we may have forgotten to install Terraform itself. I think you can safely remove that check.

dmrzzz:
> It was there to make the error message clearer, especially when running the tests in a CI job where we may have forgotten to install Terraform itself. I think you can safely remove that check.

Removed, thanks.

> It's possible the (seeming unrelated) test failure you are seeing will be fixed by #82.

Nope; I've rebased my branch onto fba5d8884644009aa8cf1f1268728ceb6be457af (and squashed it into one commit), but I still see the same unrelated test failure.  However, everything else now works as expected with terraform not on the PATH (and TERRAGRUNT_TFPATH set correctly).
[terragrunt_tests_20161220.txt](https://github.com/gruntwork-io/terragrunt/files/665247/terragrunt_tests_20161220.txt)

Is there anything else you need from me?

brikis98:
Hm, very weird that the tests aren't passing. I'll try to merge this in and see if the CI job passes. Thanks!

brikis98:
Looks like the tests passed. I just created a new release and the new binaries should show up in a few minutes: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.7.1

dmrzzz:
Awesome, thanks!

PR #77, opened 2016-12-17
Add environment variable support to .terragrunt

josesa:
Granted, might not be that useful feature for everyone but it was handy for me in the specific context where I use terraform.

Also, my first time with playing around with go. Nice features in a language.


brikis98:
Thanks for the PR!

Before diving into the code, could you explain the use case for this feature? 

One of the goals with Terragrunt was to make it as declarative as possible, to prevent mistakes like forgetting to configure remote state storage. If the remote state storage depends on an environment variable, rather than the contents of a file, you are more likely to make a mistake. 

I'm curious how you're using Terragrunt where you may need to dynamically change these values.

josesa:
Is just a way to easily test terraform changes without breaking a production environment.

We use terraform with prefix variables in our resources names to test environment changes before applying them to production. We use terraform variables and environment variables for manage these.

```bash
TF_VAR_prefix=test terraform apply
```
This means that we can create a truly copy of our live environment for testing purposed without having to change any of the code and accidentally overwriting our production `tfstate`. We iterate on a `test|development` environment (with an independent `tfstate` stored in S3) and when happy, we change to `production`.

Is possible to use completely different configuration files but we found it to be more error prone and easier to leave something behind or even not being 100% sure which environment was being affected. 
Before `terragrunt` we used wrapper scripts that ensure that the correct remote file was configured but it was not that clear. `terragrunt` makes it much easier to maintain but I just missed this feature of having multiple environments for not having to commit the file with the production setup and accidentally destroying its `tfstate` file when doing experimental changes in our parallel setup.

But I also see that is quite limited to the use case and that might be weird for a lot of people. For example, I considered not having a default and instead just break if the variable is not defined to prevent accidental overwrites.

brikis98:
Making it easier to use Terraform and Terragrunt with multiple environments is something we've been thinking about a lot. We have some ideas about how to do it, but are still coming up with a good design. I'd like to learn more about your env-var based approach. I'm also curious how it compares with the [approach recommended by Kief Morris](https://medium.com/@kief/https-medium-com-kief-using-pipelines-to-manage-environments-with-infrastructure-as-code-b37285a1cbf5).

Also, one other question: don't you have to set different variables for each environment? E.g. different VPC IDs, IPs, DNS entries, AMI IDs, etc? If so, are you managing those through environment variables too?

josesa:
@brikis98 Yes. We use terraform variables for that

Here is a simple example: 
```hcl
/*************************
  ElastiCache Production
************************ */
resource "aws_elasticache_cluster" "production-cache" {
  cluster_id = "${var.prefix}-${var.application["name"]}-prod"

  # General Config
  engine               = "redis"
  num_cache_nodes      = 1
  parameter_group_name = "default.redis3.2"

  # Instance Config
  node_type = "${var.cache["instance_type"]}"

  # Network
  port               = 6379
  subnet_group_name  = "${aws_elasticache_subnet_group.cache.name}"
  security_group_ids = ["${aws_security_group.instance.id}"]
}
```
`${var.prefix}` is a variable that needs to be set on `tfvars` or (preferably) via ENV variables:

```bash
TF_VAR_prefix="production" terraform  plan
```

brikis98:
Ah, interesting. Env vars strike me as a riskier way of doing this than `.tfvars` files, but I'm not totally against them. Let me look at the code...

README.md
```diff
@@ -464,7 +464,57 @@ remote_state = {
...
+### Environment variables replacement
+
+With terraform is possible to set variables via Environment variables. This allows to control similar infrastructures with the same configuration files. Is also possible to use these variables in `.terragrunt` file by using the `get_env()` helper function.
+
+```hcl
+remote_state = {
+  backend = "s3"
+  config {
+    encrypt = "true"
+    bucket = "${get_env(ENVIRONMENT_VARIABLE, 'development')}-bucket"
```
brikis98: Shouldn't the environment variable name be quoted? I think that would be more idiomatic for HCL syntax.

config/config_helpers.go
```diff
@@ -32,22 +34,49 @@ func ResolveTerragruntConfigString(terragruntConfigString string, include *Inclu
...
+	if len(matches) == 3 {
+		return executeTerragruntHelperFunction(matches[1], include, terragruntOptions, matches[2])
 	} else {
 		return "", errors.WithStackTrace(InvalidInterpolationSyntax(str))
 	}
 }
 
 // Execute a single Terragrunt helper function and return its value as a string
-func executeTerragruntHelperFunction(functionName string, include *IncludeConfig, terragruntOptions *options.TerragruntOptions) (string, error) {
+func executeTerragruntHelperFunction(functionName string, include *IncludeConfig, terragruntOptions *options.TerragruntOptions, parameters string) (string, error) {
```
brikis98: I would put `parameters` right after `functionName` in the parameter list.

config/config_helpers.go
```diff
@@ -10,7 +11,8 @@ import (
 )
 
 var INTERPOLATION_SYNTAX_REGEX = regexp.MustCompile("\\$\\{.*?\\}")
-var HELPER_FUNCTION_SYNTAX_REGEX = regexp.MustCompile("\\$\\{(.*?)\\(\\)\\}")
+var HELPER_FUNCTION_SYNTAX_REGEX = regexp.MustCompile(`\$\{(.*?)\((.*?)\)\}`)
+var HELPER_FUNCTION_GET_ENV_PARAMETERS_SYNTAX_REGEX = regexp.MustCompile(`(?P<ENV>[a-zA-Z_]+[a-zA-Z0-9_]+?)\,\s*\'(?P<default>.*?)\'`)
```
brikis98: If both the env var name and default value are quoted, the regex becomes a bit simpler. Nice use of named capture groups, BTW. I didn't realize Go supported those.

config/config_helpers.go
```diff
@@ -32,22 +34,49 @@ func ResolveTerragruntConfigString(terragruntConfigString string, include *Inclu
...
 
+func extractParametersFromFunction(parameters string) (map[string]string, error) {
+	matches := HELPER_FUNCTION_GET_ENV_PARAMETERS_SYNTAX_REGEX.FindStringSubmatch(parameters)
+	if (len(matches)) < 2 {
+		return nil, errors.WithStackTrace(InvalidFunctionParameters(parameters))
+	}
+	result := make(map[string]string)
+	for i, name := range HELPER_FUNCTION_GET_ENV_PARAMETERS_SYNTAX_REGEX.SubexpNames() {
+		if i != 0 { result[name] = matches[i] }
+  }
```
brikis98: White space issue on this curly brace?

config/config_helpers.go
```diff
@@ -32,22 +34,49 @@ func ResolveTerragruntConfigString(terragruntConfigString string, include *Inclu
...
+
+	return result, nil
+}
+
+func getEnvironmentVariables(parameters string) (string, error) {
+	result, err := extractParametersFromFunction(parameters)
```
brikis98: `parameterMap` might be a better name for this variable than `result`.

config/config_helpers_test.go
```diff
@@ -316,4 +317,86 @@ func TestResolveTerragruntConfigString(t *testing.T) {
...
+			"",
+			InvalidFunctionParameters("Invalid Parameters"),
+			nil,
+		},
+		{
+			"foo/${get_env(ENV,\"\")}/bar",
+			nil,
+			options.TerragruntOptions{TerragruntConfigPath: "/root/child/.terragrunt", NonInteractive: true},
+			"",
+			InvalidFunctionParameters("ENV,\"\""),
```
brikis98: Why is this test case invalid?

config/config_helpers_test.go
```diff
@@ -316,4 +317,86 @@ func TestResolveTerragruntConfigString(t *testing.T) {
...
+			nil,
+			map[string]string{"TEST_ENV_TERRAGRUNT_OTHER": "SOMETHING"},
+		},
+		{
+			"foo/${get_env(TEST_ENV_TERRAGRUNT_HIT,'default')}/bar",
+			nil,
+			options.TerragruntOptions{TerragruntConfigPath: "/root/child/.terragrunt", NonInteractive: true},
+			"foo/HIT/bar",
+			nil,
+			map[string]string{"TEST_ENV_TERRAGRUNT_HIT": "HIT"},
```
brikis98: Do any of these test cases validate that `get_env` falls back to the default value?

config/config_helpers_test.go
```diff
@@ -316,4 +317,86 @@ func TestResolveTerragruntConfigString(t *testing.T) {
...
+		},
+	}
+
+	for _, testCase := range testCases {
+		if len(testCase.env) > 0 {
+			for key, value := range testCase.env {
+				os.Setenv(key, value)
```
brikis98: This will not work if we have multiple tests running in parallel that use env vars...

Perhaps a way to resolve this is to add a field to `TerragruntOptions` called `Env`, which is a `map[string]string`. The default value for it, set by `parseTerragruntOptionsFromArgs`, will be the parsed version of `os.Environ()`. But at test time, you could provide a custom map of env vars.

josesa:
Updated according to the suggestions and notes. 

Point 1. is quite interesting and I was also thinking about it. Might be more relevant now that we are pulling all variables and not just the ones that were declared in the file.
Good idea of using the `terragruntOptions` for this. Much better than using the `os` directly.

using `TF_VAR_*` as a filter would reduce this but might not be so clear to the user since we are not really getting `terraform` variables.
Maybe 
 - put a note on the documentation to use the full format if one wants to reuse?
 - change the function name to `get_env_tf_var("variable", "")` for `TF_VAR_variable`.

Regarding the tests I'm still fighting a bit this as is quite new to me.
 - I had to change all imports to local paths since using the `github` modules would of course not load my changed files.

```diff
-       "github.com/gruntwork-io/terragrunt/options"
+       "../options"
```
Is there a better way to do this?

I've been running the tests in each module. Is there a way to run them all with one command?

*config*
```
go test -v -parallel 128
=== RUN   TestPathRelativeToInclude
=== RUN   TestFindInParentFolders
=== RUN   TestResolveTerragruntInterpolation
=== RUN   TestResolveTerragruntConfigString
=== RUN   TestResolveEnvInterpolationConfigString
=== RUN   TestParseTerragruntConfigDynamoLockMinimalConfig
=== RUN   TestParseTerragruntConfigDynamoLockFullConfig
=== RUN   TestParseTerragruntConfigDynamoLockMissingStateFileId
=== RUN   TestParseTerragruntConfigRemoteStateMinimalConfig
=== RUN   TestParseTerragruntConfigRemoteStateMissingBackend
=== RUN   TestParseTerragruntConfigRemoteStateFullConfig
=== RUN   TestParseTerragruntConfigDependenciesOnePath
=== RUN   TestParseTerragruntConfigDependenciesMultiplePaths
=== RUN   TestParseTerragruntConfigRemoteStateDynamoDbAndDependenciesFullConfig
=== RUN   TestParseTerragruntConfigInvalidLockBackend
=== RUN   TestParseTerragruntConfigInclude
=== RUN   TestParseTerragruntConfigIncludeWithFindInParentFolders
=== RUN   TestParseTerragruntConfigIncludeOverrideRemote
=== RUN   TestParseTerragruntConfigIncludeOverrideAll
=== RUN   TestParseTerragruntConfigTwoLevels
=== RUN   TestParseTerragruntConfigThreeLevels
=== RUN   TestParseTerragruntConfigEmptyConfig
=== RUN   TestMergeConfigIntoIncludedConfig
--- PASS: TestPathRelativeToInclude (0.00s)
--- PASS: TestParseTerragruntConfigDependenciesMultiplePaths (0.00s)
--- PASS: TestParseTerragruntConfigDependenciesOnePath (0.00s)
--- PASS: TestParseTerragruntConfigInvalidLockBackend (0.00s)
--- PASS: TestParseTerragruntConfigRemoteStateFullConfig (0.00s)
--- PASS: TestParseTerragruntConfigRemoteStateMissingBackend (0.00s)
--- PASS: TestParseTerragruntConfigInclude (0.00s)
--- PASS: TestParseTerragruntConfigIncludeOverrideAll (0.00s)
--- PASS: TestParseTerragruntConfigRemoteStateMinimalConfig (0.00s)
--- PASS: TestParseTerragruntConfigRemoteStateDynamoDbAndDependenciesFullConfig (0.00s)
--- PASS: TestMergeConfigIntoIncludedConfig (0.00s)
--- PASS: TestParseTerragruntConfigDynamoLockMissingStateFileId (0.00s)
--- PASS: TestParseTerragruntConfigDynamoLockFullConfig (0.00s)
--- PASS: TestParseTerragruntConfigThreeLevels (0.00s)
--- PASS: TestParseTerragruntConfigDynamoLockMinimalConfig (0.00s)
--- PASS: TestParseTerragruntConfigEmptyConfig (0.00s)
--- PASS: TestResolveEnvInterpolationConfigString (0.00s)
--- PASS: TestParseTerragruntConfigTwoLevels (0.00s)
--- PASS: TestResolveTerragruntConfigString (0.00s)
--- PASS: TestResolveTerragruntInterpolation (0.00s)
--- PASS: TestParseTerragruntConfigIncludeOverrideRemote (0.00s)
--- PASS: TestFindInParentFolders (0.00s)
--- PASS: TestParseTerragruntConfigIncludeWithFindInParentFolders (0.00s)
PASS
ok  	_/Users/jose/source/terragrunt/config	0.261s
```

*cli*
```
 go test -v -parallel 128
=== RUN   TestParseTerragruntOptionsFromArgs
=== RUN   TestFilterTerragruntArgs
=== RUN   TestParseEnvironmentVariables
--- PASS: TestParseEnvironmentVariables (0.00s)
--- PASS: TestFilterTerragruntArgs (0.00s)
--- PASS: TestParseTerragruntOptionsFromArgs (0.00s)
PASS
ok  	_/Users/jose/source/terragrunt/cli	0.012s
```

brikis98:
> put a note on the documentation to use the full format if one wants to reuse?

I guess since we don't have a concrete use case, for now, we shouldn't try to guess and end up violating YAGNI. Let's stick with the approach you have now and add a note to the README to call `get_env("TF_VAR_XXX")` if you're trying to reuse Terraform variables.

> I had to change all imports to local paths since using the github modules would of course not load my changed files.

Go is very picky about the `GOPATH`. If you check this code out into your `GOPATH`, then those imports will work correctly without any changes. For example, my `GOPATH` is `~/source/go` and that code is checked out into `~/source/go/src/github.com/gruntwork-io/terragrunt`. You can read more about it here: https://golang.org/cmd/go/#hdr-GOPATH_environment_variable

> I've been running the tests in each module. Is there a way to run them all with one command?

Yup, as it says in the README, you can run the following in the root `terragrunt` directory:

```
go test -v $(glide novendor)
```

brikis98:
BTW, the code looks good, so if you are able to sort out the import issues and all the tests are passing, let me know, and I'll merge.

brikis98:
@josesa Were you able to resolve the import issues and get all tests to pass?

josesa:
@brikis98 Not yet, I have been on vacations with limited internet connection. For sure until the weekend.

josesa:
@brikis98 Fixed the imports and run the tests. One of them is failing, (but is failing on master for me as well)

This is the failing test:
http://pastebin.com/yfeqHVND

And all of them them:
http://pastebin.com/L7xKkW0S

brikis98:
> One of them is failing, (but is failing on master for me as well)

Make sure you pull in the latest version of master and re-run. I believe the test failure you're seeing was fixed with this commit: https://github.com/gruntwork-io/terragrunt/commit/559f20340f5d35d91787ad2296e8be81468f595a

README.md
```diff
@@ -464,7 +464,75 @@ remote_state = {
 ``` 
 
 The result is that when you run `terragrunt` commands in the `qa/my-app` folder, you get the `lock` settings from the 
-parent, but the `remote_state` settings of the child. 
+parent, but the `remote_state` settings of the child.
+
+### Environment variables replacement
+
+With terraform is possible to set variables via Environment variables. This allows to control similar infrastructures with the same configuration files. Is also possible to use these variables in `.terragrunt` file by using the `get_env()` helper function.
```
brikis98: I would shorten this first paragraph to simply: 

You can read in environment variables in your `.terragrunt` file using the `get_env()` helper function:

cli/args.go
```diff
@@ -53,9 +53,23 @@ func parseTerragruntOptionsFromArgs(args []string) (*options.TerragruntOptions,
...
+		Env: parseEnvironmentVariables(os.Environ()),
 	}, nil
 }
 
+func parseEnvironmentVariables(environment []string) map[string]string {
+	environmentMap := make(map[string]string)
+
+	for i := 0; i < len(environment); i++ {
+		variableSplit := strings.SplitN(environment[i], "=", 2)
+		if (len(variableSplit) == 2) {
```
brikis98: In Go, you don't need parens around the if-statement clause. So this can be shortened to:

```go
if len(variableSplit) == 2 {
  ...
}
```

cli/args_test.go
```diff
@@ -148,4 +148,41 @@ func TestFilterTerragruntArgs(t *testing.T) {
...
+
+		{
+			[]string{"foo=composite=bar"},
+			map[string]string{"foo": "composite=bar"},
+		},
+	}
+
+	for _, testCase := range testCases {
+		actualVariables := parseEnvironmentVariables(testCase.environmentVariables)
+			assert.Equal(t, testCase.expectedVariables, actualVariables)
```
brikis98: Extra whitespace before the `assert.Equal` call?

config/config_helpers.go
```diff
@@ -10,7 +10,8 @@ import (
 )
 
 var INTERPOLATION_SYNTAX_REGEX = regexp.MustCompile("\\$\\{.*?\\}")
-var HELPER_FUNCTION_SYNTAX_REGEX = regexp.MustCompile("\\$\\{(.*?)\\(\\)\\}")
+var HELPER_FUNCTION_SYNTAX_REGEX = regexp.MustCompile(`\$\{(.*?)\((.*?)\)\}`)
+var HELPER_FUNCTION_GET_ENV_PARAMETERS_SYNTAX_REGEX = regexp.MustCompile(`\"(?P<ENV>[a-zA-Z_]+[a-zA-Z0-9_]+?)\"\,\s*\"(?P<default>.*?)\"`)
```
brikis98: Nice use of named regular expressions! 

However, a few questions:

First, I notice you're escaping double quotes. Since your regex string uses backticks, is that really necessary?

Second, I'm looking at this expression:

> `[a-zA-Z_]+[a-zA-Z0-9_]+?`

Is there a benefit to being strict with the types of characters allowed? For example, I'd have to go dig through the rules for Unix/Linux/OS X/Windows to verify if environment variables are indeed limited to alphanumeric/underscore. I'm guessing that's correct, but it seems like that could be simplified to a simple `.+?` without losing much.

Third, I noticed you're only handling whitespace here:

> "\,\s*

I think you'd need a `\s*` before the comma too. Moreover, people might use whitespace before or after the parens: e.g. `get_env(   "foo"   ,  "bar"    )`.

And finally, fourth, `ENV` and `default` should probably use consistent case: either both upper case or both lower case.

config/config_helpers.go
```diff
@@ -32,22 +33,52 @@ func ResolveTerragruntConfigString(terragruntConfigString string, include *Inclu
...
 	case "find_in_parent_folders": return findInParentFolders(terragruntOptions)
 	case "path_relative_to_include": return pathRelativeToInclude(include, terragruntOptions)
+	case "get_env": return getEnvironmentVariables(parameters, terragruntOptions)
 	default: return "", errors.WithStackTrace(UnknownHelperFunction(functionName))
 	}
 }
 
+func extractParametersFromFunction(parameters string) (map[string]string, error) {
+	matches := HELPER_FUNCTION_GET_ENV_PARAMETERS_SYNTAX_REGEX.FindStringSubmatch(parameters)
+	if (len(matches)) < 2 {
```
brikis98: No need for parens on the if-clause

config/config_helpers.go
```diff
@@ -32,22 +33,52 @@ func ResolveTerragruntConfigString(terragruntConfigString string, include *Inclu
...
+		return nil, errors.WithStackTrace(InvalidFunctionParameters(parameters))
+	}
+	result := make(map[string]string)
+	for value, name := range HELPER_FUNCTION_GET_ENV_PARAMETERS_SYNTAX_REGEX.SubexpNames() {
+		if value != 0 {
+			result[name] = matches[value]
+		}
+  }
```
brikis98: Incorrect whitespace on the closing curly brace?

config/config_helpers.go
```diff
@@ -32,22 +33,52 @@ func ResolveTerragruntConfigString(terragruntConfigString string, include *Inclu
...
 
+func extractParametersFromFunction(parameters string) (map[string]string, error) {
+	matches := HELPER_FUNCTION_GET_ENV_PARAMETERS_SYNTAX_REGEX.FindStringSubmatch(parameters)
+	if (len(matches)) < 2 {
+		return nil, errors.WithStackTrace(InvalidFunctionParameters(parameters))
+	}
+	result := make(map[string]string)
+	for value, name := range HELPER_FUNCTION_GET_ENV_PARAMETERS_SYNTAX_REGEX.SubexpNames() {
+		if value != 0 {
+			result[name] = matches[value]
```
brikis98: You may want to call `strings.TrimSpace` on both the `name` and `value`.

josesa:
You are right. Somehow I didn't rebase it correctly.
Now they are passing 🎉 :

http://pastebin.com/xQcwaiTF

config/config_helpers.go
```diff
@@ -32,22 +33,52 @@ func ResolveTerragruntConfigString(terragruntConfigString string, include *Inclu
...
 	}
 }
 
+func extractParametersFromFunction(parameters string) (map[string]string, error) {
+	matches := HELPER_FUNCTION_GET_ENV_PARAMETERS_SYNTAX_REGEX.FindStringSubmatch(parameters)
+	if (len(matches)) < 2 {
+		return nil, errors.WithStackTrace(InvalidFunctionParameters(parameters))
+	}
+	result := make(map[string]string)
+	for value, name := range HELPER_FUNCTION_GET_ENV_PARAMETERS_SYNTAX_REGEX.SubexpNames() {
```
brikis98: I think `index` is a better name than `value`, since that parameter is the index of the iteration.

config/config_helpers.go
```diff
@@ -32,22 +33,52 @@ func ResolveTerragruntConfigString(terragruntConfigString string, include *Inclu
...
+		if value != 0 {
+			result[name] = matches[value]
+		}
+  }
+
+	return result, nil
+}
+
+func getEnvironmentVariables(parameters string, terragruntOptions *options.TerragruntOptions) (string, error) {
+	parameterMap, err := extractParametersFromFunction(parameters)
```
brikis98: `extractParametersFromFunction` makes it sound like a generic function that could parse any type of parameter, but the reality is that the function uses a specific regex to parse a specific type of parameter (an environment variable and a default fallback value). To use the opaque `map[string]string` that it returns, you have to know what regex was used. I found this a little confusing.

Perhaps a cleaner approach is to:

1. Rename the function to `parseGetEnvParameters`.
1. Instead of returning a `map[string]string`, have the function return a struct with clearly named fields:
    ```go
    type EnvVar struct {
      Name         string
      DefaultValue string
    }
    ```

This way, all the regex details are fully encapsulated within the function, and it's easier to see what the function does and how to use its return value.

config/config_helpers.go
josesa: Regarding the expression, I confess I used as reference: http://stackoverflow.com/questions/2821043/allowed-characters-in-linux-environment-variable-names?answertab=votes#tab-top and totally forgot about the windows case.

I avoided the `.` since it seems a bit to aggressive even with `go` ignoring new lines. What about `(?P<env>[^=]+?)` that excludes `=` that is not allowed in any system (*nix and windows) and could lead to confusion.

I've added a few more tests to cover edge cases.

config/config_helpers.go
josesa: Not sure what you mean. In this context `name` are the `"env"` and `"default"` strings from the regex.
anyway, I've added some logic to trim the variable name in both the hcl and as well from the environment but let the value contents untouched since the white space is a valid content.

cli/args.go.orig
```diff
@@ -0,0 +1,150 @@
+package cli
```
brikis98: I think you may have accidentally checked in some files `.orig` files, probably from a merge conflict. Could you remove these?

brikis98:
Awesome, thx!

brikis98:
I just created a new release: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.9.1. If the build passes, the new binaries will show up there shortly.

acesaro:
Love this, thank you!

PR #82, opened 2016-12-20
Fix spin-up missing deeper subfolders. Fix AWS API getting overloaded during tests. 

### LINKED ISSUE #81: Spin-up misses modules in subfolders.
Given the following folder structure:

```
infrastructure-live
- stage
  - data-stores
    - postgres
    - redis
  - ecs-cluster
  - kms-master-key
  - vpc
```

I ran `terragrunt spin-up` and expected Terragrunt to apply all the above folders as long as each had a `.terragrunt` file. But the `stage/data-stores` is ignored, even though both `stage/data-stores/postgres` and `stage/data-stores/redis` are confirmed to have a `.terragrunt` file .

Here's a [zip file of the full infrastructure-live directory](https://drive.google.com/open?id=0B161Wobnisu-cVo0RHE3bzJESVk) in a private Google Drive folder available to Gruntworkers only so someone else can reproduce.

Here's the output. Note the inclusion of ignored modules.

```
josh@Joshs-MacBook-Pro /r/g/b/i/stage> terragrunt spin-up
[terragrunt]  Module /repos/gruntwork-clients/acme/infrastructure-live/stage/vpc depends on module /repos/gruntwork-clients/acme/infrastructure-live/mgmt/vpc, which is an external dependency outside of the current working directory. Should Terragrunt skip over this external dependency? Warning, if you say 'no', Terragrunt will make changes in /repos/gruntwork-clients/acme/infrastructure-live/mgmt/vpc as well! (y/n) y
[terragrunt]  Module /repos/gruntwork-clients/acme/infrastructure-live/stage/ecs-cluster depends on module /repos/gruntwork-clients/acme/infrastructure-live/mgmt/bastion-host, which is an external dependency outside of the current working directory. Should Terragrunt skip over this external dependency? Warning, if you say 'no', Terragrunt will make changes in /repos/gruntwork-clients/acme/infrastructure-live/mgmt/bastion-host as well! (y/n) y
[terragrunt]  Module /repos/gruntwork-clients/acme/infrastructure-live/stage/ecs-cluster depends on module /repos/gruntwork-clients/acme/infrastructure-live/global/sns-topics, which is an external dependency outside of the current working directory. Should Terragrunt skip over this external dependency? Warning, if you say 'no', Terragrunt will make changes in /repos/gruntwork-clients/acme/infrastructure-live/global/sns-topics as well! (y/n) y
[terragrunt] 2016/12/19 15:07:23 Stack at /repos/gruntwork-clients/acme/infrastructure-live/stage:
  => Module /repos/gruntwork-clients/acme/infrastructure-live/stage/ecs-cluster (dependencies: [/repos/gruntwork-clients/acme/infrastructure-live/stage/vpc, /repos/gruntwork-clients/acme/infrastructure-live/mgmt/bastion-host, /repos/gruntwork-clients/acme/infrastructure-live/global/sns-topics])
  => Module /repos/gruntwork-clients/acme/infrastructure-live/stage/kms-master-key (dependencies: [])
  => Module /repos/gruntwork-clients/acme/infrastructure-live/stage/vpc (dependencies: [/repos/gruntwork-clients/acme/infrastructure-live/mgmt/vpc])
  => Module /repos/gruntwork-clients/acme/infrastructure-live/mgmt/vpc (dependencies: [])
  => Module /repos/gruntwork-clients/acme/infrastructure-live/mgmt/bastion-host (dependencies: [/repos/gruntwork-clients/acme/infrastructure-live/mgmt/vpc, /repos/gruntwork-clients/acme/infrastructure-live/global/sns-topics])
  => Module /repos/gruntwork-clients/acme/infrastructure-live/global/sns-topics (dependencies: [])
[terragrunt]  Are you sure you want to run 'terragrunt apply' in each folder of the stack described above? (y/n)
```

josh-padnick:
This PR fixes two separate issues:

- First, it allows `spin-up` and `tear-down` to look for a `.terragrunt` file in more folders than just the immediate subfolders. I implemented unit tests to validate this functionality.
- Second, Terragrunt tests were failing arbitrarily when the AWS API complained of being too much at once. This was because when many parallel Terragrunt tests created a DynamoDB Table and then waited 10 seconds to check again, all checks came at exactly the same time. I resolved this by introducing a random sleep between tests while keeping the API the same for non-test code (i.e. no random sleep for non-test code).

Fixes #81.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,8 +1,8 @@
 hash: f70d8455d9f5f68d4449263ab8d0485d5b938b65b4c31716be50aac64fedc6ae
-updated: 2016-12-19T15:20:41.695295361-07:00
+updated: 2016-12-19T23:23:04.584262219Z
 imports:
 - name: github.com/aws/aws-sdk-go
-  version: 7a81396c57134038e554b2ac86be4653018b20f9
+  version: 665c623d7f3e0ee276596b006655ba4dbe0565b0
   subpackages:
   - aws
   - aws/awserr
@@ -12,15 +12,12 @@ imports:
   - aws/corehandlers
   - aws/credentials
   - aws/credentials/ec2rolecreds
-  - aws/credentials/endpointcreds
-  - aws/credentials/stscreds
   - aws/defaults
   - aws/ec2metadata
   - aws/request
   - aws/service/dynamodb
   - aws/service/s3
   - aws/session
-  - aws/signer/v4
   - private/endpoints
   - private/protocol
   - private/protocol/json/jsonutil
@@ -30,20 +27,21 @@ imports:
   - private/protocol/rest
   - private/protocol/restxml
   - private/protocol/xml/xmlutil
+  - private/signer/v4
   - private/waiter
   - service/dynamodb
   - service/s3
   - service/sts
 - name: github.com/davecgh/go-spew
-  version: 2df174808ee097f90d259e432cc04442cf60be21
+  version: 04cdfd42973bb9c8589fd6a731800cf222fde1a9
   subpackages:
   - spew
 - name: github.com/go-errors/errors
-  version: 8fa88b06e5974e97fbf9899a7f86a344bfd1f105
+  version: a41850380601eeb43f4350f7d17c6bbd8944aaf8
 - name: github.com/go-ini/ini
-  version: 6e4869b434bd001f6983749881c7ead3545887d8
+  version: 12f418cc7edc5a618a51407b7ac1f1f512139df3
 - name: github.com/hashicorp/hcl
-  version: 80e628d796135357b3d2e33a985c666b9f35eee1
+  version: 9a905a34e6280ce905da1a32344b25e81011197a
   subpackages:
   - hcl/ast
   - hcl/parser
@@ -54,17 +52,17 @@ imports:
   - json/scanner
   - json/token
 - name: github.com/jmespath/go-jmespath
-  version: bd40a432e4c76585ef6b72d3fd96fb9b6dc7b68d
+  version: 0b12d6b521d83fc7f755e7cfc1b1fbdd35a01a74
 - name: github.com/mitchellh/mapstructure
-  version: bfdb1a85537d60bc7e954e600c250219ea497417
+  version: f3009df150dadf309fdee4a54ed65c124afad715
 - name: github.com/pmezard/go-difflib
   version: d8ed2627bdf02c080bf22230dbb337003b7aba2d
   subpackages:
   - difflib
 - name: github.com/stretchr/testify
-  version: f390dcf405f7b83c997eac1b06768bb9f44dec18
+  version: 18a02ba4a312f95da08ff4cfc0055750ce50ae9e
   subpackages:
   - assert
 - name: github.com/urfave/cli
-  version: 0bdeddeeb0f650497d603c4ad7b20cfe685682f6
+  version: 9a18ffad6c548ca4d458d4c30a8aa4d181cf92fa
 testImports: []
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
 hash: f70d8455d9f5f68d4449263ab8d0485d5b938b65b4c31716be50aac64fedc6ae
-updated: 2016-12-19T23:23:04.584262219Z
+updated: 2016-12-19T19:45:59.216824394-07:00
 imports:
 - name: github.com/aws/aws-sdk-go
   version: 665c623d7f3e0ee276596b006655ba4dbe0565b0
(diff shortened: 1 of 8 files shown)


spin/stack.go
```diff
@@ -47,7 +49,10 @@ func (stack *Stack) CheckForCycles() error {
 // Find all the Terraform modules in the subfolders of the working directory of the given TerragruntOptions and
 // assemble them into a Stack object that can be applied or destroyed in a single command
 func FindStackInSubfolders(terragruntOptions *options.TerragruntOptions) (*Stack, error) {
-	terragruntConfigFiles, err := filepath.Glob(fmt.Sprintf("%s/**/%s", terragruntOptions.WorkingDir, config.DefaultTerragruntConfigPath))
+	// Ideally, we'd use a builin Go library like filepath.Glob here, but per https://github.com/golang/go/issues/11862,
+	// the current go implementation doesn't support treating ** as zero or more directories, just zero or one.
+	// So we use a third-party library.
```
brikis98: Good catch.

I really think Go is a shitty language. Don't name the method `Glob` if it doesn't implementing globbing the way 99.9% of platforms do. And then refuse to fix clearly broken behavior in the name of backwards compatibility.

util/random.go
```diff
@@ -0,0 +1,36 @@
...
+func GetRandomTime(lowerBound, upperBound time.Duration) time.Duration {
+	if lowerBound == upperBound {
+		return lowerBound
+	}
+
+	var lowerBoundMs, upperBoundMs float64
+
+	if upperBound < 1 * time.Second {
+		lowerBoundMs = 1
+		upperBoundMs = 1000
```
brikis98: Why this special case?

util/random.go
```diff
@@ -0,0 +1,36 @@
...
+	"time"
+	"math/rand"
+)
+
+// Get a random time duration between the lower bound and upper bound. This is useful because some of our automated tests
+// wound up flooding the AWS API all at once, leading to a "Subscriber limit exceeded" error.
+func GetRandomTime(lowerBound, upperBound time.Duration) time.Duration {
+	if lowerBound == upperBound {
+		return lowerBound
+	}
```
brikis98: What about the case where `lowerBound < upperBound`?

util/random_test.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+func TestGetRandomTime(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct{
+		lowerBound time.Duration
+		upperBound time.Duration
+	}{
+		{ 1 * time.Second, 10 * time.Second },
```
brikis98: Might be worth testing a bunch of different ranges here, such as:

* (0, 0)
* (-1, -3)
* (1, 1.000001)
* (1, 10)

It may also be worth running a loop for each test case and testing it multiple times, as there are random numbers involved, so it's possible a test will pass as a pure fluke simply because the random number didn't catch a corner case.

glide.lock
```diff
@@ -1,8 +1,8 @@
 hash: f70d8455d9f5f68d4449263ab8d0485d5b938b65b4c31716be50aac64fedc6ae
```
brikis98: I think you're missing `glide.yaml` in this PR.

spin/stack.go
josh-padnick: Yeah, I hear you. Their minimalism is great for the learning curve but you constantly find yourself re-implementing boilerplate.

util/random.go
josh-padnick: Ah, that's a good point, but what to return? I'll play around with a couple of options.

util/random.go
josh-padnick: It was an early implementation vestige. The thinking was that at a certain level, time ranges are so small as to not be meaningful, but no reason I need to enforce that. Will remove.

util/random.go
brikis98: Either an error or `upperBound`, I guess.

glide.lock
josh-padnick: Well, tests pass, so I guess `glide.lock` is good enough, but I manually added to `glide.yaml` for consistency's sake.

brikis98: Without `glide.yaml`, the next time someone runs `glide up`, it'll lose the new zglob dependency.

josh-padnick:
Ok, all comments responded to. If this passes tests, unless you strongly object to some of the edge cases I left out of `util.GetRandomTime()`, I think we're good to go here.

util/random.go
```diff
@@ -0,0 +1,42 @@
...
+		upperBound = -1 * upperBound
+	}
+
+	if lowerBound > upperBound {
+		return upperBound
+	}
+
+	if lowerBound == upperBound {
+		return lowerBound
+	}
```
brikis98: You could remove this case and just change the previous one to `>=` for exactly the same behavior.

brikis98:
(Feel free to merge when ready)

Issue #24, opened 2016-08-30
terragrunt graph adds header, invalidates dot format


lsc:
Just a cosmetics thing, if you run `terragrunt graph | dot -Tpng > mygraph.png`
you will get an error like so `Error: <stdin>: syntax error in line 1 near '['` because terragrunt adds the two lines `[terragrunt] 2016/08/30 11:41:34 Remote state is already configured for backend s3
[terragrunt] 2016/08/30 11:41:34 Running command: terraform graph`

It's easy to just run `terraform graph` instead, but I wanted to give a heads up

brikis98:
Ah, good point. I suppose there are two options: either a) add a `--silent` flag that suppresses terragrunt logging or b) detect if we're being piped into another process, at least for certain commands like `graph`, and automatically suppress logging in those cases.

lsc:
I was thinking of just disable logging if the command was graph, but maybe figuring out if we are piping the command is the nicer way to solve it.

A flag like --silent would be confusing I think, as it isn't clear if we silence terragrunt only or if it is a flag passed to terraform.

dln:
One solution would be to log to stderr instead of stdout. This would ensure "transparent" wrapping around terraform.

As a personal preference, I would also opt for being quiet by default.

brikis98:
This should be fixed by #94.

PR #94, opened 2017-01-08
Log terragrunt output to stderr

bshelton229:
Logging to stderr is a potential solution for #24. We alias `terraform -> terragrunt` everywhere to avoid any mistakes (we love terragrunt) and would like to still be able to rely on piping and reading from terraform's stdout.

This should fix commands like `terragrunt graph | dot -Tpng > mygraph.png` and allow reading json from stdout from commands like `terragrunt output -json | json "my-key.my-value"`.

I also wouldn't mind helping with a `-silent` flag if that is a preferred approach, or something that could be implemented as well as writing terragrunt output to stderr.

### DIFF
--- a/util/logger.go
+++ b/util/logger.go
@@ -1,9 +1,9 @@
 package util
 
 import (
-	"os"
-	"log"
 	"fmt"
+	"log"
+	"os"
 )
 
 // Create a logger with the given prefix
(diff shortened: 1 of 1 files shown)


brikis98:
Oh, good call. I think this even fits better with the purpose of stdout/stderr. That is, stdout is for the actual output of the program (in this case, the output of Terraform itself) and stderr is for diagnostic info, errors, etc (such as what Terragrunt is logging). 

Thanks!

brikis98:
I just created a new release here: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.8.0. If the build passes, the new binaries should be there in a few minutes.

PR #100, opened 2017-01-13
Normalize path usage to use forward slashes to make things more platform agnostic

### LINKED ISSUE #89: Nested terragrunt file not working on Windows
I am trying to use the nested terragrunt file on Windows. The directory structure is:

```
.terragrunt
env
  dev
    web
      .terragrunt
```

The root terragrunt file content.
```
lock = {
  backend = "dynamodb"
  config {
    state_file_id = "${path_relative_to_include()}"
    aws_region = "eu-west-1"
    table_name = "terragrunt_locks"
  }
}

remote_state = {
  backend = "s3"
  config {
    encrypt = "true"
    bucket = "tfstate"
    key = "${path_relative_to_include()}/terraform.tfstate"
    region = "eu-west-1"
  }
}
```

The content of terragrunt file in dev/web.
```
include = {
}
```

Running `terragrunt plan` in dev/web, I got the following error:
```
[terragrunt] 2016/12/27 15:56:03 Reading Terragrunt config file at C:\Users\userA\dev\test-aws\envs\dev\web\.terragrunt
[terragrunt] 2016/12/27 15:56:03 Error parsing Terragrunt config file C:\Users\userA\dev\test-aws\envs\dev\web\.terragrunt: At 2:20: illegal char escape
```

When I replace `${find_in_parent_folders()}` with hardcoded path: `../../../.terragrunt`, I got error further along.
```
[terragrunt] 2016/12/27 15:52:07 Reading Terragrunt config file at C:\Users\userA\dev\test-aws\envs\dev\web\.terragrunt
[terragrunt] 2016/12/27 15:52:07 Error parsing Terragrunt config file C:\Users\userA\dev\test-aws\envs\dev\web\.terragrunt: Error parsing Terragrunt config file C:\Users
\userA\dev\test-aws\.terragrunt: At 15:17: illegal char escape
```

Could this be another Windows issue?

cstavro:
For your consideration, this is a different approach that fixes #89 .

I've changed the relevant filepath calls to route through a helper method which calls filepath.ToSlash to convert them to forward slashes. 

The path/filepath pkg will convert slashes back and forth dependent upon the underlying OS so converting them all to forward slashes for the purposes of manipulating them within the application should be okay and also passes the tests!

I think this is a better approach than trying to explicitly handle two different path types throughout the app.

There are a few test failures around concurrency with dynamodb and deleting tables that are still in use.

### DIFF
--- a/util/file.go
+++ b/util/file.go
@@ -19,13 +19,14 @@ func FileExists(path string) bool {
 // components (e.g. "../") fully resolved, which makes it safe to compare paths as strings.
 func CanonicalPath(path string, basePath string) (string, error) {
 	if !filepath.IsAbs(path) {
-		path = filepath.Join(basePath, path)
+		path = JoinPath(basePath, path)
 	}
 	absPath, err := filepath.Abs(path)
 	if err != nil {
 		return "", err
 	}
-	return filepath.Clean(absPath), nil
+
+	return CleanPath(absPath), nil
 }
 
 // Return the canonical version of the given paths, relative to the given base path. That is, if a given path is a
@@ -83,7 +84,7 @@ func GetPathRelativeTo(path string, basePath string) (string, error) {
 		return "", errors.WithStackTrace(err)
 	}
 
-	return relPath, nil
+	return filepath.ToSlash(relPath), nil
 }
 
 // Return the contents of the file at the given path as a string
@@ -94,4 +95,12 @@ func ReadFileAsString(path string) (string, error) {
 	}
 
 	return string(bytes), nil
+}
+
+func JoinPath(elem ...string) string {
+	return filepath.ToSlash(filepath.Join(elem...))
+}
+
+func CleanPath(path string) string {
+	return filepath.ToSlash(filepath.Clean(path))
 }
--- a/util/file.go
+++ b/util/file.go
@@ -104,10 +104,14 @@ func ReadFileAsString(path string) (string, error) {
 	return string(bytes), nil
 }
 
+// Windows systems use \ as the path separator *nix uses /
+// Use this function when joining paths to force the returned path to use / as the path separator
+// This will improve cross-platform compatibility
 func JoinPath(elem ...string) string {
 	return filepath.ToSlash(filepath.Join(elem...))
 }
 
+// Use this function when cleaning paths to ensure the returned path uses / as the path separator to improve cross-platform compatibility
 func CleanPath(path string) string {
 	return filepath.ToSlash(filepath.Clean(path))
 }
--- a/util/file.go
+++ b/util/file.go
@@ -69,6 +69,13 @@ func Grep(regex *regexp.Regexp, glob string) (bool, error) {
 
 // Return the relative path you would have to take to get from basePath to path
 func GetPathRelativeTo(path string, basePath string) (string, error) {
+	if path == "" {
+		path = "."
+	}
+	if basePath == "" {
+		basePath = "."
+	}
+
 	inputFolderAbs, err := filepath.Abs(basePath)
 	if err != nil {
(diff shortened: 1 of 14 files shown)


cstavro:
Here's the latest test run with all the changes in this PR thus far.

[test_output.txt](https://github.com/gruntwork-io/terragrunt/files/704696/test_output.txt)

brikis98:
> test_output.txt

Almost all of these test failures have to do with concurrent modifications to DynamoDB. We generate a random name for the table and the odds of collision should be very low, and yet, the test output shows multiple tests using the exact same name. Here's the code used to generate a unique ID for the table:

```go
func uniqueId() string {
	const BASE_62_CHARS = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
	const UNIQUE_ID_LENGTH = 6 // Should be good for 62^6 = 56+ billion combinations

	var out bytes.Buffer

	random := rand.New(rand.NewSource(time.Now().UnixNano()))
	for i := 0; i < UNIQUE_ID_LENGTH; i++ {
		out.WriteByte(BASE_62_CHARS[random.Intn(len(BASE_62_CHARS))])
	}

	return out.String()
}
```

Perhaps `time.Now().UnixNano()` doesn't work properly on Windows and returns the same seed for everyone?

brikis98:
BTW, this is an interesting approach! Is the idea here to replace all path separators with a "/"? I assume Windows can handle forward slashes as well as backslashes when loading things from the file system?

cstavro:
> BTW, this is an interesting approach! Is the idea here to replace all path separators with a "/"? I assume Windows can handle forward slashes as well as backslashes when loading things from the file system?

Effectively, yes. I did some testing and have had no issues in windows with both slashes. Additionally, the path/filepath pkg in Go does the OS mangling for you already anyway. The path pkg on its own does not.

brikis98:
> Effectively, yes. I did some testing and have had no issues in windows with both slashes. Additionally, the path/filepath pkg in Go does the OS mangling for you already anyway. The path pkg on its own does not.

Awesome. In that case, I think this approach makes sense. The only thing left to figure out is why the DynamoDB stuff is failing, including even a panic in some cases. Do those tests pass if you run just one of them? E.g. `go test -v -run TestAcquireLockHappyPath`.

util/file.go
```diff
@@ -94,4 +102,12 @@ func ReadFileAsString(path string) (string, error) {
 	}
 
 	return string(bytes), nil
+}
+
+func JoinPath(elem ...string) string {
+	return filepath.ToSlash(filepath.Join(elem...))
+}
+
+func CleanPath(path string) string {
```
brikis98: Could you add comments to these two methods explaining what they do and why the should be used (i.e. for cross-OS compatibility).

cstavro:
> Awesome. In that case, I think this approach makes sense. The only thing left to figure out is why the DynamoDB stuff is failing, including even a panic in some cases. Do those tests pass if you run just one of them? E.g. go test -v -run TestAcquireLockHappyPath

They seem to, yes. So this definitely has something to do with the parallelization of the tests. I'll take a look at the uniqueid function.

cstavro:
Fixed!
I changed the way the seeding of rand works so it only initializes it once and things now seem to work swimmingly.
This fixes ALL the tests on Windows now.

[test_output.txt](https://github.com/gruntwork-io/terragrunt/files/705305/test_output.txt)

brikis98:
@cstavro You rock. This is awesome work! Thank you!

Merging now :)

brikis98:
Just created a new release here: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.9.0. If the build passes, the new binaries should show up in a few minutes.

PR #101, opened 2017-01-15
Read remote state file from specified working dir

### LINKED ISSUE #95: Remote state config not detected when using --terragrunt-working-dir
**Issue**
When calling terragrunt from the root of directory of project using the `--terragrunt-working-dir` CLI option to reference a sub directory, terragrunt does not see that the remote state is already configure for the S3 backend, and instead runs the `terraform remote config` command.

**Steps to Reproduce**
1. From the root of the project, run `terragrunt apply --terragrunt-workingdir=subdirectory`
2. Run `terragrunt output --terragrunt-working-dir=subdirectory`
Notice that terragrunt still calls `terraform remote config`, even though it was configured by the apply in step 1.
3. Change directory to `subdirectory` and run `terragrunt output`, and notice that terragrunt does not issue the call to `terraform remote config`

**Expected Behavior**
By supplying the `--terragrunt-working-dir` option, terragrunt should notice that the remote state has been configuted by looking in the working dir and not issue that command again

This came up when attempting to use the output from terragrunt as input elsewhere.  Moving terragrunt's output to stderr was very helpful, but the output that results from calling Terraform remote config muddies the waters here.

asobrien:
This changes `shouldConfigureRemoteState` in `remote_state.go` to always pass the `WorkingDir` option to `ParseTerraformStateFileFromLocation` in `terraform_state_file.go`. If `--terragrunt-working-dir` is specified the path to state files in that working dir should always take precedence over `DEFAULT_PATH_TO_LOCAL_STATE_FILE` and `DEFAULT_PATH_TO_REMOTE_STATE_FILE`.

Fixes #95.

### DIFF
--- a/remote/terraform_state_file.go
+++ b/remote/terraform_state_file.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"github.com/gruntwork-io/terragrunt/errors"
 	"fmt"
+	"path"
 	"github.com/gruntwork-io/terragrunt/util"
 )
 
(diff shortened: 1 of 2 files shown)


remote/terraform_state_file.go
```diff
@@ -44,10 +45,18 @@ func (state *TerraformState) IsRemote() bool {
...
 }
 
-// Parse the Terraform .tfstate file from its default locations. If the file doesn't exist at any of the default
-// locations, return nil.
-func ParseTerraformStateFileFromDefaultLocations() (*TerraformState, error) {
-	if util.FileExists(DEFAULT_PATH_TO_LOCAL_STATE_FILE) {
+// Parse the Terraform .tfstate file from the specified locations. If no location is specified,
+// search the default locations. If the file doesn't exist at any of the default locations, return nil.
+func ParseTerraformStateFileFromLocation(workingDir string) (*TerraformState, error) {
+	if workingDir != "" && util.FileExists(workingDir) {
```
brikis98: I don't think this check is necessary. If `workingDir` is empty, then `path.Join(workingDir, foo)` will just return `foo`. Moreover, `workingDir` should always be set, so that's not a particularly big concern.

In short, you can reduce this code back to the original if/else-if/else structure, but the paths in the if and else-if should both be `Join`ed with `workingDir`.

asobrien: 👍 Let's do that

asobrien:
I was concerned about the case of the `workingDir` not existing, but that's handled gracefully downstream by `ParseTerraformStateFile` in any case. 

@brikis98 042ae90 reverts this back to the original structure if/else-if/else structure.

brikis98:
Just created a new release here: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.9.2. If the build passes, the binaries should show up in a few minutes.

PR #103, opened 2017-01-20
Alternate messages when lock item or table doesn't exist

### LINKED ISSUE #16: Releasing a non-existing lock does not return a meaningful error message.
If you run `terragrunt release-lock` when a corresponding lock that doesn't exist, you get the same output as if the lock did exist.

asobrien:
Fixes #16 

When a non-existent lock is released via `terragrunt release-lock` this will emit a corresponding message indicating that:

```
[terragrunt]  Are you sure you want to release DynamoDB lock for state file .? (y/n) y
[terragrunt] 2017/01/19 20:58:40 Attempting to release lock for state file . in DynamoDB
[terragrunt] 2017/01/19 20:58:40 Nothing to release, specified lock does not exist in table terragrunt_lock_table
```

Additionally, if attempting to release a lock on a non-existent table, the message will reflect this:

```
[terragrunt]  Are you sure you want to release DynamoDB lock for state file .? (y/n) y
[terragrunt] 2017/01/19 20:56:46 Attempting to release lock for state file . in DynamoDB
[terragrunt] 2017/01/19 20:56:47 Table non_existent_table does not exist in DynamoDB!
```

### DIFF
--- a/locks/dynamodb/dynamo_lock.go
+++ b/locks/dynamodb/dynamo_lock.go
@@ -4,11 +4,12 @@ import (
 	"fmt"
 	"strconv"
 
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/dynamodb"
+	"github.com/gruntwork-io/terragrunt/aws_helper"
 	"github.com/gruntwork-io/terragrunt/errors"
 	"github.com/gruntwork-io/terragrunt/locks"
-	"github.com/gruntwork-io/terragrunt/aws_helper"
 	"github.com/gruntwork-io/terragrunt/options"
 )
 
@@ -86,8 +87,19 @@ func (dynamoDbLock DynamoDbLock) ReleaseLock(terragruntOptions *options.Terragru
 		return err
 	}
 
-	if err := removeItemFromLockTable(dynamoDbLock.StateFileId, dynamoDbLock.TableName, client); err != nil {
-		return err
+	err = removeItemFromLockTable(dynamoDbLock.StateFileId, dynamoDbLock.TableName, client)
+
+	// If the table or lock doesn't exist log it, else return the error
+	if err != nil {
+		if awsErr, isAwsErr := errors.Unwrap(err).(awserr.Error); isAwsErr && awsErr.Code() == "ResourceNotFoundException" {
+			terragruntOptions.Logger.Printf("Table %s does not exist in DynamoDB!", dynamoDbLock.TableName)
+			return nil
+		} else if awsErr, isAwsErr := errors.Unwrap(err).(awserr.Error); isAwsErr && awsErr.Code() == "ConditionalCheckFailedException" {
+			terragruntOptions.Logger.Printf("Nothing to release, specified lock does not exist in table %s", dynamoDbLock.TableName)
+			return nil
+		} else {
+			return err
+		}
 	}
 
 	terragruntOptions.Logger.Printf("Lock released!")
--- a/locks/dynamodb/dynamo_lock.go
+++ b/locks/dynamodb/dynamo_lock.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"strconv"
 
-	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/dynamodb"
 	"github.com/gruntwork-io/terragrunt/aws_helper"
@@ -87,19 +86,8 @@ func (dynamoDbLock DynamoDbLock) ReleaseLock(terragruntOptions *options.Terragru
 		return err
 	}
 
-	err = removeItemFromLockTable(dynamoDbLock.StateFileId, dynamoDbLock.TableName, client)
-
-	// If the table or lock doesn't exist log it, else return the error
-	if err != nil {
-		if awsErr, isAwsErr := errors.Unwrap(err).(awserr.Error); isAwsErr && awsErr.Code() == "ResourceNotFoundException" {
-			terragruntOptions.Logger.Printf("Table %s does not exist in DynamoDB!", dynamoDbLock.TableName)
-			return nil
-		} else if awsErr, isAwsErr := errors.Unwrap(err).(awserr.Error); isAwsErr && awsErr.Code() == "ConditionalCheckFailedException" {
-			terragruntOptions.Logger.Printf("Nothing to release, specified lock does not exist in table %s", dynamoDbLock.TableName)
-			return nil
-		} else {
-			return err
-		}
+	if err := removeItemFromLockTable(dynamoDbLock.StateFileId, dynamoDbLock.TableName, client); err != nil {
+		return err
 	}
 
 	terragruntOptions.Logger.Printf("Lock released!")
--- a/locks/dynamodb/dynamo_lock_table.go
+++ b/locks/dynamodb/dynamo_lock_table.go
@@ -1,14 +1,14 @@
 package dynamodb
 
 import (
-	"time"
 	"fmt"
-	"github.com/aws/aws-sdk-go/service/dynamodb"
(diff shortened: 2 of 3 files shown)


locks/dynamodb/dynamo_lock.go
```diff
@@ -86,8 +87,19 @@ func (dynamoDbLock DynamoDbLock) ReleaseLock(terragruntOptions *options.Terragru
...
+	if err != nil {
+		if awsErr, isAwsErr := errors.Unwrap(err).(awserr.Error); isAwsErr && awsErr.Code() == "ResourceNotFoundException" {
+			terragruntOptions.Logger.Printf("Table %s does not exist in DynamoDB!", dynamoDbLock.TableName)
+			return nil
+		} else if awsErr, isAwsErr := errors.Unwrap(err).(awserr.Error); isAwsErr && awsErr.Code() == "ConditionalCheckFailedException" {
+			terragruntOptions.Logger.Printf("Nothing to release, specified lock does not exist in table %s", dynamoDbLock.TableName)
+			return nil
+		} else {
+			return err
+		}
```
brikis98: Could you extract this error-handling code within the `if err != nil` check into a separate method? The method could be called something like `handleRemoveItemFromLockTableError`.

Also, it seems strange to return `nil` in the case of recognized AWS errors. Why not wrap it with a clear error message? For example, you could declare:

```go
type TableDoesNotExist struct {
    TableName string
    Underlying  error
}
func (err TableDoesNotExist) Error() string {
  return fmt.Sprintf("Table %s does not exist in DynamoDB! Original error from AWS: %v", err.TableName, err.Underlying)
}
```

You could then return `TableDoesNotExist{dynamoDbLock.TableName, awsErr}` from the `handleRemoveItemFromLockTableError` method, and a nice, clean error message will be shown for you automatically rather than having to log it separately.

locks/dynamodb/dynamo_lock_table.go
```diff
@@ -129,9 +129,19 @@ func waitForTableToBeActiveWithRandomSleep(tableName string, client *dynamodb.Dy
...
+		return errors.WithStackTrace(err)
+	}
+
+	// Unless you specify conditions, the DeleteItem is an idempotent operation;
+	// running it multiple times on the same item or attribute does not result in an error response.
+	// https://docs.aws.amazon.com/sdk-for-go/api/service/dynamodb/#DynamoDB.DeleteItem
+	_, err = client.DeleteItem(&dynamodb.DeleteItemInput{
+		Key:                 createKeyFromItemId(itemId),
+		TableName:           aws.String(tableName),
+		ConditionExpression: aws.String(fmt.Sprintf("attribute_exists(%s)", ATTR_STATE_FILE_ID)),
```
brikis98: Is it possible to specify a `ConditionExpression` that checks if the table exist so we don't have to do a separate check in the previous line?

locks/dynamodb/dynamo_lock.go
brikis98: Also, if you do return the custom error type, you could move this code into the `removeItemFromLockTable` code itself so all callers benefit from better error handling.

locks/dynamodb/dynamo_lock_table_test.go
```diff
@@ -106,6 +107,22 @@ func TestWriteAndRemoveItemFromLockTable(t *testing.T) {
...
+			assert.Equal(t, awsErr.Code(), "ConditionalCheckFailedException", "Unexpected awsErr code")
+		}
+
+		// And then, verify removing an item from a non-existent table returns an expected error
+		err = removeItemFromLockTable(itemId, uniqueTableNameForTest(), client)
+		if assert.NotNil(t, err, "Error expected") {
+			awsErr, isAwsErr := errors.Unwrap(err).(awserr.Error)
+			assert.True(t, isAwsErr, "Expected error to be an awserr error")
+			assert.Equal(t, awsErr.Code(), "ResourceNotFoundException", "Unexpected awsErr code")
+		}
```
brikis98: If you update the `removeItemFromLockTable` to return the custom error type as mentioned above, then this test code could verify that it gets those clear error messages back rather than parsing the AWS errors here a second time.

locks/dynamodb/dynamo_lock_table.go
asobrien: There's no `ConditionExpression` function to check whether the table exists, but a `ResourceNotFoundException` error will be returned if the table doesn't exist. The previous check is redundant, I'll remove it.

asobrien:
I think using custom errors and a separate error method make things a lot clearer.

asobrien:
@brikis98 Thanks for comments, I think these commits should address them. 

I moved the error handling for AWS errors resulting from `removeItemFromLockTable` directly into that function. I think it makes much more sense to do it this way, as you had suggested, since these errors are now available to all callers. Also, the redundant call to check for the existence of the table was removed.

With custom errors returned from `removeItemFromLockTable`, rather than suppressing the error as was done previously, this allows the error to bubble up and return a non-zero exit code. For instance:

```
$ terragrunt release-lock
    ...
[terragrunt]  Are you sure you want to release DynamoDB lock for state file .? (y/n) y
[terragrunt] 2017/01/22 14:04:16 Attempting to release lock for state file . in DynamoDB
Item . does not exist in terragrunt_locks DynamoDB table! Original error from AWS: ConditionalCheckFailedException: The conditional request failed
	status code: 400, request id: AIKS88***************************************
[terragrunt] 2017/01/22 14:04:16 Item . does not exist in terragrunt_locks DynamoDB table! Original error from AWS: ConditionalCheckFailedException: The conditional request failed
	status code: 400, request id: AIKS88***************************************
# exit code 1
```

Do you expect that `release-lock` should return non-zero exit codes in the case of an `Item` or `Table` not existing in dynamodb? I would think so. 

It's outside the scope of this PR, but It seems like all errors are getting duplicated since `urfave/cli` will print any error returned by `app.Run` to stderr, and terragrunt handles and logs the error in `checkForErrorsAndExit`. I have some thoughts on how to suppress terragrunt internal errors from being printed by `cli` and having them handled by the terragrunt logger to keep behavior consistent.

As for test output, do you want me to dump the results of the full test suite or just the relevant tests? Here's the output of from the test that was modified:

```
$ go test -v -parallel 128 -run TestWriteAndRemoveItemFromLockTable
[terragrunt] 2017/01/22 14:22:30 Lock table terragrunt_test_R5iMpW does not exist in DynamoDB. Will need to create it just this first time.
[terragrunt] 2017/01/22 14:22:30 Creating table terragrunt_test_R5iMpW in DynamoDB
[terragrunt] 2017/01/22 14:22:30 Table terragrunt_test_R5iMpW is not yet in active state. Will check again after 10s.
[terragrunt] 2017/01/22 14:22:40 Table terragrunt_test_R5iMpW is not yet in active state. Will check again after 10s.
[terragrunt] 2017/01/22 14:22:50 Success! Table terragrunt_test_R5iMpW is now in active state.
--- PASS: TestWriteAndRemoveItemFromLockTable (21.01s)
PASS
ok  	github.com/gruntwork-io/terragrunt/locks/dynamodb	21.017s
```

brikis98:
> Do you expect that release-lock should return non-zero exit codes in the case of an Item or Table not existing in dynamodb? I would think so.

Yup!

> It's outside the scope of this PR, but It seems like all errors are getting duplicated since urfave/cli will print any error returned by app.Run to stderr, and terragrunt handles and logs the error in checkForErrorsAndExit. I have some thoughts on how to suppress terragrunt internal errors from being printed by cli and having them handled by the terragrunt logger to keep behavior consistent.

Sounds like a bug. If you have an idea how to fix that, another PR would be most welcome :)

> As for test output, do you want me to dump the results of the full test suite or just the relevant tests? 

I ask for it as a sanity check that you did, in fact, run the tests. I can't rely on Circle CI to do it for PRs due to security reasons. Of course, I'm going to merge now and all the tests will run anyway, so we'll see if things are still working shortly.

Thank you for contributing!

brikis98:
I just created a new release here: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.9.3. If the build passes, the new binaries should show up shortly.

Issue #107, opened 2017-01-24
Migrating to Terragrunt when previously using the -state option


trevorr:
My Terraform state file is currently stored separately from my templates, because the state file is specific to the environment (e.g. production vs. QA/staging) but the templates are not (generally speaking). Therefore, when I run Terraform with local state, my scripts always specify the `-state` option.

I'm now trying to migrate to remote state using Terragrunt. When Terragrunt runs `terraform remote config`, there doesn't seem to be a way to pass the [`-state`](https://www.terraform.io/docs/commands/remote-config.html#_state_path) option such that it's used by `remote config` but not by the actual command. This results in a Terraform error:

```
export TERRAGRUNT_CONFIG=/opt/deploy/state/dev/.terragrunt
cd /opt/deploy/generated/dev && terragrunt refresh -state=/opt/deploy/state/dev/terraform.tfstate -var-file=dev.tfvars
[terragrunt] 2017/01/24 09:58:10 Reading Terragrunt config file at /opt/deploy/state/dev/.terragrunt
[terragrunt] 2017/01/24 09:58:10 Initializing remote state for the s3 backend
[terragrunt] 2017/01/24 09:58:10 Configuring remote state for the s3 backend
[terragrunt] 2017/01/24 09:58:10 Running command: terraform remote config -backend s3 -backend-config=key=terraform.tfstate -backend-config=encrypt=false -backend-config=region=us-east-1 -backend-config=bucket=my-bucket
Remote configuration updated
Remote state configured and pulled.
[terragrunt] 2017/01/24 09:58:13 Attempting to acquire lock for state file dev in DynamoDB
[terragrunt] 2017/01/24 09:58:13 Attempting to create lock item for state file dev in DynamoDB table terragrunt_locks
[terragrunt] 2017/01/24 09:58:14 Lock acquired!
[terragrunt] 2017/01/24 09:58:14 Running command: terraform refresh -state=/opt/deploy/state/dev/terraform.tfstate -var-file=dev.tfvars
Failed to load state: Remote state found, but state file '/opt/deploy/state/dev/terraform.tfstate' also present.
[terragrunt] 2017/01/24 09:58:15 Attempting to release lock for state file dev in DynamoDB
[terragrunt] 2017/01/24 09:58:15 Lock released!
exit status 1
[terragrunt] 2017/01/24 09:58:15 exit status 1
```

[...]

PR #109, opened 2017-01-25
Allow explicitly setting up the remote state

Bowbaq:
This adds a new `setup-state` command which basically only runs `configureRemoteState`. This allows explicitly pulling down the remote state (in a CI environment for example) without other side effects.

Our use case is that we use some values from the state to configure the application layer (DNS names, things like that). Running `terragrunt output <some output>` works, however it writes a bunch of "logs" to `stdout`, so a script doing something like `DNS_NAME="$(terragrunt output dns_name)" ends up with unexpected values.

We currently run a `terragrunt validate` beforehand to pull down the state, but it feels a bit hacky.

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -22,6 +22,7 @@ const OPT_WORKING_DIR = "terragrunt-working-dir"
 var ALL_TERRAGRUNT_BOOLEAN_OPTS = []string{OPT_NON_INTERACTIVE}
 var ALL_TERRAGRUNT_STRING_OPTS = []string{OPT_TERRAGRUNT_CONFIG, OPT_TERRAGRUNT_TFPATH, OPT_WORKING_DIR}
 
+const CMD_SETUP_STATE = "setup-state"
 const CMD_ACQUIRE_LOCK = "acquire-lock"
 const CMD_RELEASE_LOCK = "release-lock"
 const CMD_SPIN_UP = "spin-up"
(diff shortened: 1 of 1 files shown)


brikis98:
>  Running `terragrunt output <some output>` works, however it writes a bunch of "logs" to `stdout`, so a script doing something like `DNS_NAME="$(terragrunt output dns_name)` ends up with unexpected values.

What version of Terragrunt are you using? As of [v0.8.0](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.8.0), all of Terragrunt's logs go to `stderr`. Or are you saying that the logs from Terraform's `remote config` command are getting in the way?

Bowbaq:
It's the latter. I guess another solution would be to fix that instead.

brikis98:
If you redirect the `remote config` logs to `stderr` when Terragrunt is executing it, will that fix your issue? If so, that might be simpler than adding (and then maintaining) a whole new command.

Bowbaq:
It would fix the issue. I still find it strange that there is no way to explicitly setup the remote state without side effects (unless I overlooked one?)

brikis98:
I agree there should be some sort of `init` command for Terragrunt. It came up in #107 as well. See my [latest comment](https://github.com/gruntwork-io/terragrunt/issues/107#issuecomment-274973923) for some ideas on what that command could look like.

brikis98:
Since #110 was merged, I'm going to close this. I'm still open to PRs for some sort of `terragrunt-init` command as discussed in #107.

PR #110, opened 2017-01-25
Redirect the output of `terraform remote config` to stderr

Bowbaq:
@brikis98 alternative solution to #109

### DIFF
--- a/shell/run_shell_cmd.go
+++ b/shell/run_shell_cmd.go
@@ -5,14 +5,16 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
+	"reflect"
 	"strings"
+
 	"github.com/gruntwork-io/terragrunt/errors"
 	"github.com/gruntwork-io/terragrunt/options"
 )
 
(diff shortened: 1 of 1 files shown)


brikis98:
Thanks for the PR!

From a look at the code, it's not obvious why we're redirecting that output. The real reason is that we run `remote config` before running other commands, so instead of your `stdout` containing the text from the command you wanted, it also has all the `remote config` output. Perhaps a cleaner and more general way to express that is: 

1. Compare `terragruntOptions.TerraformCliArgs` and the `args` parameter of the `RunShellCommand` method and to redirect `stdout` to `stderr` when those two are not equal.
1. Add a comment explaining this

Bowbaq:
@brikis98 took another crack at it

brikis98:
LGTM, thanks!

brikis98:
I just created a new release here: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.9.6. If the build passes, the new binaries should show up there in a few minutes.

PR #111, opened 2017-01-25
Add ability for Terragrunt to download Terraform configurations

brikis98:
This PR adds a new element you can specify in the `.terragrunt` file:

```hcl
terraform {
  source = "git::git@github.com:foo/bar.git//my-terraform-configurations?ref=v0.0.1"
}
```

When you specify this parameter, Terragrunt will download the Terraform configurations at the `source` URL to a temporary folder and run Terraform in that temporary folder. 

This makes it easy to version all of your Terraform configurations and deploy a specific version automatically, similar to Kief Morris’ [pipelines idea](https://medium.com/@kief/https-medium-com-kief-using-pipelines-to-manage-environments-with-infrastructure-as-code-b37285a1cbf5). 

You can also override the `source` parameter with the `--terragrunt-source` command-line flag and `TERRAGRUNT_SOURCE` environment variable, which makes it easy to use a local checkout for iterative development.

See the Readme for full details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -783,9 +783,10 @@ prefix `--terragrunt-`. The currently available options are:
   current working directory. Note that for the `spin-up` and `tear-down` directories, this parameter has a different 
   meaning: Terragrunt will apply or destroy all the Terraform modules in the subfolders of the 
   `terragrunt-working-dir`, running `terraform` in the root of each module it finds.
-* `--terragrunt-source`: Check out templates from the specified source, which uses the [module 
-  source](https://www.terraform.io/docs/modules/sources.html) syntax, into a temporary folder, and run Terraform in
-  that temporary folder. May also be specified via the `TERRAGRUNT_SOURCE` environment variable.  
+* `--terragrunt-source`: Download Terraform configurations from the specified source into a temporary folder, and run 
+  Terraform in that temporary folder. May also be specified via the `TERRAGRUNT_SOURCE` environment variable. The 
+  source should use the same syntax as the [Terraform module source](https://www.terraform.io/docs/modules/sources.html) 
+  parameter.  
 
 ## Developing terragrunt
 
--- a/README.md
+++ b/README.md
@@ -682,9 +682,7 @@ infrastructure-live
         └ vars.tf
         └ outputs.tf
     └ backend-app
-    └ search-app
     └ mysql
-    └ redis
     └ vpc
   └ prod
     └ frontend-app
@@ -692,9 +690,7 @@ infrastructure-live
         └ vars.tf
         └ outputs.tf
     └ backend-app
-    └ search-app
     └ mysql
-    └ redis
     └ vpc
 ```
 
@@ -716,9 +712,7 @@ infrastructure-modules
       └ vars.tf
       └ outputs.tf
   └ backend-app
-  └ search-app
   └ mysql
-  └ redis
   └ vpc
 ```
 
@@ -756,18 +750,14 @@ infrastructure-live
         └ .terragrunt
         └ terraform.tfvars
     └ backend-app
-    └ search-app
     └ mysql
-    └ redis
     └ vpc
   └ prod
     └ frontend-app
         └ .terragrunt
         └ terraform.tfvars
     └ backend-app
-    └ search-app
     └ mysql
-    └ redis
     └ vpc
 ```
  
--- a/README.md
+++ b/README.md
@@ -669,6 +669,105 @@ you've fixed the error, it's usually safe to re-run the `spin-up` or `tear-down`
 for the modules that already deployed successfully, and should only affect the ones that had an error the last time
 around.
 
+## Remote Terraform configurations
+
+Terragrunt has the ability to work with Terraform configurations that are not in your working directory. This is 
+especially useful if you have versioned all of your Terraform code and want to `apply` a specific version of it
+(see [Using Pipelines to Manage Environments with Infrastructure as 
+Code](https://medium.com/@kief/https-medium-com-kief-using-pipelines-to-manage-environments-with-infrastructure-as-code-b37285a1cbf5)).
+
+In your `.terragrunt` file, you can specify a `terraform` block with a `source` parameter that tells Terragrunt where 
+to find the Terraform configuration files:
+
+```hcl
+terraform {
+  source = "git::git@github.com:foo/bar.git//my-terraform-configurations?ref=v0.0.1"
+}
+```
+
+When Terragrunt finds a `terraform` block, it will:
+ 
+1. Download the configurations specified via the `source` parameter into a temporary folder. This is done using the
+   [terraform init command](https://www.terraform.io/docs/commands/init.html), so the `source` parameter supports the
+   exact same syntax as the [module source](https://www.terraform.io/docs/modules/sources.html) parameter, including
+   local file paths, Git URLs, and Git URLs with `ref` parameters (useful for checking out a specific tag, commit, or
+   branch of Git repo).
+1. Copy all files from the current working directory into the temporary folder. This is useful if you have `.tfvars`
+   files in your local path that you are using to set variables within your templates.
+1. Execute whatever Terraform command you specified in that temporary folder. **Note**: if you are passing any file
+   paths (other than paths to files in the current working directory) to Terraform via command-line options, those 
+   paths must be absolute paths since we will be running Terraform from the temporary folder!
+
+With this approach, you could define all of your Terraform code in one repo called, for example, 
+infrastructure-modules:
+
+```
+infrastructure-modules
+  └ frontend-app
+      └ main.tf
+      └ vars.tf
+      └ outputs.tf
+  └ backend-app
+  └ search-app
+  └ mysql
+  └ redis
+  └ vpc
+```
+
+You could then use this infrastructure in a separate repo called, for example, infrastructure-live, as follows:
+
+```
+infrastructure-live
+  └ stage
+    └ frontend-app
+        └ .terragrunt
+        └ terraform.tfvars
+    └ backend-app
+    └ search-app
+    └ mysql
+    └ redis
+    └ vpc
+  └ prod
+    └ frontend-app
+        └ .terragrunt
+        └ terraform.tfvars
+    └ backend-app
+    └ search-app
+    └ mysql
+    └ redis
+    └ vpc
+```
+
+Each `.terragrunt` file would set the `source` parameter to a specific version of the relevant module in 
+infrastructure-modules:
+
+```hcl
+terraform {
+  source = "git::git@github.com:my-company/infrastructure-modules.git//frontend-app?ref=v0.0.1"
+}
+```
+
+The `terraform.tfvars` files would provide all the variables needed for that module:
+
+```hcl
+app_name = "my-frontend-app"
+num_instances = 5
+instance_type = "m4.large"
+```
+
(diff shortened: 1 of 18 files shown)


brikis98:
@josh-padnick Please take a look when you have a chance. I'd like to merge this in today.

brikis98:
Also, after using this for a bit with a real-world use case, here are a few follow-up tasks:

1. Instead of downloading the code into a new tmp folder each time, find a way to reuse an existing folder and intelligently update it. Otherwise, if your code includes lots of modules, having to run `terraform get` from scratch each time can be slow.
1. Find some way to override the `source` parameter for the `spin-up` and `tear-down` commands so you can spin up an entire stack from a local checkout.
1. Consider if Terragrunt should be able to read its configuration not only from a `.terragrunt` file, but also a `terragrunt` key within a `terraform.tfvars` file: 

    ```hcl
    # Normal params in terraform.tfvars
    foo = "bar"

    # Terragrunt params
    terragrunt {
      lock { ... }
      remote_state { ... }
    }
    ```

    The advantage of that is that you have one less file and folder to manage for each component:

    ```
    infrastructure-live
      └ stage
        └ frontend-app.tfvars
        └ backend-app.tfvars
        └ search-app.tfvars
        └ mysql.tfvars
        └ redis.tfvars
        └ vpc.tfvars
      └ prod
        └ frontend-app.tfvars
        └ backend-app.tfvars
        └ search-app.tfvars
        └ mysql.tfvars
        └ redis.tfvars
        └ vpc.tfvars
    ```

    Moreover, instead of worrying about a special "root" `.terragrunt` file, you can a root `.tfvars` file with shared configs (or even multiple shared `.tfvars` files), as Terraform merges configs together automatically.

josh-padnick:
Looking now...

README.md
```diff
@@ -669,6 +669,105 @@ you've fixed the error, it's usually safe to re-run the `spin-up` or `tear-down`
 for the modules that already deployed successfully, and should only affect the ones that had an error the last time
 around.
 
+## Remote Terraform configurations
+
+Terragrunt has the ability to work with Terraform configurations that are not in your working directory. This is 
+especially useful if you have versioned all of your Terraform code and want to `apply` a specific version of it
+(see [Using Pipelines to Manage Environments with Infrastructure as 
+Code](https://medium.com/@kief/https-medium-com-kief-using-pipelines-to-manage-environments-with-infrastructure-as-code-b37285a1cbf5)).
+
```
josh-padnick: This might be hard to understand as someone's first into to this concept. Consider first describing the problem that we're solving (e.g. "A common gripe of Terragrunt users is that launching a new environment is cumbersome. Much copy and pasting is required, and it can be cumbersome to identify the specific vars that change between environments. One approach to solving this is to...")

README.md
```diff
@@ -669,6 +669,105 @@ you've fixed the error, it's usually safe to re-run the `spin-up` or `tear-down`
...
+In your `.terragrunt` file, you can specify a `terraform` block with a `source` parameter that tells Terragrunt where 
+to find the Terraform configuration files:
+
+```hcl
+terraform {
+  source = "git::git@github.com:foo/bar.git//my-terraform-configurations?ref=v0.0.1"
+}
+```
```
josh-padnick: +1 on this syntax, which mirrors Terraform's.

README.md
```diff
@@ -669,6 +669,105 @@ you've fixed the error, it's usually safe to re-run the `spin-up` or `tear-down`
...
+
+When Terragrunt finds a `terraform` block, it will:
+ 
+1. Download the configurations specified via the `source` parameter into a temporary folder. This is done using the
+   [terraform init command](https://www.terraform.io/docs/commands/init.html), so the `source` parameter supports the
+   exact same syntax as the [module source](https://www.terraform.io/docs/modules/sources.html) parameter, including
+   local file paths, Git URLs, and Git URLs with `ref` parameters (useful for checking out a specific tag, commit, or
+   branch of Git repo).
```
josh-padnick: Nice use of `terraform init`. Didn't even realize this command existed!

README.md
```diff
@@ -669,6 +669,105 @@ you've fixed the error, it's usually safe to re-run the `spin-up` or `tear-down`
...
+  └ frontend-app
+      └ main.tf
+      └ vars.tf
+      └ outputs.tf
+  └ backend-app
+  └ search-app
+  └ mysql
+  └ redis
+  └ vpc
+```
```
josh-padnick: This isn't the ideal place for this comment, but it occurs to me that this infra-modules folder structure should use folder groupings like:

```
- services
  - frontend-app
    - main.tf
  - backend-app
- data-stores
  - mysql
  - redis
```

README.md
```diff
@@ -669,6 +669,105 @@ you've fixed the error, it's usually safe to re-run the `spin-up` or `tear-down`
...
+  └ prod
+    └ frontend-app
+        └ .terragrunt
+        └ terraform.tfvars
+    └ backend-app
+    └ search-app
+    └ mysql
+    └ redis
+    └ vpc
+```
```
josh-padnick: Love this folder structure. I agree with your follow-up comment that combining `.terragrunt` and `terraform.tfvars` removes the need for a folder and keeps things simpler.

README.md
```diff
@@ -669,6 +669,105 @@ you've fixed the error, it's usually safe to re-run the `spin-up` or `tear-down`
...
+instance_type = "m4.large"
+```
+
+Now, all of your infrastructure is versioned, and there is no duplication of Terraform code to use it across many
+environments. You just run `terragrunt apply` and Terragrunt will check out the specified version of the code and run
+`terraform apply` on it.
+
+Note that you can also use the `--terragrunt-source` command-line option and `TERRAGRUNT_SOURCE` environment variable
+to override the `source` parameter. This is useful to point Terragrunt at a local checkout of your code so you can do 
+rapid, iterative, make-a-change-and-rerun development:
```
josh-padnick: +1

README.md
```diff
@@ -669,6 +669,105 @@ you've fixed the error, it's usually safe to re-run the `spin-up` or `tear-down`
...
+   
+```
+cd infrastructure-live/stage/frontend-app
+terragrunt apply --terragrunt-source ../../../infrastructure-modules/frontend-app
+```   
+
```
josh-padnick: First, this is fantastic. I am really excited to start playing around with this.

Second, just a sanity check. So, let's say one of our customers is using our vpc modules. Then here's what the chain of calls looks like:

- client/infrastructure-live: `vpc`
- client/infrastructure-modules: `vpc`
- gruntwork/module-vpc: `vpc-app`

So to state the obvious, this adds one more layer to our module calls than we had before. But I don't really see any downsides to this. Of course, this does raise the question of how we handle our traditional infrastructure-modules, like `ecs-cluster`. For example:

- client/infrastructure-live: `stage/ecs-cluster`
- client/infrastructure-modules: `ecs-cluster`
- client/infrastructure-modules: `modules/ecs-cluster`
- gruntwork/module-ecs: `ecs-cluster`

It means one infrastructure-module might want to make a versioned reference to another infrastructure-module. How are you addressing that? Might be better for offline or out-of-band discussion since it doesn't directly pertain to this PR.

cli/cli_app.go
```diff
@@ -124,6 +127,18 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
...
+		if err := checkoutTerraformSource(terragruntOptions.Source, terragruntOptions); err != nil {
+			return err
+		}
+	}
+
+	if terragruntOptions.Source == "" && conf.Terraform != nil && len(conf.Terraform.Source) > 0 {
+		if err := checkoutTerraformSource(conf.Terraform.Source, terragruntOptions); err != nil {
+			return err
+		}
+	}
```
josh-padnick: Can you explain what condition this is meant to handle?

README.md
brikis98: Good point. I just updated the README to make the problem clearer: https://github.com/gruntwork-io/terragrunt/pull/111/commits/cd3f23b497d5ba903a774ec4bd8fb5dae0ef2a77

README.md
brikis98: > So to state the obvious, this adds one more layer to our module calls than we had before.

It's not *really* one more layer. There is no Terraform code in `infrastructure-live`. The only Terraform code is in `infrastructure-modules`, which has a module dependency on `module-vpc`. The `infrastructure-live` repo just fills in the environment-specific variables for the `infrastructure-modules` code.

> Of course, this does raise the question of how we handle our traditional infrastructure-modules, like ecs-cluster. 

In your diagram, you had two entries for `ecs-cluster` in `infrastructure-modules`, but we'd really only have one. So in that regard, this code is simplified.

cli/cli_app.go
brikis98: There are two ways to tell Terragrunt that it needs to download some Terraform templates:

1. The `--terragrunt-source` command-line option.
1. A `terraform` block in `.terragrunt`. 

These if-statements pick them out in proper priority order. That said, this code was a bit convoluted for that, so I've refactored it as follows: https://github.com/gruntwork-io/terragrunt/pull/111/commits/8ecd053a8ba9d3d19d41aafc77eb0aa962238baf

brikis98:
OK, I'm going to merge this one in.

README.md
```diff
@@ -671,34 +671,42 @@ around.
...
-1. Copy all files from the current working directory into the temporary folder. This is useful if you have `.tfvars`
-   files in your local path that you are using to set variables within your templates.
-1. Execute whatever Terraform command you specified in that temporary folder. **Note**: if you are passing any file
-   paths (other than paths to files in the current working directory) to Terraform via command-line options, those 
-   paths must be absolute paths since we will be running Terraform from the temporary folder!
+For each environment, you have to copy/paste `main.tf`, `vars.tf`, and `outputs.tf` for each component (e.g. 
+frontend-app, backend-app, vpc, etc). As the number of components and environments grows, having to maintain more and 
+more code can become error prone. You can significantly reduce the amount of copy paste using [Terraform 
+modules](https://blog.gruntwork.io/how-to-create-reusable-infrastructure-with-terraform-modules-25526d65f73d), but even 
+the code to use the module module and set up input variables, output variables, providers, and remote state can still
```
josh-padnick: s/module module/module.

README.md
```diff
@@ -737,29 +770,41 @@ infrastructure-live
...
+Just as importantly, since the Terraform code is now defined in a single repo, you can version it (e.g. using Git
+tags and referencing them using the `ref` parameter in the `source` URL), and promote a single, immutable version 
+through each environment (e.g. qa -> stage -> prod). This idea is inspired by Kief Morris' blog post [Using Pipelines 
+to Manage Environments with Infrastructure as 
+Code](https://medium.com/@kief/https-medium-com-kief-using-pipelines-to-manage-environments-with-infrastructure-as-code-b37285a1cbf5)).
 
-Note that you can also use the `--terragrunt-source` command-line option and `TERRAGRUNT_SOURCE` environment variable
+Note that you can also use the `--terragrunt-source` command-line option or the `TERRAGRUNT_SOURCE` environment variable
 to override the `source` parameter. This is useful to point Terragrunt at a local checkout of your code so you can do 
 rapid, iterative, make-a-change-and-rerun development:
```
josh-padnick: These docs are great.

README.md
brikis98: Fixed that in a [follow-up commit](https://github.com/gruntwork-io/terragrunt/pull/111/commits/796fae2f95265ffd0dca77d721baa067bcce41dd) :)

PR #114, opened 2017-01-29
Cache temporary folder

brikis98:
In https://github.com/gruntwork-io/terragrunt/pull/111, I added the ability for Terragrunt to download Terraform files from a given `source` URL. The implementation in that PR downloaded the files to a different temp folder every time. As a result, it also had to run `terraform get` and `terraform remote config` from scratch every time. All together, this added a significant amount of overhead to every single Terraform command.

In this PR, I reuse the same temp folder for a given source URL. The exact caching logic is a bit complicated, as there are different cases to handle for local source URLs vs remote source URLs vs remote source URLs with version numbers (e.g. `ref=v0.0.3`). The result is you only have to download the source code once on your computer. After that, everything is exactly as fast as the normal way of using Terraform.

### DIFF
--- a/cli/download_source.go
+++ b/cli/download_source.go
@@ -14,6 +14,7 @@ import (
 	"fmt"
 )
 
+// This struct represents information about Terraform source code that needs to be downloaded
 type TerraformSource struct {
 	// A canonical version of RawSource, in URL format
 	CanonicalSourceURL *url.URL
--- a/cli/download_source.go
+++ b/cli/download_source.go
@@ -1,0 +1,95 @@
+package cli
+
+import (
+	"github.com/gruntwork-io/terragrunt/options"
+	"github.com/gruntwork-io/terragrunt/util"
+	"github.com/gruntwork-io/terragrunt/config"
+	"os"
+	"github.com/gruntwork-io/terragrunt/errors"
+	"path/filepath"
+)
+
+// 1. Check out the given source URL, which should use Terraform's module source syntax, into a temporary folder
+// 2. Copy the contents of terragruntOptions.WorkingDir into the temporary folder.
+// 3. Set terragruntOptions.WorkingDir to the temporary folder.
+func downloadTerraformSource(source string, terragruntOptions *options.TerragruntOptions) error {
+	tmpFolder, err := prepareTempDir(terragruntOptions)
+	if err != nil {
+		return err
+	}
+
+	terragruntOptions.Logger.Printf("Downloading Terraform configurations from %s into %s", source, tmpFolder)
+	if err := terraformInit(source, tmpFolder, terragruntOptions); err != nil {
+		return err
+	}
+
+	terragruntOptions.Logger.Printf("Copying files from %s into %s", terragruntOptions.WorkingDir, tmpFolder)
+	if err := util.CopyFolderContents(terragruntOptions.WorkingDir, tmpFolder); err != nil {
+		return err
+	}
+
+	terragruntOptions.Logger.Printf("Setting working directory to %s", tmpFolder)
+	terragruntOptions.WorkingDir = tmpFolder
+
+	return nil
+}
+
+// Prepare a temp folder into which Terragrunt can download Terraform code. We take the canonical path of the current
+// working directory and create a folder under the system temporary folder with the same path. This allows us to reuse
+// the same temp folder for a given working directory so that if you run Terragrunt from the same folder multiple
+// times, we only have to download modules and configure remote state once (however, we download the source code every
+// time).
+func prepareTempDir(terragruntOptions *options.TerragruntOptions) (string, error) {
+	canonicalPath, err := util.CanonicalPath(terragruntOptions.WorkingDir, "")
+	if err != nil {
+		return "", err
+	}
+
+	tmpFolder := filepath.Join(os.TempDir(), "terragrunt-downloads", canonicalPath)
+
+	if util.FileExists(tmpFolder) {
+		terragruntOptions.Logger.Printf("Temp folder %s already exists. Will delete Terraform configurations within it before downloading the latest ones.", tmpFolder)
+		if err := cleanupTerraformFiles(tmpFolder); err != nil {
+			return "", err
+		}
+	} else {
+		terragruntOptions.Logger.Printf("Creating temp folder %s to store downloaded Terraform configurations.", tmpFolder)
+		if err := os.MkdirAll(tmpFolder, 0777); err != nil {
+			return "", errors.WithStackTrace(err)
+		}
+	}
+
+	return tmpFolder, nil
+}
+
+// If this temp folder already exists, simply delete all the Terraform configurations (*.tf) within it
+// (the terraform init command will redownload the latest ones), but leave all the other files, such
+// as the .terraform folder with the downloaded modules and remote state settings.
+func cleanupTerraformFiles(path string) error {
+	files, err := filepath.Glob(filepath.Join(path, "*.tf"))
+	if err != nil {
+		return errors.WithStackTrace(err)
+	}
+	return util.DeleteFiles(files)
+}
+
+// There are two ways a user can tell Terragrunt that it needs to download Terraform configurations from a specific
+// URL: via a command-line option or via an entry in the .terragrunt config file. If the user used one of these, this
+// method returns the source URL and the boolean true; if not, this method returns an empty string and false.
+func getTerraformSourceUrl(terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig) (string, bool) {
+	if terragruntOptions.Source != "" {
+		return terragruntOptions.Source, true
+	} else if terragruntConfig.Terraform != nil && terragruntConfig.Terraform.Source != "" {
+		return terragruntConfig.Terraform.Source, true
+	} else {
+		return "", false
+	}
+}
+
+// Download the code from source into dest using the terraform init command
+func terraformInit(source string, dest string, terragruntOptions *options.TerragruntOptions) error {
+	terragruntInitOptions := terragruntOptions.Clone(terragruntOptions.TerragruntConfigPath)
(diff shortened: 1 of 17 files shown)


brikis98:
@josh-padnick Please review when you have a chance. This caching technique makes a pretty massive difference in usability. For example, for our Jenkins module, which itself relies in ~8 other modules, without the caching, every single Terragrunt command has a ~45 second overhead to run `terraform get`, `terraform remote config`, etc. With the caching, the overhead is 0.

josh-padnick:
Just reviewing this now, before I even get into the code, can we make the case that this feature of Terragrunt adds value above and beyond the new "supercharged" [terraform init](https://github.com/hashicorp/terraform/pull/11286) command (CTRL-F for "Features (External)")? Conceptually they seems very similar, though it's not clear to me how much caching they do.

brikis98:
> Just reviewing this now, before I even get into the code, can we make the case that this feature of Terragrunt adds value above and beyond the new "supercharged" terraform init command (CTRL-F for "Features (External)")? Conceptually they seems very similar, though it's not clear to me how much caching they do.

Good question. I'm not sure until I actually use it, but I'm guessing they download code into the local folder as a way to bootstrap a new project. Our use case is to download code into a tmp folder as a way to deploy an immutable version of your code.

cli/download_source.go
```diff
@@ -0,0 +1,259 @@
...
+// code is downloaded (or rather, copied) every single time. See the processTerraformSource method for more info.
+func alreadyHaveLatestCode(terraformSource *TerraformSource) (bool, error) {
+	if 	isLocalSource(terraformSource.CanonicalSourceURL) ||
+		!util.FileExists(terraformSource.DownloadDir) ||
+		!util.FileExists(terraformSource.VersionFile) {
+
+		return false, nil
+	}
+
+	currentVersion := encodeSourceVersion(terraformSource.CanonicalSourceURL)
```
josh-padnick: Looking at the definition of `encodeSourceVersion()`, you appear to be hashing the URL. But what if the URL is the same but its contents are different? This is the case, for example, when we re-release the same git tag. I believe this is the thought behind the `-update` in `terraform get -update`.

cli/download_source.go
```diff
@@ -0,0 +1,259 @@
...
+// 1. T is the OS temp dir (e.g. /tmp).
+// 2. W is the base 64 encoded sha1 hash of w. This ensures that if you are running Terragrunt concurrently in
+//    multiple folders (e.g. during automated tests), then even if those folders are using the same source URL s, they
+//    do not overwrite each other.
+// 3. S is the base 64 encoded sha1 has of s without its query string. For remote source URLs (e.g. Git
+//    URLs), this is based on the assumption that the scheme/host/path of the URL 
+//    (e.g. git::github.com/foo/bar//some-module) identifies the module name, and we always want to download the same
+//    module name into the same folder (see the encodeSourceName method). We also assume the version of the module is
+//    stored in the query string (e.g. ref=v0.0.3), so we store the base 64 encoded sha1 of the query string in a
+//    file called .terragrunt-source-version within S.
```
josh-padnick: s/S is the base 64 encoded sha1 **has** of s/S is the base 64 encoded sha1 of s/

cli/download_source.go
```diff
@@ -0,0 +1,259 @@
...
+	encodedWorkingDir := util.EncodeBase64Sha1(canonicalWorkingDir)
+	downloadDir := filepath.Join(os.TempDir(), "terragrunt-download", encodedWorkingDir, moduleName)
+	versionFile := filepath.Join(downloadDir, ".terragrunt-source-version")
+
+	return &TerraformSource{
+		CanonicalSourceURL: canonicalSourceUrl,
+		DownloadDir: downloadDir,
+		VersionFile: versionFile,
+	}, nil
+}
```
josh-padnick: Wow, lots of nuance here. Nicely handled.

cli/download_source.go
```diff
@@ -0,0 +1,259 @@
...
+		return "", false
+	}
+}
+
+// Download the code from the Canonical Source URL into the Download Folder using the terraform init command
+func terraformInit(terraformSource *TerraformSource, terragruntOptions *options.TerragruntOptions) error {
+	terragruntOptions.Logger.Printf("Downloading Terraform configurations from %s into %s", terraformSource.CanonicalSourceURL, terraformSource.DownloadDir)
+
+	terragruntInitOptions := terragruntOptions.Clone(terragruntOptions.TerragruntConfigPath)
+	terragruntInitOptions.TerraformCliArgs = []string{"init", terraformSource.CanonicalSourceURL.String(), terraformSource.DownloadDir}
```
josh-padnick: Consider reviewing https://github.com/hashicorp/terraform/pull/11286 for forward-compatibility.

cli/download_source_test.go
```diff
@@ -0,0 +1,208 @@
...
+	canonicalUrl := fmt.Sprintf("file://%s", absPath(t, "../test/fixture-download-source/hello-world"))
+	downloadDir := tmpDir(t)
+	defer os.Remove(downloadDir)
+
+	copyFolder(t, "../test/fixture-download-source/hello-world-2", downloadDir)
+
+	testDownloadTerraformSourceIfNecessary(t, canonicalUrl, downloadDir, "# Hello, World")
+}
+
+// TODO re-enable these two tests once this is merged into master, since they are trying to download code from master
```
josh-padnick: Roger. Nice way of handling tests with git.

glide.lock
```diff
@@ -1,43 +1,102 @@
...
+  - service/cloudwatchevents
+  - service/dynamodb/dynamodbattribute
+  - service/ecr
+  - service/firehose
+  - service/inspector
+  - service/marketplacecommerceanalytics
+  - service/mobileanalytics
+  - service/cloudfront/sign
+  - service/dynamodb/dynamodbiface
+  - service/sqs/sqsiface
```
josh-padnick: Do we really need this many additional AWS services? Will this bloat the binary?

util/file.go
```diff
@@ -46,6 +46,16 @@ func CanonicalPaths(paths []string, basePath string) ([]string, error) {
...
 
+// Delete the given list of files
+func DeleteFiles(files []string) error {
+	for _, file := range files {
+		if err := os.Remove(file); err != nil {
+			return errors.WithStackTrace(err)
+		}
+	}
+	return nil
+}
```
josh-padnick: Does this need to be able to handle recursive deletes?

cli/download_source.go
```diff
@@ -0,0 +1,259 @@
...
+type TerraformSource struct {
+	// A canonical version of RawSource, in URL format
+	CanonicalSourceURL *url.URL
+
+	// The folder where we should download the source to
+	DownloadDir         string
+
+	// The path to a file in DownloadDir that stores the version number of the code
+	VersionFile         string
+}
```
josh-padnick: Would be helpful to add a comment on what this struct as a whole represents.

josh-padnick: Would be helpful to add a comment on what this struct as a whole represents. It took me a minute to figure out this represents a set of "source code" for Terraform.

cli/download_source.go
```diff
@@ -0,0 +1,259 @@
...
+func processTerraformSource(source string, terragruntOptions *options.TerragruntOptions) (*TerraformSource, error) {
+	canonicalWorkingDir, err := util.CanonicalPath(terragruntOptions.WorkingDir, "")
+	if err != nil {
+		return nil, err
+	}
+
+	rawSourceUrl, err := getter.Detect(source, canonicalWorkingDir, getter.Detectors)
+	if err != nil {
+		return nil, errors.WithStackTrace(err)
+	}
```
josh-padnick: Good call using go-getter to normalize URLs. Looks like a great library.

cli/download_source.go
brikis98: Added: https://github.com/gruntwork-io/terragrunt/pull/114/commits/6de18efb9906293f86a4c5cd0b474807b32d8b1e

brikis98: Added: https://github.com/gruntwork-io/terragrunt/pull/114/commits/6de18efb9906293f86a4c5cd0b474807b32d8b1e

cli/download_source.go
brikis98: Yes, I considered this. In the future, if this is a common case, we could add a `terragrunt-update` or similar flag. For now, the trivial workaround is to delete the relevant tmp folder manually. My hope, however, is that this is a relatively rare case.

cli/download_source.go
brikis98: Fixed: https://github.com/gruntwork-io/terragrunt/pull/114/commits/f937e393dfdbb2b18941c3e1dc31fdc54ed4fc4e

cli/download_source.go
brikis98: It's actually the library Terraform/Packer/etc use to download files, including the `terraform init` command.

cli/download_source.go
brikis98: Looks like `init` may be changed in a backwards incompatible way to be an interactive command. Not sure I can do much about that for now. One alternative is to try to use go-getter directly to reimplement the download functionality, but I'd rather wait and see if with the next version of Terraform, we can leverage `init` with special flags to do what we need.

glide.lock
brikis98: Gah. Stupid glide. All I did was add one damn dependency. Hopefully this new version works, and has far fewer extras: https://github.com/gruntwork-io/terragrunt/pull/114/commits/15ed21e254a4ef42b63b5d296c5a42ef772e792e

util/file.go
brikis98: It's designed solely to delete files. I've updated the comment to make that clearer: https://github.com/gruntwork-io/terragrunt/pull/114/commits/931215de4b4a61f043ca735767af497ea949a93f

brikis98:
Thanks for the feedback, merging now.

PR #117, opened 2017-02-01
Add a --terragrunt-source-update flag

brikis98:
In https://github.com/gruntwork-io/terragrunt/pull/114, I added the ability for Terraform to cache the temporary folder so we don’t have to redownload code unnecessarily. This PR adds a new flag that I (naively) hoped wouldn’t be necessary: `—terragrunt-source-update`. This flag is a bit like the `-update` flag in terraform. It deletes the temporary folder before downloading into it.

### DIFF
--- a/cli/args.go
+++ b/cli/args.go
@@ -60,6 +60,8 @@ func parseTerragruntOptionsFromArgs(args []string) (*options.TerragruntOptions, 
 		return nil, err
 	}
 
+	updateSource := parseBooleanArg(args, OPT_TERRAGRUNT_SOURCE_UPDATE, false)
+
 	return &options.TerragruntOptions{
 		TerragruntConfigPath: filepath.ToSlash(terragruntConfigPath),
 		TerraformPath: filepath.ToSlash(terraformPath),
@@ -69,6 +71,7 @@ func parseTerragruntOptionsFromArgs(args []string) (*options.TerragruntOptions, 
 		Logger: util.CreateLogger(""),
 		RunTerragrunt: runTerragrunt,
 		Source: terraformSource,
+		UpdateSource: updateSource,
 		Env: parseEnvironmentVariables(os.Environ()),
 	}, nil
 }
--- a/cli/args.go
+++ b/cli/args.go
@@ -60,7 +60,7 @@ func parseTerragruntOptionsFromArgs(args []string) (*options.TerragruntOptions, 
 		return nil, err
 	}
 
-	updateSource := parseBooleanArg(args, OPT_TERRAGRUNT_SOURCE_UPDATE, false)
+	sourceUpdate := parseBooleanArg(args, OPT_TERRAGRUNT_SOURCE_UPDATE, false)
 
 	return &options.TerragruntOptions{
 		TerragruntConfigPath: filepath.ToSlash(terragruntConfigPath),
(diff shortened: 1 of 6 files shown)


cli/args.go
```diff
@@ -69,6 +71,7 @@ func parseTerragruntOptionsFromArgs(args []string) (*options.TerragruntOptions,
 		Logger: util.CreateLogger(""),
 		RunTerragrunt: runTerragrunt,
 		Source: terraformSource,
+		UpdateSource: updateSource,
```
josh-padnick: Is the preferred phrasing "updateSource" or "sourceUpdate"? It seems to be used in both ways in different places.

cli/cli_app.go
```diff
@@ -56,6 +57,7 @@ GLOBAL OPTIONS:
    terragrunt-non-interactive    Assume "yes" for all prompts.
    terragrunt-working-dir        The path to the Terraform templates. Default is current directory.
    terragrunt-source             Download Terraform configurations from the specified source into a temporary folder, and run Terraform in that temporary folder
+   terragrunt-source-update      Delete the contents of the temporary folder before downloading source code into it
```
josh-padnick: Although this perfectly describes what's happening, the user still has to make the leap "...and that's valuable because the temporary folder acts as a cache so this is a way of busting the cache." Consider writing what this is doing more so than how it works.

cli/download_source_test.go
```diff
@@ -131,17 +131,32 @@ func TestDownloadTerraformSourceIfNecessaryRemoteUrlToAlreadyDownloadedDirSameVe
...
 		CanonicalSourceURL: parseUrl(t, canonicalUrl),
 		DownloadDir: downloadDir,
 		VersionFile: filepath.Join(downloadDir, "version-file.txt"),
 	}
 
-	err := downloadTerraformSourceIfNecessary(terraformSource, options.NewTerragruntOptionsForTest("./should-not-be-used"))
+	terragruntOptions := options.NewTerragruntOptionsForTest("./should-not-be-used")
+	terragruntOptions.UpdateSource = updateSource
+
+	err := downloadTerraformSourceIfNecessary(terraformSource, terragruntOptions)
```
josh-padnick: It was a little hard for me to following how this test was working. Looks like you're just testing that `terragruntOptions.UpdateSource == true` allows it to run successfully. Thinking it through I don't think there's a great way to test that this works except possibly reading file timestamps in the temp folder before and after, but that's a little cumbersome.

cli/args.go
brikis98: Renamed everywhere to `sourceUpdate` for consistency: https://github.com/gruntwork-io/terragrunt/pull/117/commits/1bf8ed3e3a23e2be6d8bc36c32f7b49ce7eadbe3

cli/cli_app.go
brikis98: Updated description to: "Delete the contents of the temporary folder to clear out any old, cached source code before downloading new source code into it"

cli/download_source_test.go
brikis98: This is a helper function that is used by a bunch of actual test cases up above. The `TestDownloadTerraformSourceIfNecessaryRemoteUrlToAlreadyDownloadedDirSameVersion` and `TestDownloadTerraformSourceIfNecessaryRemoteUrlOverrideSource` test cases both set the `source` URL to the same value and both run against folders that contain downloaded code of exactly the same version hash as the `source` URL. The difference is that the first test case doesn't set `UpdateSource`, so it verifies that nothing is downloaded, while the second test case does set `UpdateSource`, so it verifies that the new code is downloaded from the `source` URL.

brikis98:
Merging now.

Issue #113, opened 2017-01-29
Support multiple terraform source locations


dave-pollock:
I am following the basic approach outlined in the Terragrunt docs under "Remote Terraform Configurations" and hitting a problem using modules. Let me know if I'm missing something.

An example of my project structure is:
* terraform
  * environments
    * stage
      * frontend-app
        └ .terragrunt   <-- uses "./../../../modules/frontend-app" as terraform source
        └ terraform.tfvars
      * backend-app
        └ .terragrunt
        └ terraform.tfvars
  * modules
    * reusable_module
      └ main.tf
    * frontend-app
        └ main.tf <-- contains a module using "${path.module}/../reusable_module" as source
    * backend-app
        └ main.tf

The issue is, Terragrunt will copy the files from modules/frontend-app to a temp directory and execute Terraform, but the reusable_module module source path is no longer valid. This means in order to use modules I need to either:
 - Use another source type such as GitHub, but I'd prefer to use the locally checked out copy during development.
 - Hard code the full path to the module (obviously bad).

I'm thinking simply adding the option to specify multiple source paths for terragrunt to copy into the temp directory would solve this. Something like:

```
terraform {
  source = [
    "./../../../modules/frontend-app",
    "./../../../modules/reusable_module"
  ]
}
```
Does this sound like a good solution or am I on the wrong track?

brikis98:
Good point. This is definitely something we need to fix. We need a way for the `source` usage to take into account that your Terraform code may contain relative paths pointing to files outside of the module.

A list of `source` URLs is one option, but a little tedious. Another option is to specify a root `source` that Terragrunt downloads and a way to separately specify the working directory within that root. Something like:

```hcl
terraform {
  source {
    root = "../../../modules"
    working_dir = "frontend-app"
  }
}
```

Terragrunt would download `root` and `cd` into `${root}/${working_dir}`. I'll need to think on this one a bit more.

brikis98:
Actually, thinking about it more, the syntax doesn't necessarily have to change. Terraform's `source` URLs, at least for Git URLs, already take this into account. Consider:

```hcl
terraform {
  source = "github.com/foo/bar.git//frontend-app"
}
```

The double-slash (`//`) separates the repo from the folder you want. Terragrunt could check out the entire repo and then `cd` into the folder you want. 

It's a little odd looking, but things seem to work just fine if you use a double-slash with local file paths too:

```hcl
terraform {
  source = "../../../modules//frontend-app"
}
[...]

PR #121, opened 2017-02-03
Download root folders of repos so relative paths work

brikis98:
This is a fix for the bug described in #113. Consider the following `.terragrunt` file:

```hcl
terraform {
  source = "github.com/acme/infrastructure-modules.git//foo/bar"
}
```

Originally, Terragrunt would download the `foo/bar` folder from the `infrastructure-modules` repo into a temp folder. Now imagine that `foo/bar/main.tf` contains the following:

```hcl
module "my_module" {
  source = "../../my-module"
}
```

This code won’t work. That’s because it uses a relative path to the `my-module` folder, which is in the `infrastructure-modules` repo, but Terragrunt didn’t download it, since it wasn’t in the `foo/bar` folder.

This PR updates the downloading logic to download the entire repo, identified by the part of the `source` URL before the double-slash `//` into a temporary folder, and to set the working directory to the `foo/bar` folder within that temporary folder. This way, relative paths should work as you expect. Local file paths can specify a double-slash as well for the same purpose. If no double-slash is specified, Terragrunt logs a warning, and downloads solely the one folder.


joslynesser:
Perfect timing! I've been setting up a terragrunt foundation for our org's ops and bumped into this very same issue due to having all of our blueprint modules stored in the same repo as our live stage/prod houses. Thought I was going to need to go the route of having another layer of modules and boilerplate variables passing through the stack, but this will let us go back to using `source` and DRY up a lot of code! Thanks for cranking this out so quickly 🍺

brikis98:
@josh-padnick I'm going to merge this in so I can build the next increment, but please review when you have a chance.

PR #126, opened 2017-02-14
When cleaning up files, don't delete files in .terraform folders

brikis98:
This PR fixes a bug in Terragrunt’s support for remote source directories where it was deleting `*.tf` files in the tmp download folder so it could update them with the latest code. Unfortunately, this also deleted files downloaded from modules into `.terraform` folders, so you would get “you need to run terraform get” style errors after multiple re-runs. 

I’ve updated the test cases to catch this error.


brikis98:
@josh-padnick This is a pretty straightforward fix, so I'm going to merge it. Feedback welcome.

PR #128, opened 2017-02-16
Adding `extra_arguments` config option

realloc:
Sometimes you may need to pass extra arguments to Terraform on each run.
For example if you have a separate file with secret variables you may
use `extra_arguments` option in `terraform` section of Terragrunt
configuration to do it automatically.

Example:

  terraform = {
    extra_arguments = [
      "-var-file=terraform.tfvars",
      "-var-file=terraform-secret.tfvars"
    ]
  }

### DIFF
--- a/README.md
+++ b/README.md
@@ -594,6 +594,57 @@ terragrunt = {
 }
 ```
 
+### Passing extra command line arguments to Terraform
+
+Sometimes you may need to pass extra arguments to Terraform on each run. For example if you have a separate file with
+secret variables you may use extra_arguments option in terraform section of Terragrunt configuration to do it
+automatically.
+
+Each set of arguments will be appended only if current Terraform command is in `commands` list. If more than one set is
+applicable, they will be added in the order of of appearance in config.
+
+Sample config:
+
+``` hcl
+terragrunt= {
+  terraform = {
+    {
+      extra_arguments "secrets" {
+        arguments = [
+          "-var-file=terraform.tfvars",
+          "-var-file=terraform-secret.tfvars"
+        ]
+        commands = [
+          "apply",
+          "plan",
+          "import",
+          "push",
+          "refresh"
+        ]
+      }
+
+      extra_arguments "json_output" {
+        arguments = [
+          "-json"
+        ]
+        commands = [
+          "output"
+        ]
+      }
+
+      extra_arguments "fmt_diff" {
+        arguments = [
+          "-diff=true"
+        ]
+        commands = [
+          "fmt"
+        ]
+      }
+
+    }
+  }
+```
+
 ### The spin-up and tear-down commands
 
 Let's say you have a single environment (e.g. `stage` or `prod`) that has a number of Terraform modules within it:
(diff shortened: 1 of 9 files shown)


realloc:
1. Agree. Will fix.
2. Hmm... Guess you use IDEA? =) Ok, will do it without `go fmt`.
3. Ok, will try to do it.

Thank You for review!

brikis98:
> 1. Agree. Will fix.

Great. One option is to have a way to list each individual command, but that might get tedious, so perhaps we'd also have some "presets", such as "var-commands" (can't think of a better name...), which automatically fills in the list of commands that take `-var` or `-var-file` args.

> 2. Hmm... Guess you use IDEA? =) Ok, will do it without go fmt.

Yes :)

I just filed https://github.com/gruntwork-io/terragrunt/issues/129 to run `go fmt` globally on the codebase, but I'd rather avoid doing one-offs, as it makes the diff hard to follow. 

> 3. Ok, will try to do it.

Great, thanks! If you have questions, just ask.

brikis98:
@justicel Check this out, this Terragrunt change may do exactly what we discussed the other day!

realloc:
> Great. One option is to have a way to list each individual command, but that might get tedious, so perhaps we'd also have some "presets", such as "var-commands" (can't think of a better name...), which automatically fills in the list of commands that take -var or -var-file args.

I will try to implement something like this and update PR. If multiple entries fit current action, arguments should be appended in the order of appearance in config.

``` hcl
terragrunt= {
  terraform = {
    {
      extra_arguments "secrets" {
        arguments = [
          "-var-file=terraform.tfvars",
          "-var-file=terraform-secret.tfvars"
        ]
        actions = [
          "apply",
          "plan",
          "import",
          "push",
          "refresh"
        ]
      }
      
      extra_arguments "json_output" {
        arguments = [
          "-json"
        ]
        actions = [
          "output"
        ]
      }
      
      extra_arguments "fmt_diff" {
        arguments = [
          "-diff=true"
        ]
        actions = [
          "fmt"
        ]
      }
      
    }
  }
```

realloc:
Please review updated change.

Each set of arguments will be appended only if current Terraform command is in `commands` list.

If more than one set is applicable, they will be added in the order of appearance in config.

Sample config:

``` hcl
terragrunt= {
  terraform = {
    {
      extra_arguments "secrets" {
        arguments = [
          "-var-file=terraform.tfvars",
          "-var-file=terraform-secret.tfvars"
        ]
        commands = [
          "apply",
          "plan",
          "import",
          "push",
          "refresh"
        ]
      }

      extra_arguments "json_output" {
        arguments = [
          "-json"
        ]
        commands = [
          "output"
        ]
      }

      extra_arguments "fmt_diff" {
        arguments = [
          "-diff=true"
        ]
        commands = [
          "fmt"
        ]
      }
    }
  }
```

realloc:
Fixed.

brikis98:
Perfect, thank you! Merging now.

brikis98:
I just created a new release here: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.10.2. If the build passes, the new binaries should show up in a few minutes.

Issue #130, opened 2017-02-16
terragrunt ignores --terragrunt-config / TERRAGRUNT_CONFIG files


tmclaugh:
contrary to docs, terragrunt ignores --terragrunt-config arg, TERRAGRUNT_CONFIG env var, and .terragrunt file when looking for tfvars.  terragrunt will fail of the given file does not exist but it does not appear to read the file.  I'm currently working around this by symlinking terraform.tfvars back to the tfvars file I want.  This does not work long term for me as I'm trying to drive different environments entirely by separate tfvars files.

Ignored given file:
```
[tmclaughlin@tomcat-ts:aws-straycat straycat-terraform-aws-vpc(modularize)*]$ terragrunt plan --terragrunt-config straycat.tfvars
[terragrunt] 2017/02/16 10:17:06 Reading Terragrunt config file at straycat.tfvars
[terragrunt] 2017/02/16 10:17:06 Remote state is already configured for backend s3
[terragrunt] 2017/02/16 10:17:06 WARNING: you have not configured locking in your Terragrunt configuration. Concurrent changes to your .tfstate files may cause conflicts!
[terragrunt] 2017/02/16 10:17:06 Running command: terraform plan
var.aws_account
  Enter a value: ^C
Error configuring: Error asking for aws_account: interrupted

[tmclaughlin@tomcat-ts:aws-straycat straycat-terraform-aws-vpc(modularize)*]$ ls -al straycat.tfvars
-rw-r--r--  1 tmclaughlin  staff  789 Feb 15 17:25 straycat.tfvars
```

Failed on non-existent file:
```
[tmclaughlin@tomcat-ts:aws-straycat straycat-terraform-aws-vpc(modularize)*]$ terragrunt plan --terragrunt-config doesnotexist.tfvars
[terragrunt] 2017/02/16 10:18:26 Reading Terragrunt config file at doesnotexist.tfvars
Error reading file at path doesnotexist.tfvars: open doesnotexist.tfvars: no such file or directory
[terragrunt] 2017/02/16 10:18:26 Error reading file at path doesnotexist.tfvars: open doesnotexist.tfvars: no such file or directory
```

brikis98:
```
[terragrunt] 2017/02/16 10:17:06 Reading Terragrunt config file at straycat.tfvars
[terragrunt] 2017/02/16 10:17:06 Remote state is already configured for backend s3
[...]

PR #131, opened 2017-02-16
Clarify Terragrunt vs. Terraform config

tmclaugh:
using --terragrunt-config or other options for overriding .tfvars with Terragrunt does not override Terraform's own .tfvars file processing.

resolves: gruntwork-io/terragrunt#130

### DIFF
--- a/README.md
+++ b/README.md
@@ -917,6 +917,12 @@ Terragrunt figures out the path to its config file according to the following ru
 1. A `terraform.tfvars` file in the current working directory, if it exists.
 1. If none of these are found, exit with an error.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -917,6 +917,12 @@ Terragrunt figures out the path to its config file according to the following ru
 1. A `terraform.tfvars` file in the current working directory, if it exists.
 1. If none of these are found, exit with an error.
 
+Setting `--terragrunt-config` does not override the config file used by Terraform and its own configuration file processing still applies. If you use a single `.tfvars` file for Terragrunt and Terraform configuration that is not named `terraform.tfvars` then remember to pass the `.tfvars` file name by using the `--var-file` argument.
```
brikis98: I'd rephrase this a little bit to:

The `--terragrunt-config` parameter is only used by Terragrunt and has no effect on which variable files are loaded by Terraform. Terraform will automatically read variables from a file named `terraform.tfvars`, but if you want it to read variables from some other `.tfvars` file, you must pass it in using the `--var-file` argument:

brikis98:
Looks good, thank you!

PR #141, opened 2017-02-21
Wait for shell when forwarding signals; shell package test coverage

### LINKED ISSUE #120: Interrupt of `terragrunt plan` doesn't wait for the child proess to stop.
Reproduction: run `terragrunt plan`, `^C` quickly thereafter, which drops you back to bash while seas of debug output, which you can't stop, overwhelm you.

This happens because `terragrunt plan` didn't `wait(3)` for `terraform plan` to finish, so the second `^C` may not be delivered, which, in case of `terraform plan` would have killed it, with the first one enabling prints of debug output.

asobrien:
This PR fixes #120.

The changes include:

* Modifying the `NewSignalsForwarder` so that the signal forwarder waits for the command to finish. It is possible that multiple signals need to be passed the command. Additionally, this fixes a bug where the signal forwarder previously panicked by passed the `*exec.Cmd`, instead of `os.Process`, to the signal forwarder since `os.Process` is only allocated after the command starts running.
* Adds test coverage for the `shell` package
* Creates a `GetExitCode` function, the exit code logic in `main.go` was migrated into the `shell` package so it could be tested.

### DIFF
--- a/main.go
+++ b/main.go
@@ -37,6 +37,7 @@ func checkForErrorsAndExit(err error) {
 		exitCode, exitCodeErr := shell.GetExitCode(err)
 		if exitCodeErr != nil {
 			exitCode = 1
+			logger.Println("Unable to determine underlying exit code, so Terragrunt will exit with error code 1")
 		}
 		os.Exit(exitCode)
 	}
--- a/main.go
+++ b/main.go
@@ -34,12 +34,11 @@ func checkForErrorsAndExit(err error) {
 			logger.Println(err)
 		}
 		// exit with the underlying error code
-		var retCode int
-		retCode, err := shell.GetExitCode(err)
-		if err != nil {
-			retCode = 1
+		exitCode, exitCodeErr := shell.GetExitCode(err)
+		if exitCodeErr != nil {
+			exitCode = 1
 		}
-		os.Exit(retCode)
+		os.Exit(exitCode)
 	}
 
 }
--- a/main.go
+++ b/main.go
@@ -2,8 +2,7 @@ package main
 
 import (
 	"os"
-	"os/exec"
-	"syscall"
+	"github.com/gruntwork-io/terragrunt/shell"
 	"github.com/gruntwork-io/terragrunt/util"
 	"github.com/gruntwork-io/terragrunt/cli"
 	"github.com/gruntwork-io/terragrunt/errors"
@@ -35,10 +34,10 @@ func checkForErrorsAndExit(err error) {
 			logger.Println(err)
 		}
 		// exit with the underlying error code
-		var retCode int = 1
-		if exiterr, ok := errors.Unwrap(err).(*exec.ExitError); ok {
-			status := exiterr.Sys().(syscall.WaitStatus)
-			retCode = status.ExitStatus()
+		var retCode int
+		retCode, err := shell.GetExitCode(err)
+		if err != nil {
+			retCode = 1
 		}
 		os.Exit(retCode)
 	}
--- a/shell/run_shell_cmd.go
+++ b/shell/run_shell_cmd.go
@@ -1,6 +1,7 @@
 package shell
 
 import (
+	"fmt"
 	"log"
 	"os"
 	"os/exec"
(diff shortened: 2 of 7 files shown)


shell/run_shell_cmd.go
```diff
@@ -33,30 +34,47 @@ func RunShellCommand(terragruntOptions *options.TerragruntOptions, command strin
...
+
+// Return the exit code of a command. If the error is not an exec.ExitError type,
+// the error is returned.
+func GetExitCode(err error) (int, error) {
+	var retCode int
+	if exiterr, ok := errors.Unwrap(err).(*exec.ExitError); ok {
+	status := exiterr.Sys().(syscall.WaitStatus)
+	retCode = status.ExitStatus()
+	return retCode, nil
+		}
```
brikis98: Minor stylistic issue: the indentation of the body of this if statement and the closing curly brace seem off.

shell/run_shell_cmd.go
```diff
@@ -33,30 +34,47 @@ func RunShellCommand(terragruntOptions *options.TerragruntOptions, command strin
...
+
+
+// Return the exit code of a command. If the error is not an exec.ExitError type,
+// the error is returned.
+func GetExitCode(err error) (int, error) {
+	var retCode int
+	if exiterr, ok := errors.Unwrap(err).(*exec.ExitError); ok {
+	status := exiterr.Sys().(syscall.WaitStatus)
+	retCode = status.ExitStatus()
+	return retCode, nil
```
brikis98: Could just return `status.ExitStatus(), nil` instead of setting `retCode`.

brikis98: In fact, I wouldn't declare `retCode` at all. Just return `status.ExitStatus(), nil` inside the if-statement and return `0` outside of it.

main.go
```diff
@@ -35,10 +34,10 @@ func checkForErrorsAndExit(err error) {
...
 		}
 		// exit with the underlying error code
-		var retCode int = 1
-		if exiterr, ok := errors.Unwrap(err).(*exec.ExitError); ok {
-			status := exiterr.Sys().(syscall.WaitStatus)
-			retCode = status.ExitStatus()
+		var retCode int
+		retCode, err := shell.GetExitCode(err)
+		if err != nil {
+			retCode = 1
```
brikis98: You should log the error here explicitly, otherwise, it won't be clear why the exit code is 1 instead of some underlying exit code returned by Terraform.

```go
logger.Printf("Error retrieving underlying exit code, so Terragrunt will exit with error code 1: %v", err)).
```

main.go
```diff
@@ -35,10 +34,10 @@ func checkForErrorsAndExit(err error) {
 			logger.Println(err)
 		}
 		// exit with the underlying error code
-		var retCode int = 1
-		if exiterr, ok := errors.Unwrap(err).(*exec.ExitError); ok {
-			status := exiterr.Sys().(syscall.WaitStatus)
-			retCode = status.ExitStatus()
+		var retCode int
+		retCode, err := shell.GetExitCode(err)
```
brikis98: It's very confusing here to shadow both the `retCode` and `err` variables. I'd recommend removing the extra `retCode` declaration and renaming this `err` to `exitCodeErr`.

shell/run_shell_cmd_test.go
```diff
@@ -0,0 +1,121 @@
+package shell
```
brikis98: These tests are great!

shell/run_shell_cmd_test.go
```diff
@@ -0,0 +1,121 @@
...
+	for {
+		time.Sleep(500 * time.Millisecond)
+		select {
+		case err = <-runChannel:
+			break loop
+		default:
+			cmd.Process.Signal(os.Interrupt)
+			interrupts++
+		}
+	}
```
brikis98: Instead of a label, could you put this loop in a function and use a `return`?

testdata/test_sigint_multiple.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
```
brikis98: Add a `set -e` so this script exits on any error.

testdata/test_exit_code.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
```
brikis98: Add a `set -e` so this script exits on any error. May seem silly for such a tiny script, but you never know if it'll grow in the future.

main.go
asobrien: `GetExitCode` will just return the error it was passed if an exit code can't be determined. That error will be  logged on line 34. It does make sense to log something, perhaps something like 

```go
logger.Println("Unable to determine underlying exit code, so Terragrunt will exit with error code 1")
```

would be more appropriate so that the logs aren't redundant. I know for sure that calling `terragrunt` from a location that doesn't have a config file satisfies this condition (i.e., terragrunt has to be coerced to exit with a non-zero error code).

asobrien:
Thanks for the feedback! The requested changes are above.

brikis98:
Just published a new release: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.10.3. If the build passes, the new binaries will be there in a few minutes.

Issue #129, opened 2017-02-16
Run go fmt on the code


brikis98:
This repo should run `go fmt` as a commit hook so the code is formatted consistently.

brikis98:
Fixed by #148.

PR #148, opened 2017-02-25
Run all source through `gofmt`, add formatting docs and helpers

fajpunk:
Addresses https://github.com/gruntwork-io/terragrunt/issues/129.

Two parts to this PR:
1. Run all of the source through `gofmt`
2. Add docs specifying that all contributions should be `gofmt`ed, and add helper scripts:
    * Makefile target and script to check for formatting
    * Makefile target to format all source
    * Makefile target to install git pre-commit script that will check formatting and stop commits with unformatted files.

The format checker and some of the make targets were taken from the terraform project.

All tests passed when run locally with:
`AWS_PROFILE=terragrunt-test go test -v -parallel 128 $(glide novendor)`

This is the way I've personally dealt with pre-commit hooks.  I'm relatively new to golang, so I figured terraform would be a good place to crib from as to how a big public golang project handles the formatting.  If this is not the way you want to handle either of these things for this project, I'm completely open to anything else; just let me know what you want to do!


config/config_test.go
```diff
@@ -18,7 +18,7 @@ func TestParseTerragruntConfigDynamoLockMinimalConfig(t *testing.T) {
 	t.Parallel()
 
 	config :=
-`
+		`
```
brikis98: Hm, these changes are annoying and look weird. Is there a cleaner way to format multi-line strings in Go? Perhaps the opening backtick should go on the same line as the `:=`.

locks/lock_test.go
```diff
@@ -23,10 +24,16 @@ func TestWithLockNoop(t *testing.T) {
...
 var ErrorOnAcquire = fmt.Errorf("error-on-acquire")
-func (lock ErrorOnAcquireLock) AcquireLock(terragruntOptions *options.TerragruntOptions) error { return ErrorOnAcquire }
-func (lock ErrorOnAcquireLock) ReleaseLock(terragruntOptions *options.TerragruntOptions) error { return nil }
+
+func (lock ErrorOnAcquireLock) AcquireLock(terragruntOptions *options.TerragruntOptions) error {
+	return ErrorOnAcquire
+}
+func (lock ErrorOnAcquireLock) ReleaseLock(terragruntOptions *options.TerragruntOptions) error {
+	return nil
+}
```
brikis98: How weird. A few lines above, `gofmt` is perfectly happy with one-liner functions and even aligns their curly braces, and yet here, it makes them multiline functions... Nothing to do about it, just idly commenting :)

scripts/pre-commit
```diff
@@ -0,0 +1 @@
+make fmtcheck
```
brikis98: I'd make the precommit hook simply run `gofmt` to automatically format the code. After all, if the code isn't formatted, this script will exit with an error, and the *only* thing the developer can do is run `gofmt` anyway, so why not do it for them?

util/random_test.go
```diff
@@ -8,11 +8,11 @@ import (
 func TestGetRandomTime(t *testing.T) {
 	t.Parallel()
 
-	testCases := []struct{
+	testCases := []struct {
 		lowerBound time.Duration
 		upperBound time.Duration
 	}{
-		{ 1 * time.Second, 10 * time.Second },
+		{1 * time.Second, 10 * time.Second},
```
brikis98: Another `gofmt` oddity: for a lot of arithmetic, it removes spaces (e.g. `max-min` in the file above), but here, it leaves them. I prefer it with spaces, but it's weird that it would be inconsistent.

README.md
```diff
@@ -1079,6 +1079,20 @@ Here is how the `errors` package should be used:
    `errors.WithStackTrace`. This gives us a stacktrace as close to the source as possible.
 1. If you need to get back the underlying error, you can use the `errors.IsError` and `errors.Unwrap` functions.
 
+#### Formatting
+
+Every source file in this project should be formatted with `go fmt`. Feel free to accomplish this any way you choose. There are also few helper scripts and targets in the Makefile that can help with this (mostly taken from the [terraform repo](https://github.com/hashicorp/terraform/)):
```
brikis98: I would actually explicitly ask folks to install the precommit hook to automatically format their code.

config/config_test.go
fajpunk: Agreed; I tried putting the opening backtick on the same line as the assignment operator.

locks/lock_test.go
fajpunk: Definitely weird.  Maybe it's a line length thing?  Though I thought gofmt didn't care about line length.

scripts/pre-commit
fajpunk: Done

fajpunk:
I think I added `make fmtcheck` to the `test` section of the CircleCI config.  I'm not too familiar with CircleCI, I assume I just put the command in that list...

PR #153, opened 2017-03-08
fix terragrunt config example

jmlrt:
This example for passing extra arguments to terraform doesn't work due to some formating typos:

```hcl
terraform = {
  {
    extra_arguments "secrets" {
    ...
```

instead of 

```hcl
terraform {
  extra_arguments "secrets" {
  ...
```


brikis98:
Thank you for the fix!

PR #154, opened 2017-03-13
Add support for AssumeRole for DynamoDB locks and S3 remote state

npepinpe:
As it says, adds basic support for aws_role_arn (DynamoDB config) and role_arn (S3 remote state config), similar to the terraform AWS provider.

I realise it's not particularly well tested, but reading through the current tests, I was unsure about the way to go there. I'm nowhere near a Go expert, so I'll defer to your judgement, and gladly make any necessary adjustments.

### DIFF
--- a/aws_helper/config.go
+++ b/aws_helper/config.go
@@ -2,12 +2,15 @@ package aws_helper
 
 import (
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/gruntwork-io/terragrunt/errors"
 )
 
 // Returns an AWS session object for the given region, ensuring that the credentials are available
-func CreateAwsSession(awsRegion, awsProfile string) (*session.Session, error) {
+func CreateAwsSession(awsRegion, awsProfile, awsRoleArn string) (*session.Session, error) {
 	session, err := session.NewSessionWithOptions(session.Options{
 		Config:            aws.Config{Region: aws.String(awsRegion)},
 		Profile:           awsProfile,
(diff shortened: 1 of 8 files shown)


aws_helper/config.go
```diff
@@ -22,5 +25,29 @@ func CreateAwsSession(awsRegion, awsProfile string) (*session.Session, error) {
 		return nil, errors.WithStackTraceAndPrefix(err, "Error finding AWS credentials (did you set the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environment variables?)")
 	}
 
+	if awsRoleArn != "" {
+		err = assumeRole(session, awsRoleArn)
+		if err != nil {
```
brikis98: You can combine these two lines into one:

```go
if err := assumeRole(session, awsRoleArn); err != nil {
  ...
}
```

aws_helper/config.go
```diff
@@ -22,5 +25,29 @@ func CreateAwsSession(awsRegion, awsProfile string) (*session.Session, error) {
 		return nil, errors.WithStackTraceAndPrefix(err, "Error finding AWS credentials (did you set the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environment variables?)")
 	}
 
+	if awsRoleArn != "" {
+		err = assumeRole(session, awsRoleArn)
```
brikis98: This method calls `Credentials.Get()` a few lines above, so you should probably handle the role stuff earlier. Moreover, I think it can be done simpler, without the need for an extra method:

```go
func CreateAwsSession(awsRegion, awsProfile, awsRoleArn string) (*session.Session, error) {
  session, err := // ...

  if awsRoleArn != "" {
    session.Config.Credentials = stscreds.NewCredentials(sess, assumeRoleArn)
  }

  _, err = session.Config.Credentials.Get()
  // ...
```

brikis98: BTW, `session` is a bad choice for a variable name here, as it shadows the `session` package that is imported. That has nothing to do with your code, just a screw up on my behalf. Could you rename it to something like `sess` instead to avoid confusion?

remote/remote_state_test.go
```diff
@@ -13,10 +13,11 @@ func TestToTerraformRemoteConfigArgs(t *testing.T) {
...
 		Config: map[string]string{
-			"encrypt": "true",
-			"bucket":  "my-bucket",
-			"key":     "terraform.tfstate",
-			"region":  "us-east-1",
+			"encrypt":  "true",
+			"bucket":   "my-bucket",
+			"key":      "terraform.tfstate",
+			"region":   "us-east-1",
+			"role_arn": "arn:aws:iam::318125143396:role/delegated-admin.apps-sandbox",
```
brikis98: Is this a real ARN from your AWS account?

npepinpe: Oh wow, rookie mistake there, I was using this to test in a sandbox, my bad.

npepinpe:
Some of the integration tests are failing at the moment, mostly related to includes, I'll update with the test output once I get them all to pass. I'm also not sure if to add integration tests for this case, as there are none currently for the profile either. Not sure what the right approach would be.

brikis98:
> Some of the integration tests are failing at the moment, mostly related to includes, I'll update with the test output once I get them all to pass. 

Ah, let me know when you have that sorted.

> I'm also not sure if to add integration tests for this case, as there are none currently for the profile either. Not sure what the right approach would be.

It would be great to add integration tests for this, but you'd probably have to dynamically generate IAM roles and then clean them up after to make it work... I'm not sure on the best approach for that.

brikis98:
@npepinpe This PR would need some serious rebasing, as Terragrunt no longer manages locking by itself since Terraform v0.9 came out. I'm going to close this one for now, but please reopen if you have a chance to revisit this stuff.

PR #159, opened 2017-04-05
Add plan-all option to execute plan on all stacks

jocgir:


brikis98:
This is a good idea, but there is a fundamental problem: if one of your Terraform modules depends on another through remote state, and none of the modules have been deployed yet, this command will fail, since the remote state won't be there yet... There may still be some value when everything is already deployed and you're just running this to see small diffs, or even to see as much of the plan as you can until you hit an error... But we may need better error handling/messaging to make that clear.

jocgir:
Good point, I will try to add significant error message. Thank you for the comment.

brikis98:
@jocgir Did this PR accidentally pull in the changes from #160?

jocgir:
Hi @brikis98, oups! You are right. I made this first PR directly on our master, but I wanted to separate them, so I rolled back then change and made a separate branch. But this pull request is continuing to follow our master branch instead of the newly created branch.

Issue #158, opened 2017-03-31
How Terragrunt Will Handle Terraform 0.9.0 Compatibility


josh-padnick:
Hello Terragrunt Community,

As of today, Terragrunt is not compatible with Terraform 0.9.x. We want to fix that. This GitHub Issue will outline our current thinking on how we plan to do that.

- **Handling Remote State in Terraform 0.9:** Unfortunately, Terraform 0.9.x removed the `terraform remote config` commands, which is why Terragrunt is not usable with Terraform 0.9.x. To resolve this, we’ve been exploring the best way to configure remote state in Terraform 0.9.x.

   Ideally, Terragrunt could simply retire this feature altogether, as Terraform 0.9 allows you to define your remote state configuration in your .tf files. Unfortunately, Terraform does not support all the features Terragrunt had, such as  `path_relative_to_include()`, manual lock controls, and automatic retry when a lock cannot be obtained.

   Our current thinking is to move to Terraform’s remote state configuration, but with variables where you’d like to use Terragrunt features:

   ```hcl
   # In some .tf file
   terraform {
      backend "s3" {
         bucket = "my-bucket"
         key    = "${var.remote_state_key}"
         region = "us-east-1"
      }
   }
   ```

   You can then configure Terragrunt to pass in the variables, using the interpolation helpers as before:

   ```hcl
   # In terraform.tfvars or .terragrunt
   terragrunt = {
      terraform {
         extra_arguments "remote_state" {
            arguments = [
                "-var", "remote_state_key=${path_relative_to_include()}",
            ]
            commands = [
               "apply",
               "plan",
               "import",
               "push",
               "refresh"
            ]
          }
       }
   }
   ```

- **User Migration Strategy:** Many of our users (and Gruntwork clients) have multiple folders configured to use Terragrunt. We plan to explicitly think through a recommended migration strategy. This could be a simple recommendation to manually make updates, or some kind of automation. We’re still thinking through this.

- **Handling Retries When a Lock Cannot be Obtained:** When Terragrunt cannot obtain a lock, it keeps trying to get one until it times out. Currently, Terraform 0.9.x fails immediately if a lock cannot be obtained. Some CI scripts may depend on having an automatic retry, so we intend to wrap the native Terraform backends code with our own lock-retry mechanism.

### Timeline

[...]

PR #160, opened 2017-04-06
Features/new remote state support

jocgir:
Hi, here is our proposition to add support for terraform 0.9.0+.

Basically, we inject a temporary file into the current folder with the required terraform 0.9+ configuration to support the new remote state format. We rely on tarraform init instead of terraform remote for new versions.

The temporary file is always generated based on the terragrunt configuration.

The code continue to support oldest terraform versions.

Comments are welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,8 +47,6 @@ a simple, free locking mechanism, and enforcing best practices around CLI usage 
 
 ## Install
 
-**NOTE**: Terraform 0.9 includes backwards incompatible changes and is NOT currently supported by Terragrunt. 
-
 1. Install [Terraform](https://www.terraform.io/), and let Terragrunt know where to find it using one of the following options:
 
     * Place `terraform` in a directory on your PATH.
--- a/circle.yml
+++ b/circle.yml
@@ -7,8 +7,7 @@ dependencies:
     # Install the gruntwork-module-circleci-helpers and use it to configure the build environment and run tests.
     - curl -Ls https://raw.githubusercontent.com/gruntwork-io/gruntwork-installer/master/bootstrap-gruntwork-installer.sh | bash /dev/stdin --version 0.0.13
     - gruntwork-install --module-name "gruntwork-module-circleci-helpers" --repo "https://github.com/gruntwork-io/module-ci" --tag "v0.3.2"
-    # Have to stick with Terraform 0.8.x, as Terraform 0.9 has backwards incompatible changes that Terragrunt does not yet support
-    - configure-environment-for-gruntwork-module --terraform-version 0.8.8 --packer-version NONE --terragrunt-version NONE --go-src-path .
+    - configure-environment-for-gruntwork-module --terraform-version 0.9.2 --packer-version NONE --terragrunt-version NONE --go-src-path .
 
   cache_directories:
     - ~/glide
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -4,8 +4,6 @@ import (
 	"fmt"
 	"regexp"
 
-	"io"
-
 	"github.com/gruntwork-io/terragrunt/config"
 	"github.com/gruntwork-io/terragrunt/configstack"
 	"github.com/gruntwork-io/terragrunt/errors"
@@ -15,6 +13,7 @@ import (
 	"github.com/gruntwork-io/terragrunt/shell"
 	"github.com/gruntwork-io/terragrunt/util"
 	"github.com/urfave/cli"
+	"io"
 )
 
 const OPT_TERRAGRUNT_CONFIG = "terragrunt-config"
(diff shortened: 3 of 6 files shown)


brikis98:
@jocgir Thank you for putting this PR together. I'm very grateful to see someone pitching in with this important upgrade :)

I'm a bit concerned with the approach of generating a file on the fly. Here's why:

1. It's a bit messy. A random file will pop in and out, which will be confusing to users, and problematic if due to a bug, crash, or some other reason, the file gets left behind.

1. It's not clear what the migration path is for getting users to move to Terraform 0.9 backends. I'd prefer that users actually define the backends in their config, but can still somehow take advantage of Terragrunt's other features (e.g. the `path_relative_to_include` helper). That way, as Terraform improves backends even more, we can gradually phase that aspect of Terragrunt out completely. 

We discussed all of this in https://github.com/gruntwork-io/terragrunt/issues/158 and [this comment in particular](https://github.com/gruntwork-io/terragrunt/issues/158#issuecomment-290787745) has an interesting way to accomplish this by piggybacking on the `terraform init` command.

jocgir:
@brikis98, I agree that generating file (and mostly put it in the code folder) is not the ideal way. It would be preferable if terraform allowed us to add a .tf file through args like -var-file. I took the precaution to ensure the file is cleaned up in case of most crashes or interrupts, but you are right, the file can still be left behind in some occasion. In most terragrunt use case, the inserted file is added to the temporary folder (when using source in terragrunt config). So the file is not added directly to the original folder.

The idea was inspired by the piggybacking one. I did that originally, but it required a «more» complex deployment on user setup, so I preferred to put it directly into the terragrunt executable.

I am not sure to understand your second concern. I felt that this technique allowed users to move to Terraform 0.9 whenever they are ready. When users install terraform 0.9, they get a warning and they can decide if they want to keep the lecacy way (which will be supported until 0.11) or move to the new backend. However, once they move to the new format, the whole team has to move together since the tfstate is not backward compatible.

brikis98:
> I am not sure to understand your second concern. I felt that this technique allowed users to move to Terraform 0.9 whenever they are ready.

To clarify what I mean, consider what a user would have to do with the file generation strategy: they would have to pick between stick with Terragrunt for remote state management and not using the new Terraform format OR they can move to the new format, but then they lose access to Terragrunt features such as the `path_relative_to_include` helper. 

As discussed in #158 (make sure to read the comments too), Terraform's backends + locking are missing many features that were available in Terragrunt, so ideally, the migration strategy would let you start using as much of the Terraform 0.9 functionality as possible, while still using Terragrunt for pieces here and there.

jocgir:
Hi @brikis98, be assured that I read all comments prior to make this PR.

I am not sure that you understood what this code is doing. In all cases, the users stick with all advantages of using terragrunt. But, when they move to the new terraform format, terragrunt simply add temporary terraform configuration to match the one that is configured in the terragrunt section which still benefits of the path_relative_to_include and others helpers. They don't have to change a single line of code or loose any of terragrunt functionality.

The @tamsky wrapper trick is also possible (instead of generating a file on the fly), but it have to be OS aware. We cannot presume that bash or any other sh is available on user system. We work in a mixed team (OSX, Linux and Windows). To make it work, we can set the TERRAGRUNT_TFPATH, redirecting call to terragrunt itself to handle the remote to init conversion, then, unset the variable and make the actual terraform call. We can also generate a shell script in a temp folder, but we have to handle multiple os, i.e. .bat or .ps1 on Windows. It is also possible to add a Go terraform wrapper companion to terragrunt which would be portable, but that seems a little overkill for a simple arts conversion problem. Moreover, I realized that this trick is not working with many terraform calls, while mine is always working. Just try other terraform command like terraform state list or terraform console to see the limitation.

brikis98:
@jocgir 

Let me try to explain what I was thinking of as a transition.

Initially, Terragrunt users have their remote state configured in a `terraform.tfvars` file:

```hcl
terragrunt = {
  remote_state {
    backend = "s3"
    config {
      encrypt = "true"
      bucket = "my-bucket"
      key = "${path_relative_to_include()}"
      region = "us-east-1"
    }
  }
}
```

To use the new version of Terragrunt with Terraform 0.9, the only thing they should have to change is to add the following somewhere in their `.tf` files:

```hcl
terraform {
  backend "s3" {}
}
```

With the above configuration, when you run `terragrunt XXX`, Terragrunt would, the first time, do the following:

```
terraform init \
  -backend-config=encrypt=true \
  -backend-config=bucket=my-bucket \
  -backend-config=key=(some computed value) \
  -backend-config=region=us-east-1
```

There are no bash scripts involved.  No files to auto-generate and clean up. Just piggy back on the `init` command and pass the values using the `-backend-config` option.

Does that make sense?

jocgir:
Hi @brikis98, that could work. I didn't realized that init with args was enough to store the configuration and that it was not necessary to provision the terraform - backend section to make it work on further calls.

brikis98:
@jocgir Thank you for the PR. I ended up implementing the approach I mentioned in #167, so I'm going to close this one.

PR #162, opened 2017-04-11
Build binaries with CGO_ENABLED=0 

### LINKED ISSUE #161: Build terragrunt releases built as static binaries
Currently terragrunt binary releases are not built statically, but linked to libc:

```
$ ldd terragrunt_linux_amd64 
	linux-vdso.so.1 =>  (0x00007fffe704b000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f143f603000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f143f23e000)
	/lib64/ld-linux-x86-64.so.2 (0x000055a7004a3000)
```

This prevents me to use inside of [an alpine container](https://hub.docker.com/_/alpine/).

I suggest you build them using static: 

```
$ CGO_ENABLED=0 go  get  github.com/gruntwork-io/terragrunt
...
$ ldd terragrunt 
	not a dynamic executable
```

josh-padnick:
This PR updates circle.yml to use the latest `build-go-binaries` module (v0.3.10). That version has been updated to build Go binaries with CGO_ENABLED=0 by default. Fixes #161.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -6,7 +6,7 @@ dependencies:
   override:
     # Install the gruntwork-module-circleci-helpers and use it to configure the build environment and run tests.
     - curl -Ls https://raw.githubusercontent.com/gruntwork-io/gruntwork-installer/master/bootstrap-gruntwork-installer.sh | bash /dev/stdin --version 0.0.13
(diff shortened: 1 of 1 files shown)


josh-padnick:
@brikis98 Feedback welcome but since you're on vacation, I'll merge now.

PR #167, opened 2017-04-20
Update Terragrunt to support Terraform 0.9

### LINKED ISSUE #158: How Terragrunt Will Handle Terraform 0.9.0 Compatibility

brikis98:
This PR updates Terragrunt so it works with Terraform 0.9.x. This finally fixes #158.

Here’s the approach I used:

1. Instead of calling `terraform remote config` (which was removed in Terraform 0.9), we now call `terraform init` and pass `-backend-config` parameters as necessary.

1. I have removed support for DynamoDB locking, since this is now built into Terraform itself. To enable it, you just specify a `lock_table` attribute in your s3 backend configuration. I’ve updated Terragrunt to automatically create this table if it doesn’t exist already. If you want to Terraform to retry locks, you can use an `extra_arguments` block.

1. I’ve completely refactored the docs to emphasize that Terragrunt is now focused not on locking and remote state, but working with multiple modules and keeping your Terraform code DRY.

I’ve included a section in the README on how to migrate from the old Terragrunt and Terraform to this new version. Feedback is welcome!

It may be easier to browse the README directly in the branch: https://github.com/gruntwork-io/terragrunt/tree/terraform-0.9. You can find the migration instructions here: https://github.com/gruntwork-io/terragrunt/tree/terraform-0.9#migrating-from-terragrunt-v011x-and-terraform-08x-and-older.


README.md
```diff
@@ -1,636 +1,539 @@
...
+Terragrunt has the ability to download remote Terraform configurations. The idea is that you define the Terraform code 
+for all of your infrastructure in a single repo, called, for example, `modules`:
 
-[terragrunt] 2016/05/27 00:39:19 Attempting to release lock for state file my-app in DynamoDB
-[terragrunt] 2016/05/27 00:39:19 Lock released!
+```
+└── modules
+    ├── prod
+    │   ├── app
+    │   │   └── main.tf
```
tomstockton: Doesn't this pattern still end up with a lot of copy and paste between the resources defined in `prod/app/main.tf` and `otherenv/app/main.tf`?  Does the `modules` repo actually need to be environment specific?

brikis98: Whoops! Good catch. That's a copy/paste error. I've updated that example with the proper folder structure for the `modules` repo: https://github.com/gruntwork-io/terragrunt/pull/167/commits/ef92f6e7c47e4f2c90eb2e0f18293b7f0bc7e730.

tomstockton:
How does this handle environment specific remote state files?  I imagine this would be best done using [state environments](https://www.terraform.io/docs/state/environments.html) but I didn't notice it in the code.

brikis98:
> How does this handle environment specific remote state files? 

What are you referring to with "this"?

tomstockton:
I'm referring to this change / PR.

brikis98:
I'm not sure this PR has any impact on remote state files... Can you give an example of what you mean?

tomstockton:
I've reread the README, let me rewrite my initial question better!

Are you considering using the Terraform native support for [state environments](https://www.terraform.io/docs/state/environments.html) or is it not needed due to how the remote statefile path is derived using `"${path_relative_to_include()}/terraform.tfstate"`?

brikis98:
I think Terraform's new state environments are a completely alternative way to manage multiple environments from what we describe in the Terragrunt README. The Terragrunt approach is to define all those environments in files on disk so that you can read the files, version them, revert them, etc. The state environments approach is to run a command to pick a different environment and perhaps to put conditional statements throughout your code so it behaves differently based on which environment was selected. I find the former approach easier to understand and less error prone, but AFAIK, nothing in Terragrunt or this PR prevents you from using the latter approach.

tomstockton:
Thanks for the clarification.  I was slightly confused about the use case for the state environments as it seemed to conflict with your 'Golden Rules' of Terraform!  Good to hear your thoughts on it.

README.md
```diff
@@ -643,340 +531,322 @@ terragrunt = {
         "refresh"
       ]
     }
+  }
+}
+```
```
josh-padnick: I had to read this a few times to get it. I think the word "switch" through me. Maybe reword? Also, I didn't immediately make the connection you were still using a list of strings but had the first and second string as arg and arg-value.

README.md
```diff
@@ -643,340 +531,322 @@ terragrunt = {
...
 ```
 
-### The apply-all, destroy-all, and output-all commands
 
-Let's say you have a single environment (e.g. `stage` or `prod`) that has a number of Terraform modules within it:
+
+
+
+
+### Work with multiple Terraform modules
```
josh-padnick: Maybe better wording here would be "Quickly deploy multiple Terraform modules."

README.md
```diff
@@ -643,340 +531,322 @@ terragrunt = {
...
+Inside each `terraform.tfvars` file, add a `terragrunt = { ... }` block to identify this as a module managed by 
+Terragrunt (the block can be empty or include any of the configs described in this documentation):
+
+```hcl
+terragrunt = {
+  # Put your Terragrunt configuration here
+}
+```
+
+Now you can go into the root folder and deploy all the modules within it by using the `apply-all` command:
```
josh-padnick: Can you clarify what you mean by "root folder". If there are multiple environments, each with their own folder tree, does "root" mean any folder that contains one or more subfolders that contain` terraform.tfvars` files?

README.md
```diff
@@ -643,340 +531,322 @@ terragrunt = {
...
 ```
-cd my-terraform-repo/stage
+cd root
 terragrunt apply-all
 ```
 
-When you run this command, Terragrunt will find all `terraform.tfvars` files in the subfolders of the current working 
-directory that contain `terragrunt = { ... }` blocks, and run `terragrunt apply` in each one concurrently.
+When you run this command, Terragrunt will look through all the subfolders of the current working directory, find all 
+`terraform.tfvars` files with a `terragrunt = { ... }` block, and run `terragrunt apply` in each one concurrently.
```
josh-padnick: Ok, I guess you what root means here...

josh-padnick: But what about recursive subfolders?

README.md
```diff
@@ -643,340 +531,322 @@ terragrunt = {
...
   dependencies {
     paths = ["../vpc", "../backend-app"]
   }
 }
 ```
 
-Once you've specified the depenedencies in each `terraform.tfvars` file, when you run the `terragrunt apply-all` and 
+Once you've specified the dependencies in each `terraform.tfvars` file, when you run the `terragrunt apply-all` or 
 `terragrunt destroy-all`, Terragrunt will ensure that the dependencies are applied or destroyed, respectively, in the
 correct order. For the example at the start of this section, the order for the `apply-all` command would be:
```
josh-padnick: ...while applying or destroying Terraform modules in parallel where possible.

README.md
```diff
@@ -643,340 +531,322 @@ terragrunt = {
...
+as the environment variable `TF_VAR_foo` and to read that value in using this `get_env` helper function.
 
-## CLI Options
+
+
+
+### CLI Options
 
 Terragrunt forwards all arguments and options to Terraform. The only exceptions are the options that start with the
 prefix `--terragrunt-`. The currently available options are:
```
josh-padnick: The other exception is `--version`.

README.md
```diff
@@ -643,340 +531,322 @@ terragrunt = {
...
 
 * `--terragrunt-config`: A custom path to the `terraform.tfvars` file. May also be specified via the `TERRAGRUNT_CONFIG`
   environment variable. The default path is `terraform.tfvars` in the current directory (see 
   [Terragrunt config files](#terragrnt-config-files) for a slightly more nuanced explanation).
```
josh-padnick: Does not apply with `apply-all`, etc. subcommands.

README.md
```diff
@@ -643,340 +531,322 @@ terragrunt = {
...
+
 * `--terragrunt-working-dir`: Set the directory where Terragrunt should execute the `terraform` command. Default is the
   current working directory. Note that for the `apply-all` and `destroy-all` directories, this parameter has a different 
   meaning: Terragrunt will apply or destroy all the Terraform modules in the subfolders of the 
   `terragrunt-working-dir`, running `terraform` in the root of each module it finds.
+
 * `--terragrunt-source`: Download Terraform configurations from the specified source into a temporary folder, and run 
   Terraform in that temporary folder. May also be specified via the `TERRAGRUNT_SOURCE` environment variable. The 
   source should use the same syntax as the [Terraform module source](https://www.terraform.io/docs/modules/sources.html) 
   parameter.
```
josh-padnick: Does not apply with apply-all, etc. subcommands.

README.md
```diff
@@ -1039,7 +911,122 @@ To migrate, all you need to do is:
...
+          "untaint"
+        ]
+      
+        arguments = [
+          "-lock-timeout=20m"
+        ]    
+      }     
+    }     
+    ``` 
+
```
josh-padnick: These migration instructions are great: two simple steps.

circle.yml
```diff
@@ -7,8 +7,7 @@ dependencies:
     # Install the gruntwork-module-circleci-helpers and use it to configure the build environment and run tests.
     - curl -Ls https://raw.githubusercontent.com/gruntwork-io/gruntwork-installer/master/bootstrap-gruntwork-installer.sh | bash /dev/stdin --version 0.0.13
```
josh-padnick: Upgrade gruntwork-installer to latest version.

cli/cli_app.go
```diff
@@ -111,18 +99,13 @@ func CreateTerragruntCli(version string, writer io.Writer, errwriter io.Writer)
...
 	app.Writer = writer
 	app.ErrWriter = errwriter
-	app.UsageText = fmt.Sprintf(`Terragrunt is a thin wrapper for [Terraform](https://www.terraform.io/) that supports locking
-   via Amazon's DynamoDB and enforces best practices. Terragrunt forwards almost all commands, arguments, and options
-   directly to Terraform, using whatever version of Terraform you already have installed. However, before running
-   Terraform, Terragrunt will ensure your remote state is configured according to the settings in %s.
-   Moreover, for the apply and destroy commands, Terragrunt will first try to acquire a lock using DynamoDB. For
-   documentation, see https://github.com/gruntwork-io/terragrunt/.`, config.DefaultTerragruntConfigPath)
+	app.UsageText = `Terragrunt is a thin wrapper for Terraform that provides extra tools for working with multiple
+   Terraform modules, remote state, and locking. For documentation, see https://github.com/gruntwork-io/terragrunt/.`
```
josh-padnick: Nice update.

dynamodb/dynamo_lock_table.go
```diff
@@ -15,6 +16,26 @@ import (
 // running many automated tests in parallel, we use a counting semaphore
 var tableCreateDeleteSemaphore = NewCountingSemaphore(10)
 
+// Terraform requires the DynamoDB table to have a primary key with this name
+const ATTR_LOCK_ID = "LockID"
```
josh-padnick: What are the implications of this with multiple Terraform/Terragrunt modules each writing a lock? Does that mean that otherwise parallelizable Terraform modules can no longer be run in parallel?

dynamodb/dynamo_lock_table_test.go
```diff
@@ -0,0 +1,70 @@
...
+		waitGroup.Add(1)
+		go func() {
+			defer waitGroup.Done()
+			err := CreateLockTableIfNecessary(tableName, client, mockOptions)
+			assert.Nil(t, err, "Unexpected error: %v", err)
+		}()
+	}
+
+	waitGroup.Wait()
+}
```
josh-padnick: Nice test.

remote/remote_state_s3.go
```diff
@@ -14,11 +15,12 @@ import (
...
-	Bucket  string
-	Key     string
-	Region  string
-	Profile string
+	Encrypt   string `mapstructure:"encrypt"`
+	Bucket    string `mapstructure:"bucket"`
+	Key       string `mapstructure:"key"`
+	Region    string `mapstructure:"region"`
+	Profile   string `mapstructure:"profile"`
+	LockTable string `mapstructure:"lock_table"`
```
josh-padnick: Just observing that it's a little weird that `LockTable` is part of the `RemoteStateConfigS3` struct (versus something like a `RemoteStateConfigAWS` struct, but still makes sense.

brikis98:
> One thought I had is that after reading everything, the reader still has to make one more leap on their own: What is the ideal, complete setup with Terragrunt that combines all these use cases? For that reason, it'd be nice if we had a _sample folder with maybe live and modules subfolders which mimicked a real setup and showed exactly what the root terraform.tfvars file and each "child" terraform.tfvars file looks like.

That's a *great* idea, but I think it's outside the scope of this PR. Could you file a separate bug to track this?

README.md
```diff
@@ -1,636 +1,524 @@
...
+  state storage and the `bucket` you specify in `remote_state.config` doesn't already exist, Terragrunt will create it 
+  automatically, with [versioning enabled](http://docs.aws.amazon.com/AmazonS3/latest/dev/Versioning.html). 
+
+* **DynamoDB table**: If you are using the [S3 backend](https://www.terraform.io/docs/backends/types/s3.html) for 
+  remote state storage and you specify a `lock_table` (a [DynamoDB table used for 
+  locking](https://www.terraform.io/docs/backends/types/s3.html#lock_table)) in `remote_state.config`, if that table
+  doesn't already exist, Terragrunt will create it automatically, including a primary key called `LockID`.
+
+**Note*: If you specify a `profile` key in `remote_state.config`, Terragrunt will automatically use this AWS profile 
+when creating the S3 bucket or DynamoDB table.
```
brikis98: My preference is to keep all Terraform code (`.tf`) in the `modules` repo, even if it's only used in one environment. Versioning and consistency are both good things and the amount of overhead this adds, especially with the `--terragrunt-source` param, is fairly minor if you're already using this for everything else.

README.md
brikis98: Sorry, what does this refer to?

josh-padnick:
> That's a *great* idea, but I think it's outside the scope of this PR. Could you file a separate bug to track this?

Thanks! https://github.com/gruntwork-io/terragrunt/issues/169

README.md
```diff
@@ -1,636 +1,524 @@
...
+}
 
-#### Remote state management configuration
+instance_count = 3
+instance_type = "t2.micro"
+```
 
-For remote state management, Terragrunt supports the following settings in `terraform.tfvars`:
+*(Note: the double slash (`//`) is intentional and required. It's part of Terraform's Git syntax for [module 
+sources](https://www.terraform.io/docs/modules/sources.html).)
```
brikis98: Good catch. Fixed: https://github.com/gruntwork-io/terragrunt/pull/167/commits/5af50b0d64c2cf113a35458edbac78085597401c

README.md
```diff
@@ -1,636 +1,524 @@
...
+environment-specific values in the `.tfvars` files, which is about as close to the "essential complexity" of the 
+problem as you can get.
 
-* Includes
-* Find parent helper
-* Relative path helper
-* Overriding included settings
-* The `apply-all`, `destroy-all`, and `output-all` commands
-* Dependencies between modules
+Just as importantly, since the Terraform code is now defined in a single repo, you can version it (e.g., using Git
```
brikis98: Fixed: https://github.com/gruntwork-io/terragrunt/pull/167/commits/5af50b0d64c2cf113a35458edbac78085597401c

README.md
```diff
@@ -1,636 +1,524 @@
...
+│   └── terraform.tfvars
+└── vpc
+    ├── main.tf
+    └── terraform.tfvars
+```
 
-You can read in environment variables in the `terragrunt = { ... }` portion of your `terraform.tfvars` file using the 
-`get_env()` helper function:
+In your root `terraform.tfvars` file, you can define your entire remote state configuration just once in a 
+`remote_state` block, as follows:
```
brikis98: Fixed: https://github.com/gruntwork-io/terragrunt/pull/167/commits/5af50b0d64c2cf113a35458edbac78085597401c

README.md
```diff
@@ -1,636 +1,524 @@
...
-is mandatory but can be empty (e.g. `${get_env("ENVIRONMENT_VARIABLE_NAME", "")}`).
+The `remote_state` block supports all the same [backend types](https://www.terraform.io/docs/backends/types/index.html) 
+as Terraform. The next time you run `terragrunt`, it will automatically configure all the settings in the
+`remote_state.config` block, if they aren't configured already, by calling [terraform 
+init](https://www.terraform.io/docs/commands/init.html).
 
-If there is no environment variable with that name registered in the system, the configuration file would be evaluated 
-to:
+In each of the child `terraform.tfvars` files, such as `mysql/terraform.tfvars`, you can tell Terragrunt to 
+automatically include all the settings from the root `terraform.tfvars` file as follows:
```
brikis98: Fixed: https://github.com/gruntwork-io/terragrunt/pull/167/commits/5af50b0d64c2cf113a35458edbac78085597401c

README.md
```diff
@@ -1,636 +1,524 @@
...
-      bucket = "value-bucket"
-      key = "/foo/bar/terraform.tfstate"
-      region = "us-west-2"
-    }
-  }
-}
-```
+* `path_relative_to_include()`: This helper function returns the relative path between the current `terraform.tfvars` 
+  file and the path specified in its `include` block. We use this so that each Terraform module stores its Terraform
+  state at a different `key`. For example, the `mysql` module will have its `key` parameter resolve to
```
brikis98: Fixed: https://github.com/gruntwork-io/terragrunt/pull/167/commits/5af50b0d64c2cf113a35458edbac78085597401c

README.md
brikis98: Good catch. Fixed: https://github.com/gruntwork-io/terragrunt/pull/167/commits/5af50b0d64c2cf113a35458edbac78085597401c

README.md
```diff
@@ -1,636 +1,524 @@
...
+    extra_arguments "retry_lock" {
+      commands = [
+        "init",
+        "apply",
+        "refresh",
+        "import",
+        "plan",
+        "taint",
+        "untaint"
+      ]
```
brikis98: > This list of terraform commands that apply to locks is non-obvious. It'd be nice if we could another helper command here that generates this array for you.

This is another good candidate for a separate issue, if you don't mind filing it.

> BTW, where does retry_lock come from? I didn't see it in Terraform docs after a quick search, so I'm guessing it's an arbitrary var name? Would be good to clarify either way.

Arbitrary name. I updated the docs to clarify: https://github.com/gruntwork-io/terragrunt/pull/167/commits/5af50b0d64c2cf113a35458edbac78085597401c

README.md
brikis98: Updated the docs to make this clearer: https://github.com/gruntwork-io/terragrunt/pull/167/commits/5af50b0d64c2cf113a35458edbac78085597401c

README.md
brikis98: Well, it also allows you to destroy and run output for multiple modules, so I'll stick with "work with".

README.md
brikis98: The root here was literally the folder named root in example folder structure above. I added syntax highlighting to the word to clarify that: https://github.com/gruntwork-io/terragrunt/pull/167/commits/5af50b0d64c2cf113a35458edbac78085597401c

README.md
brikis98: That includes recursive subfolders. I added the word "recursive" to clarify: https://github.com/gruntwork-io/terragrunt/pull/167/commits/5af50b0d64c2cf113a35458edbac78085597401c

README.md
brikis98: Added: https://github.com/gruntwork-io/terragrunt/pull/167/commits/5af50b0d64c2cf113a35458edbac78085597401c

README.md
brikis98: Added: https://github.com/gruntwork-io/terragrunt/pull/167/commits/5af50b0d64c2cf113a35458edbac78085597401c

README.md
brikis98: Added: https://github.com/gruntwork-io/terragrunt/pull/167/commits/5af50b0d64c2cf113a35458edbac78085597401c

README.md
josh-padnick: Consider the use case of adding Route53 Public Hosted Zones. So your position would be to write a trivial Terraform module that does the following:

```hcl
resource "aws_route53_zone" "primary" {
  name = "${var.domain_name}"
}
```

...so that you could have the following Terragrunt files in the `live` repo:

```hcl
terragrunt = {
  include {
  }
  terraform {
    source = "git::git@github.com:foo/modules.git//route53_hosted_zone?ref=v0.0.3"
  }
}

domain_name = "example.com"
# Alternatively
domain_names = ["example.com"]
```

I guess you're arguing the standardization gives us the `apply-all` command, and forces any complexity into a versionable module. In this extreme trivial case, the overhead outweighs the value but still gives us consistency.

I'd say the one potential area is this may break down a little is if Terraform's HCL isn't expressive enough to capture what we want. For example, I often wish modules accepted lists of lists, but since we can only pass "flat" lists, we may have to instantiate one module per each "thing".

Mostly thinking out loud here...

circle.yml
brikis98: Updated: https://github.com/gruntwork-io/terragrunt/pull/167/commits/854f00e57f939d6980fe62aa7ce117a8a2d4190f

dynamodb/dynamo_lock_table.go
brikis98: `LockID` is the name of the column in DynamoDB. The value that gets written to it is the `key` parameter of the S3 bucket (perhaps with the S3 bucket name prefixed, I forget). Therefore, it's a unique lock for each unique state file.

remote/remote_state_s3.go
brikis98: It's part of the S3 backend definition: https://www.terraform.io/docs/backends/types/s3.html#lock_table

brikis98:
Merging!

README.md
josh-padnick: https://github.com/gruntwork-io/terragrunt/issues/171

brikis98: Thx!

PR #175, opened 2017-04-22
New helper functions

jocgir:
Added two helper functions: `path_relative_from_include()` which is the counter part of `path_relative_to_include()` and `get_aws_account_id()` which returns the current AWS account number.
Updated the documentation accordingly _(Read documentation to view use cases for the new interpolation functions)_.
Added state to terraform remote state configuration.

_Note that there are a lot of changes in the `README.md` file. This is because my editor automatically removed trailing blanks. But only the documentation relative to the two new functions has really been modified._

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Terragrunt
 
-Terragrunt is a thin wrapper for [Terraform](https://www.terraform.io/) that provides extra tools for working with 
+Terragrunt is a thin wrapper for [Terraform](https://www.terraform.io/) that provides extra tools for working with
 multiple Terraform modules, remote state, and locking.
 
 
@@ -14,17 +14,17 @@ multiple Terraform modules, remote state, and locking.
    the binary for your OS, renaming it to `terragrunt`, and adding it to your PATH.
 
 1. Go into a folder with your Terraform configurations (`.tf` files) and create a `terraform.tfvars` file with a
-   `terragrunt = { ... }` block that contains the configuration for Terragrunt (check out the [Use cases](#use-cases) 
+   `terragrunt = { ... }` block that contains the configuration for Terragrunt (check out the [Use cases](#use-cases)
    section for the types of configuration Terragrunt supports):
-   
+
     ```hcl
     terragrunt = {
       # (put your Terragrunt configuration here)
     }
     ```
-    
+
 1. Now, instead of running `terraform` directly, run all the standard Terraform commands using `terragrunt`:
-   
+
     ```bash
     terragrunt get
     terragrunt plan
@@ -32,14 +32,14 @@ multiple Terraform modules, remote state, and locking.
     terragrunt output
     terragrunt destroy
     ```
-   
+
    Terragrunt forwards almost all commands, arguments, and options directly to Terraform, using whatever version of
    Terraform you already have installed. However, based on the settings in your `terraform.tfvars` file, Terragrunt can
-   configure remote state, locking, extra arguments, and lots more. 
-    
-       
-       
-       
+   configure remote state, locking, extra arguments, and lots more.
+
+
+
+
 
 
 ## Use cases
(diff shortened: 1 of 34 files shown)


config/config_helpers.go
```diff
@@ -157,6 +165,33 @@ func pathRelativeToInclude(include *IncludeConfig, terragruntOptions *options.Te
...
+	if err != nil {
+		return "", errors.WithStackTrace(err)
+	}
+
+	return filepath.ToSlash(filepath.Dir(relativePath)), nil
+}
+
+// Return the AWS account id associated to the current set of credentials
+func getAWSAccountID(terragruntOptions *options.TerragruntOptions) (string, error) {
+	iamSession := iam.New(session.Must(session.NewSession()))
```
brikis98: Will `session.Must` panic? If so, I'd rather return an explicit error than rely on a panic, as that keeps the code easier to reason about.

config/config_helpers.go
```diff
@@ -157,6 +165,33 @@ func pathRelativeToInclude(include *IncludeConfig, terragruntOptions *options.Te
...
+
+	result, err := iamSession.ListUsers(&iam.ListUsersInput{MaxItems: aws.Int64(1)})
```
brikis98: Instead of calling `ListUsers`, I think [GetCallerIdentity](https://docs.aws.amazon.com/sdk-for-go/api/service/sts/#STS.GetCallerIdentity) is a better fit. It returns the account ID (so no need to parse the ARN) and it requires no extra permissions.

jocgir:
Hi @brikis98, I pushed a new version of the documentation (without the trailing blanks removal to help reading).

Regarding get_aws_account_id(), I understand your concern. Our problem is that we use the account id to make sure that our remote state bucket is unique. We have several account and it is the way we use to ensure that the bucket name is unique. We are facing an egg and chicken problem because we cannot access the terraform data source that gives us the accound id while we are configuring terragrunt to use that bucket name.

It can also help to determine the environment automatically based on AWS credentials. This could avoid an error of pushing the wrong environment in the wrong account (which could happen by relying solely on environment variables).

I also realized after implementing path_relative_from include() that what we really needed was a function like get_tfvars_dir() but returning the absolute path of the root configuration instead of the leaf. I added it (get_parent_dir(), but I wonder if path_relative_from_include() is still required).

Finally, I have a question for you. Is the limitation of not being able to have cascading find_in_parent_folders() is by design or you have any other reason. We structured our terraform folders with a subdirectory named private when we have sensitive element like secrets and don't want to make them visible to everybody. We can then control the access to the remote state file through IAM rules and ensure that secrets are not exposed in tfstate that have to be referred by other terraform folders.

config/config_helpers.go
jocgir: Done

config/config_helpers.go
jocgir: Nice idea, we never realized that this call existed and alway relied on iam. Thank you!

README.md
```diff
@@ -944,6 +1019,58 @@ For the example above, this path will resolve to `/terraform-code/frontend-app/.
 what you want.
 
 
+#### get_parent_dir
```
brikis98: Ah, this is a nice helper to have for sure. Perhaps a clearer and more consistent name would be `get_parent_tfvars_dir`?

brikis98: Also, could you add an example usage here in the docs?

brikis98:
> Our problem is that we use the account id to make sure that our remote state bucket is unique. We have several account and it is the way we use to ensure that the bucket name is unique.

Don't the AWS creds you use determine which account gets used? Therefore, if you accidentally plug in the wrong AWS creds, you'd end up with the wrong bucket name by accident...

> Is the limitation of not being able to have cascading find_in_parent_folders() is by design or you have any other reason.

It's mainly to keep things simpler. If you know there is only one level of includes, it's easier to reason about the code, and the implementation is simpler too. I wasn't aware of any real-world use cases for multiple levels of includes, so this seemed like a reasonable limitation for now.

> We structured our terraform folders with a subdirectory named private when we have sensitive element like secrets and don't want to make them visible to everybody. We can then control the access to the remote state file through IAM rules and ensure that secrets are not exposed in tfstate that have to be referred by other terraform folders.

How does having multiple levels of includes help with this? Perhaps a concrete example of the folder structure and various `.tfvars` files would help me understand.

README.md
jocgir: Done

I did not add a test for get_aws_account_id() since the expected result would be different and validating the expected result would require to implement the exact same function in the test.

brikis98:
Thanks!

brikis98:
Just created a new release here: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.12.7

Issue #176, opened 2017-04-22
Backend not populated


andrewrynhard:
I am trying to follow the migration steps and running into an issue with S3 backend.
```
$ terragrunt --version
terragrunt version v0.12.2
$ terraform version
Terraform v0.9.3
```
## Main Repo
```
tree .
.
├── module
│   └── terraform.tfvars
└── terraform.tfvars

1 directory, 2 files
```
```
cat terraform.tfvars
terragrunt = {
  remote_state {
    backend = "s3"
    config {
      bucket     = "example-tfstate"
      region     = "us-west-1"
      key        = "terraform.tfstate"
      profile     = "production"
      lock_table = "example-lock"
    }
  }
}
```
```
cat ./module/terraform.tfvars
terragrunt = {
  include {
    path = "${find_in_parent_folders()}"
  }

  terraform {
    source = "/Users/user/Desktop/example-module//"

    extra_arguments "common" {
      arguments = [
        "-var-file=terraform.tfvars"
      ]
      commands = [
        "apply",
        "plan",
        "destroy",
        "import",
        "push",
        "refresh"
      ]
    }
  }
}
```
## Module Repo
```
tree .
.
└── main.tf
```
```
cat ~/Desktop/example-module/main.tf
terraform {
  # # The configuration for this backend will be filled in by Terragrunt
  backend "s3" {}
}

provider "aws" {
  region = "us-west-1"
  profile = "production"
}

resource "aws_key_pair" "example" {
  key_name   = "test"
  public_key = "${file("~/.ssh/id_rsa.pub")}"
}
```
```
$ pwd
/Users/user/Desktop/test/module
$ terragrunt plan
[terragrunt] 2017/04/22 12:13:57 Reading Terragrunt config file at /Users/user/Desktop/test/module/terraform.tfvars
[terragrunt] 2017/04/22 12:13:57 WARNING: no double-slash (//) found in source URL /Users/user/Desktop/module. Relative paths in downloaded Terraform code may not work.
[terragrunt] 2017/04/22 12:13:57 Downloading Terraform configurations from file:///Users/user/Desktop/module into /var/folders/sy/qh4qyxxj76340bgw0ntv_3y80000gn/T/terragrunt-download/n8h83LGwSiRPlykhWzKbI-yw4mo=/DB3gG7ab1KD1epljDruWOCYL230=
[...]

PR #177, opened 2017-04-23
Fix init command used to download remote configurations

### LINKED ISSUE #174: How to remote configurations + remote state + multiple modules, goes wrong
I have to following and I'm not sure what I'm doing wrong, I want to be able to run two commands:

```shell
terragrunt apply # on the network folder to create just the network
```

and/or

```shell
terragrunt apply-all # on the root folder `terraform` to create network, database and compute resources
```

My project structure looks like:

screen shot 2017-04-22 at 13 18 46

My `terraform.tfvars` files looks like:

**terraform/network/terraform.tfvars**
```hcl
terragrunt = {
    terraform {
        source = "../../modules//network"
    },
    include {
    }
}
```

**terraform/terraform.tfvars**
```hcl
terragrunt = {
    terraform {
        extra_arguments "custom_vars" {
            commands = [
                "plan",
                "apply",
                "destroy"
            ]
            arguments = [
                "-var-file=${get_tfvars_dir()}/../variables.tfvars"
            ]
        }
    },
    remote_state {
        backend = "s3"
        config {
            bucket  = "my-bucket"
            key     = "aws/myapp/${path_relative_to_include()}/terraform.tfstat"
            region  = "eu-west-1"
        }
    }
}
```
I'm looking forward to move those `modules` into their own repo and then be able to do a proper `source=git::git@...` and pull the source code from git however I can't manage to make it work, it looks like the tmp folder got created properly and the code from my modules it's there however when it comes to run the terraform command it can not find the variables and terraform ask me to introduce them, on the tmp folder there are not `variables.tfvars` file so I guess that's why it's asking me for them.

I reckon I have misunderstood something and mixing concepts in here hope someone can point me out in the right direction, ta

brikis98:
This PR fixes #174 and #176. We’ve been using the `terraform init` command to download remote Terraform configurations for a while now. As of Terraform 0.9, this same command also tries to configure backends, download modules, etc. When updating Terragrunt to support Terraform 0.9, we forgot to update this original `init` call to pass `backend` parameters.

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -42,6 +42,24 @@ var DEPRECATED_COMMANDS = map[string]string{
 	CMD_TEAR_DOWN: CMD_DESTROY_ALL,
 }
 
+var TERRAFORM_COMMANDS_THAT_USE_STATE = []string{
+	"apply",
+	"destroy",
+	"env",
+	"import",
+	"graph",
+	"output",
+	"plan",
+	"push",
+	"refresh",
+	"show",
+	"taint",
+	"untaint",
+	"validate",
+	"force-unlock",
+	"state",
+}
+
 // Since Terragrunt is just a thin wrapper for Terraform, and we don't want to repeat every single Terraform command
 // in its definition, we don't quite fit into the model of any Go CLI library. Fortunately, urfave/cli allows us to
 // override the whole template used for the Usage Text.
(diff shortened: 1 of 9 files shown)


cli/cli_app.go
```diff
@@ -42,6 +42,24 @@ var DEPRECATED_COMMANDS = map[string]string{
...
+	"plan",
+	"push",
+	"refresh",
+	"show",
+	"taint",
+	"untaint",
+	"validate",
+	"force-unlock",
+	"state",
+}
```
brikis98: This change is not directly related to this PR, but also something I missed when updating Terragrunt to work with Terraform 0.9.

test/integration_test.go
```diff
@@ -282,6 +284,48 @@ func TestLocalWithRelativeExtraArgs(t *testing.T) {
 	runTerragrunt(t, fmt.Sprintf("terragrunt apply --terragrunt-non-interactive --terragrunt-working-dir %s", TEST_FIXTURE_LOCAL_RELATIVE_ARGS_DOWNLOAD_PATH))
 }
 
+func TestLocalWithBackend(t *testing.T) {
```
brikis98: This test case would fail before this PR

brikis98:
@josh-padnick I'm going to merge this in so I can submit a couple more fixes. Feedback welcome!

PR #178, opened 2017-04-23
Update Terraform state parsing to use backends

brikis98:
Terragrunt parses the `.terraform/terraform.tfstate` file to check whether the remote state configuration has changed and `terraform init` needs to be run. Before Terraform 0.9, the `.terraform/terraform.tfstate` file stored remote state configuration under the `remote` key. As of Terraform 0.9, it is now under the `backend` key. This PR updates Terragrunt accordingly.

As a result of applying this fix, I realized that my fix for #177 was not quite right. Trying to download the source code and init backends in one command is a bad idea, as some of our checks (e.g. whether an s3 bucket should be created or modules downloaded) depend on the code already being in place. Therefore, I've changed that fix to call `init` to download the source code *without* doing any remote state setup, which then allows the `init` call later on to handle all the remote state stuff separately.

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -175,7 +175,7 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 	}
 
 	if sourceUrl, hasSourceUrl := getTerraformSourceUrl(terragruntOptions, conf); hasSourceUrl {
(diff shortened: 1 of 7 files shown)


brikis98:
@josh-padnick Merging this one too. Feedback welcome!

Issue #172, opened 2017-04-21
Ensure minimum terraform version


robkinyon:
With all the migration of features back into Terraform, Terragrunt needs to be able to guarantee a minimum Terraform version. Otherwise, if a tfvars file doesn't specify locking and someone accidentally uses an old version of Terraform, bad things can happen.

There should be a default minimum version and teams should be able to override that version if desired.

brikis98:
Ah, that's a great point. PRs for that are definitely welcome.

brikis98:
Basic fix in #180. Should be available here: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.12.5

robkinyon:
In addition to a minimum Terraform version, there should also be a way of specifying a minimum *Terragrunt* version within configuration. If I knew golang better, I'd offer a PR for this.

brikis98:
Ah, that would be handy too. Could you file a separate issue for that?

brikis98:
Closing this issue so we can focus on #182.

PR #180, opened 2017-04-23
Add basic support for minimum version number check

brikis98:
This is a fix for #172. 

This PR updates Terragrunt to check that a minimum version of Terraform is installed. For now, this is hard-coded to `v0.9.3`. 

In the future, we could make this configurable. That said, Terraform has its own [required_version](https://www.terraform.io/docs/configuration/terraform.html) support, so perhaps more fine-grained control should be specified in the `.tf` files themselves.

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -97,6 +97,8 @@ AUTHOR(S):
 
 var MODULE_REGEX = regexp.MustCompile(`module[[:blank:]]+".+"`)
 
+const DEFAULT_TERRAFORM_VERSION_CONSTRAINT = ">= v0.9.3"
+
 const TERRAFORM_EXTENSION_GLOB = "*.tf"
 
 // Create the Terragrunt CLI App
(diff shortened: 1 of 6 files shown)


brikis98:
@josh-padnick Merging this one too

cli/cli_app.go
```diff
@@ -97,6 +97,8 @@ AUTHOR(S):
 
 var MODULE_REGEX = regexp.MustCompile(`module[[:blank:]]+".+"`)
 
+const DEFAULT_TERRAFORM_VERSION_CONSTRAINT = ">= v0.9.3"
```
josh-padnick: Consider adding comment that this expression will be evaluated by https://github.com/hashicorp/go-version

brikis98: Done: https://github.com/gruntwork-io/terragrunt/commit/86721319ffb0a454f05ace37940f85b7d78f3a60

brikis98:
> What if there's a regression in t0.9.3 so that some users want to use t0.9.2 or lower? It may be worth either adding a flag that disables this check or better yet, allowing a custom flag to be specified by the user to avoid this admittedly rare situation.

Yea, I suspect we'll eventually add the ability to specify minimum Terraform and Terragrunt versions directly in the Terragrunt configuration in `terraform.tfvars`. Not sure it's a high enough priority right now.

Issue #179, opened 2017-04-23
Backend encryption is broken


andrewrynhard:
### terraform.tfvars
```
terragrunt = {
  remote_state {
    backend = "s3"
    config {
      bucket     = "example"
      encrypt = true
      key        = "${path_relative_to_include()}/terraform.tfstate"
      lock_table = "example"
      profile     = "example"
      region     = "us-west-1"
    }
  }
}
```

 and the output is:
```
[terragrunt] 2017/04/23 08:04:46 Reading Terragrunt config file at /Users/andrewrynhard/workspace/projects/LogicMonitor/Kubernetes/k8s-terraform/production/us-west-1/blue/terraform.tfvars
At 6:17: root.terragrunt.remote_state.config.encrypt: unknown type for string *ast.LiteralType
[terragrunt] 2017/04/23 08:04:46 At 6:17: root.terragrunt.remote_state.config.encrypt: unknown type for string *ast.LiteralType
[terragrunt] 2017/04/23 08:04:46 Unable to determine underlying exit code, so Terragrunt will exit with error code 1
```

brikis98:
Ah, another good catch. We used to use strings for all remote state configuration. Looks like the `backend` section in Terraform 0.9 uses proper types, including booleans for the `encrypt` param. Fixing coming shortly :)

brikis98:
Should be fixed by #181. Available shortly here: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.12.6

PR #181, opened 2017-04-23
Backend config should use proper types, including bool for encrypt

brikis98:
This is a fix for #179. While Terragrunt’s original `remote_state` config used solely strings, the `backend` config in Terraform 0.9 uses proper types, including `bool` for the `encrypt` parameter. This PR updates `remote_state` to use a `bool` for `remote_state`.


brikis98:
Merging.

Issue #97, opened 2017-01-10
Run Terragrunt tests on CI system with Windows support


brikis98:
Currently, we only run our automated tests in CircleCI, which supports Linux. As a result, we have bugs in Windows, and many of our tests fail on Windows.

We should add the ability to run tests in a CI system that works with Windows. [AppVeyor](https://www.appveyor.com) may be a good choice, as they have a free plan for open source projects.

PR #184, opened 2017-04-24
Some adjustments to make config_helpers tests pass on Windows.

jocgir:
Change getTfVarsDir() to convert result to unix format path (forward ……slash).
Fix test for config_helpers that were not processed correctly on Windows.
Fix Makefile (install hook was not working)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ fmt:
 	gofmt -w $(GOFMT_FILES)
 
 install-pre-commit-hook:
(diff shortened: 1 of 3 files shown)


brikis98:
Ah, good catch. We have not had a chance to resolve #97 yet, so Windows support isn't being tested on a regular basis :(

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.12.8

PR #189, opened 2017-04-25
Fix problem with incorrect detection of change in the remote state configuration

jocgir:
On our setup (latest terragrunt and terraform 0.9.3), since the encrypt field has been changed to a bool, terragrunt always detect a difference between the terragrunt configuration and the terraform remote state configuration. 

`[terragrunt] 2017/04/25 16:10:53 Setting working directory to /var/folders/4.../terragrunt-...

[terragrunt]  WARNING: Terraform remote state is already configured for backend s3 with config map[bucket:bucket-name encrypt:true key:dev/us-east-1/keyfile lock_table:terraform_locks region:us-east-1], but your Terragrunt configuration specifies config map[encrypt:true bucket:bucket-name region:us-east-1 key:dev/us-east-1/keyfile lock_table:terraform_locks]. Overwrite? (y/n)`

As you can see, both configuration appears to be equal except the order of element, which is not a problem for DeepEqual. The problem is caused by the fact that the type of encrypt is different, bool vs string.

We don't have the problem if we cleanup the folder using --terragrunt-source-update.

To solve the problem, I just converted the field from string to bool before the DeepEqual comparison. There is probably a better solution, but I don't know why the encrypt field has been forced to be a bool initially.

### DIFF
--- a/remote/remote_state.go
+++ b/remote/remote_state.go
@@ -5,7 +5,9 @@ import (
 	"github.com/gruntwork-io/terragrunt/errors"
 	"github.com/gruntwork-io/terragrunt/options"
 	"github.com/gruntwork-io/terragrunt/shell"
+	"github.com/gruntwork-io/terragrunt/util"
 	"reflect"
+	"strconv"
 )
 
 // Configuration for Terraform remote state
@@ -96,6 +98,12 @@ func shouldOverrideExistingRemoteState(existingBackend *TerraformBackend, remote
 		return shell.PromptUserForYesNo(prompt, terragruntOptions)
 	}
 
+	if util.KindOf(existingBackend.Config["encrypt"]) == reflect.String && util.KindOf(remoteStateFromTerragruntConfig.Config["encrypt"]) == reflect.Bool {
+		// If encrypt in remoteStateFromTerragruntConfig is a bool and a string in existingBackend, DeepEqual will consider the maps to be different1.
+		// So we convert the value from string to bool to make them equivalent.
+		value, _ := strconv.ParseBool(existingBackend.Config["encrypt"].(string))
+		existingBackend.Config["encrypt"] = value
+	}
 	if !reflect.DeepEqual(existingBackend.Config, remoteStateFromTerragruntConfig.Config) {
 		prompt := fmt.Sprintf("WARNING: Terraform remote state is already configured for backend %s with config %v, but your Terragrunt configuration specifies config %v. Overwrite?", existingBackend.Type, existingBackend.Config, remoteStateFromTerragruntConfig.Config)
 		return shell.PromptUserForYesNo(prompt, terragruntOptions)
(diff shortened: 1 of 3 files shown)


remote/remote_state.go
```diff
@@ -96,6 +98,12 @@ func shouldOverrideExistingRemoteState(existingBackend *TerraformBackend, remote
 		return shell.PromptUserForYesNo(prompt, terragruntOptions)
 	}
 
+	if util.KindOf(existingBackend.Config["encrypt"]) == reflect.String && util.KindOf(remoteStateFromTerragruntConfig.Config["encrypt"]) == reflect.Bool {
+		// If encrypt in remoteStateFromTerragruntConfig is a bool and a string in existingBackend, DeepEqual will consider the maps to be different1.
+		// So we convert the value from string to bool to make them equivalent.
+		value, _ := strconv.ParseBool(existingBackend.Config["encrypt"].(string))
```
brikis98: Don't throw the error away

remote/remote_state.go
```diff
@@ -96,6 +98,12 @@ func shouldOverrideExistingRemoteState(existingBackend *TerraformBackend, remote
 		return shell.PromptUserForYesNo(prompt, terragruntOptions)
 	}
 
+	if util.KindOf(existingBackend.Config["encrypt"]) == reflect.String && util.KindOf(remoteStateFromTerragruntConfig.Config["encrypt"]) == reflect.Bool {
```
brikis98: Please add a comment along the following lines:

```
Terraform's `backend` configuration uses a boolean for the `encrypt` parameter. However, perhaps for backwards compatibility reasons, Terraform stores that parameter as a string in the `terraform.tfstate` file. Therefore, we have to convert it accordingly, or `DeepEqual` will fail.
```

remote/remote_state.go
```diff
@@ -96,6 +98,12 @@ func shouldOverrideExistingRemoteState(existingBackend *TerraformBackend, remote
...
+		// If encrypt in remoteStateFromTerragruntConfig is a bool and a string in existingBackend, DeepEqual will consider the maps to be different1.
```
brikis98: s/different1/different/

remote/remote_state.go
jocgir: I thew the error away because in that case, it will be catched by the normal code. My only goal was to replace the string by the bool.

brikis98: Not sure I follow. If for some reason the `encrypt` value was set to "garbage", then here, you'd end up replacing it with a boolean `false`, which is probably not the behavior we want.

remote/remote_state.go
jocgir: Done

remote/remote_state.go
jocgir: Done

remote/remote_state.go
jocgir: OK, you are right. I changed the code to add a message if `encrypt` contains invalid value.

brikis98:
Perfect, thank you again!

brikis98:
New release should show up here soon: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.12.9

PR #190, opened 2017-04-25
Fix how we execute Terraform commands so stdout isn't dirtied with unrelated output

brikis98:
When you run a single `terragrunt` command, under the hood, Terragrunt may execute several `terraform` commands. For example, when you run `terragrunt apply`, Terragrunt may run `terraform init`, `terraform get`, and finally, `terraform apply`. 

Only the original Terraform command should log to `stdout`. All other commands should log to `stderr`. We already have this logic in the `shell` package, but due to some dumb usage of cloning the `TerragruntOptions` object, we accidentally worked around this logic. This PR should fix that.

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -197,7 +197,7 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


brikis98:

Issue #80, opened 2016-12-19
apply-all should summarize any errors at the end of its run.


josh-padnick:
When you run `apply-all` on multiple modules, it's not obvious whether any modules had an error because Terragrunt itself won't exit with an error. Instead, you have to just scroll up to look for red text. It'd be better if Terragrunt reported its errors in a summary at the end like this:

```
[terragrunt] 2016/12/19 14:38:49 Terragrunt spin-up completed:

   5 Modules applied successfully [green text]
   1 Module failed to apply [red text]

Failed Modules
- ecs-cluster: Error downloading modules: module ecs_cluster: not found, may need to be downloaded using 'terraform get'
```

brikis98:
I think the `spin-up` command does show you the list of errors that happened, if any did. However, it does not have a nice summary message. That would be a great addition.

PR #195, opened 2017-04-26
Add plan-all option

jocgir:
I submitted this one few weeks ago, but it was refused because plan-all will fail on a stack with dependencies that has never been deployed.

As suggested, I added a mechanism to give user an explicit message in the case where the plan fail because its dependencies were not satisfied.

Also, as specified in #194, I relied on `RunModules` instead of `RunModulesReverseOrder` for both plan-all and output-all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -752,10 +752,11 @@ terragrunt output-all
 
 Finally, if you make some changes to your project, you could evaluate the impact by using `plan-all` command:
 
-Note: It is important to realize that you could get errors running `plan-all` if you have dependencies between your projects.
+Note: It is important to realize that you could get errors running `plan-all` if you have dependencies between your projects
+and some of those dependencies haven't been applied yet.
 
-_Ex: If you import a state file from one of your dependencies, `plan-all` will not generate or update that state file, resulting
-in possible errors in the dependant project._
+_Ex: If module A depends on module B and module B hasn't been applied yet, then plan-all will show the plan for B,
+but exit with an error when trying to show the plan for A._
 
 ```
 cd root
--- a/README.md
+++ b/README.md
@@ -693,7 +693,7 @@ to complete, and then run `terraform apply` in the next subfolder. How do you av
 process?
 
 
-#### The apply-all, destroy-all, and output-all commands
+#### The apply-all, destroy-all, output-all and plan-all commands
 
 To be able to deploy multiple Terraform modules in a single command, add a `terraform.tfvars` file to each module:
 
@@ -743,11 +743,23 @@ cd root
 terragrunt destroy-all
 ```
 
-Finally, to see the currently applied outputs of all of the subfolders, you can use the `output-all` command:
+To see the currently applied outputs of all of the subfolders, you can use the `output-all` command:
 
 ```
 cd root
 terragrunt output-all
+```
+
+Finally, if you make some changes to your project, you could evaluate the impact by using `plan-all` command:
+
+Note: It is important to realize that you could get errors running `plan-all` if you have dependencies between your projects.
+
+_Ex: If you import a state file from one of your dependencies, `plan-all` will not generate or update that state file, resulting
+in possible errors in the dependant project._
+
+```
+cd root
+terragrunt plan-all
 ```
 
 If your modules have dependencies between them—for example, you can't deploy the backend-app until MySQL and redis are
@@ -1184,7 +1196,7 @@ start with the prefix `--terragrunt-`. The currently available options are:
 * `--terragrunt-config`: A custom path to the `terraform.tfvars` file. May also be specified via the `TERRAGRUNT_CONFIG`
   environment variable. The default path is `terraform.tfvars` in the current directory (see
   [Terragrunt config files](#terragrunt-config-files) for a slightly more nuanced explanation). This argument is not
-  used with the `apply-all`, `destroy-all`, and `output-all` commands.
+  used with the `apply-all`, `destroy-all`, `output-all` and `plan-all` commands.
 
 * `--terragrunt-tfpath`: A custom path to the Terraform binary. May also be specified via the `TERRAGRUNT_TFPATH`
   environment variable. The default is `terraform` in a directory on your PATH.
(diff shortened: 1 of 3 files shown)


configstack/stack.go
```diff
@@ -25,6 +26,40 @@ func (stack *Stack) String() string {
...
+			if len(stack.Modules[n].Dependencies) > 0 && strings.Contains(output, ": Resource 'data.terraform_remote_state.") {
+				terragruntOptions.Logger.Printf(
+					"%v contains dependencies to %v and refers to remote state, "+
+						"you may have to apply your changes in the dependencies prior running terragrunt plan-all.\n",
+					stack.Modules[n].Path,
+					stack.Modules[n].Config.Dependencies.Paths,
+				)
+			}
+		}
+	}
```
brikis98: This for-loop should be moved into its own method called something like `summarizePlanAllErrors`.

configstack/stack.go
```diff
@@ -25,6 +26,40 @@ func (stack *Stack) String() string {
...
+	err := RunModules(stack.Modules)
+
+	// We inspect the error streams to give an explicit message if the plan failed because there were references to
+	// remote states. `terraform plan` will fail if it tries to access remote state from dependencies and the plan
+	// has never been applied on the dependency.
+	for n, o := range errorStreams {
+		output := o.String()
+		if strings.Contains(output, "Error running plan:") {
+			terragruntOptions.Logger.Println(output)
+			if len(stack.Modules[n].Dependencies) > 0 && strings.Contains(output, ": Resource 'data.terraform_remote_state.") {
```
brikis98: It's easy to forget to include a module in the `Dependencies` list, so perhaps that check is superfluous, and searching for the `data.terraform_remote_state` error is enough.

configstack/stack.go
```diff
@@ -25,6 +26,40 @@ func (stack *Stack) String() string {
...
+		module.TerragruntOptions.ErrWriter = &errorStreams[n]
+	}
+
+	// Start the plan on each module
+	err := RunModules(stack.Modules)
+
+	// We inspect the error streams to give an explicit message if the plan failed because there were references to
+	// remote states. `terraform plan` will fail if it tries to access remote state from dependencies and the plan
+	// has never been applied on the dependency.
+	for n, o := range errorStreams {
```
brikis98: `i` is a more common name for an index variable. `errorStream` is probably an easier to understand name than `o`.

configstack/stack.go
```diff
@@ -25,6 +26,40 @@ func (stack *Stack) String() string {
...
+		output := o.String()
+		if strings.Contains(output, "Error running plan:") {
+			terragruntOptions.Logger.Println(output)
```
brikis98: Won't this print out the entire plan error message again?

configstack/stack.go
jocgir: Yes, your are right, I have tendency to use small variables and rename them later.
Fixed

configstack/stack.go
jocgir: OK

jocgir:
The goal of plan-all is mainly to evaluate the impact of code modification on an already applied infrastructure. When we submit a PR on our infrastructure, our CI apply a plan-all to let us see the impact and we have to confirm if we wish to apply them.

It can still work on new infrastructure, but not if there are true dependencies.

configstack/stack.go
jocgir: No, the error message is not printed initially. Is is only printed there.

But some messages are

jocgir:
Regarding #80, I agree with you. The xxx-all outputs are difficult to read mainly due to the multithreading. Capturing the output (stdout and stderr) for each and summarize at the end would give a more readable output.

Also, there is a problem with the exit code. We should recuperate exit code (err) for each module and summarize them with the output.

I especially have a problem with `terraform plan -detailed-exitcode`. It returns 2 if there are modification, but terragrunt exit with 1 because it is not able to parse the error summary.

`[terragrunt] 2017/04/26 19:07:07 Unable to determine underlying exit code, so Terragrunt will exit with error code 1`

Also, I don't know if it is a bug in terraform, but for some specific cases, when we use 

```
<= module.domain.data.aws_route53_zone.main
    caller_reference:          "<computed>"
    comment:                   "<computed>"
    name:                      "dev.xxx.com."
    private_zone:              "true"
    resource_record_set_count: "<computed>"
    tags.%:                    "<computed>"
    vpc_id:                    "<computed>"
    zone_id:                   "<computed>"

Plan: 0 to add, 0 to change, 0 to destroy.
[terragrunt] [/terragrunt/infra] 2017/04/26 19:44:10 Module /terragrunt/infra has finished with an error: exit status 2```

As you can see, we receive `Plan: 0 to add, 0 to change, 0 to destroy.`, but terraform still exit with status 2.

File: README.md [hunk 5]
```diff
@@ -743,13 +743,25 @@ cd root
...
+To see the currently applied outputs of all of the subfolders, you can use the `output-all` command:
 
 ```
 cd root
 terragrunt output-all
 ```
 
+Finally, if you make some changes to your project, you could evaluate the impact by using `plan-all` command:
+
+Note: It is important to realize that you could get errors running `plan-all` if you have dependencies between your projects.
```
brikis98: s/if you have dependencies between your projects/if you have dependencies between your projects and some of those dependencies haven't been applied yet/

README.md
```diff
@@ -743,13 +743,25 @@ cd root
...
+
+_Ex: If you import a state file from one of your dependencies, `plan-all` will not generate or update that state file, resulting
+in possible errors in the dependant project._
```
brikis98: I'd rephrase the example to: "If module A depends on module B and module B hasn't been applied yet, then plan-all will show the plan for B, but exit with an error when trying to show the plan for A"

brikis98:
> I especially have a problem with terraform plan -detailed-exitcode. It returns 2 if there are modifications, but terragrunt exit with 1 because it is not able to parse the error summary.

Please file a bug for that. AFAIK, that used to work, so we may have a regression that broke that functionality.

> As you can see, we receive Plan: 0 to add, 0 to change, 0 to destroy., but terraform still exit with status 2.

Never seen that before. Is there a simple case to repro that with? Does it only happen with Terragrunt or does the same problem happen when using Terraform directly?

jocgir:
The fact that exit code 2 is given when it should not is a terraform problem. But terragrunt still has problem reporting it and convert the exit code 2 (changes) to exit code 1 because of this:
```go
type MultiError struct {
	Errors []error
}

func (err MultiError) Error() string {
	errorStrings := []string{}
	for _, err := range err.Errors {
		errorStrings = append(errorStrings, err.Error())
	}
	return fmt.Sprintf("Encountered the following errors:\n%s", strings.Join(errorStrings, "\n"))
}
```

The extra formatting make the error difficult to analyse at the end of the program.

README.md
jocgir: Done

README.md
jocgir: Done

brikis98:
> The fact that exit code 2 is given when it should not is a terraform problem. But terragrunt still has problem reporting it and convert the exit code 2 (changes) to exit code 1 because of this:

Ah, I see. Worth filing a separate bug to track it. Not clear what exit code to use with multiple `plan` commands though, but better logging would definitely be a good idea.

brikis98:
OK, LGTM. Merging. Thank you!

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.12.13

PR #197, opened 2017-04-27
Bug/exit status on modules

jocgir:
Here is a fix to better handle the exit code when using xxx-all command. I simply added an interface to get the exit code and implemented it in `DependencyFinishedWithError` and `MultiError`.

That does not solve totally the problem with `terraform plan -detailed-exitcode` since if terraform detects changes on dependencies, terragrunt will simply not process the dependants. But the messages are easy to understand and the final result code is the good one.

### DIFF
--- a/configstack/running_module.go
+++ b/configstack/running_module.go
@@ -3,6 +3,7 @@ package configstack
 import (
 	"fmt"
 	"github.com/gruntwork-io/terragrunt/errors"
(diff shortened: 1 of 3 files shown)


brikis98:
Thank you!

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.12.14

Issue #198, opened 2017-04-28
plan-all swallows error when it can't acquire a lock


mrtyler:
I don't have a canned minimal repro case but hopefully it's simple enough: in two separate checkouts of a terraform project with locking enabled, run `terragrunt plan-all` simultaneously.

Here's what happens to the `plan-all` that loses the race:
```
[terragrunt] [/Users/tyler/demo/gpii-terraform-live/stg/base] 2017/04/27 19:21:56 Running command: terraform plan
Acquiring state lock. This may take a few moments...
[terragrunt] [/Users/tyler/demo/gpii-terraform-live/stg/base] 2017/04/27 19:22:00 Module /Users/tyler/demo/gpii-terraform-live/stg/base has finished with an error: exit status 1
[terragrunt] [/Users/tyler/demo/gpii-terraform-live/stg/worker] 2017/04/27 19:22:00 Dependency /Users/tyler/demo/gpii-terraform-live/stg/base of module /Users/tyler/demo/gpii-terraform-live/stg/worker just finished with an error. Module /Users/tyler/demo/gpii-terraform-live/stg/worker will have to return an error too.
```
`exit status 1` but no explanation of what went wrong.

`terragrunt plan` produces a clear error message in the same situation:
```
[terragrunt] 2017/04/27 19:26:10 Running command: terraform plan
Acquiring state lock. This may take a few moments...
Error locking state: Error acquiring the state lock: ConditionalCheckFailedException: The conditional request failed
        status code: 400, request id: G4UCJHJN43DF8LDE7GJ8EIS4IRVV4KQNSO5AEMVJF66Q9ASUAAJG
Lock Info:
  ID:        00e440d2-608d-a5bd-fa88-5d1667292200
  Path:      gpii-terraform-state/stg/base/terraform.tfstate
  Operation: OperationTypePlan
  Who:       tyler@toaster.local
  Version:   0.9.4
  Created:   2017-04-28 01:26:13.442812785 +0000 UTC
  Info:
```

`apply-all` does as well:
```
[terragrunt] [/Users/tyler/demo/gpii-terraform-live/stg/base] 2017/04/27 19:28:21 Running command: terraform apply -input=false
Acquiring state lock. This may take a few moments...
Error locking state: Error acquiring the state lock: ConditionalCheckFailedException: The conditional request failed
[...]

PR #199, opened 2017-04-28
Improve the global behaviour of -all commands

jocgir:
The output of -all command could be very difficult to read. I modified several files to implement a new mechanism and improve the global readability.

I think that this implementation could help for #80. However, I did not implemented a summary for apply-all and destroy-all. It works very well with output-all and plan-all. And the same technique could easily be applied to summarize apply-all and destroy-all.

It works especially well when we redirect the stderr to /dev/null. The output of the distinct terraform command are then grouped together and easier to associate with the source module.

Since the addition of the new remote state, a lot of output related to the terraform init are displayed. I changed the behaviour to only display this output if there is a problem.

I also refactored the newly added plan-all to implement a summary at the end and also support correctly the -detailed-exitcode option from terraform.

Finally, this new implementation should solve the problem #198.

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -233,7 +233,7 @@ func downloadModules(terragruntOptions *options.TerragruntOptions) error {
 			return err
 		}
 		if shouldDownload {
-			return shell.RunTerraformCommand(terragruntOptions, "get", "-update")
+			return shell.RunTerraformCommandAndIgnoreOutput(terragruntOptions, "get", "-update")
 		}
 	}
 
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -233,7 +233,7 @@ func downloadModules(terragruntOptions *options.TerragruntOptions) error {
 			return err
 		}
 		if shouldDownload {
-			return shell.RunTerraformCommandAndIgnoreOutput(terragruntOptions, "get", "-update")
+			return shell.RunTerraformCommandAndRedirectOutputToLogger(terragruntOptions, "get", "-update")
 		}
 	}
 
--- a/configstack/running_module.go
+++ b/configstack/running_module.go
@@ -257,7 +257,9 @@ func (this MultiError) ExitStatus() (int, error) {
 	for i := range this.Errors {
 		if code, err := shell.GetExitCode(this.Errors[i]); err != nil {
 			return -1, this
-		} else if code > exitCode {
+		} else if code == 1 || code == 2 && exitCode == 0 {
+			// The exit code 1 is more significant that the exit code 2 because it represents an error
+			// while 2 represent a warning.
 			exitCode = code
 		}
 	}
--- a/configstack/running_module.go
+++ b/configstack/running_module.go
@@ -1,9 +1,11 @@
 package configstack
 
 import (
+	"bytes"
 	"fmt"
 	"github.com/gruntwork-io/terragrunt/errors"
 	"github.com/gruntwork-io/terragrunt/shell"
+	"io"
 	"strings"
 	"sync"
 )
@@ -25,6 +27,9 @@ type runningModule struct {
 	DependencyDone chan *runningModule
 	Dependencies   map[string]*runningModule
 	NotifyWhenDone []*runningModule
+	OutStream      bytes.Buffer
+	Writer         io.Writer
+	Handler        ModuleHandler
 }
 
 // This controls in what order dependencies should be enforced between modules
@@ -34,6 +39,12 @@ const (
 	NormalOrder DependencyOrder = iota
 	ReverseOrder
 )
+
+// Handler function prototype to inject interaction during the processing.
+// The function receive the current module, its output and its error in parameter.
+// Normally, the handler should return the same error as received in parameter, but it is possible to
+// alter the normal course of the proccess by changing the error result.
+type ModuleHandler func(TerraformModule, string, error) error
 
 // Create a new RunningModule struct for the given module. This will initialize all fields to reasonable defaults,
 // except for the Dependencies and NotifyWhenDone, both of which will be empty. You should fill these using a
@@ -45,6 +56,7 @@ func newRunningModule(module *TerraformModule) *runningModule {
 		DependencyDone: make(chan *runningModule, 1000), // Use a huge buffer to ensure senders are never blocked
 		Dependencies:   map[string]*runningModule{},
 		NotifyWhenDone: []*runningModule{},
+		Writer:         module.TerragruntOptions.Writer,
 	}
 }
 
@@ -52,22 +64,43 @@ func newRunningModule(module *TerraformModule) *runningModule {
 // TerragruntOptions object. The modules will be executed in an order determined by their inter-dependencies, using
 // as much concurrency as possible.
 func RunModules(modules []*TerraformModule) error {
-	runningModules, err := toRunningModules(modules, NormalOrder)
-	if err != nil {
-		return err
-	}
-	return runModules(runningModules)
+	return RunModulesWithHandler(modules, nil, NormalOrder)
 }
 
 // Run the given map of module path to runningModule. To "run" a module, execute the RunTerragrunt command in its
 // TerragruntOptions object. The modules will be executed in the reverse order of their inter-dependencies, using
 // as much concurrency as possible.
 func RunModulesReverseOrder(modules []*TerraformModule) error {
-	runningModules, err := toRunningModules(modules, ReverseOrder)
+	return RunModulesWithHandler(modules, nil, ReverseOrder)
+}
+
+// Run the given map of module path to runningModule. To "run" a module, execute the RunTerragrunt command in its
+// TerragruntOptions object. The modules will be executed in an order determined by their inter-dependencies, using
+// as much concurrency as possible.
+// This version accepts a function as parameter (see: ModuleHander). The handler is called when the command is
+// completed (either succeeded or failed).
+func RunModulesWithHandler(modules []*TerraformModule, handler ModuleHandler, order DependencyOrder) error {
+	runningModules, err := toRunningModules(modules, order)
 	if err != nil {
 		return err
 	}
-	return runModules(runningModules)
+
+	var waitGroup sync.WaitGroup
+
+	for _, module := range runningModules {
+		waitGroup.Add(1)
+		module.Handler = handler
+		go func(module *runningModule) {
+			defer waitGroup.Done()
+			module.Module.TerragruntOptions.Writer = &module.OutStream
+			module.Module.TerragruntOptions.ErrWriter = &module.OutStream
+			module.runModuleWhenReady()
+		}(module)
+	}
+
+	waitGroup.Wait()
+
+	return collectErrors(runningModules)
 }
 
 // Convert the list of modules to a map from module path to a runningModule struct. This struct contains information
@@ -105,25 +138,6 @@ func crossLinkDependencies(modules map[string]*runningModule, dependencyOrder De
 	}
 
 	return modules, nil
-}
-
-// Run the given map of module path to runningModule. To "run" a module, execute the RunTerragrunt command in its
-// TerragruntOptions object. The modules will be executed in an order determined by their inter-dependencies, using
-// as much concurrency as possible.
-func runModules(modules map[string]*runningModule) error {
-	var waitGroup sync.WaitGroup
-
-	for _, module := range modules {
-		waitGroup.Add(1)
-		go func(module *runningModule) {
-			defer waitGroup.Done()
-			module.runModuleWhenReady()
-		}(module)
-	}
-
-	waitGroup.Wait()
-
-	return collectErrors(modules)
 }
 
 // Collect the errors from the given modules and return a single error object to represent them, or nil if no errors
@@ -186,11 +200,18 @@ func (module *runningModule) runNow() error {
 
 // Record that a module has finished executing and notify all of this module's dependencies
 func (module *runningModule) moduleFinished(moduleErr error) {
-	if moduleErr == nil {
-		module.Module.TerragruntOptions.Logger.Printf("Module %s has finished successfully!", module.Module.Path)
-	} else {
-		module.Module.TerragruntOptions.Logger.Printf("Module %s has finished with an error: %v", module.Module.Path, moduleErr)
+	status := "successfully!"
+	if moduleErr != nil {
+		status = fmt.Sprintf("with an error: %v", moduleErr)
 	}
+
+	separator := strings.Repeat("-", 132)
+	module.Module.TerragruntOptions.Logger.Printf("Module %s has finished %s", module.Module.Path, status)
+
+	if module.Handler != nil {
+		moduleErr = module.Handler(*module.Module, module.OutStream.String(), moduleErr)
+	}
+	fmt.Fprintf(module.Writer, "\n%s\n%v\n\n%v%s\n\n", separator, module.Module.Path, module.OutStream.String(), separator)
 
 	module.Status = Finished
 	module.Err = moduleErr
--- a/configstack/running_module.go
+++ b/configstack/running_module.go
@@ -13,11 +13,21 @@ import (
 // Represents the status of a module that we are trying to apply as part of the apply-all or destroy-all command
 type ModuleStatus int
 
+const NORMAL_EXIT_CODE = 0
+const ERROR_EXIT_CODE = 1
+const UNDEFINED_EXIT_CODE = -1
+
 const (
 	Waiting ModuleStatus = iota
 	Running
 	Finished
 )
+
+// Using CreateMultiErrors as a variable instead of a function allows us to override the function used to compose multi error object.
+// It is used if a command wants to change the default behaviour of the severity analysis that is implemented by default.
+var CreateMultiErrors = func(errs []error) error {
+	return MultiError{Errors: errs}
+}
 
 // Represents a module we are trying to "run" (i.e. apply or destroy) as part of the apply-all or destroy-all command
 type runningModule struct {
@@ -157,7 +167,7 @@ func collectErrors(modules map[string]*runningModule) error {
 	if len(errs) == 0 {
 		return nil
 	} else {
-		return errors.WithStackTrace(MultiError{Errors: errs})
+		return errors.WithStackTrace(CreateMultiErrors(errs))
 	}
 }
 
@@ -260,13 +270,11 @@ func (err MultiError) Error() string {
 }
 
 func (this MultiError) ExitStatus() (int, error) {
-	exitCode := 0
+	exitCode := NORMAL_EXIT_CODE
 	for i := range this.Errors {
 		if code, err := shell.GetExitCode(this.Errors[i]); err != nil {
-			return -1, this
-		} else if code == 1 || code == 2 && exitCode == 0 {
-			// The exit code 1 is more significant that the exit code 2 because it represents an error
-			// while 2 represent a warning.
+			return UNDEFINED_EXIT_CODE, this
+		} else if code > exitCode {
 			exitCode = code
 		}
 	}
--- a/configstack/running_module.go
+++ b/configstack/running_module.go
@@ -30,6 +30,7 @@ type runningModule struct {
 	OutStream      bytes.Buffer
 	Writer         io.Writer
 	Handler        ModuleHandler
+	Mutex          *sync.Mutex // A shared mutex pointer to ensure that there is no concurrency problem when job finish and report
 }
 
 // This controls in what order dependencies should be enforced between modules
@@ -49,7 +50,7 @@ type ModuleHandler func(TerraformModule, string, error) error
 // Create a new RunningModule struct for the given module. This will initialize all fields to reasonable defaults,
 // except for the Dependencies and NotifyWhenDone, both of which will be empty. You should fill these using a
 // function such as crossLinkDependencies.
-func newRunningModule(module *TerraformModule) *runningModule {
+func newRunningModule(module *TerraformModule, mutex *sync.Mutex) *runningModule {
 	return &runningModule{
 		Module:         module,
 		Status:         Waiting,
@@ -57,6 +58,7 @@ func newRunningModule(module *TerraformModule) *runningModule {
 		Dependencies:   map[string]*runningModule{},
 		NotifyWhenDone: []*runningModule{},
 		Writer:         module.TerragruntOptions.Writer,
+		Mutex:          mutex,
 	}
 }
 
@@ -107,9 +109,11 @@ func RunModulesWithHandler(modules []*TerraformModule, handler ModuleHandler, or
 // about executing the module, such as whether it has finished running or not and any errors that happened. Note that
 // this does NOT actually run the module. For that, see the RunModules method.
 func toRunningModules(modules []*TerraformModule, dependencyOrder DependencyOrder) (map[string]*runningModule, error) {
+	var mutex sync.Mutex
+
 	runningModules := map[string]*runningModule{}
 	for _, module := range modules {
-		runningModules[module.Path] = newRunningModule(module)
+		runningModules[module.Path] = newRunningModule(module, &mutex)
 	}
 
 	return crossLinkDependencies(runningModules, dependencyOrder)
@@ -198,6 +202,8 @@ func (module *runningModule) runNow() error {
 	}
 }
 
+var separator = strings.Repeat("-", 132)
+
 // Record that a module has finished executing and notify all of this module's dependencies
 func (module *runningModule) moduleFinished(moduleErr error) {
 	status := "successfully!"
@@ -205,13 +211,14 @@ func (module *runningModule) moduleFinished(moduleErr error) {
 		status = fmt.Sprintf("with an error: %v", moduleErr)
 	}
 
-	separator := strings.Repeat("-", 132)
+	module.Mutex.Lock()
+	defer module.Mutex.Unlock()
 	module.Module.TerragruntOptions.Logger.Printf("Module %s has finished %s", module.Module.Path, status)
 
 	if module.Handler != nil {
 		moduleErr = module.Handler(*module.Module, module.OutStream.String(), moduleErr)
 	}
-	fmt.Fprintf(module.Writer, "\n%s\n%v\n\n%v%s\n\n", separator, module.Module.Path, module.OutStream.String(), separator)
+	fmt.Fprintf(module.Writer, "%s\n%v\n\n%v\n", separator, module.Module.Path, module.OutStream.String())
 
 	module.Status = Finished
 	module.Err = moduleErr
--- a/configstack/stack_plan.go
+++ b/configstack/stack_plan.go
@@ -141,9 +141,9 @@ func (err countError) Error() string {
 // If there are changes, the exit status must be = 2
 func (err countError) ExitStatus() (int, error) {
 	if err.count > 0 {
-		return 2, nil
+		return CHANGE_EXIT_CODE, nil
 	}
-	return 0, nil
+	return NORMAL_EXIT_CODE, nil
 }
 
 // This is a specialized version of MultiError type
--- a/configstack/stack_plan.go
+++ b/configstack/stack_plan.go
@@ -8,7 +8,6 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
-	"sync"
 )
 
 // The returned information for each module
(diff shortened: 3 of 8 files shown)


configstack/running_module.go
```diff
@@ -184,13 +202,23 @@ func (module *runningModule) runNow() error {
...
-		module.Module.TerragruntOptions.Logger.Printf("Module %s has finished successfully!", module.Module.Path)
-	} else {
-		module.Module.TerragruntOptions.Logger.Printf("Module %s has finished with an error: %v", module.Module.Path, moduleErr)
+	status := "successfully!"
+	if moduleErr != nil {
+		status = fmt.Sprintf("with an error: %v", moduleErr)
+	}
+
+	module.Mutex.Lock()
+	defer module.Mutex.Unlock()
```
brikis98: Why do you need a lock here? I'd prefer to avoid locks as much as possible and use solely goroutines and channels.

configstack/running_module.go
```diff
@@ -236,7 +264,9 @@ func (this MultiError) ExitStatus() (int, error) {
 	for i := range this.Errors {
 		if code, err := shell.GetExitCode(this.Errors[i]); err != nil {
 			return -1, this
-		} else if code > exitCode {
+		} else if code == 1 || code == 2 && exitCode == 0 {
+			// The exit code 1 is more significant that the exit code 2 because it represents an error
+			// while 2 represent a warning.
```
brikis98: Doesn't this only apply to the `plan` command with the `detailed-exit-code` param? Seems like we should NOT apply that logic to all other commands...

shell/run_shell_cmd.go
```diff
@@ -57,10 +68,13 @@ func RunShellCommand(terragruntOptions *options.TerragruntOptions, command strin
 
 // Run the specified shell command with the specified arguments. Capture the command's stdout and return it as a
 // string.
-func RunShellCommandAndCaptureOutput(terragruntOptions *options.TerragruntOptions, command string, args ...string) (string, error) {
+func runShellCommandAndCaptureOutput(terragruntOptions *options.TerragruntOptions, copyWorkingDir bool, command string, args ...string) (string, error) {
 	stdout := new(bytes.Buffer)
 
 	terragruntOptionsCopy := terragruntOptions.Clone(terragruntOptions.TerragruntConfigPath)
+	if copyWorkingDir {
+		terragruntOptionsCopy.WorkingDir = terragruntOptions.WorkingDir
```
brikis98: Isn't the `WorkingDir` copied by default as part of `Clone`?

cli/cli_app.go
```diff
@@ -233,7 +233,7 @@ func downloadModules(terragruntOptions *options.TerragruntOptions) error {
 			return err
 		}
 		if shouldDownload {
-			return shell.RunTerraformCommand(terragruntOptions, "get", "-update")
+			return shell.RunTerraformCommandAndIgnoreOutput(terragruntOptions, "get", "-update")
```
brikis98: I don't think throwing the output away is a good approach. This output isn't useful during normal runs, but it's very useful for debugging issues.

The proper approach here is to use different levels of logging. We set the default debug level to INFO and output like this would be logged at the DEBUG level so that, normally, it doesn't show up. However, you could set the debug level to DEBUG or TRACE via config or a command-line flag to see more to help with debugging. 

Something like [go-logging](https://github.com/op/go-logging) would probably make this straightforward.

configstack/stack_plan.go
```diff
@@ -0,0 +1,135 @@
...
+
+	return err
+}
+
+// Returns the handler that will be executed after each completion of `terraform plan`
+func getResultHandler(detailedExitCode bool, results *[]moduleResult) ModuleHandler {
+	return func(module TerraformModule, output string, err error) error {
+		inspectPlanResult(module, output)
+		if exitCode, convErr := shell.GetExitCode(err); convErr == nil && detailedExitCode && exitCode == 2 {
+			err = nil
```
brikis98: Not sure I follow the logic here of setting `err` to `nil`. At the very least add a comment.

configstack/stack_plan.go
```diff
@@ -0,0 +1,135 @@
...
+		if result.Err != nil {
+			errMsg = fmt.Sprintf(", Error: %v", result.Err)
+		}
+
+		fmt.Fprintf(terragruntOptions.Writer, "    %v : %v%v\n", result.Module.Path, result.Message, errMsg)
+	}
+}
+
+// Check the output message
+func inspectPlanResult(module TerraformModule, output string) {
```
brikis98: The name `inspect` is a bit vague. This method seems to log warnings about unapplied dependencies, so perhaps rename it to `warnAboutMissingDependencies`.

configstack/stack_plan.go
```diff
@@ -0,0 +1,135 @@
...
+			"you may have to apply your changes in the dependencies prior running terragrunt plan-all.\n",
+			module.Path,
+			dependenciesMsg,
+		)
+	}
+}
+
+// Parse the output message to extract a summary
+func extractSummaryResultFromPlan(output string) (string, int) {
+	const noChange = "No changes. Infrastructure is up-to-date."
```
brikis98: I'm a bit nervous about parsing the `plan` output string, as it is very likely to change in future releases. Is it possible to get this from the exit code or some other more stable API?

configstack/stack_plan.go
```diff
@@ -0,0 +1,135 @@
...
+	if strings.Contains(output, noChange) {
+		return "No change", 0
+	}
+
+	re := regexp.MustCompile(`(\d+) to add, (\d+) to change, (\d+) to destroy.`)
```
brikis98: These sorts of regexes should be defined at the top and not inline to avoid a potential panic in the middle of running code.

configstack/stack_plan.go
```diff
@@ -0,0 +1,135 @@
...
+		message, count := extractSummaryResultFromPlan(output)
+
+		// We add the result to the result list (there is no concurrency problem because it is handled by the running_module)
+		*results = append(*results, moduleResult{module, err, message, count})
+
+		return err
+	}
+}
+
+// Print a little summary of the plan execution
```
brikis98: What does the summary look like?

configstack/stack_plan.go
```diff
@@ -0,0 +1,135 @@
...
+	// Sometimes, terraform returns 0 add, 0 change and 0 destroy. We return a more explicit message
+	return "No effective change", 0
+}
+
+// This is used to return the total number of changes if -detail-exitcode is specified and return an exit code of 2
+// to be compliant with the terraform specification.
+type countError struct{ count int }
+
+func (err countError) Error() string {
+	article, s := "is", ""
```
brikis98: What's `s`? Single-character variable name gives no hint as to its purpose. I'm guessing it's some sort of plural suffix, but a better name would make that clearer.

configstack/stack_plan.go
```diff
@@ -0,0 +1,135 @@
...
+	if err.count > 1 {
+		article, s = "are", "s"
+	}
+	return fmt.Sprintf("There %s %v change%s to apply", article, err.count, s)
+}
+
+// If there are changes, the exit status must be = 2
+func (err countError) ExitStatus() (int, error) {
+	if err.count > 0 {
+		return 2, nil
```
brikis98: The exit code "magic numbers" (0, 1, 2) are copy/pasted all over the place. Define named constants for them to make the code easier to follow.

configstack/running_module.go
jocgir: The objective of the lock is to avoid mix-up of outputs in the case where multiple terraform commands finish at the same time.

configstack/running_module.go
jocgir: Yes you are right. However, it is an important behaviour for the plan command. I should find a way to implement this without disturbing potential other command.

I didn't see other terraform command that have exit status other that 0 and 1.

shell/run_shell_cmd.go
jocgir: No, that is why I had to make this change.

```go
func (terragruntOptions *TerragruntOptions) Clone(terragruntConfigPath string) *TerragruntOptions {
	workingDir := filepath.Dir(terragruntConfigPath)

	return &TerragruntOptions{
		TerragruntConfigPath: terragruntConfigPath,
		TerraformPath:        terragruntOptions.TerraformPath,
		NonInteractive:       terragruntOptions.NonInteractive,
		TerraformCliArgs:     terragruntOptions.TerraformCliArgs,
		WorkingDir:           workingDir,
		Logger:               util.CreateLogger(workingDir),
		Env:                  terragruntOptions.Env,
		Source:               terragruntOptions.Source,
		SourceUpdate:         terragruntOptions.SourceUpdate,
		RunTerragrunt:        terragruntOptions.RunTerragrunt,
		Writer:               terragruntOptions.Writer,
		ErrWriter:            terragruntOptions.ErrWriter,
	}
}
```

cli/cli_app.go
jocgir: I agree with you. But the goal is not only limit the quantity of output, but also not "pollute" the output of the effective target command with artefacts that are required by terragrunt.

i.e. If i do `terragrunt plan`, I am interested in the output of `terraform plan`. Any other command that are issued by `terragrunt` like `terraform init`, `terraform get` are only required to achieve the `terraform plan` so their output should not be mixed with the output of the plan and should be disabled if required.

configstack/stack_plan.go
jocgir: I will add a comment.

The logic is simply that we do not want to consider exit code 2 as an error and not execute the dependants because there is an "error" in the dependencies. 2 is not an error in this case, it is simply a status. We reintroduce the exit code at the very end to mimic the behaviour of the `terrafrom plan -detailed-exitcode` to exit with 2 if there are changes in any of the module in the stack.

configstack/stack_plan.go
jocgir: OK

configstack/stack_plan.go
jocgir: ok

configstack/stack_plan.go
jocgir: The exit code gives information only if we specify the -detailed-exitcode. I may force it and rely on the code. Otherwise, I am not aware of an API that could help up on that point.

configstack/stack_plan.go
jocgir: OK

shell/run_shell_cmd.go
```diff
@@ -57,10 +68,13 @@ func RunShellCommand(terragruntOptions *options.TerragruntOptions, command strin
...
+	}
 	terragruntOptionsCopy.Writer = stdout
 	terragruntOptionsCopy.ErrWriter = stdout
```
jocgir: @brikis98 I noticed that the following "unmodified" code has the effect of ignoring the output in case of error, which is bad. I get error on Windows with the init and the get command in the case where I have modules in my module. I think it is a terragrunt bug on Windows that I will check later. But the effect was that I didn't get the error using terragrunt and I though that it was caused by my modifications, but it appears that the problem was there before.

configstack/stack_plan.go
jocgir:
```
------------------------------------------------------------------------------------------------------------------------------------
Summary:
    /Source/Infra/infrastructure-module/terragrunt : No change
    /Source/Infra/infrastructure-module/terragrunt/common : No change
    /Source/Infra/infrastructure-module/terragrunt/deploy_base : No change
    /Source/Infra/infrastructure-module/terragrunt/infra : No effective change
    /Source/Infra/infrastructure-module/terragrunt/containers_cluster : 1 to add, 1 to change, 1 to destroy.
```

configstack/stack_plan.go
jocgir: Your gess is good. It was simply s as in "s" :-), I renamed it plural. As I already said, I'am little cryptic by nature

cli/cli_app.go
jocgir: Implementing a different logging mechanism is beyond this PR. I will make changes to always keep the ouput, but isolate the significant output from the artefacts.

jocgir:
@brikis98 I had to merge #201 in this PR because I was not able to test it. Normally I develop on osx, but this weekend, I am on Windows. And also, the whole redirect thing was not working properly with the current implementation of `runShellCommandAndCaptureOutput()`. It caused a lost of error output when there are errors on redirected commands.

jocgir:
Example output of plan-all (stdout only):
```
------------------------------------------------------------------------------------------------------------------------------------
C:/Source/Infra/infrastructure-module/terragrunt

Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

No changes. Infrastructure is up-to-date.

This means that Terraform did not detect any differences between your
configuration and real physical resources that exist. As a result, Terraform
doesn't need to do anything.

------------------------------------------------------------------------------------------------------------------------------------
C:/Source/Infra/infrastructure-module/terragrunt/common

Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

data.template_file.prefix: Refreshing state...
data.template_file.replication_prefix: Refreshing state...
data.aws_caller_identity.current: Refreshing state...
No changes. Infrastructure is up-to-date.

This means that Terraform did not detect any differences between your
configuration and real physical resources that exist. As a result, Terraform
doesn't need to do anything.

------------------------------------------------------------------------------------------------------------------------------------
C:/Source/Infra/infrastructure-module/terragrunt/deploy_tf

Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

data.terraform_remote_state.common: Refreshing state...
data.aws_iam_policy_document.deploy: Refreshing state...
aws_iam_policy.deploy: Refreshing state... (ID: arn:aws:iam::0000000000:policy/xdev-deploy)
aws_iam_user.deploy: Refreshing state... (ID: xdev-deploy)
aws_iam_access_key.deploy: Refreshing state... (ID: AKIAJCXXXXXXXXXXXXXXX)
aws_iam_policy_attachment.deploy: Refreshing state... (ID: xdev-deploy)
No changes. Infrastructure is up-to-date.

This means that Terraform did not detect any differences between your
configuration and real physical resources that exist. As a result, Terraform
doesn't need to do anything.

------------------------------------------------------------------------------------------------------------------------------------
C:/Source/Infra/infrastructure-module/terragrunt/infra

Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

data.terraform_remote_state.common: Refreshing state...
tls_private_key.keys: Refreshing state... (ID: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx)
aws_eip.gw_eip: Refreshing state... (ID: eipalloc-xxxxxxx
aws_customer_gateway.vpn_gw: Refreshing state... (ID: cgw-xxxxxxxx)
aws_vpc_dhcp_options.dhcp: Refreshing state... (ID: dopt-xxxxxxxxxx)
aws_vpc.vpc: Refreshing state... (ID: vpc-xxxxxxx)
...

Note: You didn't specify an "-out" parameter to save this plan, so when
"apply" is called, Terraform can't guarantee this is what will execute.

<= module.domain.data.aws_route53_zone.main
    caller_reference:          "<computed>"
    comment:                   "<computed>"
    name:                      "xxxx.cloud.xxxxxx.org."
    private_zone:              "true"
    resource_record_set_count: "<computed>"
    tags.%:                    "<computed>"
    vpc_id:                    "<computed>"
    zone_id:                   "<computed>"


Plan: 0 to add, 0 to change, 0 to destroy.

------------------------------------------------------------------------------------------------------------------------------------
C:/Source/Infra/infrastructure-module/terragrunt/containers_cluster

Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

data.terraform_remote_state.common: Refreshing state...
data.terraform_remote_state.common: Refreshing state...
data.terraform_remote_state.infra: Refreshing state...
aws_ecs_cluster.cluster: Refreshing state... (ID: arn:aws:ecs:us-east-1:00000000000:cluster/dev-ECS_Cluster)
data.aws_ami.windows_container: Refreshing state...
data.template_file.user_data: Refreshing state...
aws_security_group.node: Refreshing state... (ID: sg-aaaaaaaa)
aws_launch_configuration.config: Refreshing state... (ID: ECS-Nodes-Windows-aaaaaaaaaaaaa)
aws_autoscaling_group.asg: Refreshing state... (ID: Cluster_Beta)
The Terraform execution plan has been generated and is shown below.
Resources are shown in alphabetical order for quick scanning. Green resources
will be created (or destroyed and then created if an existing resource
exists), yellow resources are being changed in-place, and red resources
will be destroyed. Cyan entries are data sources to be read.

Note: You didn't specify an "-out" parameter to save this plan, so when
"apply" is called, Terraform can't guarantee this is what will execute.

module.ecs.aws_autoscaling_group.asg
  launch_configuration: "ECS-Nodes-Windows-aaaaaaaaaaaa" => "${aws_launch_configuration.config.name}"

-/+ module.ecs.aws_launch_configuration.config
    associate_public_ip_address: "false" => "false"
    ebs_block_device.#:          "0" => "<computed>"
    ebs_optimized:               "false" => "<computed>"
    enable_monitoring:           "true" => "true"
    iam_instance_profile:        "ecsInstanceRole" => "ecsInstanceRole"
    image_id:                    "ami-9667ef80" => "ami-9667ef80"
    instance_type:               "m3.large" => "m3.large"
    key_name:                    "dev-infrastructure" => "dev-infrastructure"
    name:                        "ECS-Nodes-Windows-aaaaaaaaaaaaaa" => "<computed>"
    name_prefix:                 "ECS-Nodes-Windows-" => "ECS-Nodes-Windows-"
    root_block_device.#:         "0" => "<computed>"
    security_groups.#:           "2" => "2"
    security_groups.2618005233:  "sg-xxxxxxx" => "sg-yyyyyyyy"
    security_groups.3145808010:  "sg-xxxxxxx" => "sg-yyyyyyyy"
    spot_price:                  "0.15" => "0.20"
    user_data:                   "dada0cf5264fa15d5d1186cccce2c645759b87ba" => "7e51566e162b61376bd460ce9c82a5c591a7dfa8" (forces new resource)


Plan: 1 to add, 1 to change, 1 to destroy.

------------------------------------------------------------------------------------------------------------------------------------
Summary:
    C:/Source/Infra/infrastructure-module/terragrunt : No change
    C:/Source/Infra/infrastructure-module/terragrunt/common : No change
    C:/Source/Infra/infrastructure-module/terragrunt/deploy_tf : No change
    C:/Source/Infra/infrastructure-module/terragrunt/infra : No effective change
    C:/Source/Infra/infrastructure-module/terragrunt/containers_cluster : 1 to add, 1 to change, 1 to destroy.
```

jocgir:
Here is an example of the output-all (stdout only):

```
------------------------------------------------------------------------------------------------------------------------------------
C:/Source/Infra/infrastructure-module/terragrunt/common

account_id = 00000000000
env_long = develop
env_name = Development
env_short = dev
region = us-east-1
region_name = North Virginia
replication_prefix = odev
replication_region = us-west-2

------------------------------------------------------------------------------------------------------------------------------------
C:/Source/Infra/infrastructure-module/terragrunt

last_applied_on = 2017-04-24T17:26:24Z

------------------------------------------------------------------------------------------------------------------------------------
C:/Source/Infra/infrastructure-module/terragrunt/deploy_tf

access_key = AKIAXXXXXXXXXXXXXXX

------------------------------------------------------------------------------------------------------------------------------------
C:/Source/Infra/infrastructure-module/terragrunt/infra

availability_zones = [
    us-east-1a,
    us-east-1b,
    us-east-1c,
    us-east-1e
]
buckets = {
  backup = uniqueid-xdev-backup
  binaries = uniqueid-xdev-binaries
  config = uniqueid-xdev-config
  logs = uniqueid-xdev-logs
}
buckets_arns = {
  backup = arn:aws:s3:::uniqueid-xdev-backup
  binaries = arn:aws:s3:::uniqueid-xdev-binaries
  config = arn:aws:s3:::uniqueid-xdev-config
  logs = arn:aws:s3:::uniqueid-xdev-logs
}
domain_id = ZWAEKHF4231
domain_name = cloud.uniqueid.com
keypairs = {
  infrastructure = dev-infrastructure
}
ops_sg = sg-xxxxxxx
peer_vpc_cidr = 10.0.0.0/16
subnets_cidr = {
  container_clusters = [10.1.16.0/22 10.1.20.0/22 10.1.24.0/22 10.1.28.0/22]
  public = [10.1.0.0/26 10.1.0.64/26 10.1.0.128/26 10.1.0.192/26]
}
subnets_id = {
  container_clusters = [subnet-xxxx subnet-xxxx subnet-xxxxx subnet-xxxx]
  public = [subnet-xxxx subnet-xxxx subnet-xxxx subnet-xxxx]
}
vpc_cidr = 10.1.0.0/16
vpc_id = vpc-xxxxxxxx

------------------------------------------------------------------------------------------------------------------------------------
C:/Source/Infra/infrastructure-module/terragrunt/containers_cluster

windows_cluster = {
  config = ECS-Nodes-004d8df540fce0365ee86a4f7e
  id = arn:aws:ecs:us-east-1:00000000000:cluster/Cluster_Beta
  image_id = ami-9667ef80
  instance_type = m3.large
  name = Cluster_Beta
  spot_price = 0.16
}
```

configstack/running_module.go
brikis98: Are you saying `Printf` calls are being interleaved on a per-character basis? I've never seen that and we do a TON of logging in parallel in Go as part of our automated tests in dozens of projects... Is that a Windows-specific issue? Or a result of calling `fmt.Fprintf` as opposed to `Logger.Printf`?

configstack/running_module.go
brikis98: The modules themselves would have to change their error codes before returning them

configstack/running_module.go
brikis98: Are you saying the logging is being interleaved on a per-character basis? I've never seen that and we do a TON of logging in parallel in Go as part of our automated tests in over a dozen projects... Is that a Windows-specific issue? Or a result of calling fmt.Fprintf as opposed to Logger.Printf?

cli/cli_app.go
brikis98: I completely agree about not polluting output, but the way to do that is not to throw the output away, but to use different logging levels or, as you said, to format the output differently.

configstack/stack_plan.go
brikis98: My immediate thought was "what's the difference between no change and no effective change?". I'd just list "no change" for both...

configstack/stack_plan.go
brikis98: I'd rather rely on a public API than text in logging.

configstack/stack_plan.go
brikis98: Oh, and that summary is quite nice!

shell/run_shell_cmd.go
brikis98: Ah, yea, that's a bit of a problem. Should be logged at the least.

brikis98: Or just return `stdout.String(), err` without any if-statement.

brikis98:
The output looks pretty nice!

shell/run_shell_cmd.go
brikis98: Ah, I see you're handling this in #201.

brikis98:
Also, all this new behavior will need some tests.

configstack/running_module.go
jocgir: Not the logging, it is thread safe.

But I got that kind of mixup on the output that is written directly to stdout and I got it with terragrunt without any of the recent change.

Just made tests and yes, I got that kind of mixup only on Windows...

But with the lock, there is no problem on Windows, nor on ios.

brikis98: Gah, Windows :(

The lock code is currently fairly simple, but typically, locks only get more and more complicated. Moreover, they don't compose. As a result, I'd like to think about an alternative strategy.

For example, one option is to have a goroutine listening for strings on a channel. Every string it gets, it logs to stdout. You could then send strings to that channel any time you want to write them to `stdout`, and you don't need any lock, as the channel takes care of ordering for you.

jocgir:
Any idea of tests? In fact, there is not a lot of new behaviour. It is just a reorganization of how we process what is already existing...

configstack/stack_plan.go
```diff
@@ -0,0 +1,167 @@
...
+			// We do not want to consider CHANGE_EXIT_CODE as an error and not execute the dependants because there is an "error" in the dependencies.
+			// CHANGE_EXIT_CODE is not an error in this case, it is simply a status. We will reintroduce the exit code at the very end to mimic the behaviour
+			// of the native terrafrom plan -detailed-exitcode to exit with CHANGE_EXIT_CODE if there are changes in any of the module in the stack.
+			err = nil
+		}
+
+		message, count := extractSummaryResultFromPlan(output)
+
+		// We add the result to the result list (there is no concurrency problem because it is handled by the running_module)
+		*results = append(*results, moduleResult{module, err, message, count})
```
brikis98: Couldn't this function be called concurrently from multiple goroutines? If so, this mutation may cause issues.

brikis98:
> Any idea of tests? In fact, there is not a lot of new behaviour. It is just a reorganization of how we process what is already existing...

Well, as a first step, we should make sure the existing tests are passing :)

I'll leave some comments in the code with ideas for other tests.

configstack/stack_plan.go
jocgir: Any suggestion? I checked in terraform code and those strings are coded in private const.

configstack/stack_plan.go
```diff
@@ -0,0 +1,167 @@
...
+	fmt.Fprintf(terragruntOptions.Writer, "%s\nSummary:\n", separator)
+	for _, result := range results {
+		errMsg := ""
+		if result.Err != nil {
+			errMsg = fmt.Sprintf(", Error: %v", result.Err)
+		}
+
+		fmt.Fprintf(terragruntOptions.Writer, "    %v : %v%v\n", result.Module.Path, result.Message, errMsg)
+	}
+}
```
brikis98: This method could return the summary as a string rather than writing it to a `Writer`. If you do that, it should be easy to write a unit test for it.

configstack/stack_plan.go
```diff
@@ -0,0 +1,167 @@
...
+		if len(module.Dependencies) > 0 {
+			dependenciesMsg = fmt.Sprintf(" contains dependencies to %v and", module.Config.Dependencies.Paths)
+		}
+		module.TerragruntOptions.Logger.Printf("%v%v refers to remote state "+
+			"you may have to apply your changes in the dependencies prior running terragrunt plan-all.\n",
+			module.Path,
+			dependenciesMsg,
+		)
+	}
+}
```
brikis98: This method could return a string or list of strings with warnings in them rather than logging directly. That way, you could write a unit test for it.

configstack/stack_plan.go
```diff
@@ -0,0 +1,167 @@
...
+
+// Parse the output message to extract a summary
+func extractSummaryResultFromPlan(output string) (string, int) {
```
brikis98: There should be unit tests for this method.

configstack/stack_plan.go
```diff
@@ -0,0 +1,167 @@
...
+		if code, err := shell.GetExitCode(this.Errors[i]); err != nil {
+			return UNDEFINED_EXIT_CODE, this
+		} else if code == ERROR_EXIT_CODE || code == CHANGE_EXIT_CODE && exitCode == NORMAL_EXIT_CODE {
+			// The exit code 1 is more significant that the exit code 2 because it represents an error
+			// while 2 represent a warning.
+			return UNDEFINED_EXIT_CODE, this
+		}
+	}
+	return exitCode, nil
+}
```
brikis98: You could have unit tests or integration tests check that `PlanMultiError` and `countError` handle exit codes as expected.

configstack/stack_plan.go
brikis98: The easiest option is probably to always run `plan-all` with `-detailed-exitcode`. That way, the summary at the end just reduces to the exit codes you got for each module: 

```
Module XXX: no changes (exit code 0)
Module YYY: error (exit code 1)
Module ZZZ: there are changes to apply (exit code 2)
```

This should be reasonably reliable and provide decently useful info.

configstack/stack_plan.go
jocgir: In fact, it is protected by the lock 😄 . At the beginning, I put the lock around it, but moved it to running_modules to get a larger concurrent protection.

brikis98: Hm, that's not obvious, and will be quite error prone during refactors. I'd again vote in favor of using goroutines/channels. That is, instead of editing the `results` array here, you send a message on a channel, and the goroutine  listening on that channel is the only thing allowed to modify `results`.

jocgir: Yes, I figured it out.

brikis98:
@jocgir Let me know what's the status with this PR.

jocgir:
Sorry, lack of time.

One of my colleague is supposed to write the tests to complete this PR.

davidvuong:
Can we get an update on this? `plan-all/apply-all` swallowing errors really sucks.

PR #201, opened 2017-04-29
Fix problem on Windows with temp directory name encoding

jocgir:
I just found out that there is a bug on Windows.

The algorithm used to encode directory name use a base 64 encoding. But this generates names with non alphanumeric characters that are invalid for some command on Windows.

When we call `terraform get`, it tries to issue a `mklink` command to link the `.terraform\modules\<tempname>` to the actual module source folder. Since `terragrunt` is working in a temp directory with mangled names containing invalid characters (especially the = at the end), that cause the `mklink` command to fail on windows.

Also, I fixed a problem with `RunShellCommandAndCaptureOutput()`. In case of error, this function didn't return any captured output. I think the function should returns the captured buffer in any circumstance otherwise, we are loosing the output to stderr.

### DIFF
--- a/util/hash.go
+++ b/util/hash.go
@@ -3,10 +3,18 @@ package util
 import (
 	"crypto/sha1"
 	"encoding/base64"
+	"regexp"
 )
 
 // Returns the base 64 encoded sha1 hash of the given string
 func EncodeBase64Sha1(str string) string {
 	hash := sha1.Sum([]byte(str))
 	return base64.URLEncoding.EncodeToString(hash[:])
+}
+
+var removeNonAlphaNumericCharsRegex = regexp.MustCompile("[^a-zA-Z0-9]+")
+
+// Returns the base 64 encoded sha1 hash of the given string keeping only alphanumeric characters
+func EncodeBase64Sha1AlphaNum(str string) string {
+	return removeNonAlphaNumericCharsRegex.ReplaceAllString(EncodeBase64Sha1(str), "")
 }
--- a/util/hash.go
+++ b/util/hash.go
@@ -3,18 +3,10 @@ package util
 import (
 	"crypto/sha1"
 	"encoding/base64"
-	"regexp"
 )
 
 // Returns the base 64 encoded sha1 hash of the given string
 func EncodeBase64Sha1(str string) string {
 	hash := sha1.Sum([]byte(str))
-	return base64.URLEncoding.EncodeToString(hash[:])
-}
-
-var removeNonAlphaNumericCharsRegex = regexp.MustCompile("[^a-zA-Z0-9]+")
-
-// Returns the base 64 encoded sha1 hash of the given string keeping only alphanumeric characters
-func EncodeBase64Sha1AlphaNum(str string) string {
-	return removeNonAlphaNumericCharsRegex.ReplaceAllString(EncodeBase64Sha1(str), "")
+	return base64.RawURLEncoding.EncodeToString(hash[:])
 }
--- a/cli/download_source.go
+++ b/cli/download_source.go
@@ -195,7 +195,7 @@ func processTerraformSource(source string, terragruntOptions *options.Terragrunt
 		return nil, err
 	}
 
-	encodedWorkingDir := util.EncodeBase64Sha1AlphaNum(canonicalWorkingDir)
+	encodedWorkingDir := util.EncodeBase64Sha1(canonicalWorkingDir)
 	downloadDir := util.JoinPath(os.TempDir(), "terragrunt-download", encodedWorkingDir, rootPath)
 	workingDir := util.JoinPath(downloadDir, modulePath)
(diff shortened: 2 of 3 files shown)


util/hash.go
```diff
@@ -3,10 +3,18 @@ package util
...
 func EncodeBase64Sha1(str string) string {
 	hash := sha1.Sum([]byte(str))
 	return base64.URLEncoding.EncodeToString(hash[:])
 }
+
+var removeNonAlphaNumericCharsRegex = regexp.MustCompile("[^a-zA-Z0-9]+")
+
+// Returns the base 64 encoded sha1 hash of the given string keeping only alphanumeric characters
+func EncodeBase64Sha1AlphaNum(str string) string {
+	return removeNonAlphaNumericCharsRegex.ReplaceAllString(EncodeBase64Sha1(str), "")
```
brikis98: Throwing away characters will result in unwanted collisions. A better way to solve this is to create our own encoding instead of using `base64.URLEncoding`:

```go
URLEncodingAlphaNumOnly := base64.NewEncoding("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789").WithPadding(base64.NoPadding)

URLEncodingAlphaNumOnly.EncodeToString(hash[:])
```

jocgir: I get:
```
encoding alphabet is not 64-bytes long
```

brikis98: Whoops, copy/paste error. Use:

```
ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_
```

As the alphabet (I lost the `-_` at the end in my first copy/pate).

brikis98: (I'm pretty sure `-` and `_` are valid in Windows file paths)

brikis98: Oh, as it turns out, `base64.RawURLEncoding` is exactly this encoding, so you could use that instead of creating a new one :)

jocgir: OK, Nice. Thank you!

brikis98:
LGTM, thanks!

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.12.15

brikis98:
@jocgir Looks like this PR causes some test failures:

```
--- FAIL: TestAlreadyHaveLatestCodeRemoteFilePathDownloadDirExistsWithVersionAndVersionFile (0.00s)
	Error Trace:	download_source_test.go:180
			download_source_test.go:66
	Error:      	Not equal: 
	            	expected: true
	            	received: false
	Messages:   	For terraform source TerraformSource{CanonicalSourceURL = http://www.some-url.com?ref=v0.0.1, DownloadDir = ../test/fixture-download-source/download-dir-version-file, WorkingDir = ../test/fixture-download-source/download-dir-version-file, VersionFile = ../test/fixture-download-source/download-dir-version-file/version-file.txt}

--- FAIL: TestAlreadyHaveLatestCodeRemoteFilePathDownloadDirExistsNoVersionWithVersionFile (0.00s)
	Error Trace:	download_source_test.go:180
			download_source_test.go:48
	Error:      	Not equal: 
	            	expected: true
	            	received: false
	Messages:   	For terraform source TerraformSource{CanonicalSourceURL = http://www.some-url.com, DownloadDir = ../test/fixture-download-source/download-dir-version-file-no-query, WorkingDir = ../test/fixture-download-source/download-dir-version-file-no-query, VersionFile = ../test/fixture-download-source/download-dir-version-file-no-query/version-file.txt}

--- FAIL: TestDownloadTerraformSourceIfNecessaryRemoteUrlToAlreadyDownloadedDir (7.30s)
	Error Trace:	download_source_test.go:166
			download_source_test.go:110
	Error:      	Not equal: 
	            	expected: "# Hello, World 2"
	            	received: "# Hello, World"
	Messages:   	For terraform source TerraformSource{CanonicalSourceURL = github.com/gruntwork-io/terragrunt//test/fixture-download-source/hello-world, DownloadDir = /tmp/download-source-test096492682, WorkingDir = /tmp/download-source-test096492682, VersionFile = /tmp/download-source-test096492682/version-file.txt}

--- FAIL: TestDownloadTerraformSourceIfNecessaryRemoteUrlToAlreadyDownloadedDirSameVersion (7.67s)
	Error Trace:	download_source_test.go:166
			download_source_test.go:134
	Error:      	Not equal: 
	            	expected: "# Hello, World version remote"
	            	received: "# Hello, World"
	Messages:   	For terraform source TerraformSource{CanonicalSourceURL = github.com/gruntwork-io/terragrunt//test/fixture-download-source/hello-world?ref=v0.9.7, DownloadDir = /tmp/download-source-test825628568, WorkingDir = /tmp/download-source-test825628568, VersionFile = /tmp/download-source-test825628568/version-file.txt}
```

My guess is that all of these tests have a base64 hash in them that needs to be updated to the new encoding. Could you take a look?

PR #203, opened 2017-04-30
Fix the expected values from the tests.

jocgir:
The base64 encoding were changing by #201 and tests had expected values hardcode.


brikis98:
Thanks!

brikis98:
OK, tests are now passing. New build should be here soon: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.12.15

PR #204, opened 2017-05-03
Update README.md to add snap install instructions

popey:
This pull request adds installation instructions for users of [snap enabled Linux distributions](https://snapcraft.io/docs/core/install). Snaps enable ISVs and projects to make their software available to millions of Linux users via the snap store and directly control delivery of software updates to their users. Snaps of stable desktop applications will also be available via the desktop Software Center, improving the discover-ability of your software. Your software can be installed with a simple `snap install terragrunt` simplifying the installation instructions for your users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ multiple Terraform modules, remote state, and locking.
 
 1. Install [Terraform](https://www.terraform.io/).
 
-1. Install Terragrunt by going to the [Releases Page](https://github.com/gruntwork-io/terragrunt/releases), downloading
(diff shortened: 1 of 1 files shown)


brikis98:
Thanks for the PR! It occurs to me that Terragrunt is available in [Homebrew](https://libraries.io/homebrew/terragrunt) as well, so we should probably update the instructions to say "Install Terragrunt using one of the following options" and then list:

1. Manually by going to the releases page
1. On OS X: `brew install terragrunt`
1. On Linux: `snap install terragrunt`

conorgil:
I just submitted PR #216, which includes the instructions for installing with snap listed in this issue.

Issue #171, opened 2017-04-21
Add helper function for listing all terraform commands to which locking applies.


josh-padnick:
Here's a common recommended `terraform.tfvars`:

```hcl
terragrunt = {
   ...
   terraform {
     # Force Terraform to keep trying to acquire a lock for up to 20 minutes if someone else already has the lock 
      extra_arguments "retry_lock" {
         commands = [
            "init",
            "apply",
            "refresh",
            "import",
            "plan",
            "taint",
            "untaint"
         ]
      }
   }
}
```

But this list of terraform commands that apply to locks is non-obvious. It'd be nice if we could add another helper command here that generates this array for you. It would make the `terraform.tfvars` look something like:

```hcl
terraform { 
   extra_arguments "retry_lock" {
      commands = [ get_terraform_commands_that_need_locking() ]
   }
}
```

brikis98:
Should be fixed by #205.

Issue #142, opened 2017-02-21
import with extra arguments fails due to argument order


bdashrad:
Using the following config, import does not work as expected, due to resulting argument order:

```
terragrunt= {
  terraform = {
    {
      extra_arguments "secrets" {
        arguments = [
          "-var-file=terraform.tfvars",
          "-var-file=terraform-secret.tfvars"
        ]
        commands = [
          "apply",
          "plan",
          "import",
          "push",
          "refresh"
        ]
      }
    }
  }
```

```
$ terragrunt import module.legacy.aws_elb.elb staging-balancer
[terragrunt] 2017/02/21 12:50:35 Reading Terragrunt config file at /Users/bradclark/codeterraform/environments/staging/terraform.tfvars
[terragrunt] 2017/02/21 12:50:35 Remote state is already configured for backend s3
[terragrunt] 2017/02/21 12:50:35 Attempting to acquire lock for state file staging in DynamoDB
[terragrunt] 2017/02/21 12:50:35 Attempting to create lock item for state file staging in DynamoDB table terragrunt_locks
[terragrunt] 2017/02/21 12:50:36 Lock acquired!
[terragrunt] 2017/02/21 12:50:36 Running command: terraform import module.legacy.aws_elb.elb staging-balancer -var-file=terraform.tfvars -var-file=staging.tfvars
The import command expects two arguments.
Usage: terraform import [options] ADDR ID
```

The extra arguments (options) need to be passed between `terraform import` and the resources to import, like this:

```
$ terraform import -var-file=terraform.tfvars -var-file=staging.tfvars module.legacy.aws_elb.elb staging-balancer
module.legacy.aws_elb.elb: Importing from ID "staging-balancer"...
module.legacy.aws_elb.elb: Import complete!
  Imported aws_elb (ID: staging-balancer)
module.legacy.aws_elb.elb: Refreshing state... (ID: staging-balancer)

Import success! The resources imported are shown above. These are
now in your Terraform state. Import does not currently generate
configuration, so you must do this next. If you do not create configuration
for the above resources, then the next `terraform plan` will mark
them for destruction.
```

brikis98:
Ah, good catch. I suspect there are probably other commands that have a similar issue. PRs to fix this are very welcome!

brikis98:
Should be fixed by #205.

PR #205, opened 2017-05-03
New interpolation functions to handle terraform commands

jocgir:
This should solve #171.

Add new interpolation functions to list terraform commands that make use of -var, -var-file and -lock-timeout.

`get_terraform_commands_that_need_locking()`
`get_terraform_commands_that_need_vars()`

```hcl
terragrunt = {
  terraform {
    # Force Terraform to keep trying to acquire a lock for up to 20 minutes if someone else already has the lock
    extra_arguments "retry_lock" {
      commands  = "${get_terraform_commands_that_need_locking()}"
      arguments = ["-lock-timeout=20m"]
    }
  }
}
```

Note that the implementation has been a little bit complicated by the fact that we cannot just implement it as suggested:
```hcl
      commands  = ${get_terraform_commands_that_need_locking()}
```

This syntax is not compliant with HCL as interpolation function should be between quotes.
```hcl
      commands  = "${get_terraform_commands_that_need_locking()}"
```

But, since these functions are returning a list of string, we do not want to get this as result:
```hcl
      commands  = "["apply", "plan", ...]"
```

Which is invalid. We expect to get rid of the surrounding quotes and get:
```hcl
      commands  = ["apply", "plan", ...]
```

So, I had to change the return of `resolveTerragruntInterpolation` and `executeTerragruntHelperFunction` to handle `interface{}` instead of `string`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -460,7 +460,7 @@ when creating the S3 bucket or DynamoDB table.
 
 ### Keep your CLI flags DRY
 
-#### <a name="extra_arguments"></a>Motivation
+#### Motivation
 
 Sometimes you may need to pass extra CLI arguments every time you run certain `terraform` commands. For example, you
 may want to set the `lock-timeout` setting to 20 minutes for all commands that may modify remote state so that
@@ -507,7 +507,7 @@ terragrunt = {
   terraform {
     # Force Terraform to keep trying to acquire a lock for up to 20 minutes if someone else already has the lock
     extra_arguments "retry_lock" {
-      commands  = "${get_terraform_commands_that_need_locking()}"
+      commands  = ["${get_terraform_commands_that_need_locking()}"]
       arguments = ["-lock-timeout=20m"]
     }
   }
@@ -1169,18 +1169,43 @@ The common.tfvars located in the terraform root folder will be included by all a
 
 `get_terraform_commands_that_need_vars()`
 
-Returns the list of terraform commands that accept -var and -var-file parameters. This function is used when defining [extra_arguments](#extra_arguments):
+Returns the list of terraform commands that accept -var and -var-file parameters. This function is used when defining [extra_arguments](#keep-your-cli-flags-dry).
+
+```
+terragrunt = {
+  terraform = {
+    ...
+
+    extra_arguments "common_var" {
+      commands  = ["${get_terraform_commands_that_need_vars()}"]
+      arguments = ["-var-file=${get_aws_account_id()}.tfvars"]
+    }
+  }
+}
+```
 
 #### get_terraform_commands_that_need_locking
 
 `get_terraform_commands_that_need_locking()`
 
-Returns the list of terraform commands that accept -lock-timeout parameter. This function is used when defining [extra_arguments](#extra_arguments):
+Returns the list of terraform commands that accept -lock-timeout parameter. This function is used when defining [extra_arguments](#keep-your-cli-flags-dry).
+
+```hcl
+terragrunt = {
+  terraform {
+    # Force Terraform to keep trying to acquire a lock for up to 20 minutes if someone else already has the lock
+    extra_arguments "retry_lock" {
+      commands  = ["${get_terraform_commands_that_need_locking()}"]
+      arguments = ["-lock-timeout=20m"]
+    }
+  }
+}
+```
 
 _Note: Functions that return a list of values must be used in a single declaration like:_
 
 ```hcl
-commands = "${get_terraform_commands_that_need_vars()}"
+commands = ["${get_terraform_commands_that_need_vars()}"]
 
 # which result in:
 commands = ["apply", "console", "destroy", "import", "plan", "push", "refresh"]
@@ -1218,7 +1243,7 @@ terragrunt = {
     ...
 
     extra_arguments "common_var" {
-      commands = "${get_terraform_commands_that_need_vars()}"
+      commands = ["${get_terraform_commands_that_need_vars()}"]
       arguments = ["-var-file=${get_aws_account_id()}.tfvars"]
     }
   }
--- a/README.md
+++ b/README.md
@@ -460,7 +460,7 @@ when creating the S3 bucket or DynamoDB table.
 
 ### Keep your CLI flags DRY
 
-#### Motivation
+#### <a name="extra_arguments"></a>Motivation
 
 Sometimes you may need to pass extra CLI arguments every time you run certain `terraform` commands. For example, you
 may want to set the `lock-timeout` setting to 20 minutes for all commands that may modify remote state so that
@@ -496,6 +496,23 @@ terragrunt = {
 Each `extra_arguments` block includes an arbitrary name (in the example above, `retry_lock`), a list of `commands` to
 which the extra arguments should be add, a list of `arguments` or `required_var_files` or `optional_var_files` to add.
 With the configuration above, when you run `terragrunt apply`, Terragrunt will call Terraform as follows:
+
+When available, it is preferable to use interpolation functions such as
+[get_terraform_commands_that_need_locking](#get_terraform_commands_that_need_locking) and
+[get_terraform_commands_that_need_vars](#get_terraform_commands_that_need_vars)
+since they provide the complete list of terraform commands that make use of the desired parameter:
+
+```hcl
+terragrunt = {
+  terraform {
+    # Force Terraform to keep trying to acquire a lock for up to 20 minutes if someone else already has the lock
+    extra_arguments "retry_lock" {
+      commands  = "${get_terraform_commands_that_need_locking()}"
+      arguments = ["-lock-timeout=20m"]
+    }
+  }
+}
+```
 
 ```
 > terragrunt apply
@@ -865,6 +882,8 @@ Here are the supported helper functions:
 * [get_env](#get_env)
 * [get_tfvars_dir](#get_tfvars_dir)
 * [get_parent_tfvars_dir](#get_parent_tfvars_dir)
+* [get_terraform_commands_that_need_vars](#get_terraform_commands_that_need_vars)
+* [get_terraform_commands_that_need_locking](#get_terraform_commands_that_need_locking)
 * [get_aws_account_id](#get_aws_account_id)
 
 
@@ -1146,6 +1165,34 @@ terragrunt = {
 
 The common.tfvars located in the terraform root folder will be included by all applications, whatever their relative location to the root.
 
+#### get_terraform_commands_that_need_vars
+
+`get_terraform_commands_that_need_vars()`
+
+Returns the list of terraform commands that accept -var and -var-file parameters. This function is used when defining [extra_arguments](#extra_arguments):
+
+#### get_terraform_commands_that_need_locking
+
+`get_terraform_commands_that_need_locking()`
+
+Returns the list of terraform commands that accept -lock-timeout parameter. This function is used when defining [extra_arguments](#extra_arguments):
+
+_Note: Functions that return a list of values must be used in a single declaration like:_
+
+```hcl
+commands = "${get_terraform_commands_that_need_vars()}"
+
+# which result in:
+commands = ["apply", "console", "destroy", "import", "plan", "push", "refresh"]
+
+# They shall not be used to do string composition like:
+commands = "Some text ${get_terraform_commands_that_need_locking()}"
+
+# which result in:
+commands = "Some text [apply destroy import init plan refresh taint untaint]"
+```
+
+
 #### get_aws_account_id
 
 `get_aws_account_id()` returns the AWS account id associated with the current set of credentials. Example:
@@ -1171,17 +1218,8 @@ terragrunt = {
     ...
 
     extra_arguments "common_var" {
-      commands = [
-        "apply",
-        "plan",
-        "import",
-        "push",
-        "refresh"
-      ]
-
-      arguments = [
-        "-var-file=${get_aws_account_id()}.tfvars",
-      ]
+      commands = "${get_terraform_commands_that_need_vars()}"
+      arguments = ["-var-file=${get_aws_account_id()}.tfvars"]
     }
   }
 }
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -460,7 +460,7 @@ when creating the S3 bucket or DynamoDB table.
 
 ### Keep your CLI flags DRY
 
-#### Motivation
+#### <a name="extra_arguments"></a>Motivation
```
brikis98: GitHub headers automatically have anchor links. You could just link to #keep-your-cli-flags-dry.

README.md
```diff
@@ -1146,6 +1165,34 @@ terragrunt = {
 
 The common.tfvars located in the terraform root folder will be included by all applications, whatever their relative location to the root.
 
+#### get_terraform_commands_that_need_vars
+
+`get_terraform_commands_that_need_vars()`
+
+Returns the list of terraform commands that accept -var and -var-file parameters. This function is used when defining [extra_arguments](#extra_arguments):
```
brikis98: Replace the `:` with a `.` at the end of the sentence.

brikis98: Or, better yet, add some example code after the `:` :)

README.md
```diff
@@ -497,6 +497,23 @@ Each `extra_arguments` block includes an arbitrary name (in the example above, `
...
+[get_terraform_commands_that_need_locking](#get_terraform_commands_that_need_locking) and
+[get_terraform_commands_that_need_vars](#get_terraform_commands_that_need_vars)
+since they provide the complete list of terraform commands that make use of the desired parameter:
+
+```hcl
+terragrunt = {
+  terraform {
+    # Force Terraform to keep trying to acquire a lock for up to 20 minutes if someone else already has the lock
+    extra_arguments "retry_lock" {
+      commands  = "${get_terraform_commands_that_need_locking()}"
```
brikis98: Would the implementation and docs be simplified if `get_terraform_commands_that_need_locking()` returned a comma-separated list of command names and you the user provided the brackets themselves?

```
commands = ["${get_terraform_commands_that_need_locking()}"]
```

brikis98: That's also more readable.

README.md
jocgir: OK, I wasn't sure how to link to title with spaces, thank you.

README.md
jocgir: OK for the . I am hesitant to add more example since there is already an example of both commands.

README.md
jocgir: That doesn't really simplify the implementation because we still have to generate a comma separated list of quoted strings and still have to get rid of the surrounding quotes around the declaration itself.

This is the technique used by terraform. When a command returns an array, the result is an array and there is no need to add the brackets.The only advantage would be to allow the user to specify more commands, but since the list is complete, there is no real use.

jocgir:
@brikis98, If you don't mind, I found a bug with the current management of extra_arguments. I fixed it, but will include the fix in this PR.

command like `terraform taint -module xxx yyy` which mark the resource yyy of module xxx for refresh is not handling the `extra_arguments` properly.


If terragrunt is configured to inject -lock-timeout on command.
```
terragrunt taint -module xxx yyy
```
will result in:
```
terraform taint -module xxx yyy -lock-timeout=20m
```
which give an error. `terraform` expect to receive it's options between the command `taint` and the arguments. So it expect:
```
terraform taint -module xxx -lock-timeout=20m  yyy 
```
But is is difficult to find the proper insertion point. So the easiest way to do thing is to insert the extra arguments directly after the command:
```
terraform taint -lock-timeout=20m -module xxx yyy 
```
My suggested implementation handle the fact that some `terraform` commands require sub command like `terraform state <sub-command>`.

This also has the advantage of not overwriting a desired behaviour. If we automatically inject `-var region-us-east-1` in the terragrunt configuration.
```
terragrunt apply -var region=us-west-2
```
We get:
```
terraform apply -var region=us-west-2 -var region=us-east-1
```
Which override the explicitly declared value supplied by the user

With the proposed change, we get
```
terraform apply -var region=us-east-1 -var region=us-west-2
```
Which result in a more compliant command to the user intent.

brikis98:
Ah, yes, that's #142. I like your idea for a fix!

brikis98:
@jocgir What is the latest status with this PR?

jocgir:
Hi @brikis98,

If I remind well, I gave made the modifications you proposed and I was waiting for your comment or approval.

README.md
```diff
@@ -1146,6 +1165,59 @@ terragrunt = {
...
+
+_Note: Functions that return a list of values must be used in a single declaration like:_
+
+```hcl
+commands = ["${get_terraform_commands_that_need_vars()}"]
+
+# which result in:
+commands = ["apply", "console", "destroy", "import", "plan", "push", "refresh"]
+
+# They shall not be used to do string composition like:
```
brikis98: s/They shall not/They cannot/`

cli/cli_app.go
```diff
@@ -182,7 +188,17 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
...
+		if util.ListContainsElement(TERRAFORM_COMMANDS_WITH_SUBCOMMAND, terragruntOptions.TerraformCliArgs[0]) {
+			commandLength = 2
+		}
+		var args []string
+		args = append(args, terragruntOptions.TerraformCliArgs[:commandLength]...)
+
+		// Options must be inserted after command but before the other args
+		// command is either 1 word or 2 words
+		args = append(args, filterTerraformExtraArgs(terragruntOptions, conf)...)
+		terragruntOptions.TerraformCliArgs = append(args, terragruntOptions.TerraformCliArgs[commandLength:]...)
```
brikis98: It would be more readable to use a consistent structure for building up the list of `args`:

```go
var args []string

// Options must be inserted after command but before the other args
args = append(args, filterTerraformExtraArgs(terragruntOptions, conf)...)
args = append(args, terragruntOptions.TerraformCliArgs[commandLength:]...)

terragruntOptions.TerraformCliArgs = args
```

brikis98: Also, we need a test case for this!

config/config_helpers.go
```diff
@@ -28,19 +52,46 @@ type EnvVar struct {
 func ResolveTerragruntConfigString(terragruntConfigString string, include *IncludeConfig, terragruntOptions *options.TerragruntOptions) (resolved string, finalErr error) {
 	// The function we pass to ReplaceAllStringFunc cannot return an error, so we have to use named error
 	// parameters to capture such errors.
-	resolved = INTERPOLATION_SYNTAX_REGEX.ReplaceAllStringFunc(terragruntConfigString, func(str string) string {
+
+	resolved = INTERPOLATION_SYNTAX_REGEX_SINGLE.ReplaceAllStringFunc(terragruntConfigString, func(str string) string {
+		// We do a special treatment for function returning an array of values because they need
+		// to be declared like that "${function()" and the result must get rid of the surrounding quote
```
brikis98: Might be clearer to say:

```
All interpolation functions are called using the syntax "${function_name()}". For functions that return a non-string value we have to get rid of the surrounding quotes and convert the output to HCL syntax. For example, for an array, we need to return [1, 2, 3].
```

config/config_helpers.go
```diff
@@ -28,19 +52,46 @@ type EnvVar struct {
...
+		internalStr := str[1 : len(str)-1] // We get rid of the surrounding quote
+		out, err := resolveTerragruntInterpolation(internalStr, include, terragruntOptions)
+		if err != nil {
+			finalErr = err
+		}
+
+		switch out := out.(type) {
+		case []string:
+			return util.ListToHCLArray(out)
+		}
```
brikis98: There should be two `case` statements here: the one for `[]string` you already have and a new one for `string` that returns the string unchanged. There should also be a `default` that sets `finalErr` to an error message indicating that an unsupported type was returned and using `reflect.TypeOf` to display what that type was.

config/config_helpers.go
```diff
@@ -28,19 +52,46 @@ type EnvVar struct {
...
+
+		// The function is not returning an array of string, so left the string unchanged
+		// Remaining functions will be processed by the next statement
+		return str
+	})
+
+	resolved = INTERPOLATION_SYNTAX_REGEX.ReplaceAllStringFunc(resolved, func(str string) string {
```
brikis98: Why are there two regex now and why should both be executed?

jocgir: There is one for the case where the functions is directly included within "" as `"${get_terraform_commands_that_need_vars()}"` which does not return the same result as something hypothetic like `"Hello ${get_terraform_commands_that_need_vars()} world!"`. In the first case, we return a list and in the second, we return the expansion of the list within a string.

README.md
jocgir: I changed it, but in fact, this is not resulting in an error. It is just that the result would be useless. That explain why I used shall not instead of cannot.

config/config_helpers.go
jocgir: I am not sure that this would be a plus. That would just add a restriction to future interpolation function. The goal here is to properly process the array return and let other any other values unchanged.

Then, the next statement take care of the rest.

README.md
brikis98: A useless result is worse than an error, as it's an error that's going to be hard to track down :)

How about "We do not recommend using them in string composition"?

config/config_helpers.go
brikis98: Ah, OK. TBH, this structure with two regular expressions and various shuffling of types is a bit confusing. At the very least, we need clear comments explaining how this works. I'll also think about a way to simplify this...

brikis98: Also, we need tests for this stuff too, as it has lots of corner cases.

util/collections.go
```diff
@@ -49,3 +54,12 @@ func removeDuplicatesFromList(list []string, keepLast bool) []string {
 	}
 	return out
 }
+
+// Returns an HCL compliant formated list of strings
+func ListToHCLArray(list []string) string {
```
brikis98: This function makes it sound like it'll return a string of the format `["foo", "bar", "baz"]`, but in reality, it returns `"foo", "bar", "baz"` (no brackets). Therefore, it should probably be called something like `ToCommaSeparatedStrings` or similar.

config/config_helpers.go
brikis98: OK, after thinking about this for a few minutes more, it just seems that Terraform's interpolation syntax, by requiring double quotes around interpolations, is inherently messy. I don't want to waste any more time on looking for a more elegant way to parse/process it, so for now, let's add do the following:

1. Move each regex `ReplaceAllStringFunc` call into its own function. The first one can be called `processSingleInterplationInString` and the second `processMultipleInterpolationsInString`. Each one should have a comment above it with an example of what it does. E.g. the comment above `processSingleInterplationInString` would look something like this:

    ```
    Process a string that has a single interpolation in it and nothing else, such as "${function()}". These sorts of interpolations allow us to render types other than a string. For example, if "${function}" returns an array, we can drop the wrapping quotes and render [1, 2, 3].
    ```
1. Add automated tests for these new interpolation behaviors.

util/collections.go
jocgir: Initially, this function was returning a list ;-) but you preferred that the user add the bracket itself.

config/config_helpers.go
```diff
@@ -49,47 +49,83 @@ type EnvVar struct {
...
+	// First, we replace all single interpolation syntax (i.e. function directly enclosed within quotes "${function()}")
+	terragruntConfigString, err := processSingleInterpolationInString(terragruntConfigString, include, terragruntOptions)
+	if err != nil {
+		return terragruntConfigString, err
+	}
+	// Then, we replace all other interpolation functions (i.e. functions not directly enclosed within quotes)
+	return processMultipleInterpolationsInString(terragruntConfigString, include, terragruntOptions)
+}
+
+// Execute a single Terragrunt helper function and return its value as a string
```
brikis98: s/and return its value as a string/and return the result/

config/config_helpers.go
```diff
@@ -49,47 +49,83 @@ type EnvVar struct {
...
-	// The function we pass to ReplaceAllStringFunc cannot return an error, so we have to use named error
-	// parameters to capture such errors.
+func ResolveTerragruntConfigString(terragruntConfigString string, include *IncludeConfig, terragruntOptions *options.TerragruntOptions) (string, error) {
+	// First, we replace all single interpolation syntax (i.e. function directly enclosed within quotes "${function()}")
+	terragruntConfigString, err := processSingleInterpolationInString(terragruntConfigString, include, terragruntOptions)
+	if err != nil {
+		return terragruntConfigString, err
+	}
+	// Then, we replace all other interpolation functions (i.e. functions not directly enclosed within quotes)
+	return processMultipleInterpolationsInString(terragruntConfigString, include, terragruntOptions)
```
brikis98: Ah, much easier to understand now, thx!

config/config_helpers_test.go
```diff
@@ -492,6 +493,27 @@ TERRAGRUNT_HIT","")}/bar`,
...
+			fmt.Sprintf("commands = [%s]", util.CommaSeparatedStrings(TERRAFORM_COMMANDS_NEED_VARS)),
+			nil,
+		},
+		{
+			`commands = "test-${get_terraform_commands_that_need_vars()}"`,
+			nil,
+			options.TerragruntOptions{TerragruntConfigPath: "/root/child/" + DefaultTerragruntConfigPath, NonInteractive: true},
+			fmt.Sprintf(`commands = "test-%v"`, TERRAFORM_COMMANDS_NEED_VARS),
+			nil,
+		},
```
brikis98: Nice tests! Could you add one more that calls two interpolation functions in one string?

brikis98:
OK, we are almost there. Two final comments and this is good to merge. Thank you for this important fix!

jocgir:
You are right, more tests we have, more problems we can identify. Read my latest commit to check what problem I faced :-)

brikis98:
Hahah, awesome. Merging now. Thanks!

brikis98:
New build is here: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.12.17

PR #209, opened 2017-05-04
--terragrunt-ignore-dependency-errors

mrtyler:
See #193.


brikis98:
👍 

Thank you for the PR. Nice work on the code, tests, and docs. I don't see anything that needs to be changed, so merging now!

brikis98:
Assuming the build passes, the new binaries will be here: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.12.17

brikis98:
Ah, looks like a previous PR caused some test failures. I'll look into that.

brikis98:
OK, build issues fixed, new binaries are here: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.12.16

PR #206, opened 2017-05-03
Update regex that checks Terraform version compat

luis-silva:
This PR is meant to fix #200 as well as to provide the functionality to run user compiled terraform binaries e.g. when fixing terraform bugs or adding extra functionality locally and before it is merged into the mainline terraform codebase.

PR #210, opened 2017-05-04
Fix version check test failures

brikis98:
Fixes test failures that came from #206.


luis-silva:
@brikis98 Thanks for the review, I didn't have a setup to run those tests so thank you for giving them a once over.

Issue #196, opened 2017-04-26
TestNewSignalsForwarderMultiple is flaky when run from top level


mrtyler:
Spinning this off from #194.

I was not able to cause this test to fail by running its module alone: `cd shell ; go test -v -parallel 128`.

However, I still see this test fail about 25% of the time when running as part of the full suite. Here's the setup:
* macos sierra 10.12.3 on a macbook pro. I was doing stuff on the machine during the test run (browser, shell, youtube videos) but no unusual load or anything I would expect to cause a test to fail so often
* terragrunt at aa9ae94a868b53d9e2f381c363313d6114e6f25d
* glide 0.12.3
* terraform 0.9.3
* Run this from top of terragrunt working copy: `for i in $(seq 1 22) ; do rm -f test.out ; go test -v -parallel 128 $(glide novendor) > test.out 2>&1 ; rc=$? ; if grep "FAIL" test.out ; then echo "FAIL with return code $rc" ; else echo "PASS with return code $rc" ; fi ; done`

This resulted in 17 successes and 5 failures. TestNewSignalsForwarderMultiple was the only test that failed during the run.

Flaky tests are difficult to debug, especially when they only flake in complex scenarios like 128 threads of simultaneous test action, so I won't be offended if you choose to close this out without a lot of action. I just wanted y'all to be aware.

brikis98:
Thank you for filing! I'm not sure what would cause this. cc @asobrien and @mildred who looked at this code more recently than me.

asobrien:
I haven't been able to recreate the failures @mrtyler has been seeing, even when putting my system (similar hardware) under `stress`.

The error you're seeing:

```
--- FAIL: TestNewSignalsForwarderMultiple (7.54s)
        assertions.go:226: ^M                          ^M       Error Trace:    run_shell_cmd_test.go:121
                ^M      Error:          Not equal:
                ^M                      expected: 10
                ^M                      received: 12
FAIL
FAIL    github.com/gruntwork-io/terragrunt/shell        7.564s
```

stems from the fact more `SIGINT` signals were emitted from the parent process than were received by the subprocess (`test_sigint_multiple.sh` in this case).

I'm curious, if disabling parallelism from this test would mitigate the issue. The relevant line is [here](https://github.com/gruntwork-io/terragrunt/blob/master/shell/run_shell_cmd_test.go#L83).

[...]

PR #213, opened 2017-05-05
Loosen assertions in TestNewSignalsForwarderMultiple.

mrtyler:
See https://github.com/gruntwork-io/terragrunt/issues/196.

I ran the tests a couple times and they passed.


brikis98:
Great, thanks!

Issue #166, opened 2017-04-18
Dealing with Hard Coded Credentials for Private Repos


tomstockton:
Hi Terragrunt team!

First of all, love your work, I've bought the 'Up & Running' book and am following your advice on using Terraform as closely as possible!

Specifically, I'm using the [pipelines pattern that you referenced](https://medium.com/@kief/https-medium-com-kief-using-pipelines-to-manage-environments-with-infrastructure-as-code-b37285a1cbf5) and have created the following structure:

````
project_live repo:
  environment/
    component/
      terraform.tfvars # defines git source to private 'project_modules repo'

project_modules repo:
  component/
    main.tf # defines modules from the private 'core_modules repo'

code_modules repo:
  module/
    main.tf # defines groups of resources
````

All these repos are private, hence needing to define in the format (as per the [documentation](https://www.terraform.io/docs/modules/sources.html#private-github-repos)) which hard codes user credentials into the git URL.  Obviously this isn't ideal as has been recognised in https://github.com/hashicorp/terraform/issues/1439 and I'm looking for a better solution!

Do you guys have any thoughts on how to deal with this?  Would it be a worthwhile feature that could be added to Terragrunt in order to allow variable interpolation into the source parameter?

I've achieved this with some success using a wrapper shell script to replace string `{VCS_CREDS}` with env var `${VCS_CREDS}` but it only works as far as pulling the `project_modules` repo.  Any references in `project_modules` to `core_modules` still have to be hardcoded due to `terraform init <url>` downloading any modules defined in `project_modules`.

Thanks!

**Comment by brikis98** [comment]
Use SSH URLs. 

```hcl
terragrunt = {
  terraform {
    source = "git::git@github.com:gruntwork-io/modules-vpc.git//modules/vpc-mgmt?ref=v1.0.8"
  }
}
```

This will use your SSH creds, so no need to store any passwords in the code itself.

tomstockton:
Ah, thanks.  I'd tried this but it failed.  I'm using bitbucket rather than github though and it seems that Terraform interprets `git::git@bitbucket.org:foo/bar.git//baz` as a reference to a local path.

I'll create an issue / PR for terraform to properly handle bitbucket SSH URLs.

brikis98:
For BitBucket, try:

```
git::ssh://git@bitbucket.org:foo/bar.git//baz
```

tomstockton:
Almost!  Running that gives:
```
[...]

PR #219, opened 2017-05-16
add documentation on github ssh authentication

fosrias:
@brikis98 This PR addresses https://github.com/gruntwork-io/terragrunt/issues/166#issuecomment-301110407 adding documentation on using ssh with terragrunt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,9 +301,10 @@ In particular:
 
 
 
-#### Github using SSH
-To use terragrunt with SSH github authentication, use the following syntax. Note the `/` after 
-github.com instead of `:` and the uses of `//`. 
+#### Using Terragrunt with private Git repos
+The easiest way to use Terragrunt with private Git repos is to use SSH authentication. 
+Configure your Git account so you can use it with SSH (see the [guide for GitHub here](https://help.github.com/articles/connecting-to-github-with-ssh/)) and use the SSH URL for your repo, 
+prepended with `git::ssh://`: 
 
 ```hcl
 terragrunt = {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -301,6 +301,25 @@ In particular:
 
 
 
+#### Github using SSH
+To use terragrunt with SSH github authentication, use the following syntax. Note the `/` after 
+github.com instead of `:` and the uses of `//`.
```
brikis98: I would change the heading title to "Using Terragrunt with private Git repos" and the first sentence to say: 

```
The easiest way to use Terragrunt with private Git repos is to use SSH authentication. Configure your Git account so you can use it with SSH (see the [guide for GitHub here](https://help.github.com/articles/connecting-to-github-with-ssh/) and use the SSH URL for your repo, prepended with `git::ssh://`:
```

README.md
```diff
@@ -301,6 +301,25 @@ In particular:
...
+github.com instead of `:` and the uses of `//`. 
+
+```hcl
+terragrunt = {
+  terraform {
+    source = "git::ssh://git@github.com/foo/modules.git//path/to/module?ref=v0.0.1"
+  }
+}
+```
+For private git repositories, replace github.com with the private repository host.
```
brikis98: I would replace this sentence with "Look up the Git repo for your repository to find the proper format." After all, GitHub has private repos too, and other types of private Git repos may have significantly different URL formats. For example, for private BitBucket repos, you may need to use a URL like `git::ssh://git@my-bitbucket-account.com:7999/team-name/repo-name.git//path/to/module?ref=v0.0.1`.

fosrias: @brikis98 I think I should leave in the comment about `:` as the point of this extra documentation is  partially that using it does not work after the host and your bitbucket example here included it ;). Past that will update later today.

brikis98: Which `:`? The one before the port number? If so, that definitely works with BitBucket...

fosrias: K. I will play with it a bit and double check. IIRC, it did not with github.

fosrias: Ah. Nm. Missed that it was a port. That should work. Problem is `:` vs `/` for path in normal github ssh path.

fosrias:
@brikis98 Comments addressed.

brikis98:
Wonderful, thanks!

PR #223, opened 2017-05-24
Fix a bug introduced by refactoring of new interpolation functions.

jocgir:
I introduced a bug when I refactored the new interpolation functions. I did not use the right variable in the regex FindStringSubmatch function.

Also, I think that we should not continue with the match function if there is an error. We should exit the function without returning a changed value.


jocgir:
Yes. I will add a test that would have catched the problem.

brikis98:
Perfect, thanks!

Issue #147, opened 2017-02-24
Fold "extra_arguments" into terragrunt when using "include" to retrieve remote TF configurations


rkr-kununu:
So, I've got the following setup:
* /terraform.tfvars
  * /preview/terraform.tfvars

The contents of `/terraform.tfvars` is:
```
terragrunt = {
  lock { ... }
  remote_state { ... }
  terraform {
    extra_arguments "global" {
      arguments = [ "-var", "account_id=12345" ]
    }
  }
}
```

The contents of `/preview/terraform.tfvars` is:
```
terragrunt = {
  include {
    path = "${find_in_parent_folders()}"
  }
  terraform {
    source = "git::git@github.com:foo/bar.git//app?ref=v0.0.1"
  }
}
```

The idea being that in app I would have access to `var.account_id`.  This is of course not possible (currently) because the terraform source-block will override the terraform extra_arguments-block.

The solution described in https://github.com/gruntwork-io/terragrunt/issues/143 could be applied (put my configuration into a separate tfvars), but it would mean I'd need to cut-and-paste the extra_arguments blocks into /preview/terraform.tfvars (and future staging environments).

brikis98:
Ah, yea, that's not ideal. 

I think the way we are merging keys from `include` is fairly coarse grained, where we simply override top-level keys: `lock`, `remote_state`, `terraform`. We could merge the values together, but that will leave other holes, as then there would be no way to unset a value...

I'm open to ideas here.

One thing I've been considering is a new interpolation function: 

```
read_var_from_file(VAR, FILE)
```

The idea is that it returns the value of a variable `VAR`  read from `.tfvars` file `FILE`. It could replace most use cases for `include`, as you could just explicitly pull in the values you want, but also allow much finer grained control:

```hcl
terragrunt = {
  lock = "${read_var_from_file("terragrunt.lock", find_in_parent_folders())}"
  remote_state = "${read_var_from_file("terragrunt.remote_state", find_in_parent_folders())}"
  terraform {
    source = "git::git@github.com:foo/bar.git//app?ref=v0.0.1"
    extra_arguments {
      global = "${read_var_from_file("terragrunt.terraform.extra_arguments", find_in_parent_folders())}"
    }
  }
}
```

This may also solve #132, as it will allow you to reuse values of arbitrary variables (not just Terragrunt ones) from your `.tfvars` files:

```
aws_region = "us-east-1"

terragrunt = {
  remote_state {
    backend = "s3"
[...]

Issue #132, opened 2017-02-17
Use ${var.} in terragrunt config


krjensen:
Hi,

I have a lot of environments that all havs it's own .tfvars file for setting up base infrastructure vars like:

I define my environment in a variable called `environment_name ` and i dont really want to define that more that one time, but my problem is that I want to reuse that variable in the terragrunt config as a path to my state in consul
```
environment_name = "dev"
environment_id = "01"
external_network = "external-vlan1"
user_data_linux = "../../Environments/global/user_data_linux_dev.yaml"
user_data_windows = "../../Environments/global/user_data_windows_dev.yaml"

terragrunt = {
  remote_state {
    backend = "consul"
    config {
	address = "my-consul.dev.local:8500"
        path = "terraform/state/**${var.environment_name }**"
	datacenter = "dus2-shared"
    }
  }
}
```

brikis98:
Sharing variables between Terragrunt and Terraform has come up before, but to be honest, I'm not sure we know the best way to handle it. The problem is that Terraform can load variables from many places:

1. `default` values in the variable definition.
1. `TF_VAR_XXX` environment variables.
1. `terraform.tfvars`, if it exists.
1. `-var` argument.
1. `-var-file` argument.

We could certainly read and parse all the variables in the `.tfvars` file you use to configure Terragrunt, but that would only give us access to a small subset of the variables Terraform can access. So `var.environment_name` might work, but `var.foo` would not if `foo` was set via some other means (e.g. env var or `default`). I worry that inconsistency would get very confusing.

If someone has ideas on a clean solution to this, I'm open to ideas.

JohnEmhoff:
I think I'm more interested in variables in the terragrunt file being available to terraform. The workaround is break them out and specify a couple`-var-file` arguments. At the very least, the documentation is a bit misleading here:

```Terraform also uses terraform.tfvars as a place where you can set values for your variables, but so long as your Terraform code doesn't define any variables named terragrunt, Terraform will safely ignore this value.```

This pretty strongly implies that non-terragrunt variables are passed to terraform, which doesn't seem to be the case in `include`-ed files.

brikis98:
> I think I'm more interested in variables in the terragrunt file being available to terraform.

[...]

PR #224, opened 2017-05-25
Found another cornercase

jocgir:
There was a bug in certain circumstance whit processSingleInterpolationInString.

The regex can consider multiple interpolations as only one. The main problem is that INTERPOLATION_SYNTAX_REGEX is too general. I did not change it to avoid undesired side effects. But I added a test in processSingleInterpolationInString to ensure that there is only one match between the quotes.

I also added a test that catch the misbehaviour.

Sorry for the inconvenience.

### DIFF
--- a/README.md
+++ b/README.md
@@ -993,6 +993,7 @@ Terragrunt allows you to use [Terraform interpolation syntax](https://www.terraf
 * [get_tfvars_dir()](#get_tfvars_dir)
 * [get_parent_tfvars_dir()](#get_parent_tfvars_dir)
 * [get_terraform_commands_that_need_vars()](#get_terraform_commands_that_need_vars)
+* [get_terraform_commands_that_need_input()](#get_terraform_commands_that_need_input)
 * [get_terraform_commands_that_need_locking()](#get_terraform_commands_that_need_locking)
 * [get_aws_account_id()](#get_aws_account_id)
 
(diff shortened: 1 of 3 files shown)


jocgir:
It could work. I thought about that, but the problem with that kind of exclusion in regex is that it may have some undesired effect. Like that:
```
key = "${get_env("SOME_VAR", "SOME{VALUE}")}:
```

Which is doubtful, but perfectly valid.

brikis98:
Ah, good point. Do we have a test case for that?

jocgir:
Not yet.

I am trying something on regex101 to address this point

https://regex101.com/r/wG3aM3/302

I must say that in our forked terragrunt version, we made several «enhancements» that are not ready to be submitted as PR. One of these is direct support for ${var.whatever} in terragrunt config. It would be nice if we can have a talk regarding those.

brikis98:
> I am trying something on regex101 to address this point https://regex101.com/r/MVuA1t/1

Heh, nice. That said, there are diminishing returns here. A small tweak to the regex to avoid extra if-statements is good, but if the fix requires a huge, hard-to-understand regex, then an if-statement is probably easier to maintain.

> One of them is direct support for ${var.whatever} in terragrunt config. It would be nice if we can have a talk regarding those.

Nice. We have a couple relevant discussions going on around this topic: 

https://github.com/gruntwork-io/terragrunt/issues/132
https://github.com/gruntwork-io/terragrunt/issues/147#issuecomment-282276693

brikis98:
@jocgir Let me know if you found a reasonable regex to fix this. If not, happy to merge the if-statement once we add the test case mentioned above.

jocgir:
Yes, I think it would be preferabable to improve the regex (bug segmenting it in smaller, more readable parts).

brikis98:
New regex looks promising! That said, it seems to still do a greedy capture, so is the reason that it works because you are now matching on word characters (`\w`) so that `}` will no longer match? Does it work correctly with the test case you mentioned earlier?

```
key = "${get_env("SOME_VAR", "SOME{VALUE}")}:
```

jocgir:
Hi, I just a a test to ensure it.

The \w+ just ensure that we got `some_function` in `${   some_function()   }`. We tolerate spaces around it, but not between the function name and the opening parenthesis. Just add tests for that too.

The tricky part is more in the parameters. There are still possible corner cases. Regex will never replace a real lexer/parser.

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.12.19

PR #226, opened 2017-05-25
Add "init" to Terraform commands that use state

rwiggins:
The Terraform docs explicitly state that the "init" command is used (in
part) to initialize Terraform backends. So, the backend config needs to
be initialized by Terragrunt for the init command.

This led to quite a bit of confusion for me, as Terragrunt "mysteriously"
wasn't configuring the remote state for "init".

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -30,6 +30,8 @@ const CMD_APPLY_ALL = "apply-all"
 const CMD_DESTROY_ALL = "destroy-all"
 const CMD_OUTPUT_ALL = "output-all"
 
+const CMD_INIT = "init"
+
 // CMD_SPIN_UP is deprecated.
 const CMD_SPIN_UP = "spin-up"
 
@@ -217,6 +219,14 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 		if err := configureRemoteState(conf.RemoteState, terragruntOptions); err != nil {
 			return err
 		}
+	}
+
+	// If the command is 'init', stop here. That's because ConfigureRemoteState above will have already called
+	// terraform init if it was necessary, and the below RunTerraformCommand would end up calling init without
+	// the correct remote state arguments, which confusing.
+	if terragruntOptions.TerraformCliArgs[0] == CMD_INIT {
+		terragruntOptions.Logger.Println("Running 'init' manually is not necessary: Terragrunt will call it when needed before running Terraform commands")
+		return nil
 	}
 
 	return shell.RunTerraformCommand(terragruntOptions, terragruntOptions.TerraformCliArgs...)
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -223,7 +223,7 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 
 	// If the command is 'init', stop here. That's because ConfigureRemoteState above will have already called
 	// terraform init if it was necessary, and the below RunTerraformCommand would end up calling init without
-	// the correct remote state arguments, which confusing.
+	// the correct remote state arguments, which is confusing.
 	if terragruntOptions.TerraformCliArgs[0] == CMD_INIT {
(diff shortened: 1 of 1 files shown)


brikis98:
Hmm, this is a bit odd, as we use the `init` command to initialize remote state... So I guess this would end up calling `init` twice, once with Terragrunt-provided params, and once without. I'm guessing the second call would actually fail.

What are you trying to do?

rwiggins:
Hmmm... I see now. The way this state configuration worked was not how I imagined it. Apologies for the confusion.

I was using Terraform local state and wanted to migrate to Terragrunt w/ Remote State. So, I set up the terraform.tfvars file with the appropriate terragrunt stanzas, then per the Terraform docs on migrating to a remote backend, decided to run `terragrunt init` (mirroring the documented suggestion of `terraform init`) to kick off the migration.

Terragrunt didn't initialize the remote state, so Terraform just kept asking for the bucket name and so on. This was pretty confusing for me, since it appeared to just "not work," until I went digging in the Terragrunt source and found the array that this PR modifies.

As a workaround in the meantime, I ended up just using `terragrunt refresh`, which did end up calling `terraform init` and making the state migration go. 

Now that I look further at how this state stuff is handled, it looks more like the behavior should be either: (1) prevent calling `terragrunt init` at all, or (2) catch the case of `terragrunt init` and handle it correctly (initialize the remote backend, but do nothing else). 

With the current code (in master), `terragrunt init` ends up calling `terraform init` directly without initializing the backend, which makes terraform prompt the user for the backend config.

With the modified code (this PR), it works like you guessed: `terragrunt init` initializes the backend by calling `terraform init` with the appropriate arguments, and then calls `terraform init` bare, which makes terraform prompt the user for the backend config. So, I think this PR is not merge-worthy at the moment.

brikis98:
Ah, yea, you're right, that's definitely confusing.

Well, at the least, we should update the docs to say that with Terragrunt, you don't need to run `init`, as Terragrunt automatically does it for you when you run other commands (e.g. `refresh`, `plan`, etc).

However, a PR to handle calls to `init` in code would be good. Probably the right thing to do is to tack on the proper remote state configs to the `init` command automatically for the user, and perhaps log something to indicate that's what we're doing.

rwiggins:
I pushed another commit to this branch that at least patches over the problem for now. The commit makes it so that the code will bail before calling `RunTerraformCommand` in the case of `init`. So, init basically makes Terragrunt call configureRemoteState, which *may* run `init` (or it may not). It also prints a message telling the user that we call `init` for them if necessary.

I'm not sure this is the best solution, but it's pretty quick and painless, so I thought I'd offer it up. A more complete solution would probably be reworking the configureRemoteState stuff to "force" a reconfigure, I guess. Up to you, though 🤷‍♂️

cli/cli_app.go
```diff
@@ -218,6 +221,14 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 		}
 	}
 
+	// If the command is 'init', stop here. That's because ConfigureRemoteState above will have already called
+	// terraform init if it was necessary, and the below RunTerraformCommand would end up calling init without
+	// the correct remote state arguments, which confusing.
```
brikis98: s/which confusing/which is confusing/

cli/cli_app.go
```diff
@@ -218,6 +221,14 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
...
+	if terragruntOptions.TerraformCliArgs[0] == CMD_INIT {
+		terragruntOptions.Logger.Println("Running 'init' manually is not necessary: Terragrunt will call it when needed before running Terraform commands")
```
brikis98: s/will call it when/will call it automatically when/
s/running Terraform commands/running other Terraform commands/

rwiggins:
Ah, thanks for the wording recs. Pushed a couple commits with the suggested changes.

brikis98:
LGTM, thank you!

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.12.20

PR #232, opened 2017-06-13
Always use the Logger rather than writing directly to stdout

brikis98:
Fix a couple places in Terragrunt where we were writing directly to `stdout` instead of using a `Logger`. Everything in Terragrunt should go through the `Logger`, especially as we’ve configured Terragrunt to log everything to `stderr` so that it’s output does not interfere with Terraform’s output (e.g. so it doesn’t interfere with the output of the `terraform output` command).


brikis98:
Simple change, so merging. Feedback welcome!

josh-padnick:
LGTM.

Issue #230, opened 2017-06-09
Remove need for "backend" block in Terraform modules.


josh-padnick:
One of the most common issues that Terragrunt users experience is that the "topmost" module that is called from your `terraform.tfvars` file is expected to have a `backend` block:

**terraform.tfvars:**

```hcl
terragrunt = {
  terraform {
    source = "/my-module/"
  }
}

key1 = val1
key2 = val2
```

**/my-module/main.tf:**

```hcl
terraform {
  # The offending block. The configuration for this backend will be filled in by Terragrunt.
  backend "s3" {}
}

variable "key1" {}
variable "key2" {}
...
```

In theory, a module should be unopinionated about which backend it implements and the calling code should express this opinion. In practice, Terragrunt works by copying the module to a temp directory and filling in the vars automatically, so it needs a `backend` block to "hook into".

An alternative way to implement this is to auto-generate a boilerplate Terraform template that calls the original module, but @brikis98 had a concern that we'd be forever playing catch-up to frequently changing Terraform code.

I don't currently agree with that argument, but it's still non-trivial to update Terragrunt to auto-generate new code, so for now, I just wanted to record the issue and let users know this is just about the most common issue we encounter with users.

conorgil:
@josh-padnick Glad you created an issue for this, as I am running into it while creating some example configurations for #169. The discussion came up in that thread in [this comment](https://github.com/gruntwork-io/terragrunt/issues/169#issuecomment-30448767), where @brikis98 suggested exposing variables that you can use to change the configuration of the S3 backend.

However, to make sure I follow your suggestion, you're saying that the larger problem is that `my-module/main.tf` defines a backend at all, right? If someone wanted to use the Atlas backend, for example, then they would not be able to use that module because it hard codes the S3 backend inline. I am running into this problem during local development of my Terraform modules because I want to use a local backend during dev (working in Terraform only land) and then the S3 backend when running in AWS (Terragrunt land). Since I am new to the tool, my workflow might be off, so feel free to suggest general improvements there.

Can you explain more your idea to "auto-generate a boilerplate Terraform template that calls the original module"?

Here are my thoughts on a potential solution (not sure if it is the same as your idea or not). Basically, I think that because the root `terraform.tfvars` file already defines the backend configuration, Terragrunt can use that configuration for all child modules automatically by generating a `backend.tf` file in the temporary directory in which Terraform actually runs. For example:

**live/terraform.tfvars**:

```hcl
terragrunt = {
[...]

Issue #169, opened 2017-04-21
Include sample "best practices Terragrunt configuration"


josh-padnick:
Even I find myself glazing over a little by the bottom of the README because there are a lot of features here. Also, after reading everything, the Terragrunt user still has to make one more leap on their own: What is the ideal, complete setup with Terragrunt that combines all these use cases?

For that reason, it'd be nice if we had a `_sample` folder with maybe `live` and `modules` subfolders which mimicked a real setup and showed exactly what the root `terraform.tfvars` file and each "child" `terraform.tfvars` file looks like.

Or alternatively, just include those two `terraform.tfvars` files (root and child) as examples.

tomstockton:
Hello,

I expect that this is probably low priority but I'd appreciate some rough guidance around how to organise / write the terraform.tfvars files, specifically on creating DRY remote state configuration.  The current README.md suggests:

* Creating parent and child `terraform.tfvars` files in the `modules` repo/folder which defines remote state config.
* Creating a `terraform.tfvars` file in the `live` repo/folder for each component in each environment ie. `prod/app/terraform.tfvars` which defines the source (and any overriding vars) of the appropriate module.

I'm finding that this fails to configure the remote state as after the initial `init` to download the source repo into a tmp dir, terragrunt then copies over the working dir (of the live repo) into the tmp dir and clobbers the terraform.tfvars file (that includes the remote state config).

I've tried putting the remote state (parent and child) `terraform.tfvars` files into the live repo and this works as I would expect so I'm wondering if the current README is incorrect?

Any advice much appreciated!

Thanks

Tom

brikis98:
@tomstockton All the `.tfvars` files should be in the live repo. None of them should be in the modules repo.

dpetzel:
I've also been struggling with this. Conceptually I think I grok all the individual examples, but when i try to combine some strategies I keep hitting some issues. 

Here is the combined structure I believe should be in place:

```
└── live
    ├── terraform.tfvars
    │
    ├── app1
    │   ├── dev
    │   │   ├── main.tf
    │   │   └── terraform.tfvars
    │   ├── qa
    │   │   ├── main.tf
    │   │   └── terraform.tfvars
    │   └── prod
    │       ├── main.tf
    │       └── terraform.tfvars
    └── app2
        ├── dev
        │   ├── main.tf
        │   └── terraform.tfvars
        ├── qa
        │   ├── main.tf
        │   └── terraform.tfvars
        └── prod
[...]

PR #234, opened 2017-06-15
WIP: Example infra using one account and one region

conorgil:
**WIP! Opened for discussion purposes. This is not ready to merge.**

There is still a lot of work and cleanup to be done on this example, but I think it is in a state that people can run it and provide some feedback. I want to get some feedback from the community to see if I am on the right track for creating something that people will find useful.

Is the level of detail in this example too much? Too little? Confusing? Helpful? Any and all feedback is appreciated so I can take it into account as I work on this and other examples.

*Before merging eventually, the Terraform modules can be recreated in repos owned by Gruntworks instead of me.*

### DIFF
--- a/_examples/01_one_account_one_region/README.md
+++ b/_examples/01_one_account_one_region/README.md
@@ -31,8 +31,54 @@ $> terragrunt apply-all
 ```
 
 ## Directory structure
-TODO: move the discussion of directory hierarchy and definitions out of the root
-`terraform.tfvars` file and put it here instead?
+Following best practices, the following directory structure contains all of our Terragrunt
+configuration:
+
+```
+live                                
+├── production                      <- environment directory
+│   ├── frontend-app                <- component directory
+│   |   └── terraform.tfvars        <- component configuration
+│   └── vpc
+│       └── terraform.tfvars
+├── staging                         <- environment directory
+│   ├── frontend-app                <- component directory
+│   |   └── terraform.tfvars        <- component configuration
+│   └── vpc
+│       └── terraform.tfvars
+└── terraform.tfvars                <- root configuration
+```
+
+The `live/` directory is the root of our Terragrunt configuration hierarchy.
+It contains a single `terraform.tfvars` file, which we will call
+the **root configuration**. The root configuration is optional, but allows 
+us to define general configuration that applies to all components in our infrastructure
+in a single place.
+
+The root contains a child directory for each environment in the infrastructure
+(e.g. production, staging), which we will call **environment directories**.
+The directory structure shown above contains two environment directories:
+`production` and `staging`.
+Environment directories do not normally contain any files. Instead, they are
+the location where Terragrunt commands such as `plan-all` and `apply-all`
+are executed to instantiate all of the infrastructure components for the
+given environment.
+
+Environment directories contain a child directory for each component of 
+the infrastructure, which we will call **component directories**.
+For example, a component might be a single app, a single MySQL DB, or a single VPC.
+The directory structure shown above contains 4 component directories: `frontend-app` and
+`vpc` in each environment.
+
+Each component directory contains a single `terraform.tfvars` file, which
+defines the configuration for that specific component. We will call this
+`terraform.tfvars` file the **component configuration**. The directory
+structure shown above contains 4 component configurations: one for
+each component in each environment.
+In a component configuration, we can use the `include` stanza to instruct 
+Terragrunt to locate and copy configuration from the root configuration file. 
+This allows us to keep our configurations DRY. Review the example
+code in this repo to see how this works in practice.
 
 ## Each infrastructure component is defined in a Terraform module
 In this example, our infrastructure contains two components: a VPC and a frontend_app.
--- a/_examples/01_one_account_one_region/README.md
+++ b/_examples/01_one_account_one_region/README.md
@@ -89,21 +89,25 @@ and
 respectively.
 
 There are a few important things to note about the Terraform modules:
-1. The code in each of these repos is regular plain old Terraform code. These
-   Terraform modules can be used without modification in any project that does
+1. The code in each of these repos is 100% regular plain old Terraform code. These
+   Terraform modules can be used, without modification, in any project that does
    not use Terragrunt and instead uses Terarform directly.
 1. Each module defines a variable for any value that might change between invocations
    of the module. For example, values that might change between production and staging
    environments.
+1. Each module enables a backend for storing remote state. The backend intentionally
+   uses a [Partial Configuration](https://www.terraform.io/docs/backends/config.html#partial-configuration)
+   to keep the Terraform modules DRY. Terragrunt will use the `remote_state` stanza in
+   our component configurations to fully configure the backend at runtime.
 1. There are no `terraform.tfvars` files in the Terraform module repos. The Terraform
    modules are generic infrastructure templates and do not define values for any
-   variables in `*.tfvar` files. The values for variables are defined in a separate
-   repo where Terragrunt conifguration also lives. In this example, you can see that
-   the values for variables are defined in this repo in the `terraform.tfvars` file
-   in each component directory.
+   variables in `*.tfvar` files. The values for variables are defined in the same repo
+   as our Terragrunt conifguration. In this example, you can see that
+   the values for variables are defined in the component configuration
+   (`terraform.tfvars`) files in each component directory.
 
-Each component of our infrastructure will be defined in its own Terraform module
-in the same way. This provides several benefits:
+As our infrastructure grows and we add components, each will be defined similarly in its own
+Terraform module. This provides several benefits:
 1. Versioning
    1. Hosting each component in its own repo means that each component can be indepdently
       versioned, which allows each environment in the infrastructure to use a different
--- a/_examples/01_one_account_one_region/README.md
+++ b/_examples/01_one_account_one_region/README.md
@@ -1,0 +1,91 @@
+# WIP
+
+This example demonstrates several basic Terragrunt best practices by
+creating a simple infrastructure with two environments (production and staging),
+which is hosted in a single AWS account in a single AWS region.
+
+**Demonstrated concepts and keywords:**
+
+* Recommended Terragrunt directory structure
+* Basic Terragrunt configuration in `terraform.tfvars` files
+* Keeping Terragrunt configurations DRY by using the `include` keyword and 
+  the `find_in_parent_folders()` Terragrunt built-in function
+* Using remotely sourced and independently versioned Terraform modules
+* Independent remote state for each individual infrastructure component
+* Defining dependencies between infrastructure components using the Terragrunt 
+  `dependencies` keyword
+* Sharing remote state between infrastructure components using `terraform_remote_state`
+
+## Directory structure
+TODO: move the discussion of directory hierarchy and definitions out of the root
+`terraform.tfvars` file and put it here instead?
+
+## Each infrastructure component is defined in a Terraform module
+In this example, our infrastructure contains two components: a VPC and a frontend_app.
+Each of these components is defined in its own Terraform module:
+[conorgil/terragrunt_sample_vpc](https://github.com/conorgil/terragrunt_sample_vpc)
+and
+[conorgil/terragrunt_sample_frontend_app](https://github.com/conorgil/terragrunt_sample_frontend_app)
+respectively.
+
+There are a few important things to note about the Terraform modules:
+1. The code in each of these repos is regular plain old Terraform code. These
+   Terraform modules can be used without modification in any project that does
+   not use Terragrunt and instead uses Terarform directly.
+1. Each module defines a variable for any value that might change between invocations
+   of the module. For example, values that might change between production and staging
+   environments.
+1. There are no `terraform.tfvars` files in the Terraform module repos. The Terraform
+   modules are generic infrastructure templates and do not define values for any
+   variables in `*.tfvar` files. The values for variables are defined in a separate
+   repo where Terragrunt conifguration also lives. In this example, you can see that
+   the values for variables are defined in this repo in the `terraform.tfvars` file
+   in each component directory.
+
+Each component of our infrastructure will be defined in its own Terraform module
+in the same way. This provides several benefits:
+1. Versioning
+   1. Hosting each component in its own repo means that each component can be indepdently
+      versioned, which allows each environment in the infrastructure to use a different
+      version of the Terraform module. This allows for straightforward development and
+      testing of Terraform module changes in QA and staging environments before applying
+      those changes to production.
+     
+      In this example, look at the `source` stanza of the `terraform` block in any of the
+      component configuration files. Notice that each specifies a specific version of the
+      Terraform module source by using the `?ref=<some version>` syntax.
+2. Independent remote state
+   1. Typically, a Terraform project will have a single top level `main.tf` file
+      for each environment, which calls a Terraform module for each component of
+      the infrastructure. As a result of having a single top level `main.tf` file,
+      the entire environment has a single remote state file.
+   
+      Hosting each Terraform module in its own repo makes it possible to setup the directory
+      structure such that each component within an environment has its own remote state file.
+      In this example, each component is defined in its own directory. This means that when
+      Terraform is run in each directory, a remote state file will be created for only the
+      infrastructure defined within that directory; a single component within the environment.
+      
+      Having a remote state file for each component makes it easier to make changes to a single
+      component; commands like `plan` and `apply` will be quicker because they only have to refresh
+      the state of the single component, not the entire environment. Also, each state file is smaller
+      in scope, which means that we have a smaller blast radius if a bug is introduced into one of
+      our components. It is more likely to break a single component instead of the entire environment.
+      
+## Dependencies between infrastructure components
+It might be best to come up with some contrived example since this is the first one and *NOT* introduce
+the `dependencies` keyword yet. Save it for a future example so that we can explain it appropriately?
+
+Some things to discuss with the `dependencies` keyword:
+* makes `apply-all` and `destroy-all` work as expected
+* issues with `plan-all`
+
+If we do leave this in the first example, this section would also explain that the dependency
+between the two modules is because the
+[conorgil/terragrunt_sample_frontend_app](https://github.com/conorgil/terragrunt_sample_frontend_app)
+Terraofrm module needs to pull from the 
+[conorgil/terragrunt_sample_vpc](https://github.com/conorgil/terragrunt_sample_vpc)
+module which subnet to use for the EC2 instance. Highlight that the 
+[conorgil/terragrunt_sample_frontend_app](https://github.com/conorgil/terragrunt_sample_frontend_app)
+module accomplishes this by using a `terraform_remote_state` data source and exposes the configuration
+for that resource as variables.
--- a/_examples/01_one_account_one_region/README.md
+++ b/_examples/01_one_account_one_region/README.md
@@ -16,6 +16,20 @@ which is hosted in a single AWS account in a single AWS region.
   `dependencies` keyword
 * Sharing remote state between infrastructure components using `terraform_remote_state`
 
+## Running the example
+To create the staging environment:
+```
+$> cd live/staging
+$> terragrunt apply-all
+```
+
+
+To create the production environment:
+```
+$> cd live/production
+$> terragrunt apply-all
+```
+
 ## Directory structure
 TODO: move the discussion of directory hierarchy and definitions out of the root
 `terraform.tfvars` file and put it here instead?
--- a/_examples/01_one_account_one_region/live/production/frontend_app/terraform.tfvars
+++ b/_examples/01_one_account_one_region/live/production/frontend_app/terraform.tfvars
@@ -1,0 +1,33 @@
+terragrunt = {
+  # This include stanza utilizes the find_in_parent_folders() Terragrunt built-in,
+  # which instructs Terragrunt to locate the root terraform.tfvars
+  # file and copy that configuration. This allows us to remain DRY and prevents
+  # the need to copy/paste general configuration, such as remote state settings.
+  include {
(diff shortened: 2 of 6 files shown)


conorgil:
@brikis98 In this example, I recommend maintaining each Terraform module in its own repo. In #169, it looks like you recommend using a single repo called `modules` that has a subdir for each Terraform module. Thoughts on the pros/cons of each approach?

Benefits to having each module in its own repo:
* It is easier to reason about version numbers throughout the infrastructure. If each module has its own version, then you can easily identify which environments are using different versions of a given module. If all modules are defined in a single repo, then you could have prod using `v0.1.1` and staging using `v0.1.2` and you're not actually sure whether `modules/A` changed between those two versions. It could have easily been a change applied to `modules/B` that caused the version number for the repo to jump. For me, this reason alone is enough motivation to host each in its own repo.
* Improved maintainability and usability
  * the README.md for each module can reside at the root of the repo and can clearly document what the module does, the variables, etc
  * the CHANGELOG.md for each module can reside at the root of the repo and clearly document important changes over time, etc
  * Logical place to add automated tests to make sure that the module works as expected

Drawbacks to having each in its own module:
* there are more repos to manage, more admin overhead
* there can be dependencies between Terraform modules and maintaining them in separate repos could make it more difficult to keep them in sync

mcalhoun:
> In this example, I recommend maintaining each Terraform module in its own repo

@conorgil I also agree with this approach. The one additional benefit it brings occurs during the development of complex packages. The automated test cycle runs on a CI platform can become very long when each module's tests need to complete with each new commit. Having a module isolated could significantly lower the "wait for tests to run" time.

_examples/01_one_account_one_region/README.md
```diff
@@ -0,0 +1,155 @@
...
+the **root configuration**. The root configuration is optional, but allows 
+us to define general configuration that applies to all components in our infrastructure
+in a single place.
+
+The root contains a child directory for each environment in the infrastructure
+(e.g. production, staging), which we will call **environment directories**.
+The directory structure shown above contains two environment directories:
+`production` and `staging`.
+Environment directories do not normally contain any files. Instead, they are
+the location where Terragrunt commands such as `plan-all` and `apply-all`
```
dpetzel: Then assuming you're in the `production` directory doesn't this ignore the **root configuration**? I didn't think terragrunt would traverse upward past your current location.

_examples/01_one_account_one_region/README.md
```diff
@@ -0,0 +1,155 @@
...
+   modules are generic infrastructure templates and do not define values for any
+   variables in `*.tfvar` files. The values for variables are defined in the same repo
+   as our Terragrunt conifguration. In this example, you can see that
+   the values for variables are defined in the component configuration
+   (`terraform.tfvars`) files in each component directory.
+
+As our infrastructure grows and we add components, each will be defined similarly in its own
+Terraform module. This provides several benefits:
+1. Versioning
+   1. Hosting each component in its own repo means that each component can be indepdently
```
dpetzel: `indepdently` -> `independently`

_examples/01_one_account_one_region/README.md
```diff
@@ -0,0 +1,155 @@
...
+respectively.
+
+There are a few important things to note about the Terraform modules:
+1. The code in each of these repos is 100% regular plain old Terraform code. These
+   Terraform modules can be used, without modification, in any project that does
+   not use Terragrunt and instead uses Terarform directly.
+1. Each module defines a variable for any value that might change between invocations
+   of the module. For example, values that might change between production and staging
+   environments.
+1. Each module enables a backend for storing remote state. The backend intentionally
```
dpetzel: Does this now impose the requirement of remote state inside the module? IE what if someone has modules they don't want to use remote state for? I can't help but feel like this sort of decision belongs in the in the component configuration, not the module.

_examples/01_one_account_one_region/README.md
```diff
@@ -0,0 +1,155 @@
...
+      infrastructure defined within that directory; a single component within the environment.
+      
+      Having a remote state file for each component makes it easier to make changes to a single
+      component; commands like `plan` and `apply` will be quicker because they only have to refresh
+      the state of the single component, not the entire environment. Also, each state file is smaller
+      in scope, which means that we have a smaller blast radius if a bug is introduced into one of
+      our components. It is more likely to break a single component instead of the entire environment.
+      
+## Dependencies between infrastructure components
+It might be best to come up with some contrived example since this is the first one and *NOT* introduce
```
dpetzel: I think its common enough and simple enough it could fit well here.

_examples/01_one_account_one_region/README.md
```diff
@@ -0,0 +1,155 @@
...
+the `dependencies` keyword yet. Save it for a future example so that we can explain it appropriately?
+
+Some things to discuss with the `dependencies` keyword:
+* makes `apply-all` and `destroy-all` work as expected
+* issues with `plan-all`
+
+If we do leave this in the first example, this section would also explain that the dependency
+between the two modules is because the
+[conorgil/terragrunt_sample_frontend_app](https://github.com/conorgil/terragrunt_sample_frontend_app)
+Terraofrm module needs to pull from the
```
dpetzel: `Terraofrm` -> `Terraform`

_examples/01_one_account_one_region/README.md
```diff
@@ -0,0 +1,155 @@
...
+* issues with `plan-all`
+
+If we do leave this in the first example, this section would also explain that the dependency
+between the two modules is because the
+[conorgil/terragrunt_sample_frontend_app](https://github.com/conorgil/terragrunt_sample_frontend_app)
+Terraofrm module needs to pull from the 
+[conorgil/terragrunt_sample_vpc](https://github.com/conorgil/terragrunt_sample_vpc)
+module which subnet to use for the EC2 instance. Highlight that the 
+[conorgil/terragrunt_sample_frontend_app](https://github.com/conorgil/terragrunt_sample_frontend_app)
+module accomplishes this by using a `terraform_remote_state` data source and exposes the configuration
```
dpetzel: I know in my case I make use of dependencies, but not a lot of `terraform_remote_state`. To your earlier point would the example be simpler if the front_end app learned of the subnet via the `aws_subnet` datasource. For me personally I understood this concept easier than extracting data from `terraform_remote_state`, but this one is probably largely dependent on the user.

dpetzel:
This is really great (wishing I had it when I first got started). I'm wondering about having different structures for one account/1 region and multiple account/multiple region, versus just laying out a structure that works for both right from the beginning? Jumping to multiple account and multiple region was a challenge for me for sure.

conorgil:
@mcalhoun Great point about CI/CD running tests more quickly.

@dpetzel thanks for all of your comments! I'll address each inline. 

Regarding your point about accounts and regions, I was thinking about this same thing today. I think that it would make sense for someone just starting out with Terragrunt to actually set up their directory structure such that they could easily add additional regions or accounts in the future even if they only start with a single one in the beginning. I have not personally had to set those situations up yet, but I'll be playing with them over the next few weeks so that I can write a good tutorial on them. Any pointers or tips you can share about your experience would be super useful!

_examples/01_one_account_one_region/README.md
conorgil: @dpetzel Terragrunt will traverse upwards beyond the directory you execute Terragrunt from to find the parent. I believe that it will traverse all the way up the tree until it finds a `terraform.tfvars` file, actually. Let me know if you have experienced otherwise and we'll sort it out in the example.

_examples/01_one_account_one_region/README.md
conorgil: Good catch. Fixed.

_examples/01_one_account_one_region/README.md
conorgil: Good catch. Fixed.

_examples/01_one_account_one_region/README.md
conorgil: @dpetzel This is a great question. The short answer is yes. As I have written the example, the remote state is defined inside the Terraform module. I decided to write the example as I did because of [this discussion](https://github.com/gruntwork-io/terragrunt/issues/169#issuecomment-299199548) in #169.

The long answer is that I think this is an open question that is being discussed as to what the "best practice" (see #230). 

As I was writing the example, it did feel "wrong" to hardcode some things into the Terraform modules; specifically, the `backend` and `terraform_remote_state`. I think that it felt "wrong" because I want to apply the dependency injection pattern to Terraform modules and hardcoding a `backend` and `terraform_remote_state` violates that pattern.

I'll ramble about `backend` and leave my thoughts on `terraform_remote_state` for another issue/time. As you said, if you define a backend in a module, then that module cannot be used by another part of your team with a different backend; they would have to change the source code to change the backend. When using plain old Terraform, the backend is defined once in the `main.tf` file. One or more Terraform modules are called using the classic syntax and passing in any values they might need as variables. The Terraform modules do not define any backend because it was defined in the `main.tf` file already and there is only a single state file. With Terragrunt, we want each component to maintain its own state file, which means that each component is defined in its own directory and also needs to configure its own backend. As you point out, this could be done in a `backend.tf` file file along side the component configuration instead of in the Terraform module itself. This would certainly make the Terraform module more generic and follow the dependency injection pattern more closely since other teams could simply define their own `backend.tf` file and the module would work without change. There is some discussion in #230 about challenges supporting this feature natively in Terragrunt, but if you were to just follow this pattern manually, then I think it should work fine. There would be more files to maintain in the Terragrunt configuration repo, but it could be worth it if generic reusable Terraform modules are important for your team. If you try this (or already do this) let me know how it works out!

I think the counter argument is one that you actually highlighted in [this comment](https://github.com/gruntwork-io/terragrunt/issues/169#issuecomment-300942218).
> your not building generic multi-use type of modules, but rather modules that are implementing a very specific configuration, unique to your scenario.

If the module is unique to each scenario, then reusability is not the main goal and the additional overhead of maintaining `backend.tf` files is likely not worth it if all groups in the team use the same backend type anyways.

_examples/01_one_account_one_region/README.md
conorgil: That is a super interesting idea. I literally did not even think about using data sources instead of `terraform_remote_state`.

The team that I was working with started using Terraform a few years ago before data sources were introduced and we never updated our configurations to use them because of the classic "everything works and we'll update it later" logic.

I agree that both approaches are functionally equivalent and come down to user preference. I will leave them as `terraform_remote_state` for now since that is what I am familiar with.

brikis98:
@conorgil This is great work, thank you!

Some thoughts:

1. It would be great to add automated tests to these examples. However, deploying real AWS resources (e.g. EC2 Instances, RDS, etc) takes a while, and has to be carefully cleaned up. Perhaps we keep the folder-structure as-is, but change the contents of each example to `template_file` data sources or other Terraform elements that are purely local and very quick to set up. The comments in the code would explain why we're doing this, and the various modules can still depend on each other (e.g. you could create an output variable with the rendered text of one of the `template_file` values in one module and pull in that rendered text in another module using `terraform_remote_state`), but we'd also be able to add fast-running automated tests on top of these examples to ensure they stay up to date over time. The `integration_test.go` folder has similar examples.

1. Keeping your modules in separate repos is completely fine. To get versioning, they should be defined in *at least* one separate repo (e.g. `live` and `modules`), but having multiple separate repo has plenty of advantages, as you and @mcalhoun mentioned.

1. It would be great to have an example that's a bit simpler as a starting point. It would have several environments (stage, prod), but each environment would contain a single `terraform.tfvars` file. This works well for small deployments, as there are fewer files, less code to manage, no need to use `terraform_remote_state`, easier testing, and so on. Of course, it has the downside of making it possible to break the entire environment with any mistake, but in some use cases, that's an acceptable trade-off, and we should call that out explicitly in the docs.

conorgil:
@brikis98 

1. Great idea to add automated tests to the examples. I have not written automated tests for Terraform code before. Do you have any guidance for the types of tests that we could write? Would you just write the tests in Go, or are you aware of any tools to help write IAC or Terraform specific tests? I looked at [test/integration_test.go](https://github.com/gruntwork-io/terragrunt/blob/master/test/integration_test.go) for some ideas. We could write a simple test to execute `terragrunt apply-all` and `terragrunt destroy-all` and make sure that each executes without error. Thoughts?

2. Agreed.

3. I agree that a simpler example might be a better starting point, but I think that might be too simple. Few questions to make sure I'm on the same page, though. Would each env have a single `main.tf` file and a `terraform.tfvars` file, or would the env only have a `terraform.tfvars` file and use remote source? If using remote source, what would the remote source look like? If using local source, then what do you imagine the configuration in `terraform.tfvars` containing?

    I think the first tutorial really needs to highlight the value of Terragrunt and which pain points it solves to make it clear why someone would want to use it. Given, we've all had different experiences with Terraform, so the content of that tutorial might vary on preference and experience.

I have been working on a presentation that I am going to give to a Terraform meetup group in Berlin in early July. My approach is to walk through a series of steps to take a Terraform project from "very disorganized" to "best practice" as I understand it to be. Each step will make a single change to the project and I will discuss the motivation for the change. As things progress, certain pain points will emerge that someone might think "that is going to a PITA with Terraform even though I see the value we're hoping to get with this change". That is when I will introduce Terragrunt and show how it can solve those pain points so that we can have our cake and eat it too.

Thoughts on that general approach for explaining the benefit of Terragrunt? We could modle the examples in this repo in a similar fashion if I get positive feedback from the meetup group and folks in this thread.

brikis98:
>  I looked at test/integration_test.go for some ideas. We could write a simple test to execute terragrunt apply-all and terragrunt destroy-all and make sure that each executes without error. Thoughts?

Yes, exactly. 

> Would each env have a single main.tf file and a terraform.tfvars file, or would the env only have a terraform.tfvars file and use remote source? If using remote source, what would the remote source look like? If using local source, then what do you imagine the configuration in terraform.tfvars containing?

In the `live` repo, you'd just have a single `terraform.tfvars` file per env. In the `modules` repo, you'd have a single `main.tf` (or a single folder of Terraform files) defining the entire environment. It's a fairly common set up for users who don't want to split up components into separate modules and use `terraform_remote_state` to connect them. 

That said, it's definitely not a blocker for this PR, but merely a suggestion for other examples we could add in the future.

> My approach is to walk through a series of steps to take a Terraform project from "very disorganized" to "best practice" as I understand it to be. 

That sounds like a great talk!

> Thoughts on that general approach for explaining the benefit of Terragrunt?

I think your approach sounds like a great way to do it. Show the pain, show the solution, and discuss trade-offs. Please share the link with us if that talk is available online in the future :)

conorgil:
@brikis98 
> In the live repo, you'd just have a single terraform.tfvars file per env. In the modules repo, you'd have a single main.tf (or a single folder of Terraform files) defining the entire environment. It's a fairly common set up for users who don't want to split up components into separate modules and use terraform_remote_state to connect them.

I have been thinking a lot about this scenario recently and I am confident that I am thinking myself in circles and could use some outside perspective. Can you help me grok why someone would want to use Terragrunt in this scenario instead of plain Terraform?

With Terragrunt, it would look like:
```
live
├─ terraform.tfvars         # Defines the configuration patterns for the S3 backend
├─production
│   └─ terraform.tfvars     # Has source="git::modules.git?ref=v0.1.1"
│                           # Defines the input var values for production.
│                           # Includes partial config of S3 backend.
└─staging
       └─ terraform.tfvars  # Has source="git::modules.git?ref=v0.2.0"
                            # Defines the input var values for staging.
                            # Includes partial config of S3 backend.
```

With Terraform, it would look like:
```
live
├─production
│   └─ terraform.tfvars     # Has source="git::modules.git?ref=v0.1.1"
│                           # Defines the input var values for production.
│                           # Defines and configures the S3 backend for production.
└─staging
       └─ terraform.tfvars  # Has source="git::modules.git?ref=v0.2.0"
                            # Defines the input var values for staging.
                            # Defines and configures the S3 backend for staging.
```

Both approaches have the same user workflow (a single `terraform/grunt plan` and `terraform/grunt apply` per env), both use versioned modules, and both store a single state file in S3 per env.

In this use-case, is the main motivation for using Terragrunt instead of Terraform to remain DRY and abstract the configuration of the S3 backend configuration in each env? I can see this providing more and more value each time an env is added, but seems like a small win with only two envs. Are there any other benefits of Terragrunt in this use-case that I'm over looking?

Generally, it seems to me that Terragrunt is *way* more valuable when working with a state file per infrastructure component than state file per env. However, I am still forming a concise argument for why that structure is a good idea in the first place. I'd love input from others as I continue to think on it!

brikis98:
> With Terraform, it would look like:

The purely Terraform example you list isn't supported by Terraform natively, at all. There is no support for `source = ...` in `.tfvars` files, so you'd have to have, at the bare minimum, a `.tf` file that calls out the module you want, sets the `source` URL in that module, and sets all the parameters within it. In practice, you also need to define a `provider`, and expose input variables, and add output variables, so instead of a single `.tfvars` or single `.tf` file, you have several, with a non-trivial amount of code. And all of that has to be copy/pasted into each environment.

conorgil:
@brikis98 Yup, the Terraform example in my previous comment is totally broken. Thanks for pointing that out. I copy/pasted and guess I didn't really read over things carefully after changing the comments in the example.

I'm still unclear why one would choose to use Terragrunt over Terraform in this simple scenario (two envs, single state file for each). In both cases, you have almost the same amount of code, just in different places.

In the Terraform setup, the local code is heavy and the remote module is lean. The main.tf file sets up the provider, the backend, and calls the remote module using the `module { source = ... }` syntax; the variables.tf file defines the input variables; and the output.tf file defines the output values. The remote module defines the resources, the input variables, and the output values.

In the Terragrunt setup, the local code is slightly lighter and the remote module is slightly heavier. The `terrform.tfvars` file now is responsible for calling the remote module using Terragrunt `source = ...` syntax and defining the input variable values. The remote module would define the provider (exposing the config as variables), define the backend partial, define input variables, and define output values. 

Yes, there is some copy/paste between environments in the Terraform approach, but I think that is a small price to pay compared to the overhead of introducing another tool into the development stack for such a simple use-case (single state file per env). I see some value using Terragrunt to remain as DRY as possible if the number of environments gets much larger, but in a small example like this it seems like overkill to me. Of course, Terragrunt would obviously work in this scenario and if someone thinks the benefits of being completely DRY outweigh the tradeoff of adding a new tool to the stack, then power to em! I'm just trying to understand the motivations for the use-case and wanted to see if I was missing some part of the bigger picture.

brikis98:
> I'm still unclear why one would choose to use Terragrunt over Terraform in this simple scenario (two envs, single state file for each). In both cases, you have almost the same amount of code, just in different places.

It's not the same amount of code. Here's an example.

Say you have some Terraform code in `infra-modules/my-infra/main.tf` that defines a way to deploy your entire infrastructure, including auto scaling groups, load balancers, security groups, IAM roles, etc:

```hcl
resource "aws_autoscaling_group" "example" {
  # ...
}

resource "aws_launch_configuration" "example" {
  # ...
}

resource "aws_alb" "example" {
  # ...
}
```

This module takes in a bunch of input variables in `infra-modules/my-infra/vars.tf`:

```hcl
variable "aws_region" {
  description = "The AWS region to deploy into (e.g. us-east-1)"
}

variable "aws_account_id" {
  description = "The ID of the AWS account to deploy into"
}

variable "min_size" {
  description = "The minimum number of instances of the microservice to run in the ASG"
}

variable "max_size" {
  description = "The maximum number of instances of the microservice to run in the ASG"
}

variable "instance_type" {
  description = "The type of instance to run in the ASG"
}

# ...
```

It may also have a bunch of output variables in `infra-modules/my-infra/outputs.tf`:

```hcl
output "alb_dns_name" {
  value = "${aws_alb.example.dns_name}"
}

output "asg_name" {
  value = "${aws_autoscaling_group.example.name}"
}

output "iam_role_id" {
  value = "${aws_iam_role.example.id}"
}

# ...
```

Now, to use this module in infra-live with pure Terraform, you would typically need the following code in `infra-live/stage/main.tf`:

```hcl
module "stage" {
  source = "../infra-live/my-infra"
  
  aws_region = "us-east-1"
  aws_account_id = "1234567890123"

  min_size = 3
  max_size = 5
  instance_type = "t2.micro"

  # ...
}
```

To keep your code DRY and to allow some of those params to be set dynamically (e.g. via env vars, `.tfvars` files, cmd-line params, etc), you'll need to expose some of those params as input variables in `infra-live/stage/vars.tf`:

```hcl
variable "aws_region" {
  description = "The AWS region to deploy into (e.g. us-east-1)"
}

variable "aws_account_id" {
  description = "The ID of the AWS account to deploy into"
}

# ...
```

And you have to mirror just about all of the output variables from the module in `infra-live/stage/outputs.tf` so they are displayed to users and stored in Terraform state:

```hcl
output "alb_dns_name" {
  value = "${module.stage.alb_dns_name}"
}

output "asg_name" {
  value = "${module.stage.asg_name}"
}

output "iam_role_id" {
  value = "${module.stage.iam_role_id}"
}

# ...
```

Now you have to copy/paste all three files into each environment: `infra-live/prod`, `infra-live/qa`, etc. 

Compare that with the Terragrunt approach, where the code in `infra-modules` is more or less the same, but the code in infra-live is reduced to a single `infra-live/stage/terraform.tfvars` file:

```hcl
terragrunt = {
  terraform {
    source = "../infra-live/my-infra"
  }
}

aws_region = "us-east-1"
aws_account_id = "1234567890123"

min_size = 3
max_size = 5
instance_type = "t2.micro"
```

The key idea is there is no `vars.tf` or `outputs.tf` to copy/paste and maintain infra-live. With a tiny bit of code and 1-2 environments, it's not a huge issue. But as the amount of infrastructure grows--as it usually does--and as you add more input and output variables, and as the number of environment grows, it becomes more and more painful. Many of our customers have large, multi-account deployments, with many different environments, and Terragrunt has proven useful for keeping their Terraform code maintainable.

conorgil:
@brikis98 Thanks for the detailed example. There is definitely more duplication that I thought. Consider me convinced!

brikis98:
PR not updated in several years, closing.

PR #235, opened 2017-06-16
Merge child extra_arguments with parent extra_arguments

philsttr:
For #147, merge child `extra_arguments` with parent `extra_arguments`

Copied and isolated @jocgir's base implementation from https://github.com/coveo/terragrunt

Added documentation and unit tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -487,8 +487,8 @@ The child `.tfvars` file's `terragrunt.terraform` settings will be merged into t
 settings as follows:
 
 * If an `extra_arguments` block in the child has the same name as an `extra_arguments` block in the parent,
-  then the child's block will override the parents.
-  * Specifying an empty `extra_arguments` block in a child will effectively remove the parent's block.
+  then the child's block will override the parent's.
+  * Specifying an empty `extra_arguments` block in a child with the same name will effectively remove the parent's block.
 * If an `extra_arguments` block in the child has a different name than `extra_arguments` blocks in the parent,
   then both the parent and child's `extra_arguments` will be effective.
 * The `source` field in the child will override `source` field in the parent
--- a/README.md
+++ b/README.md
@@ -491,6 +491,9 @@ settings as follows:
   * Specifying an empty `extra_arguments` block in a child with the same name will effectively remove the parent's block.
 * If an `extra_arguments` block in the child has a different name than `extra_arguments` blocks in the parent,
   then both the parent and child's `extra_arguments` will be effective.
+  * The child's `extra_arguments` will be placed _after_ the parent's `extra_arguments` on the terraform command line.
+  * Therefore, if a child's and parent's `extra_arguments` include `.tfvars` files with the same variable defined,
+    the value from the `.tfvars` file from the child's `extra_arguments` will be used by terraform.
 * The `source` field in the child will override `source` field in the parent
 
 Other settings in the child `.tfvars` file's `terragrunt` block (e.g. `remote_state`) override the respective
--- a/README.md
+++ b/README.md
@@ -481,8 +481,19 @@ terragrunt = {
 
 The `include` block tells Terragrunt to use the exact same Terragrunt configuration from the `terraform.tfvars` file
 specified via the `path` parameter. It behaves exactly as if you had copy/pasted the Terraform configuration from 
-the root `terraform.tfvars` file into `mysql/terraform.tfvars`, but this approach is much easier to maintain! Note
-that if you include any other settings in the `terragrunt` block of a child `.tfvars` file, it will override the
+the root `terraform.tfvars` file into `mysql/terraform.tfvars`, but this approach is much easier to maintain!
+
+The child `.tfvars` file's `terragrunt.terraform` settings will be merged into the parent file's `terragrunt.terraform`
+settings as follows:
+
+* If an `extra_arguments` block in the child has the same name as an `extra_arguments` block in the parent,
+  then the child's block will override the parents.
+  * Specifying an empty `extra_arguments` block in a child will effectively remove the parent's block.
+* If an `extra_arguments` block in the child has a different name than `extra_arguments` blocks in the parent,
+  then both the parent and child's `extra_arguments` will be effective.
+* The `source` field in the child will override `source` field in the parent
+
+Other settings in the child `.tfvars` file's `terragrunt` block (e.g. `remote_state`) override the respective
 settings in the parent.
 
 The `terraform.tfvars` files above use two Terragrunt built-in functions:
--- a/config/config.go
+++ b/config/config.go
@@ -2,14 +2,15 @@ package config
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
 	"github.com/gruntwork-io/terragrunt/errors"
 	"github.com/gruntwork-io/terragrunt/options"
 	"github.com/gruntwork-io/terragrunt/remote"
 	"github.com/gruntwork-io/terragrunt/util"
 	"github.com/hashicorp/hcl"
-	"os"
-	"path/filepath"
-	"strings"
 )
 
 const DefaultTerragruntConfigPath = "terraform.tfvars"
@@ -225,7 +226,7 @@ func parseConfigString(configString string, terragruntOptions *options.Terragrun
 		return nil, err
 	}
 
-	return mergeConfigWithIncludedConfig(config, includedConfig)
+	return mergeConfigWithIncludedConfig(config, includedConfig, terragruntOptions)
 }
 
 // Parse the given config string, read from the given config file, as a terragruntConfigFile struct. This method solely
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -481,8 +481,19 @@ terragrunt = {
...
 specified via the `path` parameter. It behaves exactly as if you had copy/pasted the Terraform configuration from 
-the root `terraform.tfvars` file into `mysql/terraform.tfvars`, but this approach is much easier to maintain! Note
-that if you include any other settings in the `terragrunt` block of a child `.tfvars` file, it will override the
+the root `terraform.tfvars` file into `mysql/terraform.tfvars`, but this approach is much easier to maintain!
+
+The child `.tfvars` file's `terragrunt.terraform` settings will be merged into the parent file's `terragrunt.terraform`
+settings as follows:
+
+* If an `extra_arguments` block in the child has the same name as an `extra_arguments` block in the parent,
+  then the child's block will override the parents.
```
brikis98: s/parents/parent's/

README.md
```diff
@@ -481,8 +481,19 @@ terragrunt = {
...
+  * Specifying an empty `extra_arguments` block in a child will effectively remove the parent's block.
```
brikis98: s/an empty `extra_arguments` block/an empty `extra_arguments` block of the same name/

config/config.go
```diff
@@ -268,6 +276,25 @@ func mergeConfigWithIncludedConfig(config *TerragruntConfig, includedConfig *Ter
...
+	result := *parentExtraArgs
+addExtra:
+	for _, child := range childExtraArgs {
+		for i, parent := range result {
+			if parent.Name == child.Name {
+				terragruntOptions.Logger.Printf("extra_arguments '%v' from child overriding parent", child.Name)
+				// For extra args, we want to keep the values specified in the child and put them after
+				// the parent ones, so if we encounter a duplicate, we just overwrite it.
+				result[i] = child
+				continue addExtra
```
brikis98: I'm a bit confused by these nested loops and label usage.

Wouldn't it be simpler to load them into a `map[string]TerraformExtraArguments`, with all the parent items first, then all the child items, and using the `Name` as the key? If you do that, any duplicate names in the children will override the parents. When you're done, you just return the map values as a lis.

philsttr: Agreed.  I just took @jocgir 's code and ran with it.  

However, now that I think about it, I think we do want to impose some ordering here (which a map can't provide).  For example, if the parent extra_arguments includes a variable file containing `variable_x`, and a child also includes a variable file containing `variable_x`, then I hope we can order the extra_arguments such that the child's file "wins".  I'm not sure if terraform honors var file ordering on the cli, or which one "wins" (first file containing the variable, or last file containing the variable).  I'll do a little investigation.

brikis98: Ah, you're right about the ordering. In that case, instead of a `map`, load them into a `list` if the list doesn't contain the element already, using a method similar to [ListContainsElement](https://github.com/gruntwork-io/terragrunt/blob/master/util/collections.go#L9) to check each item by `Name`. Yes, this is O(n^2), but given that we're dealing with a tiny number of items, it'll make no difference performance-wise, but will make the code easier to read and maintain.

philsttr: Ok, looks like the _last_ tfvars file on the command line containing a variable "wins" in the case where the variable is defined in multiple files.  I tested this out in terraform.

So, I think the current implementation is "functionally correct", in that it always adds child extra args at the end.  Therefore, terragrunt will put the child's extra_arguments` after the parent's `extra_arguments` on the terraform command line.

I have clarified this behavior in the documentation.

Let me know what you think.

philsttr:
Copy of unit test output:

```
Running tool: /usr/bin/go test -timeout 30s -tags  -run ^TestParseTerragruntConfigRemoteStateMinimalConfig|TestParseTerragruntConfigRemoteStateMissingBackend|TestParseTerragruntConfigRemoteStateFullConfig|TestParseTerragruntConfigDependenciesOnePath|TestParseTerragruntConfigDependenciesMultiplePaths|TestParseTerragruntConfigRemoteStateDynamoDbTerraformConfigAndDependenciesFullConfig|TestParseTerragruntConfigRemoteStateDynamoDbTerraformConfigAndDependenciesFullConfigOldConfigFormat|TestParseTerragruntConfigInclude|TestParseTerragruntConfigIncludeWithFindInParentFolders|TestParseTerragruntConfigIncludeOverrideRemote|TestParseTerragruntConfigIncludeOverrideAll|TestParseTerragruntConfigTwoLevels|TestParseTerragruntConfigThreeLevels|TestParseTerragruntConfigEmptyConfig|TestParseTerragruntConfigEmptyConfigOldConfig|TestMergeConfigIntoIncludedConfig|TestParseTerragruntConfigTerraformNoSource|TestParseTerragruntConfigTerraformWithSource|TestParseTerragruntConfigTerraformWithExtraArguments|TestParseTerragruntConfigTerraformWithMultipleExtraArguments|TestFindConfigFilesInPathNone|TestFindConfigFilesInPathOneNewConfig|TestFindConfigFilesInPathOneOldConfig|TestFindConfigFilesInPathMultipleConfigs$

[terragrunt] [config_test] 2017/06/16 09:41:42 extra_arguments 'overrideArgs' from child overriding parent
PASS
ok  	github.com/philsttr/terragrunt/config	0.014s
Success: Tests passed.
```

config/config.go
brikis98: That seems like the proper behavior. That said, I'm not a fan of two nested loops plus a label & continue (ie goto). I had to re-read the code a few times to understand it. Would you mind keeping the same behavior, but changing the code as described above: that is, two simple, un-nested loops that load elements into a list if that element isn't in the list already?

philsttr: Sure thing.  I'll try to make it more readable.

However, I'm not sure how to completely avoid nested loops.  You need an inner loop to determine whether the element is contained in the array.  Although, I could pull that out into a separate function to make it more clear.   (I'm fairly new to go though.  Is there a built-in way to detect whether an array contains an element?)

brikis98: Yes, moving it into a separate function is the idea, just to improve readability.

Go, unfortunately, does not include a built-in `contains` method for arrays. You just implement it yourself using a loop. And since there are no generics either, you have to implement it separately for each type... The language definitely has some weaknesses :)

jocgir: Just a note, I am not particularly proud of this piece of code. It it a reason why I did not yet submit it as a PR. And with go lacking of support for generics, I have 3 times almost the same code.

In our case, we have 4 level of inclusion. I had to do a lot of gymnastic with the include mechanism to make it work and while it work well for our use case, I did not test it over all scenarios.

The final order is definitely important.

philsttr: Totally understood @jocgir .   Makes sense given go's limitations.

@brikis98 I swapped out the continue/label with a slightly different implementation that I think is a bit more clear.  Let me know if this works for you.

philsttr:
New test result after refactor.

```
Running tool: /usr/bin/go test -timeout 30s -tags  -run ^TestParseTerragruntConfigRemoteStateMinimalConfig|TestParseTerragruntConfigRemoteStateMissingBackend|TestParseTerragruntConfigRemoteStateFullConfig|TestParseTerragruntConfigDependenciesOnePath|TestParseTerragruntConfigDependenciesMultiplePaths|TestParseTerragruntConfigRemoteStateDynamoDbTerraformConfigAndDependenciesFullConfig|TestParseTerragruntConfigRemoteStateDynamoDbTerraformConfigAndDependenciesFullConfigOldConfigFormat|TestParseTerragruntConfigInclude|TestParseTerragruntConfigIncludeWithFindInParentFolders|TestParseTerragruntConfigIncludeOverrideRemote|TestParseTerragruntConfigIncludeOverrideAll|TestParseTerragruntConfigTwoLevels|TestParseTerragruntConfigThreeLevels|TestParseTerragruntConfigEmptyConfig|TestParseTerragruntConfigEmptyConfigOldConfig|TestMergeConfigIntoIncludedConfig|TestParseTerragruntConfigTerraformNoSource|TestParseTerragruntConfigTerraformWithSource|TestParseTerragruntConfigTerraformWithExtraArguments|TestParseTerragruntConfigTerraformWithMultipleExtraArguments|TestFindConfigFilesInPathNone|TestFindConfigFilesInPathOneNewConfig|TestFindConfigFilesInPathOneOldConfig|TestFindConfigFilesInPathMultipleConfigs$

[terragrunt] [config_test] 2017/06/16 11:03:10 extra_arguments 'overrideArgs' from child overriding parent
PASS
ok  	github.com/philsttr/terragrunt/config	0.014s
Success: Tests passed.
```

brikis98:
New release is here: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.12.23. The binaries should show up in a few minutes when the build finishes.

philsttr:
Thanks!

PR #236, opened 2017-06-16
Don't copy hidden files and folders into temp folder

brikis98:
When Terragrunt downloads Terraform configurations from a `source` URL into a temp folder, it also copies the files in the current working directory (e.g. `terraform.tfvars`) into that temp folder. Unfortunately, as part of the copy process, Terragrunt would also copy hidden files and folders, such as `.git` or `.terraform`, which would lead to confusing errors. This PR updates Terragrunt to ensure it no longer copies hidden files & folders.


brikis98:
This is a fairly simple change with reasonable tests, so I'm going to merge. Feedback welcome!

PR #238, opened 2017-06-18
[WIP] Fix a bunch of Windows bugs

brikis98:
This PR is a WIP to fix a bunch of filepath issues on Windows:

* Need to use a shorter tmp folder path, or our integration tests hit the 260 character path file limit.
* Fix checks for hidden file/folders 
* Fix slashes in the paths of multiple tests
* Separate tests for relative paths with `extra_arguments`... We may want to normalize these paths in the future for teams that use both Windows and *nix.

If I can get all the tests stable on both Windows and *nix, I'll merge this in, and set up a Windows CI so we don't have these regressions in the future.


brikis98:
OK, as far as I can tell, all the tests now pass on both Windows and *nix. I'm going to merge. I'll add Windows CI in the next day or two.

PR #242, opened 2017-06-22
fixed a typo in the readme

james-powis:
Nothing sexy, just a typo correction.


brikis98:
Ah, good catch. Thanks!

Issue #208, opened 2017-05-03
plan-all fails if root has a parent terraform.tfvars


dpetzel:
This feels like a bug to me, but I admit maybe its simply not a supported configuration, but figured I'd open this for clarification. I have this tree:
```
.
├── dev
│   ├── thing1
│   │   ├── backend.tf
│   │   └── terraform.tfvars
│   ├── thing2
│   │   ├── backend.tf
│   │   └── terraform.tfvars
│   └── terraform.tfvars
```
The parrent `terraform.tfvars` is only setting remote state configuration:
```
terragrunt = {
  remote_state {
    backend = "s3"
    config {
      bucket  = "mine"
      key     = "state/${path_relative_to_include()}/terraform.tfstate"
      region  = "us-east-1"
      encrypt = true
    }
  }
}
```

Children are like this:
```
terragrunt = {
  include {
    path = "${find_in_parent_folders()}"
  }
  terraform {
    source = "blah"
  }
}
```

When I run a `terragrunt plan-all` while sitting in the `dev` folder I get the following exception:
```
[terragrunt] [/my_path] 2017/05/03 16:37:28 Module /my_path has finished successfully!
[terragrunt] 2017/05/03 16:37:28 configstack.MultiError Encountered the following errors:
exit status 1
/home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/configstack/running_module.go:142 (0x4a924c)
/home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/configstack/running_module.go:126 (0x4a903f)
/home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/configstack/running_module.go:59 (0x4a86c9)
/home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/configstack/stack.go:40 (0x4aa9c9)
/home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/cli/cli_app.go:276 (0x45d3ed)
/home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/cli/cli_app.go:215 (0x45c94c)
/home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/cli/cli_app.go:169 (0x45c374)
/home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/cli/cli_app.go:152 (0x45c097)
[...]

PR #258, opened 2017-07-27
WIP: tfvars files should be ignored during *-all commands if they have no source setting and no .tf file is present

jonahh-yeti:
Bit of a fresh GO programmer, but this change should do the trick.  The outstanding issue is that since the test data used is mostly just blank .tfvars files, any attempt to resolve these is now ignored so a ton of tests are failing.  Any advice on how to proceed or comments on the code would be welcome :)

Fix for issue https://github.com/gruntwork-io/terragrunt/issues/208


configstack/module.go
```diff
@@ -83,7 +84,18 @@ func resolveTerraformModule(terragruntConfigPath string, terragruntOptions *opti
 		return nil, err
 	}
 
-	return &TerraformModule{Path: modulePath, Config: *terragruntConfig, TerragruntOptions: opts}, nil
+	if terragruntConfig.Terraform != nil {
+		s := []string{filepath.Dir(terragruntConfigPath), "*.tf"}
```
brikis98: Shouldn't the check for `.tf` files happen whether or not a `terraform { ... }` block exists?

brikis98: Also, instead of using an array here and then joining it below, try:

```go
matches, err := filepath.Glob(filepath.Join(filepath.Dir(terragruntConfigPath), "*.tf"))
```

configstack/module.go
```diff
@@ -83,7 +84,18 @@ func resolveTerraformModule(terragruntConfigPath string, terragruntOptions *opti
...
+		matches, err := filepath.Glob(strings.Join(s, ""))
+		if err != nil {
+			return nil, err
+		}
+		if terragruntConfig.Terraform.Source != "" || matches != nil {
```
brikis98: Reading the double-negative is a bit hard. Perhaps it would be simpler to invert this check?

```go
if terragruntConfig.Terraform.Source == "" && matches == nil {
  terragruntOptions.Logger.Printf("Module %s does not have an associated terraform config, sipping...", modulePath)
  return nil, nil
}
```

configstack/module.go
jonahh-yeti: Yeah that's much cleaner, thanks

configstack/module.go
jonahh-yeti: Looks like Go does use short-circuit evaluation on its Boolean operators, the new logic wraps the nil, nil return rather than the primary return and should look much cleaner!

jonahh-yeti:
Hey!  So, I made changes around your recommendations in the code.  The logic looks much cleaner overall now.  Also changed the tests so they don't all get caught by this addition, had to both edit the files in the test/fixture-modules/ directories and the expected values in the tests.  Finally, added a new test case and data in test/fixture-modules/module-l/ to check that a .tfvars file with no Terragrunt block and associated source, and no .tf file in the directory, will be skipped.  All tests are passing locally.  

Let me know what you think!

configstack/module.go
```diff
@@ -83,6 +84,16 @@ func resolveTerraformModule(terragruntConfigPath string, terragruntOptions *opti
 		return nil, err
 	}
 
+	// Fix for https://github.com/gruntwork-io/terragrunt/issues/208
+	matches, err := filepath.Glob(filepath.Join(filepath.Dir(terragruntConfigPath), "*.tf"))
+	if err != nil {
+		return nil, err
+	}
+	if (terragruntConfig.Terraform == nil || terragruntConfig.Terraform.Source == "") && matches == nil {
+		terragruntOptions.Logger.Printf("This module does not have an associated terraform config, sipping...")
```
brikis98: s/sipping/skipping/. Also, include the module path in this output so we know which module is being skipped!

jonahh-yeti: Fixed!  Can't believe I didn't catch the spelling error myself, and it'll include the directory path when it logs.

brikis98:
Thank you, merging now!

brikis98:
If the tests pass, the new binaries should show up here soon: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.12.25

Issue #252, opened 2017-07-06
Compatibility with terraform 0.10.0


philsttr:
At least one change in [terraform 0.10.0](https://github.com/hashicorp/terraform/blob/master/CHANGELOG.md#0100-beta1-june-22-2017) has broken terragrunt...

Specifically this change:

> The `terraform init` subcommand no longer takes a SOURCE argument to copy to the current directory. The behavior has been changed to match that of `plan` and `apply`, so that a configuration can be provided as an argument on the commandline while initializing the current directory. If a module needs to be copied into the current directory before initialization, it will have to be done manually.

Terragrunt uses the old two argument form of init, and terraform complains:

```
$ terragrunt plan
[terragrunt] [/foobar] 2017/07/05 16:55:38 Running command: terraform --version
[terragrunt] 2017/07/05 16:55:39 Reading Terragrunt config file at /foobar/terraform.tfvars
[terragrunt] 2017/07/05 16:55:39 Downloading Terraform configurations from git::ssh://git@githost/org/proj.git?ref=master into /tmp/terragrunt/Q9fDMTFutsTifIUWoQCicP-_WVY/b8EcM0nU_fk63G8D7qkIzC15V0Y
[terragrunt] 2017/07/05 16:55:39 Running command: terraform init -backend=false -get=false git::ssh://git@githost/org/proj.git?ref=master /tmp/terragrunt/Q9fDMTFutsTifIUWoQCicP-_WVY/b8EcM0nU_fk63G8D7qkIzC15V0Y
The init command expects at most one argument.

Usage: terraform init [options] [DIR]
```

There might be more incompatibilities, but this is the first one that I noticed.  Would be good to do some investigation into other incompatibilities.

brikis98:
Gah, that sucks. Thanks for pointing this out. I just left a comment asking about workarounds and alternatives for this completely backwards incompatible change: https://github.com/hashicorp/terraform/pull/15032#issuecomment-313277025.

jre21:
Since the comment doesn't seem to be attracting any attention from the devs, perhaps it would be worthwhile to open it as a new issue terraform's issue tracker?

brikis98:
@jre21 A bug on Terraform's issue tracker sounds worthwhile!

jre21:
https://github.com/hashicorp/terraform/issues/15655

brikis98:
Thanks!

jre21:
[...]

Issue #239, opened 2017-06-19
Support terraform init --reconfigure


philsttr:
We have a use case for a script (and occasionally by hand) that requires reconfiguring the terraform backend _without_ copying state.

With terraform 0.9.4+, this can be accomplished by passing the `--reconfigure` flag to `terraform init` (See [here](https://www.terraform.io/docs/commands/init.html#reconfigure)).

It would be great if terragrunt exposed some cli flag that resulted in passing the `--reconfigure` flag to `terraform init`.  This could be done generically (e.g. `--extra-init-argument xxxx`) or specifically (e.g. `--terragrunt-reconfigure-backend`)

I can currently hack around this by calling terragrunt like this:

```
echo no | terragrunt --terragrunt-non-interactive plan
```

But that is not very elegant.

For reference, this is the prompt from terraform that passing `--reconfigure` will avoid.

```
Initializing the backend...
Backend configuration changed!

Terraform has detected that the configuration specified for the backend
has changed. Terraform will now reconfigure for this backend. If you didn't
intend to reconfigure your backend please undo any changes to the "backend"
section in your Terraform configuration.


Do you want to copy the state from "s3"?
  Would you like to copy the state from your prior backend "s3" to the
  newly configured "s3" backend? If you're reconfiguring the same backend,
  answering "yes" or "no" shouldn't make a difference. Please answer exactly
  "yes" or "no".

  Enter a value: no
```

brikis98:
Ah, good call. Supporting some sort of `--terragrunt-init-arg XXX`, where you can specify whatever you want for `XXX`, would make sense. PRs welcome!

philsttr:
Yeah, looks like [0.10](https://github.com/hashicorp/terraform/blob/master/CHANGELOG.md) is adding an `-upgrade` flag as well.  A generic mechanism is probably best.

philsttr:
0.10.0 also added a several flags to the init command [to control plugin installation](https://www.terraform.io/docs/commands/init.html#plugin-installation).  Would be good if there were some way to use these.

dscowan:
Is there any plan to add this anytime soon.  The workaround of "echo no" listed above does not seem to work for me.

brikis98:
As of #261, you should be able to run `terragrunt init` and pass arguments to it, including `--reconfigure`. Is that not working for you?

PR #261, opened 2017-08-03
Make terraform init command compatible with terraform 0.10.0

philsttr:
For #252, make terraform init command compatible with terraform 0.10.0

Specifically,
* terraform < 0.10.0 used `terraform init [SOURCE] [DIR]`
* terraform >= 0.10.0 uses `terraform init -from-module=[SOURCE] [DIR]`

terragrunt now requires terraform >= 0.10.0


brikis98:
Thanks for the PR! 

I'm a little concerned about making terragrunt incompatible with Terraform 0.9.x. What are your thoughts on the following:

1. In `cli_app.go`, we call [CheckTerraformVersion](https://github.com/gruntwork-io/terragrunt/blob/master/cli/cli_app.go#L158) to verify the `DEFAULT_TERRAFORM_VERSION_CONSTRAINT`. Under the hood, that command calls `getTerraformVersion`, which, in turn, shells out to `terraform --version`. If we just store this version number, we can use it in the `download_source` command to decide if the `-from-module=` flag is necessary.

1. To make that happen, first change [ParseTerragruntOptions](https://github.com/gruntwork-io/terragrunt/blob/master/cli/args.go#L17) to call `getTerraformVersion` and store the result in a new field in a new field in the [TerragruntOptions struct](https://github.com/gruntwork-io/terragrunt/blob/master/options/options.go#L16). 

1. Then update `CheckTerraformVersion` to merely compare the value stored in `TerragruntOptions` against `DEFAULT_TERRAFORM_VERSION_CONSTRAINT`.

1. Finally, update [terraformInit](https://github.com/philsttr/terragrunt/blob/3eff2094bcb4bd7396ddc721be9bd196df811e29/cli/download_source.go#L354) to include `-from-module=` only if the Terraform version is 0.10.0 and higher.

brikis98:
Hmm, one more thing I just realized. Since providers are now separate and must be downloaded via `terraform init`, calling `init` is now a required part of *ALL* Terraform usage, but Terragrunt doesn't allow it: https://github.com/gruntwork-io/terragrunt/blob/master/cli/cli_app.go#L228-L231. Also, per your comments in #239, it would be nice to expose `init` in Terragrunt to allow access to the other new `init` flags too.

Not sure if it should be part of this PR or a separate one, but to be fully compatible with Terraform 0.10.0, I think we'll need to update the check for the `init` command in Terragrunt to append the necessary remote state args automatically and run the command on behalf of the user. That way, when you call `terragrunt init`, it will do more or less exactly what you expect.

philsttr:
Yeah, I was debating myself whether it would be better to:

A) keep `terraform init` implicit within terragrunt as it is today, but add an ability to "pass-through" args to `terraform init` from any terragrunt command (similar to what we already discussed in #239), or

B) allow/require `terragrunt init`, and pass the args given by the user AND the other "auto-calculated" args already passed by terragrunt to the init command for initializing the backend properly.

I think I like #B better.  I get the impression that you do as well.

Assuming we go with #B, would you like to _also_ keep the "implicit" call to `terraform init` if the user has never called `terragrunt init`?  Or should terragrunt require that the user call `terragrunt init`?

If it's required that the user call `terragrunt init`, then this would be a somewhat-significant change to the terragrunt workflow.  

And as an extension of that, if the terragrunt workflow is changed significantly, I personally wouldn't have a problem with dropping support for terraform 0.9.x to keep the terragrunt code cleaner.  It's not particularly a problem in my workplace because of the way we manage versions  (i.e. if I want to use terraform 0.9.x, then I just run an older version of terragrunt.  if I want to use terraform 0.10.x, then I'll run a newer version of terragrunt).    However, having said that, I acknowledge that adding some complexity to terragrunt to support backwards compatibility would be beneficial to some.

brikis98:
My vote is that Terragrunt keeps working as-is, with the sole exception that we make it possible to call `init` manually, and we auto-add the remote state arguments to that command. That way, it's backwards compatible, works with the 0.10.x flow, and doesn't require too much code change.

philsttr:
Ok. Fair enough.

In that case, one thing that I think terragrunt should protect against is _if the user has ever run `terragrunt init` explicitly, terragrunt should never call `terraform init` on behalf of the user during other terragrunt commands._  If it did, then terragrunt has no way to pass user's custom args to the init command.

For example, currently terragrunt detects if the remote state config has changed, and [prompts the user to see if they want to override it](https://github.com/gruntwork-io/terragrunt/blob/cc72c3dfdafb6077bf03d0ff3b94e8347b32f0e6/remote/remote_state.go#L95-L99).  If the user answers yes, then terragrunt will execute `terraform init` on behalf of the user.   This would be bad in the case the user has explicitly run `terragrunt init` with custom args in the past.

Does that make sense?  Do you agree?

If so, how can terragrunt detect the case that the user has explicitly run `terragrunt init`?

brikis98:
> if the user has ever run terragrunt init explicitly, terragrunt should never call terraform init on behalf of the user during other terragrunt commands. If it did, then terragrunt has no way to pass user's custom args to the init command.

What custom args to the `init` command are you referring to? The only ones that look problematic are the backend ones, and Terragrunt handles those for you...

philsttr:
`-get-plugins`
`-plugin-dir`
`-verify-plugins`

Perhaps others in the future.

brikis98:
With the exception of perhaps `-plugin-dir`, those seem fairly harmless to omit. There's also a simple workaround: specify those flags in [extra_arguments](https://github.com/gruntwork-io/terragrunt#keep-your-cli-flags-dry) for the `init` command.

philsttr:
Specifying things like `-plugin-dir` and `-get-plugins` via extra_arguments makes it a hard to support running terragrunt in different environments.  For example. In prod, we don't want terraform downloading extra binaries (`-get-plugins=false`).  But during dev we don't mind.  Or in prod we use a different plugin dir than on a developer's machine (`-plugin-dir=xyz`)

From tracing through the code it doesn't look like extra_arguments is supported for the "implicit"  `terraform init` that gets run as part of other terragrunt commands.  It is currently only supported if the user executed `terragrunt init`.  The filtering for extra_arguments is done based on the arguments passed to terragrunt.  But support could be added for that as part of this change.

brikis98:
> Specifying things like -plugin-dir and -get-plugins via extra_arguments makes it a hard to support running terragrunt in different environments.

If you have a separate `.tfvars` for each environment (which you typically should), then it doesn't make it harder at all...

> From tracing through the code it doesn't look like extra_arguments is supported for the "implicit" terraform init that gets run as part of other terragrunt commands. It is currently only supported if the user executed terragrunt init. The filtering for extra_arguments is done based on the arguments passed to terragrunt. But support could be added for that as part of this change.

Adding it there would make sense too!

philsttr:
Let me clarify what I was trying to explain.

When I mentioned 'environments', I wasn't talking about the environments in which the resources are created, I was talking about the environments in which terragrunt is executed.  We _definitely_ have different `.tfvars` files for the environments in which resources are created .  Let me explain what I mean by the environments in which terragrunt is executed (i'll call these _execution environments_).

Say developer A uses windows, developer B uses a mac, and the jenkins server uses linux.  Each of them perform operations on the same set of .`tfvars` files.    For example, the developer does some testing on the tfvars files as they are writing them prior to checkin.  But then after checkin, the jenkins server actually does the "real" work.  So, if the -plugin-dir and -get-plugins options are specified as extra args, then each execution environment will be forced to use the same values.  This is not what I would like.  On developer machines, I don't really want to specify -plugin-dir, and I want -get-plugins=true, since this will allow rapid development and experimentation with various providers.  But on the jenkins server, I want to specify -plugin-dir and -get-plugins=false, because I want more control over what executes.  Even if I wanted to prevent downloading plugins in all execution environments, there's no way I could pick a value for -plugin-dir that would be valid for every execution environment.

brikis98:
Thanks for the explanation! 

So I guess that leaves us with two options:

1. Keep Terragrunt backwards compatible, in that it calls `init` automatically. However, add some sort of support for specifying custom `init` args on a per-user (per-execution-environment) basis. E.g. We could allow you to add `--terragrunt-init-arg "-foo=bar"` to any command and/or specify an env var along the lines of `TERRAGRUNT_INIT_ARGS="-foo=bar"`.  

1. Change Terragrunt to no longer call `init` automatically. Instead, you have to call `init` yourself and Terragrunt will automatically append remote state args to that `init` call. 

The second option does seem cleaner and more inline with the workflow now recommended by Terraform. The only downside is it breaks backwards compatibility; but I guess that compatibility was broken by Terraform itself in 0.10.x (due to the provider change) and not Terragrunt, so I can probably live with that.

Does anyone have any objections? @josh-padnick, @mcalhoun, thoughts?

chrisb3ll:
If I had a vote, it would be option 1. 

Terragrunt calling the init command automatically is one of the reasons I use Terragrunt. 🙂

brikis98:
Hm, I just realized the `terraform init` command now allows you to [pass a file](https://www.terraform.io/docs/backends/config.html#partial-configuration) via the `-backend-config` flag. 

This gives us another possible approach: remove the `remote_state` configuration from Terragrunt entirely! Instead, users can move all those settings into a `backend.conf` file and either manually include them when calling `terragrunt init`, or specifying the path to those files using `extra_arguments` in the Terragrunt configuration.

This seems the most idiomatic approach, although also the most backwards incompatible. Thoughts?

philsttr:
@brikis98 , I like it.

I feel that as terraform implements more and more features, any feature duplication from terragrunt should be removed  (similar to how terragrunt removed locking when terraform implemented locking).  Unfortunately, these will always break terragrunt compatibility.  But ultimately I think this will make terragrunt a better product, because it can focus on providing added value above terraform, without the added baggage of maintaining duplicate functionality.  Honestly, who knows, eventually terraform might provide all the functionality of terragrunt, and terragrunt would no longer be necessary.

I'm probably of the minority opinion on this, but I also feel that the terragrunt workflow should mimic terraform's workflow as closely as possible, and that terragrunt should remain as slim of a layer as possible above terraform.  This makes it easy and natural to transition between terragrunt and terraform.  Lots more people are familiar with terraform at this point, so keeping workflows similar will help those people also learn terragrunt.  On a personal level, I was [surprised](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) when I received an error message after issuing my first `terragrunt init`.  I had to dig into terragrunt to understand what exactly it did.  I was even more surprised to learn that it runs `terraform init` twice, and I had to dig in even more to see what each one was doing.  Now I loath every time I have to explain it to somebody that is only familiar with terraform.

So, my first preference would be:

1. Bump terragrunt to 0.13.0
2. Drop compatiblity with terraform 0.9.x  Make it only compatible with terraform 0.10.0 +.  If people want to continue to use the out-of-date terraform 0.9.x version, then they can use the out-of-date terragrunt 0.12.x version.  Any huge terragrunt bugfixes could be backported if needed.
3. Make terragrunt _require_ the same workflow as terraform.  i.e. require `terragrunt init`.  
4. Remove all implicit `terraform init` calls  (therefore avoiding the need to make those implicit calls support `extra_arguments`)
5. Remove backend config variables from terragrunt config.  Instead, provide the same functionality via `extra_arguments` for the init command to configure the backend via a backend config file.

Advantages:

* greatly simplify the terragrunt code, 
* allow it to more easily support terraform changes in the future, since all terragrunt commands will become pass-through commands to terraform, and
* be much more intuitive for users coming from terraform, since the same workflow is used
* easier to document  (e.g. terragrunt uses the same workflow as terraform.   `extra_arguments` just applies to the terragrunt commands that a user issues.  no need to document implicit `terraform init` cases.)

Disadvantages:

* Breaks compatibility with previous terragrunt/terraform versions



If that is too big of a change, my second preference would be:

1. Keep compatibility with terraform 0.9.x
2. Make terragrunt _allow_ the same workflow as terraform.  i.e. allow `terragrunt init`.
3. Keep supporting implicit `terraform init` calls.  But provide a way to disable them somehow. (`--terragrunt-no-init` ?  something in the terragrunt config block?)
4. Make implicit `terraform init` calls honor `extra_arguments`.  (Do these apply to both of the current `terraform init` calls?)
5. Keep duplicating terraform functionality for storing backend config in a file.

Advantages:

* Compatible with previous terragrunt/terraform versions
* Maintains support for not having to explicitly call `terragrunt init` for use cases that do not require customization of  `terraform init` for each user / execution environment.

Disadvantages:

* Complicates terragrunt code
* Supporting the "terraform workflow" (where init is required) is more difficult for a user to do  (e.g. they would have to add a `--terragrunt-no-init` argument or something to terragrunt's config).  Note I see this as a disadvantage, but others may see this as an advantage.
* Duplicates terraform functionality
* Have to support `extra_arguments` for implicit commands  (and which one of the current `terraform init` calls does it apply to?  One? the other? both?  Good luck documenting that ;)


Sorry that was so long.  Thanks for considering.  These are my opinions.  I won't be offended if yours differ.

brikis98:
Thanks for the detailed thoughts @philsttr. I realized that one issue with using a `backend.conf` file is that it doesn't support many of the helpers we've added to Terragrunt: e.g., setting the `key` to `${path_relative_to_include()}`. Specifying the key manual for every single module is tedious and error prone, so it's important to make sure this works. 

I suppose the usage pattern could be something like this:

```hcl
terragrunt = {
  terraform = {
    extra_arguments "init" {
      commands = ["init"]
      arguments = [
        "-backend-config=../root.conf",
        "-backend-config=key=${path_relative_to_include()}/terraform.tfstate"
      ]
    }
  }
}
```

philsttr:
That seems reasonable to me.

brikis98:
Actually, something just occurred to me. We could support both:

1. If you specify a `remote_state` block in your Terragrunt config, then Terragrunt will call `init` on your behalf. 
1. If you don't you can use `extra_arguments` instead, and call `init` yourself.

The only changes we'd have to make to Terragrunt code would be:

1. Update Terragrunt to only call `init` if you specify `remote_state`.
1. Update the `init` command to the new format in 0.10.0.
1. Allow users to call `init` explicitly and add any arguments in `remote_state` to that `init` command automatically.

As far as I can tell, this would be largely backwards compatible. Thoughts?

bitbier:
Just throwing my two cents in. However, I've only been using `terragrunt` for the past week, so not sure how important my opinion is here.

I personally agree with @philsttr's first preference. It'd be nice to have backwards compatibility, but terraform itself is not even in version 1.0.0 yet. Things are going to change and when 0.x.x comes out probably even more stuff will. Having `terragrunt` follow closer to the new terraform workflow will just make it easier to reason about going forward. It'd also make it easier for newer people to adopt `terragrunt` until 🤞 `terraform` provides some of the functionality inside terragrunt and we can completely remove it.

I think that people are adopting terraform know that these things are going to change. We already having to start using terraform 0.10.0 because some of the resources that they are adding are only available there. People are going to hit this "you need to upgrade" but just reading the documentation. The documentation on `terraform` currently doesn't show what version a resource is supported in. Until terraform becomes 1.0.0, I think it's acceptable to break backwards compatibility simply because terraform itself is going to do this until it doesn't.

brikis98:
OK, the suggestion I have in https://github.com/gruntwork-io/terragrunt/pull/261#issuecomment-321393676 seems like the best path forward. Does anyone have objections? @mcalhoun or @josh-padnick? 

If not, @philsttr, are you interested in updating your PR to do that?

mcalhoun:
The suggested path seems logical to me.

philsttr:
I'm interested in doing it, but I can't make any promises on when I'll have it complete.  I'll try to get started on it this weekend though.  I'm still a golang newb, so can't make any promises on when it'll be complete.

brikis98:
@philsttr Thanks for giving it a shot!

philsttr:
I have good news and bad news.

The good news is that I was able to get a lot of this working.

The bad news is that I discovered a change in behavior in terraform 0.10.0 that throws a new wrench into terragrunt's current design.    (By the way, I'm so happy that terragrunt has a lot of tests.  Otherwise I would not have found this so easily)

In terraform prior to 0.10.0, [you can execute the following command](https://github.com/gruntwork-io/terragrunt/blob/ba92a4c23b481c48f1c9f0245a3ac38702c98274/cli/download_source.go#L354), and terraform will happily download the source _and not complain about modules that have not been downloaded yet_...

```
$ /usr/bin/terraform-0.9.8 init -backend=false -get=false file:///home/foo/go/src/github.com/philsttr/terragrunt/test/fixture-download/hello-world /tmp/terragrunt/bOJSeF_rBy7wWYCvIe_VG0jo2UU/9I7vef99YOWnS-fk5zIR8g8lHqQ
Initializing configuration from: "file:///home/foo/go/src/github.com/philsttr/terragrunt/test/fixture-download/hello-world"...

You may now begin working with Terraform. Try running "terraform plan" to see
any changes that are required for your infrastructure. All Terraform commands
should now work.

If you ever set or change modules or backend configuration for Terraform,
rerun this command to reinitialize your environment. If you forget, other
commands will detect it and remind you to do so if necessary.
```

In terraform 0.10.0, the similar command fails, because for some reason terraform tries to access modules and plugins...

```
$ /usr/bin/terraform-0.10.0 init -backend=false -get=false -get-plugins=false -from-module=file:///home/foo/go/src/github.com/philsttr/terragrunt/test/fixture-download/hello-world /tmp/terragrunt/bOJSeF_rBy7wWYCvIe_VG0jo2UU/9I7vef99YOWnS-fk5zIR8g8lHqQ
Copying configuration from "file:///home/foo/go/src/github.com/philsttr/terragrunt/test/fixture-download/hello-world"...
Error getting plugins: Error loading modules: module hello: not found, may need to be downloaded using 'terraform get'
```


Therefore, terragrunt's current design of using `terraform init` to download module sources _before_ running `terraform init` again to initialize the module does not work with terraform 0.10.0.

In terraform 0.10.0, when the module source is at the root of the repo  (e.g. there is no `//` with a subdirectory in the module source url), terragrunt must call `terraform init` only _once_ to both download the module AND initialize it.  Incidentally, if the module source is NOT at the root of the repo (e.g. there is a `//` with a subdirectory in the module source url), and no other terraform module exists at the root of the repo, then the current behavior of calling `init` once to download and once to initialize will work.

To make this more complex, `terraform init` _must_ be called _twice_ in the case where the module source is not at the root of the repo.  Because if you only call `terraform init` _once_  (essentially at the root level), then the subdirectory where the 'real' module resides won't be properly initialized. 

So in summary:
* If the source url has `//subdir` in it, then call `terraform init` twice.
  1. Once at the root level with args to download the repo  (`-backend=false -get=false -get-plugins=false`).
     * Note that this won't work if the repo has another module at the root level
  2. Once in the subdir to initialize the backend/modules/plugins
* If the source url does _not_ have `//subdir` in it, then call `terraform init` once with args to download AND initialize the backend/module/plugins

This is a solvable problem, but it's going to take some more work, and a change to the current terragrunt design/flow

brikis98:
@philsttr Thanks for digging into this!

I guess that new behavior from Terraform is part of the fall out of making `init` a required part of every workflow, either to configure remote state, or to download providers, or to download modules, and then, recursively, download more providers and more modules for that code.

Looking back at the code, it looks like we actually call `init` in three places:

1. To [download remote Terraform configurations](https://github.com/gruntwork-io/terragrunt/blob/master/cli/cli_app.go#L210).
1. To [download modules](https://github.com/gruntwork-io/terragrunt/blob/master/cli/cli_app.go#L215).
1. To [configure remote state](https://github.com/gruntwork-io/terragrunt/blob/master/cli/cli_app.go#L220).

The first one is separate because it's not a core Terraform feature, but something Terragrunt adds. The second and third could be combined, but IIRC, I kept them separate to try to download modules or configure remote state only when absolutely necessary, as each `init` call can be quite slow. It sounds like this may have to be refactored along the following logic:

1. If remote Terraform configuration needs to be downloaded, call `init`, setting the `-from-module-source` flag, and adding remote state arguments if those are present in the Terragrunt config. Do *not* add the `-get=false` or `-backend=false` arguments we were adding before.

1. In all other cases, go into the working directory, and only run `init` if a) the user has added remote state arguments to their Terragrunt config and b) a `.terraform` folder doesn't already exist. For this `init` run, include those remote state arguments, but again, do *not* add the `-get=false` or `-backend=false` arguments we were adding before.

Do you think that would solve the issue you were seeing? Is that something you have time to tackle?

philsttr:
Yeah, let me work on it a bit more.

philsttr:
Think I have everything working now.  I've updated the PR.  There's a little bit of churn, so feel free to squash.

All of the tests that do not require AWS are working with both terraform 0.9.8 and 0.10.0.

I have not run the tests that require AWS.  So, I could have broken something there.  Please test this.

I've introduced the concept of _Auto-Init_.  See the new README.md for a description.  Basically, when Auto-Init is enabled (the default), terragrunt is free to run whatever `terraform init` commands it deems necessary.  When Auto-Init is disabled (either through command line flag or environment variable), terragrunt will never run `terraform init` implicitly.  In this case you must always run `terragrunt init` separately.

I favored having an explicit control flag like this, rather than just basing it off of the presence of remote state, because terragrunt's remote state has a couple more features over terraform (terragrunt interpolations, and "initializing" the backend storage) that can be taken advantage of in either case.

Give it a try and let me know what you think

cli/cli_app.go
```diff
@@ -183,54 +189,165 @@ func runCommand(command string, terragruntOptions *options.TerragruntOptions) (f
...
+	}
+
+	// Need to clone the terragruntOptions, so the TerraformCliArgs can be configured to run the init command
+	initOptions := terragruntOptions.Clone(terragruntOptions.TerragruntConfigPath)
+	initOptions.TerraformCliArgs = []string{CMD_INIT}
+	initOptions.WorkingDir = terragruntOptions.WorkingDir
+
+	// Only add the arguments to download source if terraformSource was specified
+	downloadSource := terraformSource != nil
+	if downloadSource {
```
brikis98: Shy the extra variable instead of just checking `terraformSource != nil` directly in the if-statement?

cli/cli_app.go
```diff
@@ -183,54 +189,165 @@ func runCommand(command string, terragruntOptions *options.TerragruntOptions) (f
...
+		initOptions.AppendTerraformCliArgs(terraformSource.DownloadDir)
 	}
 
-	return shell.RunTerraformCommand(terragruntOptions, terragruntOptions.TerraformCliArgs...)
+	return runTerragruntWithConfig(initOptions, terragruntConfig, downloadSource)
+}
+
+// Returns an error if allowSourceDownload is false, and terragruntOptions.TerraformCliArgs contains source download related arguments
+func verifySourceDownloadArguments(allowSourceDownload bool, terragruntOptions *options.TerragruntOptions) error {
+	if !allowSourceDownload && len(terragruntOptions.TerraformCliArgs) > 1 {
```
brikis98: You can reduce nesting a bit by inverting the if-statement and returning `nil` immediately if it's true. Then the body of the if statement is moved outside of it and is just the regular method body.

cli/version_check.go
```diff
@@ -2,26 +2,38 @@ package cli
...
 // The terraform --version output is of the format: Terraform v0.9.5-dev (cad024a5fe131a546936674ef85445215bbc4226+CHANGES)
 // where -dev and (commitid+CHANGES) is for custom builds or if TF_LOG is set for debug purposes
 var TERRAFORM_VERSION_REGEX = regexp.MustCompile("Terraform (v?[\\d\\.]+)(?:-dev)?(?: .+)?")
 
-// Check that the currently installed Terraform version works meets the specified version constraint and return an error
-// if it doesn't
-func CheckTerraformVersion(constraint string, terragruntOptions *options.TerragruntOptions) error {
-	currentVersion, err := getTerraformVersion(terragruntOptions)
+// Populate the currently installed version of Terraform into the given terragruntOptions
+func PopulateTerraformVersion(terragruntOptions *options.TerragruntOptions) error {
```
brikis98: Very nice. I've been meaning to add this for a while.

options/options.go
```diff
@@ -127,6 +143,28 @@ func (terragruntOptions *TerragruntOptions) Clone(terragruntConfigPath string) *
 	}
 }
 
+// Inserts the given argsToInsert after the terraform command argument, but before the remaining args
+func (terragruntOptions *TerragruntOptions) InsertTerraformCliArgs(argsToInsert ...string) {
```
brikis98: Ah, nice work extracting this into a method.

remote/remote_state.go
```diff
@@ -52,57 +53,34 @@ func (remoteState *RemoteState) Initialize(terragruntOptions *options.Terragrunt
...
-// we should override the existing remote state setting.
-func shouldOverrideExistingRemoteState(existingBackend *TerraformBackend, remoteStateFromTerragruntConfig RemoteState, terragruntOptions *options.TerragruntOptions) (bool, error) {
-	if existingBackend.Type != remoteStateFromTerragruntConfig.Backend {
-		prompt := fmt.Sprintf("WARNING: Terraform remote state is already configured, but for backend %s, whereas your Terragrunt configuration specifies %s. Overwrite?", existingBackend.Type, remoteStateFromTerragruntConfig.Backend)
-		return shell.PromptUserForYesNo(prompt, terragruntOptions)
+// Returns true if this remote state is different than the given remote state that is currently being used by terraform.
+func (remoteState *RemoteState) differsFrom(existingBackend *TerraformBackend, terragruntOptions *options.TerragruntOptions) bool {
+	if existingBackend.Type != remoteState.Backend {
+		terragruntOptions.Logger.Printf("Backend type has changed from %s to %s", existingBackend.Type, remoteState.Backend)
+		return true
```
brikis98: Why remove the prompt to override that was here before?

remote/remote_state.go
```diff
@@ -111,17 +89,13 @@ func shouldOverrideExistingRemoteState(existingBackend *TerraformBackend, remote
 		}
 	}
 
-	if !reflect.DeepEqual(existingBackend.Config, remoteStateFromTerragruntConfig.Config) {
-		prompt := fmt.Sprintf("WARNING: Terraform remote state is already configured for backend %s with config %v, but your Terragrunt configuration specifies config %v. Overwrite?", existingBackend.Type, existingBackend.Config, remoteStateFromTerragruntConfig.Config)
-		return shell.PromptUserForYesNo(prompt, terragruntOptions)
+	if !reflect.DeepEqual(existingBackend.Config, remoteState.Config) {
+		terragruntOptions.Logger.Printf("Backend config has changed from %s to %s", existingBackend.Config, remoteState.Config)
```
brikis98: Same question here.

jre21:
How much work would it be to allow setting auto-init in the terragrunt block of terraform.tfvars?  In some scenarios, I'd think it would be easier to disable auto-init in a config file, rather than every developer having to remember to set an environment variable.

README.md
```diff
@@ -676,6 +678,18 @@ With the configuration above, when you run `terragrunt apply`, Terragrunt will c
...
+#### `extra_arguments` for `init`
+
+Extra arguments for the `init` command have some additional behavior and constraints.
+
+In addition to being appended to the `terraform init` command that is run when you explicitly run `terragrunt init`,
+`extra_arguments` for `init` will also be appended to the `init` commands that are automatically
+run during other commands (see [Auto-Init](#auto-init)).
+
+You must _not_ specify the `-from-module` option (aka. the `SOURCE` argument for terraform < 0.10.0) or the `DIR` argument
+in the `extra_arguments` for `init`.  This option and argument will be provided automatically by terragrunt.
```
philsttr: Done

cli/cli_app.go
philsttr: It's also used below the if block, in the `runTerragruntWithConfig` call.

remote/remote_state.go
philsttr: Couple reasons:

1. This code is now called twice during a single run.  It's called once to determine if `terraform init` needs to be called, and again to determine if the remote state needs to be "initialized" (e.g. create s3 bucket).
2. Terraform will also prompt when the backend changes.  Seemed overkill for terragrunt to prompt as well.

I can add the prompts back in, but I'll need to solve 1. 


BTW, what do you think about changing the "Initialize" term to "Provision", to differentiate it from `terraform init` ?  i.e. provision s3 bucket and dynamodb table

https://github.com/gruntwork-io/terragrunt/blob/64de850c2870cd5c1f2e80fce32777ea81f6a5d0/remote/remote_state.go#L23-L28

cli/cli_app.go
```diff
@@ -95,6 +96,7 @@ COMMANDS:
 GLOBAL OPTIONS:
    terragrunt-config                    Path to the Terragrunt config file. Default is terraform.tfvars.
    terragrunt-tfpath                    Path to the Terraform binary. Default is terraform (on PATH).
+   terragrunt-no-auto-init              Don't automatically run 'terraform init' during other terragrunt commands.  Require using 'terragrunt init'.
```
philsttr: Done

cli/cli_app.go
```diff
@@ -183,54 +189,165 @@ func runCommand(command string, terragruntOptions *options.TerragruntOptions) (f
...
+func runTerragruntWithConfig(terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig, allowSourceDownload bool) error {
+
+	// Add extra_arguments to the command
+	if terragruntConfig.Terraform != nil && terragruntConfig.Terraform.ExtraArgs != nil && len(terragruntConfig.Terraform.ExtraArgs) > 0 {
+		terragruntOptions.InsertTerraformCliArgs(filterTerraformExtraArgs(terragruntOptions, terragruntConfig)...)
 	}
 
-	if sourceUrl, hasSourceUrl := getTerraformSourceUrl(terragruntOptions, conf); hasSourceUrl {
-		if err := downloadTerraformSource(sourceUrl, terragruntOptions); err != nil {
+	if firstArg(terragruntOptions.TerraformCliArgs) == CMD_INIT {
```
philsttr: Done

cli/cli_app.go
```diff
@@ -183,54 +189,165 @@ func runCommand(command string, terragruntOptions *options.TerragruntOptions) (f
...
-	if conf.RemoteState != nil {
-		if err := configureRemoteState(conf.RemoteState, terragruntOptions); err != nil {
+	remoteStateNeedsInit, err := remoteStateNeedsInit(terragruntConfig.RemoteState, terragruntOptions)
+	if err != nil {
+		return false, err
+	}
+	if remoteStateNeedsInit {
+		return true, nil
+	}
+	return false, nil
```
philsttr: Done

cli/cli_app.go
brikis98: Ah, gotcha

remote/remote_state.go
brikis98: Hm, alright, if Terraform prompts about it, then that's probably good enough.

Renaming "Initialize" to "Provision" probably makes sense, but perhaps as part of a different PR, to prevent this one from getting much larger.

philsttr:
@brikis98 Awesome!  Also, let me know if the tests that require AWS don't pass.

@jre21, probably not that bad.  can do that under another PR

brikis98:
The tests passed, including the AWS ones. The only thing that failed was a lack of `go fmt` :)

Fixed here: https://github.com/gruntwork-io/terragrunt/commit/920ba4c1f2316e5283700e7fb915694e233669d8

Just cut a new release here: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.13.0. Assuming the build passes, the new binaries should be there in a few minutes. Can't wait to try it out. Thanks again @philsttr!

PR #270, opened 2017-08-25
Replace lock_table config option with dynamodb_table

### LINKED ISSUE #269: lock_table deprecated
The terraform remote state configuration directive `lock_table` for the S3 backend  is deprecated in favour of the new directive `dynamodb_table`

terraform throws the following warning when initialising the backend:

```
Initializing the backend...
"lock_table": [DEPRECATED] please use the dynamodb_table attribute
```

I tried using `dynamodb_table` in the terragrunt remote config but it is not recognised.

robinbowes:
Fixes #269.

Not tested at all.

Please review and offer feedback.

### DIFF
--- a/remote/remote_state_s3.go
+++ b/remote/remote_state_s3.go
@@ -15,12 +15,12 @@ import (
 
 // A representation of the configuration options available for S3 remote state
 type RemoteStateConfigS3 struct {
-	Encrypt   bool   `mapstructure:"encrypt"`
-	Bucket    string `mapstructure:"bucket"`
-	Key       string `mapstructure:"key"`
-	Region    string `mapstructure:"region"`
-	Profile   string `mapstructure:"profile"`
-	LockTable string `mapstructure:"lock_table"`
+	Encrypt       bool   `mapstructure:"encrypt"`
+	Bucket        string `mapstructure:"bucket"`
+	Key           string `mapstructure:"key"`
+	Region        string `mapstructure:"region"`
+	Profile       string `mapstructure:"profile"`
+	DynamoDBTable string `mapstructure:"dynamodb_table"`
 }
 
 const MAX_RETRIES_WAITING_FOR_S3_BUCKET = 12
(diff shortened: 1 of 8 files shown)


mcalhoun:
Thanks for the PR @robinbowes! It looks good to me, but you'll probably also want to update the [test](https://github.com/gruntwork-io/terragrunt/blob/1b0cb787c0314681e92f5ff083486eb5a988452f/test/integration_test.go#L523) as well

robinbowes:
Good spot. How's that look? If it wasn't 02:30 here I'd think about setting up a local go env and running the tests :)

remote/remote_state_s3.go
```diff
@@ -15,12 +15,12 @@ import (
...
-	Key       string `mapstructure:"key"`
-	Region    string `mapstructure:"region"`
-	Profile   string `mapstructure:"profile"`
-	LockTable string `mapstructure:"lock_table"`
+	Encrypt       bool   `mapstructure:"encrypt"`
+	Bucket        string `mapstructure:"bucket"`
+	Key           string `mapstructure:"key"`
+	Region        string `mapstructure:"region"`
+	Profile       string `mapstructure:"profile"`
+	DynamoDBTable string `mapstructure:"dynamodb_table"`
```
brikis98: I'm a little concerned about dropping the old param. Ideally, we'd follow Terraform's lead and report a deprecation warning for at least a little while. Or, even simpler, Terragrunt should accept both params and pass them on to Terraform unchanged, and Terraform itself would log the deprecation warning.

robinbowes: Yeah, I agree, but that's more work :)

I'll see if I can make some time to figure that out soon.

R.

chapati23: not a go expert but can't you just keep both params?

```go
LockTable string `mapstructure:"lock_table"`
DynamoDBTable string `mapstructure:"dynamodb_table`"
```

brikis98: Yup, keeping both is most likely the way to go. Also need to check both in the `createLockTableIfNecessary` method.

brikis98:
@robinbowes Any update on this?

robinbowes:
I'm afraid I've not had time to re-visit yet.

brikis98:
OK, I just opened #309 with the full fix, which also allows me to run tests in the PR. Closing this one now. Thanks!

Issue #251, opened 2017-07-05
plan-all hides errors


shaharmor:
Hey,

When using `plan-all`, if one of the internal modules has a parsing/validation error it will be hidden and not shown by `plan-all`.

Example:
For this module:
```hcl
resource "aws_route53_zone" "main" {
  name   = "${var.domain_name}"
  vpc_id = "${aws_vpc.main.id}"
}

resource "aws_route53_zone" "main" {
  name   = "${var.domain_name}"
  vpc_id = "${aws_vpc.main.id}"
}
```

There is an error because the two resources have the same name, but running `plan-all` won't show it. (Running `plan` will).

brikis98:
What does the command do in that case? Does it show any messages? Could you paste some example log output?

shaharmor:
The `plan` command shows this:

```
1 error(s) occurred:

* module root: 1 error(s) occurred:

* aws_route53_zone.main: resource repeated multiple times
```

`plan-all` shows this:

```
...
...
[terragrunt] [module/vpc] 2017/07/06 14:44:44 Running command: terraform plan -var-file=/env/test/vpc/../environment.tfvars
[terragrunt] [module/vpc] 2017/07/06 14:44:54 Module /env/test/vpc has finished with an error: exit status 1
[terragrunt] 2017/07/06 14:44:54 Encountered the following errors:
exit status 1
```

brikis98:
And the error doesn't appear anywhere in the log output above?

shaharmor:
Nope.
There is only 1 line after the plan command and I pasted it

brikis98:
I believe this was fixed by #277.

Issue #74, opened 2016-12-14
The apply-all and destroy-all commands should save log output to file


brikis98:
When you run `terragrunt apply-all` or `terragrunt destroy-all`, they can make changes across a ton of different modules, in parallel. This is great, but as mentioned in #71, the amount of log output is overwhelming, and it's easy to miss something important. One small step in helping with this is to store the log output for each module somewhere. Perhaps within the module or a tmp directory will do.

justyns:
Posting this here in case it helps someone else.  You can save the output of the terraform commands by making a wrapper around terraform and having terragrunt call that.

An example wrapper script:

``` shell
#!/bin/bash

terraform $@ 2>&1 | tee ./terraform.log
```

And then using it with terragrunt:
``` shell
terragrunt apply-all --terragrunt-non-interactive --terragrunt-tfpath terraform-wrapper
```

For me, this creates a `terraform.log` file in each directory where terragrunt plan/applied something.

Ideally terragrunt could do something like this by itself, but this workaround helps until that happens.

falmotlag:
### Solution
Delay Terraform output for all modules and display it formatted at the end of Terragrunt execution.

#### CLI option
Add a ```--delay-terraform-output``` cli option to turn on this feature.

### Code Change
    - Add a ```--delay-terraform-output``` cli option to turn on this feature.
    - in stack.Plan and stack.Apply assign a buffer to Writer
    - defer to function that print buffer of each module

Example snippet:
```go
if terragruntOptions.DelayOutput == true {
    outputStreams := make([]bytes.Buffer, len(stack.Modules))
    for n, module := range stack.Modules {
        module.TerragruntOptions.Writer = &outputStreams[n]
    }
    defer stack.printDelayedOutput(terragruntOptions, outputStreams)
}
```

Code to create cli flag and to ```printDelayedOutput``` is excluded from the snippet.

I am not too familiar with the codebase so I could be missing something here.
Can someone familiar with the codebase review this?

brikis98:
Thx for the proposal!

Unfortunately, I've often found that buffering log output for a long period of time is not a great solution for a few reasons:

1. With no output, the user has no idea what's going on. Is `apply` still running? Is there an error? What's being deployed? This leads to hitting `CTRL+C`, which is not a good combination with Terraform.
1. Some CI systems will kill a build if it goes more than N minutes without log output (e.g., in CircleCi, the limit is 10m). 

[...]

PR #277, opened 2017-08-31
Have test case for showing that plan-all swallows error messages

jmahowald:
When running terragrunt plan-all any errors in terraform are swallowed.  This is in contrast to running terraform plan in the offending subdir.  

This pull doesn't actually fix the behavior, but does add an integration test to demonstrate the issue.  Do you agree that this is change worth pursuing?  If so, any tips on where best to look to make the more specific change would be appreciated.

### DIFF
--- a/cli/args.go
+++ b/cli/args.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
@@ -15,7 +16,7 @@ import (
 
 // Parse command line options that are passed in for Terragrunt
 func ParseTerragruntOptions(cliContext *cli.Context) (*options.TerragruntOptions, error) {
-	terragruntOptions, err := parseTerragruntOptionsFromArgs(cliContext)
+	terragruntOptions, err := parseTerragruntOptionsFromArgs(cliContext.Args(), cliContext.App.Writer, cliContext.App.ErrWriter)
 	if err != nil {
 		return nil, err
 	}
@@ -33,8 +34,7 @@ func ParseTerragruntOptions(cliContext *cli.Context) (*options.TerragruntOptions
 // see: https://github.com/urfave/cli/issues/533. For now, our workaround is to dumbly loop over the arguments
 // and look for the ones we need, but in the future, we should change to a different CLI library to avoid this
 // limitation.
-func parseTerragruntOptionsFromArgs(cliContext *cli.Context) (*options.TerragruntOptions, error) {
-	args := cliContext.Args()
+func parseTerragruntOptionsFromArgs(args []string, writer, errWriter io.Writer) (*options.TerragruntOptions, error) {
 	currentDir, err := os.Getwd()
 	if err != nil {
 		return nil, errors.WithStackTrace(err)
@@ -77,13 +77,13 @@ func parseTerragruntOptionsFromArgs(cliContext *cli.Context) (*options.Terragrun
 	opts.NonInteractive = parseBooleanArg(args, OPT_NON_INTERACTIVE, false)
 	opts.TerraformCliArgs = filterTerragruntArgs(args)
 	opts.WorkingDir = filepath.ToSlash(workingDir)
-	opts.Logger = util.CreateLoggerWithWriter(cliContext.App.ErrWriter, "")
+	opts.Logger = util.CreateLoggerWithWriter(errWriter, "")
 	opts.RunTerragrunt = runTerragrunt
 	opts.Source = terraformSource
 	opts.SourceUpdate = sourceUpdate
 	opts.IgnoreDependencyErrors = ignoreDependencyErrors
-	opts.Writer = cliContext.App.Writer
-	opts.ErrWriter = cliContext.App.ErrWriter
+	opts.Writer = writer
+	opts.ErrWriter = errWriter
 	opts.Env = parseEnvironmentVariables(os.Environ())
 
 	return opts, nil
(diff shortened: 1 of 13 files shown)


brikis98:
Ooh, nice, love the TDD! 

Definitely worth fixing. Looks like #251 mentioned the same thing. #74 also mentions the need to save those logs to file, as reading them all in the console for a large number of modules being deployed in parallel is overwhelming.

I'm not sure where the bug is. Some likely starting points:

* There is logic here to capture error streams and summarize them: https://github.com/gruntwork-io/terragrunt/blob/master/configstack/stack.go#L33-L38. Probably something is broken with this logic here.

* Modules are executed here: https://github.com/gruntwork-io/terragrunt/blob/master/configstack/running_module.go#L187

* `TerragruntOptions`, which contains the logger, is cloned here: https://github.com/gruntwork-io/terragrunt/blob/master/configstack/module.go#L81

jmahowald:
I like the idea of having distinct files for the different streams, and I think I know how it could be tackled now, but this fix is a fair amount easier. There was just an if condition that resulted in errors not being printed if there was a very particular error message.

brikis98:
Ah, that fix makes sense. Do all the tests pass now?

jmahowald:
All the other tests pass.  Not the new one I created though.  Something about how stderr + logger are interacting in the tests that I haven't fully worked out yet.  I'll have to punt a bit on that part for now, but will hopefully get back to that this week.

jmahowald:
@brikis98 I would love to get this merged in before I start having more people using terragrunt , but the fact that my added test failed is preventing me from doing so in good conscience.  The problem is that I spent a couple of hours trying to manipulate stdout/stderr from the tests and just had to throw up my hands.  Is there another way of verifying this behavior, or perhaps you fixed it in another way (the part where plan-all would swallow error messages).

brikis98:
```go
output := stdout.String()
errOutput := stderr.String()
```

What do those contain when you run your new test?

jmahowald:
See output here
https://asciinema.org/a/xscN5iWx8Rcr1VNujBPtKRR4v

brikis98:
> See output here https://asciinema.org/a/xscN5iWx8Rcr1VNujBPtKRR4v

Hm, the stdout/stderr flashes by before I can actually see it in that video. Could you paste it into a [GitHub gist](https://gist.github.com)?

jmahowald:
https://gist.github.com/jmahowald/1431cc274ad982ad005c0ae9c0f626cb
It's the capturing of stderr in the test that is not working correctly.

brikis98:
Ah, I see. I *think* what might be happening is that `TerragruntOptions` sets its `Logger` field via a call to `util.CreateLogger` [here](https://github.com/gruntwork-io/terragrunt/blob/master/options/options.go#L98) and [here](https://github.com/gruntwork-io/terragrunt/blob/master/options/options.go#L134). The [CreateLogger](https://github.com/gruntwork-io/terragrunt/blob/master/util/logger.go#L14) function always logs to `os.Stderr`, rather than whatever other `Writer` you may have specified. I think you fix that method to take in a `Writer` and set it to `TerragruntOptions.ErrWriter`, you'll be all set.

jmahowald:
Thanks.  I had updated the Clone method and added in a CreateLoggerWithWriter function it was using.
I've checked that in, along with some other parts I can back out of when it's all working.

Still no dice, as I'm not sure how to get it in during the New call.  I'm not yet seeing the relationship between the App which does get an errWriter and the TerragruntOptions, but I'm suspecting that if I can figure that out/ a way to pass that errWriter into the options creation that it would be good.

Thanks

brikis98:
Looks like you're on the right track! Note my original message:

> Ah, I see. I *think* what might be happening is that `TerragruntOptions` sets its `Logger` field via a call to `util.CreateLogger` [here](https://github.com/gruntwork-io/terragrunt/blob/master/options/options.go#L98) and [here](https://github.com/gruntwork-io/terragrunt/blob/master/options/options.go#L134).

I think you only updated one of those two locations with your new `CreateLoggerWithWriter` function!

jmahowald:
All clear now.  Thanks for al the help (and the great tool)

jmahowald:
Removed the binary, there is one in master though BTW.

I had indeed missed a test, but have fixed with one more minor refactoring.  All tests now pass.

```
$ go test ./...
?       github.com/gruntwork-io/terragrunt      [no test files]
?       github.com/gruntwork-io/terragrunt/aws_helper   [no test files]
ok      github.com/gruntwork-io/terragrunt/cli  12.991s
ok      github.com/gruntwork-io/terragrunt/config       0.023s
ok      github.com/gruntwork-io/terragrunt/configstack  0.028s
ok      github.com/gruntwork-io/terragrunt/dynamodb     21.271s
?       github.com/gruntwork-io/terragrunt/errors       [no test files]
?       github.com/gruntwork-io/terragrunt/options      [no test files]
ok      github.com/gruntwork-io/terragrunt/remote       0.018s
ok      github.com/gruntwork-io/terragrunt/shell        6.555s
ok      github.com/gruntwork-io/terragrunt/test 185.592s
?       github.com/gruntwork-io/terragrunt/test/helpers [no test files]
ok      github.com/gruntwork-io/terragrunt/util 0.020s
```

brikis98:
> Removed the binary, there is one in master though BTW.

Huh. So there is. Not sure when that happened, but thanks :)

> I had indeed missed a test, but have fixed with one more minor refactoring. All tests now pass.

Fantastic. Thank you! Merging now.

jmahowald:
@brikis98 Can you let me know when your next planned release is?  Trying to decide how much to jerry-rig in terms of creating my own build for our workstation pipeline.  If it's days, probably not, if it's weeks maybe so

Thanks

brikis98:
That's totally my bad. I meant to release this immediately, but just lost track of it. New release here: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.13.3

brikis98:
(binaries are building now)

PR #278, opened 2017-08-31
Fix working dir

brikis98:
After the upgrade to support terraform 0.10.0, when Terragrunt ran `init` to download remote configurations, it was running `init` with the wrong working dir.


josh-padnick:
LGTM

PR #282, opened 2017-09-03
Ensure Terragrunt doesn't pollute stdout

brikis98:
Fix a bug where the stdout from “extra” commands executed by Terragrunt (e.g., calling `init`) would pollute the stdout with extra text you wouldn’t want. For example, if you called `terragrunt output X`, you would get the value for X, but also a bunch of extra output from Terragrunt automatically calling `init`.

We’ve had this issue before, so I’ve added a test case to try to prevent this regression from happening again.


brikis98:

PR #283, opened 2017-09-04
fixing typo

jmahowald:
In the docs on using optional_var_files there's a missing comma that would cause a message like

`[terragrunt] 2017/09/04 12:41:24 At 16:9: error parsing list, expected comma or list end, got: STRING`


brikis98:
Ah, good catch, thanks!

Issue #212, opened 2017-05-04
Terragrunt should detect if backend configuration is missing from Terraform code


brikis98:
As of Terraform 0.9, you have to declare a `backend` in your Terraform code, or remote state won't work:

```hcl
terraform {
  # The configuration for this backend will be filled in by Terragrunt
  backend "s3" {}
}
```

This is very easy to miss. See #211, #207, and #183. 

If the user specifies a `remote_state` configuration in their Terragrunt configuration, Terragrunt should scan the user's `.tf` files and report an error if a `backend` block is not found. This could be done using an approach similar to how we look for modules: https://github.com/gruntwork-io/terragrunt/blob/master/cli/cli_app.go#L254

hSATAC:
Thanks @brikis98 man you just saved my day. 💯

mazubieta:
Please include this, soon.  I just unwittingly deleted my local state files thinking that I had properly setup remote state.

Thankfully, I was able to recover via backups.

brikis98:
This should be fixed by #302. New binaries will be in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.13.5 shortly.

PR #302, opened 2017-09-29
Check that a backend is defined if remote_state settings are defined

brikis98:
This is a fix for #212.

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -509,8 +509,6 @@ func outputAll(terragruntOptions *options.TerragruntOptions) error {
 
 // Custom error types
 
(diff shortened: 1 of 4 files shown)


tamsky:
https://github.com/gruntwork-io/terragrunt/pull/302/files#diff-56be1e853ecf6e7ad64321692db32678R276

The grep test in this PR only examines files matching `**/*.tf` while ignoring `*.tf.json` files.
Both are equally valid places to declare the empty block it wants.

https://www.terraform.io/docs/configuration/

Across the years I've always declared the (now-required) empty stanza in a `common.tf.json` file.
```
    "terraform": {
        "backend": {
            "s3": {
            }
        }
    },
```
In order to upgrade terragrunt past this change, 
users will need to move their existing code in (`.tf.json`) files to (`.tf`) files.

I'm going to try fixing this now.

PR #308, opened 2017-10-05
Fix how Terragrunt checks for remote state initialization

### LINKED ISSUE #229: Terragrunt only creates Dynamo table if S3 bucket doesn't exist.
We had a dynamo locking table for a previous version of terragrunt, but after updating it didn't work anymore.  No sweat, I just deleted the dynamo table and terragrunt should re-create it, right?

Unfortunately, it looks like terragrunt only checks if the dynamo table exists after it creates the S3 bucket for remote state - it should always check for the dynamo table just like it always checks for the S3 bucket.

brikis98:
Originally, Terragrunt would only initialize remote state if remote
state wasn’t initialized at all or if the config has changed. Now it
will also query each “initializer” to see if it needs initialization.
The only initializer we have right now is the one for S3, and it will
now return true if either the configured S3 bucket or DynamoDB table
don’t exist. I believe this will fix #229.


brikis98:

PR #318, opened 2017-10-13
remote: Check whether input is a TTY before prompting

### LINKED ISSUE #317: Don't prompt for yes/no answer if input is not a TTY
We call Terragrunt from inside of a CI environment. At one point, Terragrunt stops and asks:

```
Remote S3 state bucket ... does not exist. Would you like Terragrunt to create it? (y/n) 
```

However, there's no way to answer that question, since it's running in a CI environment. Terragrunt can determine this at runtime by checking whether stdin is a tty; if not, Terragrunt should exit with an error.

(It's currently hanging indefinitely)

kevinburke:
Add a new CanAcceptPrompt API for determining whether the user can
type in things to send to the running process. If the user cannot do
that, exit with an error immediately, rather than hanging forever.

I haven't checked whether this code is used in other parts of the
program, but if so we should probably update those as well.

This work was sponsored by [Otto](https://meetotto.com).

Fixes #317.

### DIFF
--- a/remote/remote_state_s3.go
+++ b/remote/remote_state_s3.go
@@ -2,6 +2,8 @@ package remote
 
 import (
 	"fmt"
+	"time"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/gruntwork-io/terragrunt/aws_helper"
@@ -10,7 +12,6 @@ import (
 	"github.com/gruntwork-io/terragrunt/options"
 	"github.com/gruntwork-io/terragrunt/shell"
 	"github.com/mitchellh/mapstructure"
-	"time"
 )
 
 // A representation of the configuration options available for S3 remote state
@@ -139,6 +140,9 @@ func validateS3Config(config *RemoteStateConfigS3, terragruntOptions *options.Te
 // confirms, create the bucket and enable versioning for it.
 func createS3BucketIfNecessary(s3Client *s3.S3, config *RemoteStateConfigS3, terragruntOptions *options.TerragruntOptions) error {
 	if !DoesS3BucketExist(s3Client, config) {
+		if !shell.CanAcceptPrompt() {
+			return fmt.Errorf("remote: S3 bucket does not exist, or you don't have permissions to access it: %q", config.Bucket)
+		}
 		prompt := fmt.Sprintf("Remote state S3 bucket %s does not exist or you don't have permissions to access it. Would you like Terragrunt to create it?", config.Bucket)
 		shouldCreateBucket, err := shell.PromptUserForYesNo(prompt, terragruntOptions)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


josh-padnick:
This is definitely a nifty way of checking for a prompt, but it also breaks the interactive experience for human CLI users. Let's resolve #317 and perhaps then we can revisit this.

kevinburke:
> it also breaks the interactive experience for human CLI users

I'm a little confused. If a human is using the CLI, terminal.IsTerminal will return true, and we'll skip the if branch. If there's no TTY hooked up to stdin, terminal.IsTerminal will return false and you'll get an error.

josh-padnick:
Ah, that's my mistake. Thanks for clarifying.

shell/prompt.go
```diff
@@ -3,12 +3,20 @@ package shell
...
+
+	"github.com/gruntwork-io/terragrunt/errors"
+	"github.com/gruntwork-io/terragrunt/options"
+	"golang.org/x/crypto/ssh/terminal"
 )
 
+// CanAcceptPrompt tells whether the program can receive input on the standard
+// input.
+func CanAcceptPrompt() bool {
+	return terminal.IsTerminal(int(os.Stdin.Fd()))
```
brikis98: Does this work on Windows?

remote/remote_state_s3.go
```diff
@@ -139,6 +140,9 @@ func validateS3Config(config *RemoteStateConfigS3, terragruntOptions *options.Te
 // confirms, create the bucket and enable versioning for it.
 func createS3BucketIfNecessary(s3Client *s3.S3, config *RemoteStateConfigS3, terragruntOptions *options.TerragruntOptions) error {
 	if !DoesS3BucketExist(s3Client, config) {
+		if !shell.CanAcceptPrompt() {
+			return fmt.Errorf("remote: S3 bucket does not exist, or you don't have permissions to access it: %q", config.Bucket)
+		}
```
brikis98: A couple lines below this, we call `shell.PromptUserForYesNo`. I think all the user prompts go through that `shell` package. It may make more sense to put this check into that method, which, BTW, also handles the `--terragrunt-non-interactive` flag, which has the opposite behavior (assumes "yes" for all prompts).

kevinburke:
This is blocked on #335 (and I guess #336).

shell/prompt.go
kevinburke: There is code that at least attempts to handle this from the Windows side... from the source code

```go
// IsTerminal returns true if the given file descriptor is a terminal.
func IsTerminal(fd int) bool {
	var st uint32
	err := windows.GetConsoleMode(windows.Handle(fd), &st)
	return err == nil
}
```

I have no idea whether that is the correct code, however, as I don't have a Windows machine.

It seems like the consequence of getting this wrong is not that big, though? If we incorrectly return true, we will wait to prompt when a Windows user can't enter any data. If we incorrectly return false, we will return an error, instead of prompting the user to enter a bucket.

brikis98: Fair enough!

brikis98:
PR not updated for over 2 years, closing

PR #320, opened 2017-10-16
Allow specifying a parameter for find_in_parent_folders

brikis98:
This PR extends the `find_in_parent_folders()` function so that it can accept an optional parameter:

```hcl
foo = "${find_in_parent_folders("foo.txt")}"
```

If you don’t specify a parameter, it defaults to looking for `terraform.tfvars`, just like it always has.


brikis98:

PR #322, opened 2017-10-17
Add support for role_arn

### LINKED ISSUE #290: Use role_arn in S3 backend
Hi,

I'm using cross-account roles with clients.
I can assume a role with terraform s3 backend config but terragrunt seems to ignore the [role_arn ](https://www.terraform.io/docs/backends/types/s3.html#role_arn) parameter.

brikis98:
Fixes #290. If you specify the `role_arn` parameter in your S3 backend config, Terragrunt will now automatically assume that IAM role before configuring remote state, creating the S3 bucket, creating the DynamoDB table, etc. This is very useful for multi-account setups where you may need to assume an IAM role in another account.

### DIFF
--- a/aws_helper/config.go
+++ b/aws_helper/config.go
@@ -2,25 +2,31 @@ package aws_helper
 
 import (
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/gruntwork-io/terragrunt/errors"
 )
 
-// Returns an AWS session object for the given region, ensuring that the credentials are available
-func CreateAwsSession(awsRegion, awsProfile string) (*session.Session, error) {
-	session, err := session.NewSessionWithOptions(session.Options{
+// Returns an AWS session object for the given region (required), profile name (optional), and IAM role to assume
+// (optional), ensuring that the credentials are available
+func CreateAwsSession(awsRegion, awsProfile string, iamRoleArn string) (*session.Session, error) {
+	sess, err := session.NewSessionWithOptions(session.Options{
 		Config:            aws.Config{Region: aws.String(awsRegion)},
 		Profile:           awsProfile,
 		SharedConfigState: session.SharedConfigEnable,
 	})
 	if err != nil {
-		return nil, errors.WithStackTraceAndPrefix(err, "Error intializing session")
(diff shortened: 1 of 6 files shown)


FernandoMiguel:
I'm quite interested in this addition.
In our setup, we run terraform from a jenkins host. that EC2 instance has an EC2 role attached that grants it access to a bucket in that account for *all* state files

different parts of terraform will then build/manage resources in different accounts.

What does this mean for terragrunt ?
Will the buckets be created in each account that it assumes the role? 
or will it be in the jenkins account? do I need a policy to allow the other accounts to write back to jenkins account?

thanks

brikis98:
If you specify a `role_arn` parameter in your backend configuration, (a) Terragrunt will assume that IAM role when checking to see if it needs to create an S3 bucket or DynamoDB table and (b) Terraform will assume that IAM role when using the S3 bucket and DynamoDB table. If that IAM role lives in another AWS account, that means the S3 bucket and DynamoDB table must be in that account too and that IAM role must have access to them.

FernandoMiguel:
yeah, thank you for confirming.

I'll have to create policies to handle this, and wont get to benefit from terragrunt to create the buckets on init :( 

maybe a feature for the future

brikis98:
> and wont get to benefit from terragrunt to create the buckets on init :(

Wait, why is that? Can you describe your use case a bit more?

FernandoMiguel:
Well, terragrunt has the ability to create the state files on s3 when the user is creating a new environment.
Since I'm creating many accounts with several envs in each in DRY format, it becomes simpler than having a terraform creating the few buckets we need and than using them for each module

brikis98:
Sorry, I'm still not following. Is your goal to use a single S3 bucket for all AWS accounts? Or a separate bucket for each AWS account?

If you just want a single bucket in a single account, then either don't specify `role_arn` at all or specify the IAM role for that one account, and Terragrunt will create the bucket in it (if it doesn't already exist). If you want a bucket in each account, then specify a different `role_arn` for each of those accounts, and Terragrunt will assume an IAM role in that account and create the bucket in it (if it doesn't already exist). You can even read `role_arn` from environment variables if you want to make it easy to modify:

```hcl
terragrunt = {
  remote_state {
    backend = "s3" 
    config {
      bucket = "..."
      role_arn = "${get_env("ROLE_ARN", "some-fallback-value")}"
    } 
  }
}
```

patkar:
@brikis98 Thank you so much for the addition, this was the final piece missing to use terragrunt

PR #335, opened 2017-10-31
Update glide.lock

kevinburke:
Generated this file by creating a new Git branch and running "glide
update". While none of the revision hashes changed, the subpackage
lists changed quite a bit, which leaves me a little confused.

Generated using glide version 0.13.0-dev.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,41 +1,43 @@
 hash: 3ec23ed9524b96844ab5cef37f5be2323a36cf7aed4a06fdc9b102b39c899d1e
-updated: 2017-10-18T17:34:37.848019211+01:00
+updated: 2017-10-31T12:33:55.087706-07:00
 imports:
 - name: github.com/aws/aws-sdk-go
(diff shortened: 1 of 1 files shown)


kevinburke:
Trying to merge this separately because I ran into a conflict in the `.lock` file attempting to add a dependency for #318.

brikis98:
Hmm, that is weird. I find glide very confusing :-\

Do the tests still pass?

kevinburke:
Not sure, see https://github.com/gruntwork-io/terragrunt/issues/336.

kevinburke:
Another option would be to check in the entire vendor directory, which would make the changes here more obvious, as they'd appear in the diff.

brikis98:
I'm not a big fan of checking in thousands of random Go src files. In fact, the sheer volume of code would likely make that diff even less useful than the `glide.lock` one.

brikis98:
Running tests on this...

brikis98:
These tests passed. Merging now.

Issue #336, opened 2017-10-31
Tests require clean AWS credentials


kevinburke:
The tests require AWS credentials that don't have access to other services (in order to be safe). This can be painful and/or expensive to set up.

It would be great if the AWS-specific tests used the `testing.Short()` helper to exit early if you run in short mode.

brikis98:
+1

yhakbar:
We use build flags to prevent them from running by default now.

Issue #334, opened 2017-10-31
Use a more permanent cache directory


kevinburke:
Terragrunt calls `os.TempDir` to allocate a temporary directory for downloading local files. However, this is vulnerable to getting cleaned up by the operating system at any time.

The purpose of the temp dir is to operate as a local cache and speed up operations.

A directory like `$HOME/Library/Caches` (on Mac), `$XDG_CACHE_HOME/terragrunt` (if specified), or `$HOME/.terragrunt` may be more appropriate, and less vulnerable to getting cleaned up.

In addition, Terragrunt cannot detect if the cache is corrupt. It would be good to check whether the cache is corrupt and trigger a remove/redownload, or at the very least report a helpful error, instead of exiting with an unhelpful error.

brikis98:
This is a great idea!

PR #338, opened 2017-10-31
Remove subdirectories before copying files to them

### LINKED ISSUE #337: Permission error when trying to copy read-only file
downloadTerraformSource works by downloading Terraform, then copying files from the terragrunt options working directory to the terraform source directory.

I have files (TLS certificates) in that directory that have permissions set to 444. Eventually we try to overwrite those files (with the exact same contents) by calling `util.CopyFile`:

```go
// Copy a file from source to destination
func CopyFile(source string, destination string) error {
	contents, err := ioutil.ReadFile(source)
	if err != nil {
		debug.PrintStack()
		fmt.Printf("source: %s, dest: %s, copy err: %#v\n", source, destination, err)
	}

	return WriteFileWithSamePermissions(source, destination, contents)
}
```

This calls `ioutil.WriteFile`:

```go
func WriteFileWithSamePermissions(source string, destination string, contents []byte) error {
	fileInfo, err := os.Stat(source)
	if err != nil {
	}

	return ioutil.WriteFile(destination, contents, fileInfo.Mode())
}
```

Which fails with an error, because the first thing it attempts to do is open the destination file for writing:

```go
// WriteFile writes data to a file named by filename.
  // If the file does not exist, WriteFile creates it with permissions perm;
  // otherwise WriteFile truncates it before writing.
  func WriteFile(filename string, data []byte, perm os.FileMode) error {
  	f, err := os.OpenFile(filename, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, perm)
  	if err != nil {
  		return err
  	}
  	n, err := f.Write(data)
  	if err == nil && n < len(data) {
  		err = io.ErrShortWrite
  	}
  	if err1 := f.Close(); err == nil {
  		err = err1
  	}
  	return err
  }
```

This is blocking deployment.

kevinburke:
If a file in a subdirectory has permissions of 444, we will attempt to
open it for writing in util.CopyFile before overwriting it. This
fails because the file is read-only.

Instead remove all files in subdirectories. This shouldn't be much
slower than the current procedure, because we are still overwriting
every file every time, we're just starting from an empty file tree
instead of a full file tree. This also avoids the problem where run
1 produces file A, and run 2 does not have file A, but it exists in
the cache.

Fixes #337.

### DIFF
--- a/util/file.go
+++ b/util/file.go
@@ -1,13 +1,13 @@
 package util
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
 
-	"fmt"
 	"github.com/gruntwork-io/terragrunt/errors"
 	"github.com/mattn/go-zglob"
 )
@@ -146,6 +146,11 @@ func CopyFolderContents(source string, destination string) error {
 		if PathContainsHiddenFileOrFolder(src) {
 			continue
 		} else if file.IsDir() {
+			// Permission and stale cache problems otherwise; see
+			// https://github.com/gruntwork-io/terragrunt/issues/337
+			if err := os.RemoveAll(dest); err != nil {
+				return errors.WithStackTrace(err)
+			}
 			if err := os.MkdirAll(dest, file.Mode()); err != nil {
 				return errors.WithStackTrace(err)
 			}
--- a/cli/download_source.go
+++ b/cli/download_source.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"net/url"
 	"os"
+	"path/filepath"
 	"regexp"
 	"strings"
 
@@ -15,7 +16,6 @@ import (
 	"github.com/hashicorp/go-getter"
 	urlhelper "github.com/hashicorp/go-getter/helper/url"
 	"github.com/mattn/go-zglob"
-	"path/filepath"
 )
 
 // This struct represents information about Terraform source code that needs to be downloaded


kevinburkeotto:
This is blocked on #335 (and I guess #336), or you adding me as a contributor to the project, you removing the requirement to have tests pass before merging changes, or you pulling down and running the tests manually in a clean AWS account.

brikis98:
Running tests locally now...

brikis98:
A bunch of tests failed. Some of the errors look like this:

```
[terragrunt] [fixture-old-terragrunt-config/download] 2017/10/31 23:36:58 Running command: terraform init -from-module=file:///Users/brikis98/source/go/src/github.com/gruntwork-io/terragrunt/test/fixture-old-terragrunt-config/hello-world /var/folders/39/w935myd14p9197wbp67pz4940000gn/T/terragrunt/lZQRK2TqaEUhn3UvKhF5fNxpljk/jjLRfkf7pYvGKBwdLKFuRON8uLo
Copying configuration from "file:///Users/brikis98/source/go/src/github.com/gruntwork-io/terragrunt/test/fixture-old-terragrunt-config/hello-world"...
Downloading modules...

Plugin reinitialization required. Please run "terraform init".
Reason: Could not satisfy plugin requirements.

Plugins are external binaries that Terraform uses to access and manipulate
resources. The configuration provided requires plugins which can't be located,
don't satisfy the version constraints, or are otherwise incompatible.

1 error(s) occurred:

* provider.template: new or changed plugin executable

Terraform automatically discovers provider requirements from your
configuration, including providers used in child modules. To see the
requirements and constraints from each module, run "terraform providers".

error satisfying plugin requirements
```

And some like this:

```
Error downloading modules: Error loading modules: module remote: No Terraform configuration files found in directory: .terraform/modules/3de0b4d8458c5b95796bcc1ab4263679
```

I'm guessing what's happening is the following:

1. Terragrunt downloads the code into a tmp folder. 
1. Terragrunt calls `init` in that tmp folder to download all the modules, providers, and initialize remote state. All of this Terraform stores in a `.terraform` folder within that tmp folder.
1. Terragrunt calls the updated `CopyFolderContents` method, which deletes everything in the folder, including the `.terraform` folder. 
1. Now when the `plan` or `apply` command are executed, none of the modules are providers are there, and you get an error.

I think you may have to delete the individual files rather than the hold folder.

kevinburkeotto:
Are you sure that isn't this problem? https://github.com/gruntwork-io/terragrunt/issues/334

Do the tests pass on master?

brikis98:
Yes, tests pass on master. I'm fairly confident the problem is as I described it above.

kevinburkeotto:
ok

brikis98:
BTW, I'm running these tests locally because I know these fixes are important to you, and I appreciate your help in making them, but setting up an AWS account where you have S3/DynamoDB permissions is about 5 min of work, so please try to run the tests after the next change!

kevinburkeotto:
I can't reproduce the test failures you saw.

kevinburkeotto:
To clarify, I ran the tests once, then ran again with 

```TF_LOG=debug aws-vault exec testing -- go test -timeout=20m -count=4 ./...```

for a total of 5 test runs, and couldn't reproduce the failure any time.

**Comment by brikis98** [comment]
Huh. I just checked out this branch again, and the tests passed. I'm... stumped.

File: util/file.go [hunk 1]
```diff
@@ -146,6 +146,11 @@ func CopyFolderContents(source string, destination string) error {
 		if PathContainsHiddenFileOrFolder(src) {
 			continue
 		} else if file.IsDir() {
+			// Permission and stale cache problems otherwise; see
+			// https://github.com/gruntwork-io/terragrunt/issues/337
+			if err := os.RemoveAll(dest); err != nil {
+				return errors.WithStackTrace(err)
+			}
```
brikis98: Is this the right fix for the permissions issue you were seeing?

I'm looking at how `CopyFolderContents` is used in [download_source](https://github.com/gruntwork-io/terragrunt/blob/0a61ebb2924ac78077f92a29123186b316126635/cli/download_source.go#L59-L61):

```go
util.CopyFolderContents(terragruntOptions.WorkingDir, terraformSource.WorkingDir)
``` 

Let's say you ran Terragrunt from a folder `foo`. If `foo` contained another folder `bar`, then this fix would kick in, deleting `bar`, and taking care of any possible permissions issues. But if `foo` contained a file named `baz`, that file would not be deleted from the previous run (since `file.IsDir()` would be false), and if had some restrictive permission, it would not be overwritten.

So it seems like you would need an `os.Remove` call in the `else` block too, right?

brikis98: Also, could you add an automated test for these cases?

brikis98:
PR not updated for over 2 years, closing

Issue #332, opened 2017-10-30
can't get get_parent_tfvars_dir() to work


ebarault:
Consider the following project:

```
/terragrunt live folder
├── terraform.tfvars <-- root
├── prd
│   ├── app1
│   |   └── terraform.tfvars <-- child
```

I'd like to access the my modules dir located on step above the folder where I have the **root** `terraform.tfvars` using `get_parent_tfvars_dir()` as in:
```source = "${get_parent_tfvars_dir()}/../modules//app"```

but `get_parent_tfvars_dir()` keeps returning the absolute location of the **child** `terraform.tfvars` `/path/to/live/prd/app1` instead of returning `/path/to/`.

Is there anything i'm doing wrong ?

**Comment by brikis98** [comment]
Do you have an `include { ... }` block in the child `terraform.tfvars`?

**Comment by ebarault** [comment]
Yes, I have the classic
```
terragrunt = {
  include {
    path = "${find_in_parent_folders()}"
  }
#...
}

brikis98:
You're right, this is a bug. I was just able to repro with your exact folder structure and the following contents in `prd/app1/terraform.tfvars`:

```hcl
 terragrunt = {
   include {
     path = "${find_in_parent_folders()}"
   }
 
   terraform {
     extra_arguments "test" {
       commands = ["plan", "apply"]
       arguments = ["-var", "foo=${get_parent_tfvars_dir()}"]
     }
   }
 }
```

Not sure when this broke, as there are tests for it. Sorry about that!

ebarault:
👍 Thanks for confirming. Unfortunately I don't know `go`... and therefore can't help you much here.

brikis98:
Understood. I'm a bit buried at the moment, so I'm not sure when I'll be able to get to this one. @jocgir Any chance you have a spare few minutes to take a look, since I believe you were the one who originally added `get_parent_tfvars_dir()`?

jocgir:
Hi, I'm a bit buried too, but I will take time to give it a quick look.

brikis98:
Thanks!

jocgir:
I found what is wrong. The `get_parent_tfvars_dir()` function has been designed to be used in the parent `terraform.tfvars`.

When the child `terraform.tfvars` configuration file is evaluated, it does not have the full context of its parent yet. So, these functions are only useful when used in the parent configuration file to retrieve the relative position of the child.

An example of utilization would be:

`
[...]

PR #339, opened 2017-11-01
Fix the situation where get_parent_tfvars() is called from within the…

jocgir:
… child configuration file. Also apply to path_relative_from_include().

Add test case.

Solve issue #332

### DIFF
--- a/options/options.go
+++ b/options/options.go
@@ -76,6 +76,9 @@ type TerragruntOptions struct {
 	// packages can use the command without a direct reference back to the cli package (which would create a
 	// circular dependency).
 	RunTerragrunt func(*TerragruntOptions) error
+
+	// Keep the last found parent folder to be able to return it when get_parent_tfvars() is called from within the child
+	LastIncludedParentFolder string
 }
 
 // Create a new TerragruntOptions object with reasonable defaults for real usage
--- a/config/config_helpers.go
+++ b/config/config_helpers.go
@@ -241,12 +241,15 @@ func findInParentFolders(parameters string, terragruntOptions *options.Terragrun
 		return "", errors.WithStackTrace(err)
 	}
 
+	configDir := filepath.Dir(terragruntOptions.TerragruntConfigPath)
+
 	// To avoid getting into an accidental infinite loop (e.g. do to cyclical symlinks), set a max on the number of
 	// parent folders we'll check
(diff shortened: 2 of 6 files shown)


options/options.go
```diff
@@ -76,6 +76,9 @@ type TerragruntOptions struct {
 	// packages can use the command without a direct reference back to the cli package (which would create a
 	// circular dependency).
 	RunTerragrunt func(*TerragruntOptions) error
+
+	// Keep the last found parent folder to be able to return it when get_parent_tfvars() is called from within the child
+	LastIncludedParentFolder string
```
brikis98: I'm not sure it makes sense to put this into `TerragruntOptions` and to modify at random points in the code. It's too hard to reason about.

We have the path of the parent folder while parsing the Terragrunt configuration. We should just save it! That is, update the [TerragruntConfig](https://github.com/gruntwork-io/terragrunt/blob/master/config/config.go#L20) struct with a `ParentPath` field. It should be set when we're working with the [Include](https://github.com/gruntwork-io/terragrunt/blob/master/config/config.go#L34) field in the `terragruntConfigFile` struct. 

After that, the `get_parent_tfvars()` could just return this field with no further work.

jocgir: It doesn't seem to be so simple. The problem is that we do not have the TerragruntConfig object when we evaluate the function.

Any suggestion.

brikis98: Actually, don't you have access to `include *IncludeConfig`? IIRC, if that isn't `nil`, it contains the path of the parent config file...

jocgir: In the context of the child, include is nil. It is the reason why this function is only working on the parent side and why I made this change to solve that problem.

brikis98: Ah, I gotcha.

I think the proper fix will require a bit of a refactor to the `parseConfigString` method. Right now, it tries to process interpolations in the string and _then_ parse it into a `terragruntConfigFile` struct. The problem is that to process those interpolations, you need info from that struct, such as any `include` setting.

So probably the better parsing approach would be to:

1. Parse the string into a struct that has just the top-level elements: `remote_state`, `include`, `terraform`. The contents of those elements are left "raw".

1. Resolve interpolations in each of those top-level elements, starting first with `include`, and make the struct available to the interpolation functions themselves.

PR #340, opened 2017-11-01
Chore/use $HOME/.terragrunt for downloadDir [resolves #334]

### LINKED ISSUE #334: Use a more permanent cache directory

mcalhoun:
Use `$HOME` for `downloadDir` per https://github.com/gruntwork-io/terragrunt/issues/334


mcalhoun:
@brikis98 can you take a look?

options/options.go
```diff
@@ -82,7 +83,13 @@ type TerragruntOptions struct {
...
 	workingDir := filepath.Dir(terragruntConfigPath)
 
-	downloadDir := filepath.Join(os.TempDir(), "terragrunt")
+	user, err := user.Current()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error: %v\n", err)
+		os.Exit(1)
+	}
+
+	downloadDir := filepath.Join(user.HomeDir, ".terragrunt")
```
brikis98: Unfortunately, this approach makes cross-compiling harder. In the past, I've used https://github.com/mitchellh/go-homedir.

options/options.go
```diff
@@ -82,7 +83,13 @@ type TerragruntOptions struct {
 func NewTerragruntOptions(terragruntConfigPath string) *TerragruntOptions {
 	workingDir := filepath.Dir(terragruntConfigPath)
 
-	downloadDir := filepath.Join(os.TempDir(), "terragrunt")
+	user, err := user.Current()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error: %v\n", err)
+		os.Exit(1)
```
brikis98: I'm not comfortable with a random method exiting. Code is much more predictable if all errors are returned. I'm sure `NewTerragrunOptions` is used in a bunch of places, but if the method can now have an error, it needs to reflect that in its signature.

options/options.go
```diff
@@ -82,7 +83,13 @@ type TerragruntOptions struct {
 func NewTerragruntOptions(terragruntConfigPath string) *TerragruntOptions {
 	workingDir := filepath.Dir(terragruntConfigPath)
 
-	downloadDir := filepath.Join(os.TempDir(), "terragrunt")
+	user, err := user.Current()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error: %v\n", err)
```
brikis98: Just a few lines below, we call `util.CreateLogger()`. Move that up here and use it to do the logging instead of `fmt.Fprintf`.

glide.lock
```diff
@@ -1,8 +1,8 @@
 hash: 3ec23ed9524b96844ab5cef37f5be2323a36cf7aed4a06fdc9b102b39c899d1e
-updated: 2017-10-31T12:33:55.087706-07:00
+updated: 2017-10-31T23:56:12.659155168-04:00
```
brikis98: Why a glide update for this change?

options/options.go
```diff
@@ -79,10 +80,18 @@ type TerragruntOptions struct {
...
-func NewTerragruntOptions(terragruntConfigPath string) *TerragruntOptions {
+func NewTerragruntOptions(terragruntConfigPath string) (*TerragruntOptions, error) {
 	workingDir := filepath.Dir(terragruntConfigPath)
 
-	downloadDir := filepath.Join(os.TempDir(), "terragrunt")
+	logger := util.CreateLogger("")
+
+	homedir, err := homedir.Dir()
+	if err != nil {
+		logger.Panicf("error: %v\n", err)
```
brikis98: This is a bit of a gotcha with golang: `Panicf` logs your message and then calls `panic`. So, yea, don't use it :)

options/options.go
```diff
@@ -79,10 +80,18 @@ type TerragruntOptions struct {
...
+		return nil, err
```
brikis98: Wrap with `errors.WithStackTrace`.

options/options.go
```diff
@@ -109,12 +118,17 @@ func NewTerragruntOptions(terragruntConfigPath string) *TerragruntOptions {
...
 }
 
 // Create a new TerragruntOptions object with reasonable defaults for test usage
 func NewTerragruntOptionsForTest(terragruntConfigPath string) *TerragruntOptions {
-	opts := NewTerragruntOptions(terragruntConfigPath)
+	logger := util.CreateLogger("")
+
+	opts, err := NewTerragruntOptions(terragruntConfigPath)
+	if err != nil {
+		logger.Panicf("error: %v\n", err)
```
brikis98: This results in a `panic`. I think this method would need to return an error obj too :-\

options/options.go
mcalhoun: Another reason I need to do more go development...I think I knew that and forgot... :)

mcalhoun:
ok @brikis98 I think I've got it all worked out now...I'd love another review... ;)

cli/args_test.go
```diff
@@ -147,7 +147,11 @@ func assertOptionsEqual(t *testing.T, expected options.TerragruntOptions, actual
 }
 
 func mockOptions(terragruntConfigPath string, workingDir string, terraformCliArgs []string, nonInteractive bool, terragruntSource string, ignoreDependencyErrors bool) *options.TerragruntOptions {
-	opts := options.NewTerragruntOptionsForTest(terragruntConfigPath)
+	logger := util.CreateLogger("")
+	opts, err := options.NewTerragruntOptionsForTest(terragruntConfigPath)
+	if err != nil {
+		logger.Panicf("error: %v\n", errors.WithStackTrace(err))
```
brikis98: I know you're trying to minimize yak shaving, but calling `panic` during tests is still problematic. I'd recommend adding a `t *testing.T` param to this test and calling `t.FatalF` here if you get an error.

configstack/module_test.go
```diff
@@ -325,7 +325,7 @@ func TestGetTerragruntSourceForModuleErrors(t *testing.T) {
 }
 
 func mockOptionsWithSource(sourceUrl string) *options.TerragruntOptions {
-	opts := options.NewTerragruntOptionsForTest("mock-for-test.tfvars")
+	opts, _ := options.NewTerragruntOptionsForTest("mock-for-test.tfvars")
```
brikis98: I would not ignore this error. It'll lead to confusing test failures. Again, add a `t *testing.T` to this method and call `t.FatalF` in case of error.

configstack/running_module_test.go
```diff
@@ -8,7 +8,7 @@ import (
 	"testing"
 )
 
-var mockOptions = options.NewTerragruntOptionsForTest("running_module_test")
+var mockOptions, _ = options.NewTerragruntOptionsForTest("running_module_test")
```
brikis98: I would not ignore this error. It'll lead to confusing test failures. This may have to be converted to a `mockOptions(t *testing.T)` function that returns a `TerragruntOptions` struct...

configstack/test_helpers.go
```diff
@@ -157,7 +157,7 @@ func state(t *testing.T, bucket string, key string) *remote.RemoteState {
 // Create a mock TerragruntOptions object and configure its RunTerragrunt command to return the given error object. If
 // the RunTerragrunt command is called, this method will also set the executed boolean to true.
 func optionsWithMockTerragruntCommand(terragruntConfigPath string, toReturnFromTerragruntCommand error, executed *bool) *options.TerragruntOptions {
-	opts := options.NewTerragruntOptionsForTest(terragruntConfigPath)
+	opts, _ := options.NewTerragruntOptionsForTest(terragruntConfigPath)
```
brikis98: Same thing here. Don't ignore errors. It always comes back to haunt you!

dynamodb/dynamo_lock_test_utils.go
```diff
@@ -19,7 +19,7 @@ func init() {
 	rand.Seed(time.Now().UnixNano())
 }
 
-var mockOptions = options.NewTerragruntOptionsForTest("dynamo_lock_test_utils")
+var mockOptions, _ = options.NewTerragruntOptionsForTest("dynamo_lock_test_utils")
```
brikis98: Same here :(

options/options.go
```diff
@@ -79,10 +80,18 @@ type TerragruntOptions struct {
...
-func NewTerragruntOptions(terragruntConfigPath string) *TerragruntOptions {
+func NewTerragruntOptions(terragruntConfigPath string) (*TerragruntOptions, error) {
 	workingDir := filepath.Dir(terragruntConfigPath)
 
-	downloadDir := filepath.Join(os.TempDir(), "terragrunt")
+	logger := util.CreateLogger("")
+
+	homedir, err := homedir.Dir()
+	if err != nil {
+		logger.Fatalf("error: %v\n", err)
```
brikis98: So... From the [FatalF docs](https://golang.org/pkg/log/#Fatalf):

```
Fatalf is equivalent to Printf() followed by a call to os.Exit(1).
```

Did I ever tell you that Go is a shitty language?

brikis98: Use a `PrintF`. In a separate PR, we should move Terragrunt off the default Go logging library and onto something that has proper `Info`, `Error`, etc methods.

options/options.go
```diff
@@ -109,16 +118,22 @@ func NewTerragruntOptions(terragruntConfigPath string) *TerragruntOptions {
...
 
 // Create a new TerragruntOptions object with reasonable defaults for test usage
-func NewTerragruntOptionsForTest(terragruntConfigPath string) *TerragruntOptions {
-	opts := NewTerragruntOptions(terragruntConfigPath)
+func NewTerragruntOptionsForTest(terragruntConfigPath string) (*TerragruntOptions, error) {
+	logger := util.CreateLogger("")
+
+	opts, err := NewTerragruntOptions(terragruntConfigPath)
+	if err != nil {
+		logger.Fatal("error: %v\n", errors.WithStackTrace(err))
```
brikis98: As per the above, replace `Fatal` with `PrintF`.

test/integration_test.go
```diff
@@ -57,7 +57,7 @@ const (
 	TERRAFORM_STATE_BACKUP                              = "terraform.tfstate.backup"
 )
 
-var mockOptions = options.NewTerragruntOptionsForTest("integration_test")
+var mockOptions, _ = options.NewTerragruntOptionsForTest("integration_test")
```
brikis98: One more...

brikis98:
@mcalhoun Are you planning on finishing this up? It's so close...

mcalhoun:
Yep @brikis98, I'm going to finish it today

mcalhoun:
ok @brikis98 I think I've finally got this all cleaned up

brikis98:
Thanks Matt. That was a lot of yak shaving for such a seemingly simple fix!

PR #341, opened 2017-11-02
Update docs about Terragrunt caching

brikis98:
Based on feedback from #311.


josh-padnick:
This is a great addition!

Issue #125, opened 2017-02-14
release-lock fails to parse .terragrunt file due to env_var function call: Invalid Interpolation Syntax


umeat:
When attempting to run `terragrunt release-lock` I get the following exception:

> [terragrunt] 2017/02/14 01:39:23 Reading Terragrunt config file at ***/.terragrunt  
>Error parsing Terragrunt config file ***/.terragrunt: Error parsing Terragrunt config file ***/terraform/.terragrunt: Invalid interpolation syntax. Expected syntax of the form '${function_name()}', but got '${get_env("BUCKET_NAME", "")}'   

Where that .terragrunt file contains:

```
remote_state = {
  backend = "s3"
  config {
    encrypt = "true"
    bucket = "${get_env("BUCKET_NAME", "")}"
    key = "${path_relative_to_include()}/terraform.tfstate"
    region = "ap-southeast-2"
  }
}
```

`terragrunt plan` and `apply` both parse this config file without issue.

brikis98:
Weird. I think `plan`, `apply`, and `release-lock` all go through the same code path for parsing...

What version of Terragrunt?

l0010o0001l:
@umeat 

You may be using an outdated version of Terragrunt. I am on latest, and after testing your bucket syntax I had no issues acquiring a lock. The obvious difference I see here is that you are using a `.terragrunt` file which is soon to be deprecated. Try moving your `terragrunt` variables into a `terraform.tfvars` file and give it another go.

brikis98:
No follow up on the issue. Moreover, since Terraform now natively supports locking, Terragrunt no longer does locking, and no longer has a `release-lock` command, so closing this issue.

PR #356, opened 2017-11-09
State locking no longer required or supported?

mcameron:
According to this issue: #125, state locking is no longer required or supported in terragrunt.

If that's the case, it's probably a good idea to pull the references for it out of the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Terragrunt is a thin wrapper for [Terraform](https://www.terraform.io/) that pro
 
    Terragrunt forwards almost all commands, arguments, and options directly to Terraform, using whatever version of
    Terraform you already have installed. However, based on the settings in your `terraform.tfvars` file, Terragrunt can
-   configure remote state, locking, extra arguments, and lots more.
+   configure remote state, extra arguments, and lots more.
 
 1. Terragrunt is a direct implementation of the ideas expressed in 
    [Terraform: Up & Running](http://www.terraformupandrunning.com). Additional background reading that will help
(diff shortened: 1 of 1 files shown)


brikis98:
Thanks for the PR, but I think there is a bit of confusion here. Let me try to clear it up.

Terraform did not used to do any sort of locking, so we did added support for locking in Terragrunt using DynamoDB. The latest versions of Terraform *do* support locking: e.g., the [S3 backend](https://www.terraform.io/docs/backends/types/s3.html) uses DynamoDB for locking, just like Terragrunt used to. So we removed the locking from Terragrunt, but Terragrunt still helps you [manage your backend (AKA remote state) configuration](https://github.com/gruntwork-io/terragrunt#keep-your-remote-state-configuration-dry)! That configuration includes the [dynamodb_table configuration](https://www.terraform.io/docs/backends/types/s3.html#dynamodb_table)  natively supported by Terraform's locking. That also includes the ability to [create the S3 bucket and DynamoDB table automatically](https://github.com/gruntwork-io/terragrunt#create-remote-state-and-locking-resources-automatically) if they don't exist already.

Therefore, I believe the docs are actually up to date.

mcameron:
Ok, I appreciate the feedback - terragrunt simply passes through that config for terraform to handle now rather than it being duplicated functionality.

mcameron:
This is why I was confused: https://github.com/gruntwork-io/terragrunt/issues/269.

So terragrunt is responsible for _creating_ the dynamodb table which terraform then _uses_ to manage the locks.

brikis98:
Yes, exactly!

PR #366, opened 2017-11-11
optionally check env var TF_INPUT to set interactive mode

### LINKED ISSUE #363: $TF_INPUT for automation
Terraform looks to the env var TF_INPUT  as an automation feature.  Terragrunt has a similar feature but with a CLI flag, --terragrunt-non-interactive (Terrafrom also as -input).

Can Terragrunt also look to TF_INPUT? The advantage of using the environment rather then the config file for this is that with the same files local, I can be interactive but in my CI jobs I can be non-interactive.

tomdavidson:
Simple and small change that closes #363 and follows the same pattern used elsewhere to set opts vis env var. The one changed line is self explanatory.

### DIFF
--- a/cli/args.go
+++ b/cli/args.go
@@ -82,7 +82,7 @@ func parseTerragruntOptionsFromArgs(args []string, writer, errWriter io.Writer) 
 
 	opts.TerraformPath = filepath.ToSlash(terraformPath)
 	opts.AutoInit = !parseBooleanArg(args, OPT_TERRAGRUNT_NO_AUTO_INIT, os.Getenv("TERRAGRUNT_AUTO_INIT") == "false")
-	opts.NonInteractive = parseBooleanArg(args, OPT_NON_INTERACTIVE, false)
+	opts.NonInteractive = parseBooleanArg(args, OPT_NON_INTERACTIVE, os.Getenv("TF_INPUT") == "true")
 	opts.TerraformCliArgs = filterTerragruntArgs(args)
 	opts.WorkingDir = filepath.ToSlash(workingDir)
 	opts.Logger = util.CreateLoggerWithWriter(errWriter, "")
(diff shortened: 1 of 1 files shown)


cli/args.go
```diff
@@ -82,7 +82,7 @@ func parseTerragruntOptionsFromArgs(args []string, writer, errWriter io.Writer)
 
 	opts.TerraformPath = filepath.ToSlash(terraformPath)
 	opts.AutoInit = !parseBooleanArg(args, OPT_TERRAGRUNT_NO_AUTO_INIT, os.Getenv("TERRAGRUNT_AUTO_INIT") == "false")
-	opts.NonInteractive = parseBooleanArg(args, OPT_NON_INTERACTIVE, false)
+	opts.NonInteractive = parseBooleanArg(args, OPT_NON_INTERACTIVE, os.Getenv("TF_INPUT") == "true")
```
brikis98: Shouldn't this be `os.Getenv("TF_INPUT") == "false"`? Also, [according to the docs](https://www.terraform.io/docs/configuration/environment-variables.html#tf_input), `0` is also valid.

tomdavidson:
I got tripped up, but corrected now. Thank you for the quick review. Do you squash the PRs from forks?

cli/args.go
```diff
@@ -82,7 +82,7 @@ func parseTerragruntOptionsFromArgs(args []string, writer, errWriter io.Writer)
 
 	opts.TerraformPath = filepath.ToSlash(terraformPath)
 	opts.AutoInit = !parseBooleanArg(args, OPT_TERRAGRUNT_NO_AUTO_INIT, os.Getenv("TERRAGRUNT_AUTO_INIT") == "false")
-	opts.NonInteractive = parseBooleanArg(args, OPT_NON_INTERACTIVE, false)
+	opts.NonInteractive = parseBooleanArg(args, OPT_NON_INTERACTIVE, os.Getenv("TF_INPUT") == "false" || os.Getenv("TF_INPUT") == 0)
```
brikis98: One more thing: `Getenv` returns a `string`, so it would have to be `== "0"`.

tomdavidson: @brikis98  Thanks, this is literally the first time I have ever touched golang. I appreciate all the help I have received in the issues and want to be net positive - hope #370 helps too.

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.13.20

PR #375, opened 2017-11-16
doc TF_INPUT feature

tomdavidson:
PR #366 adding the TF_INPUT feature did not include documentation.


brikis98:
Ah, good point. Thanks!

PR #387, opened 2017-12-01
Add TERRAGRUNT_SOURCE_UPDATE env var

brikis98:
You can now use the environment variable `TERRAGRUNT_SOURCE_UPDATE` instead of the flag `--terragrunt-source-update`. This is useful when some other script is calling Terragrunt on your behalf and you can’t override the flags it uses.

### DIFF
--- a/cli/args.go
+++ b/cli/args.go
@@ -66,7 +66,7 @@ func parseTerragruntOptionsFromArgs(args []string, writer, errWriter io.Writer) 
 		return nil, err
 	}
 
-	sourceUpdate := parseBooleanArg(args, OPT_TERRAGRUNT_SOURCE_UPDATE, false)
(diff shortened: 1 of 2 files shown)


cli/args.go
```diff
@@ -66,7 +66,7 @@ func parseTerragruntOptionsFromArgs(args []string, writer, errWriter io.Writer)
 		return nil, err
 	}
 
-	sourceUpdate := parseBooleanArg(args, OPT_TERRAGRUNT_SOURCE_UPDATE, false)
+	sourceUpdate := parseBooleanArg(args, OPT_TERRAGRUNT_SOURCE_UPDATE, os.Getenv("TERRAGRUNT_SOURCE_UPDATE") == "true" || os.Getenv("TERRAGRUNT_SOURCE_UPDATE") == "1")
```
robinbowes: Just a general comment - I prefer to be lenient about what is accepted when testing Env Vars, eg. in this case, I'd convert the env var value to lower case before comparing so the user can specify "TRUE" or "True" as well as "true".

brikis98: Seems like a good idea. PR to create a helper function for that and apply it everywhere would be very welcome!

Issue #391, opened 2017-12-07
allow get_env() to work without a "default" argument


ebarault:
Hi,

I'm using [get_env()](https://github.com/gruntwork-io/terragrunt#get_env) to wrap some environment variables expected by some modules with a different name, as follows :

```
    extra_arguments "env_var_mapping" {
      commands = ["${get_terraform_commands_that_need_vars()}"]
        arguments = ["-var", "my_var=${get_env("VAR_NAME")}"]
    }
```

I'm willingly using `get_env()` without the `DEFAULT` argument as I wan't terraform to warn/abort if the associated variable is not provided.

The thing is that Terragrunt does not currently support thus and complains:
`Invalid parameters. Expected syntax of the form '${get_env("VAR_NAME", "default")}', but got '"VAR_NAME"'`

Of course I could pass variables prefixed by `TF_VAR_` but i'd like to have the freedom not to name env variables following the name of variables expected by terraform modules, hence the use of terragrunt here.
Rationales for env var remapping: 
- the naming of a terraform variable can be relevant in the context of a module, but could lack sense in the context of a whole app
- the same variable name could be used by several modules deployed at the same time by the same cicd job

What do you think of allowing the use of `get_env()` helper without the `default` argument ?
(or do you have an alternative in mind ?)

brikis98:
A big +1 on this. I've hit the exact same issue with variables such as `password` that need to be set across different environments. A PR for this would be very welcome.

ebarault:
hey @brikis98 !
Provide me with a starting point and I could handle this i guess, it should not be too difficult.

brikis98:
Here you go: https://github.com/gruntwork-io/terragrunt/blob/master/config/config_helpers.go#L83-L84.

Thanks!

ozbillwang:
do we have any updates for this?

ebarault:
@ozbillwang we had discussions with @brikis98 on the proper way to do it (see:  https://github.com/gruntwork-io/terragrunt/pull/393) but then i got caught by somethingelse, and do not use the get_env() helper in the end.
Feel free to resume the PR

PR #393, opened 2017-12-10
allowing get_env() to run without the `DEFAULT` argument

ebarault:
Related to https://github.com/gruntwork-io/terragrunt/issues/391

This PR allows the `get_env()` helper to work with a single argument such as in `get_env("ENV_VAR_NAME")`

- existing test cases pass
- 2 new test cases are provided to test the new expectations
- updated README.md

...
I tested the new implementation in the following UC: 

```
    extra_arguments "password" {
      arguments = ["-var", "'password=${get_env("DB_PWD")}'"]
    }
```
And it works like a charm
**But** beware not to forget the single quotes surrounding the var definition as in `-var 'foo=bar'` or in `"'password=${get_env("DB_PWD")}'"`.

for some reason `terraform plan [..] -var some_var=` will assign an empty string to `some_var` instead of not setting `some_var`, when `terraform plan [..] -var 'some_var='` works OK


ebarault:
and here are test results
[test_results.txt](https://github.com/gruntwork-io/terragrunt/files/1546079/test_results.txt)

config/config_helpers.go
```diff
@@ -192,7 +192,7 @@ func getParentTfVarsDir(include *IncludeConfig, terragruntOptions *options.Terra
 func parseGetEnvParameters(parameters string) (EnvVar, error) {
 	envVariable := EnvVar{}
 	matches := HELPER_FUNCTION_GET_ENV_PARAMETERS_SYNTAX_REGEX.FindStringSubmatch(parameters)
-	if len(matches) < 2 {
+	if len(matches) < 1 {
```
brikis98: This updates the code so it's OK if a default value isn't specified, but that's not enough. You'll need to update the `DefaultValue` param to be a `*string`, so it can be set to `nil` if no value is present. Then, in `getEnvironmentVariable`, you can return an error if the env var is not defined and `DefaultValue` is `nil`.

ebarault: i get your point, but the thing is : regex's named group `default` is never nil. I spent quite a moment testing different regex formulas, and from the moment `env` group is matched, then `default` group will be assigned empty string `""` even if no default value is provided.

I had come first with a more robust implem with `*string` allowing returning `nil` but I came to the conclusion that it was not useful.

Unless i messed up somewhere with the regex it's quite hard to distinguish non provided default and empty string `""`, and at the end env var set as `my_env_var=` resolves as non set var to terraform.

Q. Do we want to deal with passing `""`(different from non set) to terraform, and if yes how ?

brikis98: Ah, I see your question.

Probably the easiest solution is to use two regex's, one with two params, one with one. That way, they will either match exactly or not at all. It'll make it a bit easier to read the regex's too.

ebarault: ok, why not.
but apart from code clarity pov, what added value would you like to see from the error handling pov?

is it to have a specific error message when no valid env var name, nor default value is provided? I find the existing one pretty much ok with this case too

brikis98: If no default is provided, and the two-param version with an empty string as the second parameter isn't used, we should return an error.

ebarault: sorry, not sure to see when it happens, please give me the example you have in mind (as in
`get_env("SOME_ENV_VAR", "")` )

brikis98: Let's say `FOO` is NOT defined.

```
// No error
get_env("FOO", "")

// Error
get_env("FOO")
```

ebarault: why do you want to raise an error when get_env("FOO") ? it's the purpose of this PR not to raise an error in such a case.
The result of `get_env("FOO")` would then be `""`. Consider the UC I was referring to:

```
    extra_arguments "password" {
      arguments = ["-var", "'password=${get_env("DB_PWD")}'"]
    }
```

It would pass `-var 'password='` to terraform who will complain about not having the `password` var defined, which is just fine. In automation, it would make the CD job fail which is safe.

brikis98: Not sure I follow. If you do `-var password=""`, Terraform will not give you an error, as it sees that as setting the variable to an empty string. You might get an error from the provider itself (e.g., AWS), but that will come quite a bit later. So I thought the point was to return an error earlier? 

If you just want the empty string, just do `get_env_("FOO", "")`.

ebarault: I specifically don't want to `-var password=""` which moreover i believe will be interpreted by terraform as `-var password='""'` 
Also note that in current implementation `get_env_("FOO", "")` is equivalent to `-var password=`, not to `-var password=""`, this is confusing. At least this PR has highlighted this point.

Having said that i do not have a way to assign a terraform var with `""` from an env var (`'""'` != `""`). But that's probably more a terraform issue then.

So in in the end `get_env_("FOO", "")` and `get_env_("FOO")` are equivalent, but I find the later more explicit on what it does. 

What about returning an error earlier then? ah.. yes, why not, if  you believe there's enough value in it and it justifies your initial motivation for this PR.

thoughts?

brikis98: It's largely a limitation of Terraform, which does not distinguish between "empty" and "omitted" params (see https://github.com/hashicorp/terraform/issues/14037). So to do what you'd want, you really should not set the param at all, rather than set it to blank or empty string or anything else.

For that, you might need some other helper method that automatically drops empty values from `args`.

ebarault: agreed, I got trapped into this and realized that (confusion "empty" and "omitted") on the way building this PR.

so not that vital : but I still believe we can improve the usability of this method by allowing to avoid the second argument instead of providing an empty string as default, even though the two would be allowed and equivalent (so far, but things might change with future terraform version).

But if you believe there's more value to keep get_env("FOO", "") for what it is now, and use get_env("FOO") to raise an error if FOO is not set, then let's go for it.

Your call, i'm good with both after all.

ebarault: on second thought, probably we should use the second function signature `get_env("SOME_VAR")` to raise an error soon, as you suggested. It's kinda orthogonal to the confusing behavior of terraform with empty strings vars.

brikis98: Here's an idea that might actually do what you want: add support for a `variable_arguments` param to `extra_arguments`:

```hcl
extra_arguments "password" {
  variable_arguments = {
    foo = "bar"
    password = "${get_env("PASSWORD")}"
  }
}
```

The idea is that `variable_arguments` is a map of (name, value) pairs that will be set as `-var name=value` arguments when calling Terraform. This is not only a more convenient syntax for setting variables, but you could also add the rule that if the value evaluates to `nil`, it will be omitted. E.g., if `get_env("PASSWORD")` returns `nil`, then `-var password=...` will not be set at all.

Thoughts?

ebarault: yes, it might be nice to have!
i'm busy right now, but I keep this on the top of my head.

ebarault: that being said, this PR should then be modified to return `nil` when using get_env("PASSWORD") with not `default` provided. Currently i guess it returns `""`

lorengordon:
Is this still in progress, or stuck? Would love to see this feature land.

ebarault:
@lorengordon : i'll try summing up the above discussion and resume the work on it soon.

What behavior would you expect from an end-user perspective?

lorengordon:
I'd say something like this:

```
export FOO=foo

get_env("FOO", "bar")
# returns "foo"

get_env("FOO")
# returns "foo"

unset FOO

get_env("FOO", "bar")
# returns "bar"

get_env("FOO")
# raises exception and exits non-zero
```

brikis98:
PR not updated for over a year, closing

PR #401, opened 2018-01-10
Extend remote_state backend check to include *.tf.json files

tamsky:
Original test in #302 only examined files matching `**/*.tf`.

This PR adds support for `*.tf.json` files.

Neither test parses the HCL or json, and instead looks for a hardcoded regex.
Given that the test is a best-effort at helping folks, I don't think we should start parsing anything.

https://github.com/gruntwork-io/terragrunt/pull/302#issuecomment-356720889

`make fmt` clean.

Test output clean:
[go-test-output.txt](https://github.com/gruntwork-io/terragrunt/files/1620412/go-test-output.txt)

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -307,12 +307,24 @@ func checkTerraformCodeDefinesBackend(terragruntOptions *options.TerragruntOptio
 	if err != nil {
 		return err
 	}
-
-	if !definesBackend {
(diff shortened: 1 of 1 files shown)


brikis98:
Binaries should be here soon: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.13.24

tamsky:
Teammates and I thank you for another very speedy release!
Best.

Issue #402, opened 2018-01-11
Getting json.SyntaxError unexpected end of JSON input


ColOfAbRiX:
**Setup**:
```
$ cat /etc/redhat-release
CentOS Linux release 7.3.1611 (Core)

$ terraform --version
Terraform v0.11.2

$ terragrunt --version
terragrunt version v0.13.23
```

My relevant file structure is the following:

```
.
├── components                    <-- For me "components" is less confusing than "modules"
│   └── network
│       ├── main.tf
│       ├── terraform.tfvars      <-- Used for the "*-all" commands
│       └── vars.tf
└─ live
    ├── test
    │   ├── network
    │   │   ├── terraform.tfstate <-- State file
    │   │   └── terraform.tfvars  <-- Network variables + terragrunt config
    │   ├── secrets.tfvars        <-- Environment secrets
    │   └── environment.tfvars    <-- Environment variables
    └── global.tfvars             <-- Global variables + terragrunt config
```

_live/test/network/terraform.tfvars_
```
terragrunt {
  terraform {
    source = "../../../components//network"
  }
  include {
    path   = "${find_in_parent_folders("global.tfvars")}"
  }
}
```

_live/global.tfvars_
```
terragrunt {
  terraform {
    extra_arguments "var_files" {
      commands = ["${get_terraform_commands_that_need_vars()}"]
      required_var_files = [
        "${get_tfvars_dir()}/../secrets.tfvars",
        "${get_tfvars_dir()}/../environment.tfvars"
      ]
    }
  }
  remote_state {
    backend = "local"
    config {
      path = "${get_tfvars_dir()}/terraform.tfstate"
    }
  }
}
```

_components/network/main.tf_
```
terraform { backend "local" {} }
provider "aws" {
  ...
}
module "networks" {
  source ="git::ssh://..."
  ...
}
module "dns" {
  source ="git::ssh://..."
  ...
}
```

**What happens**

I start cleaning `rm -rf ~/.terragrunt`.
I also have an existing and correct `live/test/network/terraform.tfstate` file:
```
$ python -mjson.tool < terraform.tfstate
{
    "lineage": "d130af51-dd8d-4e0b-b7e5-a4a8f540a517",
    "modules": [
        {
            "depends_on": [],
            "outputs": {},
...
```
Then I run `terragrunt plan` and everything works fine, I see `terraform init` runs, and also `terraform plan`, terraform refreshes the state and tells me there's nothing to do (because my infrastructure already exists):
[...]

PR #403, opened 2018-01-17
Fix of issue #402: load given state file when specified.

ColOfAbRiX:
This is a bugfix for "Getting json.SyntaxError unexpected end of JSON input" Issue #402

When using a local backend and a path is given, terragrunt should load that file instead of looking in the default working directory.

As a Go beginner, feel free to suggest any correction or semantic mistake.

### DIFF
--- a/remote/terraform_state_file.go
+++ b/remote/terraform_state_file.go
@@ -44,16 +44,15 @@ func (state *TerraformState) IsRemote() bool {
 	return state.Backend != nil && state.Backend.Type != "local"
 }
 
-// Parse the Terraform .tfstate file from the location specified by workingDir. If no location is specified,
-// search the current directory. If the file doesn't exist at any of the default locations, return nil.
+// Parses the Terraform .tfstate file. If a local backend is used then search the given path, or
+// return nil if the file is missing. If the backend is not local then parse the Terraform .tfstate
+// file from the location specified by workingDir. If no location is specified, search the current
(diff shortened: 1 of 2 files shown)


remote/terraform_state_file.go
```diff
@@ -46,7 +46,13 @@ func (state *TerraformState) IsRemote() bool {
 
 // Parse the Terraform .tfstate file from the location specified by workingDir. If no location is specified,
 // search the current directory. If the file doesn't exist at any of the default locations, return nil.
```
brikis98: Make sure to update this comment

remote/terraform_state_file.go
```diff
@@ -46,7 +46,13 @@ func (state *TerraformState) IsRemote() bool {
...
-func ParseTerraformStateFileFromLocation(workingDir string) (*TerraformState, error) {
+func ParseTerraformStateFileFromLocation(backend string, config map[string]interface{}, workingDir string) (*TerraformState, error) {
+	stateFile, ok := config["path"].(string)
+	if backend == "local" && ok {
+		if util.FileExists(stateFile) {
```
brikis98: Hm, not sure what the right behavior is here if the file doesn't exist. Should we just try to read it elsewhere? Return an empty `TerraformState`? Return an error?

brikis98:
If the build passes, the new binaries should be here in a few minutes: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.13.25

PR #416, opened 2018-02-15
WIP: fix slice out of bounds on tf cli args

### LINKED ISSUE #411: Bug in CLI arg parsing
If you run `terragrunt force-unlock` (with no ID), you get an exception:

```
[terragrunt] 2018/02/01 16:17:46 runtime.errorString runtime error: slice bounds out of range
/private/tmp/terragrunt-20180130-90428-105chku/terragrunt-0.14.0/src/github.com/gruntwork-io/terragrunt/errors/errors.go:72 (0x10ed5b3)
/usr/local/Cellar/go/1.9.3/libexec/src/runtime/asm_amd64.s:509 (0x105724b)
        call32: CALLFN(·call32, 32)
/usr/local/Cellar/go/1.9.3/libexec/src/runtime/panic.go:491 (0x102b6d3)
        gopanic: reflectcall(nil, unsafe.Pointer(d.fn), deferArgs(d), uint32(d.siz), uint32(d.siz))
/usr/local/Cellar/go/1.9.3/libexec/src/runtime/panic.go:35 (0x102a41e)
        panicslice: panic(sliceError)
/private/tmp/terragrunt-20180130-90428-105chku/terragrunt-0.14.0/src/github.com/gruntwork-io/terragrunt/options/options.go:187 (0x112c9bc)
/private/tmp/terragrunt-20180130-90428-105chku/terragrunt-0.14.0/src/github.com/gruntwork-io/terragrunt/cli/cli_app.go:252 (0x14f7bee)
/private/tmp/terragrunt-20180130-90428-105chku/terragrunt-0.14.0/src/github.com/gruntwork-io/terragrunt/cli/cli_app.go:223 (0x14f761d)
/private/tmp/terragrunt-20180130-90428-105chku/terragrunt-0.14.0/src/github.com/gruntwork-io/terragrunt/cli/cli_app.go:196 (0x14f7540)
/private/tmp/terragrunt-20180130-90428-105chku/terragrunt-0.14.0/src/github.com/gruntwork-io/terragrunt/cli/cli_app.go:177 (0x14f720e)
/private/tmp/terragrunt-20180130-90428-105chku/terragrunt-0.14.0/src/github.com/gruntwork-io/terragrunt/vendor/github.com/urfave/cli/app.go:502 (0x1498542)
/private/tmp/terragrunt-20180130-90428-105chku/terragrunt-0.14.0/src/github.com/gruntwork-io/terragrunt/vendor/github.com/urfave/cli/app.go:268 (0x1496413)
/private/tmp/terragrunt-20180130-90428-105chku/terragrunt-0.14.0/main.go:20 (0x14fcf74)
/usr/local/Cellar/go/1.9.3/libexec/src/runtime/proc.go:195 (0x102d4f6)
        main: fn()
/usr/local/Cellar/go/1.9.3/libexec/src/runtime/asm_amd64.s:2337 (0x1059a91)
        goexit: BYTE    $0x90   // NOP
```

Looks like we're missing a [bounds check here](https://github.com/gruntwork-io/terragrunt/blob/master/options/options.go#L187).

sarkis:
@brikis98 got started on this.. still a WIP: 

Fixes #411

* I decided to not use math.Min as the `math` package deals in `float64` so settled on a Min function in the util package. I'll add a test for this.

* This will still fail when you run `terragrunt force-unlock`, however, it won't be a go panic but instead terraform complaining that no ID was passed. This was the best behavior I can think of being that terragrunt is a wrapper - lmk if you think it should behave differently.

I'll need to add tests and check out different edge cases I haven't thought of yet - but hoping to get a CR early to see if I am at least on the right track with this.

### DIFF
--- a/options/options.go
+++ b/options/options.go
@@ -178,7 +178,9 @@ func (terragruntOptions *TerragruntOptions) InsertTerraformCliArgs(argsToInsert 
 
 	commandLength := 1
 	if util.ListContainsElement(TERRAFORM_COMMANDS_WITH_SUBCOMMAND, terragruntOptions.TerraformCliArgs[0]) {
-		commandLength = util.Min(2, cap(terragruntOptions.TerraformCliArgs))
+		// Since these terraform commands require subcommands which may not always be supplied by the user,
+		// using util.Min to return the minimum to avoid potential out of bounds slice errors.
+		commandLength = util.Min(2, len(terragruntOptions.TerraformCliArgs))
 	}
 
 	// Options must be inserted after command but before the other args
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -133,6 +133,74 @@ func TestTerragruntWorksWithIncludesParentUpdatedAndOldConfig(t *testing.T) {
 	runTerragrunt(t, fmt.Sprintf("terragrunt apply --terragrunt-non-interactive --terragrunt-config %s --terragrunt-working-dir %s", tmpTerragruntConfigPath, childPath))
 }
 
+func TestTerragruntReportsTerraformErrorWithSubcommandMissing(t *testing.T) {
+	t.Parallel()
+
+	cmd := fmt.Sprintf("terragrunt debug --terragrunt-non-interactive --terragrunt-working-dir %s", TEST_FIXTURE_EXTRA_ARGS_PATH)
+	var (
+		stdout bytes.Buffer
+		stderr bytes.Buffer
+	)
+	// Call runTerragruntCommand directly because this command contains failures (which causes runTerragruntRedirectOutput to abort) but we don't care.
+	if err := runTerragruntCommand(t, cmd, &stdout, &stderr); err == nil {
+		t.Fatalf("Failed to properly fail command: %v.", cmd)
+	}
+	output := stdout.String()
+	errOutput := stderr.String()
+	assert.True(t, strings.Contains(errOutput, "terraform debug") || strings.Contains(output, "terraform debug"))
+}
+
+func TestTerragruntReportsTerraformErrorWithSubcommandMissingArg(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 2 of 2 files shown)


options/options.go
```diff
@@ -178,7 +178,7 @@ func (terragruntOptions *TerragruntOptions) InsertTerraformCliArgs(argsToInsert
 
 	commandLength := 1
 	if util.ListContainsElement(TERRAFORM_COMMANDS_WITH_SUBCOMMAND, terragruntOptions.TerraformCliArgs[0]) {
-		commandLength = 2
+		commandLength = util.Min(2, cap(terragruntOptions.TerraformCliArgs))
```
brikis98: Use `len` (number of items in array), not `cap` (capacity of array).

brikis98: Also, I had to scratch my head at this logic for a min to understand it, so perhaps add a comment explaining why we're using `util.Min` here: e.g. "these Terraform commands require extra arguments, but it's possible the user didn't actually pass them, so use math.Min to make sure we don't try to read args that aren't there..."

sarkis: In this case cap and len will be the same number - I realized this after I sent the PR too. Only reason it's a cap because at first I had put it in the slices :D. I'll change to len as I think it makes more sense to use that now.

Will be adding docs and tests, shortly.

sarkis:
@brikis98 Q on tests - I looked over the `arg_test.go` - looks like those tests are just to ensure the args get parsed correctly right? 

Didn't want to re-do work, but couldn't find any tests that actually run and check terragrunt all the way through, which I don't think is right.. but wanted to confirm

brikis98:
https://github.com/gruntwork-io/terragrunt/blob/master/test/integration_test.go

sarkis:
I'll re-open this issue - looks like something weird happening with my fork on the PR

sarkis:
Can delete this for cleanup ... it was reopened and merged here: https://github.com/gruntwork-io/terragrunt/pull/420 - something wrong with this PR (can see above says it's coming from an unknown repository)

Issue #112, opened 2017-01-27
Support for non-standard regions and endpoints


Cinderhaze:
Since terragrunt intracts with s3/dynamoDB, does it support the capability for overriding regions/endpoints, similar to the following?

https://github.com/hashicorp/terraform/pull/11358
https://www.terraform.io/docs/providers/aws/index.html#s3
https://www.terraform.io/docs/providers/aws/index.html#dynamodb_endpoint

I am wondering if it will work in an AWS environment similar to [aws GovCloud Region](http://docs.aws.amazon.com/govcloud-us/latest/UserGuide/using-govcloud-endpoints.html), which has different regions, partition values, and endpoints.

brikis98:
We have not built anything special to override regions or endpoints. PR is welcome!

brikis98:
Fixed by #422 and #426. Available in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.14.2.

PR #422, opened 2018-02-26
Added support for endpoint and credentials variables in S3 remote state 

MRinalducci:
I added support for custom endpoint and credentials variable in the S3 remote state backend for Terragrunt which is used to create bucket and enable versioning if bucket does not exist.

This feature was also asked in issue [#112](https://github.com/gruntwork-io/terragrunt/issues/112).

It uses the following variable as specified in the [Terraform S3 backend documentation](https://www.terraform.io/docs/backends/types/s3.html):
* endpoint (only for S3)
* access_key
* secret_key

This update was successfully tested with integration tests.

### DIFF
--- a/aws_helper/config.go
+++ b/aws_helper/config.go
@@ -21,7 +21,7 @@ func CreateAwsSession(awsRegion, awsEndpoint string, awsAccessKey string, awsSec
 		if service == "s3" {
 			return endpoints.ResolvedEndpoint{
 				URL:           awsEndpoint,
-				SigningRegion: "custom-signing-region",
+				SigningRegion: awsRegion,
 			}, nil
 		}
 
--- a/aws_helper/config.go
+++ b/aws_helper/config.go
@@ -5,6 +5,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/gruntwork-io/terragrunt/errors"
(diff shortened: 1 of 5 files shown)


MRinalducci:
I also added a custom endpoint resolver which resolve only S3 endpoints and don't impact DynamoDB endpoints.

MRinalducci:
Thank you for reviewing the PR!
I know this is not the best practice to put credentials in the code. But I'm using it sometimes for testing.
As Terraform supports it, I think it can be confusing for a user if it is not supported by Terragrunt.
In my point of view, the user is responsible how he implements the code.

Moreover, what is the best practice to store credentials of additionnal providers other than AWS?
What do you think of encrypting a .tfvars file containing credentials before commiting it to a private GIT repo? This method could also be used with a file containing the S3 backend with credentials.

However, my principle concern was the ability to specify a custom endpoint in Terragrunt, so if you prefer a can revert the credentials changes.

Finally, I take this opportunity to say that Terragrunt is a really great tool!

MRinalducci:
> Yea, it's a tradeoff, but I've found that making it easy to do the insecure thing is almost always a bad idea. And the unfortunate reality is that putting the keys in the code really is the easiest thing to do, so people use it way more than they should, which is awful from a security/reusability/etc perspective.

I understand your point of view which I share too.

> Unless there's a really compelling use case that can only be done by putting the credentials directly in the code, my vote would be to not support that in Terragrunt for now, and limit this PR to just the custom endpoint.

No problem, I reverted the credentials change as suggested and tested it successfully.

brikis98:
A bunch of tests failed, all with essentially the same error:

```
--- FAIL: TestRemoteWithBackend (1.07s)
	integration_test.go:602: Copying fixture-download/remote-with-backend to /tmp/terragrunt-test495155519
	integration_test.go:592: Failed to run Terragrunt command 'terragrunt apply --terragrunt-non-interactive --terragrunt-working-dir /tmp/terragrunt-test495155519/fixture-download/remote-with-backend' due to error: MissingEndpoint: 'Endpoint' configuration is required for this service
	Error Trace:	integration_test.go:788
			asm_amd64.s:480
			panic.go:340
			testing.go:453
			testing.go:496
			integration_test.go:592
			integration_test.go:587
			integration_test.go:427
	Error:      	Expected nil, but got: &awserr.requestError{awsError:(*awserr.baseError)(0xc420388140), statusCode:400, requestID:"5HMQFNJ31KBRHL1IB27V92B8QVVV4KQNSO5AEMVJF66Q9ASUAAJG"}
	Messages:   	Unexpected error: ResourceNotFoundException: Requested resource not found: Table: terragrunt-lock-table-yy2q5g not found
	            		status code: 400, request id: 5HMQFNJ31KBRHL1IB27V92B8QVVV4KQNSO5AEMVJF66Q9ASUAAJG
	integration_test.go:732: Deleting test s3 bucket terragrunt-test-bucket-mkrp5i
	integration_test.go:736: Failed to list object versions in s3 bucket terragrunt-test-bucket-mkrp5i: MissingEndpoint: 'Endpoint' configuration is required for this service
```

Any idea what is the cause? Is the default endpoint resolver not working for some reason?

MRinalducci:
I didn't saw those errors before, sorry for that.
Let me have a look!

MRinalducci:
Ok I found the problem, I ommited a test if the endpoint was not specified.
I commit the changes to the same branch.

brikis98:
Thx! Since the branch is already merged, could you put the changes in a new branch & new PR?

MRinalducci:
Commit done, should I make a new PR for that ?

PR #425, opened 2018-02-27
Bug fix: If S3 endpoint is not specified use default resolver

MRinalducci:
Bug fix: If S3 endpoint is not specified use default resolver.

It is related to PR [#422](https://github.com/gruntwork-io/terragrunt/pull/422).


MRinalducci:
Leave a comment

PR #426, opened 2018-02-27
Bug fix: If S3 endpoint is not specified use default resolver 

MRinalducci:
This is a bug fix: If S3 endpoint is not specified use default resolver.

It is related to PR #422.


aws_helper/config.go
```diff
@@ -17,7 +17,7 @@ import (
 func CreateAwsSession(awsRegion, awsEndpoint string, awsProfile string, iamRoleArn string, terragruntOptions *options.TerragruntOptions) (*session.Session, error) {
 	defaultResolver := endpoints.DefaultResolver()
 	s3CustResolverFn := func(service, region string, optFns ...func(*endpoints.Options)) (endpoints.ResolvedEndpoint, error) {
-		if service == "s3" {
+		if service == "s3" && awsEndpoint != "" {
```
brikis98: Ah, interesting. Should the param be called `customS3Endpoint` or similar to make it clearer?

MRinalducci: Yes, I think this is a good idea!
But only in `func CreateAwsSession`?

brikis98: Probably in `CreateS3Client` too.

MRinalducci: Ok, why not, so it's same name overall the code!
I'll do the changes right away.

MRinalducci:
I changed awsEndpoint to customS3Endpoint variable to make it clearer as suggested.

MRinalducci:
BTW integration test are running successfully this time.
I didn't saw the regression at first pull request for the S3 backend custom endpoint feature, my bad.

brikis98:
> BTW integration test are running successfully this time.

Great, thanks!

> I didn't saw the regression at first pull request for the S3 backend custom endpoint feature, my bad.

No worries. Go's test output is hard to parse.

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.14.2

Issue #377, opened 2017-11-19
Proposal: Terragrunt hooks


brikis98:
A proposal to discuss with the Terragrunt community: adding "hooks" to Terragrunt that can be used to execute arbitrary shell commands. The idea would be to follow the basic structure of `extra_arguments`, where you can specify shell commands that will be executed either before or after specific Terraform commands. Here's the rough idea:

```hcl
terragrunt = {
  terraform {
    # This hook configures Terragrunt to copy /foo to /bar before executing apply or plan
    before_hook "copy-file" {
      commands = ["apply", "plan"]
      execute = ["cp", "/foo", "/bar"]
    }

    # This hook configures Terragrunt to do a simple echo statement after executing any Terraform command
    after_hook "copy-file" {
      commands = ["${all_terraform_commands()}"]
      execute = ["echo", "Hello World"]
    }
  }
}
```

This idea has come up before, but I pushed back against it, recommending the use of Terraform's [local-exec provisioner](https://www.terraform.io/docs/provisioners/local-exec.html) instead. However, I realize now that if you need to execute some code *before* or *after* terraform runs, `local-exec` doesn't help you. 

A pattern that seems particularly interesting is to copy files into the working directory before running Terraform. This is especially useful for cases where Terraform doesn't support interpolation or effective code reuse. Examples:

1. Copy a `version.tf` file into your working directory. This file could define the [required_version](https://www.terraform.io/docs/configuration/terraform.html#specifying-a-required-terraform-version) setting for your whole company so you don't have to copy/paste that setting into every single Terraform module. 

1. Copy a `common-vars.tf` file into your working directory. This file could define common input variables that every single module should define: e.g., `aws_region`, `aws_account_id`, etc.

1. Copy a `remote-state.tf` file into your working directory and dynamically fill in the proper `key` and `bucket` values. Instead of relying on Terragrunt for keeping your remote state configuration DRY, you could use this approach to execute arbitrary code, and have a more flexible/customizable system.

Thoughts? Feedback? Concerns?

antonbabenko:
That is exactly what I am implementing using *another shell wrapper script* and I really don't like my approach. My use case is described in example 1 and 2.

My feedback is that hooks in child `terraform.tfvars` and in a parent's one should be possible to combine, because the one in a child sometimes has some specific steps, while the one in parent should be applied for all cases.

john-mcgowan-wowza:
[...]

PR #439, opened 2018-03-19
Terragrunt hooks

eak12913:
This is my stab at implementing [#377](https://github.com/gruntwork-io/terragrunt/issues/377)

Notes:
- `run_on_error`: allows users to control whether this hook should run if a previous error occurred (from a previous hook or from terraform execution)
- `run_on_error`: if you don't specify it, `false` is the default

Before hook syntax:
```    
   before_hook "copy-file-before2" {
      commands = ["apply", "plan"]
      execute = ["echo", "Hello World"]
      run_on_error = true
    }
```

After hook syntax:
```
    after_hook "after-copy-file" {
      commands = ["apply", "plan"]
      execute = ["echo", "After Hook Running"]
      run_on_error = true
    }
```


cli/cli_app.go
```diff
@@ -223,6 +224,46 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 	return runTerragruntWithConfig(terragruntOptions, terragruntConfig, false)
 }
 
+func processBeforeAfterHookActions(beforeAfterHoook []config.BeforeAfterHook, terragruntConfig *config.TerragruntConfig, terragruntOptions *options.TerragruntOptions, previousExecError bool) bool {
```
brikis98: `processHooks` is probably a bit simpler of a name :)

config/config.go
```diff
@@ -63,10 +63,24 @@ func (deps *ModuleDependencies) String() string {
 	return fmt.Sprintf("ModuleDependencies{Paths = %v}", deps.Paths)
 }
 
+// BeforeAfterHook specifies terraform commands (apply/plan) and array of os commands to execute
+type BeforeAfterHook struct {
```
brikis98: Could just call it `Hook`.

config/config.go
```diff
@@ -63,10 +63,24 @@ func (deps *ModuleDependencies) String() string {
...
+}
+
 // TerraformConfig specifies where to find the Terraform configuration files
 type TerraformConfig struct {
-	ExtraArgs []TerraformExtraArguments `hcl:"extra_arguments"`
-	Source    string                    `hcl:"source"`
+	ExtraArgs  []TerraformExtraArguments `hcl:"extra_arguments"`
+	Source     string                    `hcl:"source"`
+	BeforeHook []BeforeAfterHook         `hcl:"before_hook"`
+	AfterHook  []BeforeAfterHook         `hcl:"after_hook"`
```
brikis98: Probably `BeforeHooks` and `AfterHooks` would be a bit clearer.

cli/cli_app.go
```diff
@@ -261,7 +302,37 @@ func runTerragruntWithConfig(terragruntOptions *options.TerragruntOptions, terra
 			return err
 		}
 	}
-	return shell.RunTerraformCommand(terragruntOptions, terragruntOptions.TerraformCliArgs...)
+
+	terragruntOptions.Logger.Println("Running terraform with: %s", terragruntOptions)
+
+	errorOccurred := true
```
brikis98: Rename to `beforeHooksErr`.

cli/cli_app.go
```diff
@@ -261,7 +302,37 @@ func runTerragruntWithConfig(terragruntOptions *options.TerragruntOptions, terra
...
+		if errorOccurred || possibleError != nil {
+			//if there was an error during before hook run
+			//or if there was an error during terraform execution
+			//make sure to pass that knowledge on to after hook execution
+			processBeforeAfterHookActions(terragruntConfig.Terraform.AfterHook, terragruntConfig, terragruntOptions, true)
+		} else {
+			processBeforeAfterHookActions(terragruntConfig.Terraform.AfterHook, terragruntConfig, terragruntOptions, false)
+		}
+	}
+	return possibleError
```
brikis98: It took a little while to figure out the intent of the code above. Some ideas on how to make this a bit clearer.

First, as this method can now generate multiple errors, create a type to capture them:

```go
type MultiError struct {
  Errors []error
}

func NewMultiError(errs ... error) error {
  if len(errs) == 0 {
    return nil
  }

  return MultiError{Errors: errs}
} 

func (errs MultiError) Error() string {
  errorMessages := []string{}
  for _, err := range errs {
    errorMessages = append (out, err.Error())
  }

  return fmt.Sprintf("Hit multiple errors:\n%s", strings.Join(errorMessages, "\n"))
}
```

Next, extract most of this functionality into small helper methods, and collect the errors as you go:

```go
beforeHooksErr := processHooks(terragruntConfig.Terraform.BeforeHooks, terragruntOptions)
terraformCmdErr := runTerraformCommandIfNoErrors(beforeHooksErr, terragruntOptions)
afterHooksErr := processHooks(terragruntConfig.Terraform.afterHooks, terragruntOptions, beforeHooksErr, terraformCmdErr)

return NewMultiError(beforeHooksErr, terraformCmdErr, afterHooksErr)
```

The key things to notice are:

1. `processHooks` takes an optional `...` of errors that happened previous as the final params. On the first call, this list is empty; on the second call, it has errors from before hooks and the Terraform command execution. If any of these are non-nil, it internally checks the `RunOnError` setting to determine if the hook should be executed anyway.

1. The `runTerraformCommandIfNoErrors` only executes if the errors passed into it are `nil`.

cli/cli_app.go
```diff
@@ -223,6 +224,46 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
...
+					errorOccurred = true
+				}
+			}
+		}
+	}
+
+	return errorOccurred
+}
+
+func contains(arr []string, str string) bool {
```
brikis98: Stupid Go and its lack of collections. We have this method in `util.ListContainsElement`.

cli/cli_app.go
```diff
@@ -223,6 +224,46 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
...
 }
 
+func processBeforeAfterHookActions(beforeAfterHoook []config.BeforeAfterHook, terragruntConfig *config.TerragruntConfig, terragruntOptions *options.TerragruntOptions, previousExecError bool) bool {
+
+	terraformCommand := terragruntOptions.TerraformCliArgs[0]
+
+	errorOccurred := previousExecError
+
+	for i := 0; i < len(beforeAfterHoook); i++ {
+		applicableCommands := beforeAfterHoook[i].Commands
```
brikis98: `beforeAfterHoook` has too many o's.

cli/cli_app.go
```diff
@@ -223,6 +224,41 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
...
+
+	terraformCommand := terragruntOptions.TerraformCliArgs[0]
+
+	errorsOccurred := []error{}
+
+	if len(hooks) > 0 {
+		terragruntOptions.Logger.Printf("Detected %d Hooks", len(hooks))
+
+		for i := 0; i < len(hooks); i++ {
+			applicableCommands := hooks[i].Commands
```
brikis98: Idiomatic way to do this loop in Go is:

```go
for _, hook := range hooks {
  applicableCommands := hook.Commands
}
```

cli/cli_app.go
```diff
@@ -223,6 +224,41 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
...
+
+			if util.ListContainsElement(applicableCommands, terraformCommand) {
+				//if there's no previous error, execute command
+				//OR if a previos error DID happen AND we want to run anyways
+				//then execute.
+				//Skip execution if there was an error AND we care about errors
+				if len(previousExecError) == 0 || hooks[i].RunOnError {
```
brikis98: To reduce nesting, could combine these if-statements into a method and call this method in the for-loop:

```go
func shouldRunHook(hook config.Hook, terragruntOptions *options.TerragruntOptions,  previousExecErrors ... error) bool {
  return util.ListContainsElement(hook.Commands, terragruntOptions..TerraformCliArgs[0]) && (len(previousExecErrors) == 0) || hook.RunOnError)
}
```

cli/cli_app.go
```diff
@@ -223,6 +224,41 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
...
+					terragruntOptions.Logger.Printf("Executing hook: %s", hooks[i].Name)
+					actionToExecute := hooks[i].Execute[0]
+					actionParams := hooks[i].Execute[1:]
```
brikis98: What does this do if `Execute` only has 1 item? Or 0?

cli/cli_app.go
```diff
@@ -223,6 +224,47 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
...
+	errorsOccurred := []error{}
+
+	terragruntOptions.Logger.Printf("Detected %d Hooks", len(hooks))
+
+	for _, curHook := range hooks {
+		allPreviousErrors := append(previousExecError, errorsOccurred...)
+		if shouldRunHook(curHook, terragruntOptions, allPreviousErrors...) {
+			terragruntOptions.Logger.Printf("Executing hook: %s", curHook.Name)
+			actionToExecute := curHook.Execute[0]
+			if actionToExecute != "" {
```
brikis98: Prob better to validate this stuff during parsing, not part way through execution.

cli/cli_app.go
```diff
@@ -223,6 +224,47 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
...
+				actionParams := curHook.Execute[1:]
```
brikis98: What if it's just a command with no args?

test/fixture-hooks/after-only/terraform.tfvars
```diff
@@ -0,0 +1,12 @@
+terragrunt = {
+  terraform {
+    // source = "test"
+
+    # This hook configures Terragrunt to do a simple echo statement after executing any Terraform command
```
brikis98: Comment isn't accurate

test/fixture-hooks/before-and-after/terraform.tfvars
```diff
@@ -0,0 +1,20 @@
+terragrunt = {
+  terraform {
+    // source = "test"
+
+    # This hook configures Terragrunt to copy /foo to /bar before executing apply or plan
```
brikis98: None of these comments seem to correspond to what the code is doing?

test/fixture-hooks/skip-on-error/terraform.tfvars
```diff
@@ -0,0 +1,37 @@
+terragrunt = {
+  terraform {
+    // source = "test"
+
+    # This hook purposely causes an error
+    before_hook "before_hook_1" {
+      commands = ["apply", "plan"]
+      execute = ["cd","garbageFolder"]
```
brikis98: Could also just `exit 1`

test/integration_test.go
```diff
@@ -60,6 +66,97 @@ func init() {
 	rand.Seed(time.Now().UnixNano())
 }
 
+func TestTerragruntBeforeHook(t *testing.T) {
+	cleanupTerraformFolder(t, TEST_FIXTURE_HOOKS_BEFORE_ONLY_PATH)
```
brikis98: Ideally, we can call `t.Parallel()` at the top of each of these tests so they all run in parallel. Note that this only works if they all run in different folders.

brikis98: One other catch: the `file.out` that may of the hooks create may be left overs from previous test runs. So probably, the tests should copy the folder you're testing (e.g. `TEST_FIXTURE_HOOKS_BEFORE_ONLY_PATH`) to a new temp folder each test run to ensure we're not seeing an older `file.out`. There are several helper methods in in `integration_test.go` to do that copying.

test/integration_test.go
```diff
@@ -60,6 +66,97 @@ func init() {
...
+
+	err := runTerragruntCommand(t, fmt.Sprintf("terragrunt apply --terragrunt-non-interactive --terragrunt-working-dir %s", TEST_FIXTURE_HOOKS_SKIP_ON_ERROR_PATH), &stdout, &stderr)
+
+	output := stderr.String()
+	if err != nil {
+		assert.Contains(t, output, "BEFORE_SHOULD_DISPLAY")
+		assert.NotContains(t, output, "BEFORE_NODISPLAY")
+
+		assert.Contains(t, output, "AFTER_SHOULD_DISPLAY")
+		assert.NotContains(t, output, "AFTER_NODISPLAY")
```
brikis98: Nice tests!

cli/cli_app.go
eak12913: This should work. I added a test to make sure it does: `TestTerragruntBeforeOneArgAction`

test/integration_test.go
eak12913: Good point - will look into it and implement

test/integration_test.go
eak12913: Thanks :-)

config/config.go
```diff
@@ -359,6 +374,15 @@ func convertToTerragruntConfig(terragruntConfigFromFile *terragruntConfigFile, t
 		terragruntConfig.RemoteState = terragruntConfigFromFile.RemoteState
 	}
 
+	//Validate before and after hooks to make sure that they are properly configured
+	allHooks := append(terragruntConfigFromFile.Terraform.BeforeHooks, terragruntConfigFromFile.Terraform.AfterHooks...)
+
+	for _, curHook := range allHooks {
+		if curHook.Execute[0] == "" {
```
brikis98: Could `curHook.Execute` be an empty list? So that this throws an out of bounds error?

errors/invalidargerror.go
```diff
@@ -0,0 +1,13 @@
...
+	s string
+}
+
+func (e InvalidArgError) Error() string {
+	return e.s
+}
+
+func NewInvalidArgError(message string) error {
+	return InvalidArgError{s: message}
+}
```
brikis98: A couple thoughts on errors:

1. Unlike Java, you don't have to define them in their own files. We often put the errors relevant to a given bit of code in the same file (often at the bottom) as that code.
1. If your error is just a wrapper for a single type, you can declare it as an "alias" for that type to simplify usage a bit:

```go
type InvalidArgError string

func (e InvalidArgError) Error() string {
	return string(e)
}

// To create this error elsewhere in the code:
return InvalidArgError("My error message")
```

test/fixture-hooks/bad-arg-action/terraform.tfvars
```diff
@@ -0,0 +1,11 @@
+terragrunt = {
+  terraform {
+
+    # This hook is purposely misconfigured to trigger an error
+    before_hook "before_hook_1" {
+      commands = ["apply", "plan"]
+      execute = [""]
```
brikis98: Test this with `[]` (empty list is different than list with an empty string)

test/fixture-hooks/interpolations/terraform.tfvars
```diff
@@ -0,0 +1,11 @@
+terragrunt = {
+  terraform {
+
+    # This hook echos out user's HOME path or HelloWorld 
+    before_hook "before_hook_1" {
+      commands = ["apply", "plan"]
+      execute = ["echo", "${get_env("HOME", "HelloWorld")}"]
```
brikis98: Nice :)

test/fixture-hooks/skip-on-error/terraform.tfvars
```diff
@@ -0,0 +1,37 @@
+terragrunt = {
+  terraform {
+    // source = "test"
```
brikis98: Remove?

config/config.go
eak12913: Good point - Added a new test for this and fixed the underlying issue

errors/invalidargerror.go
eak12913: Ah ok

eak12913:
Yikes! You’re right! I’ll try to get that done today or tomorrow.

brikis98:
I'll wait for the tests to run, and if they pass, issue a new release.

brikis98:
New release is here!

https://github.com/gruntwork-io/terragrunt/releases/tag/v0.14.4

Thanks @eak12913!

ebarault:
Excellent, a very big thank you for bringing this to terragrunt @eak12913

eak12913:
No problem :-) Happy to help!

ebarault:
@eak12913 : i tried playing a bit with new hooks and i realized hooks are running inside terragrunt's tmp dir, which means after the terragrunt has created it and moved the files there.

in my case, i'd also need hooks that would happen before in the terragrunt source code dir.

what do you think would be the simpler to implement, add an option to the `before_hook` block? or create another type of hook? such as `initial_hook`?

eak12913:
@ebarault: I discussed this with @brikis98 and he proposed an interesting idea:  What if we provided a `working_dir` parameter for the hook? This would default to terragrunt's working dir, but could be overridden to anything you need.  Could this work for you?

ebarault:
By "terragrunt's working dir", do you mean the location where the original `terraform.tfvars` sits or the tmp dir where it is copied to?

See https://github.com/gruntwork-io/terragrunt/issues/453, i have also a temporality issue with when the hooks happen: thanks to terragrunt's interpolation functions i managed running the hooks in the source code dir, but then the problem is that they happen after terragrunt's parsing of `required_var_files` / `optional_var_files`.

So if the file (which i want to generate with hooks) is missing at terragrunt's execution, even though it's generated just after, it will be skipped by terragrunt.
```
[terragrunt] 2018/03/28 19:53:13 Skipping var-file /path/to/component.tfvars as it does not exist
```

I ended up generating the file directly in the tmp dir as suggested by @lorengordon using the `.auto.tfvars` extension so it gets automatically parsed by terraform.

Not an ideal solution if you ask me as it makes the workflow difficult to understand, and we have to care about the tmp folder which should remain out of end user's concern.

A more natural solution would have the before hooks doing any required action actually before any other terragrunt step, then the regular terragrunt workflow takes place.

PR #460, opened 2018-04-05
Issue 459 fix

eak12913:


cli/cli_app.go
```diff
@@ -224,40 +224,47 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 	return runTerragruntWithConfig(terragruntOptions, terragruntConfig, false)
 }
 
-func processHooks(hooks []config.Hook, terragruntOptions *options.TerragruntOptions, previousExecError ...error) error {
+func processHooks(hooks []config.Hook, terragruntOptions *options.TerragruntOptions, previousExecErrors errors.MultiError) error {
+
 	if len(hooks) == 0 {
```
brikis98: Wouldn't it be a lot simpler to fix this by just checking if there are any non-`nil` errors?

errors/multierror.go
```diff
@@ -18,12 +18,24 @@ func NewMultiError(errs ...error) error {
 	}
 
 	if len(nilsRemoved) == 0 {
-		return nil
+		return MultiError{}
```
brikis98: My guess is this change will break things, because now we're always returning errors...

cli/cli_app.go
```diff
@@ -257,7 +257,13 @@ func shouldRunHook(hook config.Hook, terragruntOptions *options.TerragruntOption
 	//OR if a previos error DID happen AND we want to run anyways
 	//then execute.
 	//Skip execution if there was an error AND we care about errors
-	return util.ListContainsElement(hook.Commands, terragruntOptions.TerraformCliArgs[0]) && (len(previousExecErrors) == 0 || hook.RunOnError)
+
+	//resolves: https://github.com/gruntwork-io/terragrunt/issues/459
+	//by helping to filter out nil errors that were acting as false positives
+	//for the len(previousExecErrors) == 0 check that used to be here
+	multiError := errors.NewMultiError(previousExecErrors...)
```
brikis98: Do you also need to fix the `if len(hooks) == 0` in `processHooks`?

PR #468, opened 2018-04-27
Removes Terragrunt's requirement for "//" in all module sources when using --terragrunt-source

### LINKED ISSUE #445: Possible to use --terragrunt-source with modules stored in git repo root?
Consider the following source string in Terragrunt: 

`git::ssh://git@git.host/Org/sample_module.git?ref=terragrunt`

Now consider the following repo structure:

```
/sample_module (terragrunt ✔) ᐅ tree
.
├── README.md
├── main.tf
├── outputs.tf
└── variables.tf
```

Everything works as expected when sourcing the modules from git.
However, when trying to develop locally, specifying `--terragrunt-source` breaks the module sourcing.
I assume this is because there is no "double slash" in the source string, so Terragrunt does not have anything to append to the module_root when running *-all* commands.

Is there a way in Terragrunt to make this work reasonably?

The use case here is that we plan on having separate repos for most modules, which simplifies module pinning and releases for us.

geekifier:
## Overview

Removes Terragrunt's requirement for "//" in all module sources when using --terragrunt-source.

This is accomplished by utilizing Hashicorp's go-getter to parse the module URLs.

The regex was updated to also parse out the module name from each URL if the
module lives at the root of a repo.
Test cases were updated to account for HTTP paths, forward slashes in ?ref=, as well as local file sources. 

The error check for a missing "//" was removed, as there are too many edge cases based on a large amount of source formats supported by Terraform. Terraform itself does not perform this validation on module download. It's possible that there is a place for an error suggesting potential missing `//` somewhere else in the code (e.g. when Terragrunt does not find the expected modules under a path).

The testing edge cases involved paths such as `github.com/hashicorp/modules/subdir/module` or `github.com/org/some_module?ref=feature/some_feature`.

While it is possible for Terragrunt to utilize an expanded set of URL validation functions (likely one for each type of supported module source), I thought this was going out of scope of this bugfix PR.

## Regex update

The regex will only run _if_ go-getter fails to obtain a subdir for a parsed module source (e.g. the module source does not contain `//`. Terraform normally does not care about this, but Terragrunt needs a valid name to use in the local module path later.

`(?:.+/)(.+?)(?:\?.+|\.)|(?:.+/)(.+)`

The regex basically matches everything after the last slash in the URL, but before a valid termination, such as end-of-line, `?`, or `.`.
The second capture group (and the conditional) are needed, because it is possible to include slashes in an otherwise valid source string via the ` ?ref=` parameter, e.g. `?ref=feature/some_branch`.

Fixes #445.

### DIFF
--- a/configstack/module.go
+++ b/configstack/module.go
@@ -7,6 +7,7 @@ import (
 	"github.com/gruntwork-io/terragrunt/options"
 	"github.com/gruntwork-io/terragrunt/shell"
 	"github.com/gruntwork-io/terragrunt/util"
+	"github.com/hashicorp/go-getter"
 	"path/filepath"
 	"regexp"
 	"strings"
@@ -107,7 +108,7 @@ func resolveTerraformModule(terragruntConfigPath string, terragruntOptions *opti
 	return &TerraformModule{Path: modulePath, Config: *terragruntConfig, TerragruntOptions: opts}, nil
 }
 
-var moduleUrlRegexp = regexp.MustCompile(`.+//(.+?)(?:$|\?.+$)`)
+var moduleUrlRegexp = regexp.MustCompile(`(?:.+/)(.+?)(?:\?.+|\.)|(?:.+/)(.+)`)
 
 // If one of the xxx-all commands is called with the --terragrunt-source parameter, then for each module, we need to
 // build its own --terragrunt-source parameter by doing the following:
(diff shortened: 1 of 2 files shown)


configstack/module.go
```diff
@@ -128,12 +129,30 @@ func getTerragruntSourceForModule(modulePath string, moduleTerragruntConfig *con
...
 	}
 
-	matches := moduleUrlRegexp.FindStringSubmatch(moduleTerragruntConfig.Terraform.Source)
-	if len(matches) != 2 {
+	var moduleRelativePath string
+
+	// use go-getter to split the module source string into a valid URL and subdirectory (if // is present)
+	moduleURL, moduleSubdir := getter.SourceDirSubdir(moduleTerragruntConfig.Terraform.Source)
+
+	// if both URL and subdir are missing, something went terribly wrong
```
brikis98: lol 😄

configstack/module.go
```diff
@@ -128,12 +129,30 @@ func getTerragruntSourceForModule(modulePath string, moduleTerragruntConfig *con
 		return "", nil
 	}
 
-	matches := moduleUrlRegexp.FindStringSubmatch(moduleTerragruntConfig.Terraform.Source)
-	if len(matches) != 2 {
+	var moduleRelativePath string
```
brikis98: I suspect we'll need a method to separate the module path from the URL elsewhere, so would you mind extracting this into a new `getModulePathFromSourceUrl(sourceUrl string) (string, error)` method? That would also simplify the code a bit and remove the need for this `moduleRelativePath` variable, as you could `return` the relative path as soon as you found it.

configstack/module.go
```diff
@@ -107,7 +108,7 @@ func resolveTerraformModule(terragruntConfigPath string, terragruntOptions *opti
 	return &TerraformModule{Path: modulePath, Config: *terragruntConfig, TerragruntOptions: opts}, nil
 }
 
-var moduleUrlRegexp = regexp.MustCompile(`.+//(.+?)(?:$|\?.+$)`)
+var moduleUrlRegexp = regexp.MustCompile(`(?:.+/)(.+?)(?:\?.+|\.)|(?:.+/)(.+)`)
```
brikis98: This regex is hard to read and understand. I stared at it for a long time and can't fully convince myself it'll work in all cases.

Perhaps we can simplify the regex portion by pre-processing the source URL a bit:

1. Strip of the query string (anything after and including a `?`)
1. Strip off the repo name (anything after and including a `.`)

Now we simply take the portion after the last slash.

geekifier: I like the query string removal idea. This avoids having to deal with slashes in the ref parameter. We can probably just use one of the split functions for that, no need for regex.

I have tested the regex against a number of cases, but I agree simpler == better. I'll look into implementing your proposals.

configstack/module.go
geekifier: Good idea, I will work on that.

geekifier:
My apologies, I noticed some failed tests, and have reverted the latest commit. I am double checking what happened.

geekifier:
Sorry about that - I was bitten by scoped variables and trying to reduce the amount of `return` statements :). Should be good to go.

brikis98:
New binaries should be here shortly: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.14.9

PR #469, opened 2018-04-27
Merge before_hook and after_hook block from child to parent .tfvars

### LINKED ISSUE #467: Hooks are not merged when including parent config
Hello.  First of all, I'm a new user of terragrunt and I really like the direction it is headed.  It is saving us a lot of time, so thank you!

I think I've spotted an issue with the new `*_hook` feature.

We are set up like the live example where we have possible `terraform` blocks in the parent and module directories.

I'm expecting to be able to add `before_hood` and `after_hook` in the child `terraform.tfvars` (inside the module) at the same time as using `include`, however, it seems the hooks aren't merged up.

I think there is a missing merge for hooks here (this line is where `extra_arguments` are handled):
https://github.com/gruntwork-io/terragrunt/blob/master/config/config.go#L316

For now, we can work around this by copying the `extra_arguments` block to all child config files, but it would be great if it would merge up correctly.

Thank you.

Example child config:

```
terragrunt = {
    terraform {
        source = "../awesome-module"

        before_hook "before" {
            commands = [ "apply", "plan" ]
            execute = [
                "echo", "Foo"
            ]
            run_on_error = false
        }

        after_hook "after" {
            commands = [ "apply", "plan" ]
            execute = [
                "echo", "Bar"
            ]
            run_on_error = true
        }
    }

    include {
    }
}
```

Example parent config:

```
terragrunt = {

  remote_state {
    backend = "s3"

    config {
      bucket  = "tfstate"
      key     = "us-east-1/${path_relative_to_include()}/terraform.tfstate"
      region  = "us-east-1"
      encrypt = true
      acl     = "private"
      profile = "awesome-profile"
    }
  }

  terraform {
  
    extra_arguments "custom_vars" {
      optional_var_files = [
        "${get_tfvars_dir()}/${find_in_parent_folders("account.tfvars", "ignore")}"
      ]
    }
  }
}
```

dtryon:
This PR attempts to fix #467 

`BeforeHooks` and `AfterHooks` are merged from child to parent .tfvars.  If the name of the child is the same as the name of the parent, the child overrides.  Otherwise, they are appended.

This is very similar to the `TerraformExtraArgs` logic.

Thank you!


config/config_test.go
```diff
@@ -504,6 +504,36 @@ func TestMergeConfigIntoIncludedConfig(t *testing.T) {
...
+			&TerragruntConfig{Terraform: &TerraformConfig{}},
+			&TerragruntConfig{Terraform: &TerraformConfig{AfterHooks: []Hook{Hook{Name: "childHooks"}}}},
+		},
+		{
+			&TerragruntConfig{Terraform: &TerraformConfig{AfterHooks: []Hook{Hook{Name: "childHooks"}}}},
+			&TerragruntConfig{Terraform: &TerraformConfig{AfterHooks: []Hook{Hook{Name: "parentHooks"}}}},
+			&TerragruntConfig{Terraform: &TerraformConfig{AfterHooks: []Hook{Hook{Name: "parentHooks"}, Hook{Name: "childHooks"}}}},
+		},
+		{
+			&TerragruntConfig{Terraform: &TerraformConfig{AfterHooks: []Hook{Hook{Name: "overrideHooks", Commands: []string{"child-apply"}}}}},
```
brikis98: These are great tests 👍 

Could you add a few more test cases:

1. The `Terraform` config is `nil` in parent but child has hooks.
1. The `Terraform` config is `nil` in child but parent has hooks. 
1. The parent and child both have multiple hooks, a few of which overlap, and a of of which don't.

README.md
```diff
@@ -522,6 +522,11 @@ settings as follows:
   * The child's `extra_arguments` will be placed _after_ the parent's `extra_arguments` on the terraform command line.
   * Therefore, if a child's and parent's `extra_arguments` include `.tfvars` files with the same variable defined,
     the value from the `.tfvars` file from the child's `extra_arguments` will be used by terraform.
+* If a `before_hook` or `after_hook` block in the child has the same name as the hook block in the parent,
+  then the child's block will override the parent's.
+  * Specifying an empty hook block in a child with the same name will effectively remove the parent's block.
```
brikis98: We might want to add a test for this

dtryon:
@brikis98 Ok, I hope I've completed all the tests.  I decided that one integration test that exercised *most* of the functionality was better then many.  I don't think it is too big :)

I also added unit tests for `nil` config, overlapping hooks and empty child overriding parent.
Thanks again for any feedback

dtryon:
Ah, I'll add then to .gitignore and remove them +1

dtryon:
Ok, out files removed.  Thanks again for helping me along. :)

brikis98:
New release is here: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.14.8. Binaries should show up in a few minutes.

Issue #344, opened 2017-11-02
Allow S3 backend key to be defined in child


mike-schiller:
It seems very common to specify remote state in the root terraform.tfvars file.  In the context of the S3 backend this works well for the encryption, bucket, region, and lock table settings. However, specifying key = "${path_relative_to_include()}/terraform.tfstate" assumes that I'm never going to change where a child is located in my directory structure.

It would be nice if the key could be set to a fixed value in the child's .tfvars file so that the structure of the state does not depend on the relative location of the child within the Terragrunt directory structure.

brikis98:
> However, specifying key = "${path_relative_to_include()}/terraform.tfstate" assumes that I'm never going to change where a child is located in my directory structure.

You can change the child directory, but you should move its remote state in that case too. If you store remote state in S3, that's just a matter of `aws s3 --recursive mv s3://<bucket>/<old-path> s3://<bucket>/<new-path>`. Keeping the path of the code in sync with the path of the state makes it much, much easier to figure out which state files belong to which code.

Moreover, you can override the `remote_state` block in any child, so if you really do need something custom, it's possible to do it. The one catch is that, IIRC, it overrides the *entire* `remote_state` configuration, so you'd have to copy/paste the other values... I suppose the values could be merged together...

mike-schiller:
You raise a good point about moving the state in S3. Perhaps the better way to handle this would be to add a 'mv' command to Terragrunt that moves both the local folder and the remote storage location at the same time.

brikis98:
> Perhaps the better way to handle this would be to add a 'mv' command to Terragrunt that moves both the local folder and the remote storage location at the same time.

Ah, that's a good idea. I'd definitely welcome a PR for that.

PR #470, opened 2018-04-30
Allow s3 remote state key to be configured in child

Sprinkle:
Allows S3 remote state key to be part of the actual terraform module rather than specified in the terraform.tfvars.

To make it easier to use terraform_remote_state with nested module

### DIFF
--- a/remote/remote_state.go
+++ b/remote/remote_state.go
@@ -109,9 +109,18 @@ func (remoteState *RemoteState) differsFrom(existingBackend *TerraformBackend, t
 		}
 	}
 
-	if !reflect.DeepEqual(existingBackend.Config, remoteState.Config) {
-		terragruntOptions.Logger.Printf("Backend config has changed from %s to %s", existingBackend.Config, remoteState.Config)
-		return true
+	// Compare the states by looping through the map, assume it is different if either a key doesn't exist in the `existingBackend` or a value
+	// is different
+	for key, value := range remoteState.Config {
+		if existingValue, found := existingBackend.Config[key]; found {
+			if existingValue != value {
+				terragruntOptions.Logger.Printf("Backend config has changed from %s to %s", existingBackend.Config, remoteState.Config)
+				return true
+			}
+		} else {
+			terragruntOptions.Logger.Printf("Backend config has new key from %s to %s", existingBackend.Config, remoteState.Config)
+			return true
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


brikis98:
Not sure I follow. Could you explain your use case a bit more and provide some examples?

Sprinkle:
Its a little bit like #344, say I have modules structure like below.

```
├── backend-app
│   └── main.tf
├── frontend-app
│   └── main.tf
├── mysql
│   └── main.tf
└── vpc
    └── main.tf
```

In each main.tf I would have a partial s3 state configured.

```hcl
terraform {
  # The configuration for this backend will be mostly filled in by Terragrunt
  backend "s3" {
    key = "module_name/something.tfstate"
  }
}
```

So if I have to move or restructure module folders, and because I use terraform_remote_state in other modules to reference various state files, this would allow flexibility to have state files follow a different structure.

remote/remote_state.go
```diff
@@ -109,9 +109,18 @@ func (remoteState *RemoteState) differsFrom(existingBackend *TerraformBackend, t
 		}
 	}
 
-	if !reflect.DeepEqual(existingBackend.Config, remoteState.Config) {
-		terragruntOptions.Logger.Printf("Backend config has changed from %s to %s", existingBackend.Config, remoteState.Config)
-		return true
+	// Compare the states by looping through the map, assume it is different if either a key doesn't exist in the `existingBackend` or a value
+	// is different
+	for key, value := range remoteState.Config {
```
brikis98: This is only checking the top-level keys/values, whereas `DeepEqual` goes into any nested structures too, so this isn't quite doing the same check.

remote/remote_state.go
```diff
@@ -109,9 +109,18 @@ func (remoteState *RemoteState) differsFrom(existingBackend *TerraformBackend, t
...
+		if existingValue, found := existingBackend.Config[key]; found {
+			if existingValue != value {
+				terragruntOptions.Logger.Printf("Backend config has changed from %s to %s", existingBackend.Config, remoteState.Config)
+				return true
+			}
+		} else {
+			terragruntOptions.Logger.Printf("Backend config has new key from %s to %s", existingBackend.Config, remoteState.Config)
+			return true
```
brikis98: Both the `if` and `else` return `true`. What does this actually change? Just the log statement?

brikis98:
PR not updated for over a year, closing

Issue #314, opened 2017-10-12
Support `--help` flag


kevinburke:
One of `terragrunt plan -h`, `terragrunt plan --help`, `terragrunt help plan` should print help text about the command and the options it takes.

In particular I am trying to determine if I can plan the files in a directory without having it as my working directory, e.g. `terragrunt plan /path/to/directory` instead of `cd /path/to/directory && terragrunt plan`.

brikis98:
What do you see when you run `terragrunt plan --help` now?

kevinburke:
It just proceeds with a plan, e.g. if I run plan in a directory without terraform.tfvars it exits because it couldn't find the config file, or it proceeds trying to clone modules, etc.

brikis98:
Hm, definitely a bug. PRs are very welcome.

jckuester:
Here is a little fix: https://github.com/gruntwork-io/terragrunt/pull/473

brikis98:
Fixed by #473. Thank you @jckuester!

PR #473, opened 2018-05-06
Bugfix: show help for terraform commands

jckuester:
This PR fixes the issue described in https://github.com/gruntwork-io/terragrunt/issues/314:

If there is a --help, -help or -h flag provided by a user
for a command that is forwarded to terraform (e.g. terragrunt plan --help),
don't throw any error because a config file is missing
or other checks fail, but show the help text (this should have
precedence).


brikis98:
New binaries will be here shortly: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.14.10

PR #488, opened 2018-06-13
WIP: Allow tagging of s3 bucket and dynamo db

BernhardBln:
This is WIP - no tests were created for this yet, but the tagging itself works already.

Please give some feedback if that looks ok, then I will start adding some tests too.

And remember I'm totally new to go...

**Note**: I decided against putting the tags outside of config, but inside of remote_state, for two reasons:

1. (Minor) - I was not able to get access to the whole remote_state object in the remote_state_s3 file
2. (Major) - I think it does not make sense to have an option which is s3-specific top-level in remote_state - if you ever add another provider, you would have to explain that these two config items can be set only if the backend is s3, which I find worse than putting it inside of config and filtering out later (which I do in two places, when reading the config in remote_state_s3, and when iterating over the config items to prepare the command line options for terraform)

If you still think I should place them outside of remote_state, I will change that, but I find the second argument quite strong.

So configuration looks like this now:

```
terragrunt = {

  remote_state {
    backend = "s3"
    config {
      bucket = "..."
      [...]

      s3_bucket_tags {
        owner = "team name"
        name = "Terraform state storage"
      }

      dynamotable_tags {
        owner = "team name"
        name = "Terraform lock table"
      }

    }
  }
}
```

You can set arbitrary tags (just follow the limitations of AWS, I do not check e.g. if a tag value is too long etc.

In case the tag mappings are missing, nothing is tagged (when they are there but are empty, an empty tagging is made, which could be prevented if you think that's necessary). 

In case they appear multiple times (like, if s3_bucket_tags appears twice or more), an error is printed and we stop.

### DIFF
--- a/README.md
+++ b/README.md
@@ -486,6 +486,16 @@ terragrunt = {
       region         = "us-east-1"
       encrypt        = true
       dynamodb_table = "my-lock-table"
+      
+      s3_bucket_tags {
+        owner = "terragrunt integration test"
+        name = "Terraform state storage"
+      }
+
+      dynamotable_tags {
+        owner = "terragrunt integration test"
+        name = "Terraform lock table"
+      }
     }
   }
 }
@@ -494,7 +504,9 @@ terragrunt = {
 The `remote_state` block supports all the same [backend types](https://www.terraform.io/docs/backends/types/index.html)
 as Terraform. The next time you run `terragrunt`, it will automatically configure all the settings in the
 `remote_state.config` block, if they aren't configured already, by calling [terraform
-init](https://www.terraform.io/docs/commands/init.html).
+init](https://www.terraform.io/docs/commands/init.html). The config options `s3_bucket_tags` and 
+`dynamotable_tags` are only valid for backend `s3`. They are used by terragrunt and are **not** passed on to 
+terraform. See section [Create remote state and locking resources automatically](#create-remote-state-and-locking-resources-automatically).
 
 In each of the **child** `terraform.tfvars` files, such as `mysql/terraform.tfvars`, you can tell Terragrunt to
 automatically include all the settings from the root `terraform.tfvars` file as follows:
@@ -562,11 +574,19 @@ they don't already exist:
 * **S3 bucket**: If you are using the [S3 backend](https://www.terraform.io/docs/backends/types/s3.html) for remote
   state storage and the `bucket` you specify in `remote_state.config` doesn't already exist, Terragrunt will create it
   automatically, with [versioning enabled](http://docs.aws.amazon.com/AmazonS3/latest/dev/Versioning.html).
+  
+  In addition, you can let terragrunt tag the bucket with custom tags that you specify in 
+  `remote_state.config.s3_bucket_tags`. See sample configuration in section 
+  [Filling in remote state settings with Terragrunt](#filling-in-remote-state-settings-with-terragrunt).
 
 * **DynamoDB table**: If you are using the [S3 backend](https://www.terraform.io/docs/backends/types/s3.html) for
   remote state storage and you specify a `dynamodb_table` (a [DynamoDB table used for
   locking](https://www.terraform.io/docs/backends/types/s3.html#dynamodb_table)) in `remote_state.config`, if that table
   doesn't already exist, Terragrunt will create it automatically, including a primary key called `LockID`.
+  
+  In addition, you can let terragrunt tag the DynamoDB table with custom tags that you specify in
+  `remote_state.config.dynamotable_tags`. See sample configuration in section 
+  [Filling in remote state settings with Terragrunt](#filling-in-remote-state-settings-with-terragrunt).
 
 **Note**: If you specify a `profile` key in `remote_state.config`, Terragrunt will automatically use this AWS profile
 when creating the S3 bucket or DynamoDB table.
--- a/README.md
+++ b/README.md
@@ -492,7 +492,7 @@ terragrunt = {
         name = "Terraform state storage"
       }
 
-      dynamotable_tags {
+      dynamodb_table_tags {
         owner = "terragrunt integration test"
         name = "Terraform lock table"
       }
@@ -505,7 +505,7 @@ The `remote_state` block supports all the same [backend types](https://www.terra
 as Terraform. The next time you run `terragrunt`, it will automatically configure all the settings in the
 `remote_state.config` block, if they aren't configured already, by calling [terraform
 init](https://www.terraform.io/docs/commands/init.html). The config options `s3_bucket_tags` and 
-`dynamotable_tags` are only valid for backend `s3`. They are used by terragrunt and are **not** passed on to 
+`dynamodb_table_tags` are only valid for backend `s3`. They are used by terragrunt and are **not** passed on to 
 terraform. See section [Create remote state and locking resources automatically](#create-remote-state-and-locking-resources-automatically).
 
 In each of the **child** `terraform.tfvars` files, such as `mysql/terraform.tfvars`, you can tell Terragrunt to
@@ -585,7 +585,7 @@ they don't already exist:
   doesn't already exist, Terragrunt will create it automatically, including a primary key called `LockID`.
   
   In addition, you can let terragrunt tag the DynamoDB table with custom tags that you specify in
-  `remote_state.config.dynamotable_tags`. See sample configuration in section 
+  `remote_state.config.dynamodb_table_tags`. See sample configuration in section 
   [Filling in remote state settings with Terragrunt](#filling-in-remote-state-settings-with-terragrunt).
 
 **Note**: If you specify a `profile` key in `remote_state.config`, Terragrunt will automatically use this AWS profile
--- a/README.md
+++ b/README.md
@@ -489,12 +489,12 @@ terragrunt = {
       
       s3_bucket_tags {
         owner = "terragrunt integration test"
-        name = "Terraform state storage"
+        name  = "Terraform state storage"
       }
 
       dynamodb_table_tags {
         owner = "terragrunt integration test"
-        name = "Terraform lock table"
+        name  = "Terraform lock table"
       }
     }
   }
--- a/dynamodb/dynamo_lock_table.go
+++ b/dynamodb/dynamo_lock_table.go
@@ -37,7 +37,7 @@ func CreateDynamoDbClient(awsRegion, awsProfile string, iamRoleArn string, terra
 }
 
 // Create the lock table in DynamoDB if it doesn't already exist
-func CreateLockTableIfNecessary(tableName string, client *dynamodb.DynamoDB, terragruntOptions *options.TerragruntOptions) error {
+func CreateLockTableIfNecessary(tableName string, tags map[string]string, client *dynamodb.DynamoDB, terragruntOptions *options.TerragruntOptions) error {
 	tableExists, err := LockTableExistsAndIsActive(tableName, client)
 	if err != nil {
 		return err
@@ -45,7 +45,7 @@ func CreateLockTableIfNecessary(tableName string, client *dynamodb.DynamoDB, ter
 
 	if !tableExists {
 		terragruntOptions.Logger.Printf("Lock table %s does not exist in DynamoDB. Will need to create it just this first time.", tableName)
-		return CreateLockTable(tableName, DEFAULT_READ_CAPACITY_UNITS, DEFAULT_WRITE_CAPACITY_UNITS, client, terragruntOptions)
+		return CreateLockTable(tableName, tags, DEFAULT_READ_CAPACITY_UNITS, DEFAULT_WRITE_CAPACITY_UNITS, client, terragruntOptions)
 	}
 
 	return nil
@@ -67,7 +67,7 @@ func LockTableExistsAndIsActive(tableName string, client *dynamodb.DynamoDB) (bo
 
 // Create a lock table in DynamoDB and wait until it is in "active" state. If the table already exists, merely wait
 // until it is in "active" state.
-func CreateLockTable(tableName string, readCapacityUnits int, writeCapacityUnits int, client *dynamodb.DynamoDB, terragruntOptions *options.TerragruntOptions) error {
+func CreateLockTable(tableName string, tags map[string]string, readCapacityUnits int, writeCapacityUnits int, client *dynamodb.DynamoDB, terragruntOptions *options.TerragruntOptions) error {
 	tableCreateDeleteSemaphore.Acquire()
 	defer tableCreateDeleteSemaphore.Release()
 
@@ -81,7 +81,7 @@ func CreateLockTable(tableName string, readCapacityUnits int, writeCapacityUnits
 		&dynamodb.KeySchemaElement{AttributeName: aws.String(ATTR_LOCK_ID), KeyType: aws.String(dynamodb.KeyTypeHash)},
 	}
 
-	_, err := client.CreateTable(&dynamodb.CreateTableInput{
+	createTableOutput, err := client.CreateTable(&dynamodb.CreateTableInput{
 		TableName:            aws.String(tableName),
 		AttributeDefinitions: attributeDefinitions,
 		KeySchema:            keySchema,
@@ -99,7 +99,44 @@ func CreateLockTable(tableName string, readCapacityUnits int, writeCapacityUnits
 		}
 	}
 
-	return waitForTableToBeActive(tableName, client, MAX_RETRIES_WAITING_FOR_TABLE_TO_BE_ACTIVE, SLEEP_BETWEEN_TABLE_STATUS_CHECKS, terragruntOptions)
+	err = waitForTableToBeActive(tableName, client, MAX_RETRIES_WAITING_FOR_TABLE_TO_BE_ACTIVE, SLEEP_BETWEEN_TABLE_STATUS_CHECKS, terragruntOptions)
+
+	if err != nil {
+		return err
+	}
+
+	err = tagTableITagsGiven(tags, createTableOutput, client, terragruntOptions)
+
+	if err != nil {
+		return errors.WithStackTrace(err)
+	}
+
+	return nil
+}
+
+func tagTableITagsGiven(tags map[string]string, createTableOutput *dynamodb.CreateTableOutput, client *dynamodb.DynamoDB, terragruntOptions *options.TerragruntOptions) error {
+
+	if tags == nil {
+		terragruntOptions.Logger.Printf("No tags for lock table given.")
+		return nil
+	}
+
+	// we were able to create the table successfully, now add tags
+	terragruntOptions.Logger.Printf("Adding tags to lock table: %s", tags)
+
+	var tagsConverted []*dynamodb.Tag
+
+	for k, v := range tags {
+		tagsConverted = append(tagsConverted, &dynamodb.Tag{Key: aws.String(k), Value: aws.String(v)})
+	}
+
+	var input = dynamodb.TagResourceInput{
+		ResourceArn: createTableOutput.TableDescription.TableArn,
+		Tags:        tagsConverted}
+
+	_, err := client.TagResource(&input)
+
+	return err
 }
 
 // Delete the given table in DynamoDB
--- a/dynamodb/dynamo_lock_table.go
+++ b/dynamodb/dynamo_lock_table.go
@@ -105,7 +105,7 @@ func CreateLockTable(tableName string, tags map[string]string, readCapacityUnits
 		return err
 	}
 
-	err = tagTableITagsGiven(tags, createTableOutput, client, terragruntOptions)
+	err = tagTableIfTagsGiven(tags, createTableOutput, client, terragruntOptions)
 
 	if err != nil {
 		return errors.WithStackTrace(err)
@@ -114,9 +114,9 @@ func CreateLockTable(tableName string, tags map[string]string, readCapacityUnits
 	return nil
 }
 
-func tagTableITagsGiven(tags map[string]string, createTableOutput *dynamodb.CreateTableOutput, client *dynamodb.DynamoDB, terragruntOptions *options.TerragruntOptions) error {
+func tagTableIfTagsGiven(tags map[string]string, createTableOutput *dynamodb.CreateTableOutput, client *dynamodb.DynamoDB, terragruntOptions *options.TerragruntOptions) error {
 
-	if tags == nil {
+	if tags == nil || len(tags) == 0 {
 		terragruntOptions.Logger.Printf("No tags for lock table given.")
 		return nil
 	}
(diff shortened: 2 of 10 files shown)


dynamodb/dynamo_lock_table.go
```diff
@@ -99,7 +99,44 @@ func CreateLockTable(tableName string, readCapacityUnits int, writeCapacityUnits
...
+	err = tagTableITagsGiven(tags, createTableOutput, client, terragruntOptions)
+
+	if err != nil {
+		return errors.WithStackTrace(err)
+	}
+
+	return nil
+}
+
+func tagTableITagsGiven(tags map[string]string, createTableOutput *dynamodb.CreateTableOutput, client *dynamodb.DynamoDB, terragruntOptions *options.TerragruntOptions) error {
```
brikis98: s/tagTableITagsGiven/tagTableIfTagsGiven

dynamodb/dynamo_lock_table.go
```diff
@@ -99,7 +99,44 @@ func CreateLockTable(tableName string, readCapacityUnits int, writeCapacityUnits
...
+
+	if tags == nil {
```
brikis98: Check if `len(tags) == 0` too

dynamodb/dynamo_lock_table.go
```diff
@@ -99,7 +99,44 @@ func CreateLockTable(tableName string, readCapacityUnits int, writeCapacityUnits
...
+
+	var tagsConverted []*dynamodb.Tag
+
+	for k, v := range tags {
+		tagsConverted = append(tagsConverted, &dynamodb.Tag{Key: aws.String(k), Value: aws.String(v)})
+	}
+
+	var input = dynamodb.TagResourceInput{
+		ResourceArn: createTableOutput.TableDescription.TableArn,
+		Tags:        tagsConverted}
```
brikis98: Make sure to run `make fmt`

remote/remote_state_s3.go
```diff
@@ -15,6 +15,13 @@ import (
 )
 
 // A representation of the configuration options available for S3 remote state
```
brikis98: Comment needs to be moved to `RemoteStateConfigS3` and the comment here should explain that `ExtendedRemoteStateConfigS3` is used to store custom settings that are not used by Terraform, but by Terragrunt only.

remote/remote_state.go
```diff
@@ -122,6 +122,11 @@ func (remoteState *RemoteState) differsFrom(existingBackend *TerraformBackend, t
 func (remoteState RemoteState) ToTerraformInitArgs() []string {
 	backendConfigArgs := []string{}
 	for key, value := range remoteState.Config {
+
+		if key == "s3_bucket_tags" || key == "dynamotable_tags" {
+			continue
+		}
```
brikis98: Instead of doing this, couldn't you use `remoteState.Config.remoteStateConfigS3`?

remote/remote_state_s3.go
```diff
@@ -117,8 +126,28 @@ func parseS3Config(config map[string]interface{}) (*RemoteStateConfigS3, error)
...
+	var s3Config RemoteStateConfigS3
+	var extendedConfig ExtendedRemoteStateConfigS3
+
+	if err := mapstructure.Decode(config, &s3Config); err != nil {
+		return nil, errors.WithStackTrace(err)
+	}
+
+	if err := mapstructure.Decode(config, &extendedConfig); err != nil {
+		return nil, errors.WithStackTrace(err)
+	}
```
brikis98: Isn't it enough to just do this? Doesn't it automatically fill in the `remoteStateConfigS3` portion?

remote/remote_state_s3.go
```diff
@@ -78,29 +85,31 @@ func (s3Initializer S3Initializer) NeedsInitialization(config map[string]interfa
...
 	if err != nil {
 		return err
 	}
 
-	if err := validateS3Config(s3Config, terragruntOptions); err != nil {
+	if err := validateS3Config(s3ConfigExtended, terragruntOptions); err != nil {
 		return err
 	}
 
+	var s3Config = s3ConfigExtended.remoteStateConfigS3
```
brikis98: Why the intermediate variable? Can't you use `s3ConfigExtended.Region`, `s3ConfigExtended.Endpoint`, etc below?

remote/remote_state_s3.go
```diff
@@ -173,20 +212,71 @@ func checkIfVersioningEnabled(s3Client *s3.S3, config *RemoteStateConfigS3, terr
...
+	if err := EnableVersioningForS3Bucket(s3Client, &config.remoteStateConfigS3, terragruntOptions); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func TagS3Bucket(s3Client *s3.S3, config *ExtendedRemoteStateConfigS3, terragruntOptions *options.TerragruntOptions) error {
+
+	if len(config.S3BucketTags) == 0 {
```
brikis98: Check for `nil` too

remote/remote_state_s3.go
```diff
@@ -173,20 +212,71 @@ func checkIfVersioningEnabled(s3Client *s3.S3, config *RemoteStateConfigS3, terr
...
+		terragruntOptions.Logger.Print("No tags for S3 bucket given.")
```
brikis98: `Println`?

remote/remote_state_s3.go
```diff
@@ -173,20 +212,71 @@ func checkIfVersioningEnabled(s3Client *s3.S3, config *RemoteStateConfigS3, terr
...
+
 	return nil
+
+}
+
+func convertTags(tags map[string]string) []*s3.Tag {
+
+	var tagsConverted []*s3.Tag
+
+	var i = 0
```
brikis98: What's `i` for?

remote/remote_state_s3.go
```diff
@@ -173,20 +212,71 @@ func checkIfVersioningEnabled(s3Client *s3.S3, config *RemoteStateConfigS3, terr
...
+
+func TagS3Bucket(s3Client *s3.S3, config *ExtendedRemoteStateConfigS3, terragruntOptions *options.TerragruntOptions) error {
+
+	if len(config.S3BucketTags) == 0 {
+		terragruntOptions.Logger.Print("No tags for S3 bucket given.")
+		return nil
+	}
+
+	// There must be one entry in the list
+	var tags = config.S3BucketTags[0]
```
brikis98: Why only `[0]`?

remote/remote_state_s3.go
```diff
@@ -258,7 +349,12 @@ func createLockTableIfNecessary(s3Config *RemoteStateConfigS3, terragruntOptions
 		return err
 	}
 
-	return dynamodb.CreateLockTableIfNecessary(s3Config.GetLockTableName(), dynamodbClient, terragruntOptions)
+	var tags map[string]string = nil
+	if len(tagsDeclarations) == 1 {
+		tags = tagsDeclarations[0]
+	}
```
brikis98: Again, why only `[0]`?

dynamodb/dynamo_lock_table.go
BernhardBln: Done

dynamodb/dynamo_lock_table.go
BernhardBln: Done

dynamodb/dynamo_lock_table.go
BernhardBln: Done

remote/remote_state_s3.go
BernhardBln: Done

remote/remote_state.go
BernhardBln: Nope, because we are in remote_state.go, not in remote_state_s3.go, where we have this object.

The cleanest would probably be to have the individual backends return the configuration.

remote/remote_state_s3.go
BernhardBln: No, that doesn't work. It does create the extendedConfig.remoteStateConfigS3 object, but all the fields are empty.

remote/remote_state_s3.go
BernhardBln: No, s3configExtended has these fields only:

* remoteStateConfigS3
* S3BucketTags
* DynamotableTags

If I didn't have the intermediate variable, I would have to use `s3ConfigExtended.remoteStateConfigS3.Region`, which I find less readable.

My goal was to separate the actual terraform backend config from the two extra keys that are just for terragrunt.

remote/remote_state_s3.go
BernhardBln: done

remote/remote_state_s3.go
BernhardBln: Ups, yeah.

remote/remote_state_s3.go
BernhardBln: I used that before I figured out there is an append function ;-) Removed

remote/remote_state_s3.go
BernhardBln: S3BucketTags can only be empty, or have one element.

If you add several config blocks like this,

```
    config {
      bucket =...

      s3_bucket_tags {
        owner = "team name"
      }

      s3_bucket_tags {
         service = "Terraform"
      }
    }
```

I'm returning this MultipleTagsDeclarations error (see at bottom of remote_state_s3.go file), because in that case, we would have to merge them together, and make sure tags are not set twice, which I found a bit too much and hence decided to allow only one tags declaration.

In case somebody specifies a tag twice in the same configuration, it seems the last value wins, hence no problem for us.

remote/remote_state_s3.go
BernhardBln: See above

remote/remote_state.go
BernhardBln: And I just implemented that, because it does not make sense that the generic remote_state needs to know which keys to filter for remote_state_s3. I encapsulated this there now, and extended the `RemoteStateInitializer` interface by another method `GetTerraformInitArgs`.

BernhardBln:
Done fixing remarks from review; will now start to write some tests.

BernhardBln:
Done with tests. Hope they are sufficient so far.

Please check again, @brikis98 - especially my solution with the `RemoteStateInitializer`, in order to move the filtering by s3-backend-specific keys from `remote_state` to `remote_state_s3`.

dynamodb/dynamo_lock_table.go
```diff
@@ -99,7 +99,48 @@ func CreateLockTable(tableName string, readCapacityUnits int, writeCapacityUnits
...
 
-	return waitForTableToBeActive(tableName, client, MAX_RETRIES_WAITING_FOR_TABLE_TO_BE_ACTIVE, SLEEP_BETWEEN_TABLE_STATUS_CHECKS, terragruntOptions)
+	err = waitForTableToBeActive(tableName, client, MAX_RETRIES_WAITING_FOR_TABLE_TO_BE_ACTIVE, SLEEP_BETWEEN_TABLE_STATUS_CHECKS, terragruntOptions)
+
+	if err != nil {
+		return err
+	}
+
+	if createTableOutput != nil && createTableOutput.TableDescription != nil && createTableOutput.TableDescription.TableArn != nil {
+		// Do not tag in case somebody else had created the table
```
brikis98: Oh, that's a very interesting point. Checking `createTableOutput` makes sense; why check the description field?

dynamodb/dynamo_lock_table_test.go
```diff
@@ -81,7 +81,52 @@ func TestCreateLockTableIfNecessaryTableAlreadyExists(t *testing.T) {
...
+func TestTableTagging(t *testing.T) {
+	t.Parallel()
+
+	mockOptions, err := options.NewTerragruntOptionsForTest("dynamo_lock_test_utils")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var tags = make(map[string]string)
+	tags["team"] = "team A"
```
brikis98: Can be simplified to:

```go
tags := map[string]string{"team": "team A"}
```

dynamodb/dynamo_lock_table_test.go
```diff
@@ -81,7 +81,52 @@ func TestCreateLockTableIfNecessaryTableAlreadyExists(t *testing.T) {
...
+		assertTags(tags, tableName, client, t)
+
+		// Try to create the table the second time and make sure you get no errors
+		err = CreateLockTableIfNecessary(tableName, nil, client, mockOptions)
+		assert.Nil(t, err, "Unexpected error: %v", err)
+	})
+}
+
+func assertTags(expectedTags map[string]string, tableName string, client *dynamodb.DynamoDB, t *testing.T) {
+	var description, err = client.DescribeTable(&dynamodb.DescribeTableInput{TableName: &tableName})
```
brikis98: Idiomatic approach is `aws.String(tableName)` instead of `&tableName`.

dynamodb/dynamo_lock_table_test.go
```diff
@@ -81,7 +81,52 @@ func TestCreateLockTableIfNecessaryTableAlreadyExists(t *testing.T) {
...
+	}
+
+	var tags = make(map[string]string)
+	tags["team"] = "team A"
+
+	// Create the table the first time
+	withLockTableTagged(t, tags, func(tableName string, client *dynamodb.DynamoDB) {
+		assertCanWriteToTable(t, tableName, client)
+
+		assertTags(tags, tableName, client, t)
```
brikis98: Nice tests 👍

remote/remote_state.go
```diff
@@ -26,6 +26,8 @@ type RemoteStateInitializer interface {
 
 	// Initialize the remote state
 	Initialize(config map[string]interface{}, terragruntOptions *options.TerragruntOptions) error
+
+	GetTerraformInitArgs(config map[string]interface{}) map[string]interface{}
```
brikis98: Please add a comment explaining what this method does and why it exists

remote/remote_state.go
```diff
@@ -121,7 +123,16 @@ func (remoteState *RemoteState) differsFrom(existingBackend *TerraformBackend, t
 // Convert the RemoteState config into the format used by the terraform init command
 func (remoteState RemoteState) ToTerraformInitArgs() []string {
 	backendConfigArgs := []string{}
-	for key, value := range remoteState.Config {
+
+	initializer, hasInitializer := remoteStateInitializers[remoteState.Backend]
+	if !hasInitializer {
+		// return empty array
+		return backendConfigArgs
```
brikis98: Doesn't seem backwards compatible to return an empty array here. Instead, you should loop over `remoteState.Config` as before to fill `backendConfigArgs`. Or, more accurately, you should set `config = remoteState.Config` by default and then override it with `initializer.GetTerraformInitArgs` if `initializer` exists.

remote/remote_state_s3.go
brikis98: If the syntax is:

```hcl
    config {
      bucket =...

      s3_bucket_tags = {
        owner = "team name"
      }

      s3_bucket_tags = {
         service = "Terraform"
      }
    }
```

Could you make `S3BucketTags` a map rather than a list of maps?

remote/remote_state_test.go
```diff
@@ -18,10 +21,20 @@ func TestToTerraformInitArgs(t *testing.T) {
...
+				"service": "Terraform"},
+
+			"dynamotable_tags": map[string]interface{}{
+				"team":    "team name",
+				"name":    "Terraform state storage",
+				"service": "Terraform"}},
 	}
 	args := remoteState.ToTerraformInitArgs()
 
+	// must not contain s3_bucket_tags or dynamotable_tags
```
brikis98: Does this test check the S3 tags were actually created?

dynamodb/dynamo_lock_table.go
BernhardBln: That part crashed in a test, so I guessed it was necessary as well. Just to be on the safe side...

remote/remote_state_test.go
BernhardBln: No, it just calls ToTerraformInitArgs.

It seems there is no test at all that actually creates the bucket, if you guys add one later I'd be happy to add the check for the tags.

remote/remote_state.go
BernhardBln: Done

remote/remote_state.go
BernhardBln: True, fixed and also wrote a test for that.

dynamodb/dynamo_lock_table_test.go
BernhardBln: Done

dynamodb/dynamo_lock_table_test.go
BernhardBln: Done

BernhardBln:
@brikis98 - fixed everything and added a new test for the case that there is no backend available to return the config.

For some reason, I cannot reply to your comment about "Could you make S3BucketTags a map rather than a list of maps?", there is no "Reply" field. 

Is there a way to I can get this as a map, rather than slice of map?

No matter if I use

```
s3_bucket_tags = {
    service = "Terraform"
}
```

or

```
s3_bucket_tags {
    service = "Terraform"
}
```

I get them as slice of map: `'s3_bucket_tags' expected a map, got 'slice'`

dynamodb/dynamo_lock_table.go
brikis98: Crashed in a test? Because you didn't check description? Do you have the stack trace?

remote/remote_state_test.go
brikis98: A bunch of the tests in [integration_test.go](https://github.com/gruntwork-io/terragrunt/blob/master/test/integration_test.go) create an S3 bucket. See the fixtures folder.

dynamodb/dynamo_lock_table.go
BernhardBln:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xa0 pc=0x132babc]

goroutine 190 [running]:
github.com/gruntwork-io/terragrunt/dynamodb.CreateLockTable(0xc420014cc0, 0x16, 0x0, 0x1, 0x1, 0xc42000e0c8, 0xc4201e0000, 0x0, 0x0)
        /Users/bernhardmercateo/go/src/github.com/gruntwork-io/terragrunt/dynamodb/dynamo_lock_table.go:108 +0x4dc
github.com/gruntwork-io/terragrunt/dynamodb.CreateLockTableIfNecessary(0xc420014cc0, 0x16, 0x0, 0xc42000e0c8, 0xc4201e0000, 0x0, 0x0)
        /Users/bernhardmercateo/go/src/github.com/gruntwork-io/terragrunt/dynamodb/dynamo_lock_table.go:48 +0x161
github.com/gruntwork-io/terragrunt/dynamodb.TestCreateLockTableConcurrency.func1(0xc420020f80, 0xc420014cc0, 0x16, 0xc42000e0c8, 0xc4201e0000, 0xc4201c43c0)
        /Users/bernhardmercateo/go/src/github.com/gruntwork-io/terragrunt/dynamodb/dynamo_lock_table_test.go:47 +0x89
created by github.com/gruntwork-io/terragrunt/dynamodb.TestCreateLockTableConcurrency
        /Users/bernhardmercateo/go/src/github.com/gruntwork-io/terragrunt/dynamodb/dynamo_lock_table_test.go:45 +0x1b9
exit status 2
```

Hm, you think if createTableOutput is not nil, I can rely on the other two fields to be filled and don't need to check?

remote/remote_state_test.go
BernhardBln: Ok found and extended it to check for tags of both s3 and dynamo db.

The test case I changed runs on my machine, if I let the whole test run, I get this, but I assume it's not related to my changes:

```
[terragrunt] [/var/folders/3h/j7zf9_595vb97k60z03zt2nh0000gp/T/terragrunt-test258914403/fixture-output-all/env1/app1] 2018/06/15 11:42:21 Running command: terraform apply -input=false -auto-approve -var terraform_remote_state_s3_bucket="terragrunt-test-bucket-mgqpz9"
Copying configuration from "file:///Users/bernhard/go/src/github.com/gruntwork-io/terragrunt/test/fixture-download/hello-world"...
Initializing modules...
- module.hello
- module.remote
  Getting source "github.com/gruntwork-io/terragrunt.git//test/fixture-download/hello-world?ref=v0.9.9"
Error downloading modules: Error loading modules: subdir "test/fixture-download/hello-world" not found
--- FAIL: TestLocalDownloadWithHiddenFolder (0.14s)
	integration_test.go:817: Failed to run Terragrunt command 'terragrunt apply --terragrunt-non-interactive --terragrunt-working-dir fixture-download/local-with-hidden-folder' due to error: Hit multiple errors:
		exit status 1
```

dynamodb/dynamo_lock_table.go
brikis98: Oh, I misunderstood `TableDescription`; I thought it was an optional human-readable description, rather than a struct used to return data about the table such as its ARN. OK, this is probably fine then.

BernhardBln:
I extended the integration test (see my other comment), renamed the config key `dynamotable_tags` to `dynamodb_table_tags` for the sake of consistency, and also added the two new config keys to the README.

Let me know if there is anything we can do about this map/slice of map problem, otherwise I think we should be pretty much done soon :)

README.md
```diff
@@ -494,7 +504,9 @@ terragrunt = {
 The `remote_state` block supports all the same [backend types](https://www.terraform.io/docs/backends/types/index.html)
 as Terraform. The next time you run `terragrunt`, it will automatically configure all the settings in the
 `remote_state.config` block, if they aren't configured already, by calling [terraform
-init](https://www.terraform.io/docs/commands/init.html).
+init](https://www.terraform.io/docs/commands/init.html). The config options `s3_bucket_tags` and
```
brikis98: Nice! Thx for updating the docs :)

remote/remote_state_test.go
```diff
@@ -18,10 +21,37 @@ func TestToTerraformInitArgs(t *testing.T) {
...
+		Backend: "s4",
+		Config: map[string]interface{}{
+			"encrypt": true,
+			"bucket":  "my-bucket",
+			"key":     "terraform.tfstate",
+			"region":  "us-east-1"},
 	}
 	args := remoteState.ToTerraformInitArgs()
 
+	// no Backend initializer available, but command line args should still be passed on
```
brikis98: 👍

BernhardBln:
Sweet, thanks a lot, also for the patience :)

brikis98:
Whoops, almost forgot to create the new release! Here it is (new binaries should show up in a few min): https://github.com/gruntwork-io/terragrunt/releases/tag/v0.14.11

BernhardBln:
Thank you :)

Issue #471, opened 2018-05-04
apply-all -  External dependency resolution is a mess of inconsistent prompts


sgendler-stem:
If running apply-all in a subdirectory, but encountering templates with dependencies that are up a directory, it asks if I want to apply the external dependeny as well.  But it asks it with an inverted prompt, such that answering 'n' causes apply to occur, and answering 'y' causes it to be skipped.  I doubt, in the history of software development, that anyone has ever thought it intuitive to answer 'no' to make changes and 'yes' to cancel.  

Even worse, if there are lots of such dependencies, you may have to hit the same answer many times, and then the final prompt asks whether I am sure I want to do it, this time expecting me to answer 'y' to make the change and 'n' to stop.  And, of course if I am busily typing 'n', 'return', 'n', 'return', ... over and over again, it is all too likely that I will accidentally type 'n' to that final prompt, when what I actually want is 'y'.  This is the very definition of poor user experience - inconsistent, repetitive, and difficult to visually parse.  The prompts are so damn wordy, it can be hard to see exactly what module it is referring to as the string in question moves around from prompt to prompt.  

And why does it prompt me over and over again for whether I want to run the same external dependencies.  I already said 'no' 3 times (asking it to be run), so must I really answer 'no' a fourth time, too?  How about just prompting me once, no matter how many of the apply-all targets have the same external dependency?  It only seems to run apply on the external dependency once, so it clearly understands that things are supposed to be idempotent and it only needs to prompt me once, but nevertheless, it prompts me as many times as there are edges in the graph connecting to it.  It would be a whole lot easier to type the correct number of 'n' respnses before typing 'y' if I didn't have 30-50 prompts for the same template every time.

Finally, how about some help text or documentation of apply-all so I can find out if there is a flag which will skip all of that prompting?  I searched and searched, tried 'terragrunt apply-all help' which just starts prompting me, and 'terragrunt help apply-all' just says there is no help for apply-all.  I found nothing in the terragrunt docs themselves.  I'd really prefer to be able to run apply-all in my services directory without having to answer a prompt about each external dependency for each service.  Ideally, a flag which just blanket accepts or denies all external dependencies would be really useful.  As would documentation telling me what that flag is.

brikis98:
Did you try `--help`?

And `--terragrunt-non-interactive` (see [CLI options](https://github.com/gruntwork-io/terragrunt#cli-options))?

sgendler-stem:
[...]

PR #497, opened 2018-06-30
Reverse the dependency questions to match non-interactive behavior

natefaerber:
Since non-interactive answers "yes" to questions, we want the running of dependencies to depend on a "yes" answer instead of "no".

https://github.com/gruntwork-io/terragrunt/issues/471

I have run these test cases with `plan-all`:
* w/o --terragrunt-non-interactive: answer 'y' to all dependency questions
* w/o --terragrunt-non-interactive: answer 'n' to all dependency questions
* w/o --terragrunt-non-interactive: answer 'y' to first dependency question then 'n' to all others
* w/ --terragrunt-non-interactive

In all cases, dependencies were run or not-run as expected.

### DIFF
--- a/configstack/module.go
+++ b/configstack/module.go
@@ -197,12 +197,12 @@ func resolveExternalDependenciesForModules(moduleMap map[string]*TerraformModule
 				continue
 			}
 
-			alreadyApplied, err := confirmExternalDependencyAlreadyApplied(module, externalDependency, terragruntOptions)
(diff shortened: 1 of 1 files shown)


brikis98:
I had to [update a couple tests](https://github.com/gruntwork-io/terragrunt/commit/c0a0d72635d023ade569ae24cba12dd6a254f7a9), but now the build has passed. New release is here: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.15.0. New binaries should show up soon.

PR #498, opened 2018-07-01
Add minimum IAM policy permissions example

shoekstra:
For a more restrictive policy where you may have multiple terragrunt
repos storing their state files in a single bucket and using a single
table for state locking, you still want to limit what parts of the
bucket can be modified.

Signed-off-by: Stephen Hoekstra <stephenhoekstra@gmail.com>


shoekstra:
@brikis98 as requested, let me know if the wording needs amending.

Issue #388, opened 2017-12-02
Terragrunt should detect missing module and provider download errors and automatically re-run init to fix them


brikis98:
Terraform has two types of frustrating errors:

1. You add a `module` or modify its version and then next time you run `plan` or `apply`, you get an error that you need to re-run `get`. If Terraform *knows* you need to do that, why doesn't it just do it for you?

1. You run `init` to download the code for providers and get a TLS handshake error or similar timeout. This happens especially often when using `xxx-all` commands, so many downloads are happening concurrently. Not sure if this is a bug in Terraform's download code or throttling on their servers, but I've had many `xxx-all` commands fail because of this.

Terragrunt should detect these errors and retry the previous command automatically.

eefi:
Adding a second motivation for this issue... I'm running across this when I want to upgrade the version of a provider module I'm using.

We have a `boilerplate.tf` that I symlink into all of my modules that has this clause:

```HCL
provider "azurerm" {
  version = "1.0.0"
}
```

We do this so that we can test and coordinate upgrades of the provider at our leisure rather than having them randomly happen.

Unfortunately, when we update the version to `1.0.1`, `terragrunt plan-all` fails because all of the existing Terragrunt work folders have init-ed 1.0.0 and need to be re-init-ed to get 1.0.1.

A stopgap measure could be to implement a `terragrunt init-all` that allows us to re-run `init` in all the modules the other `-all` commands would run. That might be easier to implement than auto-detection?

brikis98:
`init-all` sounds like a great interim solution!

brikis98:
And two more:

```
Error installing provider "aws": error fetching checksums: Get https://releases.hashicorp.com/terraform-provider-aws/0.1.4/terraform-provider-aws_0.1.4_SHA256SUMS: net/http: TLS handshake timeout.
```

```
Initializing provider plugins...
- Checking for available provider plugins on https://releases.hashicorp.com...

Error installing provider "template": error fetching checksums: Get https://releases.hashicorp.com/terraform-provider-template/1.0.0/terraform-provider-template_1.0.0_SHA256SUMS: net/http: TLS handshake timeout.
```

brikis98:
One more!

```
Failed to load backend: 
Error configuring the backend "s3": RequestError: send request failed
caused by: Post https://sts.amazonaws.com/: net/http: TLS handshake timeout
```

brikis98:
[...]

PR #503, opened 2018-07-02
GH #388 -- describe auto-retry;

lhagemann:
auto-retry will address transient errors and necessary module inits

### DIFF
--- a/README.md
+++ b/README.md
@@ -1137,6 +1137,7 @@ This section contains detailed documentation for the following aspects of Terrag
 1. [Interpolation Syntax](#interpolation-syntax)
 1. [Before & After Hooks](#before-and-after-hooks)
 1. [Auto-Init](#auto-init)
+1. [Auto-Retry](#auto-retry)
 1. [CLI options](#cli-options)
 1. [Configuration](#configuration)
 1. [Migrating from Terragrunt v0.11.x and Terraform 0.8.x and older](#migrating-from-terragrunt-v011x-and-terraform-08x-and-older)
@@ -1718,9 +1719,7 @@ As mentioned [above](#extra_arguments-for-init), `extra_arguments` can be config
 to allow customization of the `terraform init` command.
 
 Note that there might be cases where terragrunt does not properly detect that `terraform init` needs be called.
-In this case, terraform would fail.  Just run `terragrunt init` to correct this situation.
-
-
+In this case, terraform would fail.  Running `terragrunt init` again corrects this situation. See [Auto-Retry](#auto-retry).
 
 For some use cases, it might be desirable to disable Auto-Init.
 For example, if each user wants to specify a different `-plugin-dir` option to `terraform init` (and therefore it cannot be put in `extra_arguments`).
@@ -1730,6 +1729,17 @@ To disable Auto-Init, use the `--terragrunt-no-auto-init` command line option or
 Disabling Auto-Init means that you _must_ explicitly call `terragrunt init` prior to any other terragrunt commands for a particular configuration.
 
 If Auto-Init is disabled, and terragrunt detects that `terraform init` needs to be called, then terragrunt will fail.
+
+### Auto-Retry
+
+_Auto-Retry_ is a feature of `terragrunt` that will automatically address situations where a `terraform` command needs to be re-run.
+
+Terraform can fail with transient errors which can be addressed by simply retrying the command again. In the event `terragrunt` finds one of these errors, the command will be re-run again automatically.
+
+In the event of a module or provider update, `terraform` will fail with a message to run `terraform init`, in this case `terragrunt` will re-run the `init` command and then retry the command. 
+
+The flag `--terragrunt-no-auto-retry` can be sent to tell `terragrunt` to not retry these commands.
+
 
 ### CLI Options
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1731,6 +1730,17 @@ Disabling Auto-Init means that you _must_ explicitly call `terragrunt init` prio
 
 If Auto-Init is disabled, and terragrunt detects that `terraform init` needs to be called, then terragrunt will fail.
 
+### Auto-Retry
+
+_Auto-Retry_ is a feature of `terragrunt` that will automatically address situations where a `terraform` command needs to be re-run.
+
+Terraform can fail with transient errors which can be addressed by simply retrying the command again. In the event `terragrunt` finds one of these errors, the command will be re-run again automatically.
```
brikis98: How is the list of transient errors maintained? I presume somewhere in the code we have a default list; perhaps we can link to the file (e.g., `retryable-errors.go`)? Also, can the user specify extra ones in the Terragrunt config?

README.md
```diff
@@ -1731,6 +1730,17 @@ Disabling Auto-Init means that you _must_ explicitly call `terragrunt init` prio
...
+
+In the event of a module or provider update, `terraform` will fail with a message to run `terraform init`, in this case `terragrunt` will re-run the `init` command and then retry the command.
```
brikis98: Will Terragrunt re-run only on `init` or any command that fails (e.g., `apply`, `destroy`)?

README.md
brikis98: Also, consider adding an example here: "You run `terragrunt apply` and get an error such as [...] Terragrunt knows this is a transient error and will automatically retry..."

README.md
brikis98: Also, how many times will Terragrunt retry? Once? Infinite? Is there a pause before retrying?

cli/cli_app.go
```diff
@@ -326,15 +326,25 @@ func runTerragruntWithConfig(terragruntOptions *options.TerragruntOptions, terra
...
 }
 
-func runTerraformCommandIfNoErrors(possibleErrors error, terragruntOptions *options.TerragruntOptions) error {
+func runTerraformCommandIfNoErrors(possibleErrors error, terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig) error {
 	if possibleErrors == nil {
-		return shell.RunTerraformCommand(terragruntOptions, terragruntOptions.TerraformCliArgs...)
+		output, err := shell.RunTerraformCommandAndCaptureOutput(terragruntOptions, terragruntOptions.TerraformCliArgs...)
+
+		// show the output we captured
+		terragruntOptions.Logger.Println(output)
```
brikis98: Does no output show until this point? If so, we'll want to actually stream the output to `stdout` / `stderr` *and* capture it. Otherwise, for a long-running command (e.g., RDS), it may be 20 min with no output. See here for how to do both: https://github.com/gruntwork-io/terratest/blob/master/modules/shell/command.go#L51

cli/cli_app.go
```diff
@@ -500,6 +510,16 @@ func runTerraformInit(terragruntOptions *options.TerragruntOptions, terragruntCo
...
 }
 
+// automatically run `init` for us when terraform says we should and continues to run the original command from terrgruntConfig
+// returns error if unable to run init (possibly because auto-init is false), or if the underlying command fails
+func autorunTerrafromInit (terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig, tfoutput string) error {
+	terragruntOptions.Logger.Printf("terraform returned %s, running init to fix it\n\n", tfoutput)
+	if err := runTerraformInit(terragruntOptions, terragruntConfig, nil); err != nil {
+			return err
+	}
+	return runTerragruntWithConfig(terragruntOptions, terragruntConfig, false)
```
brikis98: Does this end up re-running `init` further down? I just want to make sure we don't end up in an infinite retry loop here!

cli/cli_app.go
```diff
@@ -326,15 +326,25 @@ func runTerragruntWithConfig(terragruntOptions *options.TerragruntOptions, terra
...
-func runTerraformCommandIfNoErrors(possibleErrors error, terragruntOptions *options.TerragruntOptions) error {
+func runTerraformCommandIfNoErrors(possibleErrors error, terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig) error {
 	if possibleErrors == nil {
-		return shell.RunTerraformCommand(terragruntOptions, terragruntOptions.TerraformCliArgs...)
+		output, err := shell.RunTerraformCommandAndCaptureOutput(terragruntOptions, terragruntOptions.TerraformCliArgs...)
+
+		// show the output we captured
+		terragruntOptions.Logger.Println(output)
+		if err != nil {
+			if strings.Contains(output,"may need to run 'terraform init'") {
```
brikis98: Extract into a separate method named something like `isRetryableInitError`.

options/options.go
```diff
@@ -77,6 +77,9 @@ type TerragruntOptions struct {
 	// exposed here primarily so we can set it to a low value at test time.
 	MaxFoldersToCheck int
 
+	// Maximum number of times to retry errors matching RetryableTerraformErrors
+	MaxRetries int
```
brikis98: RetryableTerraformErrors is not actually defined?

cli/cli_app.go
lhagemann: yes, that is one of the concerns I had with changing the `runTerraformCommandIfNoErrors` function (called from`runTerragruntWithConfig`) to use the `CaptureOutput` command. The streaming output from the underlying terraform execution is lost until we replay it.

Looking at the `terratest` implementation, I'll have to figure out to incorporate that with the `RunShellCommand` which is what the `CaptureOutput` command ultimately uses, while opening a buffer for stdOut

Debating yet another `runTerraformCommand....` function, or modifying the `CaptureOutput` to stream the Pipe.

Issue #489, opened 2018-06-14
Add "prevent destroy" to Terragrunt


brikis98:
Terragrunt should support a `prevent_destroy` flag that, if set to `true`, will prevent `terraform destroy` from being executed. This is useful for modules you want to carefully protect, such as a database, or a module that provides auth (which isn't a "dependency" of other modules, but if you delete it, you won't be able to access anything else!).

```hcl
terragrunt = {
  prevent_destroy = true
}
```

brikis98:
Added in #525.

PR #525, opened 2018-07-15
Add prevent_destroy flag

ozerovandrei:
Add boolean flag to protect selected Terraform resources.

For #489

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -680,7 +680,7 @@ func checkProtectedModule(terragruntOptions *options.TerragruntOptions, terragru
 		return nil
 	}
 	if terragruntConfig.PreventDestroy {
-		return errors.WithStackTrace(ModuleIsProtected("Module is protected, set prevent_destroy to false or delete this flag to allow destroying of the module."))
+		return errors.WithStackTrace(ModuleIsProtected{Opts: terragruntOptions})
 	}
 	return nil
 }
@@ -723,8 +723,10 @@ func (path NoTerraformFilesFound) Error() string {
 	return fmt.Sprintf("Did not find any Terraform files (*.tf) in %s", string(path))
 }
 
-type ModuleIsProtected string
+type ModuleIsProtected struct {
+	Opts *options.TerragruntOptions
+}
 
 func (err ModuleIsProtected) Error() string {
-	return string(err)
+	return fmt.Sprintf("Module is protected by the prevent_destroy flag in %s. Set it to false or delete it to allow destroying of the module.", err.Opts.TerragruntConfigPath)
 }
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -332,6 +332,10 @@ func runTerragruntWithConfig(terragruntOptions *options.TerragruntOptions, terra
 		}
 	}
 
+	if err := checkProtectedModule(terragruntOptions, terragruntConfig); err != nil {
+		return err
+	}
+
 	beforeHookErrors := processHooks(terragruntConfig.Terraform.GetBeforeHooks(), terragruntOptions)
 	terraformError := runTerraformCommandIfNoErrors(beforeHookErrors, terragruntOptions)
 	postHookErrors := processHooks(terragruntConfig.Terraform.GetAfterHooks(), terragruntOptions, beforeHookErrors, terraformError)
@@ -670,6 +674,17 @@ func validateAll(terragruntOptions *options.TerragruntOptions) error {
 	return stack.Validate(terragruntOptions)
 }
 
+// checkProtectedModule checks if module is protected via the "prevent_destroy" flag
+func checkProtectedModule(terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig) error {
+	if firstArg(terragruntOptions.TerraformCliArgs) != "destroy" {
+		return nil
+	}
+	if terragruntConfig.PreventDestroy {
+		return errors.WithStackTrace(ModuleIsProtected("Module is protected, set prevent_destroy to false or delete this flag to allow destroying of the module."))
+	}
+	return nil
+}
+
 // Custom error types
 
 type UnrecognizedCommand string
@@ -706,4 +721,10 @@ type NoTerraformFilesFound string
 
 func (path NoTerraformFilesFound) Error() string {
 	return fmt.Sprintf("Did not find any Terraform files (*.tf) in %s", string(path))
+}
+
+type ModuleIsProtected string
+
+func (err ModuleIsProtected) Error() string {
+	return string(err)
 }
--- a/README.md
+++ b/README.md
@@ -1815,6 +1815,23 @@ Terragrunt figures out the path to its config file according to the following ru
   terragrunt plan --terragrunt-config example.tfvars --var-file example.tfvars		
  ```
 
+#### prevent_destroy
+
+Terragrunt `prevent_destroy` boolean flag allows you to protect selected Terraform module. It will prevent `destroy`
+or `destroy-all` command to actually destroy resources of the protected module. This is useful for modules you want
+to carefully protect, such as a database, or a module that provides auth.
+
+Example:
+
+```hcl
+terragrunt = {
+  terraform {
+    source = "git::git@github.com:foo/modules.git//app?ref=v0.0.3"
+  }
+
+  prevent_destroy = true
+}
+```
 
 ##### Previous Versions of Terragrunt
 
--- a/config/config.go
+++ b/config/config.go
@@ -18,23 +18,25 @@ const OldTerragruntConfigPath = ".terragrunt"
 
 // TerragruntConfig represents a parsed and expanded configuration
 type TerragruntConfig struct {
-	Terraform    *TerraformConfig
-	RemoteState  *remote.RemoteState
-	Dependencies *ModuleDependencies
+	Terraform      *TerraformConfig
+	RemoteState    *remote.RemoteState
+	Dependencies   *ModuleDependencies
+	PreventDestroy bool
 }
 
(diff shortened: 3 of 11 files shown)


ozerovandrei:
@brikis98 Hello.

I started working on this for the issue #489 and I want to ask how to better deal with this configuration parameter.

As far as I understand, it's better to use the Terraform's native `prevent_destroy` [lifetime flag](https://www.terraform.io/docs/configuration/resources.html#prevent_destroy)?
If so, better way to do it is to pass 
```
lifecycle {
  prevent_destroy = true
}
```
to all of resources that are declared in the modules if user provided
```
terragrunt {
  prevent_destroy = true
}
```

Am I thinking into right direction?

brikis98:
@ozerovandrei Thanks for taking a crack at this! This would be a very useful feature.

The code looks good so far. I think the next step is to check which Terraform command is being executed by the user, and if it is `destroy`, and `prevent_destroy` is set to `true`, just log a message explaining this module is protected (and the user can only destroy it by removing the `prevent_destroy` flag), and gracefully exit. I don't think you need to integrate with Terraform's `prevent_destroy`, as that is on a per-resource basis, and it would be very tricky to make it work.

ozerovandrei:
@brikis98 thank you for the explanation.
What about `destroy-all` command? Does check need to be added to this command too?

brikis98:
`destroy-all` eventually goes through the same code path as `destroy`, so as long as you put your check in the right place, it'll work automatically for `destroy-all` too :)

ozerovandrei:
@brikis98 seems that I found a right place to put my checks 😄 

I've noticed lack of `cli/cli_app_test.go` file to test methods of the `cli_app.go`. Can I create such file to add a test for my new method? Or you have some different plans to test `cli_app`?

Aside from tests this PR only needs documentation updates for a new configuration parameter. So it's almost ready for a full review.

brikis98:
> I've noticed lack of cli/cli_app_test.go file to test methods of the cli_app.go. Can I create such file to add a test for my new method?

Sure! If there are unit tests that make sense to add for `cli_app.go`, putting them in `cli_app_test.go` makes sense. If they are more integration tests (i.e., they run the whole app, rather than executing a function or two), those should go under `test/integration_test.go`, which has quite a few integration tests already.

> Aside from tests this PR only needs documentation updates for a new configuration parameter. So it's almost ready for a full review.

Fantastic. Let me know when it's ready. Thanks!

ozerovandrei:
@brikis98 I've decided to add just a single integration test with a fixture.

This PR is ready for review.

cli/cli_app.go
```diff
@@ -670,6 +674,17 @@ func validateAll(terragruntOptions *options.TerragruntOptions) error {
 	return stack.Validate(terragruntOptions)
 }
 
+// checkProtectedModule checks if module is protected via the "prevent_destroy" flag
+func checkProtectedModule(terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig) error {
+	if firstArg(terragruntOptions.TerraformCliArgs) != "destroy" {
+		return nil
+	}
+	if terragruntConfig.PreventDestroy {
+		return errors.WithStackTrace(ModuleIsProtected{Opts: terragruntOptions})
```
brikis98: Hm, there's a bit of gotcha here, when someone runs `destroy-all`, this module will exit with an error, causing anything that depends on it to exit with an error, and causing the overall process to exit with an error. On the one hand, this is good, for if module B depends on A, and you put `prevent_destroy` in B, then not only does that protect B, but it'll also prevent A from being destroyed (without which B would also break!). On the other hand, if this error prevents unrelated parts of the dependency graph from executing, that would be unfortunate.

Could you add a test for this? E.g., add a test fixture with:

- Module A
- Module B which depends on A and has `prevent_destroy`
- Module C
- Module D which depends on C
- Module E

If you run `destroy-all` on this graph, my hope is everything but module A and B would be deleted...

ozerovandrei: @brikis98 it seems that I stuck there.

How is better to check if module was destroyed in the integration tests?
I created separate directory with fixtures of modules (that are using `hello-world` source), I can easily run `destroy-all` on them and check if the error has kind of `configstack.MultiError`.

But how can I check if module C, D, E were actually destroyed or not?

brikis98: I can think of a few options. One way to do it is to run `terragrunt output` or `terragrunt show` in each module, which should tell you if it has been applied or not. After running `apply-all`, you should see all have been applied; after running `destroy-all`, you should see that C, D, and E have been destroyed.

ozerovandrei: @brikis98 Thank you for the advice with `terragrunt show` (I just started working with Terragrunt, so I haven't got all the stuff yet).
I added the `TestPreventDestroyDependencies` that tests several modules.

brikis98:
Hm, weird. `TestPreventDestroyDependencies` is failing consistently in CircleCI... But when I run it locally, it passes... I'll have to look into this tomorrow.

```
 [terragrunt] [/home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/test/fixture-download/local-with-prevent-destroy-dependencies/module-c] 2018/07/21 11:44:43 Module / home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/test/fixture-download/local-with-prevent-destroy-dependencies/module-c has finished with an error: Hit multiple     errors:^M
 exit status 1^M
 [terragrunt] [/home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/test/fixture-download/local-with-prevent-destroy-dependencies/module-d] 2018/07/21 11:44:43          Dependency /home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/test/fixture-download/local-with-prevent-destroy-dependencies/module-c of module /home/ubuntu/.         go_workspace/src/github.com/gruntwork-io/terragrunt/test/fixture-download/local-with-prevent-destroy-dependencies/module-d just finished with an error. Module /home/ubuntu/.        go_workspace/src/github.com/gruntwork-io/terragrunt/test/fixture-download/local-with-prevent-destroy-dependencies/module-d will have to return an error too.^M
 [terragrunt] [/home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/test/fixture-download/local-with-prevent-destroy-dependencies/module-d] 2018/07/21 11:44:43 Module / home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/test/fixture-download/local-with-prevent-destroy-dependencies/module-d has finished with an error: Cannot process   module Module /home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/test/fixture-download/local-with-prevent-destroy-dependencies/module-d (dependencies: [/home/        ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/test/fixture-download/local-with-prevent-destroy-dependencies/module-c]) because one of its dependencies, Module /home/  ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/test/fixture-download/local-with-prevent-destroy-dependencies/module-c (dependencies: []), finished with an error: Hit   multiple errors:^M
 exit status 1^M
 --- FAIL: TestPreventDestroyDependencies (38.98s)^M
   integration_test.go:903: Failed to run Terragrunt command 'terragrunt apply-all --terragrunt-non-interactive --terragrunt-working-dir fixture-download/local-with-prevent-destroy- dependencies' due to error: Encountered the following errors:^M
     Cannot process module Module /home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/test/fixture-download/local-with-prevent-destroy-dependencies/module-b            (dependencies: [/home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/test/fixture-download/local-with-prevent-destroy-dependencies/module-a]) because one of its        dependencies, Module /home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/test/fixture-download/local-with-prevent-destroy-dependencies/module-a (dependencies: []),    finished with an error: Hit multiple errors:^M
     exit status 1^M
     Hit multiple errors:^M
     exit status 1^M
     Cannot process module Module /home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/test/fixture-download/local-with-prevent-destroy-dependencies/module-d            (dependencies: [/home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/test/fixture-download/local-with-prevent-destroy-dependencies/module-c]) because one of its        dependencies, Module /home/ubuntu/.go_workspace/src/github.com/gruntwork-io/terragrunt/test/fixture-download/local-with-prevent-destroy-dependencies/module-c (dependencies: []),    finished with an error: Hit multiple errors:^M
     exit status 1^M
     Hit multiple errors:^M
     exit status 1^M
     Hit multiple errors:^M
     exit status 1^M
```

brikis98:
@ozerovandrei Apologies for the delay on this. I'm traveling this whole week and won't be able to dig into this test failure for a while. If you're able to repro it locally and have any idea of the cause in the meantime, let me know! This is a great feature and I want to get it released as soon as I can, but sometimes things get a bit busy, so thank you for your patience.

ozerovandrei:
@brikis98 I'll try to look into it, sure.

brikis98:
OK, finally had time to sit down and figure this one out. Turns out that there was a bug in the file copying code which prevented the `terraform.tfvars` files from being copied when running in CircleCI, as the `GOPATH` in CircleCi is in a folder with a dot in it  (`~/.go_workspace`). Fix: https://github.com/gruntwork-io/terragrunt/commit/0bdb5f9ebbcad4ebe2c104bb1c51c186d3ad10f7.

Release with this new feature: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.16.4. New binaries should show up shortly.

PR #540, opened 2018-08-12
Adds options attribute to track state of the terraform command

### LINKED ISSUE #539: after_hook on init command executes twice
I like to use an `after_hook` on `init` to further reduce code duplication, usually at least copying a common or backend .tf file into the tmp dir, but this setup ends up executing the hook twice... once after `init -from-module` retrieves the source, and again after `init` executes to initalize modules, the backend, and plugins.

I really only need the hook to execute after retrieving the source. It's not too big a deal if all I'm doing is copying a file like this, but I have other less common use cases where I'm doing something more (rendering the .tf file using jinja 😭 ), where the duplication really slows things down.

Not too sure on the best approach to avoid this, but could maybe create new command-types recognized by the hooks. Such as `init-from-module` and `init-configure` or somesuch, to identify the different steps.

lorengordon:
The "init" hook command was previously overloaded, and would execute
hooks twice. Once when retrieving source using -from-module, and once
to configure the backend, modules, and plugins.

The new options attribute, `TerraformCommand`, is first set to the
terraform command when the `TerragruntOptions` are parsed from the
CLI arguments.

The `TerraformCommand` attribute is updated when `runTerraformInit`
determines that it needs to retrieve the source using `-from-module`.

Finally, the `TerraformCommand` is now used to determine whether a
hook should execute.

There are now two "init" commands that execute hooks:

* `init-from-module`: executes when the source is retrieved
* `init`: executes for any other init invocation, to configure
the backend, modules, and plugins

Fixes #539


options/options.go
```diff
@@ -97,6 +100,7 @@ func NewTerragruntOptions(terragruntConfigPath string) (*TerragruntOptions, erro
 	return &TerragruntOptions{
 		TerragruntConfigPath:   terragruntConfigPath,
 		TerraformPath:          "terraform",
+		TerraformCommand:       "",
```
brikis98: Please add this to the `Clone` method further down below.

brikis98:
> Can you maybe point me at the detail I'm missing?

IIRC, Terragrunt calls in the "second time" only if (a) it needs to download modules or (b) it needs to configure a remote state backend. Perhaps if your code has neither, it's skipping the second `init`?

lorengordon:
I'm having some trouble tracing the logic through the code. Gonna write the test cases for different condition and see if that helps me figure it out. Here are the test cases I think are relevant...

* no-source-no-backend
* no-source-with-backend
* with-source-no-backend
* with-source-with-backend

brikis98:
The code you're probably looking for is this: https://github.com/gruntwork-io/terragrunt/blob/master/cli/cli_app.go#L448. Those test cases sound good 👍

lorengordon:
Hmm, those test cases are still not quite demonstrating the nuance I'm seeing. Think I might need to split out local vs remote sources...

* no-source-no-backend
* no-source-with-backend
* local-source-no-backend
* local-source-with-backend
* remote-source-no-backend
* remote-source-with-backend

lorengordon:
Aha, no, I just had mapped one of the constants to the wrong path. 😭

lorengordon:
Ok, so the only failing test case with this approach is `with-source-no-backend`. I think I understand why now. I didn't quite have the terragrunt workflow correct.

I think the best approach to better align with how terragrunt works will actually be to rework this so that instead of an `init-from-module` command, we implement an `init-backend` command...

This way, the `command` remains as `init` throughout, and where the backend is configured it would be set to `init-backend`.

brikis98:
> I think the best approach to better align with how terragrunt works will actually be to rework this so that instead of an init-from-module command, we implement an init-backend command...
> This way, the command remains as init throughout, and where the backend is configured it would be set to init-backend.

Not sure I follow. The `init-from-module` and `init` workflow made sense to me. What is the distinction with `init` and `init-backend`?

lorengordon:
The problem I observed was that hooks on the `init` command would execute twice. I always use a source and a backend, which seems to be the only config that causes this to happen. 

For the first hook execution, I would see the command:

```
terraform init // -backend-config ... // -from-module ...
```

And for the second hook execution, I would see the same command, just without the `-from-module ...`:

```
terraform init // -backend-config ... //
```

This code path is triggered within `runTerragrunt()` when the test for `getTerraformSourceUrl()` calls `downloadTerraformSource()`. `downloadTerraformSource()` eventually calls `runTerraformInit()`, which calls `runTerragruntWithConfig()` with the necessary init options. `terraform init -from-module` then executes.

From here, things operate basically the same as they do without a `source` argument... `runTerragrunt()` executes `runTerragruntWithConfig()` again, but now with the original non-init command. And it's `prepareNonInitCommand()` that triggers the second init.

Ok, I mostly needed to write that out for myself to better understand the workflow. 😁 To the point at hand, the trouble with `init-from-module` basically comes down to expectations...

* `no-source-no-backend`:
  * `init-from-module` *does not* trigger hook, `init` command *does*. This behavior is EXPECTED
* `no-source-with-backend`:
  * `init-from-module` *does not* trigger hook, `init` command *does*. This behavior is EXPECTED
* `with-source-no-backend`:
  * `init-from-module` *does* trigger hook, `init` command *does not*. This behavior is **NOT EXPECTED**
* `with-source-with-backend`:
  * `init-from-module` *does* trigger hook, `init` command also triggers hook. This behavior is EXPECTED

I feel like you _should_ be able to trigger on `init` reliably, but this approach doesn't work without a backend (or another condition that causes `needsInit()` to return true).

Unfortunately, now having a better understanding of the workflow that is triggering the second init, I'm not sure I see a simple way to set the command accordingly... and yeah, we can take `init-backend` off the table as it would not be accurate, since there's really several conditions that trigger `needsInit()`.

I mean, I could just change the test to match what actually happens. . .

brikis98:
An alternative is to have just a single `init` command that hooks fire on and to track if we've already run a hook for the `init` command using a new field in `TerragruntOptions`.

lorengordon:
I'll give that a go tomorrow. I would like to preserve the use case where a before or after hook can copy/render .tf files in the working directly, though. That requires executing after the source is pulled in, but before the backend initializes...

lorengordon:
Reverted the prior changes and updated the tests to demonstrate clearly the use case failure. Go TDD, right?

```
=== RUN   TestTerragruntInitHookWithSourceWithBackend
=== PAUSE TestTerragruntInitHookWithSourceWithBackend
=== CONT  TestTerragruntInitHookWithSourceWithBackend
--- FAIL: TestTerragruntInitHookWithSourceWithBackend (9.19s)
        integration_test.go:1054: Copying fixture-hooks/init-once to /tmp/terragrunt-test858288851
        Error Trace:    integration_test.go:185
        Error:          Not equal:
                        expected: 2
                        actual: 4
        Messages:       Hooks on init command executed more than once
        integration_test.go:1237: Deleting test s3 bucket terragrunt-test-bucket-wpdrri
FAIL
exit status 1
FAIL    github.com/gruntwork-io/terragrunt/test 9.221s
```

Working on the other approach now. It's a little more difficult, since the hooks have state (before and after), but that state is not actually part of the hook attributes. So we need to track the hook state through the code. Hmm...

brikis98:
+1 for TDD.

FWIW, I tried to clean up the `init` logic before, but Terraform is a bit finicky with `init`, `-from-module`, and all the other params. E.g., See https://github.com/gruntwork-io/terragrunt/pull/523. Ideally, we'd only ever call `init` once... but I haven't been able to get that to work yet.

lorengordon:
Setting a flag on the `TerragruntOptions` doesn't seem to work... I think because the options are cloned in `runTerraformInit()`? Would changes to option attributes when init is executed also get applied on the original options? Fairly new to Go, so not sure how some of these things are supposed to work...

brikis98:
Ah, yea, you're right, we do clone it, specifically so we don't make accidental modifications to the original. 

I'm thinking the original implementation you had in [this comment](https://github.com/gruntwork-io/terragrunt/pull/540#issuecomment-412589620) may be our best bet as it most accurately reflects Terragrunt's actual behavior. In a separate PR, it may be worth looking into changing how we call `init` and `init -from-module`.

lorengordon:
Are there tests that are known to be failing right now? `go test -parallel 128 $(glide novendor)` is exiting non-zero, but I can't figure out which test is actually failing because there is so much stdout/stderr output from the integration tests.

brikis98:
I'm not aware of any existing test failures, but I'll let you know what happens when the tests finish for https://github.com/gruntwork-io/terragrunt/pull/542.

lorengordon:
Found the failures, but not sure what I could have done to cause them:

```
--- FAIL: TestLocalWithBackend (2.37s)
        integration_test.go:1062: Copying fixture-download to /tmp/terragrunt-test799597791
        integration_test.go:1084: Error walking file path fixture-download due to error: read fixture-download/local/.terragrunt-cache/nWnQU8XO3IG8-Pc7IW1iTN2kR6Y/K9T2yS3Cjskt4KDaVB9rdK7fFq0/.terraform/modules/123b8e4fd617c6cd09164b9df977046a: is a directory
        Error Trace:    integration_test.go:1301
                        asm_amd64.s:574
                        panic.go:377
                        testing.go:553
                        testing.go:597
                        integration_test.go:1084
                        integration_test.go:711
        Error:          Expected nil, but got: &awserr.requestError{awsError:(*awserr.baseError)(0xc420715300), statusCode:400, requestID:"__REQUESTID__"}
        Messages:       Unexpected error: ResourceNotFoundException: Requested resource not found: Table: terragrunt-lock-table-zuvqsi not found
                                status code: 400, request id: __REQUESTID__
        integration_test.go:1245: Deleting test s3 bucket terragrunt-test-bucket-stkhzd
        integration_test.go:1249: Failed to list object versions in s3 bucket terragrunt-test-bucket-stkhzd: NoSuchBucket: The specified bucket does not exist
                        status code: 404, request id: 056D080CFFB2239F, host id: __HOSTID__
--- FAIL: TestLocalWithMissingBackend (2.38s)
        integration_test.go:1062: Copying fixture-download to /tmp/terragrunt-test843155301
        integration_test.go:1084: Error walking file path fixture-download due to error: read fixture-download/local/.terragrunt-cache/nWnQU8XO3IG8-Pc7IW1iTN2kR6Y/K9T2yS3Cjskt4KDaVB9rdK7fFq0/.terraform/modules/123b8e4fd617c6cd09164b9df977046a: is a directory
        Error Trace:    integration_test.go:1301
                        asm_amd64.s:574
                        panic.go:377
                        testing.go:553
                        testing.go:597
                        integration_test.go:1084
                        integration_test.go:732
        Error:          Expected nil, but got: &awserr.requestError{awsError:(*awserr.baseError)(0xc420405140), statusCode:400, requestID:"__REQUESTID__"}
        Messages:       Unexpected error: ResourceNotFoundException: Requested resource not found: Table: terragrunt-lock-table-t0mwvr not found
                                status code: 400, request id: __REQUESTID__
        integration_test.go:1245: Deleting test s3 bucket terragrunt-test-bucket-d9emhi
        integration_test.go:1249: Failed to list object versions in s3 bucket terragrunt-test-bucket-d9emhi: NoSuchBucket: The specified bucket does not exist
                        status code: 404, request id: F51D38555F253BD8, host id: __HOSTID__
```

brikis98:
OK, https://github.com/gruntwork-io/terragrunt/pull/542 is now merged. Merge the latest from `master` and see if that helps your test cases behave more like you expect!

brikis98:
> Found the failures, but not sure what I could have done to cause them:

This turned out to be a bug in Terragrunt's test code if you ran the tests more than once. I fixed it in https://github.com/gruntwork-io/terragrunt/pull/542.

lorengordon:
>Merge the latest from master and see if that helps your test cases behave more like you expect!

@brikis98 Done and done! Behaves exactly as I expect now! 💯 Tests and Readme updated to reflect the now correct workflow.

> This turned out to be a bug in Terragrunt's test code if you ran the tests more than once. I fixed it in #542.

So, even with that patch those same two tests are still failing on my local dev box. Might just be an artifact of my setup.

brikis98:
> Done and done! Behaves exactly as I expect now! 💯 Tests and Readme updated to reflect the now correct workflow.

Fantastic! 🎉  

> So, even with that patch those same two tests are still failing on my local dev box. Might just be an artifact of my setup.

You're still getting the "Error walking file path fixture-download due to error: read xxx: is a directory" error? You're sure your local code has this patch?

https://github.com/gruntwork-io/terragrunt/blob/master/test/integration_test.go#L951-L955

The fix was that `info.IsDir()` incorrectly returns `false` for symlinked directories, so now the code checks for symlinks....

lorengordon:
Not quite, the errors are slightly different, and yes I definitely have the patch...

```
--- FAIL: TestLocalWithMissingBackend (6.98s)
        integration_test.go:1062: Copying fixture-download to /tmp/terragrunt-test164692187
        integration_test.go:1086: Error walking file path fixture-download due to error: lstat fixture-download/local-relative/.terragrunt-cache/2-wjG_ETsLVrHU1L8Ejo8E7yRbQ/sTtvdyfU8Hlx-1oGm8v0ocjiYS8/relative/main.tf: no such file or directory
        Error Trace:    integration_test.go:1303
                        asm_amd64.s:574
                        panic.go:377
                        testing.go:553
                        testing.go:597
                        integration_test.go:1086
                        integration_test.go:732
--- FAIL: TestLocalWithBackend (7.32s)
        integration_test.go:1062: Copying fixture-download to /tmp/terragrunt-test252649925
        integration_test.go:1086: Error walking file path fixture-download due to error: lstat fixture-download/local-relative/.terragrunt-cache/2-wjG_ETsLVrHU1L8Ejo8E7yRbQ/sTtvdyfU8Hlx-1oGm8v0ocjiYS8/relative/main.tf: no such file or directory
        Error Trace:    integration_test.go:1303
                        asm_amd64.s:574
                        panic.go:377
                        testing.go:553
                        testing.go:597
                        integration_test.go:1086
                        integration_test.go:711
```

brikis98:
That's weird. What version of Go? What OS are you on?

There are also some open Go bugs for this that sound similar: https://github.com/golang/go/issues/23444

lorengordon:
I had just installed/setup go before beginning this work, so latest at the time: `go version go1.10.3 linux/amd64`.

As for OS... well, that's where it gets complicated. It's Windows 10, but I use the Windows Subsystem for Linux (WSL), which appears pretty seamlessly to most everything as Ubuntu 18.04. I clone the repo on the Windows-side and do all my editing and git work there, but I mount the repo directory on the Linux side, and create a symlink into `$GOHOME`. Then I run the tests from WSL. So, yeah, weird is the name of the game... very well might be a local setup issue of some kind...

cli/cli_app.go
```diff
@@ -544,6 +546,7 @@ func runTerraformInit(terragruntOptions *options.TerragruntOptions, terragruntCo
 		} else {
 			// Terraform versions >= 0.10.0 specify the module source using the -from-module option
 			initOptions.AppendTerraformCliArgs("-from-module="+terraformSource.CanonicalSourceURL.String(), "-no-color")
+			initOptions.TerraformCommand = CMD_INIT_FROM_MODULE
```
brikis98: Might want to set this in the if-clause above too... Terraform before 0.10.0 did the behavior of `-from-module` without the `-from-module` param...

test/integration_test.go
```diff
@@ -30,51 +30,169 @@ import (
...
+		t.Errorf("Did not expect to get error: %s", err.Error())
+	}
+
+	// `init` hook should execute only once (2 occurrences due to the echo and its output)
+	assert.Equal(t, 2, strings.Count(output, "AFTER_INIT_ONLY_ONCE"), "Hooks on init command executed more than once")
+	// `init-from-module` hook should execute only once (2 occurrences due to the echo and its output)
+	assert.Equal(t, 2, strings.Count(output, "AFTER_INIT_FROM_MODULE_ONLY_ONCE"), "Hooks on init-from-module command executed more than once")
+}
+
+func TestTerragruntInitHookWithSourceWithBackend(t *testing.T) {
```
brikis98: Nice tests 👍

brikis98:
Oh, I gotcha. OK, then this error may be some symlink / Windows issue. Probably not worth digging into those two test failures, assuming all the other tests are passing.

test/integration_test.go
lorengordon: Thanks, the name-spacing in `go test` is nice... Just run `go test -v TestTerragruntInitHook` to run the four related tests. Good stuff.

cli/cli_app.go
lorengordon: Done

brikis98:
Merging now. I'll let the tests run and create a new release when they pass.

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.16.7

lorengordon:
Thanks @brikis98!

brikis98:
Thank you for the PR! 👍

Issue #355, opened 2017-11-09
ignore specified modules when running xxx-all commands (plan-all, apply-all, etc.)


ebarault:
Sometime it might be handy to prevent some modules from being handled by terragrunt **xxx-all** commands in a terragrunt project hierarchy while keeping the project hierarchy as-is for the sake of keeping the remote states coherent.

Consider this project structure:
```
root
├── app
│   ├── main.tf
│   └── terraform.tfvars
├── db
│   ├── main.tf
│   └── terraform.tfvars
└── vpc
    ├── main.tf
    └── terraform.tfvars
```

Say for example when i execute `terragrunt apply-all` at the root I want **app** and **db** to be handled, but not **vpc** which I manage separately. I think this is not currently supported.

Maybe a new directive inside the terragrunt config block could be used to indicate a given module should be ignored by `xxx-all` commands?

```
terragrunt = {
  dependencies {
    ignore = true
  }
}
```

brikis98:
I've run into similar circumstances before. Putting some in the `.tfvars` file is one option. Another is a CLI flag along the lines of `--terragrunt-ignore some/path/to/ignore`.

ebarault:
yes, i thought about this option too. that would fit too and be probably more flexible as it would enable multiple scenarios with the same terragrunt repo, which would be nice for CI/CD integration :
- job **A** loads the full repo and triggers infrastructure components deploy only
- job **B** loads the same repo and triggers service components deploy only

so +1 for `--terragrunt-ignore`

brikis98:
I'd definitely welcome a PR for that :)

ebarault:
I don't know go, but I can try if you considerer it feasible (entry-level PR) and provide a starting point

brikis98:
It's probably closer to intermediate rather than entry-level. But key points are:

* [Argument parsing](https://github.com/gruntwork-io/terragrunt/blob/master/cli/args.go#L18).
* [Where modules are found for the xxx-all command](https://github.com/gruntwork-io/terragrunt/blob/master/configstack/module.go#L38). 

You'd read store the `--terragrunt-ignore` arguments in `TerragruntOptions` and then in `ResolveTerraformModules`, skip over any modules that match the ignore args.

ebarault:
ok, thanks, I'll try to have a look in the coming days

tomdavidson:
What about globs for the workdir instead? For the OP it could be:

[...]

PR #563, opened 2018-09-07
implement exclude-dir cli flag

msvechla:
This implements https://github.com/gruntwork-io/terragrunt/issues/355 by adding a `-terragrunt-exclude-dir` CLI flag.

It supports unix-style globs and will exclude all modules within or under the specified directories during execution.

Let me know if additional things need to be adjusted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1801,6 +1801,7 @@ start with the prefix `--terragrunt-`. The currently available options are:
   specified via the `TERRAGRUNT_IAM_ROLE` environment variable. This is a convenient way to use Terragrunt and 
   Terraform with multiple AWS accounts.
 
+* `--terragrunt-exclude-dir`: Unix-style glob of directories to exclude when running `*-all` commands. Modules under these direcotires will be excluded during execution of the commands. Flag can be specified multiple times.
 
 ### Configuration
 
--- a/README.md
+++ b/README.md
@@ -1801,7 +1801,7 @@ start with the prefix `--terragrunt-`. The currently available options are:
   specified via the `TERRAGRUNT_IAM_ROLE` environment variable. This is a convenient way to use Terragrunt and 
   Terraform with multiple AWS accounts.
 
-* `--terragrunt-exclude-dir`: Unix-style glob of directories to exclude when running `*-all` commands. Modules under these direcotires will be excluded during execution of the commands. Flag can be specified multiple times.
+* `--terragrunt-exclude-dir`: Unix-style glob of directories to exclude when running `*-all` commands. Modules under these directories will be excluded during execution of the commands. If a relative path is specified, it should be relative from `--terragrunt-working-dir`. Flag can be specified multiple times.
 
 ### Configuration
 
--- a/cli/args.go
+++ b/cli/args.go
@@ -87,7 +87,7 @@ func parseTerragruntOptionsFromArgs(args []string, writer, errWriter io.Writer) 
 		return nil, err
 	}
 
-	excludeDirs, err := parseMultiStringArg(args, OPT_TERRAGRUNT_EXCLUDE_DIR, []string{os.Getenv("TERRAGRUNT_EXCLUDE_DIR")})
+	excludeDirs, err := parseMultiStringArg(args, OPT_TERRAGRUNT_EXCLUDE_DIR, []string{})
 	if err != nil {
 		return nil, err
 	}
@@ -236,6 +236,7 @@ func parseStringArg(args []string, argName string, defaultValue string) (string,
 // return a list of all values. If there are any present, but one of them has no value, return an error. If there aren't any present, return defaultValue.
 func parseMultiStringArg(args []string, argName string, defaultValue []string) ([]string, error) {
 	stringArgs := []string{}
+	fmt.Println(args)
 	for i, arg := range args {
 		if arg == fmt.Sprintf("--%s", argName) {
 			if (i + 1) < len(args) {
@@ -248,6 +249,9 @@ func parseMultiStringArg(args []string, argName string, defaultValue []string) (
 	if len(stringArgs) == 0 {
 		return defaultValue, nil
 	}
+
+	fmt.Printf("STRING ARGS: %s \n", stringArgs)
+
 	return stringArgs, nil
 }
 
--- a/cli/args.go
+++ b/cli/args.go
@@ -87,6 +87,11 @@ func parseTerragruntOptionsFromArgs(args []string, writer, errWriter io.Writer) 
 		return nil, err
 	}
 
+	excludeDir, err := parseStringArg(args, OPT_TERRAGRUNT_EXCLUDE_DIR, os.Getenv("TERRAGRUNT_EXCLUDE_DIR"))
+	if err != nil {
+		return nil, err
+	}
+
 	opts, err := options.NewTerragruntOptions(filepath.ToSlash(terragruntConfigPath))
 	if err != nil {
 		return nil, err
@@ -108,6 +113,7 @@ func parseTerragruntOptionsFromArgs(args []string, writer, errWriter io.Writer) 
 	opts.ErrWriter = errWriter
 	opts.Env = parseEnvironmentVariables(os.Environ())
 	opts.IamRole = iamRole
+	opts.ExcludeDir = excludeDir
 
 	return opts, nil
 }
--- a/cli/args.go
+++ b/cli/args.go
@@ -87,7 +87,7 @@ func parseTerragruntOptionsFromArgs(args []string, writer, errWriter io.Writer) 
 		return nil, err
 	}
 
-	excludeDir, err := parseStringArg(args, OPT_TERRAGRUNT_EXCLUDE_DIR, os.Getenv("TERRAGRUNT_EXCLUDE_DIR"))
+	excludeDir, err := parseMultiStringArg(args, OPT_TERRAGRUNT_EXCLUDE_DIR, []string{os.Getenv("TERRAGRUNT_EXCLUDE_DIR")})
 	if err != nil {
 		return nil, err
 	}
@@ -113,7 +113,7 @@ func parseTerragruntOptionsFromArgs(args []string, writer, errWriter io.Writer) 
 	opts.ErrWriter = errWriter
 	opts.Env = parseEnvironmentVariables(os.Environ())
 	opts.IamRole = iamRole
-	opts.ExcludeDir = excludeDir
+	opts.ExcludeDirs = ExcludeDirs
 
 	return opts, nil
 }
@@ -230,6 +230,29 @@ func parseStringArg(args []string, argName string, defaultValue string) (string,
 		}
 	}
 	return defaultValue, nil
+}
+
+// Find multiple string arguments of the same type (e.g. --foo "VALUE_A" --foo "VALUE_B") of the given name in the given list of arguments. If there are any present,
+// return a list of all values. If there are any present, but one of them has no value, return an error. If there aren't any present, return defaultValue.
+func parseMultiStringArg(args []string, argName string, defaultValue []string) ([]string, error) {
+
+	stringArgs := []string{}
+
+	for i, arg := range args {
+		if arg == fmt.Sprintf("--%s", argName) {
+			if (i + 1) < len(args) {
+				stringArgs = append(stringArgs, args[i+1])
+			} else {
+				return nil, errors.WithStackTrace(ArgMissingValue(argName))
+			}
+		}
+	}
+
+	if len(stringArgs) == 0 {
+		return defaultValue, nil
+	}
+
+	return stringArgs, nil
 }
 
 // Custom error types
--- a/cli/args.go
+++ b/cli/args.go
@@ -236,7 +236,7 @@ func parseStringArg(args []string, argName string, defaultValue string) (string,
 // return a list of all values. If there are any present, but one of them has no value, return an error. If there aren't any present, return defaultValue.
 func parseMultiStringArg(args []string, argName string, defaultValue []string) ([]string, error) {
 	stringArgs := []string{}
-	fmt.Println(args)
+
 	for i, arg := range args {
 		if arg == fmt.Sprintf("--%s", argName) {
 			if (i + 1) < len(args) {
@@ -249,8 +249,6 @@ func parseMultiStringArg(args []string, argName string, defaultValue []string) (
 	if len(stringArgs) == 0 {
 		return defaultValue, nil
 	}
-
-	fmt.Printf("STRING ARGS: %s \n", stringArgs)
 
 	return stringArgs, nil
 }
--- a/cli/args.go
+++ b/cli/args.go
@@ -87,7 +87,7 @@ func parseTerragruntOptionsFromArgs(args []string, writer, errWriter io.Writer) 
 		return nil, err
 	}
 
-	excludeDir, err := parseMultiStringArg(args, OPT_TERRAGRUNT_EXCLUDE_DIR, []string{os.Getenv("TERRAGRUNT_EXCLUDE_DIR")})
+	ExcludeDirs, err := parseStringArg(args, OPT_TERRAGRUNT_EXCLUDE_DIR, os.Getenv("TERRAGRUNT_EXCLUDE_DIR"))
 	if err != nil {
 		return nil, err
 	}
@@ -230,29 +230,6 @@ func parseStringArg(args []string, argName string, defaultValue string) (string,
 		}
 	}
 	return defaultValue, nil
-}
-
-// Find multiple string arguments of the same type (e.g. --foo "VALUE_A" --foo "VALUE_B") of the given name in the given list of arguments. If there are any present,
-// return a list of all values. If there are any present, but one of them has no value, return an error. If there aren't any present, return defaultValue.
-func parseMultiStringArg(args []string, argName string, defaultValue []string) ([]string, error) {
-
-	stringArgs := []string{}
-
-	for i, arg := range args {
-		if arg == fmt.Sprintf("--%s", argName) {
-			if (i + 1) < len(args) {
-				stringArgs = append(stringArgs, args[i+1])
-			} else {
-				return nil, errors.WithStackTrace(ArgMissingValue(argName))
-			}
-		}
-	}
-
-	if len(stringArgs) == 0 {
-		return defaultValue, nil
-	}
-
-	return stringArgs, nil
 }
 
 // Custom error types
--- a/cli/args.go
+++ b/cli/args.go
@@ -87,7 +87,7 @@ func parseTerragruntOptionsFromArgs(args []string, writer, errWriter io.Writer) 
 		return nil, err
 	}
 
-	ExcludeDirs, err := parseStringArg(args, OPT_TERRAGRUNT_EXCLUDE_DIR, os.Getenv("TERRAGRUNT_EXCLUDE_DIR"))
+	excludeDirs, err := parseMultiStringArg(args, OPT_TERRAGRUNT_EXCLUDE_DIR, []string{os.Getenv("TERRAGRUNT_EXCLUDE_DIR")})
 	if err != nil {
 		return nil, err
 	}
@@ -113,7 +113,7 @@ func parseTerragruntOptionsFromArgs(args []string, writer, errWriter io.Writer) 
 	opts.ErrWriter = errWriter
 	opts.Env = parseEnvironmentVariables(os.Environ())
 	opts.IamRole = iamRole
-	opts.ExcludeDirs = ExcludeDirs
+	opts.ExcludeDirs = excludeDirs
 
 	return opts, nil
 }
@@ -230,6 +230,25 @@ func parseStringArg(args []string, argName string, defaultValue string) (string,
 		}
 	}
 	return defaultValue, nil
+}
+
+// Find multiple string arguments of the same type (e.g. --foo "VALUE_A" --foo "VALUE_B") of the given name in the given list of arguments. If there are any present,
+// return a list of all values. If there are any present, but one of them has no value, return an error. If there aren't any present, return defaultValue.
+func parseMultiStringArg(args []string, argName string, defaultValue []string) ([]string, error) {
+	stringArgs := []string{}
+	for i, arg := range args {
+		if arg == fmt.Sprintf("--%s", argName) {
+			if (i + 1) < len(args) {
+				stringArgs = append(stringArgs, args[i+1])
+			} else {
+				return nil, errors.WithStackTrace(ArgMissingValue(argName))
+			}
+		}
+	}
+	if len(stringArgs) == 0 {
+		return defaultValue, nil
+	}
+	return stringArgs, nil
 }
 
 // Custom error types
--- a/cli/args_test.go
+++ b/cli/args_test.go
@@ -7,13 +7,14 @@ import (
 	"path/filepath"
 	"testing"
 
+	"strings"
+
 	"github.com/gruntwork-io/terragrunt/config"
 	"github.com/gruntwork-io/terragrunt/errors"
 	"github.com/gruntwork-io/terragrunt/options"
 	"github.com/gruntwork-io/terragrunt/util"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"strings"
 )
 
 func TestParseTerragruntOptionsFromArgs(t *testing.T) {
@@ -188,6 +189,27 @@ func TestFilterTerragruntArgs(t *testing.T) {
 
 	for _, testCase := range testCases {
 		actual := filterTerragruntArgs(testCase.args)
+		assert.Equal(t, testCase.expected, actual, "For args %v", testCase.args)
+	}
+}
+
+func TestParseMultiStringArg(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		args         []string
+		argName      string
+		defaultValue []string
+		expected     []string
+	}{
+		{[]string{"apply-all", "--foo", "bar"}, "foo", []string{"default_bar"}, []string{"bar"}},
+		{[]string{"apply-all", "--test", "bar"}, "foo", []string{"default_bar"}, []string{"default_bar"}},
+		{[]string{"plan-all", "--test", "--foo", "bar1", "--foo", "bar2"}, "foo", []string{"default_bar"}, []string{"bar1", "bar2"}},
+		{[]string{"plan-all", "--test", "value", "--foo", "bar1", "--foo"}, "foo", []string{"default_bar"}, nil},
+	}
+
+	for _, testCase := range testCases {
+		actual, _ := parseMultiStringArg(testCase.args, testCase.argName, testCase.defaultValue)
 		assert.Equal(t, testCase.expected, actual, "For args %v", testCase.args)
 	}
 }
--- a/cli/args_test.go
+++ b/cli/args_test.go
@@ -200,17 +200,24 @@ func TestParseMultiStringArg(t *testing.T) {
 		args         []string
 		argName      string
 		defaultValue []string
-		expected     []string
+		expectedVals []string
+		expectedErr  error
 	}{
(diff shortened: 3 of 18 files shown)


configstack/module.go
```diff
@@ -52,7 +53,68 @@ func ResolveTerraformModules(terragruntConfigPaths []string, terragruntOptions *
 		return []*TerraformModule{}, err
 	}
 
-	return crosslinkDependencies(mergeMaps(modules, externalDependencies), canonicalTerragruntConfigPaths)
+	crossLinkedModules, err := crosslinkDependencies(mergeMaps(modules, externalDependencies), canonicalTerragruntConfigPaths)
+	if err != nil {
+		return []*TerraformModule{}, err
+	}
+
+	finalModules := removeExcludedDirs(crossLinkedModules, terragruntOptions)
```
brikis98: Should modules be excluded *before* cross-linking?

cli/args.go
```diff
@@ -87,6 +87,11 @@ func parseTerragruntOptionsFromArgs(args []string, writer, errWriter io.Writer)
 		return nil, err
 	}
 
+	excludeDir, err := parseStringArg(args, OPT_TERRAGRUNT_EXCLUDE_DIR, os.Getenv("TERRAGRUNT_EXCLUDE_DIR"))
```
brikis98: Should you be able to specify `--terragrunt-exclude-dir` multiple times to capture multiple patterns to exclude?

configstack/module.go
```diff
@@ -52,7 +53,68 @@ func ResolveTerraformModules(terragruntConfigPaths []string, terragruntOptions *
...
+// removeExcludedDirs iterates over a module slice and removes all entries, which should be ignored via the terragrunt-exclude-dir CLI flag.
+// Modules are also removed if one of its dependencies are to be excluded. Returns the cleaned-up slice of modules.
+func removeExcludedDirs(modules []*TerraformModule, terragruntOptions *options.TerragruntOptions) []*TerraformModule {
+
+	// If no excludeDir is specified return the modules list instantly
+	if terragruntOptions.ExcludeDir == "" {
+		return modules
+	}
+
+	excludeGlobMatches := []string{}
```
brikis98: No need to declare ahead of time, the next line will declare this variable for you.

configstack/module.go
```diff
@@ -52,7 +53,68 @@ func ResolveTerraformModules(terragruntConfigPaths []string, terragruntOptions *
...
+
+	// If possible, expand the glob to get all excluded filepaths
+	excludeGlobMatches, err := zglob.Glob(terragruntOptions.ExcludeDir)
+
+	if err != nil {
+		// Glob can not be expanded, using literal value
```
brikis98: Could the error mean something else? E.g., invalid glob?

configstack/module.go
```diff
@@ -52,7 +53,68 @@ func ResolveTerraformModules(terragruntConfigPaths []string, terragruntOptions *
...
+		excludeGlobMatches = []string{terragruntOptions.ExcludeDir}
+	}
+
+	// If the expanded glob paths are relative, generate absolute paths
+	absoluteExludeDir := []string{}
+	for _, m := range excludeGlobMatches {
+		if strings.HasPrefix(m, "/") {
```
brikis98: Hm, this won't work on Windows. Use `filepath.IsAbs`.

configstack/module.go
```diff
@@ -52,7 +53,68 @@ func ResolveTerraformModules(terragruntConfigPaths []string, terragruntOptions *
...
+			absoluteExludeDir = append(absoluteExludeDir, m)
+		} else {
+			absoluteExludeDir = append(absoluteExludeDir, fmt.Sprintf("%s/%s", terragruntOptions.WorkingDir, m))
```
brikis98: Use `filepath.Join` so this works across operating systems.

configstack/module.go
```diff
@@ -52,7 +53,68 @@ func ResolveTerraformModules(terragruntConfigPaths []string, terragruntOptions *
...
+	excludeGlobMatches, err := zglob.Glob(terragruntOptions.ExcludeDir)
+
+	if err != nil {
+		// Glob can not be expanded, using literal value
+		excludeGlobMatches = []string{terragruntOptions.ExcludeDir}
+	}
+
+	// If the expanded glob paths are relative, generate absolute paths
+	absoluteExludeDir := []string{}
+	for _, m := range excludeGlobMatches {
```
brikis98: Please use longer, more readable names like `module` or even `ecludedGlobMatch` instead of `m`

configstack/module.go
brikis98: In fact, replace the body of this entire loop with a call to Terragrunts `util.CanonicalPath`, as, based on my reading of your code below, that's what you actually need.

configstack/module.go
```diff
@@ -52,7 +53,68 @@ func ResolveTerraformModules(terragruntConfigPaths []string, terragruntOptions *
...
+	// Create a slice of finalModules to store all modules, which should not be excluded
+	finalModules := modules[:0]
+
+	for _, e := range absoluteExludeDir {
+		for _, m := range modules {
+
+			// Check if none of the current modules dependencies are located under the excluded directory
+			dependencyExcluded := false
+			for _, d := range m.Dependencies {
+				if strings.Contains(d.Path, e) {
```
brikis98: To do string comparisons on paths, absolute paths won't do it. E.g., `/foo/bar` and `/foo/bar/../bar` are both absolute paths and point to the same folder, but are not equivalent as strings. So you'll need to use `util.CanonicalPath` above to make all the paths "canonical", which means all `..` and `.` are resolved.

configstack/module.go
```diff
@@ -52,7 +53,68 @@ func ResolveTerraformModules(terragruntConfigPaths []string, terragruntOptions *
...
+	for _, m := range excludeGlobMatches {
+		if strings.HasPrefix(m, "/") {
+			absoluteExludeDir = append(absoluteExludeDir, m)
+		} else {
+			absoluteExludeDir = append(absoluteExludeDir, fmt.Sprintf("%s/%s", terragruntOptions.WorkingDir, m))
+		}
+	}
+
+	// Create a slice of finalModules to store all modules, which should not be excluded
+	finalModules := modules[:0]
```
brikis98: Why `modules[:0]`?

configstack/module.go
```diff
@@ -52,7 +53,68 @@ func ResolveTerraformModules(terragruntConfigPaths []string, terragruntOptions *
...
+
+	for _, e := range absoluteExludeDir {
+		for _, m := range modules {
+
+			// Check if none of the current modules dependencies are located under the excluded directory
+			dependencyExcluded := false
+			for _, d := range m.Dependencies {
```
brikis98: This sort of code is easier to follow if you extract the checks into their own `shouldExcludeDependency` method that returns a `bool`. That way, instead of `continue` and mutable variables and all that, the method can just `return true` or `return false` immediately. You then call that method within this for-loop and only if it returns `false`, do you add the module to the `finalModules` list.

configstack/module_test.go
```diff
@@ -92,6 +92,80 @@ func TestResolveTerraformModulesTwoModulesWithDependencies(t *testing.T) {
 	assertModuleListsEqual(t, expected, actualModules)
 }
 
+func TestResolveTerraformModulesTwoModulesWithDependenciesExcludedDirsWithDependency(t *testing.T) {
```
brikis98: Thanks for adding tests! 👍

configstack/module.go
```diff
@@ -52,7 +53,68 @@ func ResolveTerraformModules(terragruntConfigPaths []string, terragruntOptions *
...
+		}
+	}
+
+	// Create a slice of finalModules to store all modules, which should not be excluded
+	finalModules := modules[:0]
+
+	for _, e := range absoluteExludeDir {
+		for _, m := range modules {
+
+			// Check if none of the current modules dependencies are located under the excluded directory
```
brikis98: Hm, so should a module `A` be excluded just because it depends on a module `B` that is excluded? I understand that if `B` has never been applied, then `A` might fail, but what if `B` is something like a VPC that has been applied, long ago, and you just don't want to (uselessly) keep applying it again and again?

options/options.go
```diff
@@ -80,6 +80,9 @@ type TerragruntOptions struct {
 	// exposed here primarily so we can set it to a low value at test time.
 	MaxFoldersToCheck int
 
+	// Unix-style glob of directories to exclude when running *-all commands
+	ExcludeDir string
```
brikis98: Having looked through the code, I think `ExcludeDirs []string` would be a better fit.

```
terragrunt apply-all --terragrunt-exclude-dir foo --terragrunt-exclude-dir bar/*
```

configstack/module.go
```diff
@@ -52,7 +53,68 @@ func ResolveTerraformModules(terragruntConfigPaths []string, terragruntOptions *
...
+	}
+
+	finalModules := removeExcludedDirs(crossLinkedModules, terragruntOptions)
+
+	return finalModules, nil
+}
+
+// removeExcludedDirs iterates over a module slice and removes all entries, which should be ignored via the terragrunt-exclude-dir CLI flag.
+// Modules are also removed if one of its dependencies are to be excluded. Returns the cleaned-up slice of modules.
+func removeExcludedDirs(modules []*TerraformModule, terragruntOptions *options.TerragruntOptions) []*TerraformModule {
```
brikis98: Consider adding unit tests just for this method. You could pass in your own mock set of modules with various paths and see what it returns.

msvechla:
Thanks for your very valuable feedback. This is my first bigger MR using golang, so there is definitely a lot to learn. I'll go over it and let you know once I'm done.

configstack/module.go
msvechla: I excluded them *after* cross-linking, because the depency information is needed. This information is used to exclude modules with dependencies which are in the to-be-excluded list. (See other conversation)

configstack/module.go
msvechla: I already added to unit tests for `ResolveTerraformModules` which calls `removeExcludedDirs`.
The mock set was especially trimmed for excluded dirs. However I can add even more tests if you would like.

configstack/module.go
msvechla: Thanks for the feedback, I agree this makes the code much more readable. Please take a look at my adjustments.

configstack/module.go
msvechla: However even if we remove them before cross-linking, I think `cross-linking` will throw an error, because it can no longer find some modules which are specified as dependencies.

configstack/module.go
msvechla: Makes completely sense, please check the new code.

configstack/module.go
msvechla: Used longer, more readable names throughout the code.

configstack/module.go
msvechla: This is a very good point. I think it is safe to assume, that if the user excplicitly excludes a directory, it can be assumed that all other modules will still be applied, even if there dependencies are not fulfilled. I can adjust the code to reflect this behavior, what do you think?

options/options.go
msvechla: Renamed the variable.

README.md
```diff
@@ -1801,6 +1801,7 @@ start with the prefix `--terragrunt-`. The currently available options are:
   specified via the `TERRAGRUNT_IAM_ROLE` environment variable. This is a convenient way to use Terragrunt and 
   Terraform with multiple AWS accounts.
 
+* `--terragrunt-exclude-dir`: Unix-style glob of directories to exclude when running `*-all` commands. Modules under these direcotires will be excluded during execution of the commands. Flag can be specified multiple times.
```
brikis98: s/direcotires/directories/

cli/args.go
```diff
@@ -87,6 +87,11 @@ func parseTerragruntOptionsFromArgs(args []string, writer, errWriter io.Writer)
 		return nil, err
 	}
 
+	ExcludeDirs, err := parseStringArg(args, OPT_TERRAGRUNT_EXCLUDE_DIR, os.Getenv("TERRAGRUNT_EXCLUDE_DIR"))
```
brikis98: Doesn't `parseStringArg` parse a string rather than a slice?

configstack/module.go
```diff
@@ -52,7 +54,82 @@ func ResolveTerraformModules(terragruntConfigPaths []string, terragruntOptions *
...
+	}
+
+	excludeGlobMatches := []string{}
+
+	// If possible, expand the glob to get all excluded filepaths
+	for _, dir := range terragruntOptions.ExcludeDirs {
+		matches, err := zglob.Glob(dir)
+
+		if err != nil {
+			// Glob can not be expanded, using literal value
```
brikis98: I'm still worried `zglob.Glob` may fail for legit reasons and ignoring it here and just using it as a literal may lead to unexpected behavior.

configstack/module.go
```diff
@@ -52,7 +54,82 @@ func ResolveTerraformModules(terragruntConfigPaths []string, terragruntOptions *
...
+		}
+		canonicalExcludeDir = append(canonicalExcludeDir, canonicalPath)
+	}
+
+	// Create a slice of finalModules to store all modules, which should not be excluded
+	finalModules := []*TerraformModule{}
+
+	for _, excludeDir := range canonicalExcludeDir {
+		for _, module := range modules {
+			if !shouldExcludeModuleBecauseOfPath(module, excludeDir) && !shouldExcludeModuleBecauseOfDependency(module, excludeDir) {
```
brikis98: Nice, this is much more readable now 👍

configstack/module.go
```diff
@@ -52,7 +54,82 @@ func ResolveTerraformModules(terragruntConfigPaths []string, terragruntOptions *
...
+			}
+		}
+	}
+
+	return finalModules, nil
+}
+
+// Returns true if a module is located under one of the excluded directories
+func shouldExcludeModuleBecauseOfPath(module *TerraformModule, excludeDir string) bool {
+	if strings.Contains(module.Path, excludeDir) {
```
brikis98: I can't remember, will `module.Path` already be a canonical path?

configstack/module.go
brikis98: Yea, I think we should only exclude the directories the user explicitly tells us to and not their dependencies.

cli/args.go
msvechla: Correct. Sorry my local git was messed up and I did a partial commit, should be fixed now.

msvechla:
I tried adding some integration tests, but they are not working yet as expected. Will take another look at it tomorrow.

brikis98:
Sounds good. Let me know if you need any help sorting it out. And thank you for the work you're putting into this PR!

configstack/module.go
msvechla: @brikis98 I adapted this so we are no longer excluding the modules which depend on excluded modules. However I am now running in a `DependencyNotFoundWhileCrossLinking` error. 

I tracked it down to `crossLinkDependencies` inside `running_module.go`.

This error appears because it tries to cross-link dependencies during the run, and the modules dependencies are no longer in the list of modules.

Is the best solution here, to remove the affected dependencies in `removeExcludedDirs`?

msvechla: FYI, I did not push that code yet, as it contains a lot of debug output for tests.

brikis98: Hm, is it possible to remove excluded modules after the call to `crossLinkDependencies` in `running_module.go`?

msvechla: you mean instead of removing them in `module.go`? 
I assume this would be in `toRunningModules()` then? However I am missing the terragrunt options there and I am not yet completely sure if this is the correct place to do it.

Also, why are dependencies linked here again, when this is already done in `ResolveTerraformModules`?

I would need some guidance here on where best to remove the excluded modules. Would be great if you could give some more insights.

msvechla: I fixed this now by introducing another field called `flagExcluded` to the module struct.
This flag is set in `flagExcludedDirs` in `module.go` as all necessary information is available here.

The modules and the affected dependencies are then later actually removed in `crossLinkDependencies` in `running_module.go`.

This was the cleanest way of making this work, as I was missing essential information inside `running_module.go`.

msvechla:
Did some refactoring (see discussions) and added integration tests. Would be awesome if you can take another look!

brikis98:
Thanks! I'm a bit bogged down today, but I'll take a look tomorrow.

configstack/module.go
```diff
@@ -52,7 +56,86 @@ func ResolveTerraformModules(terragruntConfigPaths []string, terragruntOptions *
...
+	// If no ExcludeDirs is specified return the modules list instantly
+	if len(terragruntOptions.ExcludeDirs) == 0 {
+		return modules, nil
+	}
+
+	excludeGlobMatches := []string{}
+
+	// If possible, expand the glob to get all excluded filepaths
+	// Change to the actual working dir to make glob expansion work correctly
+	os.Chdir(terragruntOptions.WorkingDir)
```
brikis98: Hm, I'm a bit worried about changing dirs... This can have unintended and confusing side effects, especially when things are running concurrently (e.g., multiple unit tests). 

Perhaps instead you should `filepath.Join` each excluded dir with `terragruntOptions.WorkingDir`if that excluded dir isn't an absolute path?

configstack/running_module.go
```diff
@@ -104,9 +105,27 @@ func crossLinkDependencies(modules map[string]*runningModule, dependencyOrder De
 		}
 	}
 
+	removeFlagExcluded(modules)
```
brikis98: This should probably not be called in `crossLinkDependencies` but separately to avoid confusion.

configstack/running_module.go
```diff
@@ -104,9 +105,27 @@ func crossLinkDependencies(modules map[string]*runningModule, dependencyOrder De
...
+
 	return modules, nil
 }
 
+// Removes all the modules from the map, which have been flagged as excluded in flagExcludedDirs()
+func removeFlagExcluded(modules map[string]*runningModule) {
```
brikis98: Instead mutating the existing map, I'd just create a new one that only contains the non-excluded modules.

configstack/module.go
msvechla: I had the same idea, but I thought there were some reasons why it would not work. Let me think this through again and give it another try

msvechla: yep this worked, thanks for challenging this again!

configstack/module.go
```diff
@@ -52,7 +56,86 @@ func ResolveTerraformModules(terragruntConfigPaths []string, terragruntOptions *
...
+
+	// Save original working directory
+	workingDir, err := os.Getwd()
+	if err != nil {
+		return nil, err
+	}
+
+	// If no ExcludeDirs is specified return the modules list instantly
+	if len(terragruntOptions.ExcludeDirs) == 0 {
+		return modules, nil
```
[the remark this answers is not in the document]
msvechla: fixed

cli/args_test.go
```diff
@@ -192,6 +193,27 @@ func TestFilterTerragruntArgs(t *testing.T) {
...
+	testCases := []struct {
+		args         []string
+		argName      string
+		defaultValue []string
+		expected     []string
+	}{
+		{[]string{"apply-all", "--foo", "bar"}, "foo", []string{"default_bar"}, []string{"bar"}},
+		{[]string{"apply-all", "--test", "bar"}, "foo", []string{"default_bar"}, []string{"default_bar"}},
+		{[]string{"plan-all", "--test", "--foo", "bar1", "--foo", "bar2"}, "foo", []string{"default_bar"}, []string{"bar1", "bar2"}},
+		{[]string{"plan-all", "--test", "value", "--foo", "bar1", "--foo"}, "foo", []string{"default_bar"}, nil},
```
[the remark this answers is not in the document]
msvechla: I was only checking the return value for nil and did not check the actual error. Fixed this, now also checking the error value.

configstack/running_module.go
msvechla: moved this to `toRunningModules()` which should separate the actions more clearly.

configstack/running_module.go
msvechla: Adjusted this accordingly, however the original solution was probably more efficient and had better readability.

cli/args.go
```diff
@@ -226,6 +232,29 @@ func parseStringArg(args []string, argName string, defaultValue string) (string,
...
+			} else {
+				return nil, errors.WithStackTrace(ArgMissingValue(argName))
+			}
+		}
+	}
+	if len(stringArgs) == 0 {
+		return defaultValue, nil
+	}
+
+	fmt.Printf("STRING ARGS: %s \n", stringArgs)
```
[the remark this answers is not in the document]
msvechla: Removed leftover debug-output

cli/args.go
```diff
@@ -226,6 +232,29 @@ func parseStringArg(args []string, argName string, defaultValue string) (string,
 	return defaultValue, nil
 }
 
+// Find multiple string arguments of the same type (e.g. --foo "VALUE_A" --foo "VALUE_B") of the given name in the given list of arguments. If there are any present,
+// return a list of all values. If there are any present, but one of them has no value, return an error. If there aren't any present, return defaultValue.
+func parseMultiStringArg(args []string, argName string, defaultValue []string) ([]string, error) {
+	stringArgs := []string{}
+	fmt.Println(args)
```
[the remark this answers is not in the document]
msvechla: Removed leftover debug-output

configstack/module.go
```diff
@@ -52,7 +55,89 @@ func ResolveTerraformModules(terragruntConfigPaths []string, terragruntOptions *
...
+		}
+		canonicalExcludeDirs = append(canonicalExcludeDirs, canonicalPath)
+	}
+
+	for _, module := range modules {
+		if shouldExcludeModuleBecauseOfPath(module, canonicalExcludeDirs) {
+			// Mark module itself as excluded
+			module.FlagExcluded = true
+		} else {
+			// Mark all affected dependencies as excluded
```
brikis98: Should we loop over all dependencies whether or not the module itself is excluded?

configstack/running_module.go
```diff
@@ -79,7 +80,13 @@ func toRunningModules(modules []*TerraformModule, dependencyOrder DependencyOrde
...
 	}
 
-	return crossLinkDependencies(runningModules, dependencyOrder)
+	crossLinkedModules, err := crossLinkDependencies(runningModules, dependencyOrder)
+	if err != nil {
+		return crossLinkedModules, err
+	}
+
+	finalModules := removeFlagExcluded(crossLinkedModules)
+	return finalModules, nil
```
brikis98: Slightly shorter: `return removeFlagExcluded(...), nil`

configstack/module.go
msvechla: To be consistent we should, yes. Good point.

msvechla:
I just pushed another update, found an error while running tests.

Now I am stuck at an error with test: `TestToRunningModulesTwoModulesWithDependenciesReverseOrder()`

Is this test case correct?
`runningModuleB` has no dependencies, while `moduleB` has a dependency.
Is this correct, or should the dependencies of `runningModuleB` match for the test case?
```
moduleB := &TerraformModule{
		Path:              "b",
		Dependencies:      []*TerraformModule{moduleA},
		Config:            config.TerragruntConfig{},
		TerragruntOptions: mockOptions,
	}

	runningModuleB := &runningModule{
		Module:         moduleB,
		Status:         Waiting,
		Err:            nil,
		Dependencies:   map[string]*runningModule{},
		NotifyWhenDone: []*runningModule{runningModuleA},
	}
```

msvechla:
Thanks for giving a little bit more insight here, totally makes sense. With this information I found the issue and fixed it (see latest commit). I also successfully ran all tests now, including the ones requiring AWS credentials:

```
FAIL    github.com/msvechla/terragrunt/cli      9.053s
ok      github.com/msvechla/terragrunt/config   0.031s
ok      github.com/msvechla/terragrunt/configstack      0.027s
ok      github.com/msvechla/terragrunt/dynamodb 33.841s
?       github.com/msvechla/terragrunt/errors   [no test files]
?       github.com/msvechla/terragrunt/options  [no test files]
ok      github.com/msvechla/terragrunt/remote   0.025s
ok      github.com/msvechla/terragrunt/shell    6.551s
ok      github.com/msvechla/terragrunt/test     428.479s
?       github.com/msvechla/terragrunt/test/helpers     [no test files]
ok      github.com/msvechla/terragrunt/util     0.033s
``` 

The only issue I get is with `TestCheckTerraformVersionMeetsConstraintGreaterDev`:

```
--- FAIL: TestCheckTerraformVersionMeetsConstraintGreaterDev (0.00s)
    version_check_test.go:78:
                Error Trace:    version_check_test.go:78
                                                        version_check_test.go:17
                Error:          Expected nil, but got: &errors.Error{Err:cli.InvalidTerraformVersion{CurrentVersion:(*version.Version)(0xc00042a460), VersionConstraints:version.Constraints{(*version.Constraint)(0xc0003f0da0)}}, stack:[]uintptr{0x159f7b9, 0x15a71cd, 0x15a6bc8, 0x10f8f9f, 0x105dc11}, frames:[]errors.StackFrame(nil), prefix:""}
                Test:           TestCheckTerraformVersionMeetsConstraintGreaterDev
                Messages:       Expected Terraform version v0.9.4-dev to meet constraint >= v0.9.3, but got error: The currently installed version of Terraform (0.9.4-dev) is not compatible with the version Terragrunt requires (>= v0.9.3).
``` 

Does this mean I have to bump the terragrunt version as-well? Could you guide me on how to resolve this? Thanks again for your help.

brikis98:
> Expected Terraform version v0.9.4-dev to meet constraint >= v0.9.3, but got error: The currently installed version of Terraform (0.9.4-dev) is not compatible with the version Terragrunt requires (>= v0.9.3).

Do you have a custom (i.e., locally-built) version of Terraform installed (0.9.4-dev)? If so, it looks like the version check we do doesn't take `-dev` into account. If you try it with an official release of Terraform in your `PATH`, it should work.

msvechla:
Actually this has nothing to do with the installed version. This test is simply checking hardcoded versions.

This is the entire test:

``` 
func TestCheckTerraformVersionMeetsConstraintGreaterDev(t *testing.T) {
	t.Parallel()
	testCheckTerraformVersionMeetsConstraint(t, "v0.9.4-dev", ">= v0.9.3", true)
}
``` 

Drilling a little bit down the function it actually mostly calls the https://github.com/hashicorp/go-version module. So it seems like an external module is tested here. Does this test work for you on master? Maybe something changed on the module end which breaks the check.

brikis98:
> So it seems like an external module is tested here. Does this test work for you on master?

Yes, the tests passed on master just 14 hours ago. Is it possible you have a different version of `go-version` installed and something changed upstream?

msvechla:
Sorry for the confusion, I cleaned-up my local environment and ran `dep ensure`. Now all tests finished successfully.

```
(⎈ |:) ~/go/src/github.com/gruntwork-io/terragrunt   pr/563  go test -short ./...
?       github.com/gruntwork-io/terragrunt      [no test files]
?       github.com/gruntwork-io/terragrunt/aws_helper   [no test files]
ok      github.com/gruntwork-io/terragrunt/cli  10.085s
ok      github.com/gruntwork-io/terragrunt/config       0.022s
ok      github.com/gruntwork-io/terragrunt/configstack  0.026s
ok      github.com/gruntwork-io/terragrunt/dynamodb     43.642s
?       github.com/gruntwork-io/terragrunt/errors       [no test files]
?       github.com/gruntwork-io/terragrunt/options      [no test files]
ok      github.com/gruntwork-io/terragrunt/remote       0.019s
ok      github.com/gruntwork-io/terragrunt/shell        6.541s
ok      github.com/gruntwork-io/terragrunt/test 587.301s
?       github.com/gruntwork-io/terragrunt/test/helpers [no test files]
ok      github.com/gruntwork-io/terragrunt/util 0.032s
```

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.16.11

msvechla:
Thanks so much for your help on this, it was an awesome experience 👍

PR #566, opened 2018-09-13
Support force_path_style in remote state configuration

fosrias:
@brikis98 FOR DISCUSSION ONLY

This fixes the issue with `force_path_style`. The root issue is that the two S3 Client methods `HeadBucket` and `GetBucketVersioning` always use the bucket as the first subdomain. 

Thus, terragrunt will never work as currently configured with an S3 Compatible object store using path style buckets.

Have not added tests. Will do if this makes sense to you.


fosrias:
Closing this PR as not correct solution. Will open correct implementation shortly.

Issue #565, opened 2018-09-13
Terragrunt does not support current S3 remote state configuration options


fosrias:
Here are the current terraform options: https://github.com/hashicorp/terraform/blob/v0.11.8/backend/remote-state/s3/backend.go

Here are the supported options in Terragrunt: https://github.com/gruntwork-io/terragrunt/blob/master/remote/remote_state_s3.go#L33-L42

Terragrunt exclude lots of options including the new (and very useful one) `force_path_style` that I personally need.

@brikis98 Am I missing something? If not, are you open to a PR to fix this?

brikis98:
Are you hitting an actual bug or limitation because of this?

IIRC, the arguments we parse explicitly are for internal Terragrunt usage. When calling `terraform init` to configure a backend, we [parse config as a plain old map](https://github.com/gruntwork-io/terragrunt/blob/master/remote/remote_state.go#L13) and pass _all_ the config args to Terraform. So, the design should accommodate new args being added without Terragrunt having to know about them, though of course it's possible there's a bug in the implementation!

fosrias:
Well, if that is how it is supposed to work, then it appears there is a bug. I am trying to use this with an s3 compatible object store that does not support the bucket being the first subdomain, but rather a path for the bucket.

The following is my config:

```
terragrunt = {
  remote_state {
    backend = "s3"
    config {
      bucket   = "terraform-state"
      key      = "mwf/${path_relative_to_include()}/terraform.tfstate"
      region   = "us-phoenix-1"
      encrypt  = true
      endpoint = "https://something.example.com"
      
      shared_credentials_file     = "..."
      skip_region_validation      = true
      skip_credentials_validation = true
      skip_requesting_account_id  = true
      skip_get_ec2_platforms      = true
      skip_metadata_api_check     = true
      force_path_style            = true
    }
  }
```
If I configure this in a Terraform file, it all works fine for `terraform init`. However, for `terragrunt init` I first noticed the `shared_credentials_file` was ignored. So, I just used the aws default location even though this is not for AWS. Once I fixed that, I started trying to make requests and got this output:
```
$  terragrunt init
terragrunt init      
                                                                                                                                                   
[terragrunt] [/.../identity] 2018/09/12 11:51:07 Running command: terraform --version
[terragrunt] 2018/09/12 11:51:07 Reading Terragrunt config file at /.../identity/terraform.tfvars
[...]

PR #567, opened 2018-09-14
configure s3 client session to support force_path_style

fosrias:
@brikis98 This correctly implements `force_path_style`. It maintains all the current terragrunt functionality with the following changes:
1. It graciously fails if a third party s3 API does not support versioning
2. It allows other boolean configurations without incorrectly triggering an init everytime when comparing with string based existing configuration.

Relates to https://github.com/gruntwork-io/terragrunt/issues/565

This supports the following config:

```
terragrunt = {
  remote_state {
    backend = "s3"
    config {
      bucket   = "terraform-state2"
      key      = "${path_relative_to_include()}/terraform.tfstate"
      region   = "us-west-2"
      encrypt  = true
      endpoint = "https://s3-us-west-2.amazonaws.com"

      skip_region_validation      = true
      skip_credentials_validation = true
      skip_requesting_account_id  = true
      skip_get_ec2_platforms      = true
      skip_metadata_api_check     = true
      force_path_style            = true
    }
  }
```

I have not written tests, but will if you are comfortable with this. I have manually tested this on our third party object store and on AWS (with both regular and path style configuration).

### DIFF
--- a/README.md
+++ b/README.md
@@ -509,8 +509,29 @@ terragrunt = {
 The `remote_state` block supports all the same [backend types](https://www.terraform.io/docs/backends/types/index.html)
 as Terraform. The next time you run `terragrunt`, it will automatically configure all the settings in the
 `remote_state.config` block, if they aren't configured already, by calling [terraform
-init](https://www.terraform.io/docs/commands/init.html). The config options `s3_bucket_tags` and 
-`dynamodb_table_tags` are only valid for backend `s3`. They are used by terragrunt and are **not** passed on to 
+init](https://www.terraform.io/docs/commands/init.html). 
+
+For backend `s3`, the following config options can be used for S3-compatible object stores, as necessary:
+
+```hcl
+remote_state {
+  # ...
+
+  skip_bucket_versioning = true # use only if the object store does not support versioning
+
+  skip_region_validation      = true
+  skip_credentials_validation = true
+  skip_requesting_account_id  = true
+  skip_get_ec2_platforms      = true
+  skip_metadata_api_check     = true
+  force_path_style            = true
+}
+```
+
+If you experience an error for any of these configurations, confirm you are using Terraform v0.11.2 or greater.
+
+Further, the config options `s3_bucket_tags`, 
+`dynamodb_table_tags`, and `skip_bucket_versioning` are only valid for backend `s3`. They are used by terragrunt and are **not** passed on to 
 terraform. See section [Create remote state and locking resources automatically](#create-remote-state-and-locking-resources-automatically).
 
 In each of the **child** `terraform.tfvars` files, such as `mysql/terraform.tfvars`, you can tell Terragrunt to
--- a/remote/remote_state_s3.go
+++ b/remote/remote_state_s3.go
@@ -25,8 +25,9 @@ import (
 type ExtendedRemoteStateConfigS3 struct {
 	remoteStateConfigS3 RemoteStateConfigS3
 
-	S3BucketTags    []map[string]string `mapstructure:"s3_bucket_tags"`
-	DynamotableTags []map[string]string `mapstructure:"dynamodb_table_tags"`
+	S3BucketTags         []map[string]string `mapstructure:"s3_bucket_tags"`
+	DynamotableTags      []map[string]string `mapstructure:"dynamodb_table_tags"`
+	SkipBucketVersioning bool                `mapstructure:"skip_bucket_versioning"`
 }
 
 // A representation of the configuration options available for S3 remote state
@@ -135,9 +136,10 @@ func configValuesEqual(config map[string]interface{}, existingBackend *Terraform
 		}
 	}
 
-	// Delete S3 and DynamoDB tags, as these are only stored in Terragrunt config and not in Terraform's backend
+	// Delete S3 and DynamoDB and Bucket Versioning tags, as these are only stored in Terragrunt config and not in Terraform's backend
 	delete(config, "s3_bucket_tags")
 	delete(config, "dynamodb_table_tags")
+	delete(config, "skip_bucket_versioning")
 
 	if !reflect.DeepEqual(existingBackend.Config, config) {
 		terragruntOptions.Logger.Printf("Backend config has changed from %s to %s", existingBackend.Config, config)
@@ -186,7 +188,7 @@ func (s3Initializer S3Initializer) GetTerraformInitArgs(config map[string]interf
 
 	for key, val := range config {
 
-		if key == "s3_bucket_tags" || key == "dynamodb_table_tags" {
+		if key == "s3_bucket_tags" || key == "dynamodb_table_tags" || key == "skip_bucket_versioning" {
 			continue
 		}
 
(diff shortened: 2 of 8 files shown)


remote/remote_state_s3.go
```diff
@@ -123,6 +124,17 @@ func configValuesEqual(config map[string]interface{}, existingBackend *Terraform
...
 			terragruntOptions.Logger.Printf("Remote state configuration encrypt contains invalid value %v, should be boolean.", existingBackend.Config["encrypt"])
 		}
+
+		// If other keys are bools, DeepEqual also will consider the maps to be different.
+		for key, value := range existingBackend.Config {
+			if _, isBool := value.(bool); isBool {
+				continue // We know this is already converted to a bool, e.g. encrypt
+			}
+
+			if convertedValue, err := strconv.ParseBool(value.(string)); err == nil {
```
brikis98: Will this convert every string "true" and "false" to a boolean true and false?

remote/remote_state_s3.go
```diff
@@ -390,6 +402,12 @@ func EnableVersioningForS3Bucket(s3Client *s3.S3, config *RemoteStateConfigS3, t
 		VersioningConfiguration: &s3.VersioningConfiguration{Status: aws.String(s3.BucketVersioningStatusEnabled)},
 	}
 	_, err := s3Client.PutBucketVersioning(&input)
+	if config.S3ForcePathStyle && err != nil {
+		terragruntOptions.Logger.Printf("Versioning not supported for bucket %s", config.Bucket)
```
brikis98: How do we know that `err` implies versioning is not supported? Couldn't it be some other type of error too?

remote/remote_state_s3.go
fosrias: Yes.

remote/remote_state_s3.go
fosrias: > How do we know that err implies versioning is not supported? Couldn't it be some other type of error too?

Well, that has been nagging me since the moment I coded it ;). I think the right way is to add a terragrunt internal config option`skip_bucket_versioning` configuration. Then if it fails, user can configure it. Shall I do it that way?

fosrias:
Another issue is that Terraform has schema checking on `backend-config`, for example using v0.11.1, I got:

```
$ terragrunt init
[terragrunt] [/Users/mwfoster/projects/oke-terraform/identity] 2018/09/14 08:19:48 Running command: terraform --version
[terragrunt] 2018/09/14 08:19:49 Reading Terragrunt config file at /Users/mwfoster/projects/oke-terraform/identity/terraform.tfvars
[terragrunt] 2018/09/14 08:19:49 Initializing remote state for the s3 backend
[terragrunt] 2018/09/14 08:19:49 WARNING: Versioning is not enabled for the remote state S3 bucket terraform-state2. We recommend enabling versioning so that you can roll back to previous versions of your Terraform state in case of error.
[terragrunt] 2018/09/14 08:19:49 Running command: terraform init -backend-config=skip_region_validation=true -backend-config=skip_get_ec2_platforms=true -backend-config=shared_credentials_file=true -backend-config=skip_credentials_validation=true -backend-config=skip_requesting_account_id=true -backend-config=region=us-phoenix-1 -backend-config=skip_metadata_api_check=true -backend-config=force_path_style=true -backend-config=key=mwf/identity/terraform.tfstate -backend-config=endpoint=https://... -backend-config=bucket=terraform-state2 -backend-config=encrypt=true

Initializing the backend...
Error configuring the backend "s3": 2 error(s) occurred:

* : invalid or unknown key: force_path_style
* : invalid or unknown key: skip_region_validation
[terragrunt] 2018/09/14 08:19:49 Hit multiple errors:
exit status 1
```
So, we can either trap this type of error in code and put a message about checking the Terraform version support for the config element or we we should put it in README or both. Thoughts?

Also, should I update the README to discuss this support for the other s3 options along the lines of the sample config above?

Also, `shared_credentials_file` does not work ATM. I planned to put that in another forthcoming PR, but could add it in this PR. It needs to get passed into the S3 client session for HeadBucket and GetBucketVersioning as well in like fashion to S3ForcePathStyle.

And finally, in the CreateS3Client function and CreateS3Session functions, there are a ton of arguments now. And will need one more for `shared_credentials_file`. I think we should just modify them using a struct and maybe pass in the struct with these items set and just use that. So you would have something like:

```
func CreateS3Client(config *ClientConfig, terragruntOptions *options.TerragruntOptions) (*s3.S3, error) {
...
}
Thoughts?

File: remote/remote_state_s3.go (hunk 1 above)
Comment (fosrias) [reply remote/remote_state_s3.go]: BTW, I corrected the implementation in the tests commit and it will now only convert to a bool if the new config is a bool.

File: remote/remote_state_s3.go (hunk 2 above)
Comment (fosrias) [reply remote/remote_state_s3.go]: Well, that has been nagging me since the moment I coded it ;). I think the right way is to add a terragrunt internal config optionskip_bucket_versioning configuration. Then if it fails, user can configure it. Shall I do it that way?

**Comment by fosrias** [comment]
> So, we can either trap this type of error in code and put a message about checking the Terraform version support for the config element or we we should put it in README or both. Thoughts?

So, output is from Terraform and cannot parse the error string, so will have to add to documentation. Can't add in code. Will do that.

File: README.md [hunk 3]
```diff
@@ -509,8 +509,29 @@ terragrunt = {
...
+
+For backend `s3`, the following config options can be used for S3-compatible object stores, as necessary:
+
+```hcl
+remote_state {
+  # ...
+
+  skip_bucket_versioning = true # use only if the object store does not support versioning
+
+  skip_region_validation      = true
```
fosrias: @brikis98 Let me know what you think about what I have documented here from this line down. Turns out Terraform docs are [stale](https://www.terraform.io/docs/backends/types/s3.html) on this and you have to read the code to get the currently available options.

fosrias:
@brikis98 This should do it for this PR. Address prior comments.

fosrias:
Thanks. Your welcome. This is a great tool, so happy to help.

Ran `make fmt` (using vscode so reformats continuously, and not changes with `make fmt`) and ran all the tests:
```
$ sudo go test ./...                                                                                                                                                                                        
?   	github.com/gruntwork-io/terragrunt/aws_helper	[no test files]
ok  	github.com/gruntwork-io/terragrunt/cli	38.325s
ok  	github.com/gruntwork-io/terragrunt/config	0.038s
ok  	github.com/gruntwork-io/terragrunt/configstack	0.060s
ok  	github.com/gruntwork-io/terragrunt/dynamodb	37.642s
?   	github.com/gruntwork-io/terragrunt/errors	[no test files]
?   	github.com/gruntwork-io/terragrunt/options	[no test files]
ok  	github.com/gruntwork-io/terragrunt/remote	0.050s
ok  	github.com/gruntwork-io/terragrunt/shell	6.554s
ok  	github.com/gruntwork-io/terragrunt/test	283.789s
?   	github.com/gruntwork-io/terragrunt/test/helpers	[no test files]
ok  	github.com/gruntwork-io/terragrunt/util	0.022s
```
I am going to whip up the `shared_credentials_file` fix immediately, FYI.

brikis98:
Great, thanks! Merging now. I'll let the tests run and once they pass, I'll create a new release, and paste the link here.

PR #583, opened 2018-10-03
Auto-Retry

autero1:
Auto-Retry will automatically address situations where a terraform command needs to be re-run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ Terragrunt is a thin wrapper for [Terraform](https://www.terraform.io/) that pro
    1. [Interpolation Syntax](#interpolation-syntax)
    1. [Before and After Hooks](#before-and-after-hooks)
    1. [Auto-Init](#auto-init)
+   1. [Auto-Retry](#auto-retry)
    1. [CLI options](#cli-options)
    1. [Configuration](#configuration)
    1. [Migrating from Terragrunt v0.11.x and Terraform 0.8.x and older](#migrating-from-terragrunt-v011x-and-terraform-08x-and-older)
--- a/README.md
+++ b/README.md
@@ -1761,9 +1761,7 @@ As mentioned [above](#extra_arguments-for-init), `extra_arguments` can be config
 to allow customization of the `terraform init` command.
 
 Note that there might be cases where terragrunt does not properly detect that `terraform init` needs be called.
-In this case, terraform would fail.  Just run `terragrunt init` to correct this situation.
-
-
+In this case, terraform would fail.  Running `terragrunt init` again corrects this situation. See [Auto-Retry](#auto-retry).
 
 For some use cases, it might be desirable to disable Auto-Init.
 For example, if each user wants to specify a different `-plugin-dir` option to `terraform init` (and therefore it cannot be put in `extra_arguments`).
@@ -1773,6 +1771,44 @@ To disable Auto-Init, use the `--terragrunt-no-auto-init` command line option or
 Disabling Auto-Init means that you _must_ explicitly call `terragrunt init` prior to any other terragrunt commands for a particular configuration.
 
 If Auto-Init is disabled, and terragrunt detects that `terraform init` needs to be called, then terragrunt will fail.
+
+### Auto-Retry
+
+_Auto-Retry_ is a feature of `terragrunt` that will automatically address situations where a `terraform` command needs to be re-run. 
+
+Terraform can fail with transient errors which can be addressed by simply retrying the command again. In the event `terragrunt` finds one of these errors, the command will be re-run again automatically.
+ 
+In the event of a module or provider update, `terraform` commands such as `plan` or `apply` will fail with a message to run `terraform init`. In this case `terragrunt` will re-run the `init` command and then retry the command. 
+
+**Example 1**
+
+```
+$ terragrunt plan
+...
+Error: Error loading modules: module consul: not found, may need to run 'terraform init'
+```
+
+In this case, as terraform knows it needs to re-run `init` auto-retry will rerun `init` and then call `plan` again.
+
+**Example 2**
+
+```
+$ terragrunt apply
+...
+Initializing provider plugins...
+- Checking for available provider plugins on https://releases.hashicorp.com...
+Error installing provider "template": error fetching checksums: Get https://releases.hashicorp.com/terraform-provider-template/1.0.0/terraform-provider-template_1.0.0_SHA256SUMS: net/http: TLS handshake timeout.
+```
+
+Terragrunt sees this error, and knows it is a transient error that can addressed by re-running the `apply` command.
+
+`auto-retry` will try a maximum of three times to re-run the command, at which point it will deem the error as not transient, and accept the terraform failure. Retries will occur when the error is encountered, pausing for 5 seconds between retries.
+
+Known errors that `auto-retry` will rerun, are maintained in the `TerragruntOptions.RetryableErrors` map. Future upgrades to `terragrunt` may include the ability to configure `auto-retry` - pass in additional error strings, configure max retries and set retry interval - via the `terragrunt` config.
+
+To disable `auto-retry`, use the `--terragrunt-no-auto-retry` command line option or set the `TERRAGRUNT_AUTO_RETRY` environment variable to `false`.
+
+If [Auto-Init](#auto-init) is disabled, `terragrunt` will not attempt to re-run `terraform init`.
 
 ### CLI Options
 
@@ -1793,6 +1829,9 @@ start with the prefix `--terragrunt-`. The currently available options are:
   You must run `terragrunt init` yourself in this case if needed.
   `terragrunt` will fail if it detects that `init` is needed, but auto init is disabled.
   See [Auto-Init](#auto-init)
+  
+* `--terragrunt-no-auto-retry`: Don't automatically retry commands which fail with transient errors.
+  See [Auto-Retry](#auto-retry)
 
 * `--terragrunt-non-interactive`: Don't show interactive user prompts. This will default the answer for all prompts to
   'yes'. Useful if you need to run Terragrunt in an automated setting (e.g. from a script).  May also be specified with the [TF_INPUT](https://www.terraform.io/docs/configuration/environment-variables.html#tf_input) environment variable.


README.md
```diff
@@ -1774,6 +1773,44 @@ Disabling Auto-Init means that you _must_ explicitly call `terragrunt init` prio
...
+
+Terragrunt sees this error, and knows it is a transient error that can addressed by re-running the `apply` command.
+
+`auto-retry` will try a maximum of three times to re-run the command, at which point it will deem the error as not transient, and accept the terraform failure. Retries will occur when the error is encountered, pausing for 5 seconds between retries.
+
+Known errors that `auto-retry` will rerun, are maintained in the `TerragruntOptions.RetryableErrors` map. Future upgrades to `terragrunt` may include the ability to configure `auto-retry` - pass in additional error strings, configure max retries and set retry interval - via the `terragrunt` config.
+
+To disable `auto-retry`, use the `--terragrunt-no-auto-retry` command line option or set the `TERRAGRUNT_AUTO_RETRY` environment variable to `false`.
+
+If [Auto-Init](#auto-init) is disabled, `terragrunt` will not attempt to re-run `terraform init`.
```
autero1: I would assume we don't want run init no matter what if auto-init is disabled. Is this a correct assumption?

brikis98: Yes, I think that's correct

options/auto_retry_options.go
```diff
@@ -0,0 +1,18 @@
+package options
```
autero1: I opted for a separate file to keep things readable. Nevermind the actual regexp strings - those we can tweak if the approach looks reasonable.

util/collections.go
```diff
@@ -2,9 +2,21 @@ package util
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 )
 
+func MatchesAny(regExps []string, s string) bool {
+	if nil != regExps {
+		for i := 0; i < len(regExps); i++ {
```
eak12913: Perhaps this would be better expressed as:
``` for i, regExpr := range regExps {...
```

test/integration_test.go
```diff
@@ -839,6 +840,19 @@ func TestPriorityOrderOfArgument(t *testing.T) {
 	assert.Contains(t, out.String(), fmt.Sprintf("test = %s", injectedValue))
 }
 
+//func TestAutoRetryRerun(t *testing.T) {
```
eak12913: How come this is commented out?

util/collections.go
autero1: Thanks for the tip. Also, I can probably drop the nil check as I would assume that the nil range will just produce 0 iterations. My Java bg rears its ugly head  :)

test/integration_test.go
autero1: I commented it out just before the commit. Will uncomment.

shell/run_shell_cmd.go
```diff
@@ -30,19 +32,31 @@ func RunTerraformCommandAndCaptureOutput(terragruntOptions *options.TerragruntOp
 func RunShellCommand(terragruntOptions *options.TerragruntOptions, command string, args ...string) error {
```
autero1: What I'm thinking is that *all* functions should return the output, regardless of caller. The auto-retry will quite heavily depend on that and by looking at the whole code base there aren't many cases where the output is *not* needed. As such, I would be tempted to rename some of the functions - for example RunShellCommandAndCaptureOutput to something like RunShellCommandMuted. And pipe the stdout/err to ioutil.Discard instead, as the output is captured with the RunShellCommand anyways. Opinions?

shell/run_shell_cmd.go
```diff
@@ -51,13 +65,21 @@ func RunShellCommand(terragruntOptions *options.TerragruntOptions, command strin
...
+
 	cmdChannel := make(chan error)
 	signalChannel := NewSignalsForwarder(forwardSignals, cmd, terragruntOptions.Logger, cmdChannel)
 	defer signalChannel.Close()
 
-	err := cmd.Wait()
+	err = cmd.Wait()
 	cmdChannel <- err
 
+	terragruntOptions.Logger.Println(fmt.Sprintf("And the output is: %v", output))
```
brikis98: Don't forget to remove this later (I assume it's just for debugging?).

shell/run_shell_cmd.go
```diff
@@ -30,19 +32,31 @@ func RunTerraformCommandAndCaptureOutput(terragruntOptions *options.TerragruntOp
...
+	if err != nil {
+		return errors.WithStackTrace(err)
+	}
+
 	// TODO: consider adding prefix from terragruntOptions logger to stdout and stderr
 	cmd.Stdin = os.Stdin
-	cmd.Stdout = terragruntOptions.Writer
-	cmd.Stderr = terragruntOptions.ErrWriter
+	//cmd.Stdout = terragruntOptions.Writer
+	//cmd.Stderr = terragruntOptions.ErrWriter
```
brikis98: Don't forget to remove later

options/options.go
```diff
@@ -80,6 +80,21 @@ type TerragruntOptions struct {
...
 	MaxFoldersToCheck int
 
+	// Whether we should automatically run terraform init if necessary when executing other commands
+	AutoRetry bool
+
+	// Maximum number of times to retry errors matching RetryableErrors
+	MaxRetryAttempts int
+
+	// Sleep is the duration in seconds to wait before retrying
+	Sleep int
```
brikis98: Could use `time.Duration`

options/auto_retry_options.go
```diff
@@ -0,0 +1,18 @@
...
+	".*Creating metric alarm failed.*request to update this alarm is in progress.*",
+	".*Error installing provider.*TLS handshake timeout.*",
+}
+
+var ERRORS_REQUIRING_INIT = []string{
+	".*Failed to load state.*tcp.*timeout.*",
+	".*Failed to load backend.*TLS handshake timeout.*",
+	".*Creating metric alarm failed.*request to update this alarm is in progress.*",
+	".*Error installing provider.*TLS handshake timeout.*",
+}
```
brikis98: Nice configuration

shell/run_shell_cmd.go
autero1: Yes, this is for debugging at this point. Didn't want to change the return values just yet but wanted to double check that I do capture the output.

shell/run_shell_cmd.go
```diff
@@ -137,3 +166,43 @@ func (signalChannel *SignalsForwarder) Close() error {
...
+			fmt.Fprintln(outWriter, text)
+			allOutput = append(allOutput, text)
+		} else if stderrScanner.Scan() {
+			text := stderrScanner.Text()
+			fmt.Fprintln(errWriter, text)
+			allOutput = append(allOutput, text)
+		} else {
+			break
+		}
+	}
```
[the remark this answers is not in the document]
autero1: Cool, thanks!

util/collections_test.go
```diff
@@ -5,6 +5,28 @@ import (
...
+		list     []string
+		element  string
+		expected bool
+	}{
+		{nil, "", false},
+		{[]string{}, "", false},
+		{[]string{}, "foo", false},
+		{[]string{"foo"}, "kafoot", true},
+		{[]string{"bar", "foo", ".*Failed to load backend.*TLS handshake timeout.*"}, "Failed to load backend: Error...:...  TLS handshake timeout", true},
+		{[]string{"bar", "foo", ".*Failed to load backend.*TLS handshake timeout.*"}, "Failed to load backend: Error...:...  TLxS handshake timeout", false},
```
josh-padnick: It may be worth including some real-world error messages to verify that the RegEx doesn't get tripped up on any special characters.

autero1: This was a good suggestion. Tested with real world error messages and learned that I had to add the (?s) flag in the regexp to handle multi-line.

cli/cli_app.go
```diff
@@ -377,7 +376,26 @@ func runTerraformCommandIfNoErrors(possibleErrors error, terragruntOptions *opti
...
+			if isRetryable(out, tferr, terragruntOptions) {
+				terragruntOptions.Logger.Printf("Encountered an error eligible for retrying. Sleeping %v before retrying.\n", terragruntOptions.Sleep)
+				err = tferr
+				time.Sleep(terragruntOptions.Sleep)
+			} else {
+				break
+			}
+		} else {
+			err = nil
+			break
```
brikis98: You can prob just `return nil` here.

cli/cli_app.go
```diff
@@ -377,7 +376,26 @@ func runTerraformCommandIfNoErrors(possibleErrors error, terragruntOptions *opti
...
+		}
+	}
+
+	return err
```
brikis98: Based on the above suggestions, if you get here, you should probably return an error that max retries have been exceeded

cli/cli_app.go
```diff
@@ -377,7 +376,26 @@ func runTerraformCommandIfNoErrors(possibleErrors error, terragruntOptions *opti
 		return err
 	}
 
-	return shell.RunTerraformCommand(terragruntOptions, terragruntOptions.TerraformCliArgs...)
+	var err error = nil
+
+	// Retry the command configurable time with sleep in between
```
brikis98: Could you extract this into a method named something like `runTerraformCommandWithRetries`?

cli/cli_app.go
```diff
@@ -377,7 +376,26 @@ func runTerraformCommandIfNoErrors(possibleErrors error, terragruntOptions *opti
...
+	for i := 0; i < terragruntOptions.MaxRetryAttempts; i++ {
+		if out, tferr := shell.RunTerraformCommandWithOutput(terragruntOptions, terragruntOptions.TerraformCliArgs...); tferr != nil {
+			err = tferr
+			if isRetryable(out, tferr, terragruntOptions) {
+				terragruntOptions.Logger.Printf("Encountered an error eligible for retrying. Sleeping %v before retrying.\n", terragruntOptions.Sleep)
+				err = tferr
+				time.Sleep(terragruntOptions.Sleep)
+			} else {
+				break
```
brikis98: You can `return tferr` here

cli/cli_app.go
```diff
@@ -377,7 +376,26 @@ func runTerraformCommandIfNoErrors(possibleErrors error, terragruntOptions *opti
 		return err
 	}
 
-	return shell.RunTerraformCommand(terragruntOptions, terragruntOptions.TerraformCliArgs...)
+	var err error = nil
```
brikis98: You won't need this extra var per some suggestions below

cli/cli_app.go
```diff
@@ -516,6 +534,16 @@ func runTerraformInit(terragruntOptions *options.TerragruntOptions, terragruntCo
 		return errors.WithStackTrace(InitNeededButDisabled("Cannot continue because init is needed, but Auto-Init is disabled.  You must run 'terragrunt init' manually."))
 	}
 
+	initOptions, err := prepareInitOptions(terragruntOptions, terraformSource)
+
+	if nil != err {
```
brikis98: I think it's more idiomatic in Go to check `err != nil`

options/auto_retry_options.go
```diff
@@ -0,0 +1,18 @@
+package options
+
+const DEFAULT_MAX_RETRY_ATTEMPTS = 3
+const DEFAULT_SLEEP = 5
```
brikis98: `5 * time.Second`

shell/run_shell_cmd.go
```diff
@@ -72,14 +102,13 @@ func toEnvVarsList(envVarsAsMap map[string]string) []string {
 // Run the specified shell command with the specified arguments. Capture the command's stdout and return it as a
 // string.
 func RunShellCommandAndCaptureOutput(terragruntOptions *options.TerragruntOptions, command string, args ...string) (string, error) {
-	stdout := new(bytes.Buffer)
+	stdout := ioutil.Discard
```
brikis98: Why discard?

shell/run_shell_cmd.go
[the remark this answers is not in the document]
brikis98: Hm, won't this change how the output looks? If the service was logging intermittently to stdout and stderr, I think the code as written above would write everything in stdout first and then everything in stderr after. We should add a unit test specifically for this method if we can where we execute some code with known (interleaved) stdout and stderr and make sure we get back *exactly* what we expect (and not just string contains).

test/fixture-auto-retry/re-run/main.tf
```diff
@@ -0,0 +1,17 @@
+resource "null_resource" "tf_retryable_error" {
+  provisioner "local-exec" {
+    command = "if [[ -f touched ]]; then exit 0; else echo 'Failed to load backend: Error configuring the backend 's3': RequestError: send request failed caused by: Post https://sts.amazonaws.com/: net/http: TLS handshake timeout' && touch touched && exit 1; fi"
```
brikis98: Nice. You may want to add a comment explaining what this does.

shell/run_shell_cmd.go
autero1: Because we're capturing the output with the pipe now instead and don't need the buffer for that purpose.

shell/run_shell_cmd.go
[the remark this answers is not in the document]
autero1: You might be right. I'll look into it and write a unit test.

options/auto_retry_options.go
```diff
@@ -0,0 +1,18 @@
...
+
+// List of recurring transient errors encountered when calling terraform
+// If any of these match, we'll retry the command
+var RETRYABLE_ERRORS = []string{
```
autero1: Does the list of expressions look reasonable? I think I covered all the items in https://github.com/gruntwork-io/terragrunt/issues/388

brikis98: Yup!

cli/cli_app.go
autero1: @brikis98 That means that we will not have the error from the terraform execution run_shell_cmd.RunTerraformCommandWithOutput and lose the associated information (which we do for the earlier tries, as well). Is this ok? I know this is the case only when the error is retryable and we've exhausted retries - with any other error we have the information via tferr. Still wanted to double check...

shell/run_shell_cmd.go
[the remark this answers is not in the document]
autero1: I'm not sure either approach actually work. In the original if statement, the the stdoutScanner is always evaluated first and we never reach the stderrScanner until stdout is drained.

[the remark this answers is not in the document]
yorinasub17: I think I am in agreement with @autero1 . I had suggested the above method because I observed the equivalence of outputting stdout before stderr and thought that was the intended behavior, but I had not realized the desired behavior was the interleaving.

autero1:
All the tests pass.
screen shot 2018-10-05 at 10 42 36

cli/cli_app.go
autero1: Done

brikis98:
Hm, tests failed in CI. I'll dig into it tomorrow.

autero1:
Update on the failing tests. When I removed `t.Parallel()` from the AutoRetry tests, everything started working fine, even with `-parallel 128`. This must be due to how the parallel execution deals with the shared stdout and stderr that get passed to each `runTerragruntCommand` execution. Reason I didn't spot this earlier was that I was mainly running the tests from within GoLand IDE. While it uses parallel execution, the order is probably different, so we don't hit these conditions.

Other tests that seem to be failing (this is fairly random), are `TestTerragruntStdOut` and `TestLocalWithMissingBackend` + `TestLocalWithBackend`. While the first one might very well be connected due to relying on stdout, the two latter ones usually fail with creating the dynamodb lock table. Just out of curiosity I commented out `t.Parallel()` and they haven't failed since.

autero1:
Parallel execution coupled with using environment variables like this caused the issue: 
```
func TestAutoRetryEnvVarWithRecoverableError(t *testing.T) {
	t.Parallel() 
	os.Setenv("TERRAGRUNT_AUTO_RETRY", "false")
	defer os.Unsetenv("TERRAGRUNT_AUTO_RETRY")
```
The env var value set here was visible for of all other tests executing in parallel. That caused the failure in the other auto-retry tests. Removing the parallel fixed the issue.

README.md
CornerstoneII: Hi @autero1 @brikis98 can I get sample on how to use the auto-retry function? Am a noobie using Terragrunt fro the first time.

Thanks!

PR #584, opened 2018-10-07
Auto-Retry fix failing tests

autero1:
Remove parallel from a test case that uses environment variables. 
Add one more transient error expression.

### DIFF
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -892,7 +892,6 @@ func TestAutoRetryFlagWithRecoverableError(t *testing.T) {
 }
 
 func TestAutoRetryEnvVarWithRecoverableError(t *testing.T) {
-	t.Parallel()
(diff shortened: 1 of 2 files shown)


test/integration_test.go
```diff
@@ -892,7 +892,6 @@ func TestAutoRetryFlagWithRecoverableError(t *testing.T) {
 }
 
 func TestAutoRetryEnvVarWithRecoverableError(t *testing.T) {
-	t.Parallel()
 	os.Setenv("TERRAGRUNT_AUTO_RETRY", "false")
```
brikis98: Perhaps a better fix is to not set this as an env var and instead just pass it in as a CLI arg?

autero1: There is already a test testing the cli arg `TestAutoRetryFlagWithRecoverableError`. Just wanted to cover the env var case as well.

brikis98: Ah, OK, gotcha

PR #586, opened 2018-10-11
Add a command that prints the temporary directory.

### LINKED ISSUE #352: How can i know the path to the temporary folder ?
When i run terraform all modules are been downloaded in to a temporary folder,
How can i know the path to the temporary folder that the source module will be download to ? 
Or control it using the cli ?

Thanks

tamsky:
## What's changing

This PR adds a command `terragrunt-info` that intercepts normal operations but stops just after downloading source code.

## Why

After downloading sources (after `terragrunt init`), some automation may want a method for determining the download directory.

Fixes #352 

If additional work (eg: additional changes to README.md) are required, please let me know.


brikis98:
Thanks for the PR!

A few thoughts:

1. Would it make more sense to have something like a `terragrunt-info` command? It could print out in some machine-parseable format (JSON or just key:value pairs on new lines) a variety of useful info from Terragrunt: working dir, download dir, terraform binary path, env vars, etc. You could also run `terragrunt terragrunt-info xxx` to get just the info on `xxx`, analogous to how `terraform output` works.

1. Please add this to the docs and a corresponding automated test in `integration_test.go`.

tamsky:
I've recently pushed new commits to my branch, but they don't show up here.
Hm.

cli/cli_app.go
```diff
@@ -248,6 +251,21 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 		}
 	}
 
+	if shouldPrintTerragruntInfo(terragruntOptions) {
```
brikis98: This can probably go earlier, before assume role?

cli/cli_app.go
```diff
@@ -248,6 +251,21 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
...
+			"terragrunt workingdir: %s\n"+
+			"terragrunt config path: %s\n"+
+			"terraform binary: %s\n"+
+			"terraform command: %s\n"+
+			"DownloadDir: %s\n",
+			terragruntOptions.WorkingDir,
+			terragruntOptions.TerragruntConfigPath,
+			terragruntOptions.TerraformPath,
+			terragruntOptions.TerraformCommand,
+			terragruntOptions.DownloadDir)
```
brikis98: Should this be outputted in JSON format for easier machine consumption?

cli/cli_app.go
brikis98: In fact, could this check go into `runCommand`?

conorgil:
@brikis98 I like your proposal to make it similar to `terraform output` command. In fact, mirroring it even more closely would make this more usable in scripts, etc. (see [docs](https://www.terraform.io/docs/commands/output.html)).

## Proposed command ideas
Return just the value for the terragrunt download directory
```
$> terragrunt tg-info download_dir
```

Return all terragrunt info as key/value pairs, one per line
```
$> terragrunt tg-info
```

Return all terragrunt info in json format:
```
$> terragrunt tg-info -json
```

cli/cli_app.go
tamsky: I don't claim to know where the best place to drop this logic into the call chain is.

If you think within `runCommand` is best, I'll take a look there today.

cli/cli_app.go
tamsky: I did see your earlier suggestions to output in JSON format, but I didn't see any instances where `terragrunt` performs any `write()` calls to `stdout`.

I've dug into things and now think `terragruntOptions.Writer` is the correct Writer to use, and have updated the PR to emit JSON on stdout.

cli/cli_app.go
tamsky: @brikis98 I believe I put this `shouldPrintTerragruntInfo()` test at this point in `runTerragrunt` because
- we need to first call `getTerraformSourceUrl()`
- we then need to `downloadTerraformSource()`

both of which are needed in order to get the value of `WorkingDir` updated to include
the `.../.terragrunt_cache/...` in the path.

tamsky:
@brikis98 
> Thanks! Could you add a simple test for this command in integration_test.go?

I have added a simple test for this command.

tamsky:
Relevant output from appveyor:

```
--- PASS: TestTerragruntInfo (0.21s)
    integration_test.go:1545: Copying fixture-hooks/init-once to C:\Users\appveyor\AppData\Local\Temp\1\terragrunt-test204206140
    integration_test.go:1489: [TestTerragruntInfo] Full contents of show stdout:
    integration_test.go:1492: [TestTerragruntInfo] {
    integration_test.go:1492: [TestTerragruntInfo]   "WorkingDir": "C:/.terragrunt-cache/_bTjT9woEYF9ZIPVaDDpSf-5PB8/ScawSaOez9k4W0l7rb3q_-EIgG4",
    integration_test.go:1492: [TestTerragruntInfo]   "ConfigPath": "C:/Users/appveyor/AppData/Local/Temp/1/terragrunt-test204206140/fixture-hooks/init-once/with-source-no-backend/terraform.tfvars",
    integration_test.go:1492: [TestTerragruntInfo]   "TerraformBinary": "terraform",
    integration_test.go:1492: [TestTerragruntInfo]   "TerraformCommand": "terragrunt-info",
    integration_test.go:1492: [TestTerragruntInfo]   "DownloadDir": "C:/.terragrunt-cache"
    integration_test.go:1492: [TestTerragruntInfo] }
    integration_test.go:1492: [TestTerragruntInfo] 
```

cli/cli_app.go
brikis98: OK, fair enough!

cli/cli_app.go
```diff
@@ -248,6 +252,32 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 		}
 	}
 
+	if shouldPrintTerragruntInfo(terragruntOptions) {
+		type TerragruntInfoGroup struct {
```
brikis98: Could you move the type definition to a top-level item, rather than defining it within an if-statement? This will also allow re-using the type info at test time.

test/integration_test.go
```diff
@@ -1465,6 +1465,28 @@ func TestApplyAllSkipFalse(t *testing.T) {
...
+	err := runTerragruntCommand(t, fmt.Sprintf("terragrunt terragrunt-info --terragrunt-non-interactive --terragrunt-working-dir %s", rootPath), &showStdout, &showStderr)
+	logBufferContentsLineByLine(t, showStdout, "show stdout")
+
+	stdout := showStdout.String()
+
+	assert.Nil(t, err)
+	assert.Contains(t, stdout, "{")
+	assert.Contains(t, stdout, fmt.Sprintf(`"DownloadDir": "%s/%s"`, rootPath, TERRAGRUNT_CACHE))
+	assert.Contains(t, stdout, `"TerraformCommand": "terragrunt-info",`)
+	assert.Contains(t, stdout, "}")
```
brikis98: Could you `json.Unmarshall` the output and check it contains expected values?

tamsky: done.

cli/cli_app.go
tamsky: done

tamsky:
updated to address comments.
`make fmt` run.

test/integration_test.go
```diff
@@ -1465,6 +1467,30 @@ func TestApplyAllSkipFalse(t *testing.T) {
...
+	tmpEnvPath := copyEnvironment(t, "fixture-hooks/init-once")
+	rootPath := util.JoinPath(tmpEnvPath, TEST_FIXTURE_HOOKS_INIT_ONCE_WITH_SOURCE_NO_BACKEND)
+
+	showStdout := bytes.Buffer{}
+	showStderr := bytes.Buffer{}
+
+	err := runTerragruntCommand(t, fmt.Sprintf("terragrunt terragrunt-info --terragrunt-non-interactive --terragrunt-working-dir %s", rootPath), &showStdout, &showStderr)
+	assert.Nil(t, err)
+
+	logBufferContentsLineByLine(t, showStdout, "show stdout")
```
tamsky: Do we also want to emit `stderr` here?
```suggestion
	logBufferContentsLineByLine(t, showStdout, "show stdout")
	logBufferContentsLineByLine(t, showStderr, "show stderr")
```

brikis98: I think what you have is good enough!

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.18.5

PR #599, opened 2018-11-07
Feat/terragrunt iam role in config

### LINKED ISSUE #598: Allow specifying terragrunt-iam-role in terraform.tfvars
Would it be possible to add in the ability to specify the terragrunt-iam-role in the terraform.tfvars as well as through the env var and the parameter?

lnr0626:
This closes #598


lnr0626:
I updated it with the new option, also added a blurb on precedence

lnr0626:
Awesome, thanks for the quick turnaround!

PR #615, opened 2018-11-27
Some typo fixes

udondan:
Just some typo fixes I noticed while reading through the readme.

It appears like 90% line trimming, which I hope is ok. I *could* change that but I think it's not a bad thing. Most folks auto-trim their code today. Changing it now prevents this kind of changes in future whenever someone touches the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1198,8 +1198,7 @@ means that it will automatically load credentials using the
 
 ### AWS IAM policies
 
-Your AWS user must have an [IAM
-policy](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/access-control-identity-based.html)
+Your AWS user must have an [IAM policy](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/access-control-identity-based.html)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -548,8 +548,8 @@ terragrunt = {
 ```
 
 The `include` block tells Terragrunt to use the exact same Terragrunt configuration from the `terraform.tfvars` file
-specified via the `path` parameter. It behaves exactly as if you had copy/pasted the Terraform configuration from 
-the included file's terragrunt = { ... } configuration into `mysql/terraform.tfvars`, but this approach is much easier to maintain! Including variables defined outside of the terragrunt block is discussed in #132.
+specified via the `path` parameter. It behaves exactly as if you had copy/pasted the Terraform configuration from
+the included file's terragrunt = { ... } configuration into `mysql/terraform.tfvars`, but this approach is much easier to maintain! Including variables defined outside of the terragrunt block is discussed in [#132](https://github.com/gruntwork-io/terragrunt/issues/132).
```
udondan: Unlike in issues or comments, in MD files #issues are not automatically linked. Therefore added explicit link to the issue.

PR #636, opened 2019-01-13
Add configurable parallelism

### LINKED ISSUE #511: plan-all results in connections to code commit being trottled
I think this issue is subtly different to #388

I currently have a layout that looks like:

```
.
├── _global
│   └── users
│       ├── usera
│       │   └── terraform.tfvars
│       ├── userb
│       │   └── terraform.tfvars
│       └── userc
│           └── terraform.tfvars
├── account.tfvars
├── buildspec.yml
├── eu-west-2
│   └── buildsystem
│       ├── codebuild
│       │   └── deploy_to_mgmt
│       │       └── terraform.tfvars
│       └── codecommit
│           ├── codebuild
│           │   └── terraform.tfvars
│           ├── codecommit
│           │   └── terraform.tfvars
│           ├── user
│           │   └── terraform.tfvars
│           ├── dev
│           │   └── terraform.tfvars
│           ├── devplans
│           │   └── terraform.tfvars
│           ├── mgmt
│           │   └── terraform.tfvars
│           ├── mgmtplans
│           │   └── terraform.tfvars
│           ├── prod
│           │   └── terraform.tfvars
│           ├── prodplans
│           │   └── terraform.tfvars
│           ├── spg_infra
│           │   └── terraform.tfvars
│           ├── staging
│           │   └── terraform.tfvars
│           ├── stagingplans
│           │   └── terraform.tfvars
│           ├── testing
│           │   └── terraform.tfvars
│           └── testingplans
│               └── terraform.tfvars
└── terraform.tfvars
 ```

Im just trying to get things up and running just now and if I run terraform apply-all from the root for the project I get errors like this on some of my repos, (usually only 2 or 3 but its enough to cause my builds to fail):

```
[31mError copying source module: error downloading 'ssh://git-codecommit.eu-west-2.amazonaws.com/v1/repos/codecommit': /usr/bin/git exited with 128: Cloning into '/tmp/tf139393052/module'...
Throttled: too many concurrent connections.
fatal: Could not read from remote repository.
```

Ideally I'd like to detect which of the tfvars have changed and run a plan/apply on those directories only.  My ultimate plan is to generate state files which will be committed to a different repo which will then trigger a review job for someone to then merge and run the state files.  But thats the next step. 

In the short term is there any way to reduce the rate at which terragrunt apply-all triggers the individual terraform processes?

### PR DESCRIPTION
**Pull request by stephengroat**
closes #511

this is my first stab at doing some golang, currently working on fixing current tests and adding test coverage

any feedback would be greatly appreciated!

### CODE REVIEW

File: configstack/running_module.go [hunk 1]
```diff
@@ -147,14 +147,15 @@ func removeFlagExcluded(modules map[string]*runningModule) map[string]*runningMo
 // Run the given map of module path to runningModule. To "run" a module, execute the RunTerragrunt command in its
 // TerragruntOptions object. The modules will be executed in an order determined by their inter-dependencies, using
 // as much concurrency as possible.
-func runModules(modules map[string]*runningModule) error {
+func runModules(modules map[string]*runningModule, concurrency int) error {
 	var waitGroup sync.WaitGroup
+	var s = make(chan struct{}, concurrency)
```
brikis98: Please use a more descriptive name than `s`. I can't guess what that stands for?

configstack/running_module.go
```diff
@@ -181,8 +182,12 @@ func collectErrors(modules map[string]*runningModule) error {
 }
 
 // Run a module once all of its dependencies have finished executing.
-func (module *runningModule) runModuleWhenReady() {
+func (module *runningModule) runModuleWhenReady(s chan struct{}) {
 	err := module.waitForDependencies()
+	s <- struct{}{} // Lock
+	defer func() {
+		<-s // Unlock
+	}()
```
brikis98: It's not really lock and unlock right... It's add 1 to the semaphore and subtract 1? Perhaps add an explanation here that writing to the channel will block if max concurrency is exceeded.

cli/args.go
```diff
@@ -102,6 +103,15 @@ func parseTerragruntOptionsFromArgs(args []string, writer, errWriter io.Writer)
 		return nil, err
 	}
 
+	concurrency, err := parseStringArg(args, OPT_TERRAGRUNT_CONCURRENCY, os.Getenv("TERRAGRUNT_CONCURRENCY"))
```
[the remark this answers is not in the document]
stephengroat: Added

configstack/running_module.go
stephengroat: fixed

stephengroat:
> One thought: would it be enough to set runtime.GOMAXPROCS?

i didn't know how `GOMAXPROCS` would affect the running `terraform` processes, so i didn't really want to set that. wouldn't the environment variable be passed through to `terraform` and limit its concurrency as well? i was thinking this would just limit the `*-all` concurrency

brikis98:
According to the [source](https://golang.org/src/runtime/debug.go?s=533:559#L7), that method should only affect the current Go process, I think.

cli/args.go
```diff
@@ -257,6 +264,21 @@ func parseStringArg(args []string, argName string, defaultValue string) (string,
 	return defaultValue, nil
 }
 
+// Find a int argument (e.g. --foo 1) of the given name in the given list of arguments. If it's present,
+// return its value. If it is present, but has no value, return an error. If it isn't present, return defaultValue.
+func parseIntArg(args []string, argName string, defaultValue string) (int, error) {
+	for i, arg := range args {
+		if arg == fmt.Sprintf("--%i", argName) {
```
brikis98: What is `%i`? Aren't strings usually `%s`?

cli/args.go
```diff
@@ -257,6 +264,21 @@ func parseStringArg(args []string, argName string, defaultValue string) (string,
 	return defaultValue, nil
 }
 
+// Find a int argument (e.g. --foo 1) of the given name in the given list of arguments. If it's present,
+// return its value. If it is present, but has no value, return an error. If it isn't present, return defaultValue.
+func parseIntArg(args []string, argName string, defaultValue string) (int, error) {
```
brikis98: Shouldn't the `defaultValue` already be an `int`?

configstack/running_module.go
```diff
@@ -181,8 +182,12 @@ func collectErrors(modules map[string]*runningModule) error {
 }
 
 // Run a module once all of its dependencies have finished executing.
-func (module *runningModule) runModuleWhenReady() {
+func (module *runningModule) runModuleWhenReady(semaphore chan struct{}) {
 	err := module.waitForDependencies()
+	semaphore <- struct{}{} // Add one to the buffered channel. Will block in parallelism limit is met
```
brikis98: s/in/if/

cli/args.go
stephengroat: `defaultValue` gets read in by `os.GetEnv` [here](https://github.com/gruntwork-io/terragrunt/pull/636/files/88fb562714555232b161799148c018386abcc351#diff-cfc2a8f0c663e6f451ddbff25d8cffd9R106), so i thought it would be ok to pass in the string to keep consistency and do the `Atoi` if the default value was used to return

cli/args.go
stephengroat: completely right, i tested with a num, fixed

brikis98:
OK, code looks good now. Thoughts on how to test the parallelism setting is actually working?

One idea would be to create a fixture that has a simple Terraform module that uses a `local-exec` provisioner to run, say, a Python script that listens on a port for 10 seconds. You could use this module from a couple Terragrunt `.tfvars` files, so that if you run `apply-all` with no concurrency limits, one of them will succeed and the other will fail as the port is already taken. You could then have two test cases: one that sets no concurrency limits and expects a failure and one that sets the limit to 1 and expects success.

stephengroat:
I was looking at the way that terraform tests its parallelism and thought about something like this: https://github.com/hashicorp/terraform/blob/master/command/apply_test.go#L134-L261

i was going to see if i could mock up something similar to avoid a third party test dependency. would the python script be preferable or something like above?

brikis98:
TBH, I see two approaches to testing this:

1. Unit test. To make that work, you'd extract all the code with using channels to limit parallelism into a separate package with some sort of `Execute(job)` method and create unit tests that validate `Execute` only runs a max of N things concurrently.

1. Integration test. Run `terragrunt apply-all` for real and make sure the parallelism limit is enforced through the Python code or similar trick.

Creating all sorts of mocks and counters seems like an in between that doesn't seem as effective as either of these options.

stephengroat:
Gotcha, completely agree.

To avoid a `python` dependency, I'm going to see if I can use [go Transport](https://golang.org/pkg/net/http/#Transport) to create a https proxy and count the # of connections. think that would work?

mhulscher:
Although practically the same, have you considered using the sync/semaphore package instead of a buffered channel? https://godoc.org/golang.org/x/sync/semaphore

brikis98:
@stephengroat What is the status of this PR?

ondrejklucka:
We have tested #636 (on large infra of 100s of modules) and it seems to be working as expected.

How can we help to get the patch merged?

cschroer:
@brikis98 from a real world test (huge project with tons of modules and such) this patch works fine, see also test from @ondrejklucka - any stuff missing to get it merged? we may could help here ;)

brikis98:
@ondrejklucka @cschroer Thanks for testing! I'd be happy to proceed with this PR, but we also need automated tests to be part of it. Otherwise, there may be important use cases already broken by these changes, or we may break things in the future and not realize it. I haven't heard back from @stephengroat. Anyone else want to pick up the PR?

ryno75:
Any updates on this PR?

brikis98:
@stephengroat Any plans to finish this one up?

geota:
@brikis98  I need this feature and may be willing to finish of the PR if @stephengroat doesn't plan to.

brikis98:
@geota New PR to wrap this up is very welcome!

stephengroat:
sorry, i had a kid, finishing this up probably isn't going to happen. feel free to copy and paste this code at will

ryno75:
> sorry, i had a kid, finishing this up probably isn't going to happen. feel free to copy and paste this code at will

@geota you want to run with it?

andysworkshop:
Any chance of this being merged? The conflicts are simple enough to resolve and the feature itself has been invaluable to us.

stephengroat:
in thinking some about this, i added a test for serial execution (parallelism limit of 1) and tried using golang's trace tool, but the APIs for it are internal and cannot be imported into any package.

Serial execution is really the worst case/only case that may need to be tested (i question whether testing that actually adds any value). otherwise, limiting the parallelism of go is system dependent (less cores, scheduler, etc.)

any race conditions/issues created by this PR could also naturally happen through system conditions.

i don't know if i seem much value is testing go's chan implementation

brikis98:
Not sure I follow your argument? Are you saying we shouldn't test this feature and just assume it works? 

The goal of automated tests is to give us confidence that our code works—and that it keeps working as we make more changes in the future. Here are the things I want to be confident about with this new parallelism feature:

1. That when we set parallelism to 1, exactly one module is processed at a time. 
1. That when we set parallelism to 10, that 10 modules are processed at a time.
1. That none of the existing behavior of Terragrunt is broken by adding this feature.

We can have confidence in (3) due to the large suite of automated tests already in Terragrunt. But we still need confidence in (1) and (2). We can do this using the approach I described before:

> One idea would be to create a fixture that has a simple Terraform module that uses a local-exec provisioner to run, say, a Python script that listens on a port for 10 seconds. You could use this module from a couple Terragrunt .tfvars files, so that if you run apply-all with no concurrency limits, one of them will succeed and the other will fail as the port is already taken. You could then have two test cases: one that sets no concurrency limits and expects a failure and one that sets the limit to 1 and expects success.

stephengroat:
Completely understand, don't mean to insult, I (personally) don't see any difference between `#2` and `#3`.

I can create test cases for `#3` that currently break (effectively overwhelming system resources by creating too many Go rountines, i.e. hundreds or thousands of sub directories, creating so many subprocesses that system resources are overwhelmed - stack, heap, etc.). These will be super hard to replicate in test environments, but I've done this multiple times in AWS CodeBuild on smaller runners. In those cases, the parallelism is then effectively set by the system/scheduler/etc., limiting (or attempting to limit) the number of processes created. Currently, IMO, `#3` doesn't really test the infinite number of processes that's currently possible (since that's impossible to actually test), it tests the system limits of the machine tests are being run on (some arbitrary number for GOMAXPROCS defined by the number of cores, threads, etc.). `#3` and `#2` are essentially equivalent, the difference is in `#3` the parallelism is set by available system resources vs in `#2` the limit is explicitly set in the program.

If you have confidence in the current test suite, either limiting the parallelism tests to 1 or settings `GOMAXPROCS` to `1` would be a worst case scenario (terragrunt is limited/pinned to a single core). Since there are assurances that cyclical dependencies don't exist, there should never be the need for parallelism while running module (so a limit of 1 and GOMAXPROCS=1 should always work, being valid tests). To assure multicore compatibility, the test suite should be run on a multicore system. Even on that system, unless you use `go tool trace` and find a way to implement an API using `internal/trace` (which i tried, but failed at), there's no way to assure parallel execution has occurred, even on multicore system (still at the behast of the system/scheduler without process to core pinning).

Hope this makes sense, I just don't see any value in implementing complicated testing for a simple channel semaphore implementation that is system dependent (using external test suites to validate parallelism may not be effective because of system resources, schedulers, etc.). The idea that you have (a Python port listener) is valid, but it's system dependent, as the Python and Golang processes would contend for thread execution time, limiting tests depending on system resources again. The tests would again be scheduler and system dependent, very hard to reliably profile.

I completely understand if you don't want to merge, but the current implementation of infinite isn't really tests and without cyclical dependencies, this could shouldn't be an issue (without significant code review). IMO, it's very similar to currently testing code, just setting system limiting using a process limit vs system characteristics (Hope this wasn't too rambling!)

stephengroat:
To effectively test parallelism, there really needs to be a trace tool. Golang is working on opening their trace API (https://github.com/golang/go/issues/29104), otherwise, the entire program could be implemented with the https://github.com/opentracing/opentracing-go (which seems like a lot of work, but I'm not 100% sure, not too familiar with tracing)

brikis98:
> I (personally) don't see any difference between #2 and #3.

I don't follow. In (2), multiple modules are running concurrently, so it should be possible to consistently demonstrate a race condition, such as with the port number idea I suggested. In (3), only one module is running a time, so no race condition should ever be reproducible. This is a massive and critical difference.

Nothing about this is system dependent, there's no need for infinite parallelism, overwhelming the system, some fancy CI setup, or any of that. If your code works, (2) and (3) should be have differently. If your code doesn't, they won't.

sumanmukherjee03:
This would be a very helpful feature for us as well.

thijsdev:
We really need this feature too. Keep running into max GIT en too many open file issues. Is there any insight into a merge/release date schedule?

PR #639, opened 2019-01-23
Improve test stability

autero1:
This PR tries to address intermittent test failures with terragrunt.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,10 @@ jobs:
     - persist_to_workspace:
         root: *workspace_root
         paths: vendor
-
+# We're running unit tests separately from integration tests - with no parallelization.
+# With heavy parallelization coupled with re-use of test fixtures we've witnessed slight
+# instability with the tests. The unit tests are fast to execute, so there is negligible
+# performance penalty.
   unit_test:
     <<: *defaults
     steps:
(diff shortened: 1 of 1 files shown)


autero1:
Tests that have has most failures recently are these:
`TestResolveTerraformModulesTwoModulesWithDependenciesIncludedDirsWithDependencyExcludeModuleWithNoDependency`
`TestResolveTerraformModulesTwoModulesWithDependenciesIncludedDirsWithDependency`
The tests themselves and the functionality they test look perfectly ok. Executing separately from within my IDE, the tests never fail. But increasing parallelism looks to increase the error rate. 

These tests (together with all other tests in the same directory) use the same fixture and have heavy recursive dependencies and rarely seem to fail with no parallelism. For that and "SoC" reasons I decided to split the `unit tests` and `integration tests` into separate stages with no parallelization in unit testing and default parallelization (128) in integration testing. At least the initial run was successful. And I didn't have to change or disable any tests 😄. 

Does this make sense?

yorinasub17:
Approach makes sense to me. Don't forget to also add comments into the circleci config to describe why the unit tests are run with no parallelism (and why it is run separately).

autero1:
Fixed. And after the commit second round of successful tests.

PR #645, opened 2019-02-03
 Encrypt the bucket and log access on the AWS S3 Bucket of the TFState

Xtigyro:
Encrypt the bucket and log the access on the AWS S3 Bucket of the Terraform State - both bucket-wide.

PR #646, opened 2019-02-05
Add access logging permissions

brikis98:
This is a follow-up to #645. The tests were failing with the error:

```
InvalidTargetBucketForLogging: You must give the log-delivery group WRITE and READ_ACP permissions to the target bucket
```

This PR is an attempt to grant those permissions.

### DIFF
--- a/configstack/module.go
+++ b/configstack/module.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strings"
 
 	"github.com/gruntwork-io/terragrunt/config"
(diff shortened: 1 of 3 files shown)


Xtigyro:
@brikis98 Can I assist in any way?

brikis98:
@Xtigyro Thanks, but I _think_ this PR fixes the issue, but I'm hitting some other, perhaps unrelated test failures. Will look into it tomorrow or day after, time permitting.

brikis98:
OK, the tests were failing due to an unrelated issue, but still a bug: when processing include/exclude arguments for `xxx-all` commands, we were iterating over a map, and in Go, map iteration order is random. So sometimes we'd process includes one way and sometimes another. This caused the tests to fail, but more importantly, it meant that real-world behavior could be inconsistent too. I've fixed it so we use a consistent sort order now and the tests are passing. Merging!

Xtigyro:
@brikis98 Very nice!

PR #647, opened 2019-02-05
Remote State DynamoDB Lock Table Encryption

Xtigyro:
Encrypt the DynamoDB Lock Table of the TF State - only if not already encrypted and even if it was already created.

PR #650, opened 2019-02-11
Fix nil pointer. Wait for encryption. Make DynamoDB encryption off by default.

brikis98:
Follow-up fixes for #647:

1. Fix a nil pointer dereference when looking up tables that don't have encryption enabled.
1. It turns out that it takes 1-2 minutes to actually enable encryption, so add a wait. Otherwise, we start using the table before it's ready.
1. Disable DynamoDB encryption by default as (a) Terraform doesn't store any sensitive data in DynamoDB and (b) it takes 1-2 minutes to enable it, which makes using Terragrunt a lot slower. It can be enabled on-demand using the `enable_lock_table_ssencryption` setting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -522,10 +522,10 @@ For backend `s3`, the following config options can be used for S3-compatible obj
 remote_state {
   # ...
 
-  skip_bucket_versioning        = true # use only if the object store does not support versioning
-  skip_bucket_ssencryption      = true # use only if non-encrypted Terraform State is required and/or the object store does not support server-side encryption
-  skip_bucket_accesslogging     = true # use only if the cost for the extra object space is undesirable or the object store does not support access logging
-  skip_lock_table_ssencryption  = true # use only if non-encrypted DynamoDB Lock Table for the Terraform State is required and/or the NoSQL database service does not support server-side encryption
+  skip_bucket_versioning         = true # use only if the object store does not support versioning
+  skip_bucket_ssencryption       = true # use only if non-encrypted Terraform State is required and/or the object store does not support server-side encryption
+  skip_bucket_accesslogging      = true # use only if the cost for the extra object space is undesirable or the object store does not support access logging
+  enable_lock_table_ssencryption = true # use only if non-encrypted DynamoDB Lock Table for the Terraform State is required and/or the NoSQL database service does not support server-side encryption
 
   shared_credentials_file     = "/path/to/credentials/file"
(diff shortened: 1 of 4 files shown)


brikis98:
@Xtigyro I got a number of test failures and have created this PR to fix them. Are you by any chance _just_ running the tests in the `config` or `dynamodb` folder or something that you're not seeing these?

Xtigyro:
> 
> 
> @Xtigyro I got a number of test failures and have created this PR to fix them. Are you by any chance _just_ running the tests in the `config` or `dynamodb` folder or something that you're not seeing these?

I've run the tests in the `remote` directory. The output and what was run can be checked in the other PR. I'll try again in the evening. Can I assist with anything?

brikis98:
> I run the tests in the `remote` directory. The output and what was run can be checked in the other PR.

Ah, I didn't notice that. Yea, there are tests in most of the packages, not just `remote`, and `test` in particular has integration tests that were failing. Please make sure to run all tests in the future!

I've got all the tests passing now, but it looks like enabling encryption for DynamoDB takes 1-2 minutes. That's a lot of overhead for the first call to `init` and for the tests... It makes sense to encrypt the S3 bucket by default, as Terraform state can contain secrets. But DynamoDB only contains some hashes and lock info, so encrypting it seems much less important. Given the overhead, I think I'm going to disable encryption for the lock table by default, but make it possible to encrypt via config.

Xtigyro:
> 
> 
> > I run the tests in the `remote` directory. The output and what was run can be checked in the other PR.
> 
> Ah, I didn't notice that. Yea, there are tests in most of the packages, not just `remote`, and `test` in particular has integration tests that were failing. Please make sure to run all tests in the future!
> 
> I've got all the tests passing now, but it looks like enabling encryption for DynamoDB takes 1-2 minutes. That's a lot of overhead for the first call to `init` and for the tests... It makes sense to encrypt the S3 bucket by default, as Terraform state can contain secrets. But DynamoDB only contains some hashes and lock info, so encrypting it seems much less important. Given the overhead, I think I'm going to disable encryption for the lock table by default, but make it possible to encrypt via config.

@brikis98 OK - will surely do!

Yes - I completely agree: whoever can accept the longer initialisation time, should pass that as an argument. Thank you very much!

brikis98:
OK, new release is here: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.18.0

Issue #610, opened 2018-11-20
sha256 verification for the download binaries


211217613:
I'm not familiar with githubs gpg signing commits and general binary generation for releases.

can you post a sha256 of the binaries for verification.

brikis98:
We do not currently publish checksums, but it's definitely a good idea to do so, so I've marked it as an enhancement for us to consider for the future.

lestephane:
Pending publication of said checksums, what is the workaround?

The best I can do is post  the checksum I get, so someone here can confirm:
```bash
$ sha256sum terragrunt_linux_amd64 # v0.17.4
7bdedb8c859450ea6fdd995c4c25f7bcdf6f7122f175268a896a3810a27ae85b  terragrunt_linux_amd64
```
Is this checksum correct?

211217613:
```
$ sha256sum terragrunt_linux_amd64
7bdedb8c859450ea6fdd995c4c25f7bcdf6f7122f175268a896a3810a27ae85b  terragrunt_linux_amd64
```

conn:
@211217613 commits, merge commits, and tagging that happens via the GitHub UI can optionally be signed by GitHub's signing keys. Depending on which infrastructure GitHub uses to sign commits vs hosting git repositories and their web applications, this may only be marginally more verified than no signature at all.

GitHub themselves do not generate the binaries; the binaries a compiled by CircleCI which all is configured [here](https://github.com/gruntwork-io/terragrunt/blob/master/.circleci/config.yml). Once the binaries are compiled, they are uploaded to GitHub as the release binaries that will eventually be downloaded.

Distributing checksums alongside binaries isn't really more secure than distributing without. Since you're distributing both within the same channel, at the same time, via the same methods they are both at risk if the application that serves both is compromised. A far better solution would be for Gruntwork to us a long-lived PGP signing key that is distributed separately and is available from multiple sources (their website, the git repository, etc.)

Hope this answers all your questions and concerns.

arledesma:
@brikis98

How about publishing a `.sig`, along side the release, that is created by [gruntworkci](https://keybase.io/gruntworkci) that can be easily verified.  This would align with what @conn is stating.

Proofs, or simple checksum files, could also be posted to https://keybase.pub/gruntworkci/ during the release process.

[...]

PR #669, opened 2019-03-02
Add checksums to release files

conn:
This is to address:
https://github.com/gruntwork-io/terragrunt/issues/610

The checksums of all binaries within `bin/` will be saved within
`SHA256SUMS`. The integrity of binaries can now be verified with this
command: `sha256sum --check --ignore-missing SHA256SUMS`.

In the future, Gruntwork should consider using a PGP key to
cryptographically sign this checksums file and to distribute the
signature along with the rest of the release files. Doing so will likely
require using private CI/CD infrastructure instead of CircleCI so that
Gruntwork may have full control over their private signing key.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,6 +79,7 @@ jobs:
     - checkout
     - attach_workspace:
         at: *workspace_root
+    - run: cd bin && sha256sum * > SHA256SUMS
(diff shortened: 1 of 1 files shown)


brikis98:
Thank you! 👍

conn:
You're welcome!

PR #673, opened 2019-03-08
Fix broken test introduced on PR #670

bernardoVale:
Sorry about that :D


brikis98:
Nice, thank you for the fix!

PR #675, opened 2019-03-08
Add command 'show-all'

### LINKED ISSUE #674: Add support for terraform show/show-all

troylar:
The goal of this PR is to resolve #674 to add the ability to call `show-all` to get a report of current state across all folders.


troylar:
@brikis98 Let me take a stab at it . . .

jprice-da15252:
@troylar, was a followup to this started?  I was looking at hacking together a script that did a similar thing using bash to find all output files matching a specified naming convention, and run a terraform show for automation purposes.  Before I did that, I wanted to make sure this wasn't something that was going to be implemented soon, since a native solution would probably be the better route.

aschleifer:
Is there an update on this? I really could use a `show-all` command to combine with https://docs.gitlab.com/ee/user/infrastructure/#output-terraform-plan-information-into-a-merge-request

Issue #668, opened 2019-02-28
External authentication mechanisms


jasonmcintosh:
Be nice if we could setup a hook to call an external authentication mechanism to allow remote state to work.  Example:
go acquire creds from vault and store in credentials file.  THEN init/apply/plan commands.  Currently before_hooks execute after the remote state is configured.

It'd be extra nice if these hooks would also set/get environment variables so we could pass those on vs. using credential files.

brikis98:
We've had several requests for hooks that run earlier in the lifecycle (e.g., https://github.com/gruntwork-io/terragrunt/issues/620). Seems like a generally useful feature that would support this use case too. Bonus points if hooks can affect env vars of Terragrunt/Terraform itself. This stuff is on our list, but I'm not sure when we'll get to it. If anyone is up for it, a PR to add this would be very much appreciated 👍

tiwood:
We actually have something implemented to make this possible. @tmeckel made changes to our Terraform/Terragrunt setup to allow hooks that parse their own output as variables and pass them as environment variables to subsequent hooks or to the provider configuration.

Like so:

```hcl
before_hook "get_secrets" {
      commands = [
        "init",
        "apply",
        "plan",
        "destroy",
        "import",
        "push",
        "refresh",
        "validate",
      ]

      load_env_vars {
        execute = [
          "pwsh",
          "-nop",
          "-nol",
          "-File",
          "${get_parent_tfvars_dir()}/Scripts/Get-Secrets.ps1",
        ]
      }

      run_on_error = false
    }
```

We expect KEY=value output to STDOUT from the hook. Then these variables can be used with standard interpolation (get_env()).

@tmeckel, any plans to contribute this back to this repository soon? 🙃️

tmeckel:
@brikis98  I can provide a WIP PullRequest of the current implementation over the weekend that we're currently have in place as @tiwood mentioned. The changes that we're applied to terragrunt include late binding (evaluation) of environment variables every time a command is issued and the ability to change/initialize the environment variables used by terragrunt via a hook.

tmeckel:
@brikis98  there you go .. as promised I provided a pull request (#680) with the changes we applied to Terragrunt

yhakbar:
FYI this feature exists:
https://terragrunt.gruntwork.io/docs/features/aws-authentication/#auth-provider-command

PR #680, opened 2019-03-16
WIP: Changing runtime environment and late interpolations

tmeckel:
## Description

This Pull Request contains functionality to dynamically load and change the runtime environment of Terragrunt and it adds late binding of interpolations.

## Motivation

Terragrunt interpolates the defined variables and function on startup when the configuration files are loaded. (https://github.com/tmeckel/terragrunt/blob/bf8f423f569227e339ddbe3370c227b7610850c7/config/config.go#L284) Later on it's no more possible to change the used runtime environment or reevaluate an interpolation with a changed environment. Thus to supply Terragrunt with an environment that contains dynamic login information for example with values from an Azure Keyvault it's required to initialize the runtime environment (variables) before Terragrunt is started and does the interpolation while loading the config files. Not only this might lead to a matryoshka phenomenon in which a script calls a script calls a script calls Terragrunt but it's also error prone because if the script chain isn't run but only Terragrunt this will lead to errors because the environment isn't initialized properly. Following the DRY principle it should be possible to initialize or change the runtime environment only using Terragrunt.

## New and changed functions

As described it's now possible to change the runtime environment with the new statement `load_env_vars` inside a hook definition. @tiwood described how we're are using this functionality in a customer project in his comment to issue #668 (https://github.com/gruntwork-io/terragrunt/issues/668#issuecomment-472038719). The load `load_env_vars` statement supports the definition of a command to execute (the syntax has been adopted from the the definition of command inside a hook) and two flags `overwrite` and `transient` which control the behavior how environment variables are set and the persistency of changes to the runtime environment. 

The `load_env_vars` statement can be defined in every hook definition and the command inside will always be executed before the hook command is run. The following example shows how to load secrets into the runtime environment by using a before hook definition for a set of states.

```hcl
before_hook "get_secrets" {
      commands = [
        "init",
        "apply",
        "plan",
        "destroy",
        "import",
        "push",
        "refresh",
        "validate",
      ]

      load_env_vars {
        execute = [
          "pwsh",
          "-nop",
          "-nol",
          "-File",
          "${get_parent_tfvars_dir()}/Scripts/Get-Secrets.ps1",
        ]
      }

      run_on_error = false
    }
```

Because the flags `overwrite` and `transient` aren't specified the changes to the environment will be permanent (`transient == false`) and existing environment variables won't be replaced because `overwrite == false`. With `overwrite == false` existing variables will be changed and new variables will be added. With `overwrite == true` the complete runtime environment will be changed and will only contain those variables returned by the command executed. The `transient` flags allows to setup a temporary runtime environment for a hook command when it's set to `true` because after the hook command has been run the changes to the runtime environment will be reverted.

The following table shows the effects on the environment with all combinations of those two flags.

| Flags                          | Description                                                                                                                                                                                                                                                                       |
|--------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| transient=false, overwrite=false | Changes to the environment are persitent and will be available to all hook functions executed after the current one. The runtime environment will contain all values returned from the command executed in `load_env_vars` and all environment variables which already existed.     |
| transient=true, overwrite=false  | Changes to the environment are not persitent and will not be available to hook functions executed after the current one. The runtime environment will contain all values returned from the command executed in `load_env_vars` and all environment variables which already existed. |
| transient=false, overwrite=true  | Changes to the environment are persitent and will be available to all hook functions executed after the current one. The runtime environment will be replaced and thus will only contain those variables and values returned from the command executed in `load_env_vars`.          |
| transient=true, overwrite=true   | Changes to the environment are not persitent and will not be available to hook functions executed after the current one. The runtime environment will be replaced and thus will only contain those variables and values returned from the command executed in `load_env_vars`.      |

Aside the new ability to change the runtime environment Terragrunt will now perform interpolations every time a hook command is executed and not only when the configuration is loaded (https://github.com/tmeckel/terragrunt/blob/own-load-env/shell/run_shell_cmd.go#L49). 

## Implementation remarks

The Pull Request is marked as WIP because the unit and integration tests haven't been changed yet according to the new late binding of interpolations and the ability to change the runtime environment as described above. Further more while implementing the new functions it was required to move code files to other or new packages because GO detected a loop in the import statements. Those changes should be reviewed and changed again if required. At least the if statement at https://github.com/tmeckel/terragrunt/blob/own-load-env/shell/run_shell_cmd.go#L75 should be reviewed again now that Terragrunt is doing late interpolations because the if-statement add parameters to Terraform depending on current parameter (argument) values.


cli/cli_app.go
```diff
@@ -274,16 +276,66 @@ func processHooks(hooks []config.Hook, terragruntOptions *options.TerragruntOpti
...
+				targetEnvMap := terragruntOptions.Env
+				if curHook.LoadEnvVars.Transient {
+					oldEnvMap = terragruntOptions.Env
+					targetEnvMap := make(map[string]string)
+					for key, value := range terragruntOptions.Env {
+						targetEnvMap[key] = value
+					}
+					terragruntOptions.Env = targetEnvMap
+				}
+				if curHook.LoadEnvVars.Overwrite {
```
brikis98: Should this be an `else if`?

cli/cli_app.go
```diff
@@ -274,16 +276,66 @@ func processHooks(hooks []config.Hook, terragruntOptions *options.TerragruntOpti
...
+
+				if possibleError != nil {
+					terragruntOptions.Logger.Printf("Error running hook %s with message: %s", curHook.Name, possibleError.Error())
+					errorsOccurred = append(errorsOccurred, possibleError)
+				}
 			}
 
+			if oldEnvMap != nil {
+				terragruntOptions.Env = oldEnvMap
+			}
```
brikis98: Can you explain the thinking behind `oldEnvMap`?

config/config.go
```diff
@@ -281,11 +278,7 @@ func ParseConfigFile(configPath string, terragruntOptions *options.TerragruntOpt
 
 // Parse the Terragrunt config contained in the given string.
 func parseConfigString(configString string, terragruntOptions *options.TerragruntOptions, include *IncludeConfig, configPath string) (*TerragruntConfig, error) {
-	resolvedConfigString, err := ResolveTerragruntConfigString(configString, include, terragruntOptions)
-	if err != nil {
-		return nil, err
-	}
-
+	resolvedConfigString := configString
```
brikis98: What is the motivation for this change? I suspect this will cause issues downstream...

go.mod
```diff
@@ -0,0 +1,23 @@
...
+	github.com/jmespath/go-jmespath v0.0.0-20180206201540-c2b33e8439af
+	github.com/mattn/go-zglob v0.0.0-20170920030352-4ecb59231939
+	github.com/mitchellh/go-homedir v0.0.0-20161203194507-b8bc1bf76747
+	github.com/mitchellh/go-testing-interface v0.0.0-20171004221916-a61a99592b77
+	github.com/mitchellh/mapstructure v0.0.0-20171017171808-06020f85339e
+	github.com/pmezard/go-difflib v0.0.0-20151028094244-d8ed2627bdf0
+	github.com/stretchr/testify v0.0.0-20171018052257-2aa2c176b9da
+	github.com/ulikunitz/xz v0.5.4
+	github.com/urfave/cli v0.0.0-20171014202726-7bc6a0acffa5
+)
```
brikis98: What are these `.mod` / `.sum` files?

options/options.go
```diff
@@ -105,6 +106,26 @@ type TerragruntOptions struct {
 	// packages can use the command without a direct reference back to the cli package (which would create a
 	// circular dependency).
 	RunTerragrunt func(*TerragruntOptions) error
+
+	Context TerragruntContext
```
brikis98: What's this?

remote/remote_state.go
```diff
@@ -79,28 +81,46 @@ func (remoteState *RemoteState) NeedsInit(terragruntOptions *options.TerragruntO
...
 		return initializer.NeedsInitialization(remoteState.Config, state.Backend, terragruntOptions)
-	} else if state.IsRemote() && remoteState.differsFrom(state.Backend, terragruntOptions) {
+	} else if state.IsRemote() {
+		needsInit, err := remoteState.differsFrom(state.Backend, terragruntOptions)
+		if nil != err {
+			return false, err
+		}
 		// If there's no remote state initializer, then just compare the the config values
-		return true, nil
+		return needsInit, nil
```
brikis98: Is this change related to this PR?

.gitattributes
```diff
@@ -0,0 +1,2 @@
+go.mod  text eol=lf
+go.sum  text eol=lf
```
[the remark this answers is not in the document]
tmeckel: It ensures that the go.mod and go.sum files will always be checked out with a "UNIX" file endind. I adopted this while I was working on PACKER. Sometimes git will set the state of files to changed/modified when it gets confused by the line ending.

cli/cli_app.go
```diff
@@ -274,16 +276,66 @@ func processHooks(hooks []config.Hook, terragruntOptions *options.TerragruntOpti
 	for _, curHook := range hooks {
 		allPreviousErrors := append(previousExecError, errorsOccurred...)
 		if shouldRunHook(curHook, terragruntOptions, allPreviousErrors...) {
+			var oldEnvMap map[string]string
+
 			terragruntOptions.Logger.Printf("Executing hook: %s", curHook.Name)
-			actionToExecute := curHook.Execute[0]
-			actionParams := curHook.Execute[1:]
-			possibleError := shell.RunShellCommand(terragruntOptions, actionToExecute, actionParams...)
+			if curHook.LoadEnvVars != nil {
```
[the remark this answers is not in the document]
tmeckel: I will introduce the following function
```go
// loadEnvVars loads environment variable definitions from a shell command.
// Currently the function expects KEY=value output to STDOUT
// the method returns the old evironment variables map (dictionary) if the
// changes should be transient.
func loadEnvVars(terragruntOptions *options.TerragruntOptions, curHook config.Hook) (map[string]string, error) {
}
```

cli/cli_app.go
```diff
@@ -274,16 +276,66 @@ func processHooks(hooks []config.Hook, terragruntOptions *options.TerragruntOpti
...
+				terragruntOptions.Logger.Printf("Loading environment for hook: %s", curHook.Name)
+
+				actionToExecute := curHook.LoadEnvVars.Execute[0]
+				actionParams := curHook.LoadEnvVars.Execute[1:]
```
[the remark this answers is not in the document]
tmeckel: This checked at https://github.com/tmeckel/terragrunt/blob/8066792ee5909fee702b6e03ee4a88e2cb56ca2d/config/config.go#L117. I agree as a it comes to defensive prgramming we should add a if-statemnt here as well.

cli/cli_app.go
```diff
@@ -274,16 +276,66 @@ func processHooks(hooks []config.Hook, terragruntOptions *options.TerragruntOpti
...
+					txt := scanner.Text()
+					idx := strings.Index(txt, "=")
+					if idx > 0 {
+						key := strings.TrimSpace(txt[0:idx])
+						value := txt[idx+1 : len(txt)]
+						m[key] = value
+					} else {
+						terragruntOptions.Logger.Printf("Invalid key,value pair [%s]", txt)
+					}
+				}
```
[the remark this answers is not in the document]
tmeckel: As stated above this all will move to a separate function,

cli/cli_app.go
```diff
@@ -274,16 +276,66 @@ func processHooks(hooks []config.Hook, terragruntOptions *options.TerragruntOpti
...
+				scanner := bufio.NewScanner(strings.NewReader(out.Stdout))
+				for scanner.Scan() {
+					txt := scanner.Text()
+					idx := strings.Index(txt, "=")
+					if idx > 0 {
+						key := strings.TrimSpace(txt[0:idx])
+						value := txt[idx+1 : len(txt)]
+						m[key] = value
+					} else {
+						terragruntOptions.Logger.Printf("Invalid key,value pair [%s]", txt)
```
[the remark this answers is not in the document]
tmeckel: With the new parsing function for environment variables the code inside will look like the following
```go
oldEnvMap, err := loadEnvVars(terragruntOptions, curHook)
if nil != err {
    errorsOccurred = append(errorsOccurred, err)
    continue
}
```
and we left the fact that a line from the stdout can't be parsed as `key=value` as a "warning". But we surely can make this an error.

cli/cli_app.go
tmeckel: When you look at the explanation of the two flags `transient` and `overwrite` it's required to save the old environment variables and reset them on the `terragruntOptions` instance when `transient=true`

go.mod
tmeckel: The files allows go to compile terragrunt outside GOPATH.
https://github.com/golang/go/wiki/Modules

config/config.go
tmeckel: I guess it was a leftover while I was trying to apply all changes that were required to implement late interpolations with the minial effect on the existing code. The variable can be definitly removed and `configString`can be used directly inside the function `parseConfigString`

options/options.go
tmeckel: Because terragrunt is able to to include config files we must be able to store some information what config files has been included and what's even more important from where they have been included because to fully support late interpolations we must save the include path information because the interpolation functions `path_relative_to_include` and `path_relative_from_include`rely on them.

remote/remote_state.go
tmeckel: Yes it definitly relates to this PR because I had to extend the return value of `remote_state.go::NeedsInit` with an `error` object because the method must now do a late interpolation https://github.com/tmeckel/terragrunt/blob/8066792ee5909fee702b6e03ee4a88e2cb56ca2d/remote/remote_state.go#L71 and the interpolation might result in an error. In addtion based on the interpolation of the settings for the remote state implementation it might need an initialization or not.

go.mod
brikis98: Ah, OK. Could we do this in a separate PR? It requires more thinking (i.e., do we really want _both_ Go dep and Go mod, if we remove Go dep, does that break existing users, etc).

tmeckel:
> Thanks for the PR! Some thoughts:
> 
> 1. Instead of a `load_env_vars` block, perhaps a `parse_env_vars_from_stdout` boolean param would be cleaner? That way, hooks always have the same consistent syntax:
>    ```terraform
>    before_hook "before_hook_1" {
>      commands = ["apply", "plan"]
>      execute = ["echo", "Foo"]
>      run_on_error = true
>      parse_env_vars_from_stdout = true
>    }
>    ```
The load_env_vars construct can be used to setup an runtime einvironment for an command that is executed in as a hook. With the two flags `transient`and `overwrite` this gives you the possibility to setup a special environment just for a hook function which is executed via the `execute`definition. 

> 2. What was it that created the dependency cycle? I'd like to reduce the amount of file shuffling in this PR if possible.
Sorry but I think I dont't understand you question. 

> 3. How did you test these changes?
We're using the changed version of terragrunt to deploy cloud ressources at an investment bank. The changed version works flawlessly for us up to now. Perhaps @tiwood can add some more information on this.

go.mod
tmeckel: Sure no problem. Although I didn't have any issues compiling terragrunt either inside GOPATH or outside using go.mod. GO is smart enough to handle this.

cli/cli_app.go
```diff
@@ -274,16 +276,66 @@ func processHooks(hooks []config.Hook, terragruntOptions *options.TerragruntOpti
...
 			terragruntOptions.Logger.Printf("Executing hook: %s", curHook.Name)
-			actionToExecute := curHook.Execute[0]
-			actionParams := curHook.Execute[1:]
-			possibleError := shell.RunShellCommand(terragruntOptions, actionToExecute, actionParams...)
+			if curHook.LoadEnvVars != nil {
+				terragruntOptions.Logger.Printf("Loading environment for hook: %s", curHook.Name)
+
+				actionToExecute := curHook.LoadEnvVars.Execute[0]
+				actionParams := curHook.LoadEnvVars.Execute[1:]
+				out, err := shell.RunShellCommandWithOutput(terragruntOptions, "", actionToExecute, actionParams...)
```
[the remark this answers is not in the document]
tmeckel: ad 1. maybe there are too many (verbose) logging messages still in the code. I added them to have more information during debugging. But I agree with you, that should be reconsidered.

ad 2: yes the method `RunShellCommandWithOutput` returns a struct which contains stdout adn stderr. But the parsing code only evaluates. stdout 
```go
scanner := bufio.NewScanner(strings.NewReader(out.Stdout))
for scanner.Scan() {
....
}
```

cli/cli_app.go
tmeckel: No according to the matrix I added to my initial description of the PR, you can combine the flags `transient` and `overwrite`.

cli/cli_app.go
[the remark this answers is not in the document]
tmeckel: As dicussed I restructured the code for parsing the environment variables coming from STDOUT

tmeckel:
@brikis98 I pushed a new version with the restructed parsing code for the incoming environment variables

config/config.go
tmeckel: This code has been removed in the current version of PR.

cli/cli_app.go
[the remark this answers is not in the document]
tmeckel: I once again went through the code which parses the incoming ENV var definition to ensure that no potential secrets are revealed by logging errors or by verbose logging.

cli/cli_app.go
```diff
@@ -264,6 +266,57 @@ func shouldPrintTerraformHelp(terragruntOptions *options.TerragruntOptions) bool
...
 }
 
+// loadEnvVars loads environment variable definitions from a shell command.
+// Currently the function expects KEY=value output to STDOUT
+// the method returns the old evironment variables map (dictionary) if the
+// changes should be transient.
+func loadEnvVars(terragruntOptions *options.TerragruntOptions, curHook config.Hook) (map[string]string, error) {
+	var oldEnvMap map[string]string
+
+	if curHook.LoadEnvVars != nil {
```
brikis98: To avoid a layer of nesting, you could make the first statement of this function:

```go
if curHook.LoadEnvVars == nil {
  return nil, nil
}
```

cli/cli_app.go
```diff
@@ -264,6 +266,57 @@ func shouldPrintTerraformHelp(terragruntOptions *options.TerragruntOptions) bool
...
+
+		actionToExecute := curHook.LoadEnvVars.Execute[0]
+		actionParams := curHook.LoadEnvVars.Execute[1:]
+		out, err := shell.RunShellCommandWithOutput(terragruntOptions, "", actionToExecute, actionParams...)
+		if err != nil {
+			terragruntOptions.Logger.Printf("Error running hook %s while loading environment variables with message: %s", curHook.Name, err.Error())
+			return nil, err
+		}
+		// parse out from script to map
+		m := make(map[string]string)
```
brikis98: Please use descriptive (not single letter) var names.

cli/cli_app.go
```diff
@@ -264,6 +266,57 @@ func shouldPrintTerraformHelp(terragruntOptions *options.TerragruntOptions) bool
...
+		scanner := bufio.NewScanner(strings.NewReader(out.Stdout))
+		for scanner.Scan() {
+			txt := scanner.Text()
+			idx := strings.Index(txt, "=")
+			if idx > 0 {
+				key := strings.TrimSpace(txt[0:idx])
+				value := txt[idx+1 : len(txt)]
+				m[key] = value
+			} else {
+				terragruntOptions.Logger.Print("Invalid key,value pair")
```
brikis98: Hm, this isn't much to go on when debugging. Can we specify the line number? And mention no `=` was found?

cli/cli_app.go
```diff
@@ -264,6 +266,57 @@ func shouldPrintTerraformHelp(terragruntOptions *options.TerragruntOptions) bool
...
+		}
+		// parse out from script to map
+		m := make(map[string]string)
+		scanner := bufio.NewScanner(strings.NewReader(out.Stdout))
+		for scanner.Scan() {
+			txt := scanner.Text()
+			idx := strings.Index(txt, "=")
+			if idx > 0 {
+				key := strings.TrimSpace(txt[0:idx])
+				value := txt[idx+1 : len(txt)]
```
brikis98: You can avoid some index arithmetic with `SplitN`:

```go
txt := scanner.Text()
parts := strings.SplitN(txt, "=", 2)
if len(parts) == 2 {
  key := parts[0]
  value := parts[1]
  // ... etc
```

cli/cli_app.go
```diff
@@ -264,6 +266,57 @@ func shouldPrintTerraformHelp(terragruntOptions *options.TerragruntOptions) bool
...
+			}
+		}
+		targetEnvMap := terragruntOptions.Env
+		if curHook.LoadEnvVars.Transient {
+			oldEnvMap = terragruntOptions.Env
+			targetEnvMap := make(map[string]string)
+			for key, value := range terragruntOptions.Env {
+				targetEnvMap[key] = value
+			}
+			terragruntOptions.Env = targetEnvMap
```
brikis98: Maybe I'm misreading this, but what is this code doing? It looks like `targetEnvMap` started with contents `terragruntOptions.Env`... Then you loop over `terragruntOptions.Env` and copy them again into `targetEnvMap`?

brikis98: Oh, also, I just realized you're shadowing the `targetEnvMap` variable. That makes this code even harder to follow. Please use a different variable name here.

cli/cli_app.go
```diff
@@ -264,6 +266,57 @@ func shouldPrintTerraformHelp(terragruntOptions *options.TerragruntOptions) bool
...
+				key := strings.TrimSpace(txt[0:idx])
+				value := txt[idx+1 : len(txt)]
+				m[key] = value
+			} else {
+				terragruntOptions.Logger.Print("Invalid key,value pair")
+			}
+		}
+		targetEnvMap := terragruntOptions.Env
+		if curHook.LoadEnvVars.Transient {
+			oldEnvMap = terragruntOptions.Env
```
brikis98: Will this preserve the vars correctly if you overwrite `terragruntOptions.Env` later?

cli/cli_app.go
```diff
@@ -264,6 +266,57 @@ func shouldPrintTerraformHelp(terragruntOptions *options.TerragruntOptions) bool
 	return false
 }
 
+// loadEnvVars loads environment variable definitions from a shell command.
+// Currently the function expects KEY=value output to STDOUT
+// the method returns the old evironment variables map (dictionary) if the
+// changes should be transient.
+func loadEnvVars(terragruntOptions *options.TerragruntOptions, curHook config.Hook) (map[string]string, error) {
```
brikis98: Having read through this whole method, it's trying to do many things, and may be hiding some bugs (see comments below for more context). I'd recommend breaking it down further:

1. One method that executes the LoadEnvVars hook and returns stdout as a string.
1. One method that, given a string, parses env vars from it, and returns those env vars as a map. 
1. One method that, given env vars as a map, returns a clone of `terragruntOptions` with the new env vars set.

You can then add unit tests for each of these items to make sure they work as intended.

cli/cli_app.go
```diff
@@ -277,15 +330,25 @@ func processHooks(hooks []config.Hook, terragruntOptions *options.TerragruntOpti
...
 			terragruntOptions.Logger.Printf("Executing hook: %s", curHook.Name)
-			actionToExecute := curHook.Execute[0]
-			actionParams := curHook.Execute[1:]
-			possibleError := shell.RunShellCommand(terragruntOptions, actionToExecute, actionParams...)
-
-			if possibleError != nil {
-				terragruntOptions.Logger.Printf("Error running hook %s with message: %s", curHook.Name, possibleError.Error())
-				errorsOccurred = append(errorsOccurred, possibleError)
+			oldEnvMap, err := loadEnvVars(terragruntOptions, curHook)
+			if nil != err {
```
brikis98: `err != nil`

cli/cli_app.go
```diff
@@ -264,6 +266,57 @@ func shouldPrintTerraformHelp(terragruntOptions *options.TerragruntOptions) bool
...
+			oldEnvMap = terragruntOptions.Env
+			targetEnvMap := make(map[string]string)
+			for key, value := range terragruntOptions.Env {
+				targetEnvMap[key] = value
+			}
+			terragruntOptions.Env = targetEnvMap
+		}
+		if curHook.LoadEnvVars.Overwrite {
+			targetEnvMap = make(map[string]string)
+			terragruntOptions.Env = targetEnvMap
```
brikis98: I'm not a fan of modifying `terragruntOptions` in place. It makes the code hard to follow, requires the user of this function to know to retain the old map of env vars and reset it later, and may cause issues in concurrent situations (e.g., if we have several goroutines using the same `terragruntOptions`). Instead, I'd recommend that this method returns a _new_ `terragruntOptions`, cloned from the old one, but with env vars overwritten. That way, there are no concurrency concerns, and the caller of the method can use the new `terragruntOptions` with no concern about having to reset anything in the old one.

cli/cli_app.go
brikis98: Understood, thanks. As mentioned in comments above, instead of mutating `terragruntOptions`, I think `loadEnvVars` should return a clone of it with new env vars set as necessary. That changes the flow in this method to:

```go
terragruntOptionsWithEnvVars, err := loadEnvVars(terragruntOptions, curHook)
if err != nil {
  // ...
}
if len(curHook.Execute) > 0 {
  actionToExecute := curHook.Execute[0]
  actionParams := curHook.Execute[1:]
  possibleError := shell.RunShellCommand(terragruntOptionsWithEnvVars, actionToExecute, actionParams...)
  // ...
}
```

cli/cli_app.go
```diff
@@ -338,6 +401,8 @@ func runTerragruntWithConfig(terragruntOptions *options.TerragruntOptions, terra
 		}
 	}
 
+	beforeHookErrors := processHooks(terragruntConfig.Terraform.GetBeforeHooks(), terragruntOptions)
```
brikis98: This is a fairly big behavioral change. Some use cases will want hooks to run before `init` and some after `init` but before the actual Terraform command. Therefore, the suggestion is to add a new `before_init_hook` type. This sort of change should be broken out into a separate PR.

config/config.go
```diff
@@ -111,7 +112,14 @@ func (conf *TerraformConfig) ValidateHooks() error {
 	allHooks := append(conf.GetBeforeHooks(), conf.GetAfterHooks()...)
 
 	for _, curHook := range allHooks {
-		if len(curHook.Execute) < 1 || curHook.Execute[0] == "" {
+		hasLoadEnvVars := false
```
brikis98: This var is extra. You can just check `curHook.LoadEnvVars` on line 122.

config/config.go
```diff
@@ -138,17 +146,6 @@ func (conf *TerraformExtraArguments) String() string {
 		conf.EnvVars)
 }
 
-// Return the default path to use for the Terragrunt configuration file. The reason this is a method rather than a
-// constant is that older versions of Terragrunt stored configuration in a different file. This method returns the
-// path to the old configuration format if such a file exists and the new format otherwise.
-func DefaultConfigPath(workingDir string) string {
```
brikis98: What was it that required moving this method to another package?

options/options.go
```diff
@@ -238,6 +260,21 @@ func (terragruntOptions *TerragruntOptions) AppendTerraformCliArgs(argsToAppend
...
+	var ret *models.IncludeConfig
+	ret = nil
+	if len(terragruntOptions.Context.Includes) > 0 {
+		ret = terragruntOptions.Context.Includes[0]
+	}
+	return ret
+}
+
+func (terragruntOptions *TerragruntOptions) AddIncludeConfig(includedConfig *models.IncludeConfig) {
+	if nil != includedConfig {
```
brikis98: `includedConfig != nil`

options/options.go
```diff
@@ -238,6 +260,21 @@ func (terragruntOptions *TerragruntOptions) AppendTerraformCliArgs(argsToAppend
 	terragruntOptions.TerraformCliArgs = append(terragruntOptions.TerraformCliArgs, argsToAppend...)
 }
 
+func (terragruntOptions *TerragruntOptions) GetIncludeConfig() *models.IncludeConfig {
+	var ret *models.IncludeConfig
+	ret = nil
+	if len(terragruntOptions.Context.Includes) > 0 {
+		ret = terragruntOptions.Context.Includes[0]
+	}
+	return ret
```
brikis98: Simplify method to:

```go
if len(terragruntOptions.Context.Includes) > 0 {
  return terragruntOptions.Context.Includes[0]
}
return nil
```

remote/remote_state.go
```diff
@@ -79,28 +81,46 @@ func (remoteState *RemoteState) NeedsInit(terragruntOptions *options.TerragruntO
...
 	if existingBackend.Type != remoteState.Backend {
 		terragruntOptions.Logger.Printf("Backend type has changed from %s to %s", existingBackend.Type, remoteState.Backend)
-		return true
+		return true, nil
 	}
 
-	if !reflect.DeepEqual(existingBackend.Config, remoteState.Config) {
-		terragruntOptions.Logger.Printf("Backend config has changed from %s to %s", existingBackend.Config, remoteState.Config)
-		return true
+	var pConfig = make(map[string]interface{}, len(remoteState.Config))
```
brikis98: What's the `p` in `pConfig`?

remote/remote_state.go
```diff
@@ -79,28 +81,46 @@ func (remoteState *RemoteState) NeedsInit(terragruntOptions *options.TerragruntO
...
+	include := terragruntOptions.GetIncludeConfig()
+	for key, val := range remoteState.Config {
+		if szval, ok := val.(string); ok {
+			pVal, err := shell.ResolveTerragruntConfigString(szval, include, terragruntOptions)
```
brikis98: Processing all interpolations in the `differsFrom` method feels like a big abstraction leak to me. I think we need to process them somewhere earlier, store the processed ones in `terragruntOptions`, and all other methods should not be required to know anything about interpolations.

shell/run_shell_cmd.go
```diff
@@ -42,17 +42,36 @@ func RunShellCommandWithOutput(terragruntOptions *options.TerragruntOptions, wor
...
+	 */
+	include := terragruntOptions.GetIncludeConfig()
+	argsParsed := make([]string, len(args))
+	for i := 0; i < len(args); i++ {
+		value, err := ResolveTerragruntConfigString(args[i], include, terragruntOptions)
+		if err != nil {
+			return nil, errors.WithStackTrace(err)
+		}
+		argsParsed[i] = value
+	}
```
brikis98: This again feels like an abstraction leak... I understand you're trying to do the binding just before we execute the command, but `RunTerraformCommandWithOutput` is a general purpose method that should know nothing of interpolations. I think interpolation processing needs to happen earlier, in some wrapper method.

shell/run_shell_cmd.go
```diff
@@ -86,12 +105,17 @@ func RunShellCommandWithOutput(terragruntOptions *options.TerragruntOptions, wor
...
 	for key, value := range envVarsAsMap {
-		envVarsAsList = append(envVarsAsList, fmt.Sprintf("%s=%s", key, value))
+		pvalue, err := ResolveTerragruntConfigString(value, include, terragruntOptions)
+		if err != nil {
+			return nil, err
+		}
+		envVarsAsList = append(envVarsAsList, fmt.Sprintf("%s=%s", key, pvalue))
 	}
-	return envVarsAsList
+	return envVarsAsList, nil
```
brikis98: Again, same thing here. This method should not know anything about interpolations. All the interpolation process should happen in a single location before the `shell` package is ever called.

shell/config_helpers.go
```diff
@@ -1,17 +1,16 @@
...
 	"regexp"
 	"strings"
 
-	"github.com/gruntwork-io/terragrunt/shell"
-
 	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/gruntwork-io/terragrunt/errors"
+	. "github.com/gruntwork-io/terragrunt/models"
```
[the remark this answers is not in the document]
tmeckel: I needed to refactor some type declarations and move them to new or other packages, thus I had to add new import declarations. The refactoring was required because the GO compiler detected loops in the imports and in contrast to other languages GO is not able to resolve such loops by itself. To limit the required code changes inside the affected code files I added the period in front of the new import statements, because according to the go language specification

> "If an explicit period (.) appears instead of a name, all the package's exported identifiers declared in that package's package block will be declared in the importing source file's file block and must be accessed without a qualifier."

https://golang.org/ref/spec#Import_declarations

shell/config_helpers_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	"github.com/gruntwork-io/terragrunt/errors"
+	. "github.com/gruntwork-io/terragrunt/models"
```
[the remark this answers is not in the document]
tmeckel: Refer to my comment above

cli/cli_app.go
tmeckel: Okay

cli/cli_app.go
tmeckel: I agree ... not the best way to declare variables :-D

cli/cli_app.go
tmeckel: I had the same thoughts on this when I cleaned the logging. Adding a line number would be helpful but that implies that the called script or program which is used to generate the env vars must output them in a deterministic and constant way i.e. every call to the script should output the env vars in the same sequence on STDOUT. Otherwise the line number isn't very helpful. But I agree better than nothing at all.

cli/cli_app.go
tmeckel: Neat

cli/cli_app.go
tmeckel: Okay ... I will change the variable name here.

cli/cli_app.go
tmeckel: Normally I prefer to have the constant always on the left side so that the compiler will create an error when I miss the second '=' in an equality comparison and make it up into an assignment.

https://en.wikipedia.org/wiki/Yoda_conditions

cli/cli_app.go
tmeckel: The argument about potential concurrency issues is a good one and convinced me. Although I'M not a big fan of bloating the memory with most of the useless copies. Let's hope the GC of GO is fast and clever enough :-D

cli/cli_app.go
tmeckel: Sound reasonable and I guess that's why most systems do have a PreInit step or the like :-D

config/config.go
tmeckel: Okay

config/config.go
tmeckel: As far as I remember to GO compiler complaining about an import loop.

options/options.go
tmeckel: Refer to my comment about Yoda conditions above

options/options.go
tmeckel: I'm trying to limit the number of return statements to a minimum. It's the ongoing war on the principle Dijkstra formulated back in the 70ties I think. 

".. every function, and every block within a function, should have one entry and one exit. "

If the multiple return statements principle is the way of coding for terragrunt this can easily be changed.  

https://hackerchick.com/religious-war-48293-single-vs-multiple/

remote/remote_state.go
tmeckel: My tendency to still prefix the variable name with some type information. I'm not consistent .. I know .. So the 'p' can be removed.

http://web.mst.edu/~cpp/common/hungarian.html

shell/run_shell_cmd.go
tmeckel: Okay ... no I understood your two complains here .. I was a bit confused about what you meant about the 'abstraction leak'. As I said when I filed the PR we changed the code of terraform in the customer project with the most minimal effort possible. There was not much time to think about abstraction layers.

shell/run_shell_cmd.go
tmeckel: I doubt that I understood the code of terragrunt that thoroughly so that I could be the one to add new abstraction layers here. Perhaps you can come up with a proposal here.

cli/cli_app.go
brikis98: I'm familiar with yoda conditions but they aren't necessary in Go. The following code won't even compile:

```go
if err = nil {
  // ...
}
```

And since yoda conditions are a bit weird to read, no need for them here.

cli/cli_app.go
brikis98: Perf is not going to be a concern at all in a tool like Terragrunt. Even if it was, we'd still optimize for correctness/maintenance/readability/ first, then find the bottlenecks, and optimize solely the code relevant to those.

config/config.go
brikis98: Right, I'm just trying to understand which import loop in case there's a way to avoid extra packages, extra refactoring, etc.

options/options.go
brikis98: Same as above

options/options.go
brikis98: I prefer the functional programming approach: use immutable values whenever possible and return them whenever. Multiple returns are hard to follow with lots of mutation, but straightforward otherwise. IMO, the version with two return statements conveys the intention of the code more clearly and succinctly.

remote/remote_state.go
brikis98: Nothing wrong with that, but I don't know what the `p` would even stand for in this case? Perhaps spell it out more?

shell/run_shell_cmd.go
brikis98: Well, to make a major changes to how interpolations work and when they are processed, I think understanding the flow fully is a requirement.

The way I'd approach it is to look at the places these `run_shell_cmd.go` functions are used. Is there a common path they are going through? What's the order of events? Running init, running before hooks, executing Terraform, running after hooks, etc. When are interpolations  processed now (i.e., before any of those events)? And when do you want them to be processed (i.e., at each event?).

tmeckel:
Hi @brikis98 ... Sorry I've to close this PR because currently I personally have no time to work on this PR anymore and we found a way to utilize a plain version of terragrunt in our project.

PR #702, opened 2019-05-13
Windows CI Integration

tonerdo:
This PR sets up a Windows CI for Terragrunt using Azure Pipelines (replacing AppVeyor). 

We're able to split into unit and integration test jobs
Screenshot 2019-05-13 at 10 06 59 PM

We can also publish XML test results generated by the Terratest log parser
Screenshot 2019-05-13 at 10 10 16 PM

Unlike AppVeyor (and like CircleCI) we can have organizations
Screenshot 2019-05-13 at 10 13 47 PM

Some tests are failing of course, but the point of this PR is to get the Windows CI infrastructure setup


azure-pipelines.yml
```diff
@@ -0,0 +1,72 @@
...
+- script: choco install terraform --version $(TERRAFORMVERSION)
+  displayName: 'Install Terraform'
+
+- script: dep ensure
+  workingDirectory: '$(modulePath)'
+  displayName: 'Run `dep ensure`'
+
+- pwsh: |
+    if ("$(TESTKIND)" -eq "unit") {
+      go test -v -parallel 1 $(go list ./... | where-object { $_ -notlike "*/test" }) | Tee-Object -FilePath 'logs/unit.log'
```
brikis98: I think we normally use `run-go-test` for this, which sets other options, such as `-timeout`. Can you make sure we re-use those here?

Also, perhaps a dumb question, but is there no way we can reuse any of those Bash scripts here? I'm curious if we can reduce duplication at all between Windows and *nix.

azure-pipelines.yml
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    if ("$(TESTKIND)" -eq "integration") {
+      go test -v $(go list ./... | where-object { $_ -like "*/test" }) | Tee-Object -FilePath 'logs/integration.log'
```
brikis98: Same here

azure-pipelines.yml
```diff
@@ -0,0 +1,72 @@
...
+
+- script: choco install terraform --version $(TERRAFORMVERSION)
+  displayName: 'Install Terraform'
+
+- script: dep ensure
+  workingDirectory: '$(modulePath)'
+  displayName: 'Run `dep ensure`'
+
+- pwsh: |
+    if ("$(TESTKIND)" -eq "unit") {
```
brikis98: What language is this?

azure-pipelines.yml
```diff
@@ -0,0 +1,72 @@
...
+      go test -v -parallel 1 $(go list ./... | where-object { $_ -notlike "*/test" }) | Tee-Object -FilePath 'logs/unit.log'
+    }
+
+    if ("$(TESTKIND)" -eq "integration") {
+      go test -v $(go list ./... | where-object { $_ -like "*/test" }) | Tee-Object -FilePath 'logs/integration.log'
+    }
+  workingDirectory: '$(modulePath)'
+  displayName: 'Run tests'
+  continueOnError: true
```
brikis98: Why?

azure-pipelines.yml
```diff
@@ -0,0 +1,72 @@
...
+- pwsh: |
+    if ("$(TESTKIND)" -eq "unit") {
+      $(system.defaultWorkingDirectory)\terratest_log_parser.exe --testlog logs/unit.log --outputdir ./logs/unit/
+    }
+
+    if ("$(TESTKIND)" -eq "integration") {
+      $(system.defaultWorkingDirectory)\terratest_log_parser.exe --testlog logs/integration.log --outputdir ./logs/integration/
+    }
+  workingDirectory: '$(modulePath)'
+  displayName: 'Generate Test Report'
```
brikis98: Nice!

azure-pipelines.yml
tonerdo: Powershell

azure-pipelines.yml
tonerdo: We need this here so that the task that publishes the xml test results will run. The entire build pipeline is still marked as failed so it doesn't change much

azure-pipelines.yml
tonerdo: I can run it on "Bash for Windows" (https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema#bash). The problem with this approach is, the tests will still be running in the linux subsystem and not on Windows directly. That negates the whole point of setting up a Windows CI

brikis98: Ah, OK, fair enough

azure-pipelines.yml
brikis98: Roger

tonerdo:
@brikis98 I invited you to the Gruntwork Azure DevOps org. Let me know if you have any trouble

azure-pipelines.yml
yorinasub17: Should we use python for our CI scripts once we have windows CI on track, to avoid the code duplication?

brikis98: Yea, probably. I mean, that still requires Python to be installed, and there will still be other OS differences to paper over (e.g., path lengths), but that should help.

yorinasub17: At least it isn't "share no code" 😄

tonerdo:
Just checking if this is something you guys are still interested in. So I can properly handover ownership of the Azure DevOps Gruntwork organization

brikis98:
Yes, thanks for checking @tonerdo! I've just been on vacation and haven't had time to dig into this stuff yet. And now #466 is a high priority, so I need to wrap that up, and will then come back to this.

azure-pipelines.yml
azilber: Why not Golang?  Installing Python is going to be a mess, someone's bound to install something and complain...

azilber:
Any update on this?  Building out a pipeline and wondering; when this is going to be merged?

PR #705, opened 2019-05-15
New interpolation function for parallelism

cschroer:
We do run have a huge deployment with GCP. To avoid any issues with GCP APIs we need to trottle each terraform command invoked.

New function `get_terraform_commands_that_need_parallelism` returns all terraform commands that support -parallelism


yorinasub17:
Hi @cschroer , just to clarify, is this PR ready for review? I am asking because it saids WIP at the top, so just wanted to make sure.

If still WIP, can you remove the reviewers so we know this is still WIP, and when ready, add them back in so that we get notified? Thanks!

cschroer:
@yorinasub17 my faul. Added a "WIP" by habit even to this small PR  🤦‍♂

PR #707, opened 2019-05-17
enables multiple includes

### LINKED ISSUE #303: Only one level of includes is allowed
I'm trying to enable a multi-include scenario where I can cascade includes at different points in my configurations. Given the below layout:
```
└── root
    ├── regions
    │   ├── us-east-1
    │   │   ├── regional_deployment(lambda)
    │   │   │      └── terraform.tfvars
    │   │   ├── config.tfvars
    │   │   └── terraform.tfvars
    │   └── us-east-2
    │       ├── regional_deployment(lambda)
    │       │      └── terraform.tfvars
    │       ├── config.tfvars
    │       └── terraform.tfvars
    ├── config.tfvars
    └── terraform.tfvars
 ```
I would like to be able to set the `remote_state` in the  root `terraform.tfvars` while allowing for the notion of regional configurations. Some deployments are regional by nature (aws lambdas as an example) and so I think it makes sense to allow for this kind of modeling.

Right now I'm adding a separate `extra_arguments` block in each regional deployment's `terraform.tfvars` that includes `required_var_files = ["${get_tfvars_dir()}/../config.tfvars]"` . It works but it's not very DRY. Being able to have the regional `terraform.tfvars` leverage an `include` block to roll up to its parent would be even more so.

twang817:
Allows for multiple includes.  Should fix #303.

`path_relative_from_include()` and `path_relative_to_include` include tested.  Behavior should be to evaluate within the context of the included file.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -301,20 +301,16 @@ func parseConfigString(configString string, terragruntOptions *options.Terragrun
 		return nil, err
 	}
 
-	if include != nil && terragruntConfigFile.Include != nil {
-		return nil, errors.WithStackTrace(TooManyLevelsOfInheritance{
-			ConfigPath:             terragruntOptions.TerragruntConfigPath,
-			FirstLevelIncludePath:  include.Path,
-			SecondLevelIncludePath: terragruntConfigFile.Include.Path,
-		})
-	}
-
 	includedConfig, err := parseIncludedConfig(terragruntConfigFile.Include, terragruntOptions)
 	if err != nil {
 		return nil, err
 	}
 
-	return mergeConfigWithIncludedConfig(config, includedConfig, terragruntOptions)
+	merged, err := mergeConfigWithIncludedConfig(config, includedConfig, terragruntOptions)
+	if includedConfig == nil {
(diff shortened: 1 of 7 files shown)


brikis98:
Thanks for the PR! I'm a bit buried now and will review this later on to think through the implications.

apottere:
How does this work with `*-all` commands?  Is there a way to either mark files as intermediary or mark them as leaves?

brikis98:
@twang817 When you get a chance, please rebased this PR on `master`. Note that in `v0.19.0`, Terragrunt config parsing had a major refactor to support Terraform 0.12 and move to HCL2.

icereed:
Would be happy about this feature!

asafha-etorox:
Until we get the `imports` block, I would be majorly super-duper happy to see it revisited!

maulik-vimond:
It would be great to get this one merged. We want the child modules to include some extra files (that generate TF config) and that is not possible at present.

erdnaxeli:
Hi, what is it blocking to see this merged?

notwaf:
can we get this merged in? been over a year :)

PR #714, opened 2019-05-27
Add note about TF 0.12 upgrade

brikis98:


brikis98:
Thanks for the review!

PR #722, opened 2019-06-03
WIP: Improve *-all error message output

msvechla:
Hi all, 

Thanks for this awesome project, we are using it in production with 50+ AWS accounts in our two organizations every day!

This PR aims to improve the readability of the `*-all` error messages. With our huge codebases, it can get really hard to find a small module, which errors out during a big `apply-all` run. The PR tries to improve this, by printing the root-cause error messages at the end of the execution run. Root-cause in this case means all module errors, excluding dependency errors.

Please let me know if you have any suggestions to improve this further.


msvechla:
I know using global variables is probably not the best way to achieve this, but I tried it earlier by incorporating the variables into the `Stack` struct, but this made everything way more complicated, so I decided to go for readability.

Also the go channel is currently not really needed, but might improve readability and future refactoring, as the concurrency is now explicitly coded.

Do you have any suggestions for unit tests?

brikis98:
Nice, thanks for the PR! Could you share an example of what the log output will look like now (a small snippet, not the whole thing, of course)?

Please note that we're going to hold off on merging anything until #466 is resolved, as that's very high priority. Once that one is in, please pull the latest from master, and give us a ping to review.

msvechla:
@brikis98 I just rebased from master and pushed again.

The output currently looks like this: At the end of the exuction of an `*-all` command, we see the following summary when errors occurred. The output with terraform > v0.12 is a little bit more verbose now, as it prints an additional warning:

> Warning: Skipping backend initialization pending configuration upgrade

```
[terragrunt] 2019/06/10 22:50:20 Encountered the following root-causes:
------------------------------------------------------------------------------------------------------------------------------------
Module /Users/msvechl/go/src/github.com/gruntwork-io/terragrunt/test/fixture-auto-retry/apply-all/app1:
[terragrunt] [/Users/msvechl/go/src/github.com/gruntwork-io/terragrunt/test/fixture-auto-retry/apply-all/app1] 2019/06/10 22:50:20 Running command: terraform init

Warning: Skipping backend initialization pending configuration upgrade

The root module configuration contains errors that may be fixed by running the
configuration upgrade tool, so Terraform is skipping backend initialization.
See below for more information.


Terraform has initialized, but configuration upgrades may be needed.

Terraform found syntax errors in the configuration that prevented full
initialization. If you've recently upgraded to Terraform v0.12, this may be
because your configuration uses syntax constructs that are no longer valid,
and so must be updated before full initialization is possible.

Terraform has installed the required providers to support the configuration
upgrade process. To begin upgrading your configuration, run the following:
    terraform 0.12upgrade

To see the full set of errors that led to this message, run:
    terraform validate

Error: Unsupported block type

  on main.tf line 1:
   1: outputa "app1_text" {

Blocks of type "outputa" are not expected here. Did you mean "output"?


------------------------------------------------------------------------------------------------------------------------------------
Module /Users/msvechl/go/src/github.com/gruntwork-io/terragrunt/test/fixture-auto-retry/apply-all/app2:
[terragrunt] [/Users/msvechl/go/src/github.com/gruntwork-io/terragrunt/test/fixture-auto-retry/apply-all/app2] 2019/06/10 22:50:20 Running command: terraform init

Warning: Skipping backend initialization pending configuration upgrade

The root module configuration contains errors that may be fixed by running the
configuration upgrade tool, so Terraform is skipping backend initialization.
See below for more information.


Terraform has initialized, but configuration upgrades may be needed.

Terraform found syntax errors in the configuration that prevented full
initialization. If you've recently upgraded to Terraform v0.12, this may be
because your configuration uses syntax constructs that are no longer valid,
and so must be updated before full initialization is possible.

Terraform has installed the required providers to support the configuration
upgrade process. To begin upgrading your configuration, run the following:
    terraform 0.12upgrade

To see the full set of errors that led to this message, run:
    terraform validate

Error: Unsupported block type

  on main.tf line 1:
   1: outputwas "app2_text" {

Blocks of type "outputwas" are not expected here.

 
[terragrunt] 2019/06/10 22:50:20 Encountered the following errors:
Hit multiple errors:
exit status 1
Hit multiple errors:
exit status 1
```

Still, having all errors and the related modules printed at the end of the runner is a great benefit when running large executions. Please let me know what you think.

msvechla:
Thanks for the feedback @brikis98! 

I did some refactoring and removed the channels. For sure I will add tests once this last bit is resolved. 

Do you have any idea why I still get the `terraform init` output, even though I only print the stderr stream at the end?

brikis98:
> Do you have any idea why I still get the terraform init output, even though I only print the stderr stream at the end? As far as I know the terraform init messages should be on the stdout stream.

Not sure I follow. You seem to be getting an error about a missing variable. What does stdout or stderr have to do with it?

msvechla:
What I posted is the new detailed output of my change. The idea is to have a summary of all module errors including their error messages (stderr) at the end of the execution. If you check out https://github.com/gruntwork-io/terragrunt/pull/722/files#diff-86e77ee353cd3bacb4a1f0c492bf9e2cR169 of my change, you can see that I am capturing the stderr and outputting it in the `collectErrors()` method: https://github.com/gruntwork-io/terragrunt/pull/722/files#diff-86e77ee353cd3bacb4a1f0c492bf9e2cR186.

Somehow the `terraform init` code shows up in the `stderr` stream, even though when I do a normal terraform run, it is printed to `stdout`.

So my question would be if you have any idea why the `terraform init` output shows up in `stderr.

brikis98:
Ohhhh, I gotcha, thx for providing the context 😁 

The behavior you're seeing is probably from this: https://github.com/gruntwork-io/terragrunt/blob/master/cli/cli_app.go#L606-L607

msvechla:
Yep, thats it.

In the comment [here](https://github.com/gruntwork-io/terragrunt/blob/master/cli/cli_app.go#L606-L607) it says:

> Don't pollute stdout with the stdout from Auto Init

So I assume it will not be a big issue moving this back to stdout? Of course I will run the tests to make sure.

Also I found [this part](https://github.com/gruntwork-io/terragrunt/pull/722/commits/e0aafd615c63f5a35c955f253db1386808b08c54#diff-f95053933eada5dadd7191beb755d841L196) where the logger is set by default to stderr, which also leads to some pollution of my error output.

I now adjusted both parts and now the detailed error output is clean. Can you think of any issues these changes could cause from the top of your head?

brikis98:
> Can you think of any issues these changes could cause from the top of your head?

Yes. Consider someone running the following:

```
url=$(terragrunt output url)
```

They expect that the value of the output variable `url`, and _only_ that value, is written to `stdout`. If the auto init functionality writes to `stdout`, then that assumption will break. Hence, we redirect auto init output to `stderr`. 

In general, if you run `terragrunt <cmd>`, where `<cmd>` is any standard Terraform command, what's written to `stdout` should be the same as if you had run `terraform <cmd>` directly.

msvechla:
Thanks for clarifying, of course that makes perfect sense.

In this case we would either have to live with the more verbose detailed error message at the end of the execution, or I would have to come up with some way of extracting the auto-init output from the detailed error message.

I will look into it again.

msvechla:
Alright, I made it work by saving the `auto-init` output and extracting it from the detailed error messages.

The current detailed error messages now look like this:

```
[terragrunt] 2019/07/02 23:46:37 Encountered the following errors:
====================================================================================================================================
/Users/msvechl/go/src/github.com/gruntwork-io/terragrunt/test/fixture-auto-retry/apply-all/app1 (root error): 

Hit multiple errors:
exit status 1 

[terragrunt] [/Users/msvechl/go/src/github.com/gruntwork-io/terragrunt/test/fixture-auto-retry/apply-all/app1] 2019/07/02 23:46:37 Running command: terraform init

Error: Reference to undeclared input variable

  on main.tf line 2, in output "app1_text":
   2:   value = "app1 output ${var.aasd}"

An input variable with the name "aasd" has not been declared. This variable
can be declared with a variable "aasd" {} block.


------------------------------------------------------------------------------------------------------------------------------------
/Users/msvechl/go/src/github.com/gruntwork-io/terragrunt/test/fixture-auto-retry/apply-all/app2 (dependency error): 

Cannot process module Module /Users/msvechl/go/src/github.com/gruntwork-io/terragrunt/test/fixture-auto-retry/apply-all/app2 (excluded: false, dependencies: [/Users/msvechl/go/src/github.com/gruntwork-io/terragrunt/test/fixture-auto-retry/apply-all/app3, /Users/msvechl/go/src/github.com/gruntwork-io/terragrunt/test/fixture-auto-retry/apply-all/app1]) because one of its dependencies, Module /Users/msvechl/go/src/github.com/gruntwork-io/terragrunt/test/fixture-auto-retry/apply-all/app1 (excluded: false, dependencies: [/Users/msvechl/go/src/github.com/gruntwork-io/terragrunt/test/fixture-auto-retry/apply-all/app3]), finished with an error: Hit multiple errors:
exit status 1 


------------------------------------------------------------------------------------------------------------------------------------
[terragrunt] 2019/07/02 23:46:37 Unable to determine underlying exit code, so Terragrunt will exit with error code 1 
```

What do you think? Should we also make this output optional or use it by default?

brikis98:
So that's the output you see at the end of an `xxx-all` call?

If so, I think that looks like a terrific improvement. I assume those errors are grouped by module?

msvechla:
Yep, that is correct. As the `collectErrrors()` method loops over all running modules, this automatically groups all the errors by module.

I will try and add some tests now.

msvechla:
Alright, I just added module and integration tests. Do you have further improvement suggestions?

msvechla:
I did some refactoring to separate the normal and detailed errors in the `MultiError` struct. This also fixed problems with some of the tests. Do you have any further input @brikis98?

brikis98:
Apologies for the delay! Been completely buried. I really appreciate this PR and will try to review this as soon as I can. 🍺

msvechla:
Don't worry, take your time! Just get back to me when you can 👍

configstack/running_module.go
```diff
@@ -233,6 +270,8 @@ func (module *runningModule) moduleFinished(moduleErr error) {
 		module.Module.TerragruntOptions.Logger.Printf("Module %s has finished with an error: %v", module.Module.Path, moduleErr)
 	}
 
+	fmt.Fprintf(module.Writer, "%s\n%v\n\n%v\n", OutputMessageSeparator, module.Module.Path, module.OutStream.String())
```
brikis98: Why `fmt.Fprintf` here?

options/options.go
```diff
@@ -80,6 +81,9 @@ type TerragruntOptions struct {
 	// If you want stderr to go somewhere other than os.stderr
 	ErrWriter io.Writer
 
+	// Stores output of auto-init so it can be removed later form other streams
```
brikis98:
```suggestion
	// Stores output of auto-init so it can be removed later from other streams
```

configstack/running_module.go
```diff
@@ -154,6 +168,8 @@ func runModules(modules map[string]*runningModule) error {
 		waitGroup.Add(1)
 		go func(module *runningModule) {
 			defer waitGroup.Done()
+			module.Module.TerragruntOptions.ErrWriter = io.MultiWriter(&module.OutStream, &module.ErrStream)
+			module.Module.TerragruntOptions.Writer = &module.OutStream
```
brikis98: I'm a bit confused by this... You're overwriting `TerragruntOptions.ErrWriter` to write to both `module.OutStream` and `module.ErrStream`... But what was the `Terragrunt.LOptions.ErrWriter` value set to before that? Are `module.OutStream` and `module.ErrStream` initialized to anything? Will this buffer those errors until the very end or stream to `stdout` / `stderr`?

Same questions go for `TerragruntOptions.Writer`, with the additional one of what happens when you point a second item to `module.OutStream`?

configstack/running_module.go
```diff
@@ -167,17 +183,37 @@ func runModules(modules map[string]*runningModule) error {
...
 		return nil
+	}
+
+	return errors.WithStackTrace(MultiError{Errors: errs, DetailedErrors: detailedErrs})
+}
+
+// generateDetailedErrorMessage extracts the clean stderr from a module and formats it for printing
+func generateDetailedErrorMessage(module *runningModule) error {
+	// remove the auto-init pollution from the error stream
+	cleanErrorOutput := strings.Replace(module.ErrStream.String(), module.Module.TerragruntOptions.InitStream.String(), "", -1)
```
brikis98: Hm, string replacement feels... A bit hacky. Is there any way to have the AutoInit write to `stdout` / `stderr`, but to _not_ write to `module.ErrStream`? That way, you could use `module.ErrStream` directly, without having to clean anything out...

options/options.go
```diff
@@ -80,6 +81,9 @@ type TerragruntOptions struct {
...
+	InitStream bytes.Buffer
```
brikis98: Perhaps there should be an `AutoInitWriter` and `AutoInitErrWriter` here instead? By default, these would be set to the same as `Writer` and `ErrWriter`. However, for `xxx-all` commands, these could be set to separate values that write to `stdout` / `stderr` but are not stored in those streams later used to print clean error messages?

Also, should this value be in the `Clone` method? If not, explicitly add a comment there explaining why.

PR #730, opened 2019-06-06
improve string escape handling for run_cmd

### LINKED ISSUE #684: function interpolation in run_cmd is inconsistent
This was originally just an issue with functions not being interpolated for `run_cmd` all the time, but in making and MVP to demonstrate I also found a self-contradictory error message.

`terraform.tfvars`
```hcl
terragrunt = {
  extra_arguments "it dern borkedtouch" {

    arguments = [
      "-var",
      "alone=${run_cmd("echo", "-n", "${get_tfvars_dir()}")}",
      "-var",
      "before=${run_cmd("echo", "-n", "something_before/${get_tfvars_dir()}")}",
      "-var",
      "after=${run_cmd("echo", "-n", "${get_tfvars_dir()}/something_after")}",
    ]
  }
}

```

`main.tf`
```hcl
```

aaaand the log:
```
.../test ❯❯❯ terragrunt plan
[terragrunt] [/.../test] 2019/03/19 18:39:09 Running command: terraform --version
[terragrunt] 2019/03/19 18:39:09 Reading Terragrunt config file at /.../test/terraform.tfvars
[terragrunt] 2019/03/19 18:39:09 Running command: echo -n /.../test
/.../test[terragrunt] 2019/03/19 18:39:09 run_cmd output: [/.../test]
[terragrunt] 2019/03/19 18:39:09 Running command: echo -n something_before/${get_tfvars_dir()}
something_before/${get_tfvars_dir()}[terragrunt] 2019/03/19 18:39:09 run_cmd output: [something_before/${get_tfvars_dir()}]
[terragrunt] 2019/03/19 18:39:09 Running command: echo -n ${get_tfvars_dir()}/something_after
${get_tfvars_dir()}/something_after[terragrunt] 2019/03/19 18:39:09 run_cmd output: [${get_tfvars_dir()}/something_after]
[terragrunt] 2019/03/19 18:39:09 Invalid interpolation syntax. Expected syntax of the form '${function_name()}', but got '${get_tfvars_dir()}, ${get_tfvars_dir()}'
[terragrunt] 2019/03/19 18:39:09 Unable to determine underlying exit code, so Terragrunt will exit with error code 1
.../test ❯❯❯
```

Two things of note (only one is personally causing me problems):
1. observe that "alone" interpolates fine (apologies for the sensored directories); however "before" and "after" do not interpolate before running the command:
```
Running command: echo -n /.../test
Running command: echo -n something_before/${get_tfvars_dir()}
Running command: echo -n ${get_tfvars_dir()}/something_after
```

2. That final error message seems self-contradictory:
```
Invalid interpolation syntax. Expected syntax of the form '${function_name()}', but got '${get_tfvars_dir()}, ${get_tfvars_dir()}'
```

the "but got" is a result of two outputs containing interpolation, if only one command is run it looks even more confusing:
```
[terragrunt] 2019/03/19 18:50:32 Invalid interpolation syntax. Expected syntax of the form '${function_name()}', but got '${get_tfvars_dir()}'
```

twang817:
While #684 addresses the real issues here, I made a small change for better handling of escapes in run_cmd.

This allowed me to use run_cmd in the following way:

```
$ tree .
.
├── child
│   ├── config.json
│   └── terraform.tfvars
├── plans
│   └── testplan
│       ├── v0.1
│       │   └── main.tf
│       └── v0.2
│           └── main.tf
└── terraform.tfvars
```

The child's terraform.tfvars is nothing special
```
$ cat child/terraform.tfvars
terragrunt {
  include {
    path = "${find_in_parent_folders}
```

The config.json specifies the data the run_cmd fetches:
```
$ cat child/config.json
{
    "source": "testplan",
    "version": "v0.1"
}
```

The two plans are simple:
```
$ cat plans/testplan/v0.1/main.tf
output "version" {
    value = "v0.1"
}
$ cat plans/testplan/v0.2/main.tf
output "version" {
    value = "v0.2"
}
```

Finally, the magic is in the root terraform.tfvars:
```
$ cat terraform.tfvars
terragrunt {
    terraform {
        source = "${path_relative_from_include()}/plans//${run_cmd("jq", "-j", ". | \"\\(.source)/\\(.version)\"", "config.json")}/"
    }
}
```

If you're unfamiliar with jq, the command is:

```
$ jq -j '. | "\(.source)/\(.version)" config.json
testplan/v0.1
```

This basically works by having run_cmd extract the source & version key from config.json and build a path starting from the plans// directory.

I believe this resolves #684, as I have both prefixes and suffixes around run_cmd.  It further allows better handling of escaped quotes within the run_cmd parameters.

The PR is a bit dense with regex, so I'll try to comment inline.

Lastly, I was able to run all the tests (and even added a few) in the config directory, but I was unable to pass all tests from the root.  I believe there are several tests that actually rely on AWS -- and I am unable to do so.

### DIFF
--- a/config/config_helpers.go
+++ b/config/config_helpers.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/gruntwork-io/terragrunt/shell"
@@ -16,12 +17,13 @@ import (
 	"github.com/gruntwork-io/terragrunt/util"
 )
 
-var INTERPOLATION_PARAMETERS = `(\s*"[^"]*?"\s*,?\s*)*`
-var INTERPOLATION_SYNTAX_REGEX = regexp.MustCompile(fmt.Sprintf(`\$\{\s*\w+\(%s\)\s*\}`, INTERPOLATION_PARAMETERS))
+var INTERPOLATION_QUOTED_STRING = `"(?:\\.|[^\\"])*?"`
+var INTERPOLATION_QUOTED_REGEX = regexp.MustCompile(INTERPOLATION_QUOTED_STRING)
+var INTERPOLATION_PARAMETERS = fmt.Sprintf(`\s*(?:%s(?:\s*,\s*%s)*)?\s*`, INTERPOLATION_QUOTED_STRING, INTERPOLATION_QUOTED_STRING)
+var INTERPOLATION_PARAMETERS_REGEX = regexp.MustCompile(fmt.Sprintf(`^%s$`, INTERPOLATION_PARAMETERS))
+var INTERPOLATION_SYNTAX_REGEX = regexp.MustCompile(fmt.Sprintf(`\$\{\s*(\w+)\((%s)\)\s*\}`, INTERPOLATION_PARAMETERS))
 var INTERPOLATION_SYNTAX_REGEX_SINGLE = regexp.MustCompile(fmt.Sprintf(`"(%s)"`, INTERPOLATION_SYNTAX_REGEX))
 var INTERPOLATION_SYNTAX_REGEX_REMAINING = regexp.MustCompile(`\$\{.*?\}`)
-var HELPER_FUNCTION_SYNTAX_REGEX = regexp.MustCompile(`^\$\{\s*(.*?)\((.*?)\)\s*\}$`)
-var HELPER_FUNCTION_GET_ENV_PARAMETERS_SYNTAX_REGEX = regexp.MustCompile(`^\s*"(?P<env>[^=]+?)"\s*\,\s*"(?P<default>.*?)"\s*$`)
 
 // List of terraform commands that accept -lock-timeout
 var TERRAFORM_COMMANDS_NEED_LOCKING = []string{
@@ -168,7 +170,7 @@ func processMultipleInterpolationsInString(terragruntConfigString string, includ
 // Given a string value from a Terragrunt configuration, parse the string, resolve any calls to helper functions using
 // Resolve a single call to an interpolation function of the format ${some_function()} in a Terragrunt configuration
 func resolveTerragruntInterpolation(str string, include *IncludeConfig, terragruntOptions *options.TerragruntOptions) (interface{}, error) {
-	matches := HELPER_FUNCTION_SYNTAX_REGEX.FindStringSubmatch(str)
+	matches := INTERPOLATION_SYNTAX_REGEX.FindStringSubmatch(str)
 	if len(matches) == 3 {
 		return executeTerragruntHelperFunction(matches[1], matches[2], include, terragruntOptions)
 	} else {
@@ -204,20 +206,14 @@ func getParentTfVarsDir(include *IncludeConfig, terragruntOptions *options.Terra
 
 func parseGetEnvParameters(parameters string) (EnvVar, error) {
 	envVariable := EnvVar{}
-	matches := HELPER_FUNCTION_GET_ENV_PARAMETERS_SYNTAX_REGEX.FindStringSubmatch(parameters)
-	if len(matches) < 2 {
+	parsed := parseParamList(parameters)
+
+	if len(parsed) != 2 || len(parsed[0]) == 0 {
 		return envVariable, errors.WithStackTrace(InvalidGetEnvParams(parameters))
 	}
 
-	for index, name := range HELPER_FUNCTION_GET_ENV_PARAMETERS_SYNTAX_REGEX.SubexpNames() {
-		if name == "env" {
-			envVariable.Name = strings.TrimSpace(matches[index])
-		}
-		if name == "default" {
-			envVariable.DefaultValue = strings.TrimSpace(matches[index])
-		}
-	}
-
+	envVariable.Name = parsed[0]
+	envVariable.DefaultValue = parsed[1]
 	return envVariable, nil
 }
 
@@ -305,10 +301,6 @@ func findInParentFolders(parameters string, terragruntOptions *options.Terragrun
 	return "", errors.WithStackTrace(ParentFileNotFound{Path: terragruntOptions.TerragruntConfigPath, File: fileToFindStr, Cause: fmt.Sprintf("Exceeded maximum folders to check (%d)", terragruntOptions.MaxFoldersToCheck)})
 }
 
-var oneQuotedParamRegex = regexp.MustCompile(`^"([^"]*?)"$`)
-var twoQuotedParamsRegex = regexp.MustCompile(`^"([^"]*?)"\s*,\s*"([^"]*?)"$`)
-var listQuotedParamRegex = regexp.MustCompile(`^"([^"]*?)"\s*,\s*(.*)$`)
-
 // Parse two optional parameters, wrapped in quotes, passed to a function, and return the parameter values and how many
 // of the parameters were actually set. For example, if you have a function foo(bar, baz), where bar and baz are
 // optional string parameters, this function will behave as follows:
@@ -318,22 +310,43 @@ var listQuotedParamRegex = regexp.MustCompile(`^"([^"]*?)"\s*,\s*(.*)$`)
 // foo("a", "b") -> return "a", "b", 2, nil
 //
 func parseOptionalQuotedParam(parameters string) (string, string, int, error) {
+	parsed, err := splitParamList(parameters)
+
+	if err != nil {
+		return "", "", 0, err
+	}
+
+	if len(parsed) == 0 {
+		return "", "", 0, nil
+	} else if len(parsed) == 1 {
+		return parsed[0], "", 1, nil
+	} else if len(parsed) == 2 {
+		return parsed[0], parsed[1], 2, nil
+	}
+	return "", "", 0, errors.WithStackTrace(InvalidStringParams(parameters))
+}
+
+func splitParamList(parameters string) ([]string, error) {
 	trimmedParameters := strings.TrimSpace(parameters)
 	if trimmedParameters == "" {
-		return "", "", 0, nil
+		return []string{}, nil
 	}
 
-	matches := oneQuotedParamRegex.FindStringSubmatch(trimmedParameters)
-	if len(matches) == 2 {
-		return matches[1], "", 1, nil
+	if !INTERPOLATION_PARAMETERS_REGEX.MatchString(trimmedParameters) {
+		return []string{}, errors.WithStackTrace(InvalidStringParams(parameters))
 	}
 
-	matches = twoQuotedParamsRegex.FindStringSubmatch(trimmedParameters)
-	if len(matches) == 3 {
-		return matches[1], matches[2], 2, nil
+	matches := INTERPOLATION_QUOTED_REGEX.FindAllString(trimmedParameters, -1)
+	var trimmedArgs []string
+	for _, match := range matches {
+		v, err := strconv.Unquote(match)
+		if err != nil {
+			return []string{}, errors.WithStackTrace(InvalidStringParams(parameters))
+		}
+		trimmedArgs = append(trimmedArgs, strings.TrimSpace(v))
 	}
 
-	return "", "", 0, errors.WithStackTrace(InvalidStringParams(parameters))
+	return trimmedArgs, nil
 }
 
 // parseParamList parses a string of comma separated parameters wrapped in quote and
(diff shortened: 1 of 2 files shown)


config/config_helpers.go
```diff
@@ -16,12 +17,13 @@ import (
 	"github.com/gruntwork-io/terragrunt/util"
 )
 
-var INTERPOLATION_PARAMETERS = `(\s*"[^"]*?"\s*,?\s*)*`
-var INTERPOLATION_SYNTAX_REGEX = regexp.MustCompile(fmt.Sprintf(`\$\{\s*\w+\(%s\)\s*\}`, INTERPOLATION_PARAMETERS))
+var INTERPOLATION_QUOTED_STRING = `"(?:\\.|[^\\"])*?"`
```
twang817: The crux of the change is here.  Here's a snippet from regex101 to help explain the regex:
image

twang817: In other words, `\\.` matches any escaped character, whereas `[^\\"]` matches anything except the backslash or the quote.  These are OR'd and then grouped in a non-capture group.  Finally, we use the kleene star to say we want zero or more of them (non-greedy).  Finally, the quotes are literal.

config/config_helpers.go
```diff
@@ -16,12 +17,13 @@ import (
...
+var INTERPOLATION_QUOTED_REGEX = regexp.MustCompile(INTERPOLATION_QUOTED_STRING)
+var INTERPOLATION_PARAMETERS = fmt.Sprintf(`\s*(?:%s(?:\s*,\s*%s)*)?\s*`, INTERPOLATION_QUOTED_STRING, INTERPOLATION_QUOTED_STRING)
```
twang817: With our `INTERPOLATION_QUOTED_STRING` above, we now construct a regex that will accept any number of them in a comma separate list.

The general pattern for comma separated list with at least one element: `A(?:,A)*`

To make it zero or more, we surround the entire thing with a non-capture group and a question mark for zero or one.  This makes our pattern now look like this: `(?:A(?:,A)*)?`

Finally, throw in the `\s*` everywhere, and we have our regex that will match a parameter list with zero or more parameters.

Again, regex101:
image

config/config_helpers.go
```diff
@@ -16,12 +17,13 @@ import (
...
+var INTERPOLATION_PARAMETERS_REGEX = regexp.MustCompile(fmt.Sprintf(`^%s$`, INTERPOLATION_PARAMETERS))
```
twang817: Here, the goal is to make sure it *only* matches a parameter list.  Since an empty string is a valid parameter list, we add a `^$` anchors around it to ensure that it matches the whole string.  This allows us to use this regex to validate that the string is indeed a parameter list.

config/config_helpers.go
```diff
@@ -16,12 +17,13 @@ import (
...
+var INTERPOLATION_SYNTAX_REGEX = regexp.MustCompile(fmt.Sprintf(`\$\{\s*(\w+)\((%s)\)\s*\}`, INTERPOLATION_PARAMETERS))
```
twang817: Finally, this adds the rest of the "stuff" around a properly formatted interpolation call.

The general pattern is: `\$\{(\w+)\(A)\)\}`, which should match `${anything(parameter-list)}`.  Again, throw in a bunch of `\s*`s and we have our pattern.

image

Note, I capture the \w+ and the contents of the parameter list.

twang817: With the regex's set up, note that we now have a regex that can accurately describe any properly formatted interpolation string in the format of `${foo_bar("a", "b", "c")}`.   HCL/HIL actually allows nesting, which I don't think that this will properly handle, but that's where we would have to go to a real parser as opposed to regex hackery.

Now, I no longer need to create additional regexes to help me validate or extract my strings.  I should be able to:

1. Detect an interpolation string with `INTERPOLATION_SYNTAX_REGEX`
2. Extract match[1] as the function name
3. Extract match[2] as the full parameter list
4. Use `INTERPOLATION_QUOTED_REGEX` to extract all my parameters.

This is exactly what the rest of the code attempts to do.

config/config_helpers.go
```diff
@@ -168,7 +170,7 @@ func processMultipleInterpolationsInString(terragruntConfigString string, includ
 // Given a string value from a Terragrunt configuration, parse the string, resolve any calls to helper functions using
 // Resolve a single call to an interpolation function of the format ${some_function()} in a Terragrunt configuration
 func resolveTerragruntInterpolation(str string, include *IncludeConfig, terragruntOptions *options.TerragruntOptions) (interface{}, error) {
-	matches := HELPER_FUNCTION_SYNTAX_REGEX.FindStringSubmatch(str)
+	matches := INTERPOLATION_SYNTAX_REGEX.FindStringSubmatch(str)
```
twang817: The HELPER_FUNCTION_SYNTAX_REGEX was a generalized version of the `INTERPOLATION_SYNTAX_REGEX` -- presumably with capture groups.  I couldn't think of a reason why I wouldn't just use `INTERPOLATION_SYNTAX_REGEX` after the capture groups were properly set up.

config/config_helpers.go
```diff
@@ -204,20 +206,14 @@ func getParentTfVarsDir(include *IncludeConfig, terragruntOptions *options.Terra
 
 func parseGetEnvParameters(parameters string) (EnvVar, error) {
 	envVariable := EnvVar{}
-	matches := HELPER_FUNCTION_GET_ENV_PARAMETERS_SYNTAX_REGEX.FindStringSubmatch(parameters)
-	if len(matches) < 2 {
-		return envVariable, errors.WithStackTrace(InvalidGetEnvParams(parameters))
-	}
+	parsed := parseParamList(parameters)
```
twang817: Again, since we're able to describe the entire parameter list, with zero or more parameters, we should just use that regex to extract all the parameters, then throw errors if it doesn't have the right number of args.

That's exactly what `parseParamList` does, so we just call that.

config/config_helpers.go
```diff
@@ -204,20 +206,14 @@ func getParentTfVarsDir(include *IncludeConfig, terragruntOptions *options.Terra
...
 
-	for index, name := range HELPER_FUNCTION_GET_ENV_PARAMETERS_SYNTAX_REGEX.SubexpNames() {
-		if name == "env" {
-			envVariable.Name = strings.TrimSpace(matches[index])
-		}
-		if name == "default" {
-			envVariable.DefaultValue = strings.TrimSpace(matches[index])
-		}
+	if len(parsed) != 2 || len(parsed[0]) == 0 {
```
twang817: Then its just a matter of checking that we have the right number of parameters.

Incidentally, the original `HELPER_FUNCTION_GET_ENV_PARAMETERS_SYNTAX_REGEX` also validated that the first parameter is non-empty.  So we do that here.

config/config_helpers.go
```diff
@@ -318,22 +310,43 @@ var listQuotedParamRegex = regexp.MustCompile(`^"([^"]*?)"\s*,\s*(.*)$`)
 // foo("a", "b") -> return "a", "b", 2, nil
 //
 func parseOptionalQuotedParam(parameters string) (string, string, int, error) {
+	parsed, err := splitParamList(parameters)
```
twang817: Again, we just extract the full parameter list.

Except here, there were a few tests that called this function with an invalid parameter list.  Now, in the normal flow of code, this would never happen because we're validating that up-front.  But, it's simple enough to validate it again in this function.

Here, I decided to factor out the guts of `parseParamList` to a new function `splitParamList` that would return errors if the input string wasn't a valid parameter-list.  This allows `parseOptionalQuotedParams` to return a proper error if validation fails, as on line 315 below.

config/config_helpers.go
```diff
@@ -318,22 +310,43 @@ var listQuotedParamRegex = regexp.MustCompile(`^"([^"]*?)"\s*,\s*(.*)$`)
...
+		return "", "", 0, err
+	}
+
+	if len(parsed) == 0 {
+		return "", "", 0, nil
+	} else if len(parsed) == 1 {
+		return parsed[0], "", 1, nil
+	} else if len(parsed) == 2 {
+		return parsed[0], parsed[1], 2, nil
+	}
```
twang817: Otherwise, we're just managing the return value based on the number of parameters we extracted...

config/config_helpers.go
```diff
@@ -318,22 +310,43 @@ var listQuotedParamRegex = regexp.MustCompile(`^"([^"]*?)"\s*,\s*(.*)$`)
...
 	trimmedParameters := strings.TrimSpace(parameters)
 	if trimmedParameters == "" {
-		return "", "", 0, nil
+		return []string{}, nil
 	}
 
-	matches := oneQuotedParamRegex.FindStringSubmatch(trimmedParameters)
-	if len(matches) == 2 {
-		return matches[1], "", 1, nil
+	if !INTERPOLATION_PARAMETERS_REGEX.MatchString(trimmedParameters) {
```
twang817: Here, as mentioned above, I re-validate that the input string is a valid parameter-list.  Normally, this wouldn't happen but there are a few tests that call this function directly.

config/config_helpers.go
```diff
@@ -318,22 +310,43 @@ var listQuotedParamRegex = regexp.MustCompile(`^"([^"]*?)"\s*,\s*(.*)$`)
...
-	if len(matches) == 3 {
-		return matches[1], matches[2], 2, nil
+	matches := INTERPOLATION_QUOTED_REGEX.FindAllString(trimmedParameters, -1)
+	var trimmedArgs []string
+	for _, match := range matches {
+		v, err := strconv.Unquote(match)
+		if err != nil {
+			return []string{}, errors.WithStackTrace(InvalidStringParams(parameters))
+		}
+		trimmedArgs = append(trimmedArgs, strings.TrimSpace(v))
```
twang817: Otherwise, we simply use the `INTERPOLATION_QUOTED_REGEX` regex to find all matches, then unquote each match, trim it, and stick it in the parameters.

config/config_helpers.go
```diff
@@ -342,36 +355,11 @@ func parseOptionalQuotedParam(parameters string) (string, string, int, error) {
...
-
-	for _, arg := range parsedArgs {
-		trimmedArg := strings.TrimSpace(arg)
-		if trimmedArg != "" {
-			trimmedArgs = append(trimmedArgs, trimmedArg)
-		}
-	}
-
-	return trimmedArgs
+	return parsed
```
twang817: Note, with `splitParamList` factored out, this function mainly just handles the error condition.

config/config_helpers_test.go
```diff
@@ -138,11 +138,17 @@ func TestRunCommand(t *testing.T) {
 		expectedErr       error
 	}{
 		{
-			`"/bin/bash", "-c", ""echo -n foo""`,
+			`"/bin/bash", "-c", "echo -n foo"`,
```
twang817: This test doesn't make much sense.

By splitting the arguments into an array, there's no need to rely on quoting to handle spaces.  This is something we do on the shell command-line _because_ the string hasn't already been split.

This means that `"echo -n foo"` is already treated as a single parameter being passed to `bash`.  There's no need to quote it again.

In fact, if we do, we're actually saying that we want to pass the string `"echo -n foo"` to the bash interpreter -- which is of course an invalid command.

It just so happened to work in the old implementation because there is a unconditional removal of all quotes.  With proper(ish) escape handling, this is no longer a valid test case.

Hopefully, no one was relying on this behavior.

config/config_helpers_test.go
```diff
@@ -138,11 +138,17 @@ func TestRunCommand(t *testing.T) {
...
 			terragruntOptionsForTest(t, homeDir),
 			"foo",
 			nil,
 		},
+		{
+			`"/bin/bash", "-c", "FOO=\"a test\"; echo -n $FOO"`,
```
twang817: A simple test that verifies handling of escaped quotes.  I didn't want to write an unit test that relies on `jq`.

config/config_helpers_test.go
```diff
@@ -624,8 +637,8 @@ TERRAGRUNT_HIT","")}/bar`,
 			nil,
 			terragruntOptionsForTestWithEnv(t, "/root/child/"+DefaultTerragruntConfigPath, map[string]string{"TEST_ENV_TERRAGRUNT_OTHER": "SOMETHING"}),
 			"",
-			InvalidInterpolationSyntax(`${get_env("TEST_ENV_
-TERRAGRUNT_HIT","")}`),
+			InvalidGetEnvParams(`"TEST_ENV_
```
twang817: Here, the newline would previously fail to even recognize as valid interpolation syntax.

Since the new parser actually handles newlines, the error that gets thrown is actually `InvalidGetEnvParams`

PR #731, opened 2019-06-06
Terraform 0.12 support

### LINKED ISSUE #466: Upgrading to Terraform 0.12: separate configuration file for Terragrunt?

brikis98:
This PR upgrades Terragrunt to work with Terraform 0.12. Key changes:

1. We are migrating from having the Terragrunt configuration in `terraform.tfvars` to `terragrunt.hcl`. This is necessary because Terraform 0.12 is more strict about what can be in `terraform.tfvars` files, and the `terragrunt = { ... }` configuration block we used to work is no longer allowed.

1. The configuration in `terragrunt.hcl` is largely the same as what we used to have in `terraform.tfvars`, except (a) we no longer need the `terragrunt = { ... }` wrapper, (b) we now use HCL2 syntax, (c) any input variables that need to be passed to the Terraform module need to be moved into an `inputs = { ... }` block.

1. HCL2 brings two major benefits: (a) first-class expressions, so not everything has to be wrapped in `${...}` and (b) a better parser, so we can now use Terragrunt's built-in functions _anywhere_ in `terragrunt.hcl`.

PR progress:

- [x] Update code
- [x] Update tests
- [x] Update docs
- [x] Add migration guide
- [x] Detect old Terraform version and show error
 
Note, this PR looks huge, but the vast majority of it is renaming all the `terraform.tfvars` files to `terragrunt.hcl` and updating to 0.12 syntax with first-class expressions.

Fixes #466.


_docs/migration_guides/upgrading_to_terragrunt_0.19.x.md
```diff
@@ -0,0 +1,268 @@
...
+Terraform 0.12 was released in May, 2019, and it included a few major changes:
+
+1. More strict rules around what can go in a `.tfvars` file. In particular, any variable defined in a `.tfvars` file 
+   that does not match a corresponding `variable` definition in your `.tf` files produces an error.
+1. A shift from HCL to HCL2 as the main syntax. This included support for first-class expressions (i.e., using variables
+   and functions without having to wrap everything in `${...}`).
+   
+Before version 0.19.0, Terragrunt had you define its configuration in a `terragrunt = { ... }` variable in 
+a `terraform.tfvars` file, but due to item (1) this no longer works with Terraform 0.12 and newer. That means we had to 
+move to a new file format. This requires a migration, which is unfortunate, but as a nice benefit, item (2) allows us
```
yorinasub17:
```suggestion
move to a new file format. This requires a migration, which is unfortunate, but as a nice benefit, item (2)
```

_docs/migration_guides/upgrading_to_terragrunt_0.19.x.md
```diff
@@ -0,0 +1,268 @@
...
+terraform {
+  extra_arguments "retry_lock" {
+    # Using a function within first-class expressions!
+    commands  = get_terraform_commands_that_need_locking()
+    arguments = ["-lock-timeout=20m"]
+  }
+}
+
+# This now works with Terragrunt 0.19.x and newer!
+foo = get_env("FOO", "default")
```
yorinasub17: Should this be wrapped in `inputs`?

cli/cli_app.go
```diff
@@ -404,6 +409,21 @@ func runTerragruntWithConfig(terragruntOptions *options.TerragruntOptions, terra
...
+// sets these environment variables in the given terragruntOptions.
+func setTerragruntInputsAsEnvVars(terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig) error {
+	asEnvVars, err := toTerraformEnvVars(terragruntConfig.Inputs)
+	if err != nil {
+		return err
+	}
+
+	for key, value := range asEnvVars {
+		terragruntOptions.Env[key] = value
+	}
```
yorinasub17: So this means we can't set the env var `TF_VAR_xxx` outside of terragrunt to override the inputs?

config/config.go
```diff
@@ -317,23 +261,41 @@ func parseConfigString(configString string, terragruntOptions *options.Terragrun
...
+func parseConfigStringAsTerragruntConfigFile(configString string, configPath string, terragruntOptions *options.TerragruntOptions, include *IncludeConfig) (*terragruntConfigFile, error) {
+	terragruntConfig := terragruntConfigFile{}
+	err := parseHcl(configString, configPath, &terragruntConfig, include, terragruntOptions)
+	if err != nil {
+		return nil, err
+	}
+	return &terragruntConfig, nil
+}
+
+// parseHcl uses the HCL2 parser to parse the given string into the strict specified by out.
```
yorinasub17:
```suggestion
// parseHcl uses the HCL2 parser to parse the given string into the struct specified by out.
```

config/config_helpers_test.go
```diff
@@ -560,120 +313,57 @@ func TestResolveEnvInterpolationConfigString(t *testing.T) {
 		include           *IncludeConfig
 		terragruntOptions *options.TerragruntOptions
 		expectedOut       string
-		expectedErr       error
+		expectedErr       string
```
yorinasub17: Just curious: Why did you switch to checking the string?

config/config_helpers_test.go
```diff
@@ -560,120 +313,57 @@ func TestResolveEnvInterpolationConfigString(t *testing.T) {
...
 
 	for _, testCase := range testCases {
-		actualOut, actualErr := ResolveTerragruntConfigString(testCase.str, testCase.include, testCase.terragruntOptions)
-		if testCase.expectedErr != nil {
-			assert.True(t, errors.IsError(actualErr, testCase.expectedErr), "For string '%s' include %v and options %v, expected error %v but got error %v and output %v", testCase.str, testCase.include, testCase.terragruntOptions, testCase.expectedErr, actualErr, actualOut)
-		} else {
-			assert.Nil(t, actualErr, "For string '%s' include %v and options %v, unexpected error: %v", testCase.str, testCase.include, testCase.terragruntOptions, actualErr)
-			assert.Equal(t, testCase.expectedOut, actualOut, "For string '%s' include %v and options %v", testCase.str, testCase.include, testCase.terragruntOptions)
-		}
+		testCase := testCase
```
yorinasub17: NIT: We usually add the boilerplate comment about this being done to bring the var into the scope of the forloop.

config/config_test.go
```diff
@@ -2,7 +2,6 @@ package config
```
yorinasub17: NOTE TO YORI SELF: Reviewed to here.

_docs/migration_guides/upgrading_to_terragrunt_0.19.x.md
```diff
@@ -0,0 +1,268 @@
...
+terraform {
+  # ...
+}
+
+remote_state {
+  # ...  
+}
+
+# Input variables to set for your Terraform module
+inputs = {
```
lorengordon: The block vs attr syntax is a bit confusing, but either hcl2 or terraform 0.12 impose some validation on this distinction. Just wanting to clarify here that `terraform` and `remote_state` truly are blocks, where `inputs` truly is an attr. Meaning this would not work (with an =):

```
terraform = {
  # ...
}

remote_state = {
  # ...  
}
```

and notionally, because those two are blocks, you could have several of them in the same config, somehow?

_docs/migration_guides/upgrading_to_terragrunt_0.19.x.md
```diff
@@ -0,0 +1,268 @@
...
+}
+```
+
+Since Terragrunt uses HCL2, we now have to be more strict with which parts of the Terragrunt configuration are 
+attributes and which are blocks:
+
+```hcl
+# terragrunt.hcl
+
+# terraform is a block, so make sure NOT to include an equals sign
```
lorengordon: Groan. I see you address this here.

README.md
```diff
@@ -1199,11 +1170,43 @@ This section contains detailed documentation for the following aspects of Terrag
...
+```
+$ terragrunt apply
+
+# Roughly equivalent to:
+
+TF_VAR_instance_type="t2.micro" \
+TF_VAR_instance_count=10 \
+TF_VAR_tags='{"Name":"example-app"}' \
+terraform apply
+```
```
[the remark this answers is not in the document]
brikis98: Oh, good point! I've updated the code to explicitly not override any env vars the user has already set, and added tests for the same:

https://github.com/gruntwork-io/terragrunt/pull/731/commits/fdce37cf37b5f8de94eee07d9e8b097cf45e92fa
https://github.com/gruntwork-io/terragrunt/pull/731/commits/993e1957a1f23e11961444b88f9ac94a47176a80

cli/cli_app.go
brikis98: Thanks for the great catch. Fixed and added tests:

https://github.com/gruntwork-io/terragrunt/pull/731/commits/fdce37cf37b5f8de94eee07d9e8b097cf45e92fa
https://github.com/gruntwork-io/terragrunt/pull/731/commits/993e1957a1f23e11961444b88f9ac94a47176a80

_docs/migration_guides/upgrading_to_terragrunt_0.19.x.md
brikis98: Yes, fixed! https://github.com/gruntwork-io/terragrunt/pull/731/commits/47391c61b67de90d19323c340980b5d32c353d39

config/config_helpers_test.go
brikis98: Before, the error was returned directly from my code, so I could check it exactly. Now, the error is wrapped by the HCL2 parser, so doing an exactly equality check is harder. Therefore, I'm just doing a string contains check for the error name.

config/config_helpers_test.go
brikis98: Fixed: https://github.com/gruntwork-io/terragrunt/pull/731/commits/9f19fd96d8c8301fb6078b1d86aac115a6bea00b

test/fixture-download/local-with-exclude-dir/integration-env/gce/module-gce-b/terragrunt.hcl
```diff
@@ -0,0 +1,11 @@
+inputs = {
+  name = "Module GCE B"
+}
+
+terraform {
+  source = "../../../..//hello-world-no-remote"
+}
+
+dependencies {
+  paths = ["../../aws/module-aws-a"]
```
yorinasub17: Looks like all these `module-gce` configs depend on `aws`. I know this was the same as before, but is that intentional?

brikis98:
> I think the only thing I might add is an integration test that verifies you can use functions in the inputs expression, and that that works correctly to pass through to the TF_VAR. I know in theory that is a tautology, but would be nice to have that extra verification.

Added: https://github.com/gruntwork-io/terragrunt/pull/731/commits/dc5f0aa31d89211c3a6879e22eff551b732457e9

config/config_test.go
```diff
@@ -17,17 +16,14 @@ func TestParseTerragruntConfigRemoteStateMinimalConfig(t *testing.T) {
 	t.Parallel()
 
 	config := `
-terragrunt = {
-  remote_state {
-    backend = "s3"
-  }
+remote_state {
+	backend = "s3"
+	config = {}
```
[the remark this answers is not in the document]
brikis98: Fixed: https://github.com/gruntwork-io/terragrunt/pull/731/commits/249662ef03cb27ea3a06a3a688da1493f20b901f. Shouldn't matter at all now that we're using the real HCL parser instead of regex.

test/fixture-download/local-with-exclude-dir/integration-env/gce/module-gce-b/terragrunt.hcl
brikis98: I think these were written by a contributor. Don't know if intentional or not, but probably not worth tweaking as part of this already massive PR.

brikis98:
Doh. Tests are failing due to what I _think_ is a bug in Terraform 0.12: https://github.com/hashicorp/terraform/issues/21653. This may be a blocker...

_docs/migration_guides/upgrading_to_terragrunt_0.19.x.md
brikis98: I found the HCL2 parser was a bit weird with these. To be able to define the attributes within these `terraform` and `remote_state`, I had to make them blocks... They really should only appear once, so I wanted to make them attributes, but I got confusing errors when I did that. Maybe attributes can't have well-defined properties within? Not sure.

lorengordon: Yeah, I'm frustrated with this change in hcl2 at the moment. Not sure if you've seen it already, but there is a property you can set on the schema to treat a block as an attr. The many, many PRs linked to [this issue](https://github.com/hashicorp/terraform/issues/20505#issuecomment-496603984) might point to a way forward to use attrs?

lorengordon: Here is an example: https://github.com/terraform-providers/terraform-provider-aws/pull/8228/files

README.md
```diff
@@ -611,25 +587,59 @@ they don't already exist:
...
+  skip_bucket_versioning         = true # use only if the object store does not support versioning
+  skip_bucket_ssencryption       = true # use only if non-encrypted Terraform State is required and/or the object store does not support server-side encryption
+  skip_bucket_accesslogging      = true # use only if the cost for the extra object space is undesirable or the object store does not support access logging
+  enable_lock_table_ssencryption = true # use only if non-encrypted DynamoDB Lock Table for the Terraform State is required and/or the NoSQL database service does not support server-side encryption
+
+  shared_credentials_file     = "/path/to/credentials/file"
+  skip_region_validation      = true
+  skip_credentials_validation = true
+  skip_requesting_account_id  = true
+  skip_get_ec2_platforms      = true
```
antonbabenko: skip_get_ec2_platforms and skip_requesting_account_id are deprecated and should be removed.

_docs/migration_guides/upgrading_to_terragrunt_0.19.x.md
lorengordon: Also just found the behavior better described here, https://www.terraform.io/docs/extend/terraform-0.12-compatibility.html#computed-resource-attributes

brikis98:
Just added a bunch of fixes:

1. Switch to using `go-getter` directly instead of `terraform init -from-module`. The latter seems to be poorly maintained, as it had some unfortunate bugs in 0.11 (https://github.com/hashicorp/terraform/issues/18460) and now in 0.12, it flat out crashes/panics. 

1. Since I'm no longer running `terraform init` twice, I had to backport the hooks functionality so you could still execute hooks before/after the usage of `go-getter`.

1. Terraform 0.12 stores state differently in `.tfstate` files (all possible keys are now stored, unset ones defaulting to `null`), so I had to fix the state equality check too.

brikis98:
Oh, one more fix:

Looks like with the `--terragrunt-non-interactive`, we were assuming "yes" for all prompts, including the question you get for `xxx-all` commands of whether the `xxx` command should be applied to "external dependencies" (those outside the working directory). So if you ran `destroy-all` in folder `foo`, it not only destroyed everything in folder `foo`, but also all external dependencies of anything in `foo`. That is not a safe default, so I have reversed that behavior. It was exposed by a test failure that, I guess, 0.11 was somehow covering up?

cli/download_source.go
```diff
@@ -349,16 +358,41 @@ func cleanupTerraformFiles(path string, terragruntOptions *options.TerragruntOpt
...
 
-// Download the code from the Canonical Source URL into the Download Folder using the terraform init command
-func terraformInit(terraformSource *TerraformSource, terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig) error {
-	terragruntOptions.Logger.Printf("Downloading Terraform configurations from %s into %s using terraform init", terraformSource.CanonicalSourceURL, terraformSource.DownloadDir)
+// Download the code from the Canonical Source URL into the Download Folder using the go-getter library
+func downloadSource(terraformSource *TerraformSource, terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig) error {
+	terragruntOptions.Logger.Printf("Downloading Terraform configurations from %s into %s", terraformSource.CanonicalSourceURL, terraformSource.DownloadDir)
+
+	// go-getter will not download into folders that already exist, so initially, we download into a temp folder
+	tmpDownloadDir, err := ioutil.TempDir("", "terragrunt-download-temp")
```
lorengordon: Would it make sense to respect `TERRAGRUNT_DOWNLOAD_DIR` here? I'm thinking about issues on Windows with long filepaths, and a workaround of exporting something like `TERRAGRUNT_DOWNLOAD_DIR = "C:\.terragrunt-cache"`. If we first download instead to `TEMP`, that makes the workaround ineffective...

brikis98: Ah, that's a good point. Fixed in https://github.com/gruntwork-io/terragrunt/pull/731/commits/65741dbfe6cb36a843c054051e0056cb57767d36.

cli/download_source.go
```diff
@@ -356,9 +365,38 @@ func getTerraformSourceUrl(terragruntOptions *options.TerragruntOptions, terragr
...
+		return errors.WithStackTrace(err)
+	}
+	if err := util.CopyFolderContents(tmpDownloadDir, terraformSource.DownloadDir); err != nil {
+		return err
+	}
+
+	// Clean up the tmp folder
+	if err := os.RemoveAll(tmpDownloadDir); err != nil {
+		return errors.WithStackTrace(err)
+	}
```
yorinasub17: Why not use `defer` here?

brikis98: Done: https://github.com/gruntwork-io/terragrunt/pull/731/commits/e2398083111a4c1c3d47dbf34ca7154f6ae6bcf3

README.md
brikis98: More info on this?

antonbabenko: https://www.terraform.io/docs/backends/types/s3.html#skip_get_ec2_platforms
https://www.terraform.io/docs/backends/types/s3.html#skip_region_validation
https://www.terraform.io/docs/backends/types/s3.html#skip_requesting_account_id

brikis98: Thanks! https://github.com/gruntwork-io/terragrunt/pull/731/commits/866e689a5562822a3a4d6f8b081c2ec34110b41a

brikis98:
Alright, this PR is driving me bonkers with an endless yak shave. First the bugs with `terraform init -from-module`, which I worked around by moving to `go-getter`; then bugs with copying folder contents and concurrency; I worked around those, and now `TestTerragruntStackCommands` is failing with the error:

```
NoSuchBucket: The specified bucket does not exist
```

Digging into it, I think this is happening in the part of the code that [automatically creates the S3 bucket](https://github.com/gruntwork-io/terragrunt/blob/tf12/remote/remote_state_s3.go#L206-L220). What's bewildering is that, as far as I can tell from the logs, the [WaitUntilS3BucketExists](https://github.com/gruntwork-io/terragrunt/blob/tf12/remote/remote_state_s3.go#L410) is returning successfully, which means the S3 `HeadBucket` operation is saying that the bucket _does_ exist. However, when the code moves on to [enable versioning on the bucket](https://github.com/gruntwork-io/terragrunt/blob/tf12/remote/remote_state_s3.go#L447), I get back the error the bucket doesn't exist. I even added a 5 second sleep between creation and enabling versioning, in case it was an eventual consistency issue, and it didn't help. WTF? 😕

lorengordon:
Any chance it might be a credential issue? Maybe temporary keys expiring at an inopportune moment? Or perhaps credential in combination with region, where the bucket is created in one region but then the versioning logic is somehow referencing another region?

brikis98:
> Any chance it might be a credential issue? Maybe temporary keys expiring at an inopportune moment? Or perhaps credential in combination with region, where the bucket is created in one region but then the versioning logic is somehow referencing another region?

No, I don't think it's a credentials issue, as there are a bunch of tests running in parallel, all with the same credentials, and while all the tests before/during/after `TestTerragruntStackCommands` pass, this one test fails consistently.

brikis98:
And of course `TestTerragruntStackCommands` passes without issues on my own computer... It only fails in CI...

brikis98:
OK, I've put in a workaround that does the S3 bucket creation/configuration in a retry loop with some sleep between retries. Tests seem to be passing consistently now. Finally merging this!

brikis98:
This code is now available in: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.19.0

barryib:
@brikis98 awesome. Thanks for your great work. I tried this release, but I was wondering how do we have to handle var files (required and optional) in Terragrunt, because Terraform is now rising and deprecation warning for unknown variables. I opened the following issue https://github.com/gruntwork-io/terragrunt/issues/737 to track this

PR #739, opened 2019-06-11
Switch from symlink and file copy to just file copy; upgrade dependencies; fix stack traces.

### LINKED ISSUE #733: failed to run mklink
I just upgraded to terragrunt v0.19.0 this morning.  I am running Terraform v0.12.1.  I am running on Windows with Powershell.  I am trying to setup an example repo based on terragrunt to show to my team for possible adoption.  

When I run terragrunt init or terragrunt plan I get the following:

terragrunt init
[terragrunt] [C:\Users\matt.mclane\repos\llamasoft\terraform\environments\prod\network] 2019/06/10 10:11:48 Running command: terraform --version
[terragrunt] 2019/06/10 10:11:48 Reading Terragrunt config file at C:/Users/matt.mclane/repos/llamasoft/terraform/environments/prod/network/terragrunt.hcl
[terragrunt] 2019/06/10 10:11:48 Downloading Terraform configurations from file://C:/Users/matt.mclane/repos/llamasoft/terraform/modules into C:/Users/matt.mclane/repos/llamasoft/terraform/environments/prod/network/.terragrunt-cache/j11A824preQo1pkd4TD25NjWlKs/5dESSr0bTp-_uymXIM66yNDT_Yo
[terragrunt] 2019/06/10 10:11:48 Hit multiple errors:
error downloading 'file://C:/Users/matt.mclane/repos/llamasoft/terraform/modules': failed to run mklink C:/Users/matt.mclane/repos/llamasoft/terraform/environments/prod/network/.terragrunt-cache/tmp-download-Ry4o94 C:\Users\matt.mclane\repos\llamasoft\terraform\modules: exit status 1 "Invalid switch - \"Users\".\r\n"
[terragrunt] 2019/06/10 10:11:48 Unable to determine underlying exit code, so Terragrunt will exit with error code 1
All the paths seem correct.

### LINKED ISSUE #736: Terragrunt stopped copying symlinks from modules
In the comment https://github.com/gruntwork-io/terragrunt/blob/master/util/file.go#L153 it is said that `file.Glob` was chosen to support symlinks, but then https://github.com/gruntwork-io/terragrunt/commit/264812da348c2d977ec4e7ad93a89b53a25c8cec deliberately started skipping symlinks?

Is there a reasoning behind it?
We use it extensively to keep it DRY:
```
common
     |----- infra.tf (with the terraform backend block)
     |----- providers.tf (with common providers)
     |----- variables.tf (with common variables for all modules like aws_profile and aws_region)

module1
      |------ infra.tf -> ../common/infra.tf
      |------ vars_common.tf -> ../common/variables.tf
```

I'm using `terragrunt version v0.19.0`
After reverting the commit above everything works as expected.

brikis98:
1. By default, `go-getter` uses symlinks for local files. This is fast and disk-space friendly, but it [doesn't seem to work on windows](https://github.com/gruntwork-io/terragrunt/issues/733). Moreover, you can't copy anything into a symlinked folder, as it would end up copying it back into the original, so this caused issues for Terragrunt. As a workaround, I was letting `go-getter` create the symlink in one folder and then copying the files to another, but this was effectively the worst of both worlds—slow, not friendly to disk space, and complicated. In this PR, I've added by own `Getter` implementation that does file copying instead of using symlinks. I'm _hoping_ this fixes #733. It also allow us to copy existing symlinks without hitting infinite cycles, which I _think_ fixes #736. 
1. Update all our dependency versions. Turns out we had some really old dependencies in there, including a several-year-old version of `go-getter`.
1. Fix a bug where stack traces were being lost due to `MultiError` usage.

### DIFF
--- a/cli/download_source.go
+++ b/cli/download_source.go
@@ -338,12 +338,15 @@ func cleanupDownloadDir(terraformSource *TerraformSource, terragruntOptions *opt
 	// Backup the working dir if it exists
 	backupWorkingDirFolder := util.JoinPath(terragruntOptions.DownloadDir, ".working-dir-backup")
 	if util.FileExists(terraformSource.WorkingDir) {
+		if err := os.MkdirAll(terragruntOptions.DownloadDir, 0700); err != nil {
+			return errors.WithStackTrace(err)
+		}
 		if err := os.Rename(terraformSource.WorkingDir, backupWorkingDirFolder); err != nil {
 			return errors.WithStackTrace(err)
 		}
 	}
 
-	// Delete everything in the DownloadDir
+	// Delete everything that was in the download dir
 	files, err := zglob.Glob(util.JoinPath(terraformSource.DownloadDir, "*"))
 	if err != nil {
 		return errors.WithStackTrace(err)
--- a/cli/download_source.go
+++ b/cli/download_source.go
@@ -8,6 +8,8 @@ import (
 	"regexp"
 	"strings"
 
+	"path/filepath"
+
 	"github.com/gruntwork-io/terragrunt/config"
 	"github.com/gruntwork-io/terragrunt/errors"
 	"github.com/gruntwork-io/terragrunt/options"
@@ -15,7 +17,6 @@ import (
 	"github.com/hashicorp/go-getter"
 	urlhelper "github.com/hashicorp/go-getter/helper/url"
 	"github.com/mattn/go-zglob"
-	"path/filepath"
 )
 
 // This struct represents information about Terraform source code that needs to be downloaded
@@ -375,7 +376,7 @@ func cleanupDownloadDir(terraformSource *TerraformSource, terragruntOptions *opt
 		// Filter out files in .terraform folders, since those are from modules downloaded via a call to terraform init,
 		// and we don't want to re-download them.
 		for _, terraformFile := range terraformFiles {
-			if !strings.Contains(terraformFile, ".terraform") {
+			if !util.PathContains(terraformFile, ".terraform") {
 				filteredTerraformFiles = append(filteredTerraformFiles, terraformFile)
 			}
 		}
--- a/cli/download_source.go
+++ b/cli/download_source.go
@@ -326,8 +326,8 @@ func isLocalSource(sourceUrl *url.URL) bool {
 }
 
 // If this download dir already exists, delete its contents to ensure that we don't have any old files from previous
-// runs. Leave only the .terraform folder in the working dir, as that may contain terraform state we don't want to
-// lose / reconfigure as well as module and provider code we don't want to redownload.
+// runs. Leave only the non-Terraform files in the working dir, as that dir may contain terraform state we don't want
+// to lose / reconfigure as well as module and provider code we don't want to re-download.
 func cleanupDownloadDir(terraformSource *TerraformSource, terragruntOptions *options.TerragruntOptions) error {
 	if !util.FileExists(terraformSource.DownloadDir) {
 		return nil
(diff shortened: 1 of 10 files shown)


cli/download_source.go
```diff
@@ -325,31 +325,68 @@ func isLocalSource(sourceUrl *url.URL) bool {
...
+		terraformFiles, err := zglob.Glob(util.JoinPath(terraformSource.WorkingDir, "**/*.tf"))
+		if err != nil {
+			return errors.WithStackTrace(err)
+		}
+		filteredTerraformFiles := []string{}
+
+		// Filter out files in .terraform folders, since those are from modules downloaded via a call to terraform init,
+		// and we don't want to re-download them.
+		for _, terraformFile := range terraformFiles {
+			if !strings.Contains(terraformFile, ".terraform") {
```
yorinasub17: Should this actually look at the path components? Is there a situation where you could have a hidden file that we want to actually delete, like `.terraform-test.tf`?

cli/file_copy_getter.go
```diff
@@ -0,0 +1,41 @@
+package cli
+
+import (
+	"fmt"
+	"github.com/gruntwork-io/terragrunt/util"
+	"github.com/hashicorp/go-getter"
```
yorinasub17: NIT: I like to separate out stdlib imports from thirdparty imports from local imports. E.g:

```
import (
  "fmt"
  "net/url"
  "os"

  "github.com/hashicorp/go-getter"

)
```

cli/version_check_test.go
```diff
@@ -12,11 +12,6 @@ func TestCheckTerraformVersionMeetsConstraintEqual(t *testing.T) {
 	testCheckTerraformVersionMeetsConstraint(t, "v0.9.3", ">= v0.9.3", true)
 }
 
-func TestCheckTerraformVersionMeetsConstraintGreaterDev(t *testing.T) {
-	t.Parallel()
-	testCheckTerraformVersionMeetsConstraint(t, "v0.9.4-dev", ">= v0.9.3", true)
-}
```
yorinasub17: Curious why this was removed?

ekini:
I've tested it and it actually fixes #736 👍

cli/file_copy_getter.go
ekini: https://godoc.org/golang.org/x/tools/cmd/goimports can be used instead of `go fmt`
It works faster
```
terragrunt (go-getter-update*) » time goimports -w cli/file_copy_getter.go
goimports -w cli/file_copy_getter.go  0.01s user 0.00s system 82% cpu 0.015 total
terragrunt (go-getter-update*) » time go fmt cli/file_copy_getter.go
go fmt cli/file_copy_getter.go  0.42s user 0.41s system 244% cpu 0.341 total
```
and automatically manages imports (most of the time) and formats them
```
diff -u cli/file_copy_getter.go.orig cli/file_copy_getter.go
--- cli/file_copy_getter.go.orig	2019-06-12 08:20:02.000000000 +1200
+++ cli/file_copy_getter.go	2019-06-12 08:20:02.000000000 +1200
@@ -2,10 +2,11 @@

 import (
 	"fmt"
-	"github.com/gruntwork-io/terragrunt/util"
-	"github.com/hashicorp/go-getter"
 	"net/url"
 	"os"
+
+	"github.com/hashicorp/go-getter"
 )
```

brikis98:
> I've tested it and it actually fixes #736 👍

Fantastic, thanks for trying it out!

cli/version_check_test.go
brikis98: Oh, it's in the commit message, but not obvious from the PR: the `go-version` library we use under the hood was updated as part of this PR, and they changed how they handle version comparisons between "pre-releases" (e.g., v0.9.4-dev), so this comparison isn't actually valid anymore.

cli/file_copy_getter.go
brikis98: Heh, I use IntelliJ and almost never look at the imports; all imports happen automatically while typing, so the order of them never felt important. That said, we do use `goimports` with Terratest...

brikis98: Just ran `goimports` and it updated 26 files. Probably not worth complicating this PR with that many changes. We should probably just do it in a separate PR and perhaps add it to the `make fmt` command.

cli/download_source.go
brikis98: Yes, I noticed that and felt too lazy to do it right... But, fine, _grumble grumble_. https://github.com/gruntwork-io/terragrunt/pull/739/commits/7c8a472ac48bfa9907b4a33502f74062ffff677d

cli/version_check_test.go
yorinasub17: Ah got it. Thanks for the explanation! I should have scanned the commit messages.

brikis98:
Yori, thx for the review. Merging and releasing now.

options/auto_retry_options.go
```diff
@@ -14,4 +14,6 @@ var RETRYABLE_ERRORS = []string{
 	"(?s).*Error installing provider.*TLS handshake timeout.*",
 	"(?s).*Error configuring the backend.*TLS handshake timeout.*",
 	"(?s).*Error installing provider.*tcp.*timeout.*",
+	"(?s).*Error installing provider.*tcp.*timeout.*",
```
antonbabenko: Duplicated lines 16 and 17.

brikis98: Good catch! https://github.com/gruntwork-io/terragrunt/pull/754

PR #741, opened 2019-06-12
Fix concurrency issue with go-getter usage

brikis98:
When overriding the "file" getter in our `go-getter` usage, we were accidentally doing it in the globally-shared `getter.Getters` map. This is bad because (a) we're modifying a global variable and (b) since Terragrunt may make this change from many goroutines concurrently when you call `xxx-all` methods, this was occasionally causing a `fatal error: concurrent map writes` error.


brikis98:
Thx for the review!

PR #747, opened 2019-06-13
delete .tf.json files when cleaning up terragrunt cache

### LINKED ISSUE #725: .tf.json files not cleaned up
I receive an error when I run subsequent commands after the initial command (`terragrunt plan/apply`). I believe this is because `.tf.json` files aren't cleaned up from `.terragrunt-cache`.

```
The working directory already contains files. The -from-module option requires
an empty directory into which a copy of the referenced module will be placed.

To initialize the configuration already in this working directory, omit the
-from-module option.
```

victoriaalee:
Closes #725.

Note: ` zglob` does not seem to support `{}` or `?()` syntax, so an additional glob expression to get `.tf.json` files is added.

### DIFF
--- a/cli/download_source.go
+++ b/cli/download_source.go
@@ -371,6 +371,11 @@ func cleanupDownloadDir(terraformSource *TerraformSource, terragruntOptions *opt
 		if err != nil {
 			return errors.WithStackTrace(err)
 		}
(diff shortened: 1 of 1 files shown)


brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.19.3

Issue #743, opened 2019-06-12
Will there be continuing support for 0.18.x?


cscheib:
Will there be continuing support/releases for the 0.18.x version, since there are breaking changes on both the terragrunt and terraform (with .12) sides?

relevant to maintaining a Homebrew version of the tool:
https://github.com/Homebrew/homebrew-core/pull/40926

brikis98:
No, probably not. We may do some sort of hotfix for security issues, but other than that, there will be no additional support for Terraform 0.11.x. We're doing our best to keep up with the work for one version of Terraform, and unfortunately, I just don't think we'd be able to handle more than that!

ekini:
But actually, what does really make the new terragrunt incompatible with terraform 0.11?
There are new configuration files `terragrunt.hcl`, ok, but it has nothing to do with terraform itself.

I installed terraform 0.11 into a separate directory, converted `terraform.tfvars` into `terragrunt.hcl`   and hacked the `PATH`:
```
$PATH=~/bin/terraformv11:$PATH terragrunt plan
```
It worked!

For some reason `--terragrunt-tfpath` fails with
```
terragrunt --terragrunt-tfpath=~/bin/terraformv11/terraform plan
Incorrect Usage. flag provided but not defined: -terragrunt-tfpath
```
Anyway, can we set the terraform path in `terragrunt.hcl` instead? It would:
1. Allow using 0.12 in parent `terragrunt.hcl`
2. Allow overriding the version to 0.11 in a module `terragrunt.hcl`

So we'll be able to have a mix of modules for 0.11/0.12.

brikis98:
The differences from 0.11 weren't huge:

1. We're passing the `inputs = { ... }` block as environment variables, converting them to JSON first. Not positive this works as expected with Terraform 0.11.
1. We had to change a bit how we check backend configuration equality. I _think_ it may still work with 0.11, but am not sure.
1. The most important one, and why I set the version constraint to `>= 0.12.0`: we're no longer testing with Terraform 0.11, only Terraform 0.12. So even if it works now, it may well break later, and I'm not a fan of having to develop and test against multiple versions.

The unfortunate reality is that Terraform is not a 1.x tool, so there's no promise of backwards compatibility; the same is true of anything that works with Terraform, including Terragrunt.

ekini:
Yes, I forgot to mention that I removed the version constrain from terragrunt before it worked :)

[...]

PR #751, opened 2019-06-17
Feature/tfbinary in config

ekini:
This change allows setting `terraform_binary` in `terragrunt.hcl` (a module-level one or in an included `terragrunt.hcl`.

Adds thisto the migration doc, allowing using a mix of terraform-0.11 and terraform-0.12 modules together, for example set `terraform_binary=~/bin/terraform-v11/terraform` in an included root `terragrunt.hcl` and allows to override it in specific modules.

Lowers the required terraform version to 0.11 to make it work with the previous version of terraform.

A note: checking for terraform version doesn't really make sense in this case, as it checks for only the default version of it. Maybe increase the minimum version to `0.12` again? But it will make it unclear that different terraform is checked for before running modules runs.

Makes supporting terragrunt `0.18` irrelevant as well, as per #743.

### DIFF
--- a/_docs/migration_guides/upgrading_to_terragrunt_0.19.x.md
+++ b/_docs/migration_guides/upgrading_to_terragrunt_0.19.x.md
@@ -276,12 +276,14 @@ Make sure to make the corresponding updates in your `terragrunt.hcl` file!
 
 ### Use older Terraform
 
-Although it is not officiall supported and not tested, it is still possible to use terraform<0.12 with terragrunt >=0.19.
+Although it is not officially supported and not tested, it is still possible to use terraform<0.12 with terragrunt >=0.19.
 
-Just install a different version of terraform into a directory of your choice outside of `PATH` and specify path to the binary in `terragrunt.hcl` as `terraform_binary`:
+Just install a different version of terraform into a directory of your choice outside of `PATH` and specify path to the binary in `terragrunt.hcl` as `terraform_binary`, plus you need to lower the version check constraint:
 
 ```hcl
 
 terraform_binary = "~/bin/terraform-v11/terraform"
+terraform_version_constraint = ">= 0.11"
+
 
 ```
--- a/_docs/migration_guides/upgrading_to_terragrunt_0.19.x.md
+++ b/_docs/migration_guides/upgrading_to_terragrunt_0.19.x.md
@@ -27,6 +27,7 @@ to Terragrunt 0.19.x and newer:
 1. [Use first-class expressions](#use-first-class-expressions)
 1. [Check attributes vs blocks usage](#check-attributes-vs-blocks-usage)
 1. [Rename a few built-in functions ](#rename-a-few-built-in-functions)
+1. [Use terraform \<0.12](#use-older-terraform)
 
 Check out [this PR in the terragrunt-infrastructure-live-example 
 repo](https://github.com/gruntwork-io/terragrunt-infrastructure-live-example/pull/17) for an example of what the code 
@@ -272,3 +273,15 @@ Two built-in functions were renamed:
 1. `get_parent_tfvars_dir()` is now called `get_parent_terragrunt_dir()`.   
 
 Make sure to make the corresponding updates in your `terragrunt.hcl` file!
+
+### Use older Terraform
+
+Although it is not officiall supported and not tested, it is still possible to use terraform<0.12 with terragrunt >=0.19.
+
+Just install a different version of terraform into a directory of your choice outside of `PATH` and specify path to the binary in `terragrunt.hcl` as `terraform_binary`:
+
+```hcl
+
+terraform_binary = "~/bin/terraform-v11/terraform"
+
+```
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -201,14 +201,6 @@ func runApp(cliContext *cli.Context) (finalErr error) {
 		return err
 	}
 
-	if err := PopulateTerraformVersion(terragruntOptions); err != nil {
-		return err
-	}
-
-	if err := CheckTerraformVersion(DEFAULT_TERRAFORM_VERSION_CONSTRAINT, terragruntOptions); err != nil {
-		return err
-	}
-
 	givenCommand := cliContext.Args().First()
 	command := checkDeprecated(givenCommand, terragruntOptions)
 	return runCommand(command, terragruntOptions)
@@ -246,10 +238,22 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 		return err
 	}
 
-	if terragruntOptions.TerraformPath == options.TERRAFORM_DEFAULT_PATH { // not changed from default
-		if terragruntConfig.TerraformBinary != "" {
-			terragruntOptions.TerraformPath = terragruntConfig.TerraformBinary
-		}
+	// Change the terraform binary path before checking the version
+	// if the path is not changed from default and set in the config.
+	if terragruntOptions.TerraformPath == options.TERRAFORM_DEFAULT_PATH && terragruntConfig.TerraformBinary != "" {
+		terragruntOptions.TerraformPath = terragruntConfig.TerraformBinary
+	}
+
+	if err := PopulateTerraformVersion(terragruntOptions); err != nil {
+		return err
+	}
+
+	versionConstraint := DEFAULT_TERRAFORM_VERSION_CONSTRAINT
+	if terragruntConfig.TerraformVersionConstraint != "" {
+		versionConstraint = terragruntConfig.TerraformVersionConstraint
+	}
+	if err := CheckTerraformVersion(versionConstraint, terragruntOptions); err != nil {
+		return err
 	}
 
 	if terragruntConfig.Skip {
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -150,7 +150,7 @@ AUTHOR(S):
 var MODULE_REGEX = regexp.MustCompile(`module[[:blank:]]+".+"`)
 
 // This uses the constraint syntax from https://github.com/hashicorp/go-version
-// This version of Terragrunt only works with Terraform 0.12.0 and above
+// This version of Terragrunt was tested to work with Terraform 0.12.0 and above only
 const DEFAULT_TERRAFORM_VERSION_CONSTRAINT = ">= v0.12.0"
 
 const TERRAFORM_EXTENSION_GLOB = "*.tf"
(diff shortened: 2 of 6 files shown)


_docs/migration_guides/upgrading_to_terragrunt_0.19.x.md
```diff
@@ -118,9 +119,9 @@ terragrunt = {
 # Input variables to set for your Terraform module
 instance_type  = "t2.micro"
 instance_count = 10
-``` 
+```
```
brikis98: Any chance you could remove all the whitespace changes? Makes the PR harder to review and find the real diff.

_docs/migration_guides/upgrading_to_terragrunt_0.19.x.md
```diff
@@ -260,15 +261,27 @@ dependencies {
...
 
 1. `get_tfvars_dir()` is now called `get_terragrunt_dir()`.
-1. `get_parent_tfvars_dir()` is now called `get_parent_terragrunt_dir()`.   
+1. `get_parent_tfvars_dir()` is now called `get_parent_terragrunt_dir()`.
 
 Make sure to make the corresponding updates in your `terragrunt.hcl` file!
+
+### Use older Terraform
+
+Although it is not officiall supported and not tested, it is still possible to use terraform<0.12 with terragrunt >=0.19.
```
brikis98:
```suggestion
Although it is not officially supported or tested, it is still possible to use terraform<0.12 with terragrunt >=0.19.
```

_docs/migration_guides/upgrading_to_terragrunt_0.19.x.md
ekini: Yes, definitely. I made a separate commit with this change in the first place. Reverted that one

ekini:
First of all, thanks for the feedback!

> I appreciate the PR and the desire to make migration easier, but I'm uncomfortable claiming to support older versions of Terraform with zero testing of those versions. It's fine to put disclaimers in a README, but let's be honest, no one will read those, and the result will be lots of angry bugs filed and time pulled away from making Terragrunt better to fixing bugs with old versions.

On the other hand, if terragrunt is positioned as a thin wrapper for terraform, then why does it require a specific version of it? :)

> BTW, the most likely place this will break is how variables are passed to Terraform. In previous versions, we used `.tfvars` files, so it worked automatically. Now we're passing variables via env vars, and I'm not sure TF 0.12 parses those the same way as TF 0.11. And with no testing, it just feels like taking on a bunch of tech debt...

I've tested it on almost all modules we have, and it breaks only in one place: boolean variables used as count. It happens because of [special handling](https://www.terraform.io/docs/configuration-0-11/variables.html) of `tfvars` files:

> Variables with boolean values in a tfvars file will likewise be converted to "0" and "1" values.
> Variables specified via the -var command line flag will be literal strings "true" and "false", so care should be taken to explicitly use "0" or "1".
> Variables specified with the TF_VAR_ environment variables will be literal string values, just like -var.


However, If https://github.com/gruntwork-io/terragrunt/issues/752 is accepted, then this issue will be resolved (`.tfvars.json` are handled the same way as `.tfvars` even it's not mentioned in the doc).

brikis98:
> On the other hand, if terragrunt is positioned as a thin wrapper for terraform, then why does it require a specific version of it?

Because Terraform is pre 1.0.0, so it offers no backwards compatibility guarantees on its public APIs, so to wrap it, we have to know what APIs we are wrapping. E.g., In the past, the format of the `terraform init` command changed in backwards incompatible ways; now, Terraform processes `terraform.tfvars` files differently, so we had to move away from the old file format; and both in the past and the present, the way Terraform parses variables you pass to it through various mechanisms has changed. Who knows what will change in the future?

> However, If #752 is accepted, then this issue will be resolved (.tfvars.json are handled the same way as .tfvars even it's not mentioned in the doc).

Do `.tfvars.json` files have the same restriction as `.tfvars` files? That is, you can only set variables in them that are defined in the module? If so, that will make reusing common var files across different modules more difficult.

ekini:
> > On the other hand, if terragrunt is positioned as a thin wrapper for terraform, then why does it require a specific version of it?
> 
> Because Terraform is pre 1.0.0, so it offers no backwards compatibility guarantees on its public APIs, so to wrap it, we have to know what APIs we are wrapping. E.g., In the past, the format of the `terraform init` command changed in backwards incompatible ways; now, Terraform processes `terraform.tfvars` files differently, so we had to move away from the old file format; and both in the past and the present, the way Terraform parses variables you pass to it through various mechanisms has changed. Who knows what will change in the future?

Yes, good point. Nobody knows. Though, people still need to have a way to migrate between version somehow.
Maybe put a warning to that big README with something like "we support only XXX version of terraform at the moment, and don't guarantee backwards compatibility with older versions, but we try to keep it"

> > However, If #752 is accepted, then this issue will be resolved (.tfvars.json are handled the same way as .tfvars even it's not mentioned in the doc).
> 
> Do `.tfvars.json` files have the same restriction as `.tfvars` files? That is, you can only set variables in them that are defined in the module? If so, that will make reusing common var files across different modules more difficult.

I'll answer in #752 

Meanwhile, it's not a big change to keep it working: replace boolean `true`/`false` in the inputs with `1`/`0`. Migrating every module to 0.12 is much harder.

ekini:
I've tidied it up from extra commits, resolved the conflicts and put a note to README.

ekini:
I'd really like to get it merged, as it's a real blocker for us.

It's easy to create a custom homebrew formula with the patches, but we have people who are struggling on Linux.

brikis98:
From what I can see, this PR:

1. Adds the ability to specify a custom Terraform binary in `terragrunt.hcl`. I'm fine with this.
1. Reduces the version constraint check from 0.12 to 0.11. However, Terragrunt before 0.19.x worked with Terraform `>= v0.9.3`. This of course can be fixed.
1. Does not actually fix Terragrunt to work with 0.11. E.g., The way boolean vars are passed will not work. This is a blocker. 
1. There may well be other compatibility issues with 0.11. The only way to know is to test with 0.11, but this PR does not update the tests to run against 0.11. This is also a blocker.

ekini:
This PR doesn't specifically target the compatibility with older versions of terraform, though. It just adds another way to specify the terraform binary, along with the existing env var/cli flag options. I mean, I can use terraform v0.11 right now with new terragrunt, with the tfbinary env var set by `direnv`, but it's inconvenient and forces everyone to use it.

But it can be used with the older terraform.
Do you think I need to lower the version constraint to `>=v0.9.3`?

The reason it doesn't always work with boolean vars is related to infrastructure-as-env-var path chosen by terraform. It can easily be fixed by using `1` and `0` instead of `true` and `false`, or (a better way) by implementing #752. Which is blocked by upstream.

I usually upgrade this way: remove the terragrunt block from `terraform.tfvars`, leaving only variables. Create `terragrunt.hcl` with the terragrunt configuration without variables. Especially if there are maps inside of the variables. If there are only simple vars with numbers and strings, they can be moved to `terragrunt.hcl`.

As for the tests, I'm not sure if they are needed. Is terragrunt going to actually keep compatibility with old versions? Or just _try_ to keep it?

brikis98:
> This PR doesn't specifically target the compatibility with older versions of terraform, though. It just adds another way to specify the terraform binary, along with the existing env var/cli flag options. 

As I said, I'm totally fine with that addition. But this PR _also_ changes the Terraform version check, and that's the part that worries me.

> Is terragrunt going to actually keep compatibility with old versions? Or just try to keep it?

We do not intend to support Terraform < 12 on Terragrunt >= 19. Removing the version constraint, however, will mean that a number of users will use Terragrunt with Terraform < 12, and will inevitably hit bugs. As a maintainer, I worry about lots of bug reports and a crappy user experience that will result from this.

ekini:
I'm not sure what to do then. Without lowering the checked version this PR won't help in any way with migration to terraform0.12/terragrunt0.19.

Maybe add another config variable "no-version-check"?  But that increases complexity and won't be used by many people. Or make the version configurable, but again there is the same problem.

ekini:
How about checking the binary suffix for the version check?
Say `terraform_binary="terraform-11`, then take `11`, or the default `12` otherwise.

brikis98:
> Or make the version configurable, but again there is the same problem.

I could see something like an `override_terraform_version_check = "> 0.11"` config. So the default Terragrunt will require 0.12 or greater, but if you read the docs, you can find the `override_terraform_version_check` param, and those same docs that the user is now forced to read can also make it clear that Terragrunt is only tested with 0.12 and using it with other versions is 100% at your own risk.

ekini:
@brikis98 Ok, I've left the version check as it was before and just added another config option to specify it when needed. Called it `terraform_version_constraint` because the default one is called `DEFAULT_TERRAFORM_VERSION_CONSTRAINT`.

ekini:
@brikis98 resolved the conflicts.

_docs/migration_guides/upgrading_to_terragrunt_0.19.x.md
```diff
@@ -272,3 +273,17 @@ Two built-in functions were renamed:
 1. `get_parent_tfvars_dir()` is now called `get_parent_terragrunt_dir()`.   
 
 Make sure to make the corresponding updates in your `terragrunt.hcl` file!
+
+### Use older Terraform
+
+Although it is not officiall supported and not tested, it is still possible to use terraform<0.12 with terragrunt >=0.19.
```
brikis98:
```suggestion
Although it is not officially supported and not tested, it is still possible to use terraform<0.12 with terragrunt >=0.19.
```

cli/cli_app.go
```diff
@@ -241,10 +233,30 @@ func runTerragrunt(terragruntOptions *options.TerragruntOptions) error {
...
 
 	terragruntConfig, err := config.ReadTerragruntConfig(terragruntOptions)
+
 	if err != nil {
 		return err
 	}
 
+	// change the terraform binary path before checking the version
+	if terragruntOptions.TerraformPath == options.TERRAFORM_DEFAULT_PATH { // not changed from default
+		if terragruntConfig.TerraformBinary != "" {
```
brikis98: Can combine these two if-statements into one with `&&` between the conditions

_docs/migration_guides/upgrading_to_terragrunt_0.19.x.md
```diff
@@ -272,3 +273,17 @@ Two built-in functions were renamed:
...
+
+Just install a different version of terraform into a directory of your choice outside of `PATH` and specify path to the binary in `terragrunt.hcl` as `terraform_binary`, plus you need to lower the version check constraint:
+
+```hcl
+
+terraform_binary = "~/bin/terraform-v11/terraform"
+terraform_version_constraint = ">= 0.11"
```
brikis98: This param should be documented in the root README too

ekini: done :)

ekini:
Ok, fixed the docs, rebased onto the latest `master`.

brikis98:
Great, thanks! I'll run the tests one more time, and if they pass, merge.

brikis98:
Tests passed. Merging now.

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.19.18. Thanks @ekini!

lorengordon:
@brikis98 @ekini Many thanks for this one! It's nice being able to use latest terragrunt again, even though we haven't yet updated all our terraform to v0.12!

I also really like staged migrations that update one thing at a time. Now we can update to terragrunt v0.19, and then at our leisure update to terraform v0.12, without further breaking our configs.

PR #753, opened 2019-06-17
Parse include block first so include-specific functions work in other blocks of child config

apottere:
Before this change `path_relative_to_include`, `path_relative_from_include`, and `get_parent_terragrunt_dir` were all `"."` when called from the child config.  This is because the whole config is parsed at once, and there was no way to know the value of the `include` block before parsing.

This change splits parsing up so that `include` is parsed first, and then the rest of the config is parsed.  This allows us to use the three functions above in the child config in other areas.

Test before changes, passes with changes:
```
=== RUN   TestIncludeFunctionsWorkInChildConfig
--- FAIL: TestIncludeFunctionsWorkInChildConfig (0.00s)
    config_test.go:892: 
        	Error Trace:	config_test.go:892
        	Error:      	Not equal: 
        	            	expected: "child"
        	            	actual  : "."
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-child
        	            	+.
        	Test:       	TestIncludeFunctionsWorkInChildConfig
FAIL
```


config/config_test.go
```diff
@@ -420,16 +420,6 @@ func TestMergeConfigIntoIncludedConfig(t *testing.T) {
...
 	}{
-		{
-			&TerragruntConfig{},
-			nil,
-			&TerragruntConfig{},
-		},
-		{
-			&TerragruntConfig{RemoteState: &remote.RemoteState{Backend: "foo"}},
-			nil,
-			&TerragruntConfig{RemoteState: &remote.RemoteState{Backend: "foo"}},
```
brikis98: Why are these tests and the ones below removed?

config/config_test.go
```diff
@@ -869,6 +839,29 @@ skip = false
 	assert.Equal(t, false, terragruntConfig.Skip)
 }
 
+func TestIncludeFunctionsWorkInChildConfig(t *testing.T) {
```
brikis98: Kudos on adding a failing test cause for this issue 👍

config/config.go
```diff
@@ -231,48 +237,79 @@ func ParseConfigFile(configPath string, terragruntOptions *options.TerragruntOpt
...
+	file, err := parseHcl(configString, filename)
 	if err != nil {
 		return nil, err
 	}
-	if terragruntConfigFile == nil {
-		return nil, errors.WithStackTrace(CouldNotResolveTerragruntConfigInFile(configPath))
-	}
 
-	config, err := convertToTerragruntConfig(terragruntConfigFile, configPath, terragruntOptions)
+	terragruntInclude, err := decodeAsTerragruntInclude(file, filename, terragruntOptions)
```
brikis98: Please add comments explaining why we do this two-step parse process

brikis98: In fact, now that the parsing is more complicated, each of the subsequent steps could use some explanation

apottere:
> Thanks for the PR!
> 
> 1. Please run `make fmt` if you haven't already
> 2. Did you run all the tests?

1. Done.
2. I ran the tests in `/config`, `/configstack`, and `/cli`, but I'm unable to run all the tests because I don't have a sandbox aws account to use.

config/config.go
```diff
@@ -216,13 +222,13 @@ func ReadTerragruntConfig(terragruntOptions *options.TerragruntOptions) (*Terrag
 
 // Parse the Terragrunt config file at the given path. If the include parameter is not nil, then treat this as a config
 // included in some other config file when resolving relative paths.
-func ParseConfigFile(configPath string, terragruntOptions *options.TerragruntOptions, include *IncludeConfig) (*TerragruntConfig, error) {
-	configString, err := util.ReadFileAsString(configPath)
+func ParseConfigFile(filename string, terragruntOptions *options.TerragruntOptions, include *IncludeConfig) (*TerragruntConfig, error) {
```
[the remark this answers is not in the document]
apottere: This same string was being passed along through a bunch of different functions, and halfway through it switched from `configPath` to `filename`.  As someone new to the codebase it was confusing, since it didn't actually change at all.  So this is just making the name consistent.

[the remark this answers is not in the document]
apottere: I can always revert this change if you'd like.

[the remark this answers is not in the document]
brikis98: No worries, was just checking that this wasn't due to some bug or other issue

config/config.go
brikis98: The new comments are great, thanks!

brikis98:
Sorry for the delay! Tests passed, so merging now. Thanks for the fix!

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.19.5

apottere:
Thanks for the merge!

Issue #441, opened 2018-03-20
Is GCS supported as remote state ?


ArchiFleKs:
Hi, I'm trying to use GCS as a backend for remote state, I adapted the configuration from S3 but it keep asking me for bucket name when I run `terragrunt init` and state is store in root bucket folder despite using the same syntax as documentation

brikis98:
What is the contents of your `.tfvars` file? What about your `.tf` file?

ArchiFleKs:
I found the answer here : https://github.com/gruntwork-io/terragrunt/issues/169

The main doc is unclear or maybe it is my english but it says 
> To fill in the settings via Terragrunt, create a terraform.tfvars file in the root folder and in each of the Terraform modules

But in fact it is in the root folder of your `live` folder and inside each different environment

PR #757, opened 2019-06-18
Add support for creating & managing remote state using the GCS backend

robmorgan:
This PR adds support for creating and managing remote state using a multi-regional [Google Cloud Storage bucket](https://cloud.google.com/storage/).


robmorgan:
Related upstream issue: https://github.com/hashicorp/terraform/issues/18417

robmorgan:
Unlike AWS, GCS automatically encrypts bucket contents. Should we provide a way to disable it? In addition, data is encrypted using a Google-managed key, but in the future, we could add the option to specify customer-managed encryption keys.

robmorgan:
After speaking with @yorinasub17, we agreed to leave out the ability to disable encryption. Also, customer-managed keys can be implemented at a later date as needed by a specific customer.

robmorgan:
Implements: https://github.com/gruntwork-io/terragrunt/issues/441

robmorgan:
I've added GCP configuration to the CircleCI env now and the tests pass.

robmorgan:
@vincentdeity FYI - I've added support to Terragrunt for working with remote state using the GCS backend. It's pending review.

robmorgan:
cc @brikis98 for review

remote/remote_state_gcs.go
```diff
@@ -0,0 +1,405 @@
...
+
+	return nil
+}
+
+// CreateGCSBucketWithVersioning creates the given GCS bucket and enables versioning for it.
+func CreateGCSBucketWithVersioning(gcsClient *storage.Client, config *ExtendedRemoteStateConfigGCS, terragruntOptions *options.TerragruntOptions) error {
+	err := CreateGCSBucket(gcsClient, config, terragruntOptions)
+
+	if err != nil {
+		if isBucketAlreadyOwnedByYourError(err) {
```
brikis98: Is this using the same method from the S3 code? If so, will that work for GCS?

remote/remote_state_gcs.go
```diff
@@ -0,0 +1,405 @@
...
+	terragruntOptions.Logger.Printf("Waiting for bucket %s to be created", config.Bucket)
+	for retries := 0; retries < MAX_RETRIES_WAITING_FOR_GCS_BUCKET; retries++ {
+		if DoesGCSBucketExist(gcsClient, config) {
+			terragruntOptions.Logger.Printf("GCS bucket %s created.", config.Bucket)
+			return nil
+		} else if retries < MAX_RETRIES_WAITING_FOR_GCS_BUCKET-1 {
+			terragruntOptions.Logger.Printf("GCS bucket %s has not been created yet. Sleeping for %s and will check again.", config.Bucket, SLEEP_BETWEEN_RETRIES_WAITING_FOR_GCS_BUCKET)
+			time.Sleep(SLEEP_BETWEEN_RETRIES_WAITING_FOR_GCS_BUCKET)
+		}
+	}
```
brikis98: Any reason not to use `retry.DoWithRetry` here?

remote/remote_state_gcs.go
```diff
@@ -0,0 +1,405 @@
...
+func DoesGCSBucketExist(gcsClient *storage.Client, config *RemoteStateConfigGCS) bool {
+	ctx := context.Background()
+
+	// Creates a Bucket instance.
+	bucket := gcsClient.Bucket(config.Bucket)
+
+	// TODO - the code below attempts to determine whether the storage bucket exists by making a making a number of API
+	// calls, then attemping to list the contents of the bucket. It was adapted from Google's own integration tests and
+	// should be improved once the appropriate API call is added. For more info see:
+	// https://github.com/GoogleCloudPlatform/google-cloud-go/blob/de879f7be552d57556875b8aaa383bce9396cc8c/storage/integration_test.go#L1231
```
brikis98: Thx for including these comments! Helps clarify why the code is so complicated for something that seems simple :)

remote/remote_state_gcs.go
robmorgan: No, I'll remove it.

remote/remote_state_gcs.go
robmorgan: @brikis98 I guess because `DoesGCSBucketExist` returns a bool and not an error. I could change it to return an error and a bool or is there a cleaner way to do it in the retry func?

Issue #732, opened 2019-06-07
Is there no support for explicitly defining access_key and secret_key in terragrunt backend configuration?


KadenLNelson:
I'm trying to define my access_key and secret_key explicitly as per the example below in the `.tfvars` file.  But I'm receiving this error... 

`[terragrunt] 2019/06/06 22:02:59 Error finding AWS credentials`
Is this because Terragrunt uses the AWS library to find the credentials at the shared credentials file and environmental variables? I'm utilizing DigitalOcean spaces as an S3 compatible backend. Do I have to drink the AWS koolaid on this one?

```
# Terragrunt is a thin wrapper for Terraform that provides extra tools for working with multiple Terraform modules,
# remote state, and locking: https://github.com/gruntwork-io/terragrunt
terragrunt = {
  # Configure Terragrunt to automatically store tfstate files in an S3 bucket
  remote_state {
    backend = "s3"

    config {
      skip_credentials_validation = true
      skip_metadata_api_check     = true
      access_key = "ACCESSKEY"
      secret_key = "SECRETKEY"
      endpoint = "https://sfo2.digitaloceanspaces.com"
      region = "us-east-1"
      bucket = "..."
      key = "${path_relative_to_include()}/terraform.tfstate"
    }
  }

  # Configure root level variables that all resources can inherit
  terraform {
    extra_arguments "-var-file" {
      commands = ["${get_terraform_commands_that_need_vars()}"]

      optional_var_files = [
        "${get_tfvars_dir()}/${find_in_parent_folders("account.tfvars", "ignore")}",
      ]
    }
  }
}
```

yorinasub17:
Hi @KadenLNelson,

This is a use case we haven't addressed yet: using the s3 backend with something that is not AWS. What is happening is that because you are using the `s3` backend, you are triggering the workflow where `terragrunt` tries to auto create the bucket (See: https://github.com/gruntwork-io/terragrunt#create-remote-state-and-locking-resources-automatically). In this workflow, `terragrunt` makes API calls to try to determine if the bucket needs to be precreated.

I think there are two issues here:

[...]

Issue #597, opened 2018-11-01
validate-all requires credentials


adborden:
I'd like to run some simple lint-style checks on CI. When using terraform, I usually use `validate`, so for terragrunt `validate-all` seems like the correct choice. However, it fails, complaining that the credentials are not set.

[full log](https://gist.github.com/adborden/7b5286249c3273ed6465270fd2fba2f8)
```
Error finding AWS credentials (did you set the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environment variables?): NoCredentialProviders: no valid providers in chain. Deprecated.
	For verbose messaging see aws.Config.CredentialsChainVerboseErrors
Cannot process module Module /home/adborden/projects/datagov/datagov-infrastructure-live/stage/db (dependencies: [/home/adborden/projects/datagov/datagov-infrastructure-live/stage/vpc]) because one of its dependencies, Module /home/adborden/projects/datagov/datagov-infrastructure-live/stage/vpc (dependencies: []), finished with an error: Error finding AWS credentials (did you set the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environment variables?): NoCredentialProviders: no valid providers in chain. Deprecated.
	For verbose messaging see aws.Config.CredentialsChainVerboseErrors
[terragrunt] 2018/11/01 08:32:01 Unable to determine underlying exit code, so Terragrunt will exit with error code 1
```

adborden:
I think part of the issue is that terragrunt does a backend initialize on every command. With terraform, you can run `terraform init -backend=false` which allows you to continue with the validate even without credentials. This doesn't work with terragrunt.

stefansedich:
Is there any workarounds with this one @adborden? have run into this myself.

lukasmrtvy:
@adborden  is even possible to validate terraform somehow?  
because `terraform init -backend=false && terraform validate` will not work, cuz .tfvars will not be validated and .tf files will not get variables in this case.

IMHO terragrunt should handle this.

stefansedich:
@adborden mind checking out #761? I think we can close this one off.

You should be able to use something like the following and set the environment variable during your CI process.

```
remote_state {
  disable_init = tobool(get_env("DISABLE_INIT", "false"))
}
```

c33s:
sadly `disable_init` does not work for dependencies https://github.com/gruntwork-io/terragrunt/issues/2603 (workaround included)

c33s:
unstale

PR #761, opened 2019-06-24
Provide ability to disable backend init so that validate-all can be run without a backend

stefansedich:
As an attempt to solve https://github.com/gruntwork-io/terragrunt/issues/597 and https://github.com/gruntwork-io/terragrunt/issues/732 this change provides the abilty to disable a backend using something like this, I set DISABLE_BACKEND in my CI process when running validate-all.

```
remote_state {
  backend = "s3"
  enabled = get_env("DISABLE_BACKEND", "false") != "true"
  ...
}
```

@yorinasub17 I ran with your idea but ended up in a slightly different direction, would appreicate some feedback on if this is close to what we want or if I should consider another solution as I would love to get this working so I can add an validate-all step to our CI process that does not require a backend.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -55,6 +55,7 @@ type terragruntInclude struct {
 // Configuration for Terraform remote state as parsed from a terragrunt.hcl config file
 type remoteStateConfigFile struct {
 	Backend string    `hcl:"backend,attr"`
+	Enabled *bool     `hcl:"enabled,attr"`
 	Config  cty.Value `hcl:"config,attr"`
 }
 
@@ -520,10 +521,15 @@ func convertToTerragruntConfig(terragruntConfigFromFile *terragruntConfigFile, c
 		}
 
 		remoteState := &remote.RemoteState{}
+		remoteState.FillDefaults()
+
 		remoteState.Backend = terragruntConfigFromFile.RemoteState.Backend
 		remoteState.Config = remoteStateConfig
 
-		remoteState.FillDefaults()
+		if terragruntConfigFromFile.RemoteState.Enabled != nil {
+			remoteState.Enabled = *terragruntConfigFromFile.RemoteState.Enabled
+		}
+
 		if err := remoteState.Validate(); err != nil {
 			return nil, err
 		}
--- a/config/config.go
+++ b/config/config.go
@@ -54,9 +54,9 @@ type terragruntInclude struct {
 
 // Configuration for Terraform remote state as parsed from a terragrunt.hcl config file
 type remoteStateConfigFile struct {
-	Backend string    `hcl:"backend,attr"`
-	Enabled *bool     `hcl:"enabled,attr"`
-	Config  cty.Value `hcl:"config,attr"`
+	Backend     string    `hcl:"backend,attr"`
+	DisableInit *bool     `hcl:"disable_init,attr"`
+	Config      cty.Value `hcl:"config,attr"`
 }
 
 func (remoteState *remoteStateConfigFile) String() string {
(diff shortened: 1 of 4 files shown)


stefansedich:
I had not written any tests will go back and cover that off if the direction is agreed upon.

config/config.go
```diff
@@ -520,10 +521,15 @@ func convertToTerragruntConfig(terragruntConfigFromFile *terragruntConfigFile, c
 		}
 
 		remoteState := &remote.RemoteState{}
+		remoteState.FillDefaults()
```
stefansedich: It appeared to me that `FillDefaults` should probably be called before setting any attributes? It was unused so was not fully sure of it's intention, I moved it up to execute earlier so that I could default Enabled to true when not present in config.

stefansedich:
I changed my original approach and moved from adding an attribute on remote_state and moving it up to the backend itself but am finding there is some test pain going around and setting `Enabled = true` around the place.

I will hold off doing that for now incase this approach is not how we want to tackle this issue.

yorinasub17:
I am a little confused by what is supposed to happen when `enabled = false`. Intuitively to me, this should mean that the remote state config should not be used at all in any `terraform` command. But it looks like the current implementation is to only disable all the initialization. Is that correct and intentional?

If so, I might rename the `enabled` var to `enable_initialization` or something like that so that it is clearer what is supposed to happen. Otherwise, this approach makes sense to me.

Would love a second opinion from one of the codeowners though: @autero1 @brikis98 @eak12913

stefansedich:
Agreed @yorinasub17 on second glance myself I can see the confusion and agree that `enable_init` might make more sense, I could even switch it to `disable_init` which would solve my issue of having to go and default things to true in all tests which should make things easier.

stefansedich:
@yorinasub17 I went ahead and moved to `disable_init` I think it makes much more sense!

yorinasub17:
Sorry for the radio silence here, got pulled away to other things. Did a quick review of the new things and it looks good except for the `terragrunt` binary that was checked in. Can you make a commit that removes that? Thanks!

Going to kick off the build now since that doesn't depend on that binary.

stefansedich:
Oops @yorinasub17 the binary has been removed!

lorengordon:
You might want to rebase/amend [the commit](https://github.com/gruntwork-io/terragrunt/pull/761/commits/d8600fcf50df96b61ebf837bffcd5dba9461b166) that added the binary, instead. Or a maintainer can squash/rebase when merging. Otherwise the binary will end up in the repo commit history anyway...

yorinasub17:
@lorengordon Yup will squash when merging. The only reason why I wanted the additional commit was so that I can see the build status show up on 61f5345 . Otherwise, the commit squashing will create a new history and I will have to dig up the build status.

Looks like the build passed, so going to go ahead and merge + release this. Thanks for the contribution!

yorinasub17:
Woops looks like there was a merge conflict that git failed to detect, due to the change in signature.

stefansedich:
I hope I did not mess you up @yorinasub17, I must have pushed my rebase just as you clicked merge on your side.

lorengordon:
Sorry! I was trying to help keep the repo clean of binaries, but apparently you had it under control and I just made the merge more complicated! :facepalm:

yorinasub17:
No worries! The merge conflict wasn't actually because of the squashing or anything, but rather because of feature conflicts (the GCS initializer feature wasn't updated to match the new function signatures introduced here). These are harder to catch because they don't show up as conflicts in git.

I have a branch that fixes it here: https://github.com/gruntwork-io/terragrunt/pull/769

stefansedich:
Thanks @yorinasub17 my bad there I should have picked up on this too.

yorinasub17:
Released: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.19.7

cpxPratik:
Should not readme.md be updated? I do not see `disable_init` config option on readme file. Unless one look into source code, there is no way to know if that option is available.

stefansedich:
Good idea @cpxPratik I will add it to my list of things to tackle this week.

yorinasub17:
Good suggestion @cpxPratik ! That would be great if you can handle that @stefansedich . Looking forward to the PR!

PR #763, opened 2019-06-25
Add tf12 badge

yorinasub17:
This adds the tf12 version badge to better indicate that this only works with tf12 and above.


yorinasub17:
Merging. Thanks for review!

PR #769, opened 2019-07-01
Fix merge conflict from new GCS initializer and disable_init config

yorinasub17:
This fixes a conflict in features between https://github.com/gruntwork-io/terragrunt/pull/761 and https://github.com/gruntwork-io/terragrunt/pull/757

Specifically, the initializer interface was changed to accommodate the new `disable_init` feature, but that was only applied on the S3 initializer in the branch. Since the branch didn't have the GCS initializer, it wasn't updated there and so when merged, `master` broke due to a type error.


yorinasub17:

robmorgan:
Thx @yorinasub17

PR #773, opened 2019-07-02
Adding documentation around the new disable_init feature

stefansedich:
Adding some documentation around the disable_init feature as mentioned in #761


stefansedich:
Good idea @barryib just added a snippet showing this.

Issue #764, opened 2019-06-26
.terragrunt-cache fails to update with new source "...?ref=..."


tomalok:
I'm updating my `terragrunt.hcl`'s `terraform { source = "..." } }` value with new `...?ref=<commit>` values as I make adjustments, but terragrunt isn't successfully pulling the new commit from the git repo.  Instead, it complains...
```
[terragrunt] [/Users/jake/git/codecommit.us-west-2/jn-tf/jakesys/regions/us-west-2/vpcs/dev] 2019/06/25 18:04:19 Running command: terraform --version
[terragrunt] 2019/06/25 18:04:19 Reading Terragrunt config file at /Users/jake/git/codecommit.us-west-2/jn-tf/jakesys/regions/us-west-2/vpcs/dev/terragrunt.hcl
[terragrunt] 2019/06/25 18:04:19 Cleaning up contents of download folder /Users/jake/git/codecommit.us-west-2/jn-tf/jakesys/regions/us-west-2/vpcs/dev/.terragrunt-cache/QM1-63WWHsRnztH6ooxLgsdRMGg/tieM-yGmC7c1rfHpXzt3T9uRVsA
[terragrunt] 2019/06/25 18:04:19 Downloading Terraform configurations from git::ssh://git-codecommit.us-west-2.amazonaws.com/v1/repos/jn-tf-mod?ref=86e6f62 into /Users/jake/git/codecommit.us-west-2/jn-tf/jakesys/regions/us-west-2/vpcs/dev/.terragrunt-cache/QM1-63WWHsRnztH6ooxLgsdRMGg/tieM-yGmC7c1rfHpXzt3T9uRVsA
[terragrunt] 2019/06/25 18:04:19 Hit multiple errors:
error downloading 'ssh://git-codecommit.us-west-2.amazonaws.com/v1/repos/jn-tf-mod?ref=86e6f62': /usr/local/bin/git exited with 1: error: pathspec 'master' did not match any file(s) known to git

[terragrunt] 2019/06/25 18:04:19 Unable to determine underlying exit code, so Terragrunt will exit with error code 1
```
The hash subdirectories under `.terraform-cache` are **always** the same:
```
git::ssh://git-codecommit.us-west-2.amazonaws.com/v1/repos/jn-tf-mod//vpc?ref=6e30e8a
.terragrunt-cache
└── QM1-63WWHsRnztH6ooxLgsdRMGg
[...]

PR #774, opened 2019-07-03
For local file copy create a manifest with a list of files.

ekini:
That allows to safely delete only the files we copied on the previous
copy, and not worry about stale files still living in the working directory.

Also removes some unused functions.

Initially I wanted to compare source and destination directories, but we use [CopyFolderContentsWithFilter](https://github.com/gruntwork-io/terragrunt/blob/master/util/file.go#L150) at least two times, so this will not work.

The only way I see is to create a list of files we touch, and delete them just before copying new files.

Used [gob](https://golang.org/pkg/encoding/gob) as the simplest serialiser. Added some tests.


cli/download_source.go
```diff
@@ -57,7 +56,7 @@ func downloadTerraformSource(source string, terragruntOptions *options.Terragrun
 	}
 
 	terragruntOptions.Logger.Printf("Copying files from %s into %s", terragruntOptions.WorkingDir, terraformSource.WorkingDir)
-	if err := util.CopyFolderContents(terragruntOptions.WorkingDir, terraformSource.WorkingDir); err != nil {
+	if err := util.CopyFolderContents(terragruntOptions.WorkingDir, terraformSource.WorkingDir, ".terragrunt-module-manifest"); err != nil {
```
brikis98: `".terragrunt-module-manifest"` should be stored in a `const`

util/file.go
```diff
@@ -139,18 +131,26 @@ func ReadFileAsString(path string) (string, error) {
 
 // Copy the files and folders within the source folder into the destination folder. Note that hidden files and folders
 // (those starting with a dot) will be skipped.
-func CopyFolderContents(source string, destination string) error {
-	return CopyFolderContentsWithFilter(source, destination, func(path string) bool {
+func CopyFolderContents(source, destination, manifestFile string) error {
```
brikis98: Comment on function should be updated

util/file.go
```diff
@@ -139,18 +131,26 @@ func ReadFileAsString(path string) (string, error) {
...
+	return CopyFolderContentsWithFilter(source, destination, manifestFile, func(path string) bool {
 		return !PathContainsHiddenFileOrFolder(path)
 	})
 }
 
 // Copy the files and folders within the source folder into the destination folder. Pass each file and folder through
 // the given filter function and only copy it if the filter returns true.
-func CopyFolderContentsWithFilter(source string, destination string, filter func(path string) bool) error {
+func CopyFolderContentsWithFilter(source, destination, manifestFile string, filter func(path string) bool) error {
```
brikis98: Comment on function should be updated

util/file.go
```diff
@@ -139,18 +131,26 @@ func ReadFileAsString(path string) (string, error) {
...
 	// Why use filepath.Glob here? The original implementation used ioutil.ReadDir, but that method calls lstat on all
 	// the files/folders in the directory, including files/folders you may want to explicitly skip. The next attempt
 	// was to use filepath.Walk, but that doesn't work because it ignores symlinks. So, now we turn to filepath.Glob.
```
brikis98: Comment should be moved above actual `filepath.Glob` call

util/file.go
```diff
@@ -189,10 +189,13 @@ func CopyFolderContentsWithFilter(source string, destination string, filter func
 			if err := CopyFile(file, dest); err != nil {
 				return err
 			}
+			if err := manifest.AddFile(dest); err != nil {
```
brikis98: The manifest file doesn't track folders? Just files?

util/file.go
```diff
@@ -253,20 +256,68 @@ func JoinTerraformModulePath(modulesFolder string, path string) string {
...
+	Path       string
+	encoder    *gob.Encoder
+	fileHandle *os.File
+}
 
-	for _, pathPart := range pathParts {
-		if pathPart == folderName {
-			return true
+// Clean will remove all files specified in the manifest
+func (f *fileManifest) Clean() error {
```
brikis98: Please don't use single-letter variable names. They are hard to follow more than 1-2 lines away from the declaration.

util/file.go
```diff
@@ -253,20 +256,68 @@ func JoinTerraformModulePath(modulesFolder string, path string) string {
...
+
+	// if manifest file doesn't exist, just exit
+	if !FileExists(f.Path) {
+		return nil
+	}
+	file, err := os.Open(f.Path)
+	if err != nil {
+		return err
+	}
+	decoder := gob.NewDecoder(file)
```
brikis98: Sorry, I'm probably missing something obvious, but what is the role of the `decoder` here?

util/file.go
```diff
@@ -189,10 +189,13 @@ func CopyFolderContentsWithFilter(source string, destination string, filter func
...
 				return err
 			}
+			if err := manifest.AddFile(dest); err != nil {
+				return err
+			}
 		}
 	}
 
-	return nil
+	return manifest.Close()
```
brikis98: Should this be in a `defer` to ensure the file is closed even if we hit an error part way through?

util/file_test.go
```diff
@@ -120,31 +121,27 @@ func TestJoinTerraformModulePath(t *testing.T) {
...
-	for _, testCase := range testCases {
-		// The following is necessary to make sure testCase's values don't
-		// get updated due to concurrency within the scope of t.Run(..) below
-		testCase := testCase
-		t.Run(testCase.description, func(t *testing.T) {
-			actual := PathContains(testCase.path, testCase.folderName)
-			assert.Equal(t, testCase.expected, actual)
-		})
+	manifest := newFileManifest(".terragrunt-test-manifest")
+	assert.Nil(t, manifest.Create())
```
brikis98: `require.Nil`

util/file_test.go
```diff
@@ -120,31 +121,27 @@ func TestJoinTerraformModulePath(t *testing.T) {
...
+	assert.FileExists(t, manifest.Path)
+	for _, file := range testfiles {
+		assert.Nil(t, manifest.AddFile(file))
+	}
+
+	assert.Nil(t, manifest.Clean())
```
brikis98: `require.Nil`

util/file_test.go
```diff
@@ -120,31 +121,27 @@ func TestJoinTerraformModulePath(t *testing.T) {
...
-		expected    bool
-	}{
-		{"both empty", "", "", true},
-		{"path empty, folder name not empty", "", "foo", false},
-		{"path non empty, folder name empty", "/foo", "", true},
-		{"path contains one part that matches folder name", "foo", "foo", true},
-		{"path contains one part that doesn't match folder name", "foo", "bar", false},
-		{"path contains multiple parts, one of which matches folder name", "/foo/bar/baz", "bar", true},
-		{"path contains multiple parts, none of which match folder name", "/foo/bar/baz", "will-not-match", false},
+	var testfiles = []string{"file1", "file2"}
```
brikis98: Thanks for adding tests! 👍 

However, I'd strongly recommend having tests for all the corner cases, as those are what'll kill us here:

1. Empty folder
1. Nested files/folders
1. Nested folders, some of which are empty
1. Etc

cli/file_copy_getter.go
```diff
@@ -30,7 +31,7 @@ func (g *FileCopyGetter) Get(dst string, u *url.URL) error {
 		return fmt.Errorf("source path must be a directory")
 	}
 
-	return util.CopyFolderContents(path, dst)
+	return util.CopyFolderContents(path, dst, ".terragrunt-source-manifest")
```
brikis98: Shouldn't this be an absolute path, not a relative one? It seems important that the manifest file is stored in the proper folder!

util/file.go
```diff
@@ -253,20 +256,68 @@ func JoinTerraformModulePath(modulesFolder string, path string) string {
...
 
-// Returns true if the given path contains the given folder name.
-//
-// Examples:
-//
-// PathContains("/foo/bar", "foo") => returns true
-// PathContains("/foo/bar", "baz") => returns false
-func PathContains(path string, folderName string) bool {
-	pathParts := strings.Split(path, string(filepath.Separator))
+type fileManifest struct {
```
brikis98: Please add comments explaining what this is, why it exists, and how it's meant to be used.

util/file.go
```diff
@@ -253,20 +256,68 @@ func JoinTerraformModulePath(modulesFolder string, path string) string {
...
+	// remove the manifest itself
+	if err := os.RemoveAll(f.Path); err != nil {
+		return errors.WithStackTrace(err)
+	}
 
-	return false
+	return nil
+}
+
+func (f *fileManifest) Create() error {
```
brikis98: Please add comments to these methods

cli/download_source.go
ekini: Yes, I was thinking about this too. Sounds reasonable, will fix that

util/file.go
ekini: Yes, because on directories, it calls `CopyFolderContentsWithFilter` recursively, and subfolders will have their own manifests.

util/file.go
ekini: The whole point of `gob` is that it provides ability to decode filenames one by one, instead of load the file in memory at once.

cli/file_copy_getter.go
ekini: Oh, it's not a path, but a name. It's stored in the `dst` folder.

ekini:
I've had another look and it seems the code didn't handle subfolders correctly. I'll fix that and add more tests with corner cases.

util/file.go
ekini: Now it tracks directories as well.

util/file.go
ekini: done :)

ekini:
@brikis98 added more tests, fixed subfolders handling, looks much better to me now :)

mattford63:
Hi, this doesn't seem to fix the issue I'm experiencing from the original issue.  I have to still clear the cache and the error produced is as per the originally reported bug.

https://github.com/gruntwork-io/terragrunt/issues/764#issuecomment-511704040

Sorry I can't help more at the moment.

ekini:
Ok, found another bug when the destination directory is not created, but it tries to create a manifest file there. Fixed

ekini:
I've created a homebrew formula that installs terragrunt with this patch, along with `terraform` version that doesn't complain about unset variables. Just if anyone wants to try it.
```bash
$brew install springload/tools/terragrunt
```

cli/file_copy_getter.go
```diff
@@ -30,7 +34,7 @@ func (g *FileCopyGetter) Get(dst string, u *url.URL) error {
 		return fmt.Errorf("source path must be a directory")
 	}
 
-	return util.CopyFolderContents(path, dst)
+	return util.CopyFolderContents(path, dst, SOURCE_MANIFEST_NAME)
```
brikis98: Should this be using a different constant than `MODULE_MANIFEST_NAME`?

util/file.go
```diff
@@ -253,20 +264,110 @@ func JoinTerraformModulePath(modulesFolder string, path string) string {
...
+// This allows to clean those files on subsequent runs.
+// The problem is as follows: terragrunt copies the terraform source code first to "working directory" using go-getter,
+// and then copies all files from the directory it runs from to the above dir.
+// It works fine on the first run, but if we delete a file from the current terragrunt directory, we want it
+// to be cleaned in the "working directory" as well. Since we don't really know what can get copied by go-getter,
+// we have to track all the files we tough in a manifest. This way we know exactly which files we need to clean on
+// subsequent runs.
+type fileManifest struct {
+	Path         string
+	ManifestFile string
```
brikis98: What is `Path` vs `ManifestFile`? Is the former the folder path and the latter the path to the manifest file within that folder?

util/file.go
```diff
@@ -253,20 +264,110 @@ func JoinTerraformModulePath(modulesFolder string, path string) string {
...
+func (manifest *fileManifest) clean(path string) error {
+	manifestPath := filepath.Join(path, manifest.ManifestFile)
+
+	// if manifest file doesn't exist, just exit
+	if !FileExists(manifestPath) {
+		return nil
+	}
+	file, err := os.Open(manifestPath)
+	if err != nil {
+		// the manifest doesn't exist, so there's nothing we can safely clean
```
brikis98: Didn't you check for manifest existing just one line above? Seems redundant to have the same check twice.

util/file.go
```diff
@@ -253,20 +264,110 @@ func JoinTerraformModulePath(modulesFolder string, path string) string {
...
+			} else {
+				return err
+			}
+		}
+		if manifestEntry.IsDir {
+			if err := manifest.clean(manifestEntry.Path); err != nil {
+				return errors.WithStackTrace(err)
+			}
+		} else {
+			if err := os.RemoveAll(manifestEntry.Path); err != nil {
```
brikis98: If `manifestEntry` is a file path, shouldn't this use `os.Remove` instead of `os.RemoveAll`?

util/file.go
```diff
@@ -253,20 +264,110 @@ func JoinTerraformModulePath(modulesFolder string, path string) string {
...
+				return errors.WithStackTrace(err)
+			}
+		}
+	}
+	if err := os.RemoveAll(manifestPath); err != nil {
```
brikis98: Same here: `os.Remove` for files?

util/file.go
```diff
@@ -253,20 +264,110 @@ func JoinTerraformModulePath(modulesFolder string, path string) string {
...
+		return errors.WithStackTrace(err)
+	}
+	// remove the manifest itself
+	// it will run after the close defer
+	defer os.RemoveAll(manifestPath)
```
brikis98: Wait, the line above removes `manifestPath`... Why do it again?

util/file.go
```diff
@@ -253,20 +264,110 @@ func JoinTerraformModulePath(modulesFolder string, path string) string {
...
+}
+
+// Create will create the manifest file
+func (manifest *fileManifest) Create() error {
+	var err error
+	manifest.fileHandle, err = os.OpenFile(filepath.Join(manifest.Path, manifest.ManifestFile), os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0644)
+	if err != nil {
+		return err
+	}
+	manifest.encoder = gob.NewEncoder(manifest.fileHandle)
```
brikis98: Very minor nit, but it seems slightly cleaner to do:

```go
fileHandle, err := os.OpenFile(...)
if err != nil {
  return err
}
manifest.fileHandle = fileHandle
manifest.encoder = gob.NewEncoder(...)
```

cli/file_copy_getter.go
ekini: Yes, because if a local source is used, then two separate manifest files need to be created: one for the local source, and the other for the terragrunt module directory.

util/file.go
ekini: Yes, fixed that

util/file.go
ekini: I must have been drunk. Fixed

util/file.go
ekini: Renamed `Path` to `ManifestFolder` and added comments to each var. It used to be just `Path` before handling the subdirectories properly.

util/file.go
ekini: removed the extra remove 👍

util/file.go
ekini: Done :) Not sure why it's not showing as "Outdated" on github.

cli/file_copy_getter.go
brikis98: Ohhh, I gotcha. I'll update one comment to make that clearer.

cli/download_source.go
```diff
@@ -16,9 +16,11 @@ import (
 	"github.com/gruntwork-io/terragrunt/util"
 	"github.com/hashicorp/go-getter"
 	urlhelper "github.com/hashicorp/go-getter/helper/url"
-	"github.com/mattn/go-zglob"
 )
 
+// manifest for files coped from terragrunt module folder
```
brikis98:
```suggestion
// manifest for files coped from terragrunt module folder (i.e., the folder that contains the current terragrunt.hcl)
```

cli/file_copy_getter.go
```diff
@@ -2,12 +2,16 @@ package cli
...
-	"github.com/gruntwork-io/terragrunt/util"
-	"github.com/hashicorp/go-getter"
 	"net/url"
 	"os"
+
+	"github.com/gruntwork-io/terragrunt/util"
+	"github.com/hashicorp/go-getter"
 )
 
+// manifest for files copied from terraform source
```
brikis98:
```suggestion
// manifest for files copied from the URL specified in the terraform { source = "<URL>" } config
```

brikis98:
Unit tests failed:

```
--- FAIL: TestDownloadTerraformSourceFromLocalFolderWithManifest (0.00s)
    --- PASS: TestDownloadTerraformSourceFromLocalFolderWithManifest/test-stale-file-exists (0.00s)
    --- PASS: TestDownloadTerraformSourceFromLocalFolderWithManifest/test-stale-file-doesnt-exist-after-source-update (0.00s)
    --- PASS: TestDownloadTerraformSourceFromLocalFolderWithManifest/test-tffile-exists-in-subfolder (0.00s)
    --- PASS: TestDownloadTerraformSourceFromLocalFolderWithManifest/test-tffile-doesnt-exist-in-subfolder (0.00s)
    --- FAIL: TestDownloadTerraformSourceFromLocalFolderWithManifest/test-empty-folder-gets-copied (0.00s)
        require.go:17: 
            	Error Trace:	download_source_test.go:213
            	Error:      	Condition failed!
            	Test:       	TestDownloadTerraformSourceFromLocalFolderWithManifest/test-empty-folder-gets-copied
```

ekini:
@brikis98 I can't reproduce it. It works on my machine ™️ , plus I took a plain golang image:
```
docker run -ti --rm -v "$(pwd):/go/src/github.com/gruntwork/terragrunt" golang:1.12-alpine
/go # apk add git g++ gcc terraform > /dev/null && go get -u github.com/golang/dep/cmd/dep && cd src/github.com/gruntwork/terragrunt/
/go/src/github.com/gruntwork/terragrunt # dep ensure
/go/src/github.com/gruntwork/terragrunt # go test ./cli
ok      github.com/gruntwork/terragrunt/cli     29.499s
```
I'd be glad to debug the CircleCI build, but the docker image it uses seems to be private.

brikis98:
Perhaps it's a race condition? Is `TestDownloadTerraformSourceFromLocalFolderWithManifest` copying lots of files/folders to a common destination and perhaps something is overwriting something else?

ekini:
Doesn't seem to be a race condition. Not sure what's happening there. I haven't been able to reproduce it yet.

My wild guess is something wrong with temp folder in the Docker image.

brikis98:
I'm probably missing something, but how is [this test case supposed](https://github.com/springload/terragrunt/blob/fix/file_copy/cli/download_source_test.go#L196-L200) to work? 

It looks like you're calling [copyFolder](https://github.com/springload/terragrunt/blob/fix/file_copy/cli/download_source_test.go#L212) on `"../test/fixture-manifest/version-5-empty-subfolder"`, which [solely contains a main.tf file](https://github.com/springload/terragrunt/tree/fix/file_copy/test/fixture-manifest/version-5-empty-subfolder). The test is then checking for a [sub2 folder](https://github.com/springload/terragrunt/blob/fix/file_copy/cli/download_source_test.go#L199). Where would that folder come from?

ekini:
That's a very good point! It explains everything: when I tested it in a fresh Docker container I copied the codebase I had including the `sub2` folder, which obviously doesn't get committed to git. I've modified the test to create a temp folder with another folder inside, and use it as a source.

brikis98:
OK, great, kicking off tests again

brikis98:
Integration tests failed now:

```
Not on branch master, setting the short flag to skip long-running tests.
Running Go tests in /go/src/github.com/gruntwork-io/terragrunt/ with command: go test -v -timeout 45m -parallel 128 github.com/gruntwork-io/terragrunt/test github.com/gruntwork-io/terragrunt/test/helpers  -short
# github.com/gruntwork-io/terragrunt/test [github.com/gruntwork-io/terragrunt/test.test]
test/integration_test.go:35:2: assert redeclared as imported package name
	previous declaration at test/integration_test.go:21:2
test/integration_test.go:36:2: require redeclared as imported package name
	previous declaration at test/integration_test.go:22:2
test/integration_test.go:37:2: iterator redeclared as imported package name
	previous declaration at test/integration_test.go:23:2
FAIL	github.com/gruntwork-io/terragrunt/test [build failed]
?   	github.com/gruntwork-io/terragrunt/test/helpers	[no test files]
Exited with code 2
```

ekini:
That was result of bad resolving of a conflict. I'm sure I've seen the CircleCI build process before, but can't see it now.

brikis98:
Re-running tests.

brikis98:
OK, all tests passed. Merging now. Thank you!

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.19.13

PR #787, opened 2019-07-16
Option to suppress output of run_cmd() for sensitive commands

jessebye:
Adds an option to suppress the output of `run_cmd()` for sensitive commands. In our case, we're encrypting some values with `sops` and then using `run_cmd()` to decrypt them and pass them in to terraform.

To use this option, the first argument passed to `run_cmd()` should be `--terragrunt-quiet`.

More details in the README updates.


jessebye:
@brikis98 this is the fix we discussed 😁 tests are passing, and tested on some of our terraform code successfully!

README.md
```diff
@@ -1783,6 +1783,17 @@ remote_state {
...
+If the command you are running has the potential to output sensitive values, you may wish to redact the output from
+appearing in the terminal. To do so, use the special `--terragrunt-quiet` argument which must be passed as the first 
+argument to `run_cmd()`:
+```hcl
+super_secret_value = run_cmd("--terragrunt-quiet", "./decrypt_secret.sh", "foo") 
+```
+
+**Note:** This will prevent terragrunt from displaying the output from the command in its output. 
+However, the value could still be displayed in the Terraform output if Terraform does not treat it as a 
+[sensitive value](https://www.terraform.io/docs/configuration/outputs.html#sensitive-suppressing-values-in-cli-output).
```
brikis98: Thanks for updating the docs 👍

brikis98:
Tests passed. Merging now.

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.19.9

Issue #397, opened 2017-12-18
Getting "Error with plan:" for (nearly) all modules


robinbowes:
I have a tree of 89 terragrunt configurations, ie. (`terraform.tfvars`) files and I'm seeing terragrunt report errors like this for 86/89 of them:

```
[terragrunt] 2017/12/18 10:31:25 Error with plan: [terragrunt] [/path/to/config/main-account/us-west-2/stages/qa/oauth/eb_env_worker] 2017/12/18 10:30:14 Initializing remote state for the s3 backend
```

I suspect this is actually two separate lines of output, ie. the two strings that start with `[terragrunt]`. The first instance comes from line 65 of configstack/stack.go.

I'd do some debugging, etc. but I can't figure out how to build terragrunt locally? (I'll raise another ticket for that).

robinbowes:
I forgot to add that everything seems to actually work OK, ie. this is a false positive error.

brikis98:
Any chance you could post more of the output to help debug?

robinbowes:
There's not a lot to see - just the regular tg/tf output, which is why I want to build a modified version with additional debugging output in `configstack/stack.go`

robinbowes:
I've added some debugging statements:

```
 64     } else if errorStream.Len() > 0 {
 65       terragruntOptions.Logger.Printf("errorStream.Len(): %d\n", errorStream.Len())
 66       terragruntOptions.Logger.Printf("len(output): %d\n", len(output))
 67       terragruntOptions.Logger.Printf("Error with plan:\n== start of output ==\n'%s'== end of output ==\n", output)
```

and am seeing output like this:
```
[terragrunt] 2017/12/18 17:01:12 errorStream.Len(): 3350
[terragrunt] 2017/12/18 17:01:12 len(output): 3350
[terragrunt] 2017/12/18 17:01:12 Error with plan:
== start of output ==
[terragrunt] [/Users/robin/Workspace/foo/Terraform/mps-tf-live/main-account/us-west-2/stages/rbtest/backend/eb_env] 2017/12/18 17:00:44 Initializing remote state for the s3 backend
[...]

PR #789, opened 2019-07-21
Fix error output in plan-all..

mariux:
tries to fix issues mentioned in #397: 

when runnign `plan-all` every run states "Errors with plan".. which is due to capturing wrong output.
This patch series fixes several issues and add minor improvements in output:
- fix wrongly detected error by not capturing output for non terraform commands
- fix wrongly captured output for "init-from-module" pseudo-command run
- improve plan-all error output by adding module path of captured error
- improve plan-all output by displaying errors while happening and repeating them in final summary.

it's my first go code ;) so feel free to just take it as a hint how this was fixed for me.

### DIFF
--- a/cli/download_source.go
+++ b/cli/download_source.go
@@ -95,6 +95,7 @@ func downloadTerraformSourceIfNecessary(terraformSource *TerraformSource, terrag
 	// before and after hooks (if any).
 	terragruntOptionsForDownload := terragruntOptions.Clone(terragruntOptions.TerragruntConfigPath)
 	terragruntOptionsForDownload.TerraformCommand = CMD_INIT_FROM_MODULE
+	terragruntOptionsForDownload.Logger = terragruntOptions.Logger
 	downloadErr := runActionWithHooks("download source", terragruntOptionsForDownload, terragruntConfig, func() error {
 		return downloadSource(terraformSource, terragruntOptions, terragruntConfig)
 	})
--- a/shell/run_shell_cmd.go
+++ b/shell/run_shell_cmd.go
@@ -61,6 +61,10 @@ func RunShellCommandWithOutput(terragruntOptions *options.TerragruntOptions, wor
 		outWriter = os.Stdout
 	}
 
+	if terragruntOptions.ErrWriter != os.Stderr {
+		errWriter = io.MultiWriter(terragruntOptions.ErrWriter, os.Stderr)
+	}
+
 	if workingDir == "" {
 		cmd.Dir = terragruntOptions.WorkingDir
 	} else {
--- a/shell/run_shell_cmd.go
+++ b/shell/run_shell_cmd.go
@@ -57,7 +57,8 @@ func RunShellCommandWithOutput(terragruntOptions *options.TerragruntOptions, wor
 	// command requested by the user. The output of these other commands should not end up on stdout as this
 	// breaks scripts relying on terraform's output.
 	if !reflect.DeepEqual(terragruntOptions.TerraformCliArgs, args) {
-		outWriter = terragruntOptions.ErrWriter
+		errWriter = os.Stderr
+		outWriter = os.Stdout
 	}
 
 	if workingDir == "" {
--- a/configstack/stack.go
+++ b/configstack/stack.go
@@ -62,7 +62,7 @@ func (stack *Stack) summarizePlanAllErrors(terragruntOptions *options.Terragrunt
 				)
 			}
 		} else if errorStream.Len() > 0 {
-			terragruntOptions.Logger.Printf("Error with plan: %s", output)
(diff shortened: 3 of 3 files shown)


cli/download_source.go
```diff
@@ -95,6 +95,7 @@ func downloadTerraformSourceIfNecessary(terraformSource *TerraformSource, terrag
 	// before and after hooks (if any).
 	terragruntOptionsForDownload := terragruntOptions.Clone(terragruntOptions.TerragruntConfigPath)
 	terragruntOptionsForDownload.TerraformCommand = CMD_INIT_FROM_MODULE
+	terragruntOptionsForDownload.Logger = terragruntOptions.Logger
```
brikis98: Does `Clone` not do this automatically?

shell/run_shell_cmd.go
```diff
@@ -57,7 +57,12 @@ func RunShellCommandWithOutput(terragruntOptions *options.TerragruntOptions, wor
 	// command requested by the user. The output of these other commands should not end up on stdout as this
 	// breaks scripts relying on terraform's output.
 	if !reflect.DeepEqual(terragruntOptions.TerraformCliArgs, args) {
-		outWriter = terragruntOptions.ErrWriter
+		errWriter = os.Stderr
+		outWriter = os.Stdout
+	}
+
+	if terragruntOptions.ErrWriter != os.Stderr {
+		errWriter = io.MultiWriter(terragruntOptions.ErrWriter, os.Stderr)
```
brikis98: Can you explain the thinking behind these changes a bit more?

cli/download_source.go
mariux: the clone function sets up a new logger logging to the current ErrWriter.. which is not where the current logger is logging to.. 

I know it is not the cleanest solution to just set it this way, but since it fixed it for me, maybe an actual go developer can come up with a cleaner implementation.

shell/run_shell_cmd.go
mariux: this is basically taking care of giving a summary of all errors at the end of a plan-all run and printing the errors while running... the check is needed to not print logs twice while running if used in a normal `plan` (run without being wrapped in a `plan-all`)..

mariux:
> Can you explain what you mean by these? Perhaps with an example?

> > fix wrongly captured output for "init-from-module" pseudo-command run

this is basically a fix to the issue described in #397: where at the end of a plan-all run all errors are printed by basically just capturing everything sent to stderr and postponing the output to the end of the plan-all phase.
since this "all errors from plan" should just include errors from actually running a `terraform plan` and not (as it does at the moment) also stuff printed to stderr by terragrunt (e.g. in the "init-from-module" phase or from setting up the s3-backend).

> > fix wrongly detected error by not capturing output for non terraform commands

what also got wrongly captured was stderr from hooks: e.g. from running a `cp -v ....`
which does not make sense because hooks have their own error handling so also the output does not need to be postponed...

cli/download_source.go
mariux: btw it seems that ErrWriter and Logger are not really connected in a logical consequent way. since setting ErrWriter does not change the Logger in any way and also not sets up a new Logger.

brikis98: Hm, it seems like the right fix would be to make the `Clone` method do the right thing. This small workaround means (a) that `Clone` may still be doing the wrong thing in other places and (b) if we make updates to `Clone` in the future, that may break in this one-off fix here.

shell/run_shell_cmd.go
brikis98: This is the `RunShellCommandWithOutput` method, which is used in many places in the code base, and not just during `plan-all` commands. I'm a little concerned about piling up workarounds within `RunShellCommandWithOutput` that are intended for other methods, as it's not at all obvious why this would be here from just reading this part of the code. Is there some other way to set `terragruntOptions` params correctly so this method does the right thing without one-off hacks?

brikis98:
How does your fix compare to https://github.com/gruntwork-io/terragrunt/pull/722?

shell/run_shell_cmd.go
mariux: agreed.. ;) i can try to fix it in another way..

mariux: as far as i have seen ErrWriter is not used in a consequent manner and fixing it in a nicer way would require a lot of refactoring as there are already a lot of exceptions throughout the code..

mariux:
> How does your fix compare to #722?

good question... basically this code adds fixes for already intended functionality without adding more complexity.. but feel free to close this one if the other PR also improves it.. i guess the outcome is very similar..

Krzysztof-Szyper-Epam:
I've noticed this error shows up when I add `-no-color` parameter to pass it to Terraform.
It's not happening without it.

mariux:
closed as this is outdated.

Issue #795, opened 2019-07-23
Terragrunt creates a public s3 bucket


src386:
- terragrunt version v0.19.6
- terraform v0.12.3

Steps to reproduce:

1. Define a s3 backend in terragrunt.hcl / main.tf
2. terragrunt init
3. Terragrunt offers to create the s3 bucket. Accept.
4. The bucket has public access

This seems like a dangerous default behavior. The bucket should be private.

```
[terragrunt] 2019/07/23 15:59:09 [terragrunt]  Remote state S3 bucket mys3bucket does not exist or you don't have permissions to access it. Would you like Terragrunt to create it? (y/n) 
y
[terragrunt] 2019/07/23 15:59:12 Create S3 bucket mys3bucket
[terragrunt] 2019/07/23 15:59:12 Creating S3 bucket mys3bucket
[terragrunt] 2019/07/23 15:59:12 Waiting for bucket mys3bucket to be created
[terragrunt] 2019/07/23 15:59:13 S3 bucket mys3bucket created.
[terragrunt] 2019/07/23 15:59:13 No tags specified for bucket mys3bucket.
[terragrunt] 2019/07/23 15:59:13 Enabling versioning on S3 bucket mys3bucket
[terragrunt] 2019/07/23 15:59:13 Enabling bucket-wide SSE on AWS S3 bucket mys3bucket
[terragrunt] 2019/07/23 15:59:13 Granting WRITE and READ_ACP permissions to S3 Log Delivery (http://acs.amazonaws.com/groups/s3/LogDelivery) for bucket mys3bucket. This is required for access logging.
[terragrunt] 2019/07/23 15:59:13 Waiting for ACL bucket mys3bucket to have the updated ACL for access logging.
[terragrunt] 2019/07/23 15:59:14 Bucket mys3bucket now has the proper ACL permissions for access logging!
[terragrunt] 2019/07/23 15:59:14 Enabling bucket-wide Access Logging on AWS S3 bucket "mys3bucket" - using as TargetBucket "mys3bucket"
```

brikis98:
Why do you think the bucket has public access?

yorinasub17:
[...]

PR #796, opened 2019-07-23
Enable all public access blocking

yorinasub17:
Based on investigation for https://github.com/gruntwork-io/terragrunt/issues/795, it turns out we have a few more safety valves we can enable to ensure no public access is granted to the state files and bucket created by terragrunt.

You can read more about this feature in S3 in the official docs: https://docs.aws.amazon.com/AmazonS3/latest/dev/access-control-block-public-access.html


yorinasub17:
Thanks for the review! Going to merge and release this.

Issue #768, opened 2019-06-30
terraform fmt doesn't work on *.hcl files


heydonovan:
terragrunt.hcl

```
inputs = {
  aws_region                   = "us-east-1"
  aws_profile                                 = "test-staging"
}
```

If I run `terragrunt fmt`, nothing changes. However, if I rename the file to `terragrunt.tf`, and then run `terragrunt fmt`, it works. How do other folks handle this?

barryib:
Terraform can read from stdin

```shell
cat terragrunt.hcl | terraform fmt -
```

mcalhoun:
I opened a PR to add this feature: https://github.com/hashicorp/terraform/pull/22092

yorinasub17:
It looks like the PR in terraform won't be accepted, so opened https://github.com/gruntwork-io/terragrunt/pull/801 to pull in `hclfmt` as a terragrunt sub command.

yorinasub17:
FYI https://github.com/gruntwork-io/terragrunt/releases/tag/v0.19.12 rolls out a new command `terragrunt hclfmt` that can be used to format the `terragrunt.hcl` files. Obviously not as great as having `terraform fmt` handle those, but probably a good workaround for now.

ashemedai:
@brikis98 @yorinasub17 I have used that `hclfmt` feature and it does what it should. I think that tackles the use case that OP (implicitly) wanted. Issue can be closed in my opinion.

yorinasub17:
With `terragrunt hclfmt` and with terraform community unlikely to merge our PR, going to close this.

PR #801, opened 2019-07-27
hclfmt

yorinasub17:
According to https://github.com/hashicorp/terraform/pull/22092, it seems unlikely that `terraform fmt` would support formatting `terragrunt.hcl` directly. In that PR, it was suggested that [we should standardize to using `hclfmt`](https://github.com/hashicorp/terraform/pull/22092#issuecomment-512028847). However, this is an additional utility that one needs to install, not to mention that it is not officially released yet.

As such, this PR bakes the functionality of `hclfmt` as a sub command of `terragrunt`, so that you can run `terragrunt hclfmt` to format your `terragrunt.hcl` files in the directory tree.

Reference: https://github.com/gruntwork-io/terragrunt/issues/768

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -314,17 +314,11 @@ func shouldPrintTerraformHelp(terragruntOptions *options.TerragruntOptions) bool
 }
 
 func shouldPrintTerragruntInfo(terragruntOptions *options.TerragruntOptions) bool {
-	if util.ListContainsElement(terragruntOptions.TerraformCliArgs, CMD_TERRAGRUNT_INFO) {
-		return true
-	}
-	return false
+	return util.ListContainsElement(terragruntOptions.TerraformCliArgs, CMD_TERRAGRUNT_INFO)
 }
 
 func shouldRunHCLFmt(terragruntOptions *options.TerragruntOptions) bool {
-	if util.ListContainsElement(terragruntOptions.TerraformCliArgs, CMD_HCLFMT) {
-		return true
-	}
-	return false
+	return util.ListContainsElement(terragruntOptions.TerraformCliArgs, CMD_HCLFMT)
 }
 
 func processHooks(hooks []config.Hook, terragruntOptions *options.TerragruntOptions, previousExecError ...error) error {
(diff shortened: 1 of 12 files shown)


cli/cli_app.go
```diff
@@ -308,6 +320,13 @@ func shouldPrintTerragruntInfo(terragruntOptions *options.TerragruntOptions) boo
 	return false
 }
 
+func shouldRunHCLFmt(terragruntOptions *options.TerragruntOptions) bool {
+	if util.ListContainsElement(terragruntOptions.TerraformCliArgs, CMD_HCLFMT) {
+		return true
+	}
+	return false
```
brikis98: Simplify to `return util.ListContainsElement(...)`

cli/hclfmt.go
```diff
@@ -0,0 +1,115 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// findTerragruntHclFiles will recursively look for terragrunt.hcl files in the directory tree starting at workingDir.
+func findTerragruntHclFiles(terragruntOptions *options.TerragruntOptions, workingDir string) ([]string, error) {
```
brikis98: Alternative option is to use the [zglob library](https://github.com/mattn/go-zglob) to give back a list in a one-liner.

cli/hclfmt.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// readAllFromFile will read the contents of the file at the given path.
+func readAllFromFile(path string) ([]byte, error) {
+	f, err := os.Open(path)
+	defer f.Close()
+	if err != nil {
+		return nil, err
+	}
+	return ioutil.ReadAll(f)
```
brikis98: https://github.com/gruntwork-io/terragrunt/blob/master/util/file.go#L131

yorinasub17:
UPDATE:

- Simplified `shouldRunHCLFmt` and `shouldPrintTerragruntInfo`
- Replaced `findTerragruntHclFiles` with `zglob.Glob`
- Replaced `readAllFromFile` with `util.ReadFileAsString`
- Updated working directory to pull from `options.WorkingDir`

yorinasub17:
Ok going to merge + release this now. Thanks for the review!

Issue #759, opened 2019-06-19
RFC: single terragrunt.hcl per environment?


brikis98:
### Current state

The current practice for using Terragrunt is to create one folder for each module and put a `terragrunt.hcl` file in it. You also have one `terragrunt.hcl` at the root of each environment with shared configurations for that environment. The folder structure looks like this:

```
staging
├── terragrunt.hcl
├── frontend-app
│   └── terragrunt.hcl
├── mysql
│   └── terragrunt.hcl
└── vpc
    └── terragrunt.hcl

production
├── terragrunt.hcl
├── frontend-app
│   └── terragrunt.hcl
├── mysql
│   └── terragrunt.hcl
└── vpc
    └── terragrunt.hcl
```

### New proposal 

One thing I've been contemplating is if it makes sense to instead aim for the ability to have one `terragrunt.hcl` per environment so that the folder structure looks like this:

```
staging
└── terragrunt.hcl
production
└── terragrunt.hcl
```

To make this work, `terragrunt.hcl` would allow you to define the configuration for multiple modules in a single file. Here's a rough sketch of what that might look like:

```hcl
# You can include one or more modules using a syntax nearly identical to Terraform
module "frontend-end-app" {
  # Use the source param to specify where the code for this module lives
  source = "github.com/acme/modules.git//frontend-app?ref=v0.3.4"
  
  # Specify input parameters for that module the same as how Terraform does it!
  instance_type = "t2.micro"
  instance_count = 10

  # You can reference outputs from other modules using the normal Terraform syntax!
  # This also defines dependencies between modules for apply-all style commands.
  vpc_id = module.vpc.id
  db_url = module.mysql.endpoint
}

module "mysql" {
  # Each source can point to different module and different versions
  source = "github.com/acme/modules.git//mysql?ref=v0.4.2"

  # You can reference shared inputs defined in this file below
  name = "${local.env_name}-mysql"

  memory = 4096
  disk_space = 250

  vpc_id = module.vpc.id
}

module "vpc" {
  source = "github.com/acme/modules.git//mysql?ref=v0.4.2"

  name = local.env_name
  cidr_block = "10.0.0.0/16"
}

# Local variables you can share amongst all the modules above
locals = {
  env_name = "staging"
}

# Backend configuration applied to all the modules above
remote_state {
  backend = "s3"
  config {
    bucket = "foo"
    region = "us-east-1"
    # module_name returns the name of the current module (e.g., frontend-app, mysql, etc)
[...]

Issue #752, opened 2019-06-17
Don't convert inputs to env vars. Instead, generate terragrunt.tfvars.json and use as -var-file


ekini:
I've been reading documentation for terraform 0.12 in regards to handling env vars, and https://www.terraform.io/docs/configuration/variables.html states that:

> Some special rules apply to the -var command line option and to environment variables. For convenience, Terraform defaults to interpreting -var and environment variable values as literal strings, which do not need to be quoted

and

> However, if a root module variable uses a type constraint to require a complex value (list, set,  map, object, or tuple), Terraform will instead attempt to parse its value using the same syntax  used within variable definitions files, which requires careful attention to the string escaping  rules in your shell

And it's so error-prone and counter-intuitive! Just take a look at https://github.com/gruntwork-io/terragrunt/issues/740 and https://github.com/gruntwork-io/terragrunt/issues/738

Instead, terraform directly supports loading variables from json files: https://www.terraform.io/docs/configuration/variables.html#variable-definitions-tfvars-files

Terraform 0.11 supports it as well, although it's not properly documented:
> Variables files use HCL or JSON syntax to define variable values

The main point here is that types are not lost and terraform doesn't have to "guess" them. Also, the whole logic of converting inputs to env vars can be simplified to one `json.Marshal`.

Benefits:
1. Simplified logic
2. Supported by terraform 0.12 and properly documented
3. Types are not lost in translation, no need to set constraints in `variable {}` blocks
4. Works with terraform 0.11 nicely, for compatibility sake!

I can make a PR should it be needed.

barryib:
Oh yes ! This is definitely the way to go for input variables.

Furthermore, as discussed https://github.com/gruntwork-io/terragrunt/issues/737#issuecomment-500895952 with @brikis98 we should also stick to Terraform, by defining variables explicitly.

```hcl
# In root terragrunt.hcl
terraform {
  extra_arguments "terragrunt_generated_vars" {
    commands = "${get_terraform_commands_that_need_vars()}"

    required_inputs_files = [
      "${get_terragrunt_dir()}/${find_in_parent_folders("common.hcl")}",
    ]
  }
}
```

```hcl
# stage/frontend-app/terragrunt.hcl
terraform {
  source = "..."
}

# Include all settings from the root terragrunt.hcl file
include {
  path = "${find_in_parent_folders()}"
}

inputs = {
  aws_region = get_input("aws_region")
  remote_state_bucket = get_input("remote_state_bucket")

[...]

PR #802, opened 2019-07-27
Locals

yorinasub17:
This implements the concept of `locals` in terragrunt config. This can be used to bind expressions to names for reusing within the config, so that you can DRY complex expressions within the config.

TODO:
- [x] Add more testing


jfunnell:
Wow awesome, this is an excellent improvement if it works (might test it today..). 

Like you said, I don't really see how this could have helped with issue #752 and that is fine.. I think the overall best solution to clean up some of the shortcomings of terragrunt is to implement the entire feature set in #759, which would probably require a hefty chunk of code/functionality to change.  
In short, I think this is an acceptable stopgap for now. Just my 2c

README.md
```diff
@@ -1251,6 +1252,26 @@ Note that Terragrunt will respect any `TF_VAR_xxx` variables you've manually set
 anything in `inputs` will NOT be override anything you've already set in your environment.  
 
 
+### Locals
+
+You can use locals to bind a name to values and expressions that you would like to create a reference for within your
```
brikis98:
```suggestion
You can use locals to bind a name to an expression, so you can reuse that expression without having to repeat it multiple times (keeping your Terragrunt configuration DRY).
```

README.md
```diff
@@ -1251,6 +1252,26 @@ Note that Terragrunt will respect any `TF_VAR_xxx` variables you've manually set
...
+  aws_region = "us-east-1"
+}
+
+inputs = {
+  aws_region  = local.aws_region
+  s3_endpoint = "com.amazonaws.${local.aws_region}.s3"
+}
+```
+
+You can use any valid terragrunt expression in the `locals` configuration.
```
brikis98: Prepare yourself for the "how do I reuse locals from another file?" question in 3... 2... 1... :-D

Perhaps tackle that head on and add a section on reusing/merging `yaml` files for now? And when we add `get_input` and `get_output`, we can replace it.

config/config.go
```diff
@@ -299,24 +312,48 @@ func ParseConfigString(configString string, terragruntOptions *options.Terragrun
...
 	}
 	return &terragruntInclude, nil
 }
 
-func decodeAsTerragruntConfigFile(file *hcl.File, filename string, terragruntOptions *options.TerragruntOptions, include *IncludeConfig) (*terragruntConfigFile, error) {
+// This decodes only the `locals` attribute of a terragrunt config, so its value can be used while decoding the rest of
+// the config. For consistency, `locals` in the call to `decodeHcl` is always assumed to be nil.
+func decodeAsTerragruntLocals(file *hcl.File, filename string, terragruntOptions *options.TerragruntOptions) (*terragruntLocals, error) {
+	terragruntLocals := terragruntLocals{}
+	err := decodeHcl(file, filename, &terragruntLocals, terragruntOptions, nil, nil)
```
brikis98: What happens if a local references another local?

README.md
yorinasub17: Ok I can add that in.

FWIW, I was thinking about proposing a similar construct, but called `globals` which can be merged with the parent in an `include` call. I think the explicit separating of `locals` from `globals` is important, especially if and when `include` becomes more complex to allow inclusion of multiple levels of terragrunt config.

config/config.go
yorinasub17: Ah I didn't think about this possibility. I think currently it will panic or raise a syntax error, but it would be nice if it can be recursive. Let me do a quick spike to see how hard it is to implement this.

yorinasub17:
UPDATE:

I made a huge refactor of the implementation to convert the `locals` attribute to a block. This has the advantage of being able to reference other locals in the `locals` block, but at the expense of being unable to merge directly from `yamldecode`. I think this is an ok tradeoff considering that you can still import yamls as `locals`, only that it can't be merged at the top level (see the new docs I wrote).

Note that I am still optimizing the logs and error handling, so not quite ready to merge yet but I reached the point where it would be good to get some feedback on the new approach.

yorinasub17:
Ok @brikis98 this is ready for re-review given the new implementation.

README.md
```diff
@@ -1253,6 +1254,79 @@ Note that Terragrunt will respect any `TF_VAR_xxx` variables you've manually set
...
+
+```
+locals {
+  x = 2
+  y = 40
+  answer = local.x + local.y
+}
+```
+
+##### Including globally defined locals
```
brikis98: Thx for adding 👍

config/config.go
```diff
@@ -45,6 +47,15 @@ type terragruntConfigFile struct {
 	PreventDestroy *bool                  `hcl:"prevent_destroy,attr"`
 	Skip           *bool                  `hcl:"skip,attr"`
 	IamRole        *string                `hcl:"iam_role,attr"`
+
+	// This should be ignored
```
brikis98: Why ignored?

config/config.go
```diff
@@ -299,24 +318,37 @@ func ParseConfigString(configString string, terragruntOptions *options.Terragrun
 // an include block)
 func decodeAsTerragruntInclude(file *hcl.File, filename string, terragruntOptions *options.TerragruntOptions) (*terragruntInclude, error) {
 	terragruntInclude := terragruntInclude{}
-	err := decodeHcl(file, filename, &terragruntInclude, terragruntOptions, nil)
+	err := decodeHcl(file, filename, &terragruntInclude, terragruntOptions, nil, nil)
```
brikis98: An `include` can't have its own locals?

config/locals.go
```diff
@@ -0,0 +1,313 @@
...
+	diagsWriter := util.GetDiagnosticsWriter(parser)
+
+	localsBlock, diags := getLocalsBlock(hclFile)
+	if diags.HasErrors() {
+		diagsWriter.WriteDiagnostics(diags)
+		return nil, errors.WithStackTrace(diags)
+	}
+	if localsBlock == nil {
+		// No locals block referenced in the file
+		terragruntOptions.Logger.Printf("Did not find any locals block: skipping evaluation.")
```
brikis98: Do we need to log this?

config/locals.go
```diff
@@ -0,0 +1,313 @@
...
+		return nil, nil
+	}
+
+	terragruntOptions.Logger.Printf("Found locals block: evaluating the expressions.")
```
brikis98: Same here... Terragrunt doesn't currently support log levels, so the log statements here could be quite noisy...

config/locals.go
```diff
@@ -0,0 +1,313 @@
...
+		terragruntOptions.Logger.Printf("Encountered error while decoding locals block into name expression pairs.")
+		diagsWriter.WriteDiagnostics(diags)
+		return nil, errors.WithStackTrace(diags)
+	}
+
+	// Continuously attempt to evaluate the locals until there are no more locals to evaluate, or we can't evaluate
+	// further.
+	evaluatedLocals := map[string]cty.Value{}
+	evaluated := true
+	for len(locals) > 0 && evaluated {
```
brikis98: Should we add a simple counter and ensure it stays under, say, 1000? Could be a nice safety valve in case of a bug in the code that leads to an infinite loop.

Actually, on that note, what if the code has:

```hcl
locals {
  a = local.b
  b = local.a
}
```

config/locals.go
```diff
@@ -0,0 +1,313 @@
...
+	return evaluatedLocals, nil
+}
+
+// attemptEvaluateLocals attempts to evaluate the locals block given the map of already evaluated locals, replacing
+// references to locals with the previously evaluated values. This will return:
+// - the list of remaining locals that were unevaluated in this attempt
+// - the updated map of evaluated locals after this attempt
+// - whether or not any locals were evaluated in this attempt
+// - any errors from the evaluation
+func attemptEvaluateLocals(
```
brikis98: Yikes, this is quite complicated. I have some vague memory that the HCL parser has native support for partial evaluation specifically for this use case, but can't find it in the code base quickly...

config/locals_test.go
```diff
@@ -0,0 +1,83 @@
...
+      {region = local.region},
+	  {foo = "bar"},
+	)
+  ]
+  bar = local.foo[0]["region"]
+
+  // Multiple references
+  x = 1
+  y = 2
+  z = local.x + local.y
```
brikis98: Consider testing a few corner cases:

* Reference to a `local` that isn't defined in a `locals { ... }` block
* Reference to a `local` that isn't defined elsewhere in the code
* Cyclical references
* Several layers of references (a = local.b, b = local.c, c = local.d, etc)

config/locals.go
yorinasub17: This will fail because neither can be evaluated in the first loop, so `evaluated` will be `false`.

config/locals.go
yorinasub17: I searched but the only partial evaluation I could find was at the top file level, and not at the individual block level where we have expressions.

config/config.go
yorinasub17: This needs to be skipped here because `hcl` can't decode it correctly when it self references `locals`. It is not necessary to decode it into this struct here because it is evaluated in a separate loop.

However, we don't want to completely omit the existence of it because we want a syntax error if we have blocks that terragrunt doesn't expect (e.g typos). Hence it is added here as an attribute but the struct won't attempt to parse anything in it.

config/config.go
yorinasub17: Ah good point! Updated the evaluation order so it can be there.

config/locals.go
yorinasub17: Implemented a crude version of log levels. This is actually quite useful to debug issues with locals and I expect we will need it once people start using it.

config/locals.go
yorinasub17: See above re:log levels.

config/locals.go
yorinasub17: > Should we add a simple counter and ensure it stays under, say, 1000? Could be a nice safety valve in case of a bug in the code that leads to an infinite loop.

Implemented.

config/config.go
brikis98: Perhaps expand the comment a bit to explain this?

util/logger.go
```diff
@@ -19,3 +23,22 @@ func CreateLoggerWithWriter(writer io.Writer, prefix string) *log.Logger {
...
 }
+
+// MAINTAINER'S NOTE: This is a temporary solution for logging levels in terragrunt. This is not a permanent debug
+// logging solution.
+// Debugf will only print out terragrunt logs if the TG_LOG environment variable is set to DEBUG.
+func Debugf(logger *log.Logger, fmtString string, fmtArgs ...interface{}) {
+	if os.Getenv("TG_LOG") == "DEBUG" {
+		logger.Printf(fmtString, fmtArgs...)
+	}
+}
```
brikis98: Ah, clever 👍

config/config.go
yorinasub17: Already did: https://github.com/gruntwork-io/terragrunt/pull/802/files#diff-3baf47c64847a8fb8aaa8cc2e088513bR51

config/locals_test.go
yorinasub17: Done in a107c89

yorinasub17:
UPDATE a107c89 :

- Added a bunch more test cases
- Fixed a bug where we parsed multiple locals block, but silently ignored all but the first one defined. Now we throw an error if we detect multiple locals blocks.
- Fixed a case where we weren't properly writing out diagnostics information.

With that, if the build passes, I think we are ready to merge + release!

brikis98:
LGTM!

yorinasub17:
Ok merging and releasing. Thanks for the review!

README.md
apottere: Do you have any immediate plans to implement `globals`?  We've just developed a need for something like that (having variables in our parent `terragrunt.hcl` that we can override from child `terragrunt.hcl`s).  While including a yaml file if it exists would work, it doesn't seem like the cleanest solution and we'd have to walk it back once there's a better way.

If you haven't started and you think it would be something that would be accepted, I don't mind creating a PR for it.

yorinasub17: > If you haven't started and you think it would be something that would be accepted, I don't mind creating a PR for it.

I don't think we have immediate plans for `globals`. I would suggest first creating an issue (with `Feature Request` in the title) describing:

- Your use case. Why do you need something like `globals`?
- Ideal solution. Is `globals` what you want, or do you want something like `get_local`, which can take another `terragrunt.hcl` config file as input, a `key` and return the value assigned to that `local` in the target file?
- Indicate whether you want to open a PR, either for the ideal solution or if that is challenging, a temporary workaround solution that works, but is better than what is available now.

PR #807, opened 2019-07-31
Initial check mode

thoo5ieb:


thoo5ieb:
@yorinasub17 Once you have a time, thank you!

cli/hclfmt.go
```diff
@@ -62,6 +66,16 @@ func formatTgHCL(terragruntOptions *options.TerragruntOptions, tgHclFile string)
 	}
 
 	newContents := hclwrite.Format(contents)
+	checkMode := util.ListContainsElement(terragruntOptions.TerraformCliArgs, "-check")
```
yorinasub17: I'm not sure I like this. On the one hand, it's a nice way to keep the code simple. On the other hand, this is highly dependent on how we are parsing the args and segregating what counts as Terragrunt args and what counts as Terraform args. Not to mention the awkwardness where this is a terragrunt specific command inspecting args for terraform commands...

Could use a second opinion here. What do you think @brikis98 ?

Also, what do you think about using a different command name for this mode? E.g something like `check-hclfmt`.

thoo5ieb: Yeah, totally. I already asked about in the issue. :-) 
Personally, I don't like the mixed context here, e.g. lookup Terraform structure to validate Terragrunt specific option.

thoo5ieb: Other way, an additional command `check-hclfmt` sounds odd. I don't mind to follow already existing practice where Terragrunt specific options use prefix `--terragrunt`.

thoo5ieb:
@yorinasub17 

I haven't added unit tests mainly because I was not sure about additional optional parsing. Once we agree, I will come up with unit tests as well.

cli/hclfmt.go
brikis98: I think the "idiomatic" way to handle this in the Terragrunt code base is to add this flag to `TerragruntOptions` and update the parse method for it to read out this arg. We do that for other Terragrunt-specific args already.

thoo5ieb: @brikis98 Yay! Thank you!

thoo5ieb:
@yorinasub17 Once you get some time on this, I will be more than happy to get the feedback, in the meantime have a nice day!

cli/cli_app.go
```diff
@@ -136,6 +137,7 @@ GLOBAL OPTIONS:
    terragrunt-ignore-external-dependencies  *-all commands will not attempt to include external dependencies
    terragrunt-exclude-dir                   Unix-style glob of directories to exclude when running *-all commands
    terragrunt-include-dir                   Unix-style glob of directories to include when running *-all commands
+   terragrunt-check                         Enable check mode.
```
yorinasub17:
```suggestion
   terragrunt-check                         Enable check mode in the hclfmt command.
```

cli/hclfmt_test.go
```diff
@@ -88,3 +88,46 @@ func TestHCLFmtErrors(t *testing.T) {
 		})
 	}
 }
+
+func TestHCLFmtCheck(t *testing.T) {
```
yorinasub17: Can you also add a test for the base case, where there are no files to change?

README.md
```diff
@@ -2149,6 +2149,9 @@ If you run `terragrunt hclfmt` at the `root`, this will update:
 - `root/dev/terragrunt.hcl`
 - `root/qa/terragrunt.hcl`
 
+Additionally, there's a flag `--terragrunt-check`. It allows to validating if files are properly formatted. It does not
```
yorinasub17:
```suggestion
Additionally, there's a flag `--terragrunt-check`. It can be used for validating if files are properly formatted without actually formatting the files. It does not
```

thoo5ieb:
Awesome! :-) Thank you for the help!

yorinasub17:
Thanks for resolving the conflicts. Kicking off the build now.

yorinasub17:
Build passed! Going to merge and release. Thanks for your contribution!

yorinasub17:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.19.16

Binaries should show up shortly.

PR #826, opened 2019-08-09
hclfmt should run before parsing the terragrunt config

yorinasub17:
The formatting tool should not be concerned with decoding the terragrunt config, and should only deal with HCL2 syntax errors.


barryib:
Any chance to see this merged ?

brikis98:
@yorinasub17 Is this waiting on anything? Test failure?

yorinasub17:
Yes there is a deterministic failure where the test stalls and times out. I haven't had a chance to dig in given that I couldn't repro it locally.

yorinasub17:
Ok I don't know why the build was failing before, but it is passing after a rebase so going to merge and release this now.

PR #828, opened 2019-08-09
Implement terragrunt_output

yorinasub17:
The implementation has now been completely changed. `get_output` is no longer a function and instead is a new block called `terragrunt_output`. The benefits of this approach are:

- Predictable execution paths: it is now very clear when outputs are pulled into the config. See https://github.com/gruntwork-io/terragrunt/blob/38aca82ff0890af3a4584870b4bd5eed8f788311/README.md#configuration-parsing-order for more info.
- Automatic dependency graph injection: since the parsing order is now predictable, we can seed the dependency graph with the configs specified by `terragrunt_output` blocks. See https://github.com/gruntwork-io/terragrunt/blob/38aca82ff0890af3a4584870b4bd5eed8f788311/README.md#passing-outputs-between-modules for more info.
- Reuse results of a single `terragrunt output` call: with `get_output`, `terragrunt output` had to be called each time the function was executed to keep the implementation sane. Since `get_output` couldn't be parsed before `locals`, this made it difficult to reuse the output results of a single call if you needed multiple variables from the output. Now only one call needs to be made per block and it can be referenced multiple times.

~This provides an implementation of `get_output`, an interpolation function that can be used to extract the output of another terraform module wrapped with terragrunt config. See [the README changes for more details](https://github.com/gruntwork-io/terragrunt/blob/74273a9628ec766af016056c39e56a54eea6d94b/README.md#get_output).`~

Note that as part of the implementation, it was required to refactor the configuration parsing during a `xyz-all` command to only do a partial decoding. Otherwise, `apply-all` will fail even if you specify dependencies because it will interpolate the `get_output` call in the initial pass to read the `dependencies` from the config. This is handled in a new, `PartialParseConfigFile` function which will only parse the sections specified. ~Note that because `locals`, `include`, `dependencies`, `terraform` blocks are evaluated in this section, `apply-all` in the initial setup will still fail if `get_output`. This is covered in the `README`.~


README.md
```diff
@@ -1644,6 +1645,60 @@ variables](https://www.terraform.io/docs/configuration/environment-variables.htm
...
+}
+```
+
+Note that the terragrunt run configuration is inherited from the current session. What this means is that settings set
+with environment variables (such as `TERRAGRUNT_IAM_ROLE`) are automatically passed down to the execution context of
+the target terragrunt config when reading the output data.
+
+**Important**: Because `get_output` depends on the target config having been applied already, you will need to make sure
+that the specified `TARGET_TERRAGRUNT_CONFIG` is listed as a dependency in the `dependencies` block. Otherwise, this
+will fail when you run an `apply-all` since the output will be interpolated before the target config has applied.
```
brikis98: I think `get_output` should automatically count as implicit dependency, removing the need for adding the item in the `dependencies` block. This is the same way Terraform works: a reference from one resource to another creates an implicit dependency, and you use `depends_on` solely as a workaround.

In fact, `get_output` may allow us to move away from `terraform_remote_state` and manually-managed, error-prone `dependencies` entirely. Instead, each module will expose input variables for all its dependencies (e.g., `vpc_id`) and in `infrastructure-live`, you'll fill in those variables via a bunch of `get_output` calls (e.g., `vpc_id = get_output("../../vpc", "vpc_id")`). You can then call `apply-all`, and we'll automatically apply everything in the right order based on these `get_output` dependencies. 

Moreover, instead of getting a confusing error when `terraform_remote_state` of some dependency doesn't exist, the `get_output` helper can detect if a module hasn't been applied yet, and show a corresponding error message.

README.md
```diff
@@ -1644,6 +1645,60 @@ variables](https://www.terraform.io/docs/configuration/environment-variables.htm
...
+Additionally, during an `apply-all`, the terragrunt configuration has to be partially interpolated in order to build up
+the dependency tree. This means that you will have issues using `apply-all` if `get_output` is used in the following
+blocks, as it will be interpolated prior to any modules being applied:
```
brikis98: Not sure I follow this last one? Can you give an example?

README.md
```diff
@@ -1644,6 +1645,60 @@ variables](https://www.terraform.io/docs/configuration/environment-variables.htm
...
+
+- `locals`
```
brikis98: Hm, if you need to fetch 10 different outputs from a single dependency, I suspect 10 sequential `terragrunt output` calls would be slow. Do we cache the outputs in memory in between? If not, it would be nice to allow users to cache all the outputs in a `locals` variable once and then look up values within it...

config/config_helpers.go
```diff
@@ -301,6 +307,93 @@ func getAWSAccountID(include *IncludeConfig, terragruntOptions *options.Terragru
...
+		return nil, errors.WithStackTrace(EmptyStringNotAllowed("the target terragrunt config parameter"))
+	}
+	if numParams > 2 {
+		return nil, errors.WithStackTrace(WrongNumberOfParams{Func: "get_output", Expected: "1 or 2", Actual: numParams})
+	}
+
+	// target config check: make sure the target config exists
+	targetConfig := params[0]
+	if util.IsDir(targetConfig) {
+		targetConfig = util.JoinPath(targetConfig, DefaultTerragruntConfigPath)
```
brikis98: TODO for the future: we could support reading outputs from normal Terraform modules (i.e., those not meant to be used with Terragrunt) by running `terraform output` instead of `terragrunt output`. Not particularly high priority for now though!

README.md
```diff
@@ -1644,6 +1645,60 @@ variables](https://www.terraform.io/docs/configuration/environment-variables.htm
...
+```
+output "instance_id" {}
+output "instance_name" {}
+```
+
+In `app3`, you can get each of these outputs using `get_output` in the following ways:
+
+```hcl
+inputs = {
+  instance_id_directly = get_output("${get_terragrunt_dir()}/../../app1", "instance_id")
```
brikis98: I wish relative file paths would work here so this wasn't so verbose 😕

config/config_helpers.go
```diff
@@ -301,6 +307,93 @@ func getAWSAccountID(include *IncludeConfig, terragruntOptions *options.Terragru
...
+	}
+	if !util.FileExists(targetConfig) {
+		return nil, errors.WithStackTrace(TerragruntOutputConfigNotFound{Path: targetConfig})
+	}
+
+	// use terragrunt running functions to extract the json output from the target config
```
brikis98: Extract this part into its own method

config/config_helpers.go
```diff
@@ -301,6 +307,93 @@ func getAWSAccountID(include *IncludeConfig, terragruntOptions *options.Terragru
...
+	if err != nil {
+		return nil, errors.WithStackTrace(err)
+	}
+
+	stdoutBufferWriter.Flush()
+	jsonString := stdoutBuffer.String()
+	jsonBytes := []byte(jsonString)
+	util.Debugf(terragruntOptions.Logger, "Retrieved output from %s as json: %s", targetConfig, jsonString)
+
+	if numParams == 2 {
```
brikis98: Extract this into a separate method with unit tests

config/config_helpers.go
```diff
@@ -301,6 +307,93 @@ func getAWSAccountID(include *IncludeConfig, terragruntOptions *options.Terragru
...
+	// use terragrunt running functions to extract the json output from the target config
+	var stdoutBuffer bytes.Buffer
+	stdoutBufferWriter := bufio.NewWriter(&stdoutBuffer)
+	targetOptions := terragruntOptions.Clone(targetConfig)
+	targetOptions.TerraformCliArgs = []string{"output", "-json"}
+	if numParams == 2 {
+		targetOptions.TerraformCliArgs = append(targetOptions.TerraformCliArgs, params[1])
+	}
+	targetOptions.Writer = stdoutBufferWriter
+	err := targetOptions.RunTerragrunt(targetOptions)
```
brikis98: Does this stream the `terragrunt output` call to stdout or stderr of the current session?

config/config_helpers.go
```diff
@@ -301,6 +307,93 @@ func getAWSAccountID(include *IncludeConfig, terragruntOptions *options.Terragru
...
+	}
+
+	stdoutBufferWriter.Flush()
+	jsonString := stdoutBuffer.String()
+	jsonBytes := []byte(jsonString)
+	util.Debugf(terragruntOptions.Logger, "Retrieved output from %s as json: %s", targetConfig, jsonString)
+
+	if numParams == 2 {
+		// When indexing a single output, we can return the value directly after unparsing the json.
+		retType, err := ctyjson.ImpliedType(jsonBytes)
```
brikis98: Does this know how to parse the output of `terraform output -json` and extract the type from the `type` field?

config/config_helpers.go
```diff
@@ -301,6 +307,93 @@ func getAWSAccountID(include *IncludeConfig, terragruntOptions *options.Terragru
...
+		if err != nil {
+			return nil, errors.WithStackTrace(TerragruntOutputParsingError{Path: targetConfig, Err: err})
+		}
+		val, err := ctyjson.Unmarshal(jsonBytes, retType)
+		if err != nil {
+			err = TerragruntOutputEncodingError{Path: targetConfig, Err: err}
+		}
+		return &val, errors.WithStackTrace(err)
+	} else {
+		// When getting all outputs, terraform returns a json with the data containing metadata about the types, so we
```
brikis98: Extract this into a separate method with unit tests

config/config_partial.go
```diff
@@ -0,0 +1,190 @@
...
+// - locals
+// - include
+// Note also that the following blocks are never decoded in a partial parse:
+// - inputs
+func PartialParseConfigString(
+	configString string,
+	terragruntOptions *options.TerragruntOptions,
+	includeFromChild *IncludeConfig,
+	filename string,
+	decodeList []string,
```
brikis98: What is `decodeList`? Need some more docs on this method to explain the params.

config/config_partial.go
```diff
@@ -0,0 +1,190 @@
...
+	includeForDecode, err := getIncludedConfigForDecode(terragruntInclude, terragruntOptions, includeFromChild)
+	if err != nil {
+		return nil, err
+	}
+
+	output := TerragruntConfig{IsPartial: true}
+
+	for _, decode := range decodeList {
+		switch decode {
+		case "dependencies":
```
brikis98: A bit confused on what we're doing in this loop?

config/config_helpers.go
yorinasub17: The `type` field only exists if we are reading the entire output, not just one. In this mode, we are actually reading just one output, which is actually the raw output. This is the same as https://github.com/gruntwork-io/terratest/blob/master/modules/terraform/output.go#L83

Below, in the `else` block, we have the mode where the type info is in the json output and I read that in there to convert it to a `cty.Value` object.

config/config_partial.go
yorinasub17: This makes more sense if you understand the original root problem implied in https://github.com/gruntwork-io/terragrunt/pull/828#discussion_r312653873

config/config_helpers.go
brikis98: You can read all outputs every time, but just return the single output when requested...

README.md
yorinasub17: Yes ideally. However, this is going to be significantly more work to implement.

In my experimentation with walking the AST in the `locals` implementation, it turns out that function calls are not super easy to reach in the parsed AST. I know this, because I noticed that those are not clear nodes. I was only hitting the vars when I was walking the tree. This was nice when I was parsing out the self references in `locals`; bad in this case where I want to parse out the function args to `get_output`.

yorinasub17: Not completely ruling it out: just that this should be another PR after this one.

README.md
yorinasub17: Imagine you had:

```
live
├── sql
│   └── terragrunt.hcl
└── vpc
    └── terragrunt.hcl
```

And you had:

```
locals = {
  vpc_path = "../vpc"
  vpc_id = get_output(local.vpc_path, "vpc_id")
}

dependencies {
  paths = [local.vpc_path]
}

inputs = {
  vpc_id = local.vpc_id
}
```

Now suppose we have a completely clean project and nothing is applied yet and we run `apply-all`. In order to build the dependency tree, `terragrunt` needs to parse out the `dependencies` block and decode them *before any modules have been applied*. In the above use case, the `dependencies` block also references `locals`, so we need to decode `locals` as well. As part of that, we need to interpolate `get_output`.

Well, `get_output` will try to read the output of the `vpc` module and fail, because that hasn't been applied yet, because we are still in the phase of building up the dependency tree.

config/config_partial.go
yorinasub17: See https://github.com/gruntwork-io/terragrunt/pull/828#discussion_r312662399. Now imagine the following modified config for `sql`:

```
locals = {
  vpc_path = "../vpc"
}

dependencies {
  paths = [local.vpc_path]
}

inputs = {
  vpc_id = get_output(local.vpc_path, "vpc_id")
}
```

In the original implementation of the parsing logic for `apply-all`, we will still fail because we will naively parse the entire config file, including the `inputs`, regardless of whether or not we needed it in the dependency tree builder (creating the `Stack` struct).

Looking at the routine for building up the `Stack` struct, the only sections of the config that we need at that moment is the `terraform` and `dependencies` block. So this function and for loop is providing a routine that allows you to partially parse the terragrunt config so that you only read in those blocks, and ignore the rest.

`PartialParseConfigString(..., []string{"dependencies", "terraform"})` only parses the two blocks `dependencies` and `terraform` of the config (plus `include` and `locals`, which can modify those), and ignores the rest. In this way, we can extract the information we need without parsing the entire config, which may not be ready for parsing. Later on, when we actually run the `terraform apply`, the config is parsed again in full.

The forloop is just an implementation detail of the partial parsing, so that you can better control which sections are decoded. I couldn't find a clean way to programmatically build up a struct in a clean way. So instead, we conditionally pass in different structs that only extract the pieces we need from the config.

config/config_helpers.go
yorinasub17: Yea... I could do that... That kind of felt unnecessary but I guess the type information is useful....

config/config_helpers.go
yorinasub17: Done

config/config_helpers.go
yorinasub17: Done

config/config_helpers.go
yorinasub17: It streams the `stderr` to the same `stderr` of the current session, since the `ErrWriter` is left untouched from the `Clone` function.

config/config_helpers.go
yorinasub17: Done.

config/config_helpers.go
yorinasub17: Done

config/config_partial.go
yorinasub17: Added more docs.

README.md
yorinasub17: Yea that is a good point. Under most circumstances, you can use `get_output` in `locals`. There are some caveats you have to be aware of: https://github.com/gruntwork-io/terragrunt/pull/828#discussion_r312662399

README.md
yorinasub17: Fixed.

yorinasub17:
Ok the main sticking point is finding a good way to cache the `get_output` calls. Unfortunately, I am not sure what is the best way. I am not a fan of memoizing in the global space, and using `locals` has some issues as mentioned above.

I expect that the real solution is in https://github.com/gruntwork-io/terragrunt/pull/828#discussion_r312653692, where we won't need to parse the dependencies block anymore. Although... we still need the `locals` block parsing since I expect you will want to use it for the `terraform` block parsing...

README.md
brikis98: Do we need to walk the AST ahead of time for `get_output`? Or is it enough to record dependencies while that function is being executed as part of existing config parsing, and to then use the discovered dependency graph before executing further?

README.md
brikis98: Oh, I gotcha. So then the `get_output` function would have to:

1. Is this a single-module command (e.g., `apply`) or multi-module (e.g., `apply-all`).
1. If single-module, and the dependency has not yet been applied (i.e., output is missing), return an error immediately. 
1. If multi-module, add the VPC to the dependency graph, pause config parsing until the command (e.g., `apply`) has been executed in that dependency, and then continue config parsing after.

Would that work?

README.md
brikis98: My first impression was that we should maintain an internal, in-memory cache of (module absolute path, all outputs for that module) pairs. Any time you call `get_output`, it first checks the cache for that value: if it's present, it returns it immediately; if it's absent, it calls `terraform output` and caches the result. This ensures maximal performance without any extra effort for the user...

config/config_helpers.go
```diff
@@ -301,6 +307,108 @@ func getAWSAccountID(include *IncludeConfig, terragruntOptions *options.Terragru
...
+	}
+
+	outputMap, err := terraformOutputJsonToCtyValueMap(targetConfig, jsonBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	if numParams == 2 {
+		// When indexing a single output, we can return the value directly after parsing the json.
+		tmp := outputMap[params[1]]
```
brikis98: Will this `panic` if `outputMap` is missing the key? Should the two-return value version of map lookup be used instead?

config/config_helpers.go
brikis98: We may want to test that calling `get_output` on dependencies does not dirty `stdout`. E.g., If I run `terragrunt output` on a module foo, that has a `terragrunt.hcl` with a `get_output` on module bar, we need to make sure that `stdout` contains _solely_ the result of aclling `terraform output` on foo, and not any of the logs from calling `terraform output` on bar.

config/config_partial.go
brikis98: Should those be enums rather than strings?

README.md
yorinasub17: Depending on the implementation, I worry about what happens when we are going through multiple passes of the config, as it does in the `xxx-all` command. Here is where a global memoized cache could be problematic:

- In the first pass through, it would run `get_output` on the *previous* state. This result gets cached.
- The dependency gets applied as part of `apply-all`, in the dependency order.
- We parse the config again to apply the dependent module. As part of that, we call `get_output`. We look up in the cache, and return the previous value, which is NOT the newly applied state.

Note that this only happens when you depend on the awkward chicken-and-egg scenario of using `get_output` in `locals`, and in a naive implementation using a global cache that is stored for the duration of a single `terragrunt` call.

We could have a different implementation where the cache is scoped to a single config parsing, which trades off more reuse for better predictability. I am leaning towards implementing it this way, as cache bugs are always super hard to debug. This of course requires more refactoring, since now I need to store the cache somewhere in a struct that is available throughout the parsing stage.

README.md
yorinasub17: I don't think that is possible if we are using a function, because this is rendered during the struct. We COULD build a special struct that is a placeholder result of all the `get_output` calls that is then resolved later, but I think that will get very ugly unless we restrict `get_output` to only be able to apply in `input`...

yorinasub17:
Ok new idea. [This](https://github.com/gruntwork-io/terragrunt/pull/828#discussion_r313190697) got me thinking of a potential way to solve this: what if we introduce a new block? After navigating HCL2, I think I am beginning to understand that it is significantly easier to implement partial parsing decoding of individual blocks as opposed to when it appears in the middle of an AST.

Given that, it seems like the best way to achieve all of the goals is to introduce a new block construct like `terraform_remote_state`, but at the `terragrunt.hcl` level. Here is an example, since I think it will be clear once I show it:

```
terragrunt_output "vpc" {
  config = "../vpc"
}

inputs = {
  vpc_id = terragrunt_output.vpc.vpc_id
}
```

In this model, the caching problem goes away because we can resolve all the `terragrunt_outputs` in the initial parsing and stored somewhere as a reference which can then be reused.

Parsing order problems go away as well, since it can be parsed independently of the rest of the config (except maybe `locals` and `includes`?).

Implementing partial parsing just for dependency building is easy as well, since all we need to do is parse the blocks into a list of structs to get the config: no need to walk an AST!

Another benefit is that we have full control over when the output pulling happens, since now it is a first class task in the decoding pipeline, happening before we pass the HCL struct to the decoder. So the parsing logic will now be:

1. parse `locals`
1. parse `include`
1. parse `terragrunt_output` (replaces `dependencies`)
1. Get outputs from each `terragrunt_output`
1. parse the rest of config

brikis98:
That seems like an elegant solution!

yorinasub17:
Ok @brikis98 this is ready for a rereview. I have implemented `terragrunt_output` blocks as described in https://github.com/gruntwork-io/terragrunt/pull/828#issuecomment-520655414. I believe this addresses almost all of your concerns from the initial PR:

- Predictable execution paths: it is now very clear when outputs are pulled into the config. See https://github.com/gruntwork-io/terragrunt/blob/38aca82ff0890af3a4584870b4bd5eed8f788311/README.md#configuration-parsing-order for more info.
- Automatic dependency graph injection: since the parsing order is now predictable, we can seed the dependency graph with the configs specified by `terragrunt_output` blocks. See https://github.com/gruntwork-io/terragrunt/blob/38aca82ff0890af3a4584870b4bd5eed8f788311/README.md#passing-outputs-between-modules for more info.
- Reuse results of a single `terragrunt output` call: with `get_output`, `terragrunt output` had to be called each time the function was executed to keep the implementation sane. Since `get_output` couldn't be parsed before `locals`, this made it difficult to reuse the output results of a single call if you needed multiple variables from the output. Now only one call needs to be made per block and it can be referenced multiple times.

README.md
yorinasub17: This is now implemented with the `terragrunt_output` change because it is so much easier to partially parse individual blocks than a random function call!

README.md
yorinasub17: This still exists, but is less of an issue because of the way you access `terragrunt_output`.

README.md
yorinasub17: In the new implementation, you will call `terragrunt output` at most once per dependency per config. There is still room for memoization to cache across `apply-all` calls, but I think this is world's better than `get_output` which was once per function call.

And we don't need caching (which is always a pain to maintain due to annoying cache busting bugs) :)

config/config_helpers.go
yorinasub17: Verified in test that `stdout` isn't garbled by reading `terragrunt output` call to a module with `terragrunt_output` blocks defined on `stdout`.

config/config_partial.go
yorinasub17: Switched to using enums.

config/config_helpers.go
yorinasub17: This is no longer relevant as the indexing for specific outputs is now done in HCL2, not golang.

README.md
```diff
@@ -1035,10 +1037,88 @@ If your modules have dependencies between them—for example, you can't deploy t
...
+`backend-app/terragrunt.hcl`:
+
+```
+terragrunt_output "mysql" {
+  config_path = "../mysql"
+}
+
+terragrunt_output "redis" {
+  config_path = "../redis"
+}
```
brikis98: This is very, very cool 👍 

Now that I've seen this, I can't wonder if this should be:

```hcl
dependency "mysql" {
  config_path = "../mysql"
}

dependency "redis" {
  config_path = "../redis"
}

inputs = {
  mysql_url = dependency.mysql.outputs.domain
  redis_url = dependency.redis.outputs.domain
}
```

For this PR, the only difference would be a slight naming change: `terragrunt_output` becomes `dependency` and you read outputs from it via the `.outputs` attribute. However, in a future PR, we could allow you to read _any_ part of that module's config! E.g., Perhaps `dependency.mysql.inputs.foo` would return the value of the "foo" output from the `mysql` dependency and `dependency.redis.remote_state.config.bucket` would return the bucket configured for remote state storage in the `redis` dependency.

README.md
```diff
@@ -2126,6 +2206,46 @@ The `skip` flag must be set explicitly in terragrunt modules that should be skip
 `terragrunt.hcl` file that is included by another `terragrunt.hcl` file, only the `terragrunt.hcl` file that explicitly
 set `skip = true` will be skipped.
 
+
+### Configuration parsing order
```
brikis98: Thanks for documenting this 👍

config/config.go
```diff
@@ -96,6 +96,19 @@ type ModuleDependencies struct {
 	Paths []string `hcl:"paths,attr"`
 }
 
+// Merge appends the paths in the proided ModuleDependencies object into this ModuleDependencies object.
```
brikis98:
```suggestion
// Merge appends the paths in the provided ModuleDependencies object into this ModuleDependencies object.
```

config/config.go
```diff
@@ -254,7 +267,29 @@ func ParseConfigFile(filename string, terragruntOptions *options.TerragruntOptio
...
+//    terragrunt config, so that it is accessible within the config. See PartialParseConfigString for a way to parse the
+//    blocks but avoid decoding.
+//    Allowed References:
+//      - locals
+// 4. Parse everything else. At this point, all the necessary building blocks for parsing the rest of the config are
+//    available, so parse the rest of the config.
+//    Allowed References:
+//      - locals
+//      - terragrunt_output
+// 5. Merge the included config with the parsed config. Note that all the config data is mergable except for `locals`
```
brikis98: Great docs 👍

util/logger.go
```diff
@@ -28,7 +29,7 @@ func CreateLoggerWithWriter(writer io.Writer, prefix string) *log.Logger {
 // logging solution.
 // Debugf will only print out terragrunt logs if the TG_LOG environment variable is set to DEBUG.
 func Debugf(logger *log.Logger, fmtString string, fmtArgs ...interface{}) {
-	if os.Getenv("TG_LOG") == "DEBUG" {
+	if strings.ToLower(os.Getenv("TG_LOG")) == "debug" {
```
brikis98: Do we document this anywhere?

README.md
yorinasub17: I like this. I went through and made this change.

util/logger.go
yorinasub17: Added to README

README.md
yorinasub17: fca4ac2

yorinasub17:
Ok merging and releasing this! Thanks for the review!

Issue #834, opened 2019-08-15
Update go-cty-yaml dependency


cspargo:
There are some bugs in go-cty-yaml that cause incorrect results when using the yamldecode function. These are documented in this issue https://github.com/hashicorp/terraform/issues/22223

I'm seeing this same issue when using yamldecode in terragrunt.hcl.

The go-cty-yaml dependency needs to be updated to version 1.0.1

cspargo:
actually i think this just needs the terraform dependencies updated to 0.12.6

cspargo:
to demonstrate the problem

main.tf:
```
variable "test1" { }
variable "test2" { }

output "test1" { value=var.test1 }
output "test2" { value=var.test2 }
```
terragrunt.hcl:
```
inputs = {
  test1 = yamldecode("a: '003'")
  test2 = yamldecode("b: '1.00'")
}
```

produces:

```
Outputs:

test1 = {"a":"3"}
test2 = {"b":"1"}
```

but should be

```
test1 = {"a":"003"}
test2 = {"b":"1.00"}
```

yorinasub17:
Fixed in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.19.21. Binaries should show up shortly.

PR #835, opened 2019-08-15
update go-cty-yaml dependency

cspargo:
Fixes https://github.com/gruntwork-io/terragrunt/issues/834


cspargo:
actually, it's probably best to update dep for github.com/hashicorp/terraform to 0.12.6. I'm going to close this.

Issue #837, opened 2019-08-16
Using output from a different module fails on the plan-all action, and works on apply-all


sa-mao:
Having this dependency declaration in module `example-app`:

```
dependency "infra_base" {
  config_path = "../stack-infra-base"
}

inputs = {
  public_subnets_ids = dependency.infra_base.outputs.public_subnets_ids
}
```

running `terragrunt plan-all` for the first time fails with this error:

```
example-app/terragrunt.hcl:14,53-72: Unsupported attribute; This object does not have an attribute named "public_subnets_ids".
[terragrunt] 2019/08/16 09:54:37 Unable to determine underlying exit code, so Terragrunt will exit with error code 1
```
If you run `terragrunt apply-all` the error is gone.

Is this an expected behavior?

brikis98:
When you say "`plan-all` for the first time fails", by "first time," do you mean _before_ `stack-infra-base` has been applied?

sa-mao:
Exactly, stack-infra-base is not applied yet.

brikis98:
@yorinasub17 The `dependency` block should probably detect if a module hasn't been applied and show a clear error to indicate this.

yorinasub17:
Opened PR to handle report better error messages: https://github.com/gruntwork-io/terragrunt/pull/840

In general, `plan-all` is not expected to work cleanly on a completely new infrastructure build if you have dependencies, because we don't have advanced features that stitch terraform plans together.

Internally in terraform, there is a concept of computable outputs which are then used to seed the inputs of other resources and modules, which "taint" and propagate through to predict if a resource would be created or not. In terragrunt, when you cross modules, we can't use that same terraform system to taint the plan across the modules because it is a completely different terraform module, and it is a huge overhead to implement something like that.

However, to allow you to simulate something like this, I also implemented a new feature: `default_outputs`. This is a static map that you can use as dummy outputs if no module has been applied, so that terraform can compute a `plan` or run `validate`. Otherwise, terragrunt can't compute the `inputs` so it will error out. Check out the PR and updated README for more info.

yorinasub17:
Resolved and released in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.19.23 (binaries should show up shortly). See the relevant section of the [README](https://github.com/gruntwork-io/terragrunt#unapplied-dependency-and-mock-outputs) for more info.

PR #840, opened 2019-08-16
Better error handling of dependency blocks

yorinasub17:
This addresses https://github.com/gruntwork-io/terragrunt/issues/837.

Additionally, this implements a new feature to `dependency` blocks that allow you to return default values when the target module hasn't been applied yet.

The primary use case for this is when you want to run a `terragrunt validate-all`, which previously would fail on a completely new infrastructure setup because the dependencies will fail to resolve and thus the `inputs` couldn't be passed in.

Now you can specify dummy values to use in the `outputs` attribute, and restrict to only use that if running `validate`. See the README changes for more info.


README.md
```diff
@@ -1114,6 +1114,68 @@ If any of the modules failed to deploy, then Terragrunt will not attempt to depl
...
+For example, in the previous example with a `mysql` module and `vpc` module, suppose you wanted to place in a temporary,
+dummy value for the `vpc_id` during a `validate-all` for the `mysql` module. You can specify in `mysql/terragrunt.hcl`:
+
+```
+dependency "vpc" {
+  config_path = "../vpc"
+
+  default_outputs = {
+    vpc_id = "temporary-dummy-id"
+  }
```
brikis98: Do `default_outputs` get used if the module has been applied, but doesn't have one of these outputs? Or _only_ if the module has never been applied at all?

brikis98: Also, will the user get a clear error if they forget to specify an output in `default_outputs`? E.g., let's say further down, they reference `depenency.vpc.outputs.public_subnet_ids`, but they never added `public_subnet_ids` to `default_outputs`. Will they immediately get an error that Terragrunt is using outputs, but one of the outputs they are looking up was not defined in `default_outputs`?

config/dependency.go
```diff
@@ -114,6 +115,86 @@ func getTerragruntOutput(dependencyConfig Dependency, terragruntOptions *options
...
+	targetConfig := getCleanedTargetConfigPath(dependencyConfig, terragruntOptions)
+	if !util.FileExists(targetConfig) {
+		return false, errors.WithStackTrace(DependencyConfigNotFound{Path: targetConfig})
+	}
+
+	// Check if the target terraform module (managed by terragrunt target config) has been applied.
+	// When a module hasn't been applied yet, there will be no state object, so `terragrunt show` will indicate that.
+	stdout, err := runTerragruntShow(terragruntOptions, targetConfig)
+	stdout = strings.TrimSpace(stdout)
+	isApplied := stdout != "No state."
```
brikis98: The downside to calling `show` is (a) it's an extra Terragrunt command to run, making this whole outputs thing a bit slower, (b) we are relying an an unofficial API.

Does the `terraform output` command perhaps return a specific exit code if there are no outputs or the module was never applied?

config/dependency.go
```diff
@@ -136,6 +217,23 @@ func getTerragruntOutput(dependencyConfig Dependency, terragruntOptions *options
...
+	targetOptions.TerraformCliArgs = []string{"show", "-no-color"}
+	targetOptions.Writer = stdoutBufferWriter
+	err := targetOptions.RunTerragrunt(targetOptions)
+	if err != nil {
+		return "", errors.WithStackTrace(err)
+	}
+
+	stdoutBufferWriter.Flush()
+	outputString := stdoutBuffer.String()
+	return outputString, nil
```
brikis98: Do we need a wrapper method for executing Terragrunt commands in a child module and handling all the stdout/stderr stuff? I feel like I've seen this large block of code before?

README.md
yorinasub17: > Do default_outputs get used if the module has been applied, but doesn't have one of these outputs? Or only if the module has never been applied at all?

Only if the module has never been applied at all. I think this is safer: having it pull from this default outputs on read and doesn't exist seem like it could cause some headaches.

> Also, will the user get a clear error if they forget to specify an output in default_outputs?

Ah that will not show a good error message, but this one is a bit harder to fix in the error because I can't pass through the info to the execution context, unless I implement what you suggest below.

> E.g., let's say further down, they reference depenency.vpc.outputs.public_subnet_ids, but they never added public_subnet_ids to default_outputs. Will they immediately get an error that Terragrunt is using outputs, but one of the outputs they are looking up was not defined in default_outputs?

Hmm I have to think a bit how this could be implemented. My guess is that it won't be trivial enough to include here in this PR, because I have to walk the tree to get the list of references and compare.

config/dependency.go
yorinasub17: > Does the terraform output command perhaps return a specific exit code if there are no outputs or the module was never applied?

It returns `{}` with no error, but that can be easily checked so added logic to check for this.

config/dependency.go
yorinasub17: Yes, but actually once I removed `show` it is no longer necessary, so I will keep the one output call block for now.

yorinasub17:
Ok updates based on feedback:

- No longer use `terragrunt show` to check if target module is applied and instead rely on whether or not the outputs json is empty: 677c6c8 
- Since the new way to check if a module is applied is reliant on the json output being an empty object, it conflates with the use case where there is no output or all the outputs are null, so add a way to skip getting outputs: 30d20fc  (NOTE: this is a neat feature for when we start to implement read config on the `dependency` block)


---

> is there any way to automatically guess a reasonable default value based off the types somehow?

This is non-trivial as we need to parse the terraform code to figure out the expected outputs and their types. The primary use case for the `default_outputs` is when nothing has been applied yet, which implies that there is no state file, which implies that the only way to know the output type is by inspecting the terraform code. Unless there is some magical terraform command you can use to get this about a module in a machine readable way (maybe the plan output?)

README.md
yorinasub17: NOTE: the following comment is no longer true

> Only if the module has never been applied at all. I think this is safer: having it pull from this default outputs on read and doesn't exist seem like it could cause some headaches.

Now, this will pull in the `default_outputs` if the module has not been applied, or if the module has no outputs.

config/dependency.go
yorinasub17: UPDATE: this no longer uses show and just checks for empty object.

brikis98: Hm, so this is interesting. If you run `terraform output` on an un-applied module, you get an error, with an exit code of 1:

```
$ terraform output
The state file either has no outputs defined, or all the defined
outputs are empty. Please define an output in your configuration
with the `output` keyword and run `terraform refresh` for it to
become available. If you are using interpolation, please verify
the interpolated value is not empty. You can use the 
`terraform console` command to assist.
```

But if you run `terraform output -json`, you get no error and an exit code of 0:

```
$ terraform output -json
{}
```

That... seems problematic.

brikis98:
> Unless there is some magical terraform command you can use to get this about a module in a machine readable way (maybe the plan output?)

Not that I know of. Only thing I can think of is parsing the `.tf` files in the folder, looking for `variable` and figuring out the type from that. Not impossible, but definitely not trivial.

config/dependency.go
yorinasub17: Yup...

In fact,

> If you run terraform output on an un-applied module, you get an error

also returns an error if you have either of the following conditions as well:
- You have no `output` blocks
- All the output values are `null`.

yorinasub17:
> something makes me think this will cost us something in the future.

I agree. Let me sit on this for another day and see if I come up with any ideas to improve this.

tomalok:
Are `default_outputs` actually used at any time except to provide `validate-all` with some values?

Maybe something along the lines of `validation_mock_outputs` would be clearer?

yorinasub17:
> Are default_outputs actually used at any time except to provide validate-all with some values?

`default_outputs` can also be used to construct the `plan-all` in the initial deployment. But outside of those use cases, there really isn't anything since the outputs will materialize in the `apply-all`.

tomalok:
`mock_outputs` (or something similar) would at least give the indication that these values aren't referenced at all when it comes time to `apply-all`...  "default" gives me the impression that if there's no resulting outputs for some reason, it'll be set to these values.

brikis98:
+1 on the idea around naming it `mock_outputs` (or `initial_outputs` or `unapplied_outputs`) or something else to indicate it's not used during normal operation.

yorinasub17:
UPDATE:

- Rebased on master to resolve conflict
- Renamed `default_outputs` to `mock_outputs`. Thanks for the suggestion @tomalok 

@brikis98 any objections to merging this?

brikis98:
Go for it!

yorinasub17:
Merging this in. Thanks for the review Jim, and thanks for the feedback @tomalok

PR #841, opened 2019-08-18
Fix yamldecode bug

yorinasub17:
Addresses https://github.com/gruntwork-io/terragrunt/issues/834


yorinasub17:
Thanks for the review! Merging this in.

Issue #454, opened 2018-03-28
Need some way to specify where extra arguments are placed


chrisportela:
For instance I add the following to all my terraform states

```
{
  terraform {
    extra_arguments "ouput_plan" {
      commands = [
        "plan"
      ]

      # With the get_tfvars_dir() function, you can use relative paths!
      arguments = [
        "-out=${get_tfvars_dir()}/tfplan.out"
      ]
    }

    extra_arguments "apply_plan" {
      commands = [
        "apply",
        "destroy"
      ]

      # With the get_tfvars_dir() function, you can use relative paths!
      arguments = [
        "${get_tfvars_dir()}/tfplan.out"
      ]
    }
  }
}
```

However when I run `apply-all` I get the following:
```
[snip]
[terragrunt] [/Users/cp/projects/work/infrastructure/src/aws/global] 2018/03/28 15:02:36 Running command: terraform apply /Users/cp/projects/work/infrastructure/src/aws/global/tfplan.out -input=false -auto-approve
Too many command line arguments. Configuration path expected.
[snip]
```

I run in to a lot of issues like being able to pass `-terragrunt-source` or `-terragrunt-source-update` flags (I need to use env) because the extra arguments are placed before any extra arguments given. I can see why that is since the feature is for **flags**, but I think this is still a valuable feature to have built in to terragrunt. 

As far as the feature being "need way to...": I'm not sure that's the best solution either, but maybe some field called `arguments_post` or `pre_arguments` would solve this in a way that is clear to others.

brikis98:
Looks like this method is not behaving as intended: https://github.com/gruntwork-io/terragrunt/blob/master/options/options.go#L177. PRs with a fix are welcome.

seppi91:
I also came across this issue when trying to implement a plan-all command which persists a planfile, that is used by apply-all later.
This would work perfectly fine if terragrunt wouldn't try to add `-input=false -auto-approve` at the end of the cmd. 

> [terragrunt] [/home/sebastian/Development/tg-boatyard/tg-timeseriesmonitor/aws/eu-central-1/base/rds] 2019/01/02 11:33:07 Detected 1 Hooks
[...]

PR #848, opened 2019-08-31
fix: reorg cli arguments

zot24:
Related to https://github.com/gruntwork-io/terragrunt/issues/454

This fix follows the recommendation mentioned in this comment https://github.com/gruntwork-io/terragrunt/issues/454#issuecomment-450831599 where we reorganize the hardcoded options `-input=false -auto-approve` to avoid the following error:

```
Running command: terraform apply terraform.tfplan -input=false -auto-approve
```

So we end instead with the following order in the command

```
terraform apply -input=false -auto-approve terraform.tfplan
```

Which follows the `terraform` help usage for the apply command `Usage: terraform apply [options] [DIR-OR-PLAN]`

This is the requested `gist` with the result of running the tests: https://gist.github.com/zot24/b6b2d90ba1b5b1008602dc0b8b031809 however there are a bunch of failing test that I don't see how those might be related with these changes :/ if someone can point me out in the right direction here if you thing they are I'll be happy to go over those

---

**Note:** there is a minor change as well added to this PR which is the swap of `-force` by `-auto-approve` on the `destroy` command as it will be deprecated in future versions of `terraform` as recommended on `help`

```
  -force                 Deprecated: same as auto-approve.
```

### DIFF
--- a/options/options.go
+++ b/options/options.go
@@ -235,8 +235,8 @@ func (terragruntOptions *TerragruntOptions) InsertTerraformCliArgs(argsToInsert 
 	// command is either 1 word or 2 words
 	var args []string
 	args = append(args, terragruntOptions.TerraformCliArgs[:commandLength]...)
-	args = append(args, argsToInsert...)
 	args = append(args, terragruntOptions.TerraformCliArgs[commandLength:]...)
+	args = append(args, argsToInsert...)
 	terragruntOptions.TerraformCliArgs = args
 }
 
--- a/configstack/stack.go
+++ b/configstack/stack.go
@@ -77,7 +77,7 @@ func (stack *Stack) Apply(terragruntOptions *options.TerragruntOptions) error {
 // Destroy all the modules in the given stack, making sure to destroy the dependencies of each module in the stack in
 // the proper order.
 func (stack *Stack) Destroy(terragruntOptions *options.TerragruntOptions) error {
-	stack.setTerraformCommand([]string{"destroy", "-force", "-input=false"})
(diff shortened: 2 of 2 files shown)


options/options.go
```diff
@@ -235,8 +235,8 @@ func (terragruntOptions *TerragruntOptions) InsertTerraformCliArgs(argsToInsert
 	// command is either 1 word or 2 words
 	var args []string
 	args = append(args, terragruntOptions.TerraformCliArgs[:commandLength]...)
-	args = append(args, argsToInsert...)
 	args = append(args, terragruntOptions.TerraformCliArgs[commandLength:]...)
+	args = append(args, argsToInsert...)
```
brikis98: Hm, doesn't this bypass all the `commandLength` logic above and just insert the args at the end? Does that work with all Terraform commands? That is, do `plan`, `apply`, `destroy`, `state mv`, `force-unlock <ID>`, etc all expect the args at the very end? Or do some need it in the middle?

yves-del-medico:
I'm also bumping into that issue.

Is there a real need of injecting these options `-input=false -auto-approve` automatically when running `apply-all`.

In my opinion, this should be a choice of the user (via `extra_arguments`) and not imposed silently by terragrunt.

Does it make sense ?

options/options.go
zot24: I think you are right this might not work for all commands, I think the solution I'm proposing here it's a bit sloppy and just for my specific just case

zot24:
I'm going to close this as I won't have time to look into it however I agree with @yves-del-medico and not enforcing this could be a good solution and let that to the user

Issue #844, opened 2019-08-27
[BUG] infinite loop with dependency outputs feature


jonathancolby-olx:
The dependency configuration item was introduced to pass outputs between modules.   It is indeed very useful (thanks for that!!). However,  the `dependency` configuration item also creates a run dependency,  as confirmed in the README (`Note that each dependency is automatically considered a dependency in Terragrunt. `)

This can result in an endless execution loop. 

example.

```
=================
foo/terraform.hcl

dependency "bar" {
  config_path = "../bar"
}

inputs = {
  bar_stuff = dependency.bar.outputs.domain
}

=================
bar/terraform.hcl

dependency "foo" {
  config_path = "../foo"
}

inputs = {
  foo_stuff = dependency.foo.outputs.domain
}

```

The above configuration will cause terragrunt to go back and forth between the two modules, and run terraform outputs endlessly.

A possible enhancement would be to not treat `dependency` as an execution dependency, only as a way to extract terraform outputs

brikis98:
Thanks for the report! We actually already have rudimentary logic to detect cycles, so we probably just forgot to use it when implementing the `dependency` block. @yorinasub17 When you're back from vacation, could you take a look?

yorinasub17:
I'll look into providing a clearer error message for cycles.

---

> A possible enhancement would be to not treat dependency as an execution dependency, only as a way to extract terraform outputs

I'd like to understand your use case for this better. This behavior would be very confusing to me.

The reason why we treat the dependency as an execution dependency is that it doesn't make sense to depend on outputs of the data of something that hasn't been applied. You almost always want to get the "freshest" state of data, or you can get into nasty situations in an `apply-all` where the changes don't propagate correctly. For example, if you have a module that depends on a subnet defined in a VPC module, and you have changes such that the subnet recreates or you have a new subnet, then not treating the dependency lookup as an execution dependency would cause the `apply-all` to possibly finish against the old subnet because of the execution order. Not to mention that execution order matters a lot in the initial deployment with `apply-all`, where you don't have anything deployed.

I have a hard time seeing a use case where you DON'T want to have the "freshest" dependency in an `xxx-all` situation, so it would help if you can share your use case for the above suggestion.

yorinasub17:
[...]

PR #850, opened 2019-09-03
Detect cycles in dependency block parsing

yorinasub17:
This implements a friendlier error message instead of an infinite loop for the case described in https://github.com/gruntwork-io/terragrunt/issues/844

Specifically, because `dependency` block output reading depends on `terragrunt output`, and because `terragrunt output` depends on parsing the config, you can end up in a situation where a simple `terragrunt plan` on a config that has a cyclic dependency would loop indefinitely trying to read `terragrunt output` across all the dependencies.


config/dependency.go
```diff
@@ -25,15 +26,20 @@ type Dependency struct {
 // Decode the dependency blocks from the file, and then retrieve all the outputs from the remote state. Then encode the
 // resulting map as a cty.Value object.
 // TODO: In the future, consider allowing importing dependency blocks from included config
+// NOTE FOR MAINTAINER: When implementing importation of other config blocks (e.g referencing inputs), carefully
+//                      consider whether or not the implementation of the cyclic dependency detection still makes sense.
```
yorinasub17: Depending on how this is implemented, what we have now may or may not work. E.g if someone specifies `skip_output` to `true`, should that dependency be an explicit dependency of the module such that it affects the execution order? If no, then the current implementation works because it will exclude that dependency from the dependency list.

test/integration_test.go
```diff
@@ -1652,7 +1652,29 @@ func TestDependencyOutputTypeConversion(t *testing.T) {
 	assert.Equal(t, outputs["object"].Value, map[string]interface{}{"list": []interface{}{1.0, 2.0, 3.0}, "map": map[string]interface{}{"foo": "bar"}, "num": 42.0, "str": "string"})
 	assert.Equal(t, outputs["string"].Value, "string")
 	assert.Equal(t, outputs["from_env"].Value, "default")
+}
+
+// Test that we get the expected error message about dependency cycles when there is a cycle in the dependency chain
+func TestDependencyOutputCycleHandling(t *testing.T) {
```
brikis98: What type of cycle does this test? I'd want to check:

A -> B -> A

A -> B -> C -> D -> A

A -> A

yorinasub17: Good point. Added each of these cases and another one:

A -> B -> A
 |
 -> C

yorinasub17:
Merging this in. Thanks for the review!

PR #852, opened 2019-09-05
RFC Template

yorinasub17:
This proposes a new practice of writing RFCs for proposing reasonably complex Terragrunt features that may be difficult to capture in a single Github Issue.

This PR includes:

- Update to the Contribution guideline to mention RFC practice.
- A template that future RFCs can use. This ensures a standardization of the format and makes sure all the information we expect from RFCs are included.


yorinasub17:
Merging this in. Thanks for the review! Will use the template to write up the RFC for the iteration feature.

PR #853, opened 2019-09-05
[RFC] Iteration of modules

yorinasub17:
This proposes an RFC for potentially supporting iteration of modules in Terragrunt. The document contains the background for why we should consider this.

### DIFF
--- a/_docs/rfc/for_each_iteration.md
+++ b/_docs/rfc/for_each_iteration.md
@@ -204,8 +204,13 @@ to demonstrate the approach.
 
 **Cons**:
 
-- Code generation is complex. When should you run the code generation step? How often? How do you know if there is
-  drift?
+- Code generation is complex.
+    - When should you run the code generation step?
+    - How often?
+    - How do you know if there is drift?
+    - Do you check in the generated code? If so, you have to maintain it (e.g., 19 sets of files/folders for 19 AWS
+      regions). If it's always generated at runtime, then interacting with just one of the generated things becomes
+      harder.
 - Removing an iteration still requires a bit of a dance. You need to `terragrunt destroy` the relevant folder, then
   remove that folder, and then remove the corresponding item from the iteration list.
 - Users now need to learn a templating tool and language, on top of Terragrunt and Terraform.
--- a/_docs/rfc/for_each_iteration.md
+++ b/_docs/rfc/for_each_iteration.md
@@ -1,0 +1,218 @@
+# RFC: for_each - looping variables to call module multiple times
+
+**STATUS**: In proposal _(This should be updated when you open a PR for the implementation)_
+
+
+## Background
+
+Oftentimes it is desirable to repeatedly apply a module multiple times with only a few modifications. This is common
+when you have a canonical way to provision resources and need to quickly produce copies with one or two modifications.
+For example, in Kubernetes, you might have a Namespace factory module that provisions Namespaces in a canonical way that
+you configure with a common set of variables, and the only thing that differs across multiple Namespaces is the name.
+
+Another use case is in achieving compliance with the CIS benchmark for AWS. One of the requirements in the benchmark is
+that you need to enable AWS Config in all enabled regions on your account. In this case, you would want to loop over
+each enabled region and run the module that provisions and configures AWS Config using the same args.
+
+Ideally, we can handle this in the Terraform module by looping over `module` blocks:
+
+```
+module "aws_config" {
+  for_each = var.all_enabled_regions
+  region   = each.key
+
+  # additional arguments omitted for brevity
+}
+```
+
+However, [as of 0.12.7, this is still not available](https://github.com/hashicorp/terraform/issues/17519). That said,
+this is being developed and there is reason to believe that this will eventually be available, especially since,
+starting with Terraform 0.12.0, `count` and `for_each` has been reserved on `module` blocks. What is not known is how
+long it will take before `for_each` is implemented on modules.
+
+There are a few advantages to a Terragrunt implementation of `for_each` to call modules repeatedly with different
+parameters:
+
+- Provide a workaround sooner than Terraform might implement module `for_each` and `count`.
+- You can keep separate state files for each module call. While this does not help with isolation of blast radius (due
+  to the looping interface), this helps with isolating access to individual state files. For example, you can adjust the
+  remote state config such that the bucket is parameterized by the looping argument, allowing you to store the state of
+  each module call in a different S3 bucket.
+- You can allow separate IAM roles for each loop iteration. This can be done if you parameterize the `iam_role`
+  property with the looping argument.
+
+For this reason, it is worth contemplating a possible implementation path for supporting iteration in Terragrunt to
+repeatedly call a module with different parameters. Note when considering an implementation, it is important to optimize
+for **speed of implementation and simplicity**, since it is very likely that this feature will be made obsolete by
+`for_each` on `module` blocks.
+
+
+## Proposed solution
+
+_Intentionally blank, as the solution is not obvious._
+
+
+## Alternatives
+
+### Option 1: Null option - do nothing
+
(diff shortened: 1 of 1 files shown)


yorinasub17:
cc @bwhaley

_docs/rfc/for_each_iteration.md
```diff
@@ -0,0 +1,218 @@
...
+  `.terragrunt-cache/HASH_OF_EACH_VALUE/path/to/workspace`)? Or should it be one cache per iteration value (e.g
+  `HASH_OF_EACH_VALUE/.terragrunt-cache/path/to/workspace`)?
+- Implementation requires a major refactor in the calling logic post parsing.
+- It is not immediately obvious how to handle removing an item in the `for_each` list. A naive approach would not run
+  `terragrunt destroy` on that module. Additionally, as stated, it is impossible to run the command on a single item in
+  the list and thus you can't destroy just the items you want to remove without doing complex dancing (e.g update the
+  `for_each` list to only reference the items you want to destroy, then run `terragrunt destroy` to destroy just those
+  items, and then restore the list to the version with the items removed).
+
+### Option 3: scaffolding tool that code gens live config using a template
```
yorinasub17: @bwhaley : After contemplating about this feature, I am leaning towards this approach for our purposes. The implementation for supporting iteration in Terragrunt has many nuances that suggest it is not a great idea.

Better to wait for the official solution, working around it using code generation. Code generation appears to be the lesser evil of a workaround.

_docs/rfc/for_each_iteration.md
```diff
@@ -0,0 +1,218 @@
...
+  iteration.
+- The parsing functions need to be updated to first parse the `for_each` attribute. Then, the parser should iterate each
+  item in the list and set the `each` variable accordingly as it parses the rest of the config. This means the parsing
+  order is now adjusted as:
+    - `for_each` and expand loop.
+    - `locals`
+    - `include`
+    - `dependency`
+    - Everything else
+- Support `merge_parent` in `remote_state`, which ultimately adjusts how `mergeConfigWithIncludedConfig` merges the
```
brikis98: Is there a way for us to check the user isn't accidentally overriding the config multiple times? Does `remote_state` config in Terragrunt become required to use this feature? Or do we read the `.terraform/terraform.tfstate` file? Seems a bit complicated.

_docs/rfc/for_each_iteration.md
```diff
@@ -0,0 +1,218 @@
...
+
+- Parallels `for_each` in Terraform. In the examples I used `list`, but the implementation details can mimic that of
+  Terraform (requiring `set` or `map`). That way, the syntax closely resembles an existing construct and thus it will be
+  easier to pick up.
+- Limited changes are necessary in the user config to add in iteration.
+- The config parser implementation is relatively simple.
+
+**Cons**:
+
+- The implementation of the runners can become complex, especially in the dependency detection for the `xxx-all`
```
brikis98: Yea, there are a lot of complexities here, and many, many ways to get it wrong.

_docs/rfc/for_each_iteration.md
```diff
@@ -0,0 +1,218 @@
...
+to demonstrate the approach.
+
+**Pros**:
+
+- Keeps Terragrunt clean, since no new code is introduced.
+- Available workaround now, without any work.
+
+**Cons**:
+
+- Code generation is complex. When should you run the code generation step? How often? How do you know if there is
```
brikis98: Do you check in the generated code? If so, you have to maintain it (e.g., 19 sets of files/folders for 19 AWS regions). If it's always generated at runtime, then interacting with just one of the generated things becomes harder.

antonbabenko:
@yorinasub17 Thanks for this RFC! Here are my 5 cents...

After reading the whole RFC and solving it using cookiecutter before, I like code generation as the proposed solution.

1. I think keeping all generated code in a repo (even if there are 19*19 similar folders) is better because this will allow working on any of them differently (eg, try changes in playground AWS account/region before propagating to the rest).

1. Keeping more files and folders is better for humans to understand the whole footprint (which regions do we use? how they are configured?)

1. Code generation tool can be not related to terragrunt as such because the same is needed for native terraform projects where multiple similar modules have to be managed (while we are waiting for `for_each` with modules).

1. Code generation (especially if implemented as another tool, see above) will allow to keep terragrunt user-friendly and not overload it with features.

yorinasub17:
>  I'd be tempted to do the generate in the Terraform code (.tf files), simulating for_each on the module by just generating N copies of the module from some template, and committing that. Later, when for_each is officially added to Terraform, remove the generated code, and replace it with one module with the for_each on it.

Ah that is a good point.

---

> I'm reluctantly a +1 on code generation.

The one thing I am not sure about is if this should be a first class thing in terragrunt. I think this meshes really well with the scaffolding idea (which should be a separate RFC). As in, if we have the `terragrunt scaffold` (or `terragrunt codegen`, which I like slightly better) feature, this need of iteration can be resolved by the `codegen` system supporting template processing. What do you think of that approach?

_docs/rfc/for_each_iteration.md
yorinasub17: Valid points! Adding this in as one of the concerns.

_docs/rfc/for_each_iteration.md
yorinasub17: I think local state will work because I was imagining each iteration will have it's own directory in the cache, but yes this isn't immediately obvious.

yorinasub17:
> Here are my 5 cents...

Thanks for your input @antonbabenko !

> I think keeping all generated code in a repo (even if there are 19*19 similar folders) is better because this will allow working on any of them differently (eg, try changes in playground AWS account/region before propagating to the rest).

Good point, although I would be concerned about accidentally checking in the changes in the code itself instead of the generator. This could happen if one is not aware of the codegen process.

> Code generation (especially if implemented as another tool, see above) will allow to keep terragrunt user-friendly and not overload it with features.

I understand this sentiment and we definitely want to keep terragrunt simple as possible. That said, having codegen baked into terragrunt has some advantages. I plan on writing a different RFC with our latest thoughts around this within Gruntwork which highlights this.

bwhaley:
Thanks @yorinasub17 for writing this RFC and thinking through the problem, and thanks to everyone else for your input. Seems like the consensus is that adding iteration to `terragrunt` will be hard, and given that `for_each` is coming to terraform modules natively very soon, it probably isn't worthwhile to add it here.

Does it even make sense to include code generation (including examples or docs) here in `terragrunt`? Seems tangential and perhaps should live elsewhere. 

Since we need the "run this module in all regions" functionality specifically for the AWS Config module as part of a CIS compliance project, perhaps we should just do the code generation in that repo instead, and close this one out. We could have a new PR in that repo discussing exactly how the code generation would work, including addressing some of the cons raised here. For example:

> When should you run the code generation step?
After modifying the template.
> How often?
Every time the generated code for all regions need to change.

> How do you know if there is drift?
All the generated files should have headers saying "don't touch this, it's generated code." The generated files would be overwritten any time changes are made to the master.

> Do you check in the generated code? If so, you have to maintain it (e.g., 19 sets of files/folders for 19 AWS regions). If it's always generated at runtime, then interacting with just one of the generated things becomes harder.
Yes, always check in the generated code, and never edit anything but the master.

Obviously there are some deeper complexities that will come up during development, but the point is, code generation probably doesn't belong in terragrunt.

yorinasub17:
Given all the discussions, I believe we have reached the conclusion that we shouldn't implement anything specific in Terragrunt regarding this specific use case. There may be other use cases for having code generation in Terragrunt, but those should be optimized for the use cases where it makes sense to bake into Terragrunt.

I have updated the RFC with the recommendation and conclusions. Going to merge this in now so we have it in the repo for reference.

Thanks to everyone who took the time to review this!

smitthakkar96:
Hey since terraform implemented support for `for_each`, can we revisit this?

PR #863, opened 2019-09-13
Adds support for Terragrunt JSON configuration files

### LINKED ISSUE #846: terragrunt.hcl.json ?
Terraform (and HCL parser) support json input file.
Is it possible to use json as input to terragrunt ?

https://www.terraform.io/docs/configuration/syntax-json.html

jakauppila:
Adds support for Terragrunt JSON configuration files.

Updated a number of tests that mirror the HCL configuration.

Fixes #846


config/config.go
```diff
@@ -414,12 +418,23 @@ func parseHcl(parser *hclparse.Parser, hcl string, filename string) (file *hcl.F
 		}
 	}()
 
-	file, parseDiagnostics := parser.ParseHCL([]byte(hcl), filename)
-	if parseDiagnostics != nil && parseDiagnostics.HasErrors() {
-		return nil, parseDiagnostics
+	if filepath.Ext(filename) == ".hcl" {
```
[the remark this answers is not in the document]
yorinasub17: Here are a few other options:

### Option 1: Trial and error

The code can try to parse as HCL first, and then if there are errors, fall back to parsing as JSON (or other way around). Since most of the implementations out there will be HCL, I think doing HCL first makes sense for efficiency reasons.

Note that for this implementation, you need to create a `parseJson` function as opposed to doing the if check here, because you won't be able to handle the panics correctly otherwise. The idea would be to have three functions:

```
parseHclOrJson
parseHcl
parseJson
```

where `parseHcl` remains the same as it was before.

A problem with this approach is that you don't want to report the json parsing errors if the file was meant to be parsed as HCL (or vice versa), so you will need to make sure to report both parsing errors.


### Option 2: Heuristics based approach

Since HCL compatible JSON syntax is always an object, you can check if the first non-whitespace character is `{` and the last non-whitespace character is `}`. This could be brittle if the terragrunt syntax changes and the top level item is a literal map expression in hcl2.

### Option 3: Command line option

This could be configured as a command line option `--terragrunt-syntax-type`, so that it is explicit. This would be a cumbersome interface though.

---

Having laid out these options, I'd vote for keeping the file extension approach you have, with a minor adjustment. We should treat `json` as the special case, and always default to `hcl`. So the if condition should be:

```
if filepath.Ext(filename) == ".json" {
    // parse as json
}
// parse as HCL
```

This way, we preserve the old behavior for any file extension except for those that have `.json`, and I can't imagine anyone intending to use terragrunt with a `.json` syntax expecting HCL parsing (vs you might expect that with something like a `.terragrunt` file).

config/config.go
```diff
@@ -414,12 +418,23 @@ func parseHcl(parser *hclparse.Parser, hcl string, filename string) (file *hcl.F
...
+		}
+	
+		return file, nil
+	}else if filepath.Ext(filename) == ".json" {
+		file, parseDiagnostics := parser.ParseJSON([]byte(hcl), filename)
+		if parseDiagnostics != nil && parseDiagnostics.HasErrors() {
+			return nil, parseDiagnostics
+		}
+	
+		return file, nil
```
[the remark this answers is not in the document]
yorinasub17: You are getting that because it won't fit the type. Without it, you have a code path where you don't return anything (when `filename` doesn't end with `hcl` or `json`). Golang requires you to exhaust all the options in the static type checking so you need that last `else` block.

Note that if you follow the suggestion above, where you only have the `json` condition, that last `else` block goes away.

config/config.go
```diff
@@ -19,6 +19,7 @@ import (
 )
 
 const DefaultTerragruntConfigPath = "terragrunt.hcl"
+const DefaultTerragruntJsonConfigPath = "terragrunt.hcl.json"
```
yorinasub17: There's probably a few places where you need to update the usage of `DefaultTerragruntConfigPath` to also look for `DefaultTerragruntJsonConfigPath`. At a minimum, `FindConfigFilesInPath` (
https://github.com/gruntwork-io/terragrunt/blob/master/config/config.go#L192) needs to be updated to look for the json flavor so that `xxx-all` works. This adds some complexity to the logic, but if we are to support `json`, we need to make sure all the commands work or else I can see a lot of issues.

config/config_test.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/stretchr/testify/require"
```
yorinasub17: These unit tests are good! But we will also want some [integration tests](https://github.com/gruntwork-io/terragrunt/tree/master/test) though. At a minimum, I'd like to see tests for what happens when:

- Terragrunt module only has `terragrunt.hcl.json` and you run `terragrunt plan` (no args to specify the different file).
- Multiple terragrunt modules only have `terragrunt.hcl.json` and you run `terragrunt plan-all`
- Multiple terragrunt modules with a mix of `terragrunt.hcl` and `terragrrunt.hcl.json` and you run `terragrunt plan-all`

Some more interesting test cases for to be comprehensive:

- Child terragrunt module that includes a parent terragrunt config that is in JSON format, and vice versa.
- Terragrunt module that marks a json based terragrunt module as a dependency in a `dependency` block, and vice versa.
- Terragrunt module that marks a json based terragrunt module as a dependency in a `dependencies` block, and vice versa. Also when there is a mix of json and hcl based modules as `dependency` and `dependencies` blocks.

jakauppila:
Carved out some time in my sprint to work on this, so I'll be pushing updates in the next few days.

config/config_test.go
jakauppila: @yorinasub17 For the first integration test, TerragruntOptions default the value of terragruntConfigPath to be `terragrunt.hcl` at https://github.com/gruntwork-io/terragrunt/blob/master/cli/args.go#L66.

A couple of questions:
- What is the desired behavior? Error out?
- Should `parseTerragruntOptionsFromArgs` interrogate the files in the working path to determine whether we set the default config to be .hcl or .hcl.json ?
  - If so, which file should take precedence? Or should it error if both are found?

```
[terragrunt] 2019/11/11 10:58:13 Reading Terragrunt config file at fixture-config-files/one-json-config/subdir/terragrunt.hcl
--- FAIL: TestTerragruntWorksWithSingleJsonConfig (0.00s)
    c:\go_home\go\src\github.com\gruntwork-io\terragrunt\test\integration_test.go:2248: Failed to run Terragrunt command 'terragrunt plan --terragrunt-non-interactive --terragrunt-working-dir fixture-config-files/one-json-config/subdir' due to error: Error reading file at path fixture-config-files/one-json-config/subdir/terragrunt.hcl: open fixture-config-files/one-json-config/subdir/terragrunt.hcl: The system cannot find the file specified.
```

yorinasub17: I think the logic should be to look for `terragrunt.hcl`, and if it doesn't exist, use `terragrunt.hcl.json`. I think `terragrunt.hcl` should take precedence for the same reasons mentioned in https://github.com/gruntwork-io/terragrunt/pull/863#discussion_r324063211.

jakauppila:
Pushed more changes including a handful of tests, let me know if there's any additional tests you'd like to see

yorinasub17:
Build passed, so going to merge and release this. Thanks for your contribution!

Issue #822, opened 2019-08-08
Build issue v0.19.16


forficate:
Hi, I am trying to build terragrunt from source but get the below error message:
```
github.com/gruntwork-io/terragrunt/config
# github.com/gruntwork-io/terragrunt/config
go/src/github.com/gruntwork-io/terragrunt/config/config_helpers.go:95:16: cannot use v (type "github.com/hashicorp/terraform/vendor/github.com/zclconf/go-cty/cty/function".Function) as type "github.com/zclconf/go-cty/cty/function".Function in assignment
```

Tracking this down the break in compilation seems to happen between `v0.19.3` and `v0.19.4`.

Doing a diff on the two tags I can't see anything in `config_helpers` that has changed to cause that error. I do see `go-cty-yaml` is added and not sure if that would cause the issue:

```
+[[projects]]
+  digest = "1:7922171eda409496bd43979d690c2ba5d58912d9bc5cafdc0509e7cd5a6f8314"
+  name = "github.com/zclconf/go-cty-yaml"
+  packages = ["."]
+  pruneopts = "UT"
+  revision = "fbabe99bb1c897a56ed70103a40aa94fc50a7104"
+  version = "v0.1.0"
+```

**Comment by u-quark** [comment]
@forficate Did you find a solution for this?

**Comment by forficate** [comment]
No I didn't for v0.19.16.

However after creating this issue v0.19.19 was released and that builds ok.

Note I'm using [nix](https://nixos.org/nix/) to build with `go2nix` and `dep2nix`.

For reference the `terragrunt` derivation:
```
# This file was generated by https://github.com/kamilchm/go2nix v1.3.0
{ stdenv, buildGoPackage, fetchFromGitHub, fetchhg, fetchbzr, fetchsvn, makeWrapper, terraform }:

buildGoPackage rec {
  name = "terragrunt-${version}";
  version = "0.19.19";

  goPackagePath = "github.com/gruntwork-io/terragrunt";

  src = fetchFromGitHub {
    owner = "gruntwork-io";
    repo  = "terragrunt";
    rev   = "v${version}";
    sha256 = "0j7287g9lw06xvbxv9xi0g8zib0fbkihy67g7n1yygh9c3dbdxxi";
  };

  goDeps = ./deps.nix;

  buildInputs = [ makeWrapper ];

  preBuild = ''
    buildFlagsArray+=("-ldflags" "-X main.VERSION=v${version}")
  '';

  postInstall = ''
[...]

PR #866, opened 2019-09-16
 Fix bug where the context for running terraform output on the child was incorrect

yorinasub17:
This fixes the bug reported in https://community.gruntwork.io/t/backend-local-outputs-not-found/341

The main issue was that `terragruntOptions.Clone` doesn't update the `DownloadDir` to be in the context of the dependency config. This causes problems when using local state, as the state is stored in the terragrunt download directory.

### DIFF
--- a/config/dependency.go
+++ b/config/dependency.go
@@ -290,15 +290,43 @@ func getTerragruntOutput(dependencyConfig Dependency, terragruntOptions *options
 	return &convertedOutput, isEmpty, errors.WithStackTrace(err)
 }
 
+// Clone terragrunt options and update context for dependency block so that the outputs can be read correctly
+func cloneTerragruntOptionsForDependencyOutput(terragruntOptions *options.TerragruntOptions, targetConfig string) (*options.TerragruntOptions, error) {
+	targetOptions := terragruntOptions.Clone(targetConfig)
+	targetOptions.TerraformCliArgs = []string{"output", "-json"}
+
+	// DownloadDir needs to be updated to be in the context of the new config, if using default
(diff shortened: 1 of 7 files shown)


ashemedai:
Trying to compile this, with `go get` on your checked out branch, I did run into #822. Removing `hashicorp/terraform/vendor/github.com/zclconf/go-cty` allowed me to continue compiling.

config/dependency.go
```diff
@@ -290,15 +290,43 @@ func getTerragruntOutput(dependencyConfig Dependency, terragruntOptions *options
...
+	targetOptions := terragruntOptions.Clone(targetConfig)
+	targetOptions.TerraformCliArgs = []string{"output", "-json"}
+
+	// DownloadDir needs to be updated to be in the context of the new config, if using default
+	_, originalDefaultDownloadDir, err := options.DefaultWorkingAndDownloadDirs(terragruntOptions.TerragruntConfigPath)
+	if err != nil {
+		return nil, errors.WithStackTrace(err)
+	}
+
+	// Using default, so compute new download dir and update
```
brikis98: I'm missing some context here... Why do you need to check for the default here?

yorinasub17: If you set the download dir to a non-standard location using `--terragrunt-download-dir`, then we want to directly use the one set on the original config instead of overriding that back to the default. However, since we don't have the CLI context object here, we can't detect that the user passed that CLI in in this function. So we instead rely on the heuristic that if the original is using the default download dir, then we should use the default download dir in the context of the target config.

yorinasub17:
Merging and releasing. Thanks for the review!

PR #871, opened 2019-09-20
Fix bucket creation for remote_state for gcs

brimstone:
I wasn't able to convince terragrunt v0.19.25 to create a bucket in GCS for storing remote state. It kept saying:
```
Missing required GCS remote state configuration project
```
I tracked this down to the bit where the remote_state config {} section is cleaned up BEFORE the bucket is created.

This PR fixes it in my limited testing. I'm not sure how to run only the GCS tests so pointers there would be appreciated.

### DIFF
--- a/remote/remote_state_gcs.go
+++ b/remote/remote_state_gcs.go
@@ -60,12 +60,17 @@ type GCSInitializer struct{}
 // 1. Any of the existing backend settings are different than the current config
 // 2. The configured GCS bucket does not exist
 func (gcsInitializer GCSInitializer) NeedsInitialization(remoteState *RemoteState, existingBackend *TerraformBackend, terragruntOptions *options.TerragruntOptions) (bool, error) {
+
 	if remoteState.DisableInit {
 		return false, nil
 	}
 
+	project := remoteState.Config["project"]
 	if !gcsConfigValuesEqual(remoteState.Config, existingBackend, terragruntOptions) {
 		return true, nil
+	}
+	if project != nil {
+		remoteState.Config["project"] = project
 	}
 
 	gcsConfig, err := parseGCSConfig(remoteState.Config)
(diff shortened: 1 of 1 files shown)


remote/remote_state_gcs.go
```diff
@@ -82,6 +82,11 @@ func (gcsInitializer GCSInitializer) NeedsInitialization(remoteState *RemoteStat
 		return true, nil
 	}
 
+	// Delete custom AFTER the bucket is created GCS labels that are only used in Terragrunt config and not in Terraform's backend
```
yorinasub17:
```suggestion
	// Delete custom GCS labels that are only used in Terragrunt config and not in Terraform's backend AFTER the bucket is created 
```

yorinasub17:
This test `TestGCSConfigValuesEqual/equal-ignore-gcs-labels` failed:
```
    --- FAIL: TestGCSConfigValuesEqual/equal-ignore-gcs-labels (0.00s)
        remote_state_gcs_test.go:110: 
            	Error Trace:	remote_state_gcs_test.go:110
            	Error:      	Not equal: 
            	            	expected: true
            	            	actual  : false
            	Test:       	TestGCSConfigValuesEqual/equal-ignore-gcs-labels
```

brimstone:
I think the test is fixed. I don't see a machine confirming.

yorinasub17:
Just kicked off a build!

yorinasub17:
@robmorgan can you also take a look to double check the functionality? The change makes sense to me, but since you wrote the original code, definitely could use another set of eyes here.

brimstone:
Anything I can do to help merge this?

brikis98:
@robmorgan ^

robmorgan:
Hi all, sorry about the wait. I'll take a look tomorrow.

brimstone:
@robmorgan Thanks! Not tomorrow is ok too, if it doesn't fall off your radar.

robmorgan:
Okay, all the tests seem to pass locally for me. Let me check once more on CircleCI.

robmorgan:
Released in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.20.3

PR #872, opened 2019-09-23
Obey s3 backend skip_bucket_versioning flag

eamonnotoole:
Simple patch to ensure that the s3 backend skip_bucket_versioning flag
is obeyed.


eamonnotoole:
Sorry - meant to only carry this on my fork for the moment.

Issue #745, opened 2019-06-13
terragrunt S3 backend permission error when using from another aws account


truptiak:
I am using two accounts as MainAWS account & TestAWS account. We have created the bucket in MainAWS account as "xxxxxbucket". Also created dynamodb table as "tf-state-lock" with LockID.

Created the below structure to use my module with terragrunt & keep my remote backend config DRY -

# tree
.
├── non-prod
│   ├
│   └── us-east-1
│       └── vpc
│           └── terraform.tfvars
├── README.md
└── terraform.tfvars

root level terraform.tfvars as below :

terragrunt = {
  terraform {
    source = "."
  }

  # Configure Terragrunt to automatically store tfstate files in an S3 bucket

  remote_state {
    backend = "s3"

    config {
      encrypt = true
      bucket  = "xxxxxbucket"
      key     = "${path_relative_to_include()}/terraform.tfstate"
      region  = "us-east-1"

     dynamodb_table                 = "tf-state-lock"
     enable_lock_table_ssencryption = true

      skip_bucket_versioning      = true
      skip_bucket_ssencryption    = true
      skip_bucket_accesslogging   = true
      skip_region_validation      = true
      skip_credentials_validation = true
      skip_metadata_api_check     = true
      force_path_style            = true
     }	
    }
  }
}

Child level terraform.tfvars as below : 
terragrunt = {
  terraform {
       source = "git::ssh://git@gitlab.aws.exoduspoint.com:2222/core-infrastructure/terraform-modules.git//aws/vpc/?ref=master"
  }

# Include all settings from the root terraform.tfvars file

  include {
    path = "${find_in_parent_folders()}"
  }
}

# These are the variables we have to pass in to use the module specified in the terragrunt configuration above

 name   = "my-vpc"
 cidr = "10.0.0.0/16"
 aws_region = "us-east-1"

Attached below policy to my terraform user in account TestAWS -

{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "AllowCreateAndListS3ActionsOnSpecifiedTerragruntBucket",
            "Effect": "Allow",
            "Action": [
                "s3:ListBucket",
                "s3:GetBucketVersioning",
                "s3:CreateBucket"
            ],
            "Resource": "arn:aws:s3:::xxxxxbucket"
        },
        {
            "Sid": "AllowGetAndPutS3ActionsOnSpecifiedTerragruntBucketPath",
            "Effect": "Allow",
            "Action": [
[...]

PR #874, opened 2019-09-23
Skip versioning check on S3 bucket when  skip_bucket_versioning is set to true, fixes #867,#745

### LINKED ISSUE #867: S3 - Is checking permission `GetBucketVersioning` required?
It was a bit confusing when configuring Terragrunt under CI/CD and looking at the Terraform required [S3/DynamoDB IAM policies defined in the documentation](https://www.terraform.io/docs/backends/types/s3.html#s3-bucket-permissions) yet getting 403s in execution.

I feel like creating the bucket with versioning (and disabling it if the configuration is set) is fine, but always checking it on initialization seems unnecessary ?

I see a few options:

1. Remove the `GetBucketVersioning` check on initialization, required permissions now match Terraform documentation.
2. Setting `SkipBucketVersioning` within the S3 config also skips checking on initialization.
3. Leave as-is, it's my fault for not reading through the Terragrunt readme over again

---

`SkipBucketVersioning` skips enabling bucket versioning when Terragrunt creates a bucket at https://github.com/gruntwork-io/terragrunt/blob/272c78fe20a10e35a9ce4b8b591609dcc01283c3/remote/remote_state_s3.go#L365

But it doesn't prevent Terragrunt from checking if bucket versioning is enabled on intialization at https://github.com/gruntwork-io/terragrunt/blob/272c78fe20a10e35a9ce4b8b591609dcc01283c3/remote/remote_state_s3.go#L214

mattiarossi:
I have added a check on SkipBucketVersioning before calling checkIfVersioningEnabled, and updated the README accordingly.
There was no related unit test so I didn't add one

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ jobs:
     - attach_workspace:
         at: *workspace_root
     - run: cd bin && sha256sum * > SHA256SUMS
-    - run: upload-github-release-assets
+    - run: upload-github-release-assets bin/*
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ jobs:
     - attach_workspace:
         at: *workspace_root
     - run: cd bin && sha256sum * > SHA256SUMS
-    - run: upload-github-release-assets bin/*
+    - run: upload-github-release-assets --github-repo-name terragrunt-binaries bin/*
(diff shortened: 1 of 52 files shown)


brikis98:
Tests passed. Merging now.

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.19.26

Issue #875, opened 2019-09-23
Honor the dynamodb_endpoint s3 parameter added in terraform v0.11


mattiarossi:
Hi,
terraform v0.11 added the possibility to specify a custom dynamodb endpoint for the dynamodb locking table.
I will be creating a PR to add support for the configuration parameter in terragrunt as well.
This allows to use either a custom endpoint, different from the standard aws one, or to deploy a local dynamodb instance for cases where acess to AWS infrastructure is not posible/allowed

PR #876, opened 2019-09-23
Dynamo db localendpoint

mattiarossi:
Terraform v0.11 added the possibility to specify a custom dynamodb endpoint for the dynamodb locking table.
This PR adds support for the configuration parameter in terragrunt as well.
This allows to use either a custom endpoint, different from the standard aws one, or to deploy a local dynamodb instance for cases where access to AWS infrastructure is not posible/allowed

I didn't create a specific unit test as no equivalent exists

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ jobs:
     - attach_workspace:
         at: *workspace_root
     - run: cd bin && sha256sum * > SHA256SUMS
-    - run: upload-github-release-assets
+    - run: upload-github-release-assets bin/*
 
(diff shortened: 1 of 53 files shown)


mattiarossi:
Fixes #875

brikis98:
@autero1 Could you review, kick off tests, merge, etc?

yorinasub17:
A version of this was implemented in https://github.com/gruntwork-io/terragrunt/pull/1586 and released as https://github.com/gruntwork-io/terragrunt/releases/tag/v0.28.15

PR #877, opened 2019-09-24
Return mock outputs when skip_outputs is set to `true`

barryib:
When doing `terraform validate` in CI pipeline, we disable the remote state initialization with :

```hcl
remote_state {
  # ...
  disable_init = tobool(get_env("TERRAGRUNT_DISABLE_INIT", "false"))
}
```

So we need to pass `dependency.mock_outputs` with `skip_outputs` set to `true`. But when setted to true, the dependency doesn't have output.

With this PR, we return mock outputs when `skip_outputs` is set to true. The `skip_outputs` should be used only to not pull outputs from remote state and not to disable mock outputs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,11 +5,11 @@
 
 # Terragrunt
 
-Terragrunt is a thin wrapper for [Terraform](https://www.terraform.io/) that provides extra tools for keeping your 
+Terragrunt is a thin wrapper for [Terraform](https://www.terraform.io/) that provides extra tools for keeping your
 Terraform configurations DRY, working with multiple Terraform modules, and managing remote state. Check out
-[Terragrunt: how to keep your Terraform code DRY and 
-maintainable](https://blog.gruntwork.io/terragrunt-how-to-keep-your-terraform-code-dry-and-maintainable-f61ae06959d8) 
-for a quick introduction to Terragrunt. 
+[Terragrunt: how to keep your Terraform code DRY and
+maintainable](https://blog.gruntwork.io/terragrunt-how-to-keep-your-terraform-code-dry-and-maintainable-f61ae06959d8)
+for a quick introduction to Terragrunt.
 
 
 
@@ -22,9 +22,9 @@ for a quick introduction to Terragrunt.
    downloading the binary for your OS, renaming it to `terragrunt`, and adding it to your PATH.
      * See the [Install Terragrunt](#install-terragrunt) docs for other installation options.
 
-1. Go into a folder with your Terraform configurations (`.tf` files) and create a `terragrunt.hcl` file that contains 
-   the configuration for Terragrunt (Terragrunt configuration uses the exact language, HCL, as Terraform). Here's an 
-   example of using Terragrunt to keep your Terraform backend configuration DRY (check out the [Use cases](#use-cases) 
+1. Go into a folder with your Terraform configurations (`.tf` files) and create a `terragrunt.hcl` file that contains
+   the configuration for Terragrunt (Terragrunt configuration uses the exact language, HCL, as Terraform). Here's an
+   example of using Terragrunt to keep your Terraform backend configuration DRY (check out the [Use cases](#use-cases)
    section for other types of configuration Terragrunt supports):
 
     ```hcl
@@ -37,7 +37,7 @@ for a quick introduction to Terragrunt.
         region         = "us-east-1"
         encrypt        = true
         dynamodb_table = "my-lock-table"
-      }   
+      }
     }
     ```
 
@@ -57,8 +57,8 @@ for a quick introduction to Terragrunt.
 
 1. Terragrunt is a direct implementation of the ideas expressed in
    [Terraform: Up & Running](http://www.terraformupandrunning.com). Additional background reading that will help
-   explain the motivation for Terragrunt include 
-   [Terragrunt: how to keep your Terraform code DRY and 
+   explain the motivation for Terragrunt include
+   [Terragrunt: how to keep your Terraform code DRY and
    maintainable](https://blog.gruntwork.io/terragrunt-how-to-keep-your-terraform-code-dry-and-maintainable-f61ae06959d8),
    [How to create reusable infrastructure with Terraform modules](https://blog.gruntwork.io/how-to-create-reusable-infrastructure-with-terraform-modules-25526d65f73d)
    and [How to use Terraform as a team](https://blog.gruntwork.io/how-to-use-terraform-as-a-team-251bc1104973).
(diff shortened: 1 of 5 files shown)


barryib:
@yorinasub17  @brikis98 any thoughts ?

brikis98:
@yorinasub17 Could you review?

barryib:
@yorinasub17 ready for review.

PS: I didn't run all tests.

README.md
```diff
@@ -1188,6 +1188,18 @@ dependency "vpc" {
 }
 ```
 
+When `skip_outputs` is used with `mock_outputs`, mocked outputs will be returned without pulling in the outputs from remote states. This can be usefull when you disable the backend initialization (`remote_state.disable_init`) in a CI by example.
```
yorinasub17:
```suggestion
When `skip_outputs` is used with `mock_outputs`, mocked outputs will be returned without pulling in the outputs from remote states. This can be useful when you disable the backend initialization (`remote_state.disable_init`) in CI for example.
```

config/dependency.go
```diff
@@ -1,34 +1,34 @@
...
+  "strings"
+
+  "github.com/hashicorp/hcl2/hcl"
+  "github.com/zclconf/go-cty/cty"
+  "github.com/zclconf/go-cty/cty/gocty"
+  ctyjson "github.com/zclconf/go-cty/cty/json"
+
+  "github.com/gruntwork-io/terragrunt/errors"
+  "github.com/gruntwork-io/terragrunt/options"
+  "github.com/gruntwork-io/terragrunt/util"
```
yorinasub17: Hmm these changes to the file tabs don't look correct?

test/integration_test.go
```diff
@@ -1644,6 +1644,53 @@ func TestDependencyOutputSkipOutputs(t *testing.T) {
...
+	stderr := bytes.Buffer{}
+	require.NoError(
+		t,
+		runTerragruntCommand(t, fmt.Sprintf("terragrunt output -no-color -json --terragrunt-non-interactive --terragrunt-working-dir %s", dependent3Path), &stdout, &stderr),
+	)
+	outputs := map[string]TerraformOutput{}
+	require.NoError(t, json.Unmarshal([]byte(stdout.String()), &outputs))
+	assert.Equal(t, outputs["truth"].Value, "The answer is 0")
+
+	// Now apply-all so that the dependency is applied, and verify it uses the dependency output
```
yorinasub17:
```suggestion
	// Now apply-all so that the dependency is applied, and verify it still uses the mock output
```

test/integration_test.go
```diff
@@ -1644,6 +1644,53 @@ func TestDependencyOutputSkipOutputs(t *testing.T) {
...
+	// verify expected output when mocks are used: The answer is 0
+	stdout = bytes.Buffer{}
+	stderr = bytes.Buffer{}
+	require.NoError(
+		t,
+		runTerragruntCommand(t, fmt.Sprintf("terragrunt output -no-color -json --terragrunt-non-interactive --terragrunt-working-dir %s", dependent3Path), &stdout, &stderr),
+	)
+	outputs = map[string]TerraformOutput{}
+	require.NoError(t, json.Unmarshal([]byte(stdout.String()), &outputs))
+	assert.Equal(t, outputs["truth"].Value, "The answer is 42")
```
yorinasub17: Hmm shouldn't this still be `0` because `skip_outputs` is `true`?

barryib: Right. Fixed with e4cd892

config/dependency.go
barryib: fixed.

yorinasub17:
Thanks for getting the updates in! Kicked off our build.

yorinasub17:
Looks like `TestDependencyOutput`, `TestDependencyOutputSkipOutputs`, and `TestDependencyOutputSkipOutputsWithMockOutput` failed in CI. Don't have time to dig in, but see if you can repro locally?

barryib:
For `TestDependencyOutputSkipOutputs` I think it's failing normally because of :

```
-- FAIL: TestDependencyOutputSkipOutputs (0.06s)
    integration_test.go:1992: Copying fixture-get-output to /var/folders/r8/smx4f8_x0hsfh4rqzv3d_w340000gn/T/terragrunt-test224902107
    integration_test.go:1641:
        	Error Trace:	integration_test.go:1641
        	Error:      	Received unexpected error:
        	            	/var/folders/r8/smx4f8_x0hsfh4rqzv3d_w340000gn/T/terragrunt-test224902107/fixture-get-output/integration/app1/terragrunt.hcl is a dependency of /var/folders/r8/smx4f8_x0hsfh4rqzv3d_w340000gn/T/terragrunt-test224902107/fixture-get-output/integration/empty/terragrunt.hcl but detected no outputs. Either the target module has not been applied yet, or the module has no outputs. If this is expected, set the skip_outputs flag to true on the dependency block.
        	Test:       	TestDependencyOutputSkipOutputs
```

By reading https://github.com/gruntwork-io/terragrunt/blob/master/config/dependency.go#L217-L220 
```
// This will attempt to get the outputs from the target terragrunt config if it is applied. If it is not applied, the
// behavior is different depending on the configuration of the dependency:
// - If the dependency block indicates a mock_outputs attribute, this will return that.
// - If the dependency block does NOT indicate a mock_outputs attribute, this will return an error.
```

It sounds like it normal to fail because, https://github.com/gruntwork-io/terragrunt/blob/master/test/fixture-get-output/integration/empty/terragrunt.hcl skip outputs, but doesn't have mock outputs.

yorinasub17:
> Is this worked before as expected ?

Yes this worked before as expected, and I would want to support this interface where you can use `skip_outputs` without setting mock outputs. This supports the following use cases:

- If we remove `dependencies` block in the future, you can replace them with `dependency` block with `skip_outputs = true` if you are still using `terraform_remote_state` to avoid the performance cost of looking up the outputs.
- If we enhance `dependency` to read the terragrunt config and load various configurations that you can reference (e.g using `dependency.vpc.inputs.aws_region` to read the inputs from another configuration for example), setting `mock_outputs` doesn't make sense because we have no intention of reading them.

The comment was consistent before because `skip_outputs` skipped calling that routine, which made sense because `skip_outputs` was intended to be used to mean "skip output pulling routing", not "return mock outputs".

So given all that, I think the logic needs to be updated so that when `skip_outputs` is true, but `mock_outputs` is not set, it should return empty object for outputs instead of erroring. Can you also update the comment of the routine to indicate this behavior? Thanks!

barryib:
@yorinasub17 Thanks for your explanation. The PR is ready for review.

yorinasub17:
Sorry for the delay in getting back. Kicked off another build!

yorinasub17:
Ok the build passed, so going to merge and release this. Thanks for your contribution!

PR #882, opened 2019-09-27
RFC: Terragrunt scaffolding

yorinasub17:
This is an RFC for an idea around Terragrunt scaffolding to improve the experience around common workflow tasks in Terragrunt like:

- adding a new component
- adding a new environment
- adding a new module
- adding a new account

Link to video [internal only] https://app.slack.com/client/T0PJEPZ2L/CM6RR6JE7


_docs/rfc/scaffolding_new_modules.md
```diff
@@ -0,0 +1,402 @@
...
+errors.
+
+
+## Proposed solution
+
+We will introduce a new command `terragrunt new`, which provides a code generation interface for solving each of the
+scenarios described above. The goal of `terragrunt new` is to provide the ability to templatize your configurations with
+annotations in the code that reminds the user what pieces need to change for each scenario.
+
+It is important that the templates are customizable. Not all users have the same requirements, so allowing users to
```
brikis98: +1

_docs/rfc/scaffolding_new_modules.md
```diff
@@ -0,0 +1,402 @@
...
+customize the templates to their needs is important. The goal is to allow users to define their own standards as
+templates that then is used by their engineers to generate new configurations that conform to their standards, which may
+be different from Gruntwork standards. I envision Gruntwork providing a set of base templates that can be used by
+default, which can be overridden by either forking our `infrastructure-templates` repository or by building new ones
+from scratch.
+
+`terragrunt new` should support the following scenarios:
```
brikis98: We may also want:

1. `terragrunt new basic-scaffold`, which creates an infra-live and infra-modules with a reasonable basic starting structure: e.g., a basic module in infra-modules and deploying that module in a couple envs in infra-live.
1. `terragrunt new module`: creates dummy `main.tf`, `variables.tf`, `outputs.tf`, `dependencies.tf`, `README.md`, and perhaps Terratest tests for this module.

_docs/rfc/scaffolding_new_modules.md
```diff
@@ -0,0 +1,402 @@
...
+      terminology here, as it does not mesh with GCP where you have Projects).
+
+- `terragrunt new account`, which will add a new account. The inputs for this template should be:
+    - Remote state info: the minimum information necessary to setup the remote state config.
+
+- `terragrunt new wrapper-module`, which will add a new wrapper module (what you have in `infrastructure-modules`). Note
+  that as such, this will be rendered into `infrastructure-modules`, as opposed to `infrastructure-live` like the other
+  sub commands. The inputs for this template should be:
+    - Module name: what to call this module.
+    - Library Module source: what the underlying library module you are basing the wrapper on.
```
brikis98: Oh, cool idea. If you can use this with modules in the Terraform Registry and Gruntwork IaC Lib, that would be very valuable.

brikis98: Consider generating a scaffold for Terratest tests too.

_docs/rfc/scaffolding_new_modules.md
```diff
@@ -0,0 +1,402 @@
...
+`terragrunt new` depends on a repository of templates. These templates can practically live anywhere, but for the
+purposes of versioning, it should live outside of `infrastructure-live` and `infrastructure-modules`. For the purposes
+of this example, we will assume all the templates live in a repository called `infrastructure-templates`.
+
+To support each of the scenario above, `terragrunt` will need to be configured to know which subtemplates should be used
+for each of the scenarios, in addition to knowing where the `infrastructure-live` and `infrastructure-modules`
+repositories live. This should be configured in a new configuration file for Terragrunt, which will be autogenerated
+from an interactive setup process the first time you run `terragrunt new`.
+
+The following configuration options will be required:
```
brikis98: Where does this config live? Is it a `terragrunt.hcl` somewhere on your hard drive? Or does Terragrunt prompt you for this info?

_docs/rfc/scaffolding_new_modules.md
```diff
@@ -0,0 +1,402 @@
...
+
+- `infrastructure_live_repo`: Git URL of the `infrastructure-live` repository.
+- `infrastructure_modules_repo`: Git URL of the `infrastructure-modules` repository.
+- `infrastructure_templates_repo`: Git URL of the `infrastructure-templates` repository. If you follow the convention,
```
brikis98: We could potentially support any "source" URL Terraform supports for this command

_docs/rfc/scaffolding_new_modules.md
```diff
@@ -0,0 +1,402 @@
...
+  this is the only necessary config to discover the templates for each of the scenarios listed above.
+- `new_environment_template_source`: Git URL OR local path to the template source for the `new environment` scenario.
```
brikis98: I was thinking it might be cleaner to have this work by convention. That is, your `infrastructure-templates` repo should have a folder called `scaffolding`, and for each folder `scaffolding/xxx`, you can run `terragrunt new xxx`. Therefore, the default `infrastructure-templates` repo would be at `github.com/gruntwork-io/infrastructure-templates` and it will have `scaffolding/environment`, `scaffolding/region`, `scaffolding/account`, and so on. If a user overrides `infrastructure-templates`, they can replace any of those templates with any names / scaffolds they choose.

_docs/rfc/scaffolding_new_modules.md
```diff
@@ -0,0 +1,402 @@
...
+`terragrunt new component` and it seeded the live configuration with all the different variables of the underlying
+module. Next, you will need to make the changes to set the values for each of the variables. This makes sense to edit
+directly on the generated code, so you shouldn't have to run the generator again. However, suppose that at a later point
+the module is updated and a new variable is introduced. Should you regenerate the code so you get the updated variable
+description?
+
+While "codegen drift" is a concern, I think the only practical usage of the proposed scaffolding is for bootstrapping an
+initial implementation, and not for maintaining the code through code generation. As such, you should only generate the
+code in the initial implementation, and then proceed to make manual edits to the generated code over time to maintain
+it, rather than going through `terragrunt new` each time to roll out an update.
```
brikis98: Agreed. Code generation is almost always a one-way, one-time process.

_docs/rfc/scaffolding_new_modules.md
```diff
@@ -0,0 +1,402 @@
...
+- Go templating is not necessary a good language to work with. You lose editor support, and some of the constructs are
+  not intuitive (E.g whitespace chomping).
+- Requires some work to embed into `terragrunt` so that arbitrary templates can be invoked.
+- We need to invent a DSL for handling and declaring template inputs.
+
+### Template tests
+
+Template tests can be invoked using the `shell` module in `terratest` that can be used to call `terragrunt new` using
+these templates with arbitrary input values. You can then apply the generated template to test the generated code
+against various variable inputs.
```
brikis98: Not sure I follow this?

_docs/rfc/scaffolding_new_modules.md
```diff
@@ -0,0 +1,402 @@
...
+
+Note that the templates folder is not restricted to just these 4 templates: you are free to define as many or as little
+templates as you need. Additional templates that do not fit the convention can be invoked using `terragrunt new
+unicorn`.
+
+### Template engine
+
+Templates are written and rendered using the default Go Template engine. While there are many other template engines to
+consider, given that the templates are rendered using Terragrunt as a frontend, it makes the most sense to use the built
+in templating engine of Go.
```
brikis98: Hm, this requires us to:

1. Define our own format for asking users for inputs.
1. Use Go templating syntax in folder/file paths too, as those will need to be templated.
1. Add a lot of helper functions to work around the many limitations of Go templating...

Are there really no better templating languages that have native Go support these days?

_docs/rfc/scaffolding_new_modules.md
```diff
@@ -0,0 +1,402 @@
...
+to introduce the new variable to a previously generated code. Specifically, when you introduce a new variable, and if
+you were keeping the list of variable inputs in the live code that was generated, you could end up in a situation of
+"codegen drift", where the module has a new variable that is omitted from your live config.
+
+Unfortunately, as mentioned in the answer to [When should code be generated?](#when-should-code-be-generated), it is not
+practical to always regenerate the code from the template every time the module changes. This means that we are once
+again back to relying on developer discipline for maintaining this link.
+
+The solution proposed here will not address this challenge. However, the hope is that the proposed solution will improve
+the UX over what we have currently, while acknowledging that it is not a perfect solution that solves all challenges.
```
brikis98: One potential thing we could do is add a `terragrunt new input-variable` command that adds the variable to both an existing modules in infra-modules and a `terragrunt.hcl` file (or files) in infra-live... But building scaffolding to generate code into existing files is a bit tricky.

_docs/rfc/scaffolding_new_modules.md
yorinasub17: See https://github.com/gruntwork-io/terragrunt/blob/da1e3da8eb16ff29e58449c40f3ad98809c2436f/_docs/rfc/scaffolding_new_modules.md#alternative-template-engines

_docs/rfc/scaffolding_new_modules.md
yorinasub17: To test the templates that you write, you use terratest to invoke the template by calling `terragrunt new` (using `shell`) and then run the terraform/terragrunt code.

_docs/rfc/scaffolding_new_modules.md
yorinasub17: Yes that is exactly what I was thinking. See:

> This can be used to override the location of the template, as an alternative to the convention followed in infrastructure-templates, or even source from a completely different repository.

at the end of the sentence. This interface is for situations where you don't want to follow convention, like if you wanted a repo per template.

_docs/rfc/scaffolding_new_modules.md
yorinasub17: Ah good point!

_docs/rfc/scaffolding_new_modules.md
yorinasub17: In the prototype, it lives in `~/.config/terragrunt/scaffold.toml`, and terragrunt prompts you for this info if the file doesn't exist.

We can also make it overridable via the command line.

yorinasub17:
Closing as stale

PR #888, opened 2019-10-02
Check ${TF_DATA_DIR} when looking up Terraform state files

### LINKED ISSUE #838: Terragrunt always reinitializes state if TF_DATA_DIR is specified
Terragrunt does not know that `terraform.tfstate` can be stored in a path specified in `TF_DATA_DIR` environment variable, and that causes it to call `terraform init` every time.

How to reproduce:

```
TF_DATA_DIR=somewhere terragrunt plan
TF_DATA_DIR=somewhere terragrunt plan
```

misha-tectonic:
This prevents incessant reinitialization on every invocation of Terragrunt if
TF_DATA_DIR is set

Closes #838

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -588,7 +588,7 @@ func needsInit(terragruntOptions *options.TerragruntOptions, terragruntConfig *c
 
 // Returns true if we need to run `terraform init` to download providers
 func providersNeedInit(terragruntOptions *options.TerragruntOptions) bool {
-	providersPath := util.JoinPath(terragruntOptions.WorkingDir, ".terraform/plugins")
+	providersPath := util.JoinPath(terragruntOptions.DataDir(), "plugins")
 	return !util.FileExists(providersPath)
 }
 
@@ -660,7 +660,7 @@ func runMultiModuleCommand(command string, terragruntOptions *options.Terragrunt
 // modules at all. Detecting if your downloaded modules are out of date (as opposed to missing entirely) is more
 // complicated and not something we handle at the moment.
 func modulesNeedInit(terragruntOptions *options.TerragruntOptions) (bool, error) {
-	modulesPath := util.JoinPath(terragruntOptions.WorkingDir, ".terraform/modules")
+	modulesPath := util.JoinPath(terragruntOptions.DataDir(), "modules")
 	if util.FileExists(modulesPath) {
 		return false, nil
 	}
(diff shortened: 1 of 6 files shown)


cli/cli_app.go
```diff
@@ -589,6 +590,9 @@ func needsInit(terragruntOptions *options.TerragruntOptions, terragruntConfig *c
 // Returns true if we need to run `terraform init` to download providers
 func providersNeedInit(terragruntOptions *options.TerragruntOptions) bool {
 	providersPath := util.JoinPath(terragruntOptions.WorkingDir, ".terraform/plugins")
+	if tfDataDir, ok := os.LookupEnv("TF_DATA_DIR"); ok {
+		providersPath = util.JoinPath(tfDataDir, "plugins")
```
brikis98: Should this be `plugins` or `.terraform/plugins`?

test/integration_test.go
```diff
@@ -1875,6 +1876,45 @@ func TestDependencyOutputCachePathBug(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func TestDataDir(t *testing.T) {
```
brikis98: Thanks for adding a test! 👍

cli/cli_app.go
```diff
@@ -661,6 +665,9 @@ func runMultiModuleCommand(command string, terragruntOptions *options.Terragrunt
 // complicated and not something we handle at the moment.
 func modulesNeedInit(terragruntOptions *options.TerragruntOptions) (bool, error) {
 	modulesPath := util.JoinPath(terragruntOptions.WorkingDir, ".terraform/modules")
+	if tfDataDir, ok := os.LookupEnv("TF_DATA_DIR"); ok {
+		modulesPath = util.JoinPath(tfDataDir, "modules")
```
brikis98: Seems like a lot of copy/pasted logic. Perhaps extract into a method called something like `withTfDataDir(pathParts ...string)`? If `TF_DATA_DIR` isn't set, it returns `util.JoinPath` on `pathParts`. If it is set, it joins the data dir to `pathParts`, dropping `.terraform` if it's the first arg to `pathParts`.

cli/cli_app.go
misha-tectonic: `plugins`.

cli/cli_app.go
misha-tectonic: Turns out every time `.terraform` is used, it comes from `TerragruntOptions`, so I have moved the logic there.

misha-tectonic:
Done

brikis98:

misha-tectonic:
Failing test seems to be unrelated to my change.

brikis98:
OK, re-ran tests and they passed, so looks like it was some transient issue with S3. Merging now, thanks!

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.20.1

misha-tectonic:
Thanks. Are binaries generated asynchronously, should I wait for them to appear on the releases page?

brikis98:
It takes the build ~10 min to run and upload the binaries. They should be there shortly.

Issue #887, opened 2019-10-02
Document variable precedence between dependency variable and common.tfvars variable


hehnope:
I searched the issues to see if this question was asked; did not see that it was.

If I have a variable dependency from a module, such as like below:

```
inputs = {
  region   = dependency.vpc.outputs.region
}
```

But, I also have a `common.tfvars` like below:

```
region      = "us-west-2"
```

Which variable takes precedence?

yorinasub17:
The variables from terragrunt inputs are set as environment variables (`TF_VAR_variable_name`), so they are the lowest precedence as mentioned in [the official tf docs](https://www.terraform.io/docs/configuration/variables.html#variable-definition-precedence).

But that is a good point that this isn't documented!

yorinasub17:
I updated the issue to reflect the lack of documentation and will keep this open until we resolve the missing docs.

hehnope:
@yorinasub17 can you assign it to me? I'll send a PR with updated readme.

yorinasub17:
Feel free to jump in and open the PR! Your comment above is enough to indicate ownership.

hehnope:
#889 was merged, so closing.

PR #889, opened 2019-10-02
add variable precedence

hehnope:
Adds documentation per #887 

Also, atom cleaned up a lot of extra white space.


hehnope:
white space changes undone.

Issue #868, opened 2019-09-18
Allow to ignore dependency order when using Terragrunt plan-all


JeanFred:
When using terragrunt `plan-all`, terragrunt will smartly run the plans according to the dependency tree, parallelizing as much as possible.
Additionally, when using `exclude-dir`, terragrunt will not run the downstream layers of the excluded layers.

This makes perfect sense for `apply-all` and `destroy-all`, however it should not really matter for `plan-all` since the state is not altered between the plans. 

This is a proposal to “Allow to ignore the dependency tree”, for these two purposes:
- to run all layers in parallel
- to run the downstream layers of excluded layers

Our use case is for testing on a CI server:
- running all layers in parallel would save a couple minutes.
- because reasons, we cannot plan some layers in CI (insufficient permissions), however we can plan other layers which depend on them.

(happy to split into two tickets if that’s better)

brikis98:
That's a great point. PR to change this behavior is very welcome!

JeanFred:
@brikis98 Thanks for the answer − I may try my hand at filing a PR for this :)

To clarify, what should the outcome be?
* the current default behaviour of `plan-all` is changed to ignore the tree
* a new flag/env var changes the behaviour of `plan-all` with both changes (all in parallel, still running dependencies of skipped layers)
* two different flag/env var changes the behaviour of `plan-all`

brikis98:
It's possible some users will need `plan-all` to run in dependency order: e.g., perhaps they have before/after hooks that do something important and won't work if it's all in parallel.

So, a `--terragrunt-ignore-dependency-tree` flag is probably the right approach. Thx!

JeanFred:
For clarity, I’ve split off the second part in #905. Closing this as resolved then!

PR #861, opened 2019-09-11
Added Cli Option to include external dependencies 

gcrespi:
on *-all commands without asking on the prompt.
This PR fixes #860

PR #891, opened 2019-10-03
Add option to ignore the dependency order during plan-all

JeanFred:
When running `plan-all`, we may want to run the modules while ignoring the dependency order: since the state is not altered between each `plan`, we might as well run all the plans at once.

This adds an option, `IgnoreDependencyOrder` (defaulting to False), which when set makes `plan-all` disregard the dependencies. 

To achieve that, this adds an `IgnoreOrder` run strategy and an associated `RunModulesIgnoreOrder` method.

Contributes to #868

### DIFF
--- a/README.md
+++ b/README.md
@@ -2292,6 +2292,8 @@ arguments that start with the prefix `--terragrunt-`. The currently available op
   under these directories (and all dependent modules) will be included during execution of the commands. If a relative
   path is specified, it should be relative from `--terragrunt-working-dir`. Flag can be specified multiple times.
 
+* `--terragrunt-ignore-dependency-order`: Ignore the depedencies between modules when running `*-all` commands.
+
 * `--terragrunt-ignore-external-dependencies`: Dont attempt to include any external dependencies when running `*-all`
   commands
 
--- a/configstack/running_module.go
+++ b/configstack/running_module.go
@@ -35,6 +35,7 @@ type DependencyOrder int
 const (
 	NormalOrder DependencyOrder = iota
 	ReverseOrder
+	IgnoreOrder
 )
 
 // Create a new RunningModule struct for the given module. This will initialize all fields to reasonable defaults,
@@ -73,6 +74,16 @@ func RunModulesReverseOrder(modules []*TerraformModule) error {
 	return runModules(runningModules)
 }
 
+// Run the given map of module path to runningModule. To "run" a module, execute the RunTerragrunt command in its
+// TerragruntOptions object. The modules will be executed without caring for inter-dependencies.
+func RunModulesIgnoreOrder(modules []*TerraformModule) error {
+	runningModules, err := toRunningModules(modules, IgnoreOrder)
+	if err != nil {
+		return err
+	}
+	return runModules(runningModules)
+}
+
 // Convert the list of modules to a map from module path to a runningModule struct. This struct contains information
 // about executing the module, such as whether it has finished running or not and any errors that happened. Note that
 // this does NOT actually run the module. For that, see the RunModules method.
@@ -95,6 +106,7 @@ func toRunningModules(modules []*TerraformModule, dependencyOrder DependencyOrde
 // * If dependencyOrder is NormalOrder, plug in all the modules M depends on into the Dependencies field and all the
 //   modules that depend on M into the NotifyWhenDone field.
 // * If dependencyOrder is ReverseOrder, do the reverse.
+// * If dependencyOrder is IgnoreOrder, do nothing.
 func crossLinkDependencies(modules map[string]*runningModule, dependencyOrder DependencyOrder) (map[string]*runningModule, error) {
 	for _, module := range modules {
 		for _, dependency := range module.Module.Dependencies {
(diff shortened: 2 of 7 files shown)


JeanFred:
I’m floating this as potential solution for at least the first half of #868.

- It felt natural to hook on the `DependencyOrder` structure, even in some respects it’s a bit weird: R120-R121 with the `Do Nothing` ^_^ 
- Thinking more − is `IgnoreOrder` a good name or would `NoOrder` be better?
- This includes a fair amount of unit tests, reflecting the ones for `ReverseOrder` that might be more than warranted :)
- I chose to only make this flag work on `plan-all` (which is the only reasonable use-case I can see), but other flags apply to all `-all` operations − am I trying too hard to protect the user from themselves?
- The CLI option part was heavily modeled after #861

Regarding the second part (run the downstream layers of excluded layers), I still need to dig deeper into this. One question already: should that be activated under the same CLI flag, or another one?

README.md
```diff
@@ -2244,6 +2244,8 @@ arguments that start with the prefix `--terragrunt-`. The currently available op
   under these directories (and all dependent modules) will be included during execution of the commands. If a relative 
   path is specified, it should be relative from `--terragrunt-working-dir`. Flag can be specified multiple times.
 
+* `--terragrunt-ignore-dependency-order`: Disregard the depedency between modules when using `plan-all`.
```
brikis98: I'd vote for this to be a general purpose flag that works with any `xxx-all` command. It's not a good idea to use it with other commands most of the time, but I'm not sure we should block it.

cli/args_test.go
```diff
@@ -163,7 +169,7 @@ func assertOptionsEqual(t *testing.T, expected options.TerragruntOptions, actual
 	assert.Equal(t, expected.IamRole, actual.IamRole, msgAndArgs...)
 }
 
-func mockOptions(t *testing.T, terragruntConfigPath string, workingDir string, terraformCliArgs []string, nonInteractive bool, terragruntSource string, ignoreDependencyErrors bool, includeExternalDependencies bool) *options.TerragruntOptions {
+func mockOptions(t *testing.T, terragruntConfigPath string, workingDir string, terraformCliArgs []string, nonInteractive bool, terragruntSource string, ignoreDependencyErrors bool, ignoreDependencyOrder bool, includeExternalDependencies bool) *options.TerragruntOptions {
```
brikis98: Should this method use this variable?

configstack/running_module.go
```diff
@@ -35,6 +35,7 @@ type DependencyOrder int
 const (
 	NormalOrder DependencyOrder = iota
 	ReverseOrder
+	IgnoreOrder
```
brikis98: This looks like a good name to me 👍

configstack/running_module_test.go
```diff
@@ -162,6 +162,45 @@ func TestToRunningModulesTwoModulesWithDependenciesReverseOrder(t *testing.T) {
 	testToRunningModules(t, modules, ReverseOrder, expected)
 }
 
+func TestToRunningModulesTwoModulesWithDependenciesIgnoreOrder(t *testing.T) {
```
brikis98: Nice, thanks for adding tests! 🍺

brikis98:
Also, could you rebase on the latest from master?

README.md
JeanFred: Done

configstack/running_module.go
JeanFred: Ok!

cli/args_test.go
JeanFred: Not too sure to be honest ^_^ that seemed like the correct way to get that option tested.

brikis98: Hm, right now, it's a bit deceiving, as anyone using the `mockOptions`, `mockOptionsWithIamRole` and `mockCmdOptions` methods would expect that boolean to do something, but it's actually not set on the `TerragruntOptions` struct, so it does nothing. Since it's not being used in any tests now (always set to `false`), I'd remove this new param from these methods (they have lots of params already!) and we can bring it back in the future if we have a use case that merits overriding it.

JeanFred: Hmmm, ok, but then how can I add a test that tests the `--terragrunt-ignore-dependency-errors` flag use? If I add 
```

brikis98: Probably the better bet would be an integration test in `integration_test.go` that sets the flag and makes sure things work as expected.

JeanFred: Hmm, had a look at `integration_test.go` and it’s a bit tricky… Would you say such an integration test is necessary, or it’s good enough without?
(as in, happy to look deep into it if it’s a blocker :) )

brikis98: Fairy enough, can prob live without it. I'll kick off tests now.

brikis98:
OK, tests passed. Merging now, thanks!

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.20.2

JeanFred:
Thanks for merging!

PR #897, opened 2019-10-07
get_aws_caller_identity_arn and get_aws_caller_identity_user_id functions

mcalhoun:
This PR:
* adds new `get_aws_caller_identity_arn` and `get_aws_caller_identity_user_id` functions that will return the ARN and UserId, respectively, of the currently configured user or IAM role similar to the existing `get_aws_account` function.
* updates the circle config to run unit and integration tests in parallel to save times on each CI run

### DIFF
--- a/README.md
+++ b/README.md
@@ -2036,8 +2036,8 @@ remote_state {
     bucket = "mycompany-${get_aws_account_id()}"
   }
 }
-
 ```
+
 #### get_aws_caller_identity
 
 `get_aws_caller_identity()` returns the ARN of the AWS identity associated with the current set of credentials. Example:
--- a/README.md
+++ b/README.md
@@ -1650,7 +1650,7 @@ currently available are:
 * [get_terraform_commands_that_need_locking()](#get_terraform_commands_that_need_locking)
 * [get_terraform_commands_that_need_parallelism()](#get_terraform_commands_that_need_parallelism)
 * [get_aws_account_id()](#get_aws_account_id)
-* [get_aws_caller_identity()](#get_aws_caller_identity)
+* [get_aws_caller_identity_arn()](#get_aws_caller_identity_arn)
 * [run_cmd()](#run_cmd)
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -2037,6 +2038,29 @@ remote_state {
...
+    actions = [
+      "kms:*"
+    ]
+
+    principals {
+      type        = "AWS"
+      identifiers = "${get_aws_caller_identity_arn()}"
+    }
+  }
+```
```
yorinasub17: This should probably be a terragrunt example, so maybe something like:
```
inputs = {
  caller_identity = get_aws_caller_identity_arn()
}

File: config/config_helpers.go [hunk 2]
```diff
@@ -314,14 +315,25 @@ func getAWSAccountID(include *IncludeConfig, terragruntOptions *options.Terragru
...
 
-	return *identity.Account, nil
+	return *identity, nil
 }
 
-// Custom error types
+// Return the AWS account id associated to the current set of credentials
+func getAWSAccountID(include *IncludeConfig, terragruntOptions *options.TerragruntOptions) (string, error) {
+	identity, err := getAWSCallerID(include, terragruntOptions)
+	return *identity.Account, err
```
yorinasub17: I know this is a copy from the old code, but should this actually be updated to `aws.StringValue(identity.Account)` (import is `"github.com/aws/aws-sdk-go/aws"`)?

mcalhoun:
In doing my own code review, I realized my example doesn't make any sense...I'll update that

config/config_helpers.go
mcalhoun: It looks like in the newest sdk it's `"github.com/aws/aws-sdk-go/aws/awsutil"`

mcalhoun: be30c58d

yorinasub17: Ah good to know!

PR #898, opened 2019-10-08
Change order of parsing so that include is parsed first

yorinasub17:
This is a fix for https://github.com/gruntwork-io/terragrunt/issues/814#issuecomment-518919484. Specifically, right now due to the parsing order `get_parent_terragrunt_dir` does not work in `locals`. This breaks usage of locals to refactor out a path calculation that depends on the parent dir, such as getting a default `empty.yml` file in the workaround for merging in optional variables from the directory hierarchy.

Note that this is a **backwards incompatible** change, where you can no longer reference `locals` in `include`. However, this is probably a worthy tradeoff where we are trading the not so useful function of referencing `locals` in `include` blocks (which will in 80% of use cases be `find_in_parent_folders()`) with the ability to use `include` dependent functions in the parent `locals`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2380,8 +2380,8 @@ from `dependency` into `locals`. On the other hand, for the same reason, you can
 
 Currently terragrunt parses the config in the following order:
 
-1. `locals` block
 1. `include` block
+1. `locals` block
 1. `dependencies` block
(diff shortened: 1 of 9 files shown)


yorinasub17:
Merging this in. Thanks for review!

Issue #851, opened 2019-09-04
Terragrunt CLI options in Terragrunt config


erezzarum:
Suppose i want to keep the download directory (.terragrunt-cache) in a different path than ".terragrunt-cache", currently the only solution is to provider either a "--terragrunt-download-dir=<PATH>" argument or a "TERRAGRUNT_DOWNLOAD=<PATH>" environment variable.
I couldn't find an option, but it would be greate to define this in a Terragrunt config file (terragrunt.hcl).
I tried this, but it does not work (it passes the "--terragrunt-download-dir" to Terraform)
```
  extra_arguments "terragrunt_cache" {
    commands = [
      "plan",
      "apply",
    ]
    arguments = [
      "--terragrunt-download-dir=${get_terragrunt_dir()}/${path_relative_from_include()}/.terragrunt-cache/${path_relative_to_include()}"
    ]
  }
```

The only not full workaround i could come up with is to create a shell wrapper script around the terraform binary and pass it to Terragrunt as the terraform binary path.

It would be really cool if we could provide terragrunt arguments from the terragrunt config itself.

brikis98:
PR welcome!

erezzarum:
If my skills in Go were as good as my Python skills, i wouldn't even post this issue and send a PR request.
hopefully someone talented than me could help with this :)

ekini:
I can do it as part of the Hacktoberfest challenge :)

ekini:
I've added `download_dir` to the config. @erezzarum did you want to see all of the CLI options in the config or just `download_dir` would be sufficient?

I named the config option `download_dir` as I think `terragrunt` namespacing is not necessary there.

@brikis98 There is some inconsistency in naming.

The CLI option is called `terragrunt-download-dir`, but the env var is called just `TERRAGRUNT_DOWNLOAD`.

ankitcharolia:
I face the same issue. Is there any update ?

PR #901, opened 2019-10-09
Adds download_dir to the config

ekini:
Related to #851

### DIFF
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -107,6 +107,87 @@ func init() {
 	rand.Seed(time.Now().UnixNano())
 }
 
+func TestTerragruntDownloadDir(t *testing.T) {
+	t.Parallel()
+
+	cleanupTerraformFolder(t, TEST_FIXTURE_LOCAL_RELATIVE_DOWNLOAD_PATH)
+	tmpEnvPath := copyEnvironment(t, TEST_FIXTURE_GET_OUTPUT)
+
+	/* we have 2 terragrunt dirs here. One of them doesn't set the download_dir in the config,
+	the other one does. Here we'll be checking for precedence, and if the download_dir is set
+	according to the specified settings
+	*/
+	testCases := []struct {
+		name                 string
+		rootPath             string
+		downloadDirEnv       string // download dir set as an env var
+		downloadDirFlag      string // download dir set as a flag
+		downloadDirReference string // the expected result
+	}{
+		{
+			"download dir not set",
+			util.JoinPath(tmpEnvPath, TEST_FIXTURE_GET_OUTPUT, "download-dir", "not-set"),
+			"", // env
+			"", // flag
+			util.JoinPath(tmpEnvPath, TEST_FIXTURE_GET_OUTPUT, "download-dir", "not-set", TERRAGRUNT_CACHE),
+		},
+		{
+			"download dir set in config",
+			util.JoinPath(tmpEnvPath, TEST_FIXTURE_GET_OUTPUT, "download-dir", "in-config"),
+			"", // env
+			"", // flag
+			util.JoinPath(tmpEnvPath, TEST_FIXTURE_GET_OUTPUT, "download-dir", "in-config", ".download"),
+		},
+		{
+			"download dir set in config and in env var",
+			util.JoinPath(tmpEnvPath, TEST_FIXTURE_GET_OUTPUT, "download-dir", "in-config"),
+			util.JoinPath(tmpEnvPath, TEST_FIXTURE_GET_OUTPUT, "download-dir", "in-config", ".env-var"),
+			"", // flag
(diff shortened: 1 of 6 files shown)


ekini:
will do!

ekini:
Just a head up I used the same method of determining the default download directory https://github.com/gruntwork-io/terragrunt/pull/901/files#diff-56be1e853ecf6e7ad64321692db32678R280 as in #866 which itself is somehow related to #904, so maybe it makes sense to understand what's going on there first.

The bad thing is I can't reproduce that issue.

yorinasub17:
I've been able to reproduce the issue, and is fixed in master. See https://github.com/gruntwork-io/terragrunt/pull/937. The crux of the issue was https://github.com/gruntwork-io/terragrunt/issues/906#issuecomment-545072547.

ekini:
Ok, I've added the tests!

test/integration_test.go
```diff
@@ -107,6 +107,87 @@ func init() {
...
+			stdout := bytes.Buffer{}
+			stderr := bytes.Buffer{}
+			err := runTerragruntCommand(t, fmt.Sprintf("terragrunt terragrunt-info %s --terragrunt-non-interactive --terragrunt-working-dir %s", testCase.downloadDirFlag, testCase.rootPath), &stdout, &stderr)
+			logBufferContentsLineByLine(t, stdout, "stdout")
+			logBufferContentsLineByLine(t, stderr, "stderr")
+			require.NoError(t, err)
+
+			var dat cli.TerragruntInfoGroup
+			err_unmarshal := json.Unmarshal(stdout.Bytes(), &dat)
+			assert.Nil(t, err_unmarshal)
```
yorinasub17:
```suggestion
			assert.NoError(t, err_unmarshal)
```

This has a better error message from the library when there is an error

test/integration_test.go
```diff
@@ -107,6 +107,87 @@ func init() {
 	rand.Seed(time.Now().UnixNano())
 }
 
+func TestTerragruntDownloadDir(t *testing.T) {
+	t.Parallel()
```
yorinasub17: Unfortunately, we can't run this test in parallel because it manipulates environment variables. As in, if this sets an environment variable, it can conflict with other tests running concurrently because the env is set on the test process.

test/integration_test.go
```diff
@@ -107,6 +107,87 @@ func init() {
...
+	for _, testCase := range testCases {
+		testCase := testCase
+
+		t.Run(testCase.name, func(t *testing.T) {
+			if testCase.downloadDirEnv != "" {
+				require.NoError(t, os.Setenv("TERRAGRUNT_DOWNLOAD", testCase.downloadDirEnv))
+			} else {
+				// clear the variable if it's not set
+				require.NoError(t, os.Unsetenv("TERRAGRUNT_DOWNLOAD"))
+			}
```
yorinasub17: I think we can `defer` the unset when you set it above for cleaner test isolation. At least, it will be easier to follow that way. As in, this if block should be:

```
if testCase.downloadDirEnv != "" {
  defer os.Unsetenv("TERRAGRUNT_DOWNLOAD")
  require.NoError(t, os.Setenv("TERRAGRUNT_DOWNLOAD", testCase.downloadDirEnv))
}
```

ekini: Though, it needs to be unset for some subtests, so we can't defer it.

ekini: Ah, you're right! I didn't look at the code. It's enclosed in a function.

ekini: Ok, however, it might have to unset the env var beforehand if it's been set outside of the tests.

For example, I have my `TERRAGRUNT_DOWNLOAD` permanently set to `/tmp/terragrunt` in my `.zshrc` so that I don't have to worry about cleaning old files, and it breaks the tests.

test/integration_test.go
```diff
@@ -107,6 +107,85 @@ func init() {
 	rand.Seed(time.Now().UnixNano())
 }
 
+func TestTerragruntDownloadDir(t *testing.T) {
```
yorinasub17: I forgot to mention this in the previous comment about removing `t.Parallel`, but can you add a comment here explaining why we don't run this in parallel? Specifically that it is manipulating environment variables as part of the test which edits global state?

test/integration_test.go
yorinasub17: I think it is still good practice to add the defer in the `if` case to ensure the env var doesn't leak. Also, can you update the comment for the `else` comment:

```
// Clear the variable if it is set. This is clearing the variable in case the variable is set outside the test process.
```

ekini: Added comments and the defer :)

test/integration_test.go
ekini: done 👍

yorinasub17:
Thanks for making the changes! Just kicked off the build and will merge if it passes!

yorinasub17:
Merging and releasing this. Thanks for your contribution!

ekini:
Thanks for the feedback!

Issue #905, opened 2019-10-10
Allow to run downstream layers of excluded layers when using Terragrunt *-all


JeanFred:
When using `exclude-dir`, terragrunt will not run the downstream layers of the excluded layers.

This makes perfect sense for apply-all and destroy-all, however it should not really matter for plan-all since the state is not altered between the plans.

Further to #868, this is a proposal to allow to run the downstream layers of excluded layers anyways.

Our use case is for testing on a CI server:
- because reasons, we cannot plan some layers in CI (insufficient permissions), however we can plan other layers which depend on them.

JeanFred:
@brikis98 I would have some questions.

- Do we want this behaviour to be triggered via the flag introduced in #891, `terragrunt-ignore-dependency-order`? Or another flag?
- I can see two ways of implementing this:
  - in `running_modules.go`, just conditionally avoid the call to `removeFlagExcluded` in `toRunningModules`.
  - avoid flagging the dependant modules in the first place in `module.go`

I lean towards the first option, but checking before starting :)

brikis98:
Just to make sure I understand, you're running `plan-all --exclude-dir foo`... And within folder `foo`, there are several modules, some of which you want to exclude, and some which you don't?

JeanFred:
> Just to make sure I understand, you're running `plan-all --exclude-dir foo`... And within folder `foo`, there are several modules, some of which you want to exclude, and some which you don't?

Not exactly, our infra looks something like:
```
prod
├── app
│   ├── secrets.tfvars.enc
│   └── terraform.tfvars
├── data-store
│   ├── secrets.tfvars.enc
│   └── terraform.tfvars
├── networking
│   ├── secrets.tfvars.enc
│   └── terraform.tfvars
└── terraform.tfvars
```

`app` depends on `data-store` which depends on `networking` (via ` dependencies {    paths = ["../networking"]  }`).

Because of permission reasons (that we hope to solve at some point, but cannot yet), our CI system cannot plan, say, the `data-store` layer (it needs AWS permission that we don’t want to grant ; or it relies on a submodule that we don’t want the CI system to be able to clone, etc.).

I thus want to plan all layers except `data-store`.

Currently, `plan-all --exclude-dir data-store` will exclude `data-store` and `app`, since `app` depends on `data-store`.

The new behaviour would be to run the plan against `app` anyways.

brikis98:
Thx for the explanation!

[...]

PR #912, opened 2019-10-14
Add unit tests for `removeFlagExcluded`

JeanFred:
When using the `--terragrunt-exclude-dir`, modules are flagged
with a `FlagExcluded` boolean. Later on, when Terragrunt converts
the modules into runningModules (in the `toRunningModules` method),
it calls as last step the method `removeFlagExcluded`.

That method returns a modified map of `runningModules` map,
with the ones marked with the boolean removed and modules
which depend on them have their `Dependencies` fixed accordingly.

This adds unit tests for three use cases for `removeFlagExcluded`:
 - No module is marked with `FlagExcluded`:
   nothing happens, the output is the same as the input
 - One module, with no dependencies, is marked with `FlagExcluded`:
   that module is removed from the map
 - One module, with dependencies, is marked with `FlagExcluded`:
   that module is removed from the map and its direct depedency
   is altered to not depend on the excluded module anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2286,7 +2286,8 @@ arguments that start with the prefix `--terragrunt-`. The currently available op
 
 * `--terragrunt-exclude-dir`: Unix-style glob of directories to exclude when running `*-all` commands. Modules under
   these directories will be excluded during execution of the commands. If a relative path is specified, it should be
-  relative from `--terragrunt-working-dir`. Flag can be specified multiple times.
+  relative from `--terragrunt-working-dir`. Flag can be specified multiple times. This will only exclude the module,
+  not its dependencies.
 
 * `--terragrunt-include-dir`: Unix-style glob of directories to include when running `*-all` commands. Only modules
   under these directories (and all dependent modules) will be included during execution of the commands. If a relative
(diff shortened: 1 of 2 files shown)


brikis98:
Is the context for this the questions you're trying to sort out in https://github.com/gruntwork-io/terragrunt/issues/905#issuecomment-541785592?

JeanFred:
> Is the context for this the questions you're trying to sort out in [#905 (comment)](https://github.com/gruntwork-io/terragrunt/issues/905#issuecomment-541785592)?

Not really − these tests disproved how I understood Terragrunt to work ; and I thought, might as well propose them to document the existing behaviour. :)

brikis98:
Ah, gotcha. Is it worth updating the `README.md` to clarify how `--terragrunt-exclude-dir` works too?

JeanFred:
> Ah, gotcha. Is it worth updating the `README.md` to clarify how `--terragrunt-exclude-dir` works too?

Good idea − added a line to the `README.md` :)

brikis98:
Tests passed. Merging now. Since these are docs/test updates only, there won't be a new release. Thank you!

PR #917, opened 2019-10-17
Only perform exact match when [in|ex]cluding directories

JeanFred:
The flags `--terragrunt-include-dir` and `--terragrunt-exclude-dir` are
“Unix-style glob of directories to include (resp. exclude) when running
*-all commands”

Currently, the matching on which directories to include/exclude is done
using `string.Contains`, which means that this excluding `module-a` would
also exclude `module-abba`.

We should really be doing an exact match, since we can get the multi-match
behaviour by using Unix-style globs.

Also add two unit-tests confirming the old and new behaviours.

Ref #905

### DIFF
--- a/configstack/module.go
+++ b/configstack/module.go
@@ -198,7 +198,7 @@ func flagIncludedDirs(modules []*TerraformModule, terragruntOptions *options.Ter
 // Returns true if a module is located under one of the target directories
 func findModuleinPath(module *TerraformModule, targetDirs []string) bool {
 	for _, targetDir := range targetDirs {
-		if strings.Contains(module.Path, targetDir) {
+		if module.Path == targetDir {
 			return true
 		}
 	}
--- a/configstack/module_test.go
+++ b/configstack/module_test.go
@@ -135,6 +135,109 @@ func TestResolveTerraformModulesTwoModulesWithDependenciesExcludedDirsWithDepend
 	assertModuleListsEqual(t, expected, actualModules)
 }
 
+func TestResolveTerraformModulesTwoModulesWithDependenciesExcludedDirsWithDependencyAndConflictingNaming(t *testing.T) {
+	t.Parallel()
+
+	opts, _ := options.NewTerragruntOptionsForTest("running_module_test")
+	opts.ExcludeDirs = []string{canonical(t, "../test/fixture-modules/module-a")}
+
+	moduleA := &TerraformModule{
+		Path:         canonical(t, "../test/fixture-modules/module-a"),
+		Dependencies: []*TerraformModule{},
+		Config: config.TerragruntConfig{
+			Terraform: &config.TerraformConfig{Source: ptr("test")},
+			IsPartial: true,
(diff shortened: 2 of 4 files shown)


brikis98:
@autero1 Could you review this one?

autero1:
Thanks for the PR @JeanFred !

> since we can get the multi-match behaviour by using Unix-style globs.

That is correct. We can filter out all subdirectories of `module-a` recursively with `module-a/**/*` exclude pattern without having to do `strings.Contains`. 

The PR looks good - I'll kick off tests now.

configstack/module.go
```diff
@@ -198,7 +198,7 @@ func flagIncludedDirs(modules []*TerraformModule, terragruntOptions *options.Ter
 // Returns true if a module is located under one of the target directories
 func findModuleinPath(module *TerraformModule, targetDirs []string) bool {
 	for _, targetDir := range targetDirs {
-		if strings.Contains(module.Path, targetDir) {
+		if (module.Path == targetDir) {
```
autero1: Minor style nit: You don't need the parentheses in the if statement.

autero1:
The integration test `TestExcludeDirs` failed. This is the set of patterns that fixed the test for me.
```
		{TEST_FIXTURE_LOCAL_WITH_EXCLUDE_DIR, "--terragrunt-exclude-dir **/gce/**/*", []string{"Module GCE B", "Module GCE C", "Module GCE E"}},
		{TEST_FIXTURE_LOCAL_WITH_EXCLUDE_DIR, "--terragrunt-exclude-dir production-env/**/* --terragrunt-exclude-dir **/module-gce-c", []string{"Module GCE C", "Module AWS D", "Module GCE E"}},
		{TEST_FIXTURE_LOCAL_WITH_EXCLUDE_DIR, "--terragrunt-exclude-dir integration-env/gce/module-gce-b --terragrunt-exclude-dir integration-env/gce/module-gce-c --terragrunt-exclude-dir **/module-aws*", []string{"Module AWS A", "Module GCE B", "Module GCE C", "Module AWS D"}},
```
This PR will result in a backwards incompatible release because of the change in how the `glob` patterns are applied. I do agree that now the pattern behaves more like a true `glob` pattern and as such will be more precise. But it will lead to breaking existing behaviour.

configstack/module.go
JeanFred: Fixed.

JeanFred:
Thanks for the review, the tests, and the fix for the tests! I have updated the PR accordingly with the new test fixtures.

autero1:
Re-running the tests.

autero1:
Tests passed, so merging this in!

autero1:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.21.0

msvechla:
Hi, it looks like this PR broke the `--terragrunt-include-dir` flag, see: https://github.com/gruntwork-io/terragrunt/issues/977

PR #926, opened 2019-10-24
Fix link in README

robbytaylor:
Removes a hyphen to a `Formatting terragrunt.hcl` link which breaks the link


autero1:
Thanks for the PR @robbytaylor - merging this in now!

Issue #904, opened 2019-10-09
terragrunt plan-all is flakey from 0.19.25


dvulpe:
Hi 

We recently noticed that terragrunt plan-all is flakey and would randomly fail with files not found errors, sometimes reporting that a terraform backend hasn't been configured, despite being configured correctly:
```
.terragrunt-cache/yZpa9oXGPFA9yBeYJLKvMDWHQLA/nOHeCNZgRObX4kF5VK1jboCrx_w/.terragrunt-module-manifest: no such file or directory
```

I wrote a test to reproduce the problem and if the subtest runs 100 times I get to reproduce the problem at least twice (this can be run in `integration_test.go` on the master branch):

```
func TestDependencyPlanAllFlake(t *testing.T) {

	cleanupTerraformFolder(t, TEST_FIXTURE_GET_OUTPUT)
	tmpEnvPath := copyEnvironment(t, TEST_FIXTURE_GET_OUTPUT)
	rootPath := util.JoinPath(tmpEnvPath, TEST_FIXTURE_GET_OUTPUT, "localstate", "live")

	stdout := bytes.Buffer{}
	stderr := bytes.Buffer{}
	err := runTerragruntCommand(
		t,
		fmt.Sprintf("terragrunt apply-all --terragrunt-non-interactive --terragrunt-working-dir %s", rootPath),
		&stdout,
		&stderr,
	)
	logBufferContentsLineByLine(t, stdout, "stdout")
	logBufferContentsLineByLine(t, stderr, "stderr")
	require.NoError(t, err)
	for i :=0; i<100; i++ {
		t.Run(fmt.Sprintf("take %d", i), func(t *testing.T) {
			stdout := bytes.Buffer{}
			stderr := bytes.Buffer{}
			err := runTerragruntCommand(
				t,
				fmt.Sprintf("terragrunt plan-all --terragrunt-non-interactive --terragrunt-working-dir %s", rootPath),
				&stdout,
				&stderr,
			)
			logBufferContentsLineByLine(t, stdout, "stdout")
			logBufferContentsLineByLine(t, stderr, "stderr")
			require.NoError(t, err)
		})
	}
}
```

The test output for the failed cases looks like:
```
            	            	Encountered the following errors:
[...]

PR #930, opened 2019-10-29
Revert change introduced in #866

yorinasub17:
This reverts #866 in favor of resolving the issues reported in https://github.com/gruntwork-io/terragrunt/issues/906 and https://github.com/gruntwork-io/terragrunt/issues/904. These two issues are caused by concurrent calls to terragrunt to update the source OR run `init` in the same working folder, caused by concurrent calls to `terragrunt output` due to dependency ordering.

The proper fix that will allow us to reimplement this are:

- Introduce locking mechanism to prevent concurrent calls to `terragrunt output` in the same module.
- Introduce a graph logic that will extract the outputs, once, prior to calling the steps in a `xxx-all`.

Both are non-trivial solutions that introduce a lot of complexity. I have decided that it will take enough time to implement that it seems better to revert the fix as suggested in https://github.com/gruntwork-io/terragrunt/issues/906#issuecomment-547482455.

### DIFF
--- a/config/dependency.go
+++ b/config/dependency.go
@@ -324,6 +324,7 @@ func runTerragruntOutputJson(terragruntOptions *options.TerragruntOptions, targe
 	// `cloneTerragruntOptionsForDependencyOutput`. Specifically, if you have multiple terragrunt sources that point to
 	// the same dependency, then terragrunt could attempt to download the resources multiple times causing conflicts.
 	// See https://github.com/gruntwork-io/terragrunt/issues/906 for more info.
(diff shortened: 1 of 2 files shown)


yorinasub17:
Hmm now that you mention it, yes we can probably resolve this right there in the output call. Originally I was thinking we would need some fixes in the `xxx-all` calls, but it is probably good enough to mutex the output call. Let me try that.

yorinasub17:
Closing in favor of https://github.com/gruntwork-io/terragrunt/pull/937. We can revive this if that does not resolve the issue.

Issue #924, opened 2019-10-22
What is the best way to manage access to the Terraform remote backend (like S3) ?


barryib:
Actually Terragrunt can create remote backend on S3 and GCP with audit, encryption activated. But how can we manage access to that backend automatically ?

Does Terragrunt create a bucket policy ? If not, what is the best way to manage this ?

yorinasub17:
Terragrunt currently doesn't do anything related to managing permissions on the bucket. This is a bit of an advanced configuration for terragrunt to handle as a part of automatically creating buckets. For something like this, it is probably better off managed in terraform.

See https://github.com/gruntwork-io/terragrunt/issues/859#issuecomment-530321902 for a workflow to accomplish this.

barryib:
I mostly use Terragrunt for its DRY implementation and its remote backend creation. The latest eliminate a lot of complexity in the bucket initialisation workflow. I would love to continue using that feature.

Actually Terragrunt try to enforce a sort of best practices for remote backend configuration like encryption, versioning, audit, block public acess, etc. But there is something missing. In fact, in remote states, you can have secrets and passwords, so we need something to deny/allow access from unauthorized/authorized roles.

I understand your point. But is applying bucket policy add a lot of complexity ? The idea here is to provide bucket policy (in json for exemple) to Terragrunt and let him apply it. The AWS SDK provide already method for that.

I can quickly open a PR for this.

@brikis98 @yorinasub17 are you open to discussion or this is a NO GO for you ?

yorinasub17:
The challenge is encoding this in a way in `terragrunt.hcl` such that it is not too confusing. For example, right now we have mixed terragrunt specific configurations of the S3 bucket with terraform config for the remote state backend. This leads to some confusions about which configuration maps to what.

Also, if we trend towards this direction, we are basically committing to implementing the full S3 bucket API for managing it, which duplicates logic available with terraform.

Thinking out loud, the ideal solution I would be looking for for a feature like this would be to switch the terragrunt bucket creation routine to support the terraform API. It would be really nice if you could write out the terraform resource block for S3 buckets and DynamoDB tables somewhere (possibly embedded as a subblock in `remote_state`?), and then terragrunt uses that (with the AWS provider in the backend) to create the state backend resources. It would be even better if we could store the state somewhere so that you could destroy the resources in the end.

yorinasub17:
[...]

PR #934, opened 2019-10-30
[RFC] Add Terraform Lock Table Config for S3 Remote

ThatGerber:
# Allow further customization of Terraform Lock table for S3 Remote State

**STATUS**: In proposal

## Background

When Terragrunt creates a DynamoDB table, it sets the throughput to the default,
`PROVISIONED`. Provisioned is good for stable throughput, but Terraform largely
works with variable workloads, often going unused.

AWS also offers a `PAY_PER_REQUEST` billing mode that charges based on the
number of requests to the lock table. The pricing is dirt cheap, and it will
further reduce the costs associated with using Terragrunt to manage Terraform's
base requirements.

Terraform has solved this problem: It involves creating and managing the
resource manually. But, if Terragrunt is going to be managing the table, it
should allow for additional configuration to be layered on top of it via
Terragrunt.

### `dynamodb_table_config` block in `remote_state.config`

Because our company has a need for it, I've implemented a solution in my own
fork that accomplishes this.

https://github.com/ThatGerber/terragrunt/tree/add-dynamodb-billing-mode

The `dynamodb_table_config` block accepts several arguments that are forwarded
through to the DynamoTableInput object, pulling in additional values from the
parent config as defaults.

To receive the full values, I had to write a few interfaces for the methods
being called (like `GetLockTableName`) and change the function calls to accept
the interfaces instead of strings.

### `dynamodb_table_config` block that populates `dynamodb.CreateTableInput` struct

Another option is to create a config attribute that accepts any arguments that
would be sent to `CreateTable` or `UpdateTable` map the values through to the
methods when they're being applied.

## References

- [DynamoDB Pricing](https://aws.amazon.com/dynamodb/pricing/)
- [DynamoDB Billing Mods](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BillingModeSummary.html)

### DIFF
--- a/_docs/rfc/dynamo_table_config.md
+++ b/_docs/rfc/dynamo_table_config.md
@@ -1,0 +1,47 @@
+# Allow further customization of Terraform Lock table for S3 Remote State
+
+**STATUS**: In proposal
+
+## Background
+
+When Terragrunt creates a DynamoDB table, it sets the throughput to the default,
+`PROVISIONED`. Provisioned is good for stable throughput, but Terraform largely
+works with variable workloads, often going unused.
+
+AWS also offers a `PAY_PER_REQUEST` billing mode that charges based on the
+number of requests to the lock table. The pricing is dirt cheap, and it will
+further reduce the costs associated with using Terragrunt to manage Terraform's
+base requirements.
+
+Terraform has solved this problem: It involves creating and managing the
+resource manually. But, if Terragrunt is going to be managing the table, it
+should allow for additional configuration to be layered on top of it via
+Terragrunt.
+
+## Proposed solution
+
+### `dynamodb_table_config` block in `remote_state.config`
+
+Because our company has a need for it, I've implemented a solution in my own
+fork that accomplishes this.
+
+https://github.com/ThatGerber/terragrunt/tree/add-dynamodb-billing-mode
+
+The `dynamodb_table_config` block accepts several arguments that are forwarded
+through to the DynamoTableInput object, pulling in additional values from the
+parent config as defaults.
+
+To receive the full values, I had to write a few interfaces for the methods
+being called (like `GetLockTableName`) and change the function calls to accept
+the interfaces instead of strings.
+
+### `dynamodb_table_config` block that populates `dynamodb.CreateTableInput` struct
+
+Another option is to create a config attribute that accepts any arguments that
+would be sent to `CreateTable` or `UpdateTable` map the values through to the
+methods when they're being applied.
+
+## References
+
+- [DynamoDB Pricing](https://aws.amazon.com/dynamodb/pricing/)
+- [DynamoDB Billing Mods](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BillingModeSummary.html)


_docs/rfc/dynamo_table_config.md
```diff
@@ -0,0 +1,47 @@
...
+
+AWS also offers a `PAY_PER_REQUEST` billing mode that charges based on the
+number of requests to the lock table. The pricing is dirt cheap, and it will
+further reduce the costs associated with using Terragrunt to manage Terraform's
+base requirements.
+
+Terraform has solved this problem: It involves creating and managing the
+resource manually. But, if Terragrunt is going to be managing the table, it
+should allow for additional configuration to be layered on top of it via
+Terragrunt.
```
yorinasub17: Makes a lot of sense! This actually has a lot of parallels with the challenge of setting up the S3 bucket (see https://github.com/gruntwork-io/terragrunt/issues/924).

_docs/rfc/dynamo_table_config.md
```diff
@@ -0,0 +1,47 @@
...
+
+To receive the full values, I had to write a few interfaces for the methods
+being called (like `GetLockTableName`) and change the function calls to accept
+the interfaces instead of strings.
+
+### `dynamodb_table_config` block that populates `dynamodb.CreateTableInput` struct
+
+Another option is to create a config attribute that accepts any arguments that
+would be sent to `CreateTable` or `UpdateTable` map the values through to the
+methods when they're being applied.
```
yorinasub17: While this works, the drawbacks of this approach is that it commits us to keeping this remote block up to date with the AWS API, which means that we will always be "chasing" updates in the SDK. This duplicates logic available with other tools.

This also sets a precedent that opens us up for doing the same with a bunch of other resources (e.g S3, GCS buckets) and I am not sure we want to commit to maintaining an API like that.

The ideal solution I would be looking for for a feature like this would be to switch the terragrunt routines to use terraform under the hood. It would be really nice if you could write out the terraform resource block for S3 buckets and DynamoDB tables somewhere (this could be what you have for `dynamodb_table_config`), and then terragrunt directly forwards that to the terraform AWS provider to manage the state backend resources. It would be even better if we could store the state somewhere so that you could update and destroy the resources in the end as well.

For the short term though, I think we should just default the creation routine to `PAY_PER_REQUEST` instead of `PROVISIONED`.

ThatGerber: I understand the concern. I want to say, "But AWS rarely changes the APIs in backwards incompatible ways so it won't be a big deal", but it becomes a bigger dependency and that's always going to be difficult to deal with. 

I do like the idea of reading/writing from state, and could write up something in a few days to to read the provider/resource schema to generate the resource, but as you said, it needs to be written out to a file within the repo that Terraform can read (sans -grunt) otherwise running `terraform -` won't see it. And then either way, -grunt picks up a new dependency.

If it's agreed to update the default billing mode, I'll create a new PR for now to update create table call to set it to `PAY_PER_REQUEST`, and open an additional RFC to discuss how to manage buckets and lock tables going forward?

ThatGerber: Branch here: https://github.com/ThatGerber/terragrunt/tree/dynamodb-table-billing-mode

lorengordon: Do we know that `PAY_PER_REQUEST` is available in all partitions/regions? The announcement said it was available in all commercial regions, but what about `aws-us-gov`, or `aws-cn`? Or the new partitions exposed in recent SDKs, `us-iso` and `us-isob`?

yorinasub17: Not sure about ISO, but it is available in govcloud (https://aws.amazon.com/about-aws/whats-new/2019/06/amazon-dynamodb-on-demand-and-transactions-now-available-in-the-aws-govcloud-us-regions/) and CN (https://www.amazonaws.cn/en/dynamodb/pricing/).

With that said, that is a good point. Does anyone with access to ISO partitions know if it is available and if the error message from AWS is clear if it is not available? Perhaps we can catch that error and fall back to `PROVISIONED` if the first call with `PAY_PER_REQUEST` fails?

Alternatively, we can make this configurable... although in that case we would basically be introducing `dynamodb_table_config`.

ThatGerber: DynamoDB is available in `iso`/`iso-b` https://github.com/aws/aws-sdk-go/blob/master/aws/endpoints/defaults.go#L5278 and on the gov-cloud announcement page it says:

> Support for on-demand and transactions is available in all AWS Regions where DynamoDB is available.

lorengordon: the `iso` partitions are not govcloud. gimme a bit please, i am trying to get a straight answer...

ThatGerber: Is the accepted solution then that the default provisioning should be switched to `PAY_PER_REQUEST`? Would you like me to submit a new RFC proposing this or should I submit a merge request?

yorinasub17: Thanks for confirming that @lorengordon !

@ThatGerber At this point you can actually open a PR with the change and keep this open for further discussion to enable additional configuration.

yorinasub17:
Merging this in for archival reasons.

PR #937, opened 2019-10-30
Mutex output call

yorinasub17:
This is an alternative to https://github.com/gruntwork-io/terragrunt/pull/930, where we mutex the output call. This uses a mutex lock to ensure that only one `terragrunt output` call will be made at a time, avoiding the issues highlighted in #906 and #904 

As a bonus, this will reuse the outputs across calls. This is made possible with the additional synchronization.


config/dependency.go
```diff
@@ -273,10 +282,26 @@ func getTerragruntOutput(dependencyConfig Dependency, terragruntOptions *options
...
+	if hasRun {
+		util.Debugf(terragruntOptions.Logger, "%s was run before. Using cached output.", targetConfig)
+		jsonBytes = rawJsonBytes.([]byte)
+	} else {
+		newJsonBytes, err := runTerragruntOutputJson(terragruntOptions, targetConfig)
+		if err != nil {
+			return nil, true, err
+		}
+		jsonBytes = newJsonBytes
+		jsonOutputCache.Store(targetConfig, jsonBytes)
```
brikis98: Minor nit: consider extracting the locking and caching logic into a separate method that returns the JSON byte data. That'll make the code here easier to read and in the method simpler, as you'll be able to immediately `return rawJsonBytes`.

config/dependency.go
```diff
@@ -273,10 +282,26 @@ func getTerragruntOutput(dependencyConfig Dependency, terragruntOptions *options
 		return nil, true, errors.WithStackTrace(DependencyConfigNotFound{Path: targetConfig})
 	}
 
-	jsonBytes, err := runTerragruntOutputJson(terragruntOptions, targetConfig)
-	if err != nil {
-		return nil, true, err
+	// Acquire synchronization lock to ensure only one instance of output is called per config.
+	actualLock, _ := outputLocks.LoadOrStore(targetConfig, &sync.Mutex{})
```
brikis98: Perhaps this approach will allow you to not have to cast it twice below?

```suggestion
	actualLock, _ := outputLocks.LoadOrStore(targetConfig, &sync.Mutex{}).(*sync.Mutex)
```

brikis98: Also, sanity check: I assume `targetConfig` is a *canonical* file system path?

yorinasub17:
> Is there a way to add a regression test that has several modules looking up an output from the same module? Ideally, that test would fail before this change and pass after...

I had thought about this, but this doesn't quite run into the same error when using local modules: it has to be a remote module that is downloaded. But didn't have a good one to use...

brikis98:
We have a few examples we use with remote download URLs in tests. E.g., https://github.com/gruntwork-io/terragrunt/blob/master/test/fixture-download/remote/terragrunt.hcl#L6

config/dependency.go
yorinasub17: You can't do that because `LoadOrStore` returns two values.

> I assume targetConfig is a canonical file system path?

You're right. It wasn't before. Fixed: 8a97525

config/dependency.go
yorinasub17: Done: 8a97525

yorinasub17:
Ok added a regression test where I can verify the problem from master

```
    require.go:794:
                Error Trace:    integration_test.go:1983
                                                        integration_test.go:1987
                Error:          Received unexpected error:
                                Encountered the following errors:
                                Hit multiple errors:
                                error downloading 'https://github.com/gruntwork-io/terragrunt.git?ref=v0.21.0': /usr/local/bin/git exited with 128: fatal: destination path '/var/folders/n2/pljz6dq52bd1ksmw23qyr3sr0000gn/T/terragrunt-test637312762/fixture-get-output/regression-906/common-dep/.terragrunt-cache/KDUWLIAU9C2k7ArfPrqW64jW-1w/p_piCTTWVab2Hmnj1OtnAruj8J4' already exists and is not an empty directory.

                                unlinkat /var/folders/n2/pljz6dq52bd1ksmw23qyr3sr0000gn/T/terragrunt-test637312762/fixture-get-output/regression-906/common-dep/.terragrunt-cache/KDUWLIAU9C2k7ArfPrqW64jW-1w/p_piCTTWVab2Hmnj1OtnAruj8J4: directory not empty
                                Hit multiple errors:
                                error downloading 'https://github.com/gruntwork-io/terragrunt.git?ref=v0.21.0': /usr/local/bin/git exited with 128: fatal: could not create work tree dir '/var/folders/n2/pljz6dq52bd1ksmw23qyr3sr0000gn/T/terragrunt-test637312762/fixture-get-output/regression-906/common-dep/.terragrunt-cache/KDUWLIAU9C2k7ArfPrqW64jW-1w/p_piCTTWVab2Hmnj1OtnAruj8J4': File exists

                                unlinkat /var/folders/n2/pljz6dq52bd1ksmw23qyr3sr0000gn/T/terragrunt-test637312762/fixture-get-output/regression-906/common-dep/.terragrunt-cache/KDUWLIAU9C2k7ArfPrqW64jW-1w/p_piCTTWVab2Hmnj1OtnAruj8J4: directory not empty
                                Hit multiple errors:
                                error downloading 'https://github.com/gruntwork-io/terragrunt.git?ref=v0.21.0': /usr/local/bin/git exited with 128: fatal: could not create work tree dir '/var/folders/n2/pljz6dq52bd1ksmw23qyr3sr0000gn/T/terragrunt-test637312762/fixture-get-output/regression-906/common-dep/.terragrunt-cache/KDUWLIAU9C2k7ArfPrqW64jW-1w/p_piCTTWVab2Hmnj1OtnAruj8J4': File exists

                                Hit multiple errors:
                                error downloading 'https://github.com/gruntwork-io/terragrunt.git?ref=v0.21.0': /usr/local/bin/git exited with 128: fatal: destination path '/var/folders/n2/pljz6dq52bd1ksmw23qyr3sr0000gn/T/terragrunt-test637312762/fixture-get-output/regression-906/common-dep/.terragrunt-cache/KDUWLIAU9C2k7ArfPrqW64jW-1w/p_piCTTWVab2Hmnj1OtnAruj8J4' already exists and is not an empty directory.
                Test:           TestDependencyOutputRegression906
    integration_test.go:2391: Deleting test s3 bucket terragrunt-test-bucket-mbn3o8v7llzk
```

... and is fixed when run on this branch.

config/dependency.go
```diff
@@ -292,6 +301,31 @@ func getTerragruntOutput(dependencyConfig Dependency, terragruntOptions *options
 	return &convertedOutput, isEmpty, errors.WithStackTrace(err)
 }
 
+// getOutputJsonOrCache will run terragrunt output on the target config if it is not already cached.
+func getOutputJsonOrCache(targetConfig string, terragruntOptions *options.TerragruntOptions) ([]byte, error) {
```
brikis98: Minor nit: perhaps `getOutputJsonWithCaching`?

test/integration_test.go
```diff
@@ -1951,6 +1955,46 @@ func TestDependencyOutputRegression854(t *testing.T) {
 	require.NoError(t, err)
 }
 
+// Regression testing for https://github.com/gruntwork-io/terragrunt/issues/906
+func TestDependencyOutputRegression906(t *testing.T) {
```
brikis98: Minor nit: would probably be better if the function name reflected what was tested (e.g., TestDependencyOutputConcurrency), as the issue link is already in the comment.

yorinasub17: Done in 8362023

yorinasub17:
Will wait another day to see if we can get some confirmation that this works for the original issue reporters.

yorinasub17:
Given the regression test and https://github.com/gruntwork-io/terragrunt/issues/904#issuecomment-548229229, I am going to go ahead and merge + release this. If it turns out that this doesn't resolve the issue, we can revive https://github.com/gruntwork-io/terragrunt/pull/930

PR #939, opened 2019-11-01
Go modules

### LINKED ISSUE #925: Go modules
While I was trying to help upgrade terrafrom to v0.12.12, I have found the module dependency has been changed to use `Subpackage github.com/hashicorp/hcl/v2`.

Based on the [Version-Selection doc](https://github.com/hashicorp/hcl/wiki/Version-Selection), to be able to use hcl/v2, it requires to do the go modules upgrade.

Furthermore, with go v1.14, go modules would be the default toolchain, so I think it would be nice that we can upgrade the build toolchain to use go modules. 

Kind of relates to https://github.com/Homebrew/homebrew-core/pull/45521

chenrui333:
Closes #925


yorinasub17:
This is now being done in https://github.com/gruntwork-io/terragrunt/pull/939, and will continue the work there.

PR #948, opened 2019-11-11
Add more debug logging to help debug issues in the future

yorinasub17:
This introduces two new debug logs that would be helpful for debugging https://github.com/gruntwork-io/terragrunt/issues/906:

- Log the terragrunt version if `TG_LOG=debug`
- Log the start of output retrieval for a dependency block

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -176,26 +176,6 @@
   version = "v2.0.4"
 
 [[projects]]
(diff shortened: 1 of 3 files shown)


yorinasub17:
Going to merge this in given that this is pretty straightforward and will help with the current thread in #906.

PR #969, opened 2019-12-03
export runTerragrunt (i.e. make it RunTerragrunt)

### LINKED ISSUE #968: Ability to run terragrunt programmatically
It would be really helpful to be able to run terragrunt programmatically (i.e. link terragrunt into another go executable). Unfortunately, this is made quite difficult by `runTerragrunt` in `cli_app.go` not being exported. Could we discuss the possibility of exporting it? I'd be happy to raise a PR if we felt it was a suitable direction. Naturally, if there's another/better way of achieving what I'm trying to do I'd be very open to opinions.

rvodden:
This PR resolves #968 by exporting `RunTerragrunt` so that terragrunt can be executed programmatically.


yorinasub17:
Build passed so merged! Will cut a release soon.

Issue #724, opened 2019-06-04
Terragrunt Console Garbles STDIN/STDOUT Streams


Tensho:
I experience a strange behavior with **`terragrunt console`** command. When I try to run the next command:

    $ terragrunt console --terragrunt-source ...
    ...
    [terragrunt] 2019/06/04 13:35:06 Running command: terraform console
    Acquiring state lock. This may take a few moments...

    <Ctrl + C>

    Releasing state lock. This may take a few moments...

it looks like `terragrunt` hangs on **acquiring state lock forever**. However, if I change directory to `terragrunt` cache directory an try to run console manually through `terraform` CLI everything works as expected:

```
$ cd .terragrunt-cache/rXnIXItbUZO3QgycMk-_hiaL6ek/ZzjeLbJj6m48XPN-IEhSqTgSFCE
$ terraform console
Acquiring state lock. This may take a few moments...
> aws_route53_zone.this.zone_id
FAKE5JTM3YS01O
> exit
Releasing state lock. This may take a few moments...
```
All other `terragrunt` commands work as expected.
 
Later I found **`terragrunt console`** actually hides the console prompt and typed characters, but outputs results:
```
[terragrunt] 2019/06/04 13:35:06 Running command: terraform console
Acquiring state lock. This may take a few moments...

<hit Enter>
parse error at 1:3: expected expression but found ">"

<type 1>
1

<type aws_route53_zone.this.zone_id>
FAKE5JTM3YS01O
```
Seems like `terragrunt` garbles STDIN/STDOUT streams.

Tensho:
Ah, I've just realized `terragrunt` hides the console prompt and typed characters, but outputs results:
```
[terragrunt] 2019/06/04 13:35:06 Running command: terraform console
Acquiring state lock. This may take a few moments...

<hit Enter>
parse error at 1:3: expected expression but found ">"

<type 1>
1

<type aws_route53_zone.this.zone_id>
FAKE5JTM3YS01O
```
Seems like `terragrunt` garbles STDIN/STDOUT streams.

udondan:
Confirming this, observed the same behavior

edjackson-wf:
It looks to me like the `>` prompt and keyboard input are just not echoed to the screen. While I can't see what I'm typing, the correct output is actually displayed if I enter a valid expression.

yorinasub17:
[...]

PR #970, opened 2019-12-04
Add option to use pseudotty for interactive commands

yorinasub17:
This resolves https://github.com/gruntwork-io/terragrunt/issues/724, which is caused by the fact that we are not using a pseudo-tty for interactive prompts. The terminal driver is needed for the readline to work correctly with commands like `terraform console`.

After this fix, you should be able to use the console via terragrunt with `terragrunt console --terragrunt-ptty`


shell/run_shell_cmd.go
```diff
@@ -66,16 +65,26 @@ func RunShellCommandWithOutput(terragruntOptions *options.TerragruntOptions, wor
...
 	} else {
-		cmd.Stdout = io.MultiWriter(&stdoutBuf)
+		cmdStdout = io.MultiWriter(&stdoutBuf)
 	}
 
-	if err := cmd.Start(); err != nil {
-		// bad path, binary not executable, &c
-		return nil, errors.WithStackTrace(err)
+	if terragruntOptions.AllocatePseudoTTY {
+		if err := runCommandWithPTTY(terragruntOptions, cmd, cmdStdout, cmdStderr); err != nil {
```
brikis98: Sanity check: any reason not to _always_ run with a pseudo TTY so we don't need an extra flag to document, maintain, etc? 

Or, alternatively, IIRC, the `terminal` package in Go can tell you if you're currently in a terminal / tty, and we can use that to decide this automatically.

shell/ptty_unix.go
```diff
@@ -0,0 +1,82 @@
...
+	"syscall"
+
+	"github.com/creack/pty"
+	"golang.org/x/crypto/ssh/terminal"
+
+	"github.com/gruntwork-io/terragrunt/errors"
+	"github.com/gruntwork-io/terragrunt/options"
+)
+
+func runCommandWithPTTY(terragruntOptions *options.TerragruntOptions, cmd *exec.Cmd, cmdStdout io.Writer, cmdStderr io.Writer) (err error) {
```
brikis98: Did you come up with this code yourself or are you following some sort of guide or example? It seems very complicated!

shell/run_shell_cmd.go
yorinasub17: It isn't actually whether or not we are in a terminal, but rather whether or not the command we want to run is interactive so that we get proper prompts and readline support when we call it via `exec`. So really, the question is whether the command we are about to run is interactive or not. I guess we can maintain a list of terraform commands that are interactive, since it is unlikely terraform will have another command like `console`.

Will go ahead and update that logic.

shell/ptty_unix.go
yorinasub17: Ah I forgot to add the links. This is based on the quickstart example for https://github.com/creack/pty

shell/run_shell_cmd.go
yorinasub17: > Sanity check: any reason not to always run with a pseudo TTY so we don't need an extra flag to document, maintain, etc?

Also, we don't want to do this all the time, because it messes up stderr from the commands.

yorinasub17: UPDATE (2704efe): Switched to automatically use ptty routine if the terraform command is `console`.

shell/ptty_unix.go
yorinasub17: UPDATE (2704efe): added link to project and mentioned that the code is based on quickstart example.

brikis98: Ah, got it, thanks!

shell/run_shell_cmd.go
brikis98: Roger, thx!

shell/run_shell_cmd.go
```diff
@@ -14,29 +14,41 @@ import (
 
 	"github.com/gruntwork-io/terragrunt/errors"
 	"github.com/gruntwork-io/terragrunt/options"
+	"github.com/gruntwork-io/terragrunt/util"
 )
 
+var interactiveTerraformCommands = []string{
+	"console",
```
brikis98: Is interactive the right term here? E.g., if you run `terraform apply`, it interactively prompts you for a "yes" to confirm deployment. That part works just fine with Terragrunt as-is. So is t more that we need a REPL? Or even more explicitly, `terraformCommandsThatNeedPty`?

yorinasub17: Good point. Renamed! 9df6643

yorinasub17:
Build passed on the last improvement, so going to merge this in now. Thanks for the review!

PR #1005, opened 2020-01-15
fix broken links[HSTN-909]

hngerebara:
fix broken links


hngerebara:
This actually breaks the links in the jekyll site. Not sure how to fix this so I'll bring it up with the team

hngerebara:
Closing this PR. Spoke with the team and we decided this is a very low priority fix since jekyll uses templates. We might not be able to do anything about it. Also the probability of a user transversing so deep into the tree such that they hit this error is very low.

PR #1009, opened 2020-01-20
Reference documentation

yorinasub17:
I oftentimes find myself looking for a reference to a particular block or attribute in the terragrunt docs (now website) and never find what I am looking for because it is buried in one of the use case docs or feature description. Many projects provide a comprehensive API reference style documentation that enumerates all the functions and config that is available. I think it would be useful for terragrunt to have the same, so this PR does exactly that:

- Add a 4th section in the nav, `Reference`
- Move the "CLI options" and "Built-in functions" reference docs to `Reference`
- Add a new page in `Reference` that enumerates all the blocks and attributes supported in `terragrunt.hcl`

## Bonus feature

I could not get the website to work on my Mac. I had lots of issues getting `bundle install` to work, especially compiling `libv8` for `the_ruby_racer`. In the end I gave up trying and added a `Dockerfile` with `docker-compose.yml` so we can use docker for local development. That worked beautifully.


docs/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM ruby:2.6.2-stretch
```
brikis98: Ah, nice idea to add `Dockerfile` 👍

brikis98: BTW, I'm guessing the issues you had were due to Ruby versions, which requires messing with `rbenv` (or similar)... In these cases, a known working `Dockerfile` can be a lot nicer.

tomaszantas:
Yeah, changes in the PR and a new approach are compatible with each other.

The `navigation.yml` will be removed, so less manual work will be required.

The "Configuration Blocks and Attributes" contains **Blocks**, **terraform** and **remote_state** as headings so all of them will be present in navigation.

docs/Dockerfile
yorinasub17: FWIW, the issues I had was actually not the ruby versions (I think I had `rbenv` set up correctly), but the dependency on a system library (`libv8` which requires `v8` on your system) and getting the compiler to work. Here are the issues I ran into:

- The build script uses `python` and requires python2 which was annoying because I had `pyenv` hard coded to python 3.
- I needed `v8` to already have been installed since it is dynamically linked to the ruby library. `brew install v8` solved the issue.
- I could not get `libv8` to compile, because it kept running into issues with linking to the stdlib. This appears to be a known issue with using `clang++` instead of `gcc`. The workaround is to install `gcc`, but I did not want to do that as that would mess other things up on my mac. I fiddled with the flags to see if there was a workaround but ultimately gave up and switched to docker.

brikis98: Roger.

yorinasub17:
This is now ready for review. Note that I did have an issue getting the nav to work properly. When I navigate directly to the configuration reference page (e.g enter the URL directly), the navbar is not expanded:

Configuration_Blocks_and_Attributes

But it works if I navigate from the side bar:

Configuration_Blocks_and_Attributes

I don't seem to have the same issue with the other pages ("CLI options" and "Built in functions") so 
not sure what I did wrong here.

docs/_data/navigation.yml
```diff
@@ -197,3 +139,108 @@ children:
...
+            href: /docs/reference/built-in-functions/#get_aws_account_id
+          - name: get_aws_caller_identity_arn
+            href: /docs/reference/built-in-functions/#get_aws_caller_identity_arn
+          - name: get_aws_caller_identity_user_id
+            href: /docs/reference/built-in-functions/#get_aws_caller_identity_user_id
+          - name: run_cmd
+            href: /docs/reference/built-in-functions/#run_cmd
+
+      - name: Configuration
+        href: /docs/reference/config-blocks-and-attributes
```
tomaszantas: I think that missing slash `/` at the end of `href` can be the reason. 

However, I think that you can ignore this. Next PR #1014 will remove `docs/_data/navigation.yml` at all and replace the way how the navigation is being built.

jeffdyke:
As the author of https://github.com/gruntwork-io/terragrunt/issues/1010, i've been following this and it looks like its going to be a great resource for people coming into using the project and folks that felt there way through (me) and still have a lot to learn.  Kudos!

yorinasub17:
Updated to address merge conflicts with nav updates. @brikis98 this is ready for review now! Would be good to get this in sooner than later because it makes RDD harder. E.g the page being updated for https://github.com/gruntwork-io/terragrunt/pull/1021 has been moved on this branch.

yorinasub17:
Shared screen share on slack.

yorinasub17:
Got approval on slack so going to merge this now! Thanks for the review!

PR #1014, opened 2020-01-22
Switch to self-built navigation.

tomaszantas:
Minor fixes:
- filters in use cases
- CSS rules for ticks in checkboxes


docs/_includes/collection_browser/navigation/_collection_toc.html
```diff
@@ -0,0 +1,391 @@
+{% capture tocWorkspace %}
+  {% comment %}
+    To build navigation:
```
brikis98: Wow, this is quite complex. Thx for documenting it!

Sanity check: is there something about the layout of the content/nav we could change to simplify this so future maintenance is easier?

tomaszantas:
> Thanks! This looks like a big improvement.
> 
> However, I hit a nav issue on https://altalogy.github.io/terragrunt. Here's how to repro:
> 
> 1. Click "Install" in the top nav
> 2. Expand "Features" in the left nav
> 3. Click "Keep your Terraform code DRY"
> 4. Get 404 page at https://terragrunt.gruntwork.io/terragrunt/use-cases/keep-your-terraform-code-dry
> 
> I tried to look at the nav code to see what was wrong, but it's quite complicated, so please see my question there about what we can do to simplify it. Thx!

I think that we can ignore this one.
"Use cases" under **Features** use `redirect_to` plugin. It builds the link using:
a) `url` from *_config.yml*
b) `baseurl` from *_config.yml*
c) `redirect_to` from *doc*

https://altalogy.github.io/terragrunt had the URL set to `https://terragrunt.gruntwork.io/`. I changed it on a forked repo to use `altalogy.github.io`.It only affects the forked repo, so nothing has to be changed on `github.com/gruntwork-io/terragrunt`.

docs/_includes/collection_browser/navigation/_collection_toc.html
tomaszantas: I think that it's not possible to improve it.
Liquid language is not the best tool to write such code.
GitHub Pages doesn't support custom plugins so we cannot build the nav with Ruby.

brikis98:
> I think that we can ignore this one.

Roger

docs/_includes/collection_browser/navigation/_collection_toc.html
brikis98: I completely agree Jekyll is limited for this sort of thing. But what I mean is, what assumptions/requirements can we change to reduce the amount of complexity we have to shove into the Jekyll code? Some examples:

1. What if in Jekyll, we only generate the top-level of the nav and use JavaScript to dynamically fill in the nested nav items (i.e., the section headings)? 
1. Could we re-oganize the contents of the nav itself to simplify how the nav is assembled?

If not, I understand, but figured I'd ask!

tomaszantas: **The first challenge** is to build lower levels of navigation which come from documents' headings (h1, h2, h3, etc).

To build TOC of the single document, we need to parse the HTML and find all headings. It could be done with JS after the page is loaded in the web browser. However, we don't have an access to other docs (subpages) at that moment - other than the document opened in the web browser. So we cannot build TOC of other docs in the documentation. It means that navigation could contain the collection's categories, pages and detailed Table of Content only for the current document. 

If it would be fine to not have a preview of other docs' content, we can even make things simpler with a plugin supported by GitHub Pages: https://github.com/allejo/jekyll-toc 

I think it's not totally bad idea. We could make two-part navigation. The first part would navigate between collection categories and pages. The second part would navigate within the current document. The problem is that if something should be discoverable in navigation, it would have to be a separate document (page). For instance: https://terragrunt.gruntwork.io/docs/features/built-in-functions/
If each function should be visible in the nav when a user reads another doc, then each function should be a subpage.

Alternatives:
a) predefine headings of documents in .yml file (like it's done now with `_data/navigation.yml`
b) create a custom Jekyll plugin - but it's not supported by GitHub Pages
c) extract headings from docs at Jekyll building time and store in hidden HTML tags or JS. Then we can build navigation with JS. But I think it is a worse approach than proposed in this PR where navigation is built right away as HTML code and embedded into the page.


**The second challenge** are nested collections.
The script does three things:
1. builds navigation for each document from headings (the script parses the document in HTML and looks for h1, h2, h3, etc.) 
2. adds collection's categories and documents' titles to the navigation
3. **embeds nested collections** - this actually required to rewrite again (1) and (2) in Liquid (duplicated code)

Removing nested collections would decrease the code length by circa 40%. 
Moving to JS could make things a more _DRY_, but on the other hand, it forces to store data in JS or hidden HTML instead of building the code right away.

brikis98: > If it would be fine to not have a preview of other docs' content, we can even make things simpler with a plugin supported by GitHub Pages: https://github.com/allejo/jekyll-toc

I think this is the right way to go. Having just the top-level pages in the nav and the headings solely for the currently selected page is just fine. Let's do that!

> Removing nested collections 

Sorry, dumb question alert: what do you mean by nested collections? Can you give an example?

tomaszantas: "Use cases" is a collection nested in "Docs". By _nested_ I mean that "use cases" are a part of "Docs" nav.

brikis98: Sorry, I'm still missing something. Don't *all* the pages appear somewhere in the docs nav? Is the difference with use cases that they appear both under features *and* under getting started -> use cases?

tomaszantas: You are right, there is a difference. "Use cases" under "getting started" are handled as **nested collection**, but these under "features" are **redirections** to "use cases" in "getting started".

brikis98: Does it simplify the code if we remove "Use Cases" as a top-level nav item under "Getting Started" and solely have each use case as a normal page under "Features"?

If yes, then let's do that. If not, then let's just do the simplification mentioned earlier (have just the top-level pages in the nav, plus JS to fill in subheadings for current page).

tomaszantas: Yes, it does.
Changing "Examples" in Terratest makes things also easier there, so navigations in both projects can be done in the same way.

brikis98: OK, great, let's do that!

Issue #1020, opened 2020-01-27
Add ability for a before_hook that runs before initializing remote state


sleungcy:
I am attempting to set up a hook before terragrunt initialize remote state to retrieve AWS credential from vault. However terragrunt just initializes the remote state before triggering the "init" before_hook..

[terragrunt] 2020/01/27 20:28:27 Reading Terragrunt config file at /app/microservices/terragrunt.hcl
[terragrunt] [/app/microservices/terragrunt.hcl] 2020/01/27 20:28:27 Running command: terraform --version
[terragrunt] 2020/01/27 20:28:27 Initializing remote state for the s3 backend
[terragrunt] 2020/01/27 20:28:47 Error finding AWS credentials in file '/tmp/.aws_credentials' (did you set the correct file name and/or profile?): NoCredentialProviders: no valid providers in chain. Deprecated.
	For verbose messaging see aws.Config.CredentialsChainVerboseErrors

sleungcy:
Are there other hooks i can use that are before "init"? Is there a way I can put a hook before it runs terraform --version?

yorinasub17:
I think you want to use `init-from-module`. See the last bullet in https://terragrunt.gruntwork.io/docs/features/before-and-after-hooks/ for more details.

sleungcy:
unfortunately, init-from-module also runs after the state initialization.

yorinasub17:
Gotcha. Then we don't currently support this feature, so we will need to implement it. We're unlikely to get to this anytime soon, so if you are in need of this in the short term, your best bet is to submit a PR to add support for this. Otherwise, you can wait for someone from the community or us to get around to implementing this.

sleungcy:
Do I need an RFC for this?

yorinasub17:
Thanks for checking in on that! This is simple enough that the corresponding README updates to [the before hook docs](https://github.com/gruntwork-io/terragrunt/blob/master/docs/_docs/02_features/before-and-after-hooks.md) is sufficient.

PR #1021, opened 2020-01-28
adding terragrunt-read-config

sleungcy:
https://github.com/gruntwork-io/terragrunt/issues/1020


cli/cli_app.go
```diff
@@ -52,6 +52,7 @@ const CMD_INIT = "init"
 const CMD_INIT_FROM_MODULE = "init-from-module"
 const CMD_TERRAGRUNT_INFO = "terragrunt-info"
 const CMD_HCLFMT = "hclfmt"
+const CMD_VERSION = "get-terraform-version"
```
yorinasub17: After taking a look at this, I wonder if this should actually be called `any-action`. Getting the terraform version is the first step right now, but there is no guarantee that this will be the case going forward. And once we shift things around, I don't see too much use in a before hook that wraps the `terraform version` call.

cc @brikis98 : would value a second opinion here.

brikis98: Agreed that `get-terraform-version` is not the right name, as (a) it's specific to our current implementation, which could change and (b) even if it didn't change, there's no reason users should need to know that step is part of the implementation. Not sure `any-action` makes it clear that this would run as one of the very first steps. Perhaps `terragrunt-init` or `terragrunt-run`?

sleungcy: I agree terragrunt-init sounds right, however as you said the current implement may change, so terraform version may not be the correct place to latch onto

sleungcy:
What do you think about "terragrunt-read-config" instead? This would be right before and after terragrunt initialize and finish loading terragrunt configurations. In this case, only the after_hook will mattter.

yorinasub17:
An `after_hook` for `terragrunt-read-config` feels slightly awkward to me for this use case (an "after" hook whos purpose is to be the before hook for everything else). However, with that said, it is very clear and predictable when it will fire, and is likely to be stable going forward so I think that works for me!

Issue #744, opened 2019-06-13
Terraform 0.12 map merging


justicel:
It appears that Terraform 0.12 no longer merges maps, but only selects the 'newest' one when using a chained set of tfvars files. It is documented at the bottom of this page:

https://www.terraform.io/docs/configuration/variables.html

This definitely breaks a good part of my workflow with 0.12. I'm wondering if perhaps this could be something that is supported in the terragrunt.hcl inputs = {} sections so that you can chain map variables and merge them like with 0.11.x?

brikis98:
Hm, I don't think I'd want to merge maps _by default_. I think that would be unexpected and is one of the reasons Terraform itself stopped doing it.

However, if there was an explicit way to opt into that sort of behavior, it would make sense. For example, in a few issues, the idea of adding a `get_input` helper has come up:

```hcl
inputs = {
  foo = "bar"
  baz = get_input("../common.hcl", "baz")
}
```

That allows you to explicitly reuse a single input from another `.hcl` (or `.tfvars`?) file. That could be combined with an `merge` function:

```hcl
inputs = {
  foo = "bar"
  baz = merge(get_input("../common.hcl", "baz"), { "foo": "bar" })
}
```

This is verbose, but clear and explicit... It would work if you only have a small number of such variables, but might not be effective if you have dozens.

brikis98:
As of [v0.19.4](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.19.4), `merge` (and all other Terraform built-in functions) is now available.

rdebrandt:
@brikis98 

I've been trying to figure out the best way to do this, and I'm just not clear where terraform built-in functions can be run. The end goal in my scenario is specifically merge tags where I could have account level tags (or any arbitrary directory level) that would be applied to all sub-directories and merged into a single tag map to be applied to all resources.

The documentation under locals (https://github.com/gruntwork-io/terragrunt#locals) seems to indicate the only way to do this is by adding something like the below which works. The only issue is that now we have to use another configuration yaml configuration file throughout the directory structure, and the yaml files only need to be used for this type of map merge scenario which adds a bit of overhead/confusion. 

`locals {

  account_tags = yamldecode(file("${get_terragrunt_dir()}/${find_in_parent_folders("account.yaml")}"))
  local_tags = { 
     "app"  = "local_app"
     "type" = "local_type"
  }
  tags = merge(
    local.account_tags,
    local.local_tags
  )

}
`

[...]

PR #1025, opened 2020-01-30
[RFC] Imports: soliciting feedback from community

yorinasub17:
This is an RFC that introduces `import` blocks for addressing limitations of configuration reuse with `include`.

I would like feedback from the entire community on this, as this has the potential to address a lot of limitations that were proposed throughout the lifetime of terragrunt, but is fairly opinionated in the design principle.

NOTE: It may be easier to read in the rendered form. Here is [the direct link](https://github.com/gruntwork-io/terragrunt/blob/yori-enhanced-dependency/_docs/rfc/imports.md).

### DIFF
--- a/_docs/rfc/imports.md
+++ b/_docs/rfc/imports.md
@@ -125,6 +125,7 @@ The `import` block works as follows:
 - Imports support auto merging via the `merge` setting. When `merge = true`, all the blocks and attributes of the
   imported config will be merged with the child config. Note that since HCL blocks are sequential, imports will be
   merged top to bottom. See below for more details.
+- Imports can also be deep merged using the `deep_merge` setting. This will work similar to `merge = true`.
 - Imports will also export the absolute path to the directory of the terragrunt config file as the attribute
   `terragrunt_dir`.
 - Imports are parsed first in the [configuration parsing
@@ -503,7 +504,7 @@ child
 ```hcl
 import "root" {
   config_path = find_in_parent_folders("root.hcl")
-  merge = true
+  deep_merge = true
 }
 
 remote_state {
@@ -518,8 +519,8 @@ Note how we explicitly set the relative path using only the context of the curre
 circular reference here where you need the context of the child to know the exact values of the attributes being set in
 the parent (although you won't know the full setting of `remote_state` without seeing the child).
 
-Additionally, here we take advantage of the `merge` feature to deep merge the two configurations for `remote_state`.
-This is equivalent to having the following `remote_state` block in the child config:
+Additionally, here we take advantage of the `deep_merge` feature to deep merge the two configurations for
+`remote_state`.  This is equivalent to having the following `remote_state` block in the child config:
 
 ```hcl
 remote_state {
--- a/_docs/rfc/imports.md
+++ b/_docs/rfc/imports.md
@@ -104,7 +104,7 @@ expensive solutions. The following is a summary of the solutions to be built:
 
 - [Short term: read_terragrunt_config helper function](#read_terragrunt_config-helper-function)
 - [Medium term: import blocks](#import-blocks)
-- [Long term: single terragrunt.hcl file](#single-terragrunt-hcl-file-per-environment)
+- [Long term: single terragrunt.hcl file](#single-terragrunthcl-file-per-environment)
 
 ### read_terragrunt_config helper function
 
@@ -272,7 +272,7 @@ Cons:
 Let's take a look at a few common use cases and how we might use `import` to address them:
 
 - [Hierarchical variables included across multiple terragrunt.hcl
-  files](#import-block-hierarchical-variables-included-across-multiple-terragrunt-hcl-files)
+  files](#import-block-hierarchical-variables-included-across-multiple-terragrunthcl-files)
 - [Reusing common variables](#import-block-reusing-common-variables)
 - [Reusing dependencies](#import-block-reusing-dependencies)
 - [Keeping remote state configuration DRY](#import-block-keeping-remote-state-configuration-dry)
--- a/_docs/rfc/imports.md
+++ b/_docs/rfc/imports.md
@@ -586,7 +586,127 @@ blocks in this way.
 
 ## Alternatives
 
-### Enhancing include with import semantics
+### In consideration
+
+#### hcldecode or read_terragrunt_config helper function
+
+Instead of defining a dedicated block for `import`, we could define a helper function that parses the relevant config.
+For example, the explicit triple import example in the [hierarchical variables use
+case](#hierarchical-variables-included-across-multiple-terragrunt-hcl-files) can be implemented as:
+
+```
+locals {
+  root_config = read_terragrunt_config("../../../root.hcl")
+  region_config = read_terragrunt_config("../../region.hcl")
+  env_config = read_terragrunt_config("../env.hcl")
+}
+
+inputs = merge(
+  local.root_config.inputs,
+  local.region_config.inputs,
+  local.env_config.inputs,
+  {
+    # args to module
+  },
+)
+```
+
+Pros:
+
+- Relatively simple implementation. It is significantly easier to add helper functions than it is to introduce new
+  blocks in terragrunt.
+- Can address all the same use cases as `import` blocks.
+
+Cons:
+
+- Everything has to be explicit in the config. We can not support automatic merging capabilities when using helper
+  functions, as you can't manipulate the config in a helper. This can lead to very verbose configurations if one wants
+  to use this in place of `include`.
+
+Note that to take full advantage of this approach, all the blocks in `terragrunt.hcl` need to be redefined as attributes
+so that we can use assignment to override them.
+
+Let's walk through a few more of the use cases:
+
+**Keeping remote state configuration DRY**
+
+parent
+
+```hcl
+remote_state {
+  backend = "s3"
+  config = {
+    bucket         = "my-terraform-state"
+    key            = "${path_relative_to_include()}/terraform.tfstate"
+    region         = "us-east-1"
+    encrypt        = true
+    dynamodb_table = "my-lock-table"
+  }
+}
+```
+
+child
+
+```hcl
+locals {
+  root_config = read_terragrunt_config("../../../root.hcl")
+}
+
+remote_state {
+  backend = local.root_config.remote_state.backend
+  config = merge(
+    local.root_config.remote_state.config,
+    {
+      # The relative path between the root terragrunt directory and the current terragrunt file directory.
+      key =  relpath(import.root.terragrunt_dir, get_terragrunt_dir())
+    },
+  )
+}
+```
+
+Note that we can't do:
+
+```hcl
+remote_state = deep_merge(local.root_config.remote_state, { config = { key = relpath } })
+```
+
+due to the fact that `remote_state` is a block and not an attribute.
+
+
+**Reusing dependencies**
+
+We can't reuse `dependency` blocks in this implementation because there is no way to auto merge the blocks.
+
+If `dependency` was instead an attribute, we could use the following alternative syntax:
+
+vpc_dependency_config.hcl
+
+```
+dependency = {
+  vpc = {
+    config_path = "/path/to/app/vpc/module"
+  }
+}
+```
+
+```
+locals {
+  common_deps = read_terragrunt_config("../vpc_dependency_config.hcl")
+}
+
+dependency = local.common_deps.dependency
+
+inputs = {
+  name = "unique-name"
+  vpc_id = dependency.vpc.outputs.vpc_id
+}
+```
+
+
+
+### Rejected
+
+#### Enhancing include with import semantics
 
 Instead of having a dedicated block with the new functionality, we could enhance `include` to have all the semantics of
 `import`. This reduces complexity of the configuration by being able to recycle a very similar construct that already
@@ -623,7 +743,7 @@ Cons:
 Ultimately, the potential for badly shooting yourself in the foot was enough to warrant a new block to start clean.
 
 
-### globals
+#### globals
 
 Globals was a potential solution to the problem proposed by the community. `globals` work the same way as `locals`,
 except they support merging across `include`. For example, to address the use case of [reusing common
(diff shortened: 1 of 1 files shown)


_docs/rfc/imports.md
```diff
@@ -0,0 +1,766 @@
...
+## Open Question
+
+- While a major design goal in this RFC is to keep the principles of explicit over implicit, practically speaking it may
+  not be feasible to achieve that design goal to support all the use cases of the community. In that regard, it may be
+  useful to implement a minimal set of features that don't take a whole lot of cognitive load to understand. Should the
+  following functions be implemented for use with `import`?
+    - `path_relative_to_import` and `path_relative_from_import`: These are the equivalent functions of the ones named
+      for `include`.
+    - `find_in_parent_folders_from_importing_config`: This function is the version of `find_in_parent_folders` that
+      works in the context of the config that is importing the current config.
```
yorinasub17: Use case that would benefit heavily from this: https://github.com/gruntwork-io/terragrunt/issues/744#issuecomment-580092107

brikis98:
Thanks for the great write-up Yori. I love going through all the concrete use cases to make sure it's solving the right problems 👍 

Of the "reuse" options on the table—that is, the existing `include`, globals, `read_terragrunt_config`, manually reading JSON/YAML files, and your new `import` idea—I think `import` is probably the best bet, with `read_terragrunt_config` not too far behind (though I'd want to go through all the explicit use cases with it to see what they look like). However, there is one more option to consider that isn't explicitly about "reuse", which is the proposal I made here: https://github.com/gruntwork-io/terragrunt/issues/759. 

It seems like with that approach, most of the code reuse issues go away. Everything is in one file (or, more likely, one set of `.hcl` files in a single folder), you can use `locals` to share variables, you can use `module.xxx` expressions to express dependencies, you don't need too many special Terragrunt helpers, and so on. In short, it's *almost* vanilla Terraform, which I'd argue is a very good thing, but with some extra functionality for better permissions management and avoiding everything ending up in one giant state file. If we're going to make a large change, this approach still strikes me as worth considering. How does it handle the code reuse use cases we are looking at with `merge`?

tomasbackman:
Yes, very good write up and summary. Im really looking forward to having any of these solutions come to life as the current state is a bit annoying with the limitations in reusability.
Personally I would be happy with any of the proposed solutions. Most important for me is that they do not take too long to get.. terragrunt is after all a great wrapper that makes terraform easier/better. So letting these limitations stay too long feels worse (and contradicting the goal more) for me than getting the slightly "wrong" implementation (that still solves the issue.. =)

yorinasub17:
Thanks for the feedback so far!

> However, there is one more option to consider that isn't explicitly about "reuse" ...

Ah I was thinking of that as orthogonal, but you bring up a good point that it is related. When I get back to more thinking on this, I'll update with a section discussing this alternative.

> I think import is probably the best bet, with read_terragrunt_config not too far behind (though I'd want to go through all the explicit use cases with it to see what they look like)

Given that `read_terragrunt_config` will probably be the easiest to implement, and to @tomasbackman point about speed of delivery and not letting this problem linger, I think it is worth baking that a bit more.

---

Given the above two and the feedback I've received, next steps on this RFC are for me to:

- [ ] Update section on `read_terragrunt_config` as a function to walk through the use cases
- [ ] Add section on #759 
- [ ] Reorganize alternatives so that it highlights that the three top runners are: `import`, `read_terragrunt_config`, and #759

jfunnell:
I agree with @brikis98 for what it's worth.. Flattening the structure and allowing locals to be used between modules would solve most problems I'm experiencing, though I guess there could be more complicated setups that might still need some nested, fancy shared variable hierarchy that your proposal seems to cover.

yorinasub17:
UPDATE:

- 542509 : Add walkthroughs of each scenario for `read_terragrunt_config`. Note that after walking through the examples, the only way I see this avoiding the verbosity is if we convert all the blocks to attributes.

- 2b1039d : Include commentary on how #759 addresses the details in this RFC.

_docs/rfc/imports.md
```diff
@@ -0,0 +1,963 @@
...
+
+**Hierarchical variables**
+
+In this approach, a hierarchy of variables is unnecessary because all the blocks are defined in a single scope. However,
+depending on the scope of `locals`, certain things like reusing a repetitive variable becomes more challenging.
+
+To understand this, consider a use case where you are operating in two regions, `us-east-1` and `eu-west-1`. To simplify
+this example, consider a limited application deployment where you have two modules: `vpc` and `app`.
+
+If we assume that `locals` are scoped within the file that they are defined, then you can implement this by separating
```
brikis98: Why would locals be scoped to a file? The idea with the new proposal is that all `.hcl` files in a single folder are processed as if they are in a single file, just as Terraform processes all `.tf` files in a single folder as if they are in a single file.

_docs/rfc/imports.md
```diff
@@ -0,0 +1,963 @@
...
+}
+
+module "app" {
+  # additional args omitted for brevity
+  aws_region = local.region
+}
+```
+
+However, if the namespace of `locals` was shared across the environment, then the above approach would not work and each
+region file will need to define a different name for the region variable.
```
brikis98: Yea, if you're deploying the same things across regions, you'd use a different local for each region.

_docs/rfc/imports.md
```diff
@@ -0,0 +1,963 @@
...
+have each environment be at the region level. The advantage of this approach is to define a different state bucket for
+each region, which is a better posture for disaster recovery. In this approach, it doesn't matter what the scope of the
+`locals` is. However, the downside of this approach is that there will be some repetition to pull in the AWS account ID
+info across all the different regions.
+
+**Reusing common variables**
+
+Reusing common variables depends on the scope of `locals`. If the scope of `locals` is shared across the environment,
+then you can define the common variable in `locals` blocks to share across the entier environment. If instead the scope
+of `locals` is per file, we either:
```
brikis98: Pretty sure `locals` and everything should be shared across all `.hcl` files in a folder, just like how Terraform does it for `.tf` files.

_docs/rfc/imports.md
```diff
@@ -0,0 +1,963 @@
...
+  to use this in place of `include`.
+
+Note that to take full advantage of this approach, all the blocks in `terragrunt.hcl` need to be redefined as attributes
+so that we can use assignment to override them.
+
+Let's walk through a few more of the use cases:
+
+**Keeping remote state configuration DRY**
+
+parent
```
brikis98: There is no more parent or child with the new approach. All the configs for an account are in the same folder and there is one `remote_state` config in one of those `.hcl` files that looks something like this:

```hcl
remote_state {
  backend = "s3"
  config {
    bucket = "foo"
    region = "us-east-1"
    # module_name returns the name of the current module (e.g., frontend-app, mysql, etc)
    key = "${module_name()}/terraform.tfstate"
  }
}
```

The new `module_name()` helper is how we ensure each module ends up in a different state file.

_docs/rfc/imports.md
```diff
@@ -0,0 +1,963 @@
...
+```hcl
+remote_state = deep_merge(local.root_config.remote_state, { config = { key = relpath } })
+```
+
+due to the fact that `remote_state` is a block and not an attribute.
+
+
+**Reusing dependencies**
+
+We can't reuse `dependency` blocks in this implementation because there is no way to auto merge the blocks.
```
brikis98: I'm not sure we'd need `dependency` blocks at all... Each account is standalone and dependencies between modules within an account are already captured.

_docs/rfc/imports.md
yorinasub17: There is an advantage to having vars scoped to the file so that you don't have to keep picking new names for a thing like `region`.

_docs/rfc/imports.md
yorinasub17: See above for why one would want variables that are scoped to the file. E.g function scope vs global scope to avoid namespace clashes.

_docs/rfc/imports.md
yorinasub17: This is actually the section for `read_terragrunt_config`, not the approach for single tg file. See line 709

_docs/rfc/imports.md
yorinasub17: This is actually the section for read_terragrunt_config, not the approach for single tg file. See line 709

_docs/rfc/imports.md
brikis98: Agreed, but I think in net, treating all `.hcl` files in a folder as if they were in one `.hcl` file, just as Terraform does with `.tf` files, is the better trade-off.

_docs/rfc/imports.md
brikis98: Doh!

_docs/rfc/imports.md
brikis98: Doh!

jakauppila:
Looking at the suggestions, I'd be inclined to agree that `import` or `read_terragrunt_config` is the way way to go.

Our workaround for the for the differing inputs between configurations is to simply have two separate "top level" .hcl files that we `include` depending on the configuration.

@brikis98 Flattening the structure as you proposed in #759 seems counter-intuitive to our use-case as I described in https://github.com/gruntwork-io/terragrunt/issues/759#issuecomment-503769331

jfunnell:
Hm, I guess I can agree there is still possibly a need for imports even with a flattened structure, after seeing some examples.. Orthogonal enough to warrant implementing both, at least.

I think `read_terragrunt_config` is really simple and non-magical, and would compliment both the existing layout and a unified module approach.

_docs/rfc/imports.md
yorinasub17: I'll update to note that it is preferred to have scope of locals shared, but mention the possibility of having file scoped vars, its advantages and disadvantages.

_docs/rfc/imports.md
yorinasub17: Ah I just realized why you got confused. The bold font is exactly the same as `####`. Updated to use italics for the bold heading I was using previously: d1aa8f0

yorinasub17:
Thanks for the feedback everyone! I think based on the feedback and new thoughts around #759, it sounds to me like #759 is where we want to be heading. However, given how big of a change that really is, I am not sure that will get done anytime soon and, as @tomasbackman mentioned, there is a pressing need to solve this problem sooner rather than later.

Given that, I propose that we should take an incremental approach here, implementing the features that are low cost and partially solve the problem, and slowly work towards more complete solutions.

I think:

- Given the simplicity of `read_terragrunt_config`, it is probably a nice addition to have alongside any of the other implementations and thus we should implement this.
- `import` with `deep_merge` is probably a good short/medium term solution to have inbetween status quo and #759 .
- Although it is not decided yet, my guess is that we will eventually converge towards a version of #759 in the long term.

I'm going to start implementing `read_terragrunt_config`, but further feedback on that conclusion is very welcome!

brikis98:
Yea, the more I think about it, the more I believe the direction described in https://github.com/gruntwork-io/terragrunt/issues/759#issuecomment-585124357 is the way to go: turn Terragrunt into a _preprocessor_ for more or less pure Terraform code. So many of the features we've built into Terragrunt already exist in Terraform (e.g., locals, helper functions, dependencies, etc), so many others become unnecessary with pure Terraform code (e.g., code generation of `provider` / `backend` blocks), and it does a better job of paving the way for Terraform to adopt these features we all want natively.

But yes, it'll take some time to do that, so +1 on putting in some reasonable stop gap solutions for now.

yorinasub17:
UPDATE: I have updated the RFC to indicate that:

- This is now in development
- We will implement `read_terragrunt_config`, `import` blocks, and the single terragrunt.hcl file approach in that order.
- Added relevant PR and release links

yorinasub17:
@brikis98 any objections to merging this in now?

brikis98:
No objection, go for it! Thx Yori!

yorinasub17:
Thanks for sanity check! Going to merge it in now.

dudicoco:
@yorinasub17 is there a PR open for the ```import``` blocks feature?

yorinasub17:
Unfortunately, not yet.

xcrezd:
@yorinasub17 Any news on `import` feature?

arash-bizcover:
@yorinasub17 is there any way we can reference a hcl file from another git repo either with `read_terragrunt_config` or `import`
My use case is I don't want to keep some generic hcl files in the applications repos.

yorinasub17:
We currently don't have plans to implement remote references for import, but a PR to add support for that is welcome.

michelzanini:
The `import` idea is brilliant. It will help many use cases we have.

I am not quite sure yet about #759. I would prefer to keep using the hierarchical folders. If you do get to implement #759 at some point, please make sure we can still do hierarchical folders. Mixing it all up in one single file should be optional.

As mentioned by @arash-bizcover , a remote import would be a great addition. We have repeated top level `terragrunt.hcl` files for multiple repositories. We need to keep syncing them in different repos. If we could import them all from a single Git repo it would be really good. 

For example, consider something like this:

```
import "shared" {
  config_path = "git@github.com:my-company/shared-terragrunt.git/terragrunt.hcl?ref=master"
  merge = true
}

inputs = {
  
}
```

In this case `config_path` would support `git` references similar to the way `source` supports.

jfunnell:
> I am not quite sure yet about #759. I would prefer to keep using the hierarchical folders. If you do get to implement #759 at some point, please make sure we can still do hierarchical folders. Mixing it all up in one single file should be optional.

It's not one file, it's as many as you want. You just dont need a subfolder per module. Please read the whole proposal because most people, including myself, have propogated this misinformation

damon-atkins:
How  would you implement the following if the functions are deprecated
` key = "${basename(get_parent_terragrunt_dir())}/${path_relative_to_include()}/terraform.tfstate"`

yorinasub17:
> How would you implement the following if the functions are deprecated

I seem to recall a conversation in one of the issues regarding `path_relative_to_include` in the almost 1 year that passed since I wrote this (wow time flies) where this turns out to be a critical feature for many users and most people prefer the current approach vs. the more explicit approach proposed here, so we will most likely keep this around until there is a better replacement.

As for `get_parent_terragrunt_dir`, the intention is to have `get_terragrunt_dir` return that path. That is, in `import` mode, `get_terragrunt_dir` will always return the dir of the config you wrote it in, not the dir of the importing config. We currently aren't thinking about introducing a replacement for `get_importing_terragrunt_dir`, but if there are enough use cases for it, we will most likely add it in.

Matausi29 (commented):
-

digorgonzola:
Has the been any progress on this RFC?

Issue #936, opened 2019-10-30
Auto-init is not working


GauntletWizard:
In one of my repos, I recently added a new submodule to an existing module. Since then, `plan-all` no longer works for these directories. It appears that terragrunt is not detecting and running `terraform init`. After running `terraform init` in each of these directories, subsequent `plan-all` tasks work.

```
Katara:cardsavr-deploy ted$ terragrunt plan
[terragrunt] 2019/10/30 13:06:42 Reading Terragrunt config file at /Users/ted/code/switch/terraform/tenants/acmebank/cardsavr-deploy/terragrunt.hcl
[terragrunt] [/Users/ted/code/switch/terraform/tenants/acmebank/cardsavr-deploy] 2019/10/30 13:06:42 Running command: terraform --version
[terragrunt] 2019/10/30 13:06:43 Running command: terraform plan

Error: Module not installed

  on ../../../cardsavr/deploy/cat.tf line 94:
  94: module "cleanup_pcoss_jobs" {

This module is not yet installed. Run "terraform init" to install all modules
required by this configuration.


Error: Module not installed

  on ../../../cardsavr/deploy/cat.tf line 105:
 105: module "generate_vbs_dashboard" {

This module is not yet installed. Run "terraform init" to install all modules
required by this configuration.


Error: Module not installed

  on ../../../cardsavr/deploy/cat.tf line 116:
 116: module "autoscale_vbs" {

This module is not yet installed. Run "terraform init" to install all modules
required by this configuration.

[terragrunt] 2019/10/30 13:06:45 Hit multiple errors:
exit status 1
```

yorinasub17:
Are you using any of the following in the terragrunt `source`?

- Local references (e.g `source = ../module`)
- Remote references with no ref tag
- Remote references with a ref tag that is a branch

In each of these cases, Auto Init is known to not work correctly. At a high level, the underlying mechanism is checking for changes in the source tag to decide when to run `init`.

GauntletWizard:
Yes, `source = ../module` is in use. That makes sense. Are there any plans to remedy this, i.e. running `terraform get` before operations?

ajorpheus:
@yorinasub17, the cases you mentioned, in which Auto Init is **known** to not work (or even not 'always' work), is it worth documenting these on [Auto-Init](https://terragrunt.gruntwork.io/docs/features/auto-init/) documentation please?

At the moment, that page only has a generic note like so, which is helpful, but not helpful enough:

[...]

Issue #492, opened 2018-06-22
Add support for a generic xxx-all command


brikis98:
Right now, Terragrunt allows you to execute a few specific Terraform commands—`plan`, `apply`, `destroy`, `output`—on "all" submodules. However, we often want to add another `xxx-all` command to the mix, and doing so manually each time is hard to maintain.

Perhaps what we need instead is to add support for a `--terragrunt-all` flag that takes the normal Terraform command you're running and applies it to all submodules. For example:

```
terragrunt apply --terragrunt-all
```

This would be equivalent to the current `terragrunt apply-all`. However, you could also use it with commands that are not yet supported and any commands that come out in the future:

```
terragrunt  get --terragrunt-all
```

The only gotcha is we'd probably need a way to specify the dependency "order." That is, for `apply-all`, dependencies are processed in one direction, but for `destroy-all`, they go in the other order. So probably we'd want an optional `--terragrunt-all-reverse` flag:

```
terragrunt apply --terragrunt-all
terragrunt destroy --terragrunt-all-reverse
```

troylar:
@brikis98 Starting on this one  . . .

collinstevens:
@troylar did you get anywhere with this?

yorinasub17:
This is now implemented as `run-all` introduced in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.28.1

PR #1026, opened 2020-01-30
Update Auto-Init Documentation to add known issues

ajorpheus:
This PR addresses https://github.com/gruntwork-io/terragrunt/issues/936 by amending documentation to include 
known cases when auto-init fails to work.

### DIFF
--- a/docs/_docs/02_features/auto-init.md
+++ b/docs/_docs/02_features/auto-init.md
@@ -23,7 +23,12 @@ When Auto-Init is enabled (the default), terragrunt will automatically call [`te
 
 As [mentioned]({{site.baseurl}}/docs/features/keep-your-cli-flags-dry/#extra_arguments-for-init), `extra_arguments` can be configured to allow customization of the `terraform init` command.
 
-Note that there might be cases where terragrunt does not properly detect that `terraform init` needs be called. In this case, terraform would fail. Running `terragrunt init` again corrects this situation.
+Note that there might be cases where terragrunt does not properly detect that `terraform init` needs be called. Examples of such cases include the following in the terragrunt source:
+* Local references (e.g `source = ../module`)
+* Remote references with no ref tag
+* Remote references with a ref tag that is a branch
+
+In this case, terraform would fail. Running `terragrunt init` again corrects this situation.
 
 For some use cases, it might be desirable to disable Auto-Init. For example, if each user wants to specify a different `-plugin-dir` option to `terraform init` (and therefore it cannot be put in `extra_arguments`). To disable Auto-Init, use the `--terragrunt-no-auto-init` command line option or set the `TERRAGRUNT_AUTO_INIT` environment variable to `false`.
 


docs/_docs/02_features/auto-init.md
```diff
@@ -23,7 +23,12 @@ When Auto-Init is enabled (the default), terragrunt will automatically call [`te
 
 As [mentioned]({{site.baseurl}}/docs/features/keep-your-cli-flags-dry/#extra_arguments-for-init), `extra_arguments` can be configured to allow customization of the `terraform init` command.
 
-Note that there might be cases where terragrunt does not properly detect that `terraform init` needs be called. In this case, terraform would fail. Running `terragrunt init` again corrects this situation.
+Note that there might be cases where terragrunt does not properly detect that `terraform init` needs be called. Examples of such cases include the following in the terragrunt source:
+* Local references (e.g `source = ../module`)
+* Remote references with no ref tag
+* Remote references with a ref tag that is a branch
```
brikis98: Not sure I follow these. Are you saying that if you create a new `terragrunt.hcl` with the `source` references above, and run `terragrunt apply` *the first time*, it doesn't run auto init? 

Or that if you had *already* run `terragrunt apply` (or any other Terragrunt command) at least once and then changed something in the `source` module, it doesn't run auto init *again*?

ajorpheus: This references to running Terragrunt the very first time. I would expect it to create the remote state infrastructure (the S3 Bucket and the DynamoDB table) as a result of that.

I think something key to note here is the fact that this happens with `terragrunt plan-all` (at the root of the stack).
Executing `terragrunt plan` inside the stack module itself (e.g. `stack/vpc`) works fine and auto-init is triggered.

Here is a minimal repository which demonstrates this issue: https://github.com/ajorpheus/terragunt-auto-init-issue
Of particular interest might be the logs in execution-logs folder: https://github.com/ajorpheus/terragunt-auto-init-issue/tree/master/execution-logs which is structured like so:
```
execution-logs
├── plan <--------------------------- Works Fine  (Executed for the VPC sub-module)
│   ├── terraform.log
│   └── terragrunt-plan.log
|
|
├── plan-all <----------------------- Hangs (Executed at the level of the stack)
│   ├── terraform.log
│   └── terragrunt-plan-all.log
|
├── terraform-log-diff.html <-------- Diff of the above Terraform.log 			(see Link 1 below for direct link)
└── terragrunt-CLI-log-diff.html <--- Diff of the above Terragrunt CLI logs (see Link 2 below for direct link)
```

(notes:
Each of the terragrunt commands were run in the same shell after doing a:
	```
		export TF_LOG=TRACE && export TF_LOG_PATH=./terraform.log && export TG_LOG=debug && export TERRAGRUNT_DEBUG=true && 		TG_LOG_PATH=./tg.log
	```
)

I suspect this might be an issue with `plan-all`, but I do not understand enough of the terragrunt internals to figure out why. This may not be related to the type of source links in the terragrunt config at-all (in which case this PR is not related to https://github.com/gruntwork-io/terragrunt/issues/936 at all). I would've thought that if `plan` triggers auto-init for a sub-module, so should `plan-all`, since it would just invoke `plan` for all the sub-modules.


Link 1: https://cdn.statically.io/gh/ajorpheus/terragunt-auto-init-issue/master/execution-logs/terraform-log-diff.html
Link 2: https://cdn.statically.io/gh/ajorpheus/terragunt-auto-init-issue/master/execution-logs/terragrunt-CLI-log-diff.html

ajorpheus: I am tempted to suggest that until https://github.com/gruntwork-io/terragrunt/issues/388 and/or https://github.com/gruntwork-io/terragrunt/issues/492 are resolved, the documentation at https://davidbegin.github.io/terragrunt/use_cases/execute-terraform-commands-on-multiple-modules-at-once  and 
https://davidbegin.github.io/terragrunt/terragrunt_details/auto-init.html should be updated to note this 'Auto-init ain't gonna work with *-all' caveat  (especially the multi-modules documentation).

ajorpheus: I am tempted to suggest that until https://github.com/gruntwork-io/terragrunt/issues/388 and/or https://github.com/gruntwork-io/terragrunt/issues/492 are resolved, the documentation at:

1. https://davidbegin.github.io/terragrunt/use_cases/execute-terraform-commands-on-multiple-modules-at-once  and,
2. https://davidbegin.github.io/terragrunt/terragrunt_details/auto-init.html 

should be updated to note this 'Auto-init does not work with *-all' caveat (especially the multi-modules documentation). I'd be happy to do this if you agree.

yorinasub17: > Or that if you had already run terragrunt apply (or any other Terragrunt command) at least once and then changed something in the source module, it doesn't run auto init again?

FWIW, since I was the one to list these examples, I just wanted to mention that I have found that the code implementation suggest that this is what will happen in those three scenarios.

ajorpheus: @brikis98, may I please hear back on what you think about the documentation changes?

I can appreciate that the 2 use cases below may not be common:
1. That one wishes to run Terragrunt *-all commands on a set of TF modules which haven't yet been inited
2. That one wishes expects auto-init to happen if the source links are changed (as @yorinasub17 states)

However, till these are addressed, would it please be possible to note them as caveats in the documentation pages linked in my previous comment? 

I was trying to enforce the use of Terragrunt to my team of budding TF developers as a 'best practice'. However, while creating documentation around using Terragrunt for a specific project, I came across this issue. I believe that manually running a command like `find . -not -path '*/.terragrunt-cache/*' -type f -name terragrunt.hcl -execdir terragrunt init \;` is not a deal-breaker (far from it), but the experience of new terragrunt users would surely be improved if caveats that affect the major use-cases are noted down in the documentation.

brikis98: If `plan-all` fails to `init` on the very first run, that's a bug, and not one that I'm aware of. As you said, it *should* be equivalent to running `init` in each folder individually. Are you positive that when you try this, none of the folders already have a `.terrarunt-cache`?

If you change a `source` from a URL to a local path, or to a different URL, or to a different local path, then Terragrunt should re-run auto init. If not, again, it's a bug that should be fixed.

If you have `source` pointed to a local file path, then it should re-copy the code on every run. However, it won't re-run the other aspects of `init` (e.g., updating backend config). 

*Both* `plan` and `plan-all` will skip auto-init if you're using a URL in `source`, and the contents behind the URL change. E.g., If you point `source` at the `master` branch of some URL, run `plan`, then and push new code to the branch, and run `plan` again, Terragrunt will not pull the latest code down on the second `plan`. I thought this was documented, but if not, we definitely should.

ajorpheus: > If plan-all fails to init on the very first run, that's a bug, and not one that I'm aware of. As you said, it should be equivalent to running init in each folder individually. Are you positive that when you try this, none of the folders already have a .terrarunt-cache?

Yes, I am positive that none of the folders had the cache folder since I was cleaning them up just to be sure (by doing a `gfind . -regextype posix-extended  -regex '.*(.terragrunt-cache|.terraform)' -prune -type d -exec rm -rf {} \;` in the root of the 'stack').


> If you change a source from a URL to a local path, or to a different URL, or to a different local path, then Terragrunt should re-run auto init. If not, again, it's a bug that should be fixed.

It appears this is a different bug indeed. After I first experienced this issue, I thought it was due to my usage of local file URLs. However, changing them to Git URLs did not result in auto-init being executed either.


I am glad that auto-init not working the first time is to be considered a bug rather than a non-trivial 'generic *-all command' feature

levkohimins:
Hi @ajorpheus,

Thank you for this repo https://github.com/ajorpheus/terragunt-auto-init-issue! I just tried to reproduce the issue with your example, but I didn't succeed. I run in many different ways in the hope that it will show me the mentioned error, but no, every time terragrunt runs `terraform init`. I assume that the issue is somehow resolved. Could you please confirm or deny this?

levkohimins:
I close the PR as it is no longer relevant.

PR #1036, opened 2020-02-06
Restore before and after hooks feature page, 

yorinasub17:
This puts back in the before and after hooks feature page that was naively removed due to it being redundant with reference style docs. That broke the front page link, and it didn't make sense to go from the front page to the terraform block docs, so this adds back in the before and after hook feature page, but instead of being a reference, it shows an example of how it works.


yorinasub17:
Thanks for review! Merging this in

PR #1047, opened 2020-02-13
Fix support link

brikis98:


brikis98:
Given the simple nature of the fix, I'm going to merge it right away. Feedback welcome!

PR #1049, opened 2020-02-13
Skip init on validate

### LINKED ISSUE #1039: Terragrunt validate prompts for backend creation
When running terragrunt validate I expect the command to be fully self contained. However, it presently seems to still check for the existence and access to the backend remote state.

e.g.

$ terragrunt validate
...
Remote state GCS bucket REDACTED does not exist or you don't have permissions to access it. Would you like Terragrunt to create it? (y/n)

Can we bypass remote state creation in validate?

umairidris:
Fix #1039

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -82,12 +82,12 @@ var TERRAFORM_COMMANDS_THAT_USE_STATE = []string{
 	"show",
 	"taint",
 	"untaint",
-	"validate",
 	"force-unlock",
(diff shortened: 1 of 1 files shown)


umairidris:
On further testing, it looks like if you have modules you do need to run terraform init before terraform validate.. and terraform init will check the backend GCS if it is set. So I don't think this is possible. I will close the PR and issue for now and if I have any other solution will report back. Thanks!

PR #1050, opened 2020-02-14
Codegen

yorinasub17:
This implements the `generate` block that we discussed internally. I also took the liberty to add in a `generate` attribute to the `remote_state` block for generating the `config` of `remote_state`. Doing so allows you to generate the remote state config while taking advantage of the auto init functionality to create the S3/GCS bucket automatically.


brikis98:
> I also took the liberty to add in a generate attribute to the remote_state block for generating the config of remote_state. Doing so allows you to generate the remote state config while taking advantage of the auto init functionality to create the S3/GCS bucket automatically.

I have nothing against it, but I'm curious, what's the advantage of generated code over CLI args for the remote state config?

codegen/generate.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	targetFileExists := util.FileExists(targetPath)
+	if targetFileExists && config.IfExists == ExistsError {
+		return errors.WithStackTrace(GenerateFileExistsError{path: targetPath})
+	} else if targetFileExists && config.IfExists == ExistsSkip {
+		// Do nothing since file exists and skip was configured
+		logger.Printf("The file path %s already exists and if_exists for code generation set to \"skip\". Will not regenerate file.", targetPath)
+		return nil
+	} else if targetFileExists {
```
brikis98: NIT: technically, the `GenerateConfigExists` enum has another state: `ExistsUnknown`. In theory, the code should never get here with that value, but perhaps return an error if it does?

docs/_docs/02_features/keep-your-remote-state-configuration-dry.md
```diff
@@ -135,6 +137,58 @@ The child `.hcl` file’s `terraform` settings will be merged into the parent fi
 
 Other settings in the child `.hcl` file override the respective settings in the parent.
 
+
+### Using the generate property to generate terraform code for managing remote state
+
+While the default way terragrunt manages remote state is through `terraform init` with `-backend-config`, you can also
+use the `generate` property to configure terragrunt to geneerate a `.tf` file in the terraform working directory with the
```
brikis98:
```suggestion
use the `generate` property to configure terragrunt to generate a `.tf` file in the terraform working directory with the
```

docs/_docs/02_features/keep-your-remote-state-configuration-dry.md
```diff
@@ -135,6 +137,58 @@ The child `.hcl` file’s `terraform` settings will be merged into the parent fi
...
+backend configuration.
+
+The `generate` property is an object that accepts two parameters:
+
+- `path`: The path where the generated file should be written. If a relative path, it'll be relative to the Terragrunt
```
brikis98: Does this have a default? E.g., `backend.tf`?

docs/_docs/02_features/keep-your-remote-state-configuration-dry.md
```diff
@@ -135,6 +137,58 @@ The child `.hcl` file’s `terraform` settings will be merged into the parent fi
...
+  working dir (where the terraform code lives).
+- `if_exists`: What to do if a file already exists at `path`. Valid values are: `overwrite` (overwrite the existing
+  file), `skip` (skip code generation and leave the existing file as-is), `error` (exit with an error).
+
+For example, here is a version of the root `remote_state` configuration with the `generate` property:
+
+```hcl
+remote_state {
+  backend = "s3"
+  generate = {
```
brikis98: Does this code generation happen every time you run Terragrunt? Or only during Auto Init?

docs/_docs/02_features/keep-your-remote-state-configuration-dry.md
```diff
@@ -135,6 +137,58 @@ The child `.hcl` file’s `terraform` settings will be merged into the parent fi
...
+    path      = "backend.tf"
+    if_exists = "overwrite"
```
brikis98: I just had one realizationwith the `generate` helper, both in `remote_state` and in general, in terms of how the `if_exists` logic works. Let's say you are generating a `backend.tf` into some `.terragrunt-cache/xxx` folder. There are three ways a `backend.tf` file could already exist in that folder:

1. It's part of the code you downloaded from the `source` URL.
1. It's one of the files in the working directory where you ran `terragrunt apply`, and Terragrunt copied it from that folder into the `.terragrunt-cache/xxx` folder.
1. You generated `backend.tf` on a previous run.

Some important questions about these three cases:

1. Does code generation happen before or after case (1)?
1. Does code generation happen before or after case (2)?
1. Should `if_exists` provide a way to differentiate between case (3), which will happen all the time and is normal, versus cases (1) and (2) which are more rare and probably indicative of a problem?

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -401,6 +413,41 @@ dependencies {
...
+- `contents` (attribute): The contents of the generated file.
+
+Example:
+
+```hcl
+# When using this terragrunt config, terragrunt will generate the file "provider.tf" with the aws provider block before
+# calling to terraform. Note that this will overwrite the `provider.tf` file if it already exists.
+generate "provider" {
+  path      = "provider.tf"
+  if_exists = "skip"
```
brikis98: Comment above says we will overwrite, but code example sets `skip`

docs/_docs/02_features/keep-your-remote-state-configuration-dry.md
yorinasub17: No. I think it should be forced to be set given that `backend.tf` feels like a fairly common filename.

docs/_docs/02_features/keep-your-remote-state-configuration-dry.md
yorinasub17: Every time. See the commentary right below this, where it says:

> With this configuration, `terragrunt` will generate a new file `backend.tf` in the working directory before it calls out to any `terraform` command with the following contents:

codegen/generate.go
yorinasub17: Good idea. Done: f9d88fa

docs/_docs/04_reference/config-blocks-and-attributes.md
yorinasub17: Good catch! Fixed: 667de58

docs/_docs/02_features/keep-your-remote-state-configuration-dry.md
yorinasub17: Code generation happens after case (2). There is a test case that checks this behavior by having a `backend.tf` file in the terragrunt working dir and setting a `generate` block having `if_exists = "skip"`. The test then makes sure it doesn't overwrite the `backend.tf` file. Similarly, there is a case that has a similar set up with `if_exists = "overwrite"` to make sure it overwrites the `backend.tf`.

> Should if_exists provide a way to differentiate between case (3), which will happen all the time and is normal, versus cases (1) and (2) which are more rare and probably indicative of a problem?

This is an interesting question. I think it should, but I can't come up with a clean way to implement this. The only way where I can think of a stable way to implement this is to have a comment at the top of the file when Terragrunt generates the code that can be used as a signature. If the file contains that, then we know it is a terragrunt generated file.

Let me implement that for now.

yorinasub17: Done: a6e8f80

yorinasub17:
> Add it to https://terragrunt.gruntwork.io/docs/features/keep-your-terraform-code-dry/.

Done: 58c5ea6 

> Add it to https://terragrunt.gruntwork.io/docs/getting-started/quick-start/

Done: cb79d17 

> I have nothing against it, but I'm curious, what's the advantage of generated code over CLI args for the remote state config?

I don't think there is a huge difference here. I could only come up with 2 advantages, and both of them are fairly weak:

- Since the config isn't set through CLI args, they don't show up in the logs (previously you can see all the config in the logs because terragrunt will log all the args).
- Avoid potential bugs due to quoting issues with json.

yorinasub17:
I believe I addressed all the comments, but given the amount of new stuff added, I rerequested a review.

docs/_docs/02_features/keep-your-remote-state-configuration-dry.md
brikis98: Roger

docs/_docs/02_features/keep-your-remote-state-configuration-dry.md
brikis98: Ah, nice work with the comment. That's useful too so that users can easily tell it's generated code.

brikis98:
>I don't think there is a huge difference here. I could only come up with 2 advantages, and both of them are fairly weak:

I just realized one more... You no longer need the empty `terraform { backend "xxx" { } }` block in every one of your modules! That seems very valuable to me. We may even want to update all of our docs to recommend using this `generate` feature and not specifying the `backend` block manually.

yorinasub17:
> We may even want to update all of our docs to recommend using this generate feature and not specifying the backend block manually.

This is probably a good idea, but let's do this in a separate PR so as not to hold this feature up. Going to merge this in and release now.

yorinasub17:
Filed https://github.com/gruntwork-io/terragrunt/issues/1054 so we don't forget.

sleungcy:
👏 I needed this! Thanks Yori 👍 👍 👍

arashilmg:
Thanks @yorinasub17 generate is f awesome.
How can I disable the first signature line ?
I'm generating a json file and cannot have any comments inside it.

yorinasub17:
> How can I disable the first signature line ?

Ah that's a feature we currently don't have, but an option (something like `disable_signature`?) is probably a good idea. A PR to add that feature would be much appreciated!

lorengordon:
@arashilmg @yorinasub17 I just ran into the same problem, opened an issue: https://github.com/gruntwork-io/terragrunt/issues/1121

arash-bizcover:
as a quick dirty workaround I did remove the line with Terraform `replace` function while reading the content of the file, like this:

`
replace(templatefile(GENERATEDFILE.json,{}),"/^#.*rragrunt.*\n/","")
`

_I'm using templatefile as I populate some var in my json file, it might not be needed_

Issue #1041, opened 2020-02-10
How to exclude dependencies?


dudicoco:
Hi,

I am running the following command:
```
terragrunt apply-all --terragrunt-include-dir devops/amazing-app/k8s/
```
In the ```terragrunt.hcl``` file there is a dependency block defined for another component.

When running the above command the component specified in the dependency block is being installed, I would like to install only the components specified in ```--terragrunt-include-dir``` without their dependencies.

Is there a way to achieve this?

Thanks

yorinasub17:
Is `devops/amazing-app/k8s` a single module, or a folder with multiple modules?

- If it is a single module, you should use `terragrunt apply --terragrunt-config devops/amazing-app/k8s/terragrunt.hcl` instead. This has a few advantages, like skipping the check to collect all the modules to run on and bypassing terragrunt's multi-module functionality to make it virtually impossible to have terragrunt run multiple modules.
- If it is a folder with multiple modules, you can use `--terragrunt-ignore-external-dependencies`

dudicoco:
@yorinasub17, ```devops/amazing-app/k8s``` contains a single ```terragrunt.hcl``` file with a reference to a single module.

What I would like to do, is to run ```terragrunt apply-all --terragrunt-include-dir devops/app1/k8s/ --terragrunt-include-dir devops/app2/k8s/``` and have terrgrunt to install only the components specified in ```--terragrunt-include-dir``` (app1 and app2) and install them according in the order of dependencies between them. I don't want terragrunt to install dependencies specified within these components which I have not included with ```--terragrunt-include-dir```.

Please let me know if it makes sense.
Thanks!

yorinasub17:
`--terragrunt-ignore-external-dependencies` is probably the flag you are looking for: https://terragrunt.gruntwork.io/docs/reference/cli-options/#terragrunt-ignore-external-dependencies

dudicoco:
@yorinasub17 I've already tried using this flag and it did nothing.
I assume it's meant for ignoring the dependencies of the dependencies.

yorinasub17:
That's probably a bug then. Can you share the following so we can take a look?

- Your folder structure
- `terragrunt.hcl` of `devops/amazing-app/k8s`
- `terragrunt.hcl` of a dependency being pulled in
- The FULL output of the logs from a `terragrunt` call.

dudicoco:
So it seems that the behaviour is irregular.
[...]

PR #1062, opened 2020-02-20
Make include dirs consistent when dependencies are involved

yorinasub17:
This PR addresses the issue reported in https://github.com/gruntwork-io/terragrunt/issues/1041.

Specifically:

- There was a bug in the behavior of `include-dir` where the order in which modules were added to the stack changed if dependencies would be included or not. The main issue stemmed from the fact that we immediately included dependencies when processing included directories. The included dependency could be undone if that module appeared later in the for loop and it resided outside the included dirs. This PR fixes that by handling dependencies in a second loop.

- Add a new flag `--terragrunt-strict-include` which will ignore dependencies when processing include dir.


yorinasub17:
Thanks for review! Merging and releasing.

PR #1064, opened 2020-02-24
Made sure external_id and session_name are respected when s3 backend needs assuming role

kujon:
Certain roles can be set up such that an additional `external_id` parameter is required, before assuming role can go ahead. Terraforms S3 backend [supports this functionality](https://www.terraform.io/docs/backends/types/s3.html#external_id). Terragrunt seems to ignore it.

### DIFF
--- a/aws_helper/config.go
+++ b/aws_helper/config.go
@@ -22,6 +22,8 @@ type AwsSessionConfig struct {
 	CredsFilename           string
 	S3ForcePathStyle        bool
 	DisableComputeChecksums bool
+	ExternalID              string
+	SessionName             string
 }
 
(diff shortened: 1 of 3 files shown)


kujon:
Alternatively, maybe it would be good to replace the remote_state_s3.go with this https://github.com/hashicorp/terraform/blob/master/backend/remote-state/s3/backend.go?

kujon:
sure thing!

kujon:
Done!

brikis98:
Tests passed. Merging now.

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.22.5

kujon:
Cheers!

Issue #386, opened 2017-12-01
apply-all does not set auto-approve parameter


dewang:
As of Terraform v0.11, the default behaviour for `terraform apply` is to show the plan and wait for approval. When using `apply-all` this results in the script waiting for input for every module that you want to deploy.

brikis98:
Well, `apply-all` just runs `apply` in every directory, so it inherits whatever behavior Terraform has for `apply`. If you want it to be fully non-interactive, can't you just run `terragrunt apply-all -auto-approve`?

dewang:
Interesting, I didn't think the parameter would be passed through. The stack sets `-input=false` https://github.com/gruntwork-io/terragrunt/blob/bc8bae6f1c771f5a5a38953a4bc041521c984940/configstack/stack.go#L73 so I assumed the `-auto-approve` would have to be passed the same way.
I can't imagine that anyone using `apply-all` would _not_ want it to be fully non-interactive.

Having said all that, I have to say a massive thanks for terragrunt and also your blog posts. They've really helped me get set up with terraform.

brikis98:
> Interesting, I didn't think the parameter would be passed through. The stack sets -input=false

I believe the `setTerraformCommand` method [appends your CLI args to the passed-in command](https://github.com/gruntwork-io/terragrunt/blob/bc8bae6f1c771f5a5a38953a4bc041521c984940/configstack/stack.go#L114-L118), so it *should* work. Try it out and let me know!

> Having said all that, I have to say a massive thanks for terragrunt and also your blog posts. They've really helped me get set up with terraform.

Thanks :)

ebarault:
Just faced the same problem and tested vanilla terraform option `-auto-approve` : it does work 👍 

That being said it really make the `--terragrunt-non-interactive` flag confusing since terraform v0.11.

Shouldn't the '-auto-approve` as well as `-force` options be added as well when using `--terragrung-non-interactive` ?

brikis98:
> Shouldn't the -auto-approve as well as -force options be added as well when using --terragrung-non-interactive ?

I could go either way on this one, but *probably* setting `-auto-approve` to `true` would be following "principle of least surprise."

What is `-force`?

ebarault:
`-force`: [cli option for **destroy**](https://www.terraform.io/docs/commands/destroy.html)

"If -force is set, then the destroy confirmation will not be shown"

brikis98:
[...]

PR #1065, opened 2020-02-25
Don’t auto add terraform args to apply-all

joshmyers:
#### what

When using apply-all, and setting plan files to be used as 
extra_arguments, these arguments are automatically being added to our TF
command, which we could add manually if we wanted via extra_args etc.
Because these args are automatically added and using a planfile via 
extra_args for apply ends up like[1] when using apply-all, which fails.

[1]
```
terraform apply tfplan -input=false -auto-approve
```

#### tests

https://gist.github.com/joshmyers/72408e85a90c5ac9686c80a9c91c39f1

Some failing tests but they seem to be failing on master too...

#### Related

https://github.com/gruntwork-io/terragrunt/issues/386

### DIFF
--- a/configstack/stack.go
+++ b/configstack/stack.go
@@ -5,11 +5,12 @@ import (
 	"fmt"
 	"strings"
 
+	"sort"
+
 	"github.com/gruntwork-io/terragrunt/config"
 	"github.com/gruntwork-io/terragrunt/errors"
 	"github.com/gruntwork-io/terragrunt/options"
 	"github.com/gruntwork-io/terragrunt/util"
-	"sort"
 )
 
 // Represents a stack of Terraform modules (i.e. folders with Terraform templates) that you can "spin up" or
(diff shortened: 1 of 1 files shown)


joshmyers:
@brikis98 I get that this breaks backwards compatibility, but given Terragrunt gives users several ways to set vars via a mechanism, magically auto adding feels unexpected, especially ones like `-auto-approve`.

yorinasub17:
Note that because terragrunt runs the underlying terraform calls in parallel when running `xxx-all`, the stdin is shared across all the runs. So when you don't do `-auto-approve`, you won't actually know which one you are approving everytime you get a prompt because you would be approving all the ones that has shown up to the point of submitting your answer.

For this to work, you either need to:

- Force concurrency to 1 when in manual approval mode.
- Fix the way terragrunt handles interactive prompts by micromanaging the outputs and showing each prompt with the relevant context one at a time.

steve-a-jones:
My team is also looking for this functionality  --  @joshmyers is this still being worked on given the latest response from @yorinasub17  ?

joshmyers:
@steve-a-jones Nope I dropped pursing this as it felt like a janky workflow. stdout is also shared across runs.

PR #1076, opened 2020-03-02
Update find_in_parent_folders to return absolute path

brikis98:
Currently, the `find_in_parent_folders` function returns a relative path to the file it finds. This PR updates it to return the absolute path to that file. Depending on where you are using `find_in_parent_folders`, this can be important: e.g., one place I saw the relative path cause an issue is:

1. I had a child `terragrunt.hcl` that used `include` on a parent `terragrunt.hcl`.
1. The parent `terragrunt.hcl` called `read_terragrunt_config(find_in_parent_folder("foo.hcl"))`
1. `foo.hcl` called `read_terragrunt_config(find_in_parent_folder("bar.hcl"))`

This led to an error where Terragrunt said `read_terragrunt_config` could not find `../../bar.hcl`, even though that file existed. I believe the error was because we were returning a relative path for it, but with `.terragrunt-cache/xxx/yyy` as the working dir, that ended up being the wrong path. The change in this PR to an absolute path fixed the issue.

I'm not aware of any case where a relative path would have an advantage over an absolute path, so this is a backwards incompatible change to always return an absolute path.

### DIFF
--- a/config/config_helpers.go
+++ b/config/config_helpers.go
@@ -259,7 +259,7 @@ func findInParentFolders(params []string, include *IncludeConfig, terragruntOpti
 		}
 
 		if util.FileExists(fileToFind) {
-			return util.GetPathRelativeTo(fileToFind, filepath.Dir(terragruntOptions.TerragruntConfigPath))
+			return fileToFind, nil
 		}
 
 		previousDir = currentDir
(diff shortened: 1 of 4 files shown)


brikis98:

madsonic:
this has to be updated too. `docs/_docs/02_features/keep-your-remote-state-configuration-dry.md:108`

zackproser:
Thanks @madsonic! I made a PR to update that doc as well.

dexterlakin:
Is there a way to conditionally set this?

We have a CICD pipeline that needs to support a wide range of terragrunt versions.

Issue #1079, opened 2020-03-04
Panic due to api body size. 


bpastiu:
Hello,

We've encountered an interesting issue while deploying one of our stacks.
I can't go into specifics of the configuration for obvious reasons but I'll try to explain the best I can.

We are trying to deploy a aws_api_gateway_rest_api resource with the body being a rendered openapi template file.
We didn't experience any issues up until now when there were some changes done to the openapi file.
The openapi file is now about 3500 lines big and is causing the panic. We tested that this is the issue using a smaller openapi file and it worked fine.

I would like to as if there is any limitation to this we should take into account going forward and how we could get it resolved.

Added a TF_LOG=trace to extract some logs. It happens here when it tries to create the tf state file for this module in s3. Every module has it's own tfstate file.
In the respective bucket the api-gateway key is missing in entirety.

[terragrunt] [/config/account/region/environment/api-gateway] 2020/03/04 16:30:12 Running command: terraform init -backend-config=bucket=bucket-name -backend-config=dynamodb_table=dynamo-db-table -backend-config=encrypt=true -backend-config=key=12345/environment/api-gateway/terraform.tfstate -backend-config=region=region

Here is the panig message:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xd37d71]

goroutine 53 [running]:
github.com/gruntwork-io/terragrunt/cli.runTerraformWithRetry(0xc0000db600, 0xc0000db600, 0xc0002f8900)
/go/src/github.com/gruntwork-io/terragrunt/cli/cli_app.go:531 +0x121
github.com/gruntwork-io/terragrunt/cli.runTerragruntWithConfig.func1(0x0, 0x0)
/go/src/github.com/gruntwork-io/terragrunt/cli/cli_app.go:484 +0x2a
github.com/gruntwork-io/terragrunt/cli.runActionWithHooks(0xf9a250, 0x9, 0xc0000db600, 0xc0002a8140, 0xc00043bc20, 0xc0004ee500, 0xc00060a140)
/go/src/github.com/gruntwork-io/terragrunt/cli/cli_app.go:495 +0x2ae
github.com/gruntwork-io/terragrunt/cli.runTerragruntWithConfig(0xc0000db600, 0xc0002a8140, 0x0, 0x0, 0x0)
/go/src/github.com/gruntwork-io/terragrunt/cli/cli_app.go:483 +0x2c7
[...]

PR #1083, opened 2020-03-07
Fix source of panic

yorinasub17:
This fixes the cause of the panic in https://github.com/gruntwork-io/terragrunt/issues/1079, where a catastrophic exit from `terraform` causes the shell command to return the error without output, but `runTerraformWithRetry` was expecting an output always.


cli/cli_app.go
```diff
@@ -528,6 +528,9 @@ func runTerraformWithRetry(terragruntOptions *options.TerragruntOptions) error {
 	// Retry the command configurable time with sleep in between
 	for i := 0; i < terragruntOptions.MaxRetryAttempts; i++ {
 		if out, tferr := shell.RunTerraformCommandWithOutput(terragruntOptions, terragruntOptions.TerraformCliArgs...); tferr != nil {
+			if out == nil {
+				return tferr
+			}
```
brikis98: NIT: might be cleaner below to say `if out != nil && isRetryable(...)`

yorinasub17: Ah good point! Fixed: 76bfb0c

yorinasub17:
Ok build passed on the update, so going to merge this. Thanks for the review!

PR #1097, opened 2020-03-23
Fix 404s

bwhaley:
A few links in this document led to 404s. Also fixed a small typo.


bwhaley:
Thanks for reviewing. ❤️

Issue #1091, opened 2020-03-13
Allow running hclfmt on a single file


yorinasub17:
Right now, if you are trying to format a single hcl file, you will end up formatting the entire tree including subdirectories. There should be a CLI option `--terragrunt-hclfmt-file` which when provided, will only run hclfmt on that file.

yorinasub17:
Released in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.23.3

PR #1104, opened 2020-03-26
Hclfmt specific file

EDsCODE:
Implements #1091 as @yorinasub17 suggested!
 
I added cli functionality for when hclfmt command is run if the `--terragrunt-hclfmt-file` is present with a specified file (e.g. `terragrunt hclfmt --terragrunt-hclfmt-file /path/to/terragrunt.hcl`)

Gist result of the two new tests hclfmtfile and hclfmtfileerrors [here](https://gist.github.com/EDsCODE/76e88fce4ad92f4dd76b299d6dbcef3c)

**Edit:** I didn't put up an RFC for this as I added the option in a pretty similar fashion as the other options. But as a short descriptor: if hclfmt is run with the option and a specific file is provided the `runhclfmt` command will only match and add the specific file to the array of files to process rather than the entire tree. If the option is not provided, then the entire tree of subdirectories will be added as previously implemented.


cli/hclfmt_test.go
```diff
@@ -191,3 +191,84 @@ func TestHCLFmtCheckErrors(t *testing.T) {
 		})
 	}
 }
+
+func TestHCLFmtFile(t *testing.T) {
```
yorinasub17: Is there a reason you created new fixtures for this instead of reusing the existing ones (e.g., https://github.com/gruntwork-io/terragrunt/tree/master/test/fixture-hclfmt-check)?

yorinasub17: Also, these are nice tests to be comprehensive, but since we already have comprehensive testing of the functionality of `hclfmt`, it kind of overlaps with the other tests. The real test we need here is one that checks that the `--terragrunt-hclfmt-file` option only touches the file and not recursively hit all the subdirectories.

So I think it is sufficient to have a single test where `HclFile` is set to `fixture-hclfmt-check/a/terragrunt.hcl`, and the root (or `TerragruntConfigPath`) is set to `fixture-hclfmt-check/terragrunt.hcl`, and checks that only the set file was formatted the same as `fixture-hclfmt-check/expected.hcl`. It should check to make sure that `fixture-hclfmt-check/terragrunt.hcl` and `fixture-hclfmt-check/b/c/terragrunt.hcl` were untouched.

What do you think?

cli/hclfmt.go
```diff
@@ -19,10 +19,18 @@ import (
...
 	workingDir := terragruntOptions.WorkingDir
-	tgHclFiles, err := zglob.Glob(util.JoinPath(workingDir, "**", "*.hcl"))
+	targetFile := terragruntOptions.HclFile
+	if targetFile != "" {
+		terragruntOptions.Logger.Printf("Formatting terragrunt.hcl file at: %s.", targetFile)
+		tgHclFiles, err = zglob.Glob(util.JoinPath(workingDir, targetFile))
+	} else {
+		terragruntOptions.Logger.Printf("Formatting terragrunt.hcl files from the directory tree %s.", terragruntOptions.WorkingDir)
+		tgHclFiles, err = zglob.Glob(util.JoinPath(workingDir, "**", "*.hcl"))
+	}
```
yorinasub17: I think this whole routine can be simplified to:

```go
targetFile := terragruntOptions.HclFile
if targetFile != "" {
    terragruntOptions.Logger.Printf("Formatting terragrunt.hcl file at: %s.", targetFile)
    return formatTgHCL(terragruntOptions, targetFile)
}

// what the function was before
```

EDsCODE: mmm yes—i did have to add a line to format the entire path (ensuring working directory is appended to the specified file path). at the moment it's setup so that it's expecting the file's relative path to the working directory. i was planning on specifying this in the docs or change it to require an absolute path

yorinasub17: Ah you can use https://golang.org/pkg/path/filepath/#IsAbs to check if it is absolute first.

cli/hclfmt_test.go
EDsCODE: > Is there a reason you created new fixtures for this instead of reusing the existing ones (e.g., https://github.com/gruntwork-io/terragrunt/tree/master/test/fixture-hclfmt-check)?

i definitely could've clarified before setting up the fixtures but I was under the impression we wanted very clear separation of concerns especially since there are redundancies between the hclfmt and hclfmt-check-errors fixtures.

In any case, I think your testing makes more sense and I'll give it a go!

EDsCODE: i just made the changes and targeted the test at `fixture-hclfmt` instead because `fixture-hclfmt-check` has hcl files that are all identical. otherwise, I did as suggested! it runs the formatting on the hcl file in a and then checks the other two hcl files with the contents of the original file at `fixture-hclfmt/terragrunt.hcl`

yorinasub17:
Build passed, so will merge and release. Thanks for your contribution!

Issue #1092, opened 2020-03-13
Upgrade hcl formatter


yorinasub17:
Right now the hcl formatter is using an old version and thus does not handle formatting nested maps and HEREDOC syntax correctly.

yorinasub17:
Done in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.23.4

PR #1105, opened 2020-03-28
Upgrade hclfmt (+alpha: see description)

yorinasub17:
This addresses https://github.com/gruntwork-io/terragrunt/issues/1092, where the hcl2 library was too old, and thus it did not include the fix for handling heredoc syntax correctly.

However, in order to update to the latest hcl2 version, we needed to switch to `hashicorp/hcl/v2` as the `hcl2` repo is now archived and merged into the mainstream `hcl` repo. The problem was that this did not work too well with `dep`, and so I also took this chance to update to using go modules.

Note that in the process, all the packages were updated to the latest version that supports all the imports used throughout this repo (this is done automatically when you do `go mod tidy` without explicit version locking). This includes:

- `terraform`: 0.12.9 => 0.12.24
- `aws-sdk-go`: 1.25.4 => 1.29.34
- `go-getter`: 1.3.0 => 1.4.2


cli/download_source_test.go
```diff
@@ -244,10 +244,10 @@ func TestSplitSourceUrl(t *testing.T) {
...
-		{"parent-url-multiple-children-no-double-slash", "ssh://git@github.com:foo/modules.git/foo/bar/baz/blah", "ssh://git@github.com:foo/modules.git/foo/bar/baz/blah", ""},
-		{"parent-url-one-child-with-double-slash", "ssh://git@github.com:foo/modules.git//foo", "ssh://git@github.com:foo/modules.git", "foo"},
-		{"parent-url-multiple-children-with-double-slash", "ssh://git@github.com:foo/modules.git//foo/bar/baz/blah", "ssh://git@github.com:foo/modules.git", "foo/bar/baz/blah"},
+		{"parent-url-one-child-no-double-slash", "ssh://git@github.com/foo/modules.git/foo", "ssh://git@github.com/foo/modules.git/foo", ""},
+		{"parent-url-multiple-children-no-double-slash", "ssh://git@github.com/foo/modules.git/foo/bar/baz/blah", "ssh://git@github.com/foo/modules.git/foo/bar/baz/blah", ""},
+		{"parent-url-one-child-with-double-slash", "ssh://git@github.com/foo/modules.git//foo", "ssh://git@github.com/foo/modules.git", "foo"},
+		{"parent-url-multiple-children-with-double-slash", "ssh://git@github.com/foo/modules.git//foo/bar/baz/blah", "ssh://git@github.com/foo/modules.git", "foo/bar/baz/blah"},
```
yorinasub17: This is due to a change in go 1.12 that makes `ssh://git@github.com:foo/modules.git` be interpreted as `foo` for the port number.

yorinasub17:
Thanks for review!

> BTW, please check all our docs for mention of dep and update those docs accordingly!

Ah good catch! Just updated: 949b1b8 

Will merge and release!

Issue #1116, opened 2020-04-05
Graph representation of the terragrunt dependencies similar to terraform graph


mauriciopoppe:
It would be great to have a visual representation similar to `terraform graph` but with the terragrunt dependencies, something like:

```sh
terragrunt plan-all --terragrunt-graph | dot -Tsvg > graph.svg
```

This could be applied to the *-all commands

Related https://github.com/gruntwork-io/terragrunt/issues/1093

yorinasub17:
Released as https://github.com/gruntwork-io/terragrunt/releases/tag/v0.23.7. Binaries will show up shortly

PR #1117, opened 2020-04-05
Graph representation of the terragrunt dependencies similar to terraform graph

mauriciopoppe:
Fixes https://github.com/gruntwork-io/terragrunt/issues/1116

Implementation details:

- There's a new boolean param `--terragrunt-graph` that's only used in plan-all for now, when it's set the stack is printed in the graphviz language to stdout
- I created a new integration test that's a clone of `fixture-output-all`

## Usage

```sh
$ pwd
test/fixture-plan-all-graph/env1

$ terragrunt plan-all --terragrunt-graph | dot -Tjpg > graph.jpg
```

## Output

graph

## Pending items

- Path resolution to folders outside the working dir
- Display style for ignored modules
- Early exit in all the *-all commands
- More unit tests


configstack/graphviz.go
```diff
@@ -0,0 +1,42 @@
...
+	"fmt"
+	"github.com/gruntwork-io/terragrunt/options"
+	"io"
+	"path/filepath"
+	"strings"
+)
+
+// WriteDot is used to emit a GraphViz compatible definition
+// for a directed graph. It can be used to dump a .dot file.
+func WriteDot(w io.Writer, terragruntOptions *options.TerragruntOptions, modules []*TerraformModule) error {
```
yorinasub17, line 13: Since we already have terraform included as a library, any reason why we can't use https://github.com/hashicorp/terraform/tree/master/digraph?

mauriciopoppe: I think I could use but unfortunately in their implementation there's no way to set properties on the nodes like color which are useful to know in the terragrunt graph (e.g. to know which dependency is not going to be executed)

yorinasub17: Ah gotcha. Can you add a comment here with that message so that we don't ask the same question again a few months down the line? Thanks!

docs/_docs/02_features/execute-terraform-commands-on-multiple-modules-at-once.md
```diff
@@ -274,6 +274,15 @@ If any of the modules fail to deploy, then Terragrunt will not attempt to deploy
...
 To check all of your dependencies and validate the code in them, you can use the `validate-all` command.
 
+To check the dependency graph you can use the `graph-dependencies` command (similar to the `terraform graph` command),
+the graph is output in DOT format The typical program that can read this format is GraphViz, but many web services are also available to read this format.
+
+     terragrunt graph-dependencies | dot -Tsvg > graph.svg
+
+In the example above it'll generate this graph
+
+![terragrunt graph-dependencies]({{site.baseurl}}/assets/img/collections/documentation/graph.png)
```
yorinasub17, line 284: Can you add one more sentence here that mentions that the dependency relationship is in the direction of the arrow, and that terragrunt will run the action in reverse order, bottom up?

mauriciopoppe: sure, I haven't tried this with terragrunt destroy-all and from what I read the graph is reversed

mauriciopoppe: mmm I think I should just reverse the arrow to make the graph clear

yorinasub17:
Build passed and confirmed docs update, so will merge and release. Thanks for your contribution!

mauriciopoppe:
Thank you and gruntwork.io for producing high quality software and making it open source!

PR #1120, opened 2020-04-06
Fix link

bwhaley:


bwhaley:
Thanks for the review. Normally I'd wait for tests to complete, but since this is a minor fix in the docs, I'll go ahead with the merge.

Issue #1123, opened 2020-04-07
terragrunt does not change the path when collecting output from a dependency


mburger:
let's imagine i have the following structure:
```
├── rds
│   ├── secrets.hcl.enc
│   └── terragrunt.hcl
└── rds-snapshot-schedule
    └── terragrunt.hcl
```
rds-snapshot-schedule has a dependency block like this
```
dependency "rds" {
  config_path = "../rds"
}
```
the secrets.hcl.enc file in the rds folder is decrypted by a simple step that runs before terragrunt and generates a secrets.hcl file inside the rds directory
when i run a simple plan on the rds-snapshot-schedule module it fails with the following:
```
  on __redacted__ line 18, in locals:
  18:   secret_vars = read_terragrunt_config("secrets.hcl")
Call to function "read_terragrunt_config" failed: Error reading file at path
secrets.hcl: open secrets.hcl: no such file or directory.
```
the cause is that terragrunt looks for the secrets.hcl file in the rds-snapshot-schedule directory and not in the rds directory were it is actually referenced

this is unexpected behavior and should be at least documented, better would be imho if a fix was made as there is no downside that i can see from changing into the correct directory when collecting the output of another module, but this decision is for somebody else to make :)

yorinasub17:
This is indeed a bug! Sorry about the confusion. Fixed in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.23.8 (binaries will show up shortly).

PR #1125, opened 2020-04-08
Fix issue 1123

yorinasub17:
This fixes https://github.com/gruntwork-io/terragrunt/issues/1123, which was caused by passing the wrong argument to `ParseConfigFile`.


yorinasub17:
Thanks for review! Will merge and release.

PR #1134, opened 2020-04-10
Fix #1124

### LINKED ISSUE #1124: Wrong working directory for dependency with --terragrunt-source option
# Version
terragrunt version v0.22.5
# Use case
I use before_hook to download **override.tf** file from s3 to override some default behaviour of terraform module that I use. 

Example of my terragrunt.hcl

```
include {
}

terraform {
  source = "git::https://bitbucket.org/XXXXXXX/tf-modules.git//datalake?ref=${local.common_vars["tf_modules_version"]}"

  before_hook "before_hook_1" {
    commands = get_terraform_commands_that_need_vars()
    execute = [
      "aws",
      "s3",
      "cp",
      "s3://${local.artifact_bucket}/${local.git_repository}/releases/${local.artifacts_version}/${local.current_folder}.tf",
    "override.tf"]
    run_on_error = false
  }
}

locals {
  // map that hold common variables for all modules
  common_vars_file  = "${get_terragrunt_dir()}/${find_in_parent_folders("common_vars.yaml")}"
  common_vars       = yamldecode(file(local.common_vars_file))
  tfmodules_version = local.common_vars["cmdb"]["tf-modules"]

  // get git repository name for current dataproduct
  git_repository = basename(dirname(get_terragrunt_dir()))
  current_folder = basename(get_terragrunt_dir())

  // get the artifact version for current dataproduct
  artifacts_version          = local.common_vars["cmdb"][local.current_folder]

  // global artifact bucket
  artifact_bucket = local.common_vars["artifact_bucket"]
}

inputs = {
  artifacts_version          = local.artifacts_version
  local_tags = {
    Layer              = local.current_folder
    DataProduct        = local.git_repository
    DataProductVersion = local.artifacts_version
    TFModulesVersion   = local.tfmodules_version
  }
}
```

# What's wrong
*expected behaviour (it works if I do not specify the --terragrunt-source option)*
1. terragrunt create **.terragrunt-cache** directory
2. terragrunt copy modules specify in **source** keyword to the **.terragrunt-cache**
3. terragrunt change working directory to the **.terragrunt-cache/<hash_1>/<hash2>/submodule
4. terragrunt [before_hooks]
5. terragrunt [apply]
6. terragrunt [after_hooks]

*behaviour with --terragrunt-source option*
1. terragrunt create **.terragrunt-cache** directory
2. terragrunt copy modules specify in **source** keyword to the **.terragrunt-cache**
3. terragrunt change working directory to the **.terragrunt-cache/<hash_1>/<hash2>/submodule
4. terragrunt use the same submodule for all dependencies
5. execution failed

# Steps to reproduce
1. [Archive.zip](https://github.com/gruntwork-io/terragrunt/files/4446744/Archive.zip)
2. cd live/app_2
3. terragrunt apply-all --terragrunt-non-interactive  [Works fine]
4. terragrunt apply-all --terragrunt-non-interactive --terragrunt-source ~/tmp/path_to_example/modules [before_hook for app_1 executed in a wrong working directory ]

yorinasub17:
This fixes the bug reported in #1124. The main issue is that the `Source` attribute isn't updated in the context of the target config when retrieving the dependency, which causes it to continue to use the source value of the dependent config.

To fix this, we recompute the `Source` on the target config when it is set. Note that I had to shuffle the functions around to avoid a circular dependency.

### DIFF
--- a/config/dependency.go
+++ b/config/dependency.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/hashicorp/go-getter"
 	"github.com/hashicorp/hcl/v2"
 	"github.com/zclconf/go-cty/cty"
 	"github.com/zclconf/go-cty/cty/gocty"
@@ -351,6 +352,30 @@ func cloneTerragruntOptionsForDependencyOutput(terragruntOptions *options.Terrag
 			return nil, errors.WithStackTrace(err)
 		}
 		targetOptions.DownloadDir = downloadDir
+	}
+
+	// If the Source is set, then we need to recompute it in the context of the target config.
+	if terragruntOptions.Source != "" {
+		// We need the terraform source of the target config to compute the actual source to use
+		partialParseIncludedConfig, err := PartialParseConfigFile(
+			targetConfig,
+			targetOptions,
+			nil,
+			[]PartialDecodeSectionType{TerraformBlock},
+		)
+		if err != nil {
+			return nil, err
+		}
+		// Update the source value to be everything before "//" so that it can be recomputed
+		moduleUrl, _ := getter.SourceDirSubdir(terragruntOptions.Source)
+
+		// Finally, update the source to be the combined path between the terraform source in the target config, and the
+		// value before "//" in the original terragrunt options.
+		targetSource, err := GetTerragruntSourceForModule(moduleUrl, filepath.Dir(targetConfig), partialParseIncludedConfig)
+		if err != nil {
+			return nil, err
+		}
+		targetOptions.Source = targetSource
 	}
 
 	return targetOptions, nil
--- a/config/config_helpers.go
+++ b/config/config_helpers.go
@@ -3,7 +3,10 @@ package config
 import (
 	"fmt"
 	"path/filepath"
+	"regexp"
+	"strings"
 
+	"github.com/hashicorp/go-getter"
 	"github.com/hashicorp/hcl/v2"
 	tflang "github.com/hashicorp/terraform/lang"
 	"github.com/zclconf/go-cty/cty"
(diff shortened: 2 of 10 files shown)


yorinasub17:
NOTE: As a test to ensure the regression test is actually testing the bug, see the failure on e39bdad where the new test fails without any changes, and then it passes in c960eca without any modifications to the test.

config/dependency.go
```diff
@@ -353,6 +354,30 @@ func cloneTerragruntOptionsForDependencyOutput(terragruntOptions *options.Terrag
 		targetOptions.DownloadDir = downloadDir
 	}
 
+	// If the Source is set, then we need to recompute it in the context of the target config.
```
brikis98, line 357: Just to make sure I understand here... If `terragrunt.hcl` has, say:

```hcl
terraform {
  source = "github.com/org/modules.git//module-a"
}

dependency "b" {
  path = "../module-b"
}
```

And `module-b/terragrunt.hcl` contains:

```hcl
terraform {
  source = "github.com/org/modules.git//module-b"
}
```

You're saying that if someone runs `terragrunt apply  --terragrunt-source /local/path/to/modules`, then we are overriding the `source` parameter in *both* the `terragrunt.hcl` of `module-a` *and* of `module-b`?

If so, then this highlights two issues with the `--terragrunt-source` design:

1. If you run `terragrunt apply --terragrunt-source /foo/bar` in folder `yyy`, as of now, the *only* thing it does is override the `source` URL in `yyy/terragrunt.hcl` to the exact value `/foo/bar`. However, if you run `terragrunt apply-all --terragrunt-source /foo/bar` in folder `yyy`, it overrides the `source` value of each child module within `yyy/sub/path` to something like `/foo/bar/sub/path`. The inconsistency between these two behaviors is being highlighted in this PR. Would it make sense to always use the latter behavior? That is `--terragrunt-source` should always be set to the root of your modules repo?
1. This also highlights a limitation of `--terragrunt-source`: it assumes all your modules are coming from the *same* root repo. You won't be able to do an `apply-all` if you're pulling from multiple repos (e.g., your own infra modules, Terraform Registry, Gruntwork repos, etc). 

Perhaps a better solution to this is to support some sort of mapping? 

```
terragrunt apply --terragrunt-source-map source:dest
```

The `--terragrunt-source-map` param would replace any `source` URL that starts with `source` to instead start with `dest`. 

For example:

```
terragrunt apply --terragrunt-source-map github.com/org/modules.git:/local/path/to/modules
```

The above would replace `source = "github.com/org/modules.git//xxx"` with `/local/path/to/modules//xxx` regardless of whether you were running `apply`, or `apply-all`, or using a `dependency`. This would give us a consistent behavior across all command _and_ support mapping multiple repos by specifying the param multiple times:

```
terragrunt apply \
  --terragrunt-source-map github.com/org/modules.git:/local/path/to/modules \
  --terragrunt-source-map github.com/org/another-repo.git:/local/path/to/another-repo
```

We could also consider allowing users to specify this mapping in a file:

```
terragrunt apply --terragrunt-source-map mapping.hcl
```

Where `mapping.hcl` contains:

```hcl
mappings = {
  "github.com/org/modules.git"      = "/local/path/to/modules"
  "github.com/org/another-repo.git" = "/local/path/to/another-repo"  
}
```

Note, I realize the above is a bigger change, so it could also be filed as an issue and done in a separate PR 😁

yorinasub17: Filed https://github.com/gruntwork-io/terragrunt/issues/1138

By the way, are you suggesting we close this PR in favor of solving for the mapping case, or are you ok with the band-aid fix here?

brikis98: > Filed #1138

Thanks! I just updated it with the more detailed description above. 

> By the way, are you suggesting we close this PR in favor of solving for the mapping case, or are you ok with the band-aid fix here?

Your call! I'm OK with either approach. Fixing #1138 would be great, but definitely isn't a high priority, so a band-aid fix for now would be fine too. #1138 might make for a good trial project in the future, too.

yorinasub17: Sounds good. +1 to #1138 being a good trial project. Thanks for the sanity check!

yorinasub17:
Thanks for the review! Merging this in as a bugfix for the reporter, since right now it breaks usage of `--terragrunt-source` with dependencies without any workaround. This at least enables a path forward.

Issue #1121, opened 2020-04-07
generate: Not all file types support comments


lorengordon:
Is there any way to disable the addition of a comment when using `generate` blocks? I was attempting to write an `.auto.tfvars.json` file, and it is written fine, but terragrunt injects a comment into the file that results in errors like this:

```
[terragrunt] 2020/04/07 06:47:06 Running command: terraform apply

Error: Invalid start of value

  on uri_map.auto.tfvars.json line 1:
   1: # Generated by Terragrunt. Sig: nIlQXj57tbuaRZEa

A JSON value must start with a brace, a bracket, a number, a string, or a
keyword.


Error: Root value must be object

  on uri_map.auto.tfvars.json line 1:
   1: # Generated by Terragrunt. Sig: nIlQXj57tbuaRZEa

The root value in a JSON-based configuration must be either a JSON object or a
JSON array of objects.
```

The use case is that I have a very large output I am passing from one state to the next through a `dependency` block. The value is used with `for_each` in the next state, so managing the resources in separate states avoids the "cannot be computed" style of issues. I cannot use `inputs` because the value is too large as an environment variable. So I'm trying to write a file that terraform will read the value from.

My workaround for now is to generate a .tfvars file, but it is a bit sloppy because there is no `hclencode` function. So I need to `jsonencode` the value in the .tfvars file and then `decode` it in my terraform config. Which seems silly, for something that is only needed because I can't stop terragrunt from writing a comment into my json file...

lorengordon:
For my particular use case, I figured out how to generate the .tfvars file directly, using the string template syntax to template the hcl output.

```
  contents = <<-EOF
    some_map = {
      %{ for key in dependency.<foo>.outputs.keys ~}
      "${key}" = "${<operation using built-in functions to construct the value>}"
      %{ endfor ~}
    }
    EOF
```

yorinasub17:
As mentioned in https://github.com/gruntwork-io/terragrunt/pull/1050#issuecomment-591648260, being able to disable this seems like a good idea, but we are all buried at the moment to add this in, so if anyone in the community wants to open a PR that would be great!

yorinasub17:
Has some cycles today to implement this, and released as https://github.com/gruntwork-io/terragrunt/releases/tag/v0.23.10. Binaries will show up shortly.

PR #1141, opened 2020-04-15
Add the ability to disable the signature on generate blocks

yorinasub17:
See #1121 for context


yorinasub17:
Thanks for the review! Will merge and release this now.

PR #1148, opened 2020-04-21
GH-1147: do not add -var or -var-file extra_arguments during "destroy" if plan is provided

### LINKED ISSUE #1147: Disable -var and -var-file flags when running terragrunt destroy with a plan file
This is a mirror of issue #493 that was never corrected for `destroy`.

When running `terragrunt destroy` with a plan file, terragrunt mistakenly includes `-var` and `-var-file` when delegating back to terraform. The execution plan then fails with the following error:

```
You can't set variables with the '-var' or '-var-file' flag
when you're applying a plan file. The variables used when
the plan was created will be used. If you wish to use different
variable values, create a new plan file.
```

FineWolf:
Fixes #1147 

The script now checks if the command in question is destroy and if the second argument is a file. If so, -var and -var-file appends are skipped.

Tests have been updated.

### DIFF
--- a/cli/args.go
+++ b/cli/args.go
@@ -153,7 +153,7 @@ func filterTerraformExtraArgs(terragruntOptions *options.TerragruntOptions, terr
 		for _, arg_cmd := range arg.Commands {
 			if cmd == arg_cmd {
 				lastArg := util.LastArg(terragruntOptions.TerraformCliArgs)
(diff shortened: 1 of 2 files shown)


brikis98:

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.23.11

PR #1154, opened 2020-04-27
Implement Get Terraform Command

bernardoVale:
This PR provides a new function that simply returns the current terraform command.

This will be handy in two of our use-cases. First, we use different IAM Roles a read-only and a read-write, knowing the terraform commands help us set the correct role. Second, we want to define a global variable with the terraform command to pass to some modules that need to assume cross-account roles (vpc peering accept is an example). The variable helps the module decide the IAM Role that needs to be assumed in our IAM organization.

PS: The docs use to be in the README, where are the docs now? Should I worry about documenting the function?


yorinasub17:
Thanks for the PR! This makes sense to me, but two requirements before we can merge this:

- We need an integration test of some kind. Please take a look at the `test` folder for ideas, e.g. [this one is a good starting point](https://github.com/gruntwork-io/terragrunt/blob/master/test/integration_test.go#L1270).
- This needs to be added to the [reference docs](https://github.com/gruntwork-io/terragrunt/blob/master/docs/_docs/04_reference/built-in-functions.md).

bernardoVale:
@yorinasub17 take a look now, please let me know if you need anything else.

PS: I've changed the order of some test arguments which were switched (expected/actual)

yorinasub17:
Tests passed so will merge and release this. Thanks for your contribution!

PR #1160, opened 2020-04-29
Fixing get_env function to accept single env var name as parameter an…

hristo-ganekov-sumup:
…d making it required;

If the parameter is not found it throws exception making it explicitly needed with value provided and not 'defaulting'

@yorinasub17

### DIFF
--- a/config/config_helpers.go
+++ b/config/config_helpers.go
@@ -159,22 +159,20 @@ func getParentTerragruntDir(include *IncludeConfig, terragruntOptions *options.T
 func parseGetEnvParameters(parameters []string) (EnvVar, error) {
 	envVariable := EnvVar{}
 
-	envVariable.Name = parameters[0]
-
-	if len(parameters) == 1 {
+	switch len(parameters) {
+	case 1:
 		envVariable.IsRequired = true
-	} else {
+		envVariable.Name = parameters[0]
+	case 2:
+		envVariable.Name = parameters[0]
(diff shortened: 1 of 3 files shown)


config/config_helpers.go
```diff
@@ -157,12 +158,18 @@ func getParentTerragruntDir(include *IncludeConfig, terragruntOptions *options.T
...
-	if len(parameters) != 2 {
-		return envVariable, errors.WithStackTrace(InvalidGetEnvParams{ExpectedNumParams: 2, ActualNumParams: 2, Example: `getEnv("<NAME>", "<DEFAULT>")`})
-	}
 
 	envVariable.Name = parameters[0]
-	envVariable.DefaultValue = parameters[1]
+
+	if len(parameters) == 1 {
+		envVariable.IsRequired = true
+	} else {
```
syndbg, line 166: Should move to switch case of `parameters` and `default` case is

```
return envVariable, errors.WithStackTrace(InvalidGetEnvParams{ExpectedNumParams: 2, ActualNumParams: 2, Example: `getEnv("<NAME>", "<DEFAULT>")`})
```

yorinasub17: +1 to use `switch` case here.

hristo-ganekov-sumup: Done, thanks for the hint

yorinasub17:
Tests passed so will merge and release this. Thanks for your contribution!

PR #1137, opened 2020-04-14
WIP command flag `--terragrunt-debug`

eboto:
**Work-in-progress do not merge**

Adds a new flag `--terragrunt-debug`. Right now the flag only helps with dependency and local var resolution, but could be expanded to support things like detailed logs, version output, etc.

See the included README.md for more info.

Tasks:
  - [x] Readme
  - [ ] Test-cases
  - [ ] Implementation

PR #1161, opened 2020-04-29
Adding simple environment dump functionality to 'init' when called wi…

hristo-ganekov-sumup:
…th TG_LOG=env

In a situation where parts of the input maps are expected to be pushed from parental terragrunt.hcl files
with local includes itself the transperancy of pushed variables is requirement.

I know this commit is not on par with the coding standards but please consider implementing this functionality.


yorinasub17:
Thanks for the contribution! The use case makes sense, but as you mentioned, this implementation isn't quite to snuff. That said, I believe this use case is covered by https://github.com/gruntwork-io/terragrunt/pull/1137 which we intend to complete in the near future. Does the proposed solution in that PR cover your needs?

hristo-ganekov-sumup:
Closing because its addressed in https://github.com/gruntwork-io/terragrunt/pull/1137

Issue #1143, opened 2020-04-16
Terragrunt creates an s3 bucket with invalid KMS settings


cliveza:
When running `terragrunt plan` for the first time and accepting the the prompt to create an S3 bucket for the backend. The bucket is configured with a custom KMS key even though a `kms_key_id` is not specified.

```
$ terragrunt -v
terragrunt version v0.23.10
```

When you `terragrunt apply`, Terraform will create the resource but is not able to save the state file on s3

```
Failed to save state: failed to upload state: KMS.NotFoundException: Invalid keyId aws/s3
```

To fix this you have to manually edit the KMS properties for the bucket and either:

1.  Enter a valid custom KMS ARN, or
2. Change the dropdown from "Custom KMS ARN" to "aws/s3"

My main terragrunt.hcl has the following:

```
remote_state {
  backend = "s3"

  generate = {
    path = "backend.tf"
    if_exists = "overwrite_terragrunt"
  }

  config = {
    region = "us-east-1"
    bucket = "some.bucket.somewhere"
    key    = "bob.tfstate"
  }
}
```

This screenshot shows what the bucket properties look like immediately after Terragrunt has created the bucket 
![image](https://user-images.githubusercontent.com/1900666/79488785-44b3dd80-7fe8-11ea-8de8-b7531c74e8b9.png)

jfharden:
The KMS key aws/s3 doesn't exist until aws creates it automatically for you (so if you manually make any bucket and choose AWS-KMS that will also resolve this problem). Annoyingly you are not allowed to make keys which begin with the 'aws/' prefix through KMS so you can't create this yourself.

The configuration is set here: https://github.com/gruntwork-io/terragrunt/blob/master/remote/remote_state_s3.go#L530

Possible options I can think of:
1. Instead of defaulting to `aws/s3` default to a terragrunt specific KMS key, and create it first if it doesn't exist
2. Check to see if aws/s3 KMS key exists, if not error with a helpful error message
3. Default the encryption to AES256

this: 
```
defEnc := &s3.ServerSideEncryptionByDefault{SSEAlgorithm: aws.String(s3.ServerSideEncryptionAes256)}
```

instead of:
```
defEnc := &s3.ServerSideEncryptionByDefault{KMSMasterKeyID: aws.String("aws/s3"), SSEAlgorithm: aws.String(s3.ServerSideEncryptionAwsKms)}
```

jfharden:
[...]

PR #1163, opened 2020-05-05
Remove manual definition of KMSMasterKeyID when creating new bucket

jfharden:
Simplest resolution to https://github.com/gruntwork-io/terragrunt/issues/1143

I tried hard to think how to test this, it's extremely difficult to test because it needs an AWS account which has never had the aws/s3 key created. An integration test would look something like:

1. create an AWS account
2. create an IAM admin user on that account
3. run any terragrunt apply to create the s3 bucket
4. try and copy a file to the s3 bucket without manually specifying any encryption (so it tries to use the bucket default)


yorinasub17:
Regression build passed, so going to go ahead and merge + release this. Thanks for your contribution!

PR #1170, opened 2020-05-07
Add get_terraform_cli_args function

camlow325:
Similar to #1154, this PR provides a new function that returns the cli args that are passed to the current Terraform command.

We have some cases where we need to add before and after hooks to the `terraform` block in our Terragrunt configuration. Some of those hooks need to behave differently based on CLI arguments that would be passed along to Terraform. For example, a `terragrunt plan` vs. a `terragrunt plan --destroy` needs to be handled differently. We could handle this by ensuring that the extra args are always set through the `TF_CLI_ARGS` environment variable instead of the Terragrunt command line, e.g., `TF_CLI_ARGS=--destroy terragrunt plan`, so that the hooks could see the same args that would be used by Terraform. It would be nicer, though, to be able to use standard Terragrunt CLI args while still being able to interpret those from hooks as needed.

Thank you for considering this PR and for all of your work on this project!

### DIFF
--- a/docs/_docs/04_reference/built-in-functions.md
+++ b/docs/_docs/04_reference/built-in-functions.md
@@ -36,6 +36,8 @@ Terragrunt allows you to use built-in functions anywhere in `terragrunt.hcl`, ju
   
   - [get\_terraform\_command()](#get_terraform_command)
 
+  - [get\_terraform\_cli\_args()](#get_terraform_cli_args)
+
   - [get\_aws\_account\_id()](#get_aws_account_id)
 
   - [get\_aws\_caller\_identity\_arn()](#get_aws_caller_identity_arn)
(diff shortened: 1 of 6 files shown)


docs/_docs/04_reference/built-in-functions.md
```diff
@@ -413,6 +415,16 @@ inputs = {
 }
 ```
 
+## get\_terraform\_cli\_args
+
+`get_terraform_command()` returns cli args for the current terraform command in execution. Example:
```
yorinasub17, line 420:
```suggestion
`get_terraform_cli_args()` returns cli args for the current terraform command in execution. Example:
```

docs/_docs/04_reference/built-in-functions.md
```diff
@@ -413,6 +415,16 @@ inputs = {
...
+
+``` hcl
+inputs = {
+  current_command = get_terraform_cli_args()
```
yorinasub17, line 424:
```suggestion
  current_cli_args = get_terraform_cli_args()
```

yorinasub17:
Thanks for addressing the nits! The tests passed, so will go ahead and merge + release this.

Issue #631, opened 2018-12-28
Best Practice Question: How to handle "third-party" providers?


jhoblitt:
I've been wrestling with the best way to handle the `sl1pm4t` fork of the kubernetes provider and have been trying to avoid using `TF_PLUGIN_CACHE_DIR` so that behavior is, ideally, the same for a dev running terragrunt locally and under ci/cd.

My current approach is to use a `before_hook` to trigger a makefile in the same dir as the `terraform.tfvars` for the deployment. Eg.,

```hcl
terragrunt = {
  include {
    path = "${find_in_parent_folders()}"
  }

  terraform {
    source = "git::https://github.com/....git//?ref=master"

    before_hook "tf_plugins" {
      commands = ["init", "init-from-module"]

      run_on_error = false

      execute = [
        "bash", "-lc", "cd ${get_tfvars_dir()}; make"
      ]
    }
  } # terraform
}
```

```make
UNAME := $(shell uname -s | tr A-Z a-z)
DL_DIR = downloads
ARCH = amd64
TF_PLUG_DIR := terraform.d/plugins/$(UNAME)_$(ARCH)

# the filename in the archive as an extra _x4 suffix
TF_K8S_ZIP_VER = v1.3.0-custom
TF_K8S_BIN_VER = v1.3.0-custom_x4
TF_K8S_ZIP_FILE := terraform-provider-kubernetes_$(TF_K8S_ZIP_VER)_$(UNAME)_$(ARCH).zip
TF_K8S_BIN_FILE := terraform-provider-kubernetes_$(TF_K8S_BIN_VER)
TF_K8S_ZIP_DL := $(DL_DIR)/$(TF_K8S_ZIP_FILE)
TF_K8S_BIN := $(TF_PLUG_DIR)/$(TF_K8S_BIN_FILE)
TF_K8S_URL := https://github.com/sl1pm4t/terraform-provider-kubernetes/releases/download/$(TF_K8S_ZIP_VER)/$(TF_K8S_ZIP_FILE)

.PHONY: all
all: $(TF_K8S_BIN)

$(TF_K8S_BIN): | $(TF_K8S_ZIP_DL)
	unzip -j $(TF_K8S_ZIP_DL) $(TF_K8S_BIN_FILE) -d $(TF_PLUG_DIR)

$(TF_K8S_ZIP_DL): | $(DL_DIR) $(TF_PLUG_DIR)
	wget -nc $(TF_K8S_URL) -O $@

[...]

PR #1176, opened 2020-05-12
[WIP|awaiting feedback] Custom provider support

jereksel:
Hi,

I love terraform and I want to use it for my homelab, but support for custom providers is pretty mixed - while it's very easy to write them, using them is completely different story.

This PR contains basic implementation of custom provider support for terragrunt.

Fortunatelly Go since version 1.11 support compiling programs outside of GOPATH and it allows us to easily compile them in terragrunt cache directory.

In current implementation Go compiler is required on host computer for module compilation. In future I would like to add support for zips with binary and building with Docker.

Usage
```
terraform {
    custom_provider "PROVIDER_NAME" {
        source = "PROVIDER_SOURCE"
    }
}
```

Example usage: 

```
terraform {
    custom_provider "xenserver" {
        source = "git::git@github.com:terra-farm/terraform-provider-xenserver.git?ref=87d0c36bce49e47036acba4e99712eb2672def52"
    }
}
```

Implements #631


lorengordon:
Oh, this seems super cool! ❤️

yorinasub17:
Thanks for your contribution! This is indeed an exciting feature and something I had in mind as well so love that there is a community contribution!

I think this is a good start for an incremental version focused on installing custom providers, but there are a few high level design questions that should be thought through. In particular, this sets the precedent for terragrunt to implement provider management features. Given that, I think we should at least think through the UX implications of this feature. Here are a few questions that pop to mind:

- How do we keep the `custom_providers` block DRY across modules that share the provider?
- Certain providers are likely to be reused across many modules. It seems like the system could easily get overloaded in an `apply-all` or `validate-all` if they all start compiling in parallel. Is there a way to detect and cache provider versions so that it is only compiled once?
- It is fairly common practice to regularly wipe the terragrunt cache. I'm not sure if compiling and storing the providers there is the most efficient. I wonder if we should store providers in a different folder? In fact, should this actually just store the providers where terraform expects them (`.terraform.d/plugins` or `TF_PLUGIN_CACHE_DIR`)?
- If we address caching here, would it make sense to actually expand this to a larger feature of general provider management? That is, should there be a way to specify provider versions that terragrunt should prefetch into a plugin directory? <= This actually feels like a different tool, but one way I see this working in terragrunt is that you can run a command like `terragrunt plugin install` which will install all the terraform providers in the terraform provider dir for the given tree.

To help think through these questions and more, can you actually summarize your thoughts around the features into an RFC as [suggested in our contribution guidelines](https://terragrunt.gruntwork.io/docs/community/contributing/)?

barryib:
Hummm... This is nice feature indeed. But I think that Hashicorp is already working on the provider distribution issue.

- https://github.com/gruntwork-io/terragrunt/issues/631#issuecomment-629150093
- https://www.terraform.io/docs/registry/providers/docs.html
- https://registry.terraform.io/browse/providers

PR #1178, opened 2020-05-13
Bump tf to v0.12.25

chenrui333:
Attach the changelog down below.

### NOTES:

backend/s3: Region validation now automatically supports the new af-south-1 (Africa (Cape Town)) region. For AWS operations to work in the new region, the region must be explicitly enabled as outlined in the AWS Documentation. When the region is not enabled, the Terraform S3 Backend will return errors during credential validation (e.g. error validating provider credentials: error calling sts:GetCallerIdentity: InvalidClientTokenId: The security token included in the request is invalid). (#24744)

### ENHANCEMENTS:

backend/s3: Support automatic region validation for af-south-1 (#24744)
backend/remote: Add support for force push to remote backend (#24884)

### BUG FIXES:

core: Destroy provisioners should not evaluate for_each expressions (#24163)
core: Fix races in GetVariableValue (#24599)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,23 +4,22 @@ go 1.13
 
 require (
 	cloud.google.com/go/storage v1.6.0
-	github.com/aws/aws-sdk-go v1.29.34
+	github.com/aws/aws-sdk-go v1.30.12
 	github.com/creack/pty v1.1.9
 	github.com/fatih/color v1.9.0
 	github.com/fatih/structs v1.1.0
 	github.com/go-errors/errors v1.0.2-0.20180813162953-d98b870cc4e0
-	github.com/google/go-cmp v0.4.0
 	github.com/gruntwork-io/terratest v0.26.1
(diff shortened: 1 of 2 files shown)


brikis98:
Thanks for the PR! Could you provide some context on what's in the new version that drove this update?

chenrui333:
@brikis98 attached the changelog, what I am looking for is the CI consistency in terms of the terraform version. let me know if it makes sense. :)

brikis98:
@chenrui333 The Terraform version you're updating in this PR only affects the [built-in functions](https://www.terraform.io/docs/configuration/functions.html) that you can use in your `terragrunt.hcl` files. It does NOT have any impact on which version of Terraform gets used when you run, say, `terragrunt apply`. Terragrunt will run whatever version you have installed on your `PATH`.

chenrui333:
Sounds good, closing this PR.

PR #1184, opened 2020-05-16
Fix color output on Windows

### LINKED ISSUE #614: No colour output since release 0.17.0 on Windows x64
It seems 0.17.x breaks colour output, and prints the escape sequences instead, at least on Windows x64. This worked as intended on 0.16.x.

**0.16.x:**
terragrunt_0 16 x png

**0.17.x:**
terragrunt_0 17 x

**System:**
- Windows 10 Enterprise build 17134
- Terraform 0.11.7 and 0.11.10
- Terragrunt 0.16.10, 0.16.14, 0.17.0, 0.17.1, 0.17.2

jereksel:
Sources:

https://stackoverflow.com/questions/56460651/golang-fmt-print-033c-and-fmt-print-x1bc-are-not-clearing-screenansi-es
https://github.com/containerd/console/blob/f652dc3/console_windows.go#L46

Fixes #614

### DIFF
--- a/shell/ptty_windows.go
+++ b/shell/ptty_windows.go
@@ -3,6 +3,8 @@
 package shell
 
 import (
+	"fmt"
+	"golang.org/x/sys/windows"
 	"io"
 	"os"
 	"os/exec"
@@ -10,6 +12,25 @@ import (
 	"github.com/gruntwork-io/terragrunt/errors"
 	"github.com/gruntwork-io/terragrunt/options"
 )
+
+func PrepareConsole() {
+	enableVirtualTerminalProcessing(os.Stderr)
+	enableVirtualTerminalProcessing(os.Stdout)
+}
+
+func enableVirtualTerminalProcessing(file *os.File) {
+
+	var mode uint32
+	handle := windows.Handle(file.Fd())
+	if err := windows.GetConsoleMode(handle, &mode); err != nil {
+		fmt.Printf("failed to get console mode: %v\n", err)
+	}
+
+	if err := windows.SetConsoleMode(handle, mode|windows.ENABLE_VIRTUAL_TERMINAL_PROCESSING); err != nil {
+		fmt.Printf("failed to set console mode: %v\n", err)
+	}
+
+}
 
 // For windows, there is no concept of a pseudoTTY so we run as if there is no pseudoTTY.
 func runCommandWithPTTY(terragruntOptions *options.TerragruntOptions, cmd *exec.Cmd, cmdStdout io.Writer, cmdStderr io.Writer) error {
--- a/shell/ptty_windows.go
+++ b/shell/ptty_windows.go
@@ -31,7 +31,7 @@ func enableVirtualTerminalProcessing(options *options.TerragruntOptions, file *o
 
 	if err := windows.SetConsoleMode(handle, mode|windows.ENABLE_VIRTUAL_TERMINAL_PROCESSING); err != nil {
 		options.Logger.Printf("failed to set console mode: %v\n", err)
-		_ = windows.SetConsoleMode(handle, mode)
+		windows.SetConsoleMode(handle, mode)
 	}
 }
 
--- a/shell/ptty_windows.go
+++ b/shell/ptty_windows.go
@@ -3,10 +3,11 @@
 package shell
 
 import (
-	"golang.org/x/sys/windows"
 	"io"
 	"os"
 	"os/exec"
+
+	"golang.org/x/sys/windows"
 
 	"github.com/gruntwork-io/terragrunt/errors"
 	"github.com/gruntwork-io/terragrunt/options"
@@ -36,7 +37,7 @@ func enableVirtualTerminalProcessing(options *options.TerragruntOptions, file *o
 }
 
 // For windows, there is no concept of a pseudoTTY so we run as if there is no pseudoTTY.
-func runCommandWithPTTY(cmd *exec.Cmd, cmdStdout io.Writer, cmdStderr io.Writer) error {
+func runCommandWithPTTY(terragruntOptions *options.TerragruntOptions, cmd *exec.Cmd, cmdStdout io.Writer, cmdStderr io.Writer) error {
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = cmdStdout
 	cmd.Stderr = cmdStderr
--- a/shell/ptty_windows.go
+++ b/shell/ptty_windows.go
@@ -3,7 +3,6 @@
 package shell
 
 import (
-	"fmt"
 	"golang.org/x/sys/windows"
 	"io"
 	"os"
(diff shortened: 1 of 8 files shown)


brikis98:
Thanks for the PR! I'm super booked this week, but will try to take a look in the next few days.

shell/ptty_windows.go
```diff
@@ -11,6 +13,25 @@ import (
 	"github.com/gruntwork-io/terragrunt/options"
 )
 
+func PrepareConsole() {
+	enableVirtualTerminalProcessing(os.Stderr)
+	enableVirtualTerminalProcessing(os.Stdout)
+}
+
+func enableVirtualTerminalProcessing(file *os.File) {
+
```
brikis98, line 22: NIT: extra whitespace?

shell/ptty_windows.go
```diff
@@ -11,6 +13,25 @@ import (
...
+	var mode uint32
+	handle := windows.Handle(file.Fd())
+	if err := windows.GetConsoleMode(handle, &mode); err != nil {
+		fmt.Printf("failed to get console mode: %v\n", err)
```
brikis98, line 26: 1. Should we really ignore these errors? Or return them and cause the Terragrunt command to fail? I can't tell if this is something optional or not.
1. If we are going to ignore these errors, we should log them, probably with debug logging (`util.Debugf`), using the logger in `terragrunt.Options`, which can be passed in here as an arg.

shell/ptty_windows.go
```diff
@@ -11,6 +13,25 @@ import (
...
+	}
+
+	if err := windows.SetConsoleMode(handle, mode|windows.ENABLE_VIRTUAL_TERMINAL_PROCESSING); err != nil {
+		fmt.Printf("failed to set console mode: %v\n", err)
```
brikis98, line 30: Same here with error handling

shell/ptty_windows.go
```diff
@@ -11,6 +13,25 @@ import (
 	"github.com/gruntwork-io/terragrunt/options"
 )
 
+func PrepareConsole() {
+	enableVirtualTerminalProcessing(os.Stderr)
+	enableVirtualTerminalProcessing(os.Stdout)
+}
+
+func enableVirtualTerminalProcessing(file *os.File) {
```
brikis98, line 21: Please add a comment explaining what this method does, including those links to StackOverflow / GitHub

shell/ptty_windows.go
```diff
@@ -11,6 +13,25 @@ import (
 	"github.com/gruntwork-io/terragrunt/options"
 )
 
+func PrepareConsole() {
```
brikis98, line 16: Please add a comment explaining what this method does

shell/ptty_windows.go
brikis98: Do we need [this fallback](https://github.com/containerd/console/blob/f652dc3/console_windows.go#L49-L51)?

shell/ptty_windows.go
```diff
@@ -11,6 +13,25 @@ import (
...
+	enableVirtualTerminalProcessing(os.Stderr)
+	enableVirtualTerminalProcessing(os.Stdout)
```
brikis98, line 18: Do we need to do it [for stdin too](https://github.com/containerd/console/blob/f652dc3/console_windows.go#L33-L44)?

yorinasub17:
> so I'm not sure on how to test this.

FYI I had plans to test this on my windows desktop tomorrow or Wednesday.

shell/ptty_windows.go
jereksel: I was thinking about it to be honest and I don't know, because we don't input any escaped characters. But I guess I'll add it for completeness sake.

shell/ptty_windows.go
jereksel: Yeah, we can't just ignore them, because `mode` would not be set - I've added `return` there - I can only test it on my Windows 10 and I'm not sure how it will behave on other Windows versions

shell/ptty_windows.go
jereksel: Removed

shell/ptty_windows.go
jereksel: That's good question - Windows docs does not say what happens to the state when error is returned. Just in case I've added setting console mode to previous value

shell/ptty_windows.go
jereksel: Added

shell/ptty_windows.go
jereksel: Added to `PrepareConsole`

shell/ptty_windows.go
brikis98: The `return` helps to prevent the code from continuing to run... But should the actual error be returned all the way up the stack, causing Terragrunt to exit? Or are these errors safe to ignore and just mean less pretty terminal output?

shell/ptty_windows.go
```diff
@@ -11,8 +12,31 @@ import (
...
+	var mode uint32
+	handle := windows.Handle(file.Fd())
+	if err := windows.GetConsoleMode(handle, &mode); err != nil {
+		options.Logger.Printf("failed to get console mode: %v\n", err)
+		return
+	}
+
+	if err := windows.SetConsoleMode(handle, mode|windows.ENABLE_VIRTUAL_TERMINAL_PROCESSING); err != nil {
+		options.Logger.Printf("failed to set console mode: %v\n", err)
+		_ = windows.SetConsoleMode(handle, mode)
```
brikis98, line 34: NIT: If we're going to ignore the error, you don't need the `_ = ` portion.

brikis98:
@yorinasub17 Let me know if you've had a chance to try this out.

yorinasub17:
Apologies for the delay here. I have verified that this works on my windows box!

brikis98:
Thanks Yori! @jereksel Could you resolve the last remaining (very minor) comment? After that, I can kick off tests.

shell/ptty_windows.go
jereksel: Sorry about that - GoLand without it was warning me that error is ignored

brikis98: But you're still ignoring the error, so how is this better? Either don't use it at all or at least log it...

jereksel:
Sorry that it took so long

brikis98:

brikis98:
Hm, looks like a compile error?

```
# github.com/gruntwork-io/terragrunt/shell
shell/run_shell_cmd.go:98:31: not enough arguments in call to runCommandWithPTTY
	have (*exec.Cmd, io.Writer, io.Writer)
	want (*options.TerragruntOptions, *exec.Cmd, io.Writer, io.Writer)
# github.com/gruntwork-io/terragrunt/shell [github.com/gruntwork-io/terragrunt/shell.test]
shell/run_shell_cmd.go:98:31: not enough arguments in call to runCommandWithPTTY
	have (*exec.Cmd, io.Writer, io.Writer)
	want (*options.TerragruntOptions, *exec.Cmd, io.Writer, io.Writer)
FAIL	github.com/gruntwork-io/terragrunt/cli [build failed]
FAIL	github.com/gruntwork-io/terragrunt/config [build failed]
FAIL	github.com/gruntwork-io/terragrunt/configstack [build failed]
```

jereksel:
Should be fine now

brikis98:
Re-running...

brikis98:

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.23.27

PR #1192, opened 2020-05-22
Adds `terragrunt_version_constraint` functionality

### LINKED ISSUE #849: Terragrunt version constraint
I didn't see this documented but it would be useful to support a terragrunt version constraint in the same way terraform version constraints are supported.

Ensuring a team of people use a consistent version allows for new features to be supported and avoids errors when people on older versions try to run the same code base and the error message may not be immediately obvious is down to incorrect version. It's more user friendly that the user is prompted they are not running a required version.

A second more important case is that functions may change over time with bugfixes either intentionally or as a side effect. Ensuring everyone applying from a repo is using the same terragrunt version would at least ensure consistency.

jakauppila:
Adds the ability to define a `terragrunt_version_constraint` within the Terragrunt configuration.

Since we're dependent on the version fed in via the build process, I assume we really shouldn't have to worry about it not being parsable? If it does encounter an issue, currently have it defaulting to "0.0"; while not ideal, it's probably better than just erroring out.

Are there any additional tests to add?

Fixes #849


brikis98:
Thanks for the PR! I'm super booked today, but will try to take a look in the next few days.

cli/version_check.go
```diff
@@ -42,6 +42,26 @@ func CheckTerraformVersion(constraint string, terragruntOptions *options.Terragr
 	return checkTerraformVersionMeetsConstraint(terragruntOptions.TerraformVersion, constraint)
 }
 
+// Check that the currently running Terragrunt version works meets the specified version constraint and return an error
```
yorinasub17, line 45:
```suggestion
// Check that the currently running Terragrunt version meets the specified version constraint and return an error
```

cli/version_check_test.go
```diff
@@ -90,3 +92,43 @@ func testParseTerraformVersion(t *testing.T, versionString string, expectedVersi
...
+
+func testCheckTerragruntVersionMeetsConstraint(t *testing.T, currentVersion string, versionConstraint string, versionMeetsConstraint bool) {
+	current, err := version.NewVersion(currentVersion)
+	if err != nil {
+		t.Fatalf("Invalid current version specified in test: %v", err)
+	}
+
+	err = checkTerragruntVersionMeetsConstraint(current, versionConstraint)
+	if versionMeetsConstraint && err != nil {
+		assert.Nil(t, err, "Expected Terragrunt version %s to meet constraint %s, but got error: %v", currentVersion, versionConstraint, err)
```
yorinasub17, line 130: NIT: Use `t.Errorf` here to indicate failure.

options/options.go
```diff
@@ -161,18 +164,18 @@ func NewTerragruntOptions(terragruntConfigPath string) (*TerragruntOptions, erro
...
+		MaxFoldersToCheck: DEFAULT_MAX_FOLDERS_TO_CHECK,
+		AutoRetry:         true,
+		MaxRetryAttempts:  DEFAULT_MAX_RETRY_ATTEMPTS,
+		Sleep:             DEFAULT_SLEEP,
+		RetryableErrors:   util.CloneStringList(RETRYABLE_ERRORS),
+		ExcludeDirs:       []string{},
+		IncludeDirs:       []string{},
+		StrictInclude:     false,
+		Parallelism:       DEFAULT_PARALLELISM,
+		Check:             false,
```
yorinasub17, line 178: NIT: This looks like an odd `go fmt` change. What version of go are you using?

jakauppila:
Addressed all the feedback, let me know if I missed anything

yorinasub17:
The build passed, so will go ahead and merge + release. Thanks again for your contribution!

PR #1196, opened 2020-05-23
Partial parse version checks

yorinasub17:
This improves on the recent version check functionalities that were added by leveraging partial parsing. This is especially useful for the  `terragrunt_version_constraint` functionality that was added in https://github.com/gruntwork-io/terragrunt/pull/1192 .

Specifically, this will now error out with the version error message, instead of a parsing error when using an incompatible terragrunt version.

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -228,8 +228,7 @@ func runApp(cliContext *cli.Context) (finalErr error) {
 
 	// If someone calls us with no args at all, show the help text and exit
 	if !cliContext.Args().Present() {
-		cli.ShowAppHelp(cliContext)
-		return nil
+		return cli.ShowAppHelp(cliContext)
 	}
 
 	terragruntOptions, err := ParseTerragruntOptions(cliContext)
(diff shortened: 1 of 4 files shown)


cli/cli_app.go
```diff
@@ -389,6 +367,46 @@ func RunTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 	return runTerragruntWithConfig(terragruntOptions, terragruntConfig, false)
 }
 
+// Check the version constraints of both terragrunt and terraform. Note that as a side effect this will set the
+// following settings on terragruntOptions:
+// - TerraformPath
+// - TerraformVersion
```
brikis98, line 373: NIT: Seems a bit odd to do this as a side effect. Should we have two methods, one that sets the version numbers, and one that checks constraints?

yorinasub17: I had the exact same thought, but then it is a bit hairy to refactor right now because `RunTerraformCommand` relies on `TerraformPath` being set on the `terragruntOptions`, and I don't want to make a copy in this function just to avoid the side effect.

yorinasub17: Oh wait I guess you are suggesting do the side effects in a separate function first. Hmm... I guess that is cleaner. Will fix.

yorinasub17: Ok on second thought, I'm going to keep as is. After looking through the code and attempting a refactor, splitting the function turned out to be awkward because I still need to reference the partially parsed terragrunt config after the side effects. So it sort of becomes three functions:

- partialParseTerragruntConfig
- setTerraformVersionInfo (side effect)
- checkVersionConstraints

This felt a bit too verbose for my taste, considering how big the main function is now.

yorinasub17:
Thanks for review! Will merge and release this now.

PR #1199, opened 2020-05-26
GCS remote state improvements

Jawshua:
This PR has two improvements for GCS remote state.

1. Fixes a bug in the config comparison function `gcsConfigValuesEqual` that was deleting Terragrunt specific config values from the source configuration. This bug was resulting in options such as `skip_bucket_creation` being ignored.

2. Enables GCS authentication using a fixed token defined in the `GOOGLE_OAUTH_ACCESS_TOKEN` env var, which is an option for both the Terraform GCS remote state and Google Cloud provider.


robmorgan:
Hi @Jawshua it looks good and thanks for including tests as well!

PR #1208, opened 2020-06-02
Ignore Terraform data directories when running *-all commands

### LINKED ISSUE #1169: plan-all loads `terragrunt.hcl` config from downloaded module
When running `terragrunt plan-all` the resulting `terraform` command does not include arguments from `extra_arguments` defined in the root or module `terragrunt.hcl` config.

E.g. If I `cd` to an individual module and run `terragrunt plan`, the resulting `terraform` command is 

```
terraform plan -var env=staging -var region=us-east-1 -var ticket=TIP-UP
```

I would assume running `terragrunt plan-all` from the root directory would yield the same `terraform` command for that module but it doesn't, it is just 

```
terraform plan
```

The plan-all then fails because certain required arguments are not set.

terragrunt version v0.23.14
Terraform v0.12.24

zachwhaley:
If a downloaded terraform module happens to have a `terragrunt.hcl` file
in its source code, then Terragrunt `*-all` commands will naturally assume
this `terragrunt.hcl` is part of a local module when in fact Terragrunt
should ignore this `terragrunt.hcl` and its directory.

Fixes #1169 

Test results: https://gist.github.com/zachwhaley/cd914723be1dab98e6c5534e12b6615b

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -258,8 +258,8 @@ func FindConfigFilesInPath(rootPath string, terragruntOptions *options.Terragrun
 }
 
 // Returns true if the given path with the given FileInfo contains a Terragrunt module and false otherwise. A path
-// contains a Terragrunt module if it contains a Terragrunt configuration file (terragrunt.hcl, terragrunt.hcl.json) and is not a cache
-// or download dir.
+// contains a Terragrunt module if it contains a Terragrunt configuration file (terragrunt.hcl, terragrunt.hcl.json)
+// and is not a cache, data, or download dir.
 func containsTerragruntModule(path string, info os.FileInfo, terragruntOptions *options.TerragruntOptions) (bool, error) {
 	if !info.IsDir() {
 		return false, nil
@@ -267,6 +267,15 @@ func containsTerragruntModule(path string, info os.FileInfo, terragruntOptions *
 
 	// Skip the Terragrunt cache dir
 	if strings.Contains(path, options.TerragruntCacheDir) {
+		return false, nil
+	}
+
+	// Skip the Terraform data dir
+	dataDir, found := terragruntOptions.Env["TF_DATA_DIR"]
+	if !found {
+		dataDir = options.TerraformDataDir
+	}
+	if strings.Contains(path, dataDir) {
 		return false, nil
 	}
 
--- a/config/config.go
+++ b/config/config.go
@@ -265,6 +265,11 @@ func containsTerragruntModule(path string, info os.FileInfo, terragruntOptions *
 		return false, nil
 	}
 
+	canonicalPath, err := util.CanonicalPath(path, "")
+	if err != nil {
+		return false, err
+	}
+
 	splitPath := util.SplitPath(path)
 
 	// Skip the Terragrunt cache dir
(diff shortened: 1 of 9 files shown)


config/config.go
```diff
@@ -270,6 +270,15 @@ func containsTerragruntModule(path string, info os.FileInfo, terragruntOptions *
 		return false, nil
 	}
 
+	// Skip the Terraform data dir
+	dataDir, found := terragruntOptions.Env["TF_DATA_DIR"]
+	if !found {
+		dataDir = options.TerraformDataDir
+	}
```
brikis98, line 277: There's a `DataDir` method in `options.go` that does something very similar. Could we extract the logic into one function and use it in both places?

zachwhaley: I saw this method but it prepends the workspace directory to the data directory path and I wasn't sure if this would work given that we want to ignore any occurrence of the data directory, right? e.g. if a module of a module has a terragrunt.hcl file.

Did you want me to modify that method to conditionally return the workspace + data dir or just the data dir?  Or is my concern unfounded?

brikis98: Looking at the code, and thinking about this a bit more deeply, I think both places may have subtle bugs, and I think creating a common method could fix them.

First, the bugs: 

1. As best as I can tell, `TF_DATA_DIR` can be either a relative or absolute path. If it's a relative path, then the `DataDir()` method below is broken, as it's not prepending that path with the working dir.
1. Similarly, if it's a relative path, then the code here is potentially buggy too. E.g., If `TF_DATA_DIR` is set to `foo`, then the code below would exclude `/foo`, `/foo-bar`, `/.foo`, and a dozen other variations.
1. Moreover, the code below would exclude `/.terraform`, `/.terraform-oops`, `/oops.terraform`, etc.

So I think there are a few things we should do:

1. Create a method called `TerraformDataDir()` that looks up `TF_DATA_DIR` and either returns it or `.terraform` if it's not set.
1. Update the `DataDir()` method to call `TerraformDataDir()` and prepend it with the working dir if it's not an absolute path.
1. Update the code here to (a) call `DataDir()`, (b) call [SplitList](https://golang.org/pkg/path/filepath/#SplitList) on `path`, (c) if `DataDir()` returned an absolute path, call `SplitList` on it too and check if it is an exact prefix of `path`, (d) if `DataDir()` returned a relative path, check if it an exact subsequence of `path`.

zachwhaley:
So, here's a dump of code...

I believe I've addressed the issues mentioned, but not being a regular Go dev I'm not sure I did so in the most Go like fashion.  I found I had to write some standard functions that did not seem available in the stdlib.

Also SplitList does not split on paths but on `:`, e.g. it is used for splitting things like `PATH`, so I had to write my own 😐 

Let me know what you think.  I am very open to thoughts and opinions :)

zachwhaley:
Actually, I think I missed this 😕

> if DataDir() returned an absolute path, call SplitList on it too and check if it is an exact prefix of path

util/file.go
```diff
@@ -256,6 +256,23 @@ func JoinPath(elem ...string) string {
...
 
+// SplitPath splits the given path into a list, e.g. "/foo/bar/boo.txt" -> ["foo", "bar", "boo.txt"]
+func SplitPath(path string) []string {
+	var splitPath []string
+	dir, file := path, ""
+
+	for dir != "" {
+		dir, file = filepath.Split(filepath.Clean(dir))
+		splitPath = append([]string{file}, splitPath...)
+		if len(dir) == 1 && os.IsPathSeparator(uint8(dir[0])) {
```
brikis98, line 267: Ah, sorry for misreading `filepath.Split`. 

Quick sanity check: does this implementation behave differently than `strings.Split(filepath.Clean(path), filepath.Separator)`?

brikis98:
> I believe I've addressed the issues mentioned, but not being a regular Go dev I'm not sure I did so in the most Go like fashion. I found I had to write some standard functions that did not seem available in the stdlib.

Heh, yea, Go, for some reason, is missing a ton of things you'd expect to be in the stdlib :-\

util/collections.go
```diff
@@ -15,15 +15,55 @@ func MatchesAny(regExps []string, s string) bool {
...
+				i++
+				if i >= len(l) || item != l[i] {
+					match = false
+					beg += i
+					break
+				}
+				match = true
+			}
+		}
+		return match
```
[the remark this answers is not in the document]
zachwhaley: Your code does seem simpler, thanks!

util/file.go
zachwhaley: Only slightly, strings.Split will return the root / as an empty string which is not a big deal :)

https://play.golang.org/p/asbyIPP-eiK

brikis98:
Tests passed! Merging now.

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.23.24

Issue #1217, opened 2020-06-14
prevent_destroy flag override ignored in child terragrunt.hcl


dudicoco:
Hi,

I have set `prevent_destroy = true` in my root terragrunt.hcl.
In one of the child terragrunt.hcl files I've tried to override this with `prevent_destroy = false`.

When trying to run destroy terragrunt is throwing an error that the `prevent_destroy` flag should be set to `false`.

Is this a bug?

Thanks

brikis98:
So you're saying that when you run `destroy` in the child module, which has `prevent_destroy = false`, it's acting as if `prevent_destroy = true`? What version of Terragrunt?

dudicoco:
@brikis98 correct.

Terragrunt version v0.23.23

brikis98:
Hm, this is indeed a bug! I was able to repro with the following simple test case:

Root `terragrunt.hcl`:

```hcl
prevent_destroy = true
```

In `child/terragrunt.hcl`:

```hcl
include {
  path = find_in_parent_folders()
}

prevent_destroy = false
```

Run `terragrunt apply` in `child`... Then run `terragrunt destroy` in `child` and you get:

```
[terragrunt] [/tmp/ter-prevent/child] 2020/06/16 09:29:25 Running command: terraform --version
[terragrunt] 2020/06/16 09:29:25 Reading Terragrunt config file at /tmp/ter-prevent/child/terragrunt.hcl
[terragrunt] 2020/06/16 09:29:25 Module is protected by the prevent_destroy flag in /tmp/ter-prevent/child/terragrunt.hcl. Set it to false or delete it to allow destroying of the module.
[terragrunt] 2020/06/16 09:29:25 Unable to determine underlying exit code, so Terragrunt will exit with error code 1
```

There must be some bug in the [code that merges child configs with included configs](https://github.com/gruntwork-io/terragrunt/blob/master/config/config.go#L492). Does anyone have time to take a look and submit a PR with the fix?

brikis98:
This was fixed in https://github.com/gruntwork-io/terragrunt/pull/1223 and released in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.23.31.

PR #1223, opened 2020-06-19
Fix the prevent_default flag functionality

alexkayabula:
### What does this PR do?
- This PR provides a fix for issue [1217](https://github.com/gruntwork-io/terragrunt/issues/1217).
### Description of task to be completed?
- The issue was that the ```child``` module configuration was not configured to override the `prevent_destroy` flag in the `parent` `terragrunt.hcl` file for the destroy commands to destroy resources . 

- It is important to not that the `find_in_parent_folders()` helper in the `child` module automatically searches up the directory tree to find the `parent` `terragrunt.hcl` file and inherit the `remote_state` configuration from it. 
- So for the `terragrunt detroy` command to actually destroy resources of the `child` module with `prevent_destroy` flag set to `false`, the `prevent_destroy = true` flag in the `parent` `terragrunt.hcl` file has to be overriden.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -32,7 +32,7 @@ type TerragruntConfig struct {
 	RemoteState                 *remote.RemoteState
 	Dependencies                *ModuleDependencies
 	DownloadDir                 string
-	PreventDestroy              bool
+	PreventDestroy              *bool
 	Skip                        bool
 	IamRole                     string
 	Inputs                      map[string]interface{}
@@ -493,7 +493,8 @@ func mergeConfigWithIncludedConfig(config *TerragruntConfig, includedConfig *Ter
 	if config.RemoteState != nil {
 		includedConfig.RemoteState = config.RemoteState
 	}
-	if config.PreventDestroy {
+
+	if config.PreventDestroy != nil {
 		includedConfig.PreventDestroy = config.PreventDestroy
 	}
 
@@ -730,7 +731,7 @@ func convertToTerragruntConfig(
 	}
 
 	if terragruntConfigFromFile.PreventDestroy != nil {
-		terragruntConfig.PreventDestroy = *terragruntConfigFromFile.PreventDestroy
+		terragruntConfig.PreventDestroy = terragruntConfigFromFile.PreventDestroy
 	}
 
 	if terragruntConfigFromFile.Skip != nil {
--- a/config/config_as_cty.go
+++ b/config/config_as_cty.go
@@ -25,7 +25,7 @@ func terragruntConfigAsCty(config *TerragruntConfig) (cty.Value, error) {
 	output["terragrunt_version_constraint"] = gostringToCty(config.TerragruntVersionConstraint)
 	output["download_dir"] = gostringToCty(config.DownloadDir)
 	output["iam_role"] = gostringToCty(config.IamRole)
-	output["prevent_destroy"] = goboolToCty(config.PreventDestroy)
+	output["prevent_destroy"] = goboolToCty(*config.PreventDestroy)
 	output["skip"] = goboolToCty(config.Skip)
 
 	terraformConfigCty, err := terraformConfigAsCty(config.Terraform)
--- a/config/config_as_cty.go
+++ b/config/config_as_cty.go
@@ -25,7 +25,6 @@ func terragruntConfigAsCty(config *TerragruntConfig) (cty.Value, error) {
 	output["terragrunt_version_constraint"] = gostringToCty(config.TerragruntVersionConstraint)
 	output["download_dir"] = gostringToCty(config.DownloadDir)
 	output["iam_role"] = gostringToCty(config.IamRole)
-	output["prevent_destroy"] = goboolToCty(*config.PreventDestroy)
 	output["skip"] = goboolToCty(config.Skip)
 
 	terraformConfigCty, err := terraformConfigAsCty(config.Terraform)
@@ -50,6 +49,10 @@ func terragruntConfigAsCty(config *TerragruntConfig) (cty.Value, error) {
 	}
 	if dependenciesCty != cty.NilVal {
 		output["dependencies"] = dependenciesCty
+	}
+	
+	if config.PreventDestroy != nil {
+		output["prevent_destroy"] = goboolToCty(*config.PreventDestroy)
 	}
 
 	dependencyCty, err := dependencyBlocksAsCty(config.TerragruntDependencies)
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -73,6 +73,8 @@ const (
 	TEST_FIXTURE_REMOTE_MODULE_IN_ROOT                      = "fixture-download/remote-module-in-root"
 	TEST_FIXTURE_LOCAL_MISSING_BACKEND                      = "fixture-download/local-with-missing-backend"
 	TEST_FIXTURE_LOCAL_WITH_HIDDEN_FOLDER                   = "fixture-download/local-with-hidden-folder"
+	TEST_FIXTURE_PREVENT_DESTROY_OVERRIDE                   = "fixture-prevent-destroy-override/child"
+	TEST_FIXTURE_PREVENT_DESTROY_NOT_SET                    = "fixture-prevent-destroy-not-set/child"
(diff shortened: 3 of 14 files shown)


config/config.go
```diff
@@ -493,7 +493,7 @@ func mergeConfigWithIncludedConfig(config *TerragruntConfig, includedConfig *Ter
 	if config.RemoteState != nil {
 		includedConfig.RemoteState = config.RemoteState
 	}
-	if config.PreventDestroy {
+	if config.PreventDestroy != includedConfig.PreventDestroy{
```
brikis98, line 496: Ah, nice find. One NIT: 

```suggestion
	if config.PreventDestroy != includedConfig.PreventDestroy {
```

alexkayabula:
> Thanks for the PR! Would you be up for adding a regression test that checks this behavior?
> 
> E.g., Create:
> 
>     * `test/fixture-prevent-destroy/terragrunt.hcl` with `prevent_destroy = true`
> 
>     * `test/fixture-prevent-destroy/child/terragrunt.hcl` with `prevent_destroy = true` and `include { path = find_in_parent_folders() }`
> 
>     * `test/fixture-prevent-destroy/child/main.tf` with a `null_resource`
> 
> 
> Then, in test/integration_test.go`, you create a new test method that runs `apply`and then`destroy`in`test/fixture-prevent-destroy/child/` and makes sure it gets destroyed correctly.

Sure, let me add the test

alexkayabula:
> Thanks for the PR! Would you be up for adding a regression test that checks this behavior?
> 
> E.g., Create:
> 
>     * `test/fixture-prevent-destroy/terragrunt.hcl` with `prevent_destroy = true`
> 
>     * `test/fixture-prevent-destroy/child/terragrunt.hcl` with `prevent_destroy = true` and `include { path = find_in_parent_folders() }`
> 
>     * `test/fixture-prevent-destroy/child/main.tf` with a `null_resource`
> 
> 
> Then, in test/integration_test.go`, you create a new test method that runs `apply`and then`destroy`in`test/fixture-prevent-destroy/child/` and makes sure it gets destroyed correctly.

@brikis98, with reference to your repro of the [bug](https://github.com/gruntwork-io/terragrunt/issues/1217#issuecomment-644618776),  
I think * `test/fixture-prevent-destroy/child/terragrunt.hcl` should be created with `prevent_destroy = false` other than `prevent_destroy = true` like you had suggested.

brikis98:
> @brikis98, with reference to the [repro](https://github.com/gruntwork-io/terragrunt/issues/1217#issuecomment-644618776) of the bug,
> I think `test/fixture-prevent-destroy/child/terragrunt.hcl` should be created with `prevent_destroy = false` other than `prevent_destroy = true` like you had suggested.

Yes, sorry, copy/paste error in my comment! Just updated my comment to clear that up.

config/config.go
```diff
@@ -493,7 +493,7 @@ func mergeConfigWithIncludedConfig(config *TerragruntConfig, includedConfig *Ter
 	if config.RemoteState != nil {
 		includedConfig.RemoteState = config.RemoteState
 	}
-	if config.PreventDestroy {
+	if config.PreventDestroy != includedConfig.PreventDestroy {
```
brikis98, line 496: I just a had a realization this might not be the right solution. `PreventDestroy` is `bool`, which means it defaults to `false`, even if the user doesn't set it. That means that if the root config has `prevent_destroy = true`, but the child config _does not have `prevent_destroy` set at all_, then it'll default to `false`, and when this conditional executes, it'll override the root setting with `false`.

I think the real fix here will be to make `PreventDestroy` a pointer to a bool and to differentiate the cases where it's not set at all (`nil`) to where it's explicitly set to `true` or `false`.

test/integration_test.go
```diff
@@ -1311,6 +1312,15 @@ func TestPreventDestroy(t *testing.T) {
 	}
 }
 
+func TestPreventDestroyOverride(t *testing.T) {
```
brikis98, line 1315: Thank you for adding this test case!

Given my comment above, it's prob worth adding an analogous test case, but where:

* The root module sets `prevent_destroy = true`
* The child module doesn't set `prevent_destroy` at all

config/config_as_cty.go
```diff
@@ -25,7 +25,7 @@ func terragruntConfigAsCty(config *TerragruntConfig) (cty.Value, error) {
 	output["terragrunt_version_constraint"] = gostringToCty(config.TerragruntVersionConstraint)
 	output["download_dir"] = gostringToCty(config.DownloadDir)
 	output["iam_role"] = gostringToCty(config.IamRole)
-	output["prevent_destroy"] = goboolToCty(config.PreventDestroy)
+	output["prevent_destroy"] = goboolToCty(*config.PreventDestroy)
```
brikis98, line 28: Does this need a check that it's not `nil`?

alexkayabula: I guess so, thank you for point that out.

config/config_as_cty.go
```diff
@@ -51,6 +50,14 @@ func terragruntConfigAsCty(config *TerragruntConfig) (cty.Value, error) {
...
 		output["dependencies"] = dependenciesCty
 	}
+	
+	preventDestroyCty, err := gostructToCty(config.PreventDestroy)
+	if err != nil {
+		return cty.NilVal, err
+	}
+	if preventDestroyCty != cty.NilVal {
+		output["prevent_destroy"] = preventDestroyCty
+	}
```
brikis98, line 60: A bit simpler, no?
```suggestion
	if config.PreventDestroy != nil {
		output["prevent_destroy"] = goboolToCty(*config.PreventDestroy)
	}
```

brikis98:
Looks like there was a compile error:

```
# github.com/gruntwork-io/terragrunt/config [github.com/gruntwork-io/terragrunt/config.test]
config/config_as_cty_test.go:61:3: cannot use true (type bool) as type *bool in field value
config/config_partial_test.go:32:34: cannot use terragruntConfig.PreventDestroy (type *bool) as type bool in argument to assert.False
config/config_partial_test.go:79:33: cannot use terragruntConfig.PreventDestroy (type *bool) as type bool in argument to assert.True
config/config_partial_test.go:95:34: cannot use terragruntConfig.PreventDestroy (type *bool) as type bool in argument to assert.False
config/config_partial_test.go:122:33: cannot use terragruntConfig.PreventDestroy (type *bool) as type bool in argument to assert.True
config/config_partial_test.go:142:34: cannot use terragruntConfig.PreventDestroy (type *bool) as type bool in argument to assert.False
config/config_test.go:571:21: cannot use cfg.PreventDestroy (type *bool) as type bool in argument to assert.False
?   	github.com/gruntwork-io/terragrunt/aws_helper	[no test files]
```

alexkayabula:
> Looks like there was a compile error:
> 
> ```
> # github.com/gruntwork-io/terragrunt/config [github.com/gruntwork-io/terragrunt/config.test]
> config/config_as_cty_test.go:61:3: cannot use true (type bool) as type *bool in field value
> config/config_partial_test.go:32:34: cannot use terragruntConfig.PreventDestroy (type *bool) as type bool in argument to assert.False
> config/config_partial_test.go:79:33: cannot use terragruntConfig.PreventDestroy (type *bool) as type bool in argument to assert.True
> config/config_partial_test.go:95:34: cannot use terragruntConfig.PreventDestroy (type *bool) as type bool in argument to assert.False
> config/config_partial_test.go:122:33: cannot use terragruntConfig.PreventDestroy (type *bool) as type bool in argument to assert.True
> config/config_partial_test.go:142:34: cannot use terragruntConfig.PreventDestroy (type *bool) as type bool in argument to assert.False
> config/config_test.go:571:21: cannot use cfg.PreventDestroy (type *bool) as type bool in argument to assert.False
> ?   	github.com/gruntwork-io/terragrunt	[no test files]
> ?   	github.com/gruntwork-io/terragrunt/aws_helper	[no test files]
> ```

Let me check it out

brikis98:
Kicking off tests again!

brikis98:

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.23.31

PR #1227, opened 2020-06-19
Add Terraform beta releases to version checking

### LINKED ISSUE #1224: Terraform beta version support
Upon encountering a terraform version different from normal releases or dev versions, terragrunt exits, likely due to https://github.com/gruntwork-io/terragrunt/blob/master/cli/version_check.go#L16

i.e.

```
$: TG_LOG=DEBUG terragrunt init
[terragrunt] 2020/06/19 12:55:21 Terragrunt Version: v0.23.27
[terragrunt] 2020/06/19 12:55:21 Found locals block: evaluating the expressions.
...
[terragrunt] [dir] 2020/06/19 12:55:21 Running command: terraform --version
[terragrunt] 2020/06/19 12:55:21 exit status 1
$: terraform version
Terraform v0.13.0-beta2
```

Something like this should probably fix this: 
```
var TERRAFORM_VERSION_REGEX = regexp.MustCompile("Terraform (v?[\\d\\.]+)(?:-dev)?(?:-beta\\d+)?(?: .+)?")
```

artemsablin:
Fixes #1224

https://gist.github.com/artemsablin/914fa8fb3cbf479d9023f75920a6cc6c

NB: It won't work with v0.13.0-beta2 due to a bug with terraform's  `--version` (see: https://github.com/hashicorp/terraform/issues/25276 - this is already fixed).

### DIFF
--- a/cli/version_check.go
+++ b/cli/version_check.go
@@ -11,9 +11,12 @@ import (
 	"github.com/hashicorp/go-version"
 )
 
-// The terraform --version output is of the format: Terraform v0.9.5-dev (cad024a5fe131a546936674ef85445215bbc4226+CHANGES)
-// where -dev and (commitid+CHANGES) is for custom builds or if TF_LOG is set for debug purposes
-var TERRAFORM_VERSION_REGEX = regexp.MustCompile("Terraform (v?[\\d\\.]+)(?:-dev)?(?: .+)?")
+// TerraformVersionRegex verifies that terraform --version output is in one of the following formats:
+// - Terraform v0.9.5-dev (cad024a5fe131a546936674ef85445215bbc4226+CHANGES)
+// - Terraform v0.13.0-beta2
+// - Terraform v0.12.27
+// We only make sure the "v#.#.#" part is present in the output.
+var TerraformVersionRegex = regexp.MustCompile(`Terraform (v\d+\.\d+\.\d+).*`)
 
 // Populate the currently installed version of Terraform into the given terragruntOptions
 func PopulateTerraformVersion(terragruntOptions *options.TerragruntOptions) error {
@@ -78,7 +81,7 @@ func checkTerraformVersionMeetsConstraint(currentVersion *version.Version, const
 
 // Parse the output of the terraform --version command
 func parseTerraformVersion(versionCommandOutput string) (*version.Version, error) {
-	matches := TERRAFORM_VERSION_REGEX.FindStringSubmatch(versionCommandOutput)
+	matches := TerraformVersionRegex.FindStringSubmatch(versionCommandOutput)
 
 	if len(matches) != 2 {
 		return nil, errors.WithStackTrace(InvalidTerraformVersionSyntax(versionCommandOutput))
--- a/cli/version_check.go
+++ b/cli/version_check.go
@@ -16,7 +16,7 @@ import (
 // - Terraform v0.13.0-beta2
 // - Terraform v0.12.27
 // We only make sure the "v#.#.#" part is present in the output.
-var TerraformVersionRegex = regexp.MustCompile(`Terraform (v\d+\.\d+\.\d+).*`)
+var TerraformVersionRegex = regexp.MustCompile(`Terraform (v?\d+\.\d+\.\d+).*`)
 
 // Populate the currently installed version of Terraform into the given terragruntOptions
 func PopulateTerraformVersion(terragruntOptions *options.TerragruntOptions) error {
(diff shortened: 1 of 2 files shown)


cli/version_check.go
```diff
@@ -13,7 +13,7 @@ import (
 
 // The terraform --version output is of the format: Terraform v0.9.5-dev (cad024a5fe131a546936674ef85445215bbc4226+CHANGES)
 // where -dev and (commitid+CHANGES) is for custom builds or if TF_LOG is set for debug purposes
-var TERRAFORM_VERSION_REGEX = regexp.MustCompile("Terraform (v?[\\d\\.]+)(?:-dev)?(?: .+)?")
+var TERRAFORM_VERSION_REGEX = regexp.MustCompile("Terraform (v?[\\d\\.]+)(?:-dev)?(?:-beta\\d+)?(?: .+)?")
```
brikis98, line 16: Is it worth matching exactly what comes after the dash? Before it was `-dev`; today it's `-beta`; in the future, it might be something else. It seems like we could make this slightly more maintainable by ignoring what comes after the dash, right?

artemsablin: That's a good point, I'll update the PR later this day.

artemsablin: https://gist.github.com/artemsablin/d56a2b38421751374782814dddba9b36

cli/version_check.go
```diff
@@ -13,7 +13,7 @@ import (
 
 // The terraform --version output is of the format: Terraform v0.9.5-dev (cad024a5fe131a546936674ef85445215bbc4226+CHANGES)
 // where -dev and (commitid+CHANGES) is for custom builds or if TF_LOG is set for debug purposes
-var TERRAFORM_VERSION_REGEX = regexp.MustCompile("Terraform (v?[\\d\\.]+)(?:-dev)?(?: .+)?")
+var TERRAFORM_VERSION_REGEX = regexp.MustCompile("Terraform (v?[\\d\\.]+)(?:.*)")
```
brikis98, line 16: Looking at this again, I don't think this regex is quite correct. The `[\\d\\.]+` portion accepts any combination of digits or dots. E.g., `v33333`, `v.....`, and `v3....3......3` would all be valid. Also, two NITs: the final `(?:.*)` can be simplified to just `.*`. and if we use backticks to wrap the string instead of double quotes, we won't need nearly as much escaping. 

Putting that all together, we probably want a regex that looks something more like this:

```go
var TERRAFORM_VERSION_REGEX = regexp.MustCompile(`Terraform (v?(?:\d\.)*\d).*`)
```

cli/version_check.go
```diff
@@ -13,7 +13,7 @@ import (
 
 // The terraform --version output is of the format: Terraform v0.9.5-dev (cad024a5fe131a546936674ef85445215bbc4226+CHANGES)
 // where -dev and (commitid+CHANGES) is for custom builds or if TF_LOG is set for debug purposes
```
brikis98, line 15: Could you update this comment too?

cli/version_check.go
artemsablin: I think since terraform [adheres](https://www.terraform.io/docs/extend/best-practices/versioning.html) to semver it may be cleaner to check for `MAJOR.MINOR.PATCH` numbers specifically rather than rely on a quantifier.

cli/version_check.go
artemsablin: Done!

brikis98:
Looks like one unit test failed:

```
--- FAIL: TestParseTerraformVersionWithoutV (0.00s)
    version_check_test.go:99: 
        	Error Trace:	version_check_test.go:99
        	            				version_check_test.go:44
        	Error:      	Expected nil, but got: &errors.Error{Err:"Terraform 0.9.3", stack:[]uintptr{0x10e7103, 0x10f783d, 0x10f715b, 0x511e69, 0x460f11}, frames:[]errors.StackFrame(nil), prefix:""}
        	Test:       	TestParseTerraformVersionWithoutV
    version_check_test.go:100: 
        	Error Trace:	version_check_test.go:100
        	            				version_check_test.go:44
        	Error:      	Not equal: 
        	            	expected: &version.Version{metadata:"", pre:"", segments:[]int64{0, 9, 3}, si:3, original:"0.9.3"}
        	            	actual  : (*version.Version)(nil)
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,2 +1,2 @@
        	            	-(*version.Version)(0.9.3)
        	            	+(*version.Version)(<nil>)
        	            	 
        	Test:       	TestParseTerraformVersionWithoutV
```

artemsablin:
https://gist.github.com/artemsablin/015e3f94f519d6277cd3c9f58c647cd4 Forgot to run tests last time, this time should be fine.

brikis98:
Thx, kicking off tests again...

brikis98:
OK, tests passed! Merging now, thanks!

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.23.30

Issue #1040, opened 2020-02-10
Add the possibility to customize the RETRYABLE_ERRORS using environment variables


andreluiznsilva:
My CI/CD sometimes fail because of a "ssh_exchange_identification: Connection closed by remote host" error. 

Would be nice to include the error on the RETRYABLE_ERRORS list and/or allow the list to be customized by a environment variable.

Regards

brikis98:
A big +1 on this. In fact, it would be great to be able to add to the list of retryable errors in `terragrunt.hcl`: 

```hcl
retryable_errors = {
  "a regex to match the error": "more information about the error"
}
```

A PR for this is very welcome.

PR #1231, opened 2020-06-23
Docs: make retryable errors configurable

infraredgirl:
Related issue: #1040.

PR #1232, opened 2020-06-23
Make retryable errors configurable

infraredgirl:
### Description

This change enables users to create their own custom list of retryable errors in the `terragrunt.hcl`.

### Related issue
Closes: #1040

### How to test

To test this you can use these 2 files:
`main.tf`:

```tcl
resource "null_resource" "example" {
  triggers = {
    always_recreate = timestamp()
  }

provisioner "local-exec" {
    command = "echo 'My own little error' && exit 1"
  }
}

output "text" {
  value = "Hello, World"
}
```

`terragrunt.hcl`:
```tcl
retryable_errors = [
"My own little error"
]
```

Then just run `go run main.go apply` and the error which is specified in the config file should be detected as retryable and retries attempted. To make sure other errors are not picked up as retryable, edit the `echo` statement in `main.tf` to output a different error message, and re-run `go run main.go apply`. Now `terragrunt` should just fail without attempting retries.

### Known issues / still TBD:

- [x] Add unit tests
- [x] Add integration tests
~- [ ] Change the format of the retryable errors structure. Right now it's a list of regex strings and we want a map of regexes and error descriptions (as described in the [issue](#1040)).~ We decided to stick with using the list for now.
- [x] Make the config option optional (right now terragrunt errors out if this option is not set in the config file; it should fall back to using the default hardcoded value instead)

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -326,6 +326,11 @@ func RunTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 		terragruntOptions.DownloadDir = terragruntConfig.DownloadDir
 	}
 
+	// Override the default value of retryable errors using the value set in the config file
+	if terragruntConfig.RetryableErrors != nil {
+		terragruntOptions.RetryableErrors = terragruntConfig.RetryableErrors
+	}
+
 	if sourceUrl := getTerraformSourceUrl(terragruntOptions, terragruntConfig); sourceUrl != "" {
 		if err := downloadTerraformSource(sourceUrl, terragruntOptions, terragruntConfig); err != nil {
 			return err
--- a/config/config_as_cty.go
+++ b/config/config_as_cty.go
@@ -71,6 +71,14 @@ func terragruntConfigAsCty(config *TerragruntConfig) (cty.Value, error) {
 		output["generate"] = generateCty
 	}
 
+	retryableCty, err := gostructToCty(config.RetryableErrors)
+	if err != nil {
+		return cty.NilVal, err
+	}
+	if retryableCty != cty.NilVal {
+		output["retryable_errors"] = retryableCty
+	}
+
 	inputsCty, err := convertToCtyWithJson(config.Inputs)
 	if err != nil {
 		return cty.NilVal, err
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -53,7 +53,7 @@ func TestParseTerragruntJsonConfigRemoteStateMinimalConfig(t *testing.T) {
 	require.NoError(t, err)
 
 	assert.Nil(t, terragruntConfig.Terraform)
-
+	assert.Nil(t, terragruntConfig.RetryableErrors)
 	assert.Empty(t, terragruntConfig.IamRole)
 
 	if assert.NotNil(t, terragruntConfig.RemoteState) {
@@ -109,7 +109,7 @@ remote_state {
 	}
 
 	assert.Nil(t, terragruntConfig.Terraform)
-
+	assert.Nil(t, terragruntConfig.RetryableErrors)
 	assert.Empty(t, terragruntConfig.IamRole)
 
 	if assert.NotNil(t, terragruntConfig.RemoteState) {
@@ -145,7 +145,7 @@ func TestParseTerragruntJsonConfigRemoteStateFullConfig(t *testing.T) {
 	}
 
 	assert.Nil(t, terragruntConfig.Terraform)
-
+	assert.Nil(t, terragruntConfig.RetryableErrors)
 	assert.Empty(t, terragruntConfig.IamRole)
 
 	if assert.NotNil(t, terragruntConfig.RemoteState) {
@@ -155,6 +155,48 @@ func TestParseTerragruntJsonConfigRemoteStateFullConfig(t *testing.T) {
 		assert.Equal(t, "my-bucket", terragruntConfig.RemoteState.Config["bucket"])
 		assert.Equal(t, "terraform.tfstate", terragruntConfig.RemoteState.Config["key"])
 		assert.Equal(t, "us-east-1", terragruntConfig.RemoteState.Config["region"])
+	}
+}
+
+func TestParseTerragruntHclConfigRetryableErrors(t *testing.T) {
+	t.Parallel()
+
+	config := `
+retryable_errors = [
+    "My own little error",
+    "Another one of my errors"
+]
+`
+	terragruntConfig, err := ParseConfigString(config, mockOptionsForTest(t), nil, DefaultTerragruntConfigPath)
+	require.NoError(t, err)
+
+	assert.Nil(t, terragruntConfig.Terraform)
+	assert.Empty(t, terragruntConfig.IamRole)
+
+	if assert.NotNil(t, terragruntConfig.RetryableErrors) {
+		assert.Equal(t, []string{"My own little error", "Another one of my errors"}, terragruntConfig.RetryableErrors)
+	}
+}
+
+func TestParseTerragruntJsonConfigRetryableErrors(t *testing.T) {
+	t.Parallel()
+
+	config := `
+{
+	"retryable_errors": [
+        "My own little error"
+	]
+}
+`
+
+	terragruntConfig, err := ParseConfigString(config, mockOptionsForTest(t), nil, DefaultTerragruntJsonConfigPath)
+	require.NoError(t, err)
+
+	assert.Nil(t, terragruntConfig.Terraform)
+	assert.Empty(t, terragruntConfig.IamRole)
+
+	if assert.NotNil(t, terragruntConfig.RetryableErrors) {
+		assert.Equal(t, []string{"My own little error"}, terragruntConfig.RetryableErrors)
 	}
 }
 
@@ -171,6 +213,7 @@ func TestParseIamRole(t *testing.T) {
 	assert.Nil(t, terragruntConfig.RemoteState)
 	assert.Nil(t, terragruntConfig.Terraform)
 	assert.Nil(t, terragruntConfig.Dependencies)
+	assert.Nil(t, terragruntConfig.RetryableErrors)
 
 	assert.Equal(t, "terragrunt-iam-role", terragruntConfig.IamRole)
 }
@@ -191,6 +234,7 @@ dependencies {
 
 	assert.Nil(t, terragruntConfig.RemoteState)
 	assert.Nil(t, terragruntConfig.Terraform)
+	assert.Nil(t, terragruntConfig.RetryableErrors)
 
 	assert.Empty(t, terragruntConfig.IamRole)
 
@@ -215,7 +259,7 @@ dependencies {
 
 	assert.Nil(t, terragruntConfig.RemoteState)
 	assert.Nil(t, terragruntConfig.Terraform)
-
+	assert.Nil(t, terragruntConfig.RetryableErrors)
 	assert.Empty(t, terragruntConfig.IamRole)
 
 	if assert.NotNil(t, terragruntConfig.Dependencies) {
@@ -254,7 +298,7 @@ dependencies {
 	require.NotNil(t, terragruntConfig.Terraform)
 	require.NotNil(t, terragruntConfig.Terraform.Source)
 	assert.Equal(t, "foo", *terragruntConfig.Terraform.Source)
-
+	assert.Nil(t, terragruntConfig.RetryableErrors)
 	assert.Empty(t, terragruntConfig.IamRole)
 
 	if assert.NotNil(t, terragruntConfig.RemoteState) {
@@ -302,7 +346,7 @@ func TestParseTerragruntJsonConfigRemoteStateDynamoDbTerraformConfigAndDependenc
 	require.NotNil(t, terragruntConfig.Terraform)
 	require.NotNil(t, terragruntConfig.Terraform.Source)
 	assert.Equal(t, "foo", *terragruntConfig.Terraform.Source)
-
+	assert.Nil(t, terragruntConfig.RetryableErrors)
 	assert.Empty(t, terragruntConfig.IamRole)
 
 	if assert.NotNil(t, terragruntConfig.RemoteState) {
--- a/test/fixture-auto-retry/custom-errors/script.sh
+++ b/test/fixture-auto-retry/custom-errors/script.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+FILENAME="/tmp/$1"
+
+if test -f "$FILENAME"; then
+  echo "Success"
+  rm "$FILENAME"
+  exit 0
+else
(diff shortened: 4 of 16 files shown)


config/config_test.go
```diff
@@ -158,6 +158,57 @@ func TestParseTerragruntJsonConfigRemoteStateFullConfig(t *testing.T) {
...
+	config := `
+retryable_errors = [
+    "My own little error",
+    "Another one of my errors"
+]
+`
+	terragruntConfig, err := ParseConfigString(config, mockOptionsForTest(t), nil, DefaultTerragruntConfigPath)
+	if err != nil {
+		t.Fatal(err)
+	}
```
brikis98, line 173: NIT: `require.NoError(t, err)` is slightly shorter/simpler.

config/config_test.go
```diff
@@ -158,6 +158,57 @@ func TestParseTerragruntJsonConfigRemoteStateFullConfig(t *testing.T) {
...
+
+	assert.Nil(t, terragruntConfig.Terraform)
+	assert.Empty(t, terragruntConfig.IamRole)
+
+	if assert.NotNil(t, terragruntConfig.RetryableErrors) {
+		assert.NotEmpty(t, terragruntConfig.RetryableErrors)
+		assert.Equal(t, "My own little error", terragruntConfig.RetryableErrors[0])
+		assert.Equal(t, "Another one of my errors", terragruntConfig.RetryableErrors[1])
```
brikis98, line 181: NIT: I think the following will do the same thing, but shorter, and with clearer error reporting:

```suggestion
		assert.Equal(t, []string{"My own little error", "Another one of my errors"}, terragruntConfig.RetryableErrors)
```

config/config_test.go
```diff
@@ -158,6 +158,57 @@ func TestParseTerragruntJsonConfigRemoteStateFullConfig(t *testing.T) {
...
+	}
+}
+
+func TestParseTerragruntJsonConfigRetryableErrors(t *testing.T) {
```
brikis98, line 185: Same comments as above

config/config_test.go
```diff
@@ -158,6 +158,57 @@ func TestParseTerragruntJsonConfigRemoteStateFullConfig(t *testing.T) {
...
+	t.Parallel()
+
+	config := `
+{
+	"retryable_errors": [
+        "My own little error",
+        "Another one of my errors"
```
brikis98, line 192: Consider a list of 1 item or 3 items just so it's not exactly the same as the HCL version and exercises a few more code paths

config/config_test.go
```diff
@@ -158,6 +158,57 @@ func TestParseTerragruntJsonConfigRemoteStateFullConfig(t *testing.T) {
 	}
 }
 
+func TestParseTerragruntHclConfigRetryableErrors(t *testing.T) {
```
brikis98, line 161: Consider updating some of the other existing test cases to check that `terragruntConfig.RetryableErrors` is `nil`.

infraredgirl:
Thanks so much - updated the PR to address the comments.

Last remaining item is integration test - looking into that now!

test/integration_test.go
```diff
@@ -1170,6 +1171,19 @@ func TestAutoRetryExhaustRetries(t *testing.T) {
...
+	t.Parallel()
+
+	out := new(bytes.Buffer)
+	rootPath := copyEnvironment(t, TEST_FIXTURE_AUTO_RETRY_CUSTOM_ERRORS)
+	modulePath := util.JoinPath(rootPath, TEST_FIXTURE_AUTO_RETRY_CUSTOM_ERRORS)
+	err := runTerragruntCommand(t, fmt.Sprintf("terragrunt apply --auto-approve --terragrunt-non-interactive --terragrunt-working-dir %s", modulePath), out, os.Stderr)
+
+	assert.NotNil(t, err)
+	assert.Contains(t, out.String(), "My own little error")
+	assert.NotContains(t, out.String(), "Apply complete!")
```
brikis98, line 1184: This checks that the `apply` eventually failed, but how do we know this actually did some retries? 

You probably want to test a fixture that fails the first N items but then succeeds on try N+1. Here's one approach for doing that:

1. Update the Terraform test fixture to:

    1. Take in a `url` variable.
    1. Use the [`http`](https://www.terraform.io/docs/providers/http/data_source.html) data source to make an HTTP request to `url`.
    1. Update the Bash script in the `null_resource` to exit with an error if the response the `http` data source gets is not a 200 OK.
    1.  Add `triggers { run_always = timestamp() }` to the `null_resource` so it runs every time you run `apply`, and not just on the very first time.

1. Update the integration test to:

    1. Fire up a [local HTTP server](https://golang.org/pkg/net/http/) listening on some port.
    1. Configure that HTTP server to return a 500 error on the first N requests, but a 200 OK on request N+1.
    1. Run `terragrunt apply -var url=http://localhost:<PORT>`, make sure the logs mention errors/retries, but that the `apply` ultimately succeeds.

Let me know if that makes sense!

infraredgirl: It makes sense but I wonder if firing up an HTTP server is a bit of an overkill - maybe we can do it by keeping the state in a file in `/tmp` (like we discussed this morning). What do you think?

brikis98: Yup, you could use a file-based approach too, but it might be a bit more error prone as files have to be cleaned up so that the outputs from previous test runs don't cause issues with subsequent test runs. To do it with files, you'd do something like this:

1. Use the [`random_id`](https://www.terraform.io/docs/providers/random/r/id.html) resource to generate a unique filename on the very first `apply`.

1. Update the Bash script in the `null_resource` to: 

    1. Check if the file with the name from `random_id` exists in `/tmp`. 
    1. If it does not, write it to disk, and exit with an error. 
    1. If it does, exit without an error.

1. Add `triggers { run_always = timestamp() }` to the `null_resource` so it runs every time you run `apply`, and not just on the very first time.

This way, the module will fail the first time you run `apply`, but succeed after that.

cli/cli_app.go
```diff
@@ -319,6 +319,11 @@ func RunTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 		terragruntOptions.DownloadDir = terragruntConfig.DownloadDir
 	}
 
+	// Override the default value of retryable errors using the value set in the config file
+	if terragruntConfig.RetryableErrors != nil {
+		terragruntOptions.RetryableErrors = terragruntConfig.RetryableErrors
```
yorinasub17, line 324: Not for this PR, but we should probably expose a helper function that returns all the default retryable errors. I think 90% of use cases for this will want to append to the default list, rather than replace. If we had a helper (e.g. `get_default_retryable_errors`), then one could do the append use case with:

```
retryable_errors = concat(get_default_retryable_errors(), ["my special snowflake"])
```

config/config_as_cty.go
```diff
@@ -68,6 +68,14 @@ func terragruntConfigAsCty(config *TerragruntConfig) (cty.Value, error) {
 		output["generate"] = generateCty
 	}
 
+	retryableCty, err := gostructToCty(config.RetryableErrors)
```
yorinasub17, line 71: This was at first confusing to me because `RetryableErrors` is not a struct, but then I realized that `ImpliedType` "does the right thing." We should probably rename `gostructToCty` eventually...

yorinasub17 (dismissed):
Overall implementation LGTM! Only had 2 minor comments for future improvements.

test/integration_test.go
infraredgirl: Awesome - done!

cli/cli_app.go
infraredgirl: Agreed, that makes total sense.

config/config_as_cty.go
infraredgirl: I went through the very same thought process... :)

test/fixture-auto-retry/custom-errors/script.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+FILENAME=/tmp/$1
+
+if test -f "$FILENAME"; then
+  echo "Success"
+  rm $FILENAME
```
brikis98, line 7: Oh, haha, cleaning up the file on success is clever 😁

test/fixture-auto-retry/custom-errors/script.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+FILENAME=/tmp/$1
```
brikis98, line 3: NIT: we mostly follow the [Google shell style guide](https://google.github.io/styleguide/shellguide.html). Same for a few other suggestions below.

```suggestion
FILENAME="/tmp/$1"
```

test/fixture-auto-retry/custom-errors/script.sh
brikis98, line 7:
```suggestion
  rm "$FILENAME"
```

test/fixture-auto-retry/custom-errors/script.sh
```diff
@@ -0,0 +1,13 @@
...
+
+if test -f "$FILENAME"; then
+  echo "Success"
+  rm $FILENAME
+  exit 0
+else
+  touch $FILENAME
```
brikis98, line 10:
```suggestion
  touch "$FILENAME"
```

test/integration_test.go
```diff
@@ -1170,6 +1171,19 @@ func TestAutoRetryExhaustRetries(t *testing.T) {
 	assert.NotContains(t, out.String(), "Apply complete!")
 }
 
+func TestAutoRetryCustomRetryableErrors(t *testing.T) {
```
brikis98, line 1174: Prob worth adding another test that uses the same module, but does _not_ set retryable errors to make sure that test fails!

infraredgirl: Addressed everything but this one - added a TODO as discussed on Slack

infraredgirl:
Marking as WIP while still working on this.

Issues left to address:

- [ ] Rebase and fix conflicts
- [ ] Add docs from PR #1231 to this PR and close the other PR
- [ ] Add the missing integration test (see [this comment](https://github.com/gruntwork-io/terragrunt/pull/1232#discussion_r444320783))

test/integration_test.go
infraredgirl: This is now addressed in 7dfdc8c.

cli/cli_app.go
brikis98: @infraredgirl Could you file an issue to track this?

config/config_as_cty.go
brikis98: Might want a GitHub issue for this too.

brikis98 (approved):
LGTM! I think this is ready to merge if the tests are passing :)

brikis98:
@infraredgirl Feel free to merge, do a release, and add to our newsletter!

cli/cli_app.go
infraredgirl: Done: https://github.com/gruntwork-io/terragrunt/issues/1383

config/config_as_cty.go
infraredgirl: Done: https://github.com/gruntwork-io/terragrunt/issues/1384

infraredgirl:
> LGTM! I think this is ready to merge if the tests are passing :)

Thanks for the review @brikis98! Tests are passing for me locally. `terragrunt` doesn't seem to have any automated tests/CI hooked up to run on PR submissions AFAICS, so are we good to merge when tests pass locally?

brikis98:
Ohhh, it does have tests, but they don't run for PRs submitted by people outside our org. It's weird that it doesn't update that after you've joined the org... Perhaps some GitHub bug/limit?

Might be a good chance to practice with this tool that we use to trigger test runs for PRs submitted by people outside the org: https://github.com/gruntwork-io/prototypes/tree/master/testpr

infraredgirl:
Ah, it's probably because the initial commits in this PR were submitted with my old/non-Gruntwork email. Thanks for the pointer - kicked off the CI build now. Looks like we have some tests failures - looking into those now.

infraredgirl:
Unit tests are now passing however integration tests seem to be failing on something unrelated to this PR. I can see from previous PR runs that the same 16 integration tests are failing starting 2 days ago. The errors are all the same: `Failed to list object versions in s3 bucket terragrunt-test-bucket-djstsa: NoSuchBucket: The specified bucket does not exist`. Any ideas about this?

brikis98:
> Unit tests are now passing however integration tests seem to be failing on something unrelated to this PR. I can see from previous CI runs that the same 16 integration tests are failing starting 2 days ago. The errors are all the same: `Failed to list object versions in s3 bucket terragrunt-test-bucket-djstsa: NoSuchBucket: The specified bucket does not exist`. Any ideas about this?

Might be a bit of a red herring... 

Do the tests pass when you run one of them at a time locally? If not, could you share the full log output?

infraredgirl:
Alright, after a few yak shaves (thanks @brikis98 for the help debugging) we have a passing test!

Can I have a final approval please? :)

brikis98 (approved):
Ship it! 🎉

infraredgirl:
Merged, released and updated newsletter!

Fran-ELS-AMS:
Is there currently any way to add `retryable_errors` to the default hard-coded list, instead of replacing the whole list?

Currently we're copying the whole default list from here and adding our custom errors, but if the default list gets updated we have to manually update our code

infraredgirl:
> Is there currently any way to add `retryable_errors` to the default hard-coded list, instead of replacing the whole list?
> 
> Currently we're copying the whole default list from here and adding our custom errors, but if the default list gets updated we have to manually update our code

This is not currently supported but we do plan to add it in the future - in fact we already have an open issue for this very feature: https://github.com/gruntwork-io/terragrunt/issues/1383.

Fran-ELS-AMS:
> > Is there currently any way to add `retryable_errors` to the default hard-coded list, instead of replacing the whole list?
> > Currently we're copying the whole default list from here and adding our custom errors, but if the default list gets updated we have to manually update our code
> 
> This is not currently supported but we do plan to add it in the future - in fact we already have an open issue for this very feature: #1383.

Great, thanks a lot! 🙇

jubr:
This cool new feature got me 🤔: could a similar approach also work for the
```
Error: Could not load plugin
Plugin reinitialization required. Please run "terraform init".
```
use case? I'm running into this quite regularly in my dev workflow.

infraredgirl:
> This cool new feature got me 🤔: could a similar approach also work for the
> 
> ```
> Error: Could not load plugin
> Plugin reinitialization required. Please run "terraform init".
> ```
> 
> use case? I'm running into this quite regularly in my dev workflow.

In other words, terragrunt should automatically run `init` when needed. I think it's worth considering - @jubr would you mind opening an issue for this and we can take it from there?

jubr:
@infraredgirl ✅ #1414

Issue #1166, opened 2020-05-06
--detailed-exitcode bad behavior (v0.23.2)


Thomvaill:
When we pass `--detailed-exitcode` to `terragrunt plan-all`:

- If there is no diff: everything works fine
- If there is a diff: the command has an exit code 2, as intended, but its output says it encountered some errors

```
➜  git:(master) ✗ terragrunt plan-all --detailed-exitcode
[...]
[terragrunt] 2020/05/06 16:25:31 Encountered the following errors:
Hit multiple errors:
exit status 2
➜  git:(master) ✗ echo $?
2
➜  dummy-product-infrastructure git:(master) ✗ terragrunt -v
terragrunt version v0.23.2
```

It should not appear as an error.

Related issues:

- https://github.com/gruntwork-io/terragrunt/issues/451
- https://github.com/gruntwork-io/terragrunt/issues/37

brikis98:
Ah, yea, Terragrunt just assumes any non zero exit code is an error in its logging, which is not the right thing to do for some Terraform commands. Would you be up for a PR to fix this?

loreleimccollum-work:
We are hitting this issue, because when you use --detailed-exitcode the after_hooks do not run

ahmad-hamade:
I had a requirements to get the detailed exit code for detecting the drifts in our automation system and below is the steps I followed to achieve that:

1. Override the default terraform run using [`terraform_binary`](https://terragrunt.gruntwork.io/docs/reference/config-blocks-and-attributes/#terraform_binary)
```hcl
locals {
  run_in_automation = tobool(get_env("TF_IN_AUTOMATION", false))
  tfplan_path = format("%s/%s", format("%s/../../tmp", get_parent_terragrunt_dir()), uuid())
}
terraform_binary = format("%s/../../scripts/terraform-wrapper.sh", get_parent_terragrunt_dir())

terraform {
  extra_arguments "tfplan_out" {
    commands = ["plan"]
    arguments = local.run_in_automation ? [
      path_relative_to_include(),
      local.tfplan_path,
      "-lock=false",
      "-no-color",
      "-detailed-exitcode",
      "-compact-warnings"
    ] : []
  }
}
```

In the root of your repo, create `terraform-wrapper.sh` script that containes the following:

```bash
#!/usr/bin/env bash

function suppress_tf() {
  local action=$1
  shift 1
  command terraform "$action" "$@" 2>&1
[...]

Issue #451, opened 2018-03-28
Regression in --detailed-exitcode with Terragrunt v0.14.4 and later


geekifier:
It appears that starting with the v0.14.4 release, the --detailed-exitcode behavior is broken.

```bash
# terragrunt version v0.14.2
# Terraform v0.11.5

$ terragrunt plan --detailed-exitcode
[terragrunt] 2018/03/28 09:14:13 exit status 2
$ echo $?
2
```

```bash
# terragrunt version v0.14.4
# Terraform v0.11.5

$ terragrunt plan --detailed-exitcode
[terragrunt] 2018/03/28 09:12:52 Hit multiple errors:
exit status 2
[terragrunt] 2018/03/28 09:12:52 Unable to determine underlying exit code, so Terragrunt will exit with error code 1
$ echo $?
1
```

Please let me know if you'd like more verbose logging.
Both plan commands produce a valid plan with no obvious errors.
I also verified that the same behavior occurs in v0.14.5, and absent in v0.14.3.

brikis98:
Ah, you're right. It's because we now are wrapping the error: https://github.com/gruntwork-io/terragrunt/blob/master/cli/cli_app.go#L302-L306. I'll see if I can get a fix out.

geekifier:
Thanks for the verification. A side question that I have after this: have you considered following semantic versioning for the releases?

Generally, going from version 0.14.3 -> 0.14.4 I would only expect to see bugfixes with a small chance of breakage.

For new features, I'd expect to see the middle number bumped up. E.g. 0.14.3 -> 0.15.0 or whatever.

I think there is some value there from the end-user perspective - I am curious to see if you have given this any thought.

brikis98:
Try https://github.com/gruntwork-io/terragrunt/releases/tag/v0.14.6 for the fix in a few minutes, once the binaries build.

> A side question that I have after this: have you considered following semantic versioning for the releases?

We do. It's just that we are pre-1.0.0, so the rules are a bit different. We don't use the major version number (it's always 0), so we bump the patch version for anything backwards compatible and the minor version for things that are backwards incompatible. This new feature was *supposed* to be backwards compatible, but bugs do happen, and are always backwards incompatible :)

geekifier:
I can confirm that the Terraform exit code is being passed on properly in `0.14.6`.

I think that there is side story here for better handling of the -detailed-exitcode flag.

[...]

PR #1234, opened 2020-06-24
Fix uncharacteristic behaviour of the detailed exitcode flag 

alexkayabula:
### What does this PR do?
- This PR provides a fix for issues 
[1166](https://github.com/gruntwork-io/terragrunt/issues/1166)
[451](https://github.com/gruntwork-io/terragrunt/issues/451)
### Description of task to be completed?
- Fix the issue where If there is a diff: the command `terragrunt plan-all --detailed-exitcode` returns an exit code 2, as [intended](https://www.terraform.io/docs/commands/plan.html#detailed-exitcode), but its output says it encountered some errors.


configstack/running_module.go
```diff
@@ -249,7 +249,7 @@ func (module *runningModule) moduleFinished(moduleErr error) {
 	if moduleErr == nil {
 		module.Module.TerragruntOptions.Logger.Printf("Module %s has finished successfully!", module.Module.Path)
 	} else {
-		module.Module.TerragruntOptions.Logger.Printf("Module %s has finished with an error: %v", module.Module.Path, moduleErr)
+		module.Module.TerragruntOptions.Logger.Printf("Module %s has finished with status: %v", module.Module.Path, moduleErr)
```
brikis98, line 252: This doesn't seem like the right approach, as this will affect all commands/args, and not just `plan-all --detailed-exit-code`.

configstack/running_module.go
```diff
@@ -286,9 +286,11 @@ type MultiError struct {
 func (err MultiError) Error() string {
 	errorStrings := []string{}
 	for _, err := range err.Errors {
-		errorStrings = append(errorStrings, err.Error())
+		if err.Error() == "exit status 1" {
+            errorStrings = append(errorStrings, err.Error())
+		}
```
brikis98, line 291: Again, this affects all commands/args, not just `plan-all --detailed-exit-code`, so this change runs the risk of silently swallowing legitimate errors.

configstack/running_module.go
```diff
@@ -286,9 +286,11 @@ type MultiError struct {
 func (err MultiError) Error() string {
```
brikis98, line 286: I didn't realize we had two `MultiError` structs, one in `configstack` and one in `errors`. We should consolidate on just one of them; probably the latter.

alexkayabula: I have refactored accordingly.

configstack/running_module.go
alexkayabula: I was of the view that since [exit status 2](https://www.terraform.io/docs/commands/plan.html#detailed-exitcode) is not necessarily an error, we should only return it. 
Logging it with the message `Encountered the following errors: Hit multiple errors: exit status 2` is misleading.
What do you think @brikis98 ?

configstack/running_module.go
```diff
@@ -279,19 +278,7 @@ func (this DependencyFinishedWithError) ExitStatus() (int, error) {
...
-func (err MultiError) Error() string {
-	errorStrings := []string{}
-	for _, err := range err.Errors {
-		errorStrings = append(errorStrings, err.Error())
-	}
-	return fmt.Sprintf("Encountered the following errors:\n%s", strings.Join(errorStrings, "\n"))
-}
-
-func (this MultiError) ExitStatus() (int, error) {
+func ExitStatus(this errors.MultiError) (int, error) {
```
brikis98, line 281: This method needs to be moved to the same file where `MultiError` is defined.

errors/multierror.go
```diff
@@ -32,5 +32,10 @@ func (errs MultiError) Error() string {
 		}
 	}
 
-	return fmt.Sprintf("Hit multiple errors:\n%s", strings.Join(errorMessages, "\n"))
+	for _, errStatus := range errorMessages {
+        if errStatus == "exit status 2" {
+			return errStatus
+		}
+	}
+	return fmt.Sprintf("Encountered the following errors:\nHit multiple errors:\n%s", strings.Join(errorMessages, "\n"))
```
brikis98, line 40:
```suggestion
	return fmt.Sprintf("Encountered the following errors:\n%s", strings.Join(errorMessages, "\n"))
```

errors/multierror.go
```diff
@@ -32,5 +32,10 @@ func (errs MultiError) Error() string {
 		}
 	}
 
-	return fmt.Sprintf("Hit multiple errors:\n%s", strings.Join(errorMessages, "\n"))
+	for _, errStatus := range errorMessages {
+        if errStatus == "exit status 2" {
+			return errStatus
+		}
```
brikis98, line 38: Not sure I follow the logic here. Why in `MultiError` would we only return this error message if it was present? And does this actually change how Terragrunt handles this exit code?

alexkayabula: Let me do a more thorough investigation on how this exit code is handled and see how best to fix this issue.

jorhett:
is there a reason this was never pursued?  Lots of issues point at this problem #1470 #1166 etc

errors/multierror.go
```diff
@@ -32,5 +32,10 @@ func (errs MultiError) Error() string {
...
+	}
```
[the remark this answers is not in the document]
jorhett: I don't think this comment is relevant. The current code swallows all errors by design. This change is preventing error code 2 from being swallowed. It's a specific action for a specific situation that is unlikely to change. 

This analysis is wrong.

jorhett:
This remains a problem, it would be ideal to see it fixed.

yhakbar:
This PR has gotten quite stale, so I'm going to close it out.

Sorry we never got around to resolving the PR, @alexkayabula .

@jorhett , are you able to create or find a bug report and supply a reproducible example of the issue you're seeing? There is logic in Terragrunt today to support `-detailed-exitcode` for `run --all` plans.

PR #1235, opened 2020-06-25
Add ability to set terragrunt-working-dir with env var TERRAGRUNT_WORKING_DIR

yorinasub17:
Bonus features:

- Update reference docs to include env vars at the top for easier readability
- `TERRAGRUNT_CHECK` was parsed incorrectly (it would enable the flag when it was set to `false` as opposed to `true`).


yorinasub17:
Thanks for review! Will merge this in now.

lukasmrtvy:
would be nice to add also function to get a path from which fs path is terragrunt launched

PR #1239, opened 2020-06-28
[skip ci] Fix outdated docs from 0.18 days

yorinasub17:
This fixes an outdated example in the `find_in_parent_folders` docs that worked for terragrunt <=0.18, but no longer works for terragrunt 0.19+.


yorinasub17:
Merging straightaway given content, but feedback is welcome!

Issue #749, opened 2019-06-16
Hooks init-from-module runs differently in 0.18 and 0.19


antonbabenko:
Trying to convert my scripts to 0.19 and I have a problem with understanding how hooks work.

In 0.18 it worked like this. I have a hook which copied a file which contains `backend "s3" {}` before running `terraform init -backend-config=...` in Terragrunt working directory (`.terragrunt-cache/foo/bar/`):

```terraform
after_hook "copy_common_main_providers" {
  commands = ["init-from-module"]
  execute  = ["cp", "${get_parent_tfvars_dir()}/../common/main_providers.tf", "."]
}
```

I run `terragrunt init` and `main_providers.tf` was copied just into terragrunt-cache working directory.

In 0.19 the similar code (shown below) copies file `main_providers.tf` in the wrong directory (current directory) and into the working directory:

```terraform
after_hook "copy_common_main_providers" {
  commands = ["init-from-module"]
  execute  = ["cp", "${get_parent_terragrunt_dir()}/../../common/main_providers.tf", "."]
}
```

To remove unnecessary `main_providers.tf` from the wrong directory I added another hook (aka "hack"):
```terraform
after_hook "remove_useless_copy_of_main_providers" {
  commands = ["init"]
  execute  = ["rm", "-f", "${get_parent_terragrunt_dir()}/${path_relative_to_include()}/main_providers.tf"]
}
``` 

Any ideas how to instruct Terragrunt to copy files just once into a working directory?

brikis98:
> In 0.19 the similar code (shown below) copies file main_providers.tf in the wrong directory (current directory) and into the working directory:

Not sure I follow. What directory is it copying to? Which directory do you expect it to copy to?

Also, the path in the updated code is different (`../../` instead of `../`). Is that intentional?

antonbabenko:
Sorry for the delay.

The difference with `../../` and `../` is intentional, because it does not work otherwise in 0.19 while worked in 0.18.

This is the tree structure:
```
.
├── terragrunt.hcl
└── vpc
    ├── .terragrunt-cache
    │   └── 8tw6nJm13wWcoe79ee6cpyG1tvk
    │       └── BoKn_3NNiOJEg4PmxaP5wYFi3M8
    └── terragrunt.hcl
```

Parent `terragrunt.hcl` contains the hook:
```
after_hook "copy_common_main_providers" {
  commands = ["init-from-module"]
[...]

PR #1261, opened 2020-07-24
Clarify working directory of hooks

yorinasub17:
This PR updates the docs to clarify what the working directory of hooks are.

After some experimentation and code review, I found that the working directory of the hooks actually depended on which hook was running.

It turns out that all hooks run after the `init-from-module` stage is complete will be run in the terraform directory, while anything before that is run in the terragrunt config directory. Note that because the `terragruntOptions` struct is updated AFTER `init-from-module` is done, but is passed in to the hook function BEFORE `init-from-module` is run, `after_hooks` for `init-from-module` still run in the terragrunt config directory.

I verified this using the following config:

```
terraform {
  source = "git::git@github.com:gruntwork-io/terragrunt.git//test/fixture-get-output/regression-1124/modules/app?ref=master"

  after_hook "which_dir_init_from_module" {
    commands = ["init-from-module"]
    execute  = ["pwd"]
  }

  after_hook "which_dir_init" {
    commands = ["init"]
    execute  = ["pwd"]
  }

  before_hook "which_dir_plan" {
    commands = ["plan"]
    execute  = ["pwd"]
  }
}

inputs = {
  foo = "foo"
}
```

### DIFF
--- a/docs/_docs/04_reference/config-blocks-and-attributes.md
+++ b/docs/_docs/04_reference/config-blocks-and-attributes.md
@@ -51,25 +51,45 @@ The `terraform` block supports the following arguments:
       `terraform` like `required_var_files`, only any files that do not exist are ignored.
 
 - `before_hook` (block): Nested blocks used to specify command hooks that should be run before `terraform` is called.
-  Hooks run from the terragrunt configuration directory (the directory where `terragrunt.hcl` lives).
+  Hooks run from the directory with the terraform module, except for hooks related to `terragrunt-read-config` and
+  `init-from-module`. These hooks run in the terragrunt configuration directory (the directory where `terragrunt.hcl`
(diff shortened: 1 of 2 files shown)


lorengordon:
@yorinasub17 What do you think about adding a test for this behavior? `init-from-module` was something I contributed a while back, in the terragrunt <0.19 days. My intention was that `init-from-module` would run in the terraform directory, and it did at the time. But that got lost in the [terragrunt 0.19 update](https://github.com/gruntwork-io/terragrunt/issues/749), because there was no test for this behavior...

yorinasub17:
Thanks for review. Going to merge this in now.

@lorengordon Yes I agree we should have tests, but I didn't quite have time to shave that yak so deferred it for now.

PR #1262, opened 2020-07-24
[WIP] [RDD] Introduce authenticate block

yorinasub17:
__NOTE: This is just the README updates, following RDD__

Oftentimes you will want to run `terragrunt apply-all` or `plan-all` in a directory containing multiple modules that may reach different accounts (e.g. Landing Zone). Or you may have a `dependency` block that references resources in another account. In both of these scenarios, you need terragrunt to invoke modules in different accounts.

Currently, the only way to do this is to login to an IAM user that can assume IAM roles for each account and set the `iam_role` attribute in `terragrunt.hcl`. With this feature request, I am hoping to introduce another alternative using `aws-vault`.

Here I propose an `authenticate` block that allows us to specify complex configurations for authenticating to the cloud. The main feature I want to highlight here is the use of `aws-vault`.

The idea is that when `aws_vault_profile` is set, terragrunt will fetch MFA validated authentication credentials from `aws-vault` for the given profile (as if you had run `aws-vault exec profile -- terragrunt`), and feed those credentials where needed (`remote_state` and `terraform`).

Note that experimentation is needed to figure out how to make the interactive prompts of `aws-vault` work. In particular, I am worried that the `stdin` will be jumbled when you do a `terragrunt apply-all` and multiple modules request the MFA token.

### DIFF
--- a/docs/_docs/04_reference/config-blocks-and-attributes.md
+++ b/docs/_docs/04_reference/config-blocks-and-attributes.md
@@ -26,6 +26,7 @@ The following is a reference of all the supported blocks and attributes in the c
 - [dependency](#dependency)
 - [dependencies](#dependencies)
 - [generate](#generate)
+- [authenticate](#authenticate)
 
 ### terraform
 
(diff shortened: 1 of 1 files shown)


yorinasub17:
@brikis98 This is in draft status, but I would like a review of the README updates for RDD. I am thinking that this feature might help with an alternative solution to our LZ bootstrapping problem.

yorinasub17:
> What do you think about extending this to be a more general mechanism

The main reason I had it hard coded was so that this can be used with `remote_state`. Otherwise, I think it will be tricky to have it parse and update the env vars of the process on the fly (to auth to the AWS SDK within terragrunt binary).

brikis98:
> > What do you think about extending this to be a more general mechanism
> 
> The main reason I had it hard coded was so that this can be used with `remote_state`. Otherwise, I think it will be tricky to have it parse and update the env vars of the process on the fly (to auth to the AWS SDK within terragrunt binary).

Could Terragrunt just re-execute itself in a totally separate process with the given `command` as a prefix?

yorinasub17:
> Could Terragrunt just re-execute itself in a totally separate process with the given command as a prefix?

That unfortunately will break all the caching optimizations in the `dependency` block when running `xxx-all` commands.

brikis98:
> > Could Terragrunt just re-execute itself in a totally separate process with the given command as a prefix?
> 
> That unfortunately will break all the caching optimizations in the `dependency` block when running `xxx-all` commands.

I suppose if we wanted to, we could provide a way to pass that data between processes, but it might be messy. I suppose another option is for Terragrunt to execute `command`, read the resulting env vars, and set all of those in its own environment?

Also, how does `authenticate` play with `dependency` blocks? If a module you depend on has its own `authenticate` block, I suppose you use that. But if it doesn't, do you ever inherit the authentication from another module? Or is each module totally standalone? And do `authenticate` blocks get pulled in via `include`?

lorengordon:
Many of these cli utils that federate auth for AWS now support an option that outputs json, intended for use with the `credential_process` option of AWS Shared Config profiles. Including [aws-vault](https://github.com/99designs/aws-vault/blob/master/USAGE.md#using-credential-helper). Supporting a json schema might be easier than scraping and parsing the output for envs?

yorinasub17:
> Or is each module totally standalone? And do authenticate blocks get pulled in via include?

I was envisioning that the `authenticate` blocks are all standalone within the context of each module config. So:

- Yes it will be propagated with `include` or `import`.
- When going into `dependency`, the `authenticate` block will be reset. So if there is no `authenticate` block, it uses the same creds as if there was no `authenticate` block, and if there is an `authenticate` block, it will use that.

> Many of these cli utils that federate auth for AWS now support an option that outputs json

I had this thought too, or some way to indicate which env vars should be propagated. What I am not sure about is if that same model works for GCP or Azure. Some investigation is needed for the other clouds.

yorinasub17:
Closing as stale

PR #1263, opened 2020-07-24
Implement minimal debug feature

yorinasub17:
This takes https://github.com/gruntwork-io/terragrunt/pull/1137 and reduces the scope to just outputting a workable tfvars file. This should be the minimal increment for a usable debug command that should significantly improve status quo.


cli/debug.go
```diff
@@ -0,0 +1,108 @@
...
+				varName,
+			)
+		}
+	}
+	jsonContent, err := json.MarshalIndent(jsonValuesByKey, "", "  ")
+	if err != nil {
+		return nil, errors.WithStackTrace(err)
+	}
+	return jsonContent, nil
+}
```
yorinasub17, line 92: This actually produces a usable `tfvars.json` file. This makes me wonder if we should just ALWAYS produce this, and use this to feed the inputs instead of `TF_VAR_` env vars.

Note that I workaround the "vars must all be defined" terraform issue by introspecting the variables from the module source.

The advantage of this is that it would solve all the weird type issues you run into with the env vars (e.g., I think this would allow us to set `null` values for the inputs!).

@brikis98 what do you think?

cli/debug.go
```diff
@@ -0,0 +1,108 @@
...
+	terragruntOptions *options.TerragruntOptions,
+	terragruntConfig *config.TerragruntConfig,
+	moduleVariables []string,
+) ([]byte, error) {
+	var envVars map[string]string
+	if terragruntOptions.Env == nil {
+		envVars = map[string]string{}
+	} else {
+		envVars = terragruntOptions.Env
+	}
```
brikis98, line 60: NIT: slightly shorter.

```suggestion
	envVars := map[string]string{}
	if terragruntOptions.Env != nil {
		envVars = terragruntOptions.Env
	}
```

cli/debug.go
```diff
@@ -0,0 +1,108 @@
...
+		return nil, errors.WithStackTrace(err)
+	}
+	return jsonContent, nil
+}
+
+// terraformModuleVariables will return all the variables defined in the downloaded terraform modules, taking into
+// account all the generated sources.
+func terraformModuleVariables(terragruntOptions *options.TerragruntOptions) ([]string, error) {
+	modulePath := terragruntOptions.WorkingDir
+	module, diags := tfconfig.LoadModule(modulePath)
```
brikis98, line 98: Oh, this library makes this super easy, nice!

cli/debug.go
brikis98: So you're saying that instead of Terragrunt setting env vars to pass `inputs` to a module, we always generate a `tfvars.json` and write it to somewhere in `.terragrunt-cache`? I think I'm in favor of that as it makes the mental model simpler / more visible and helps with type issues, as you mentioned. 

A few things to think through:

1. What should the `.json` file we generate be called? We want to avoid conflicts with files users may already have in their modules. We could do this by either giving it a unique name (e.g., GUID) or by writing it to separate folder from where the user's code is. 
1. How do we tell Terraform to use the file? One option is to name it `terraform.tfvars.json` so Terraform picks it up automatically, but that could clash with the user's own files. Another option is `<GUID>.auto.tfvars.json`. That will avoid file conflicts and be picked up automatically. A third option is to write the file to some other path and explicitly tell Terraform to use it via `-var-file`.
1. I think generating this file would improve interop with TFC/TFE. If we write a `<GUID>.auto.tfvars.json` file, it'll be uploaded to TFC/TFE along with the code during executions.
1. Would we even need a debug mode anymore? I suspect if Terragrunt always logged (a) the path to this file, (b) the exact command its running, and (c) the folder its using as a working dir, then it would be much clearer to users what's going on and no longer require a separate debug command to see it.
1. Longer term thinking: I feel like Terragrunt is becoming a generator for Terraform source code... Here, we're generating a `.tfvars.json` file; with `generate` blocks we also generate `.tf` code; perhaps the [preprocessor idea](https://github.com/gruntwork-io/terragrunt/issues/759#issuecomment-585124357) is also just code generation?

lorengordon: There's an issue asking for exactly this, I think... https://github.com/gruntwork-io/terragrunt/issues/752

yorinasub17: Ok I went ahead and implemented this in a separate PR, so that we can decide which is better: to use this as debug mode, or use this as normal operation. https://github.com/gruntwork-io/terragrunt/pull/1267

yorinasub17:
Closing in favor of https://github.com/gruntwork-io/terragrunt/pull/1267

yorinasub17:
Reopened given https://github.com/gruntwork-io/terragrunt/pull/1267#issuecomment-667532462

cli/debug.go
yorinasub17: Done: acf43b5

yorinasub17:
@brikis98 any objections to merging this version of debug mode in?

yorinasub17:
Thanks for review! Merging and releasing now.

PR #1267, opened 2020-07-26
Update terragrunt to generate tfvars.json file instead of using TF_VAR env var

### LINKED ISSUE #752: Don't convert inputs to env vars. Instead, generate terragrunt.tfvars.json and use as -var-file

yorinasub17:
Fixes #752 

Extends https://github.com/gruntwork-io/terragrunt/pull/1263 to always generate tfvars json file instead of relying on a debug mode.

NOTE: I decided to open a new PR on a new branch so that we have the other branch in case we decide the debug mode is better.


cli/args.go
```diff
@@ -346,41 +350,93 @@ func parseMultiStringArg(args []string, argName string, defaultValue []string) (
...
-func toTerraformEnvVars(vars map[string]interface{}) (map[string]string, error) {
-	out := map[string]string{}
+// writeTFVarsFile will create a tfvars file that can be used to invoke the terraform module with the inputs generated
+// in terragrunt.
+func writeTFVarsFile(terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig) error {
+	terragruntOptions.Logger.Printf("Generating tfvars file")
 
-	for varName, varValue := range vars {
-		envVarName := fmt.Sprintf("TF_VAR_%s", varName)
+	variables, err := terraformModuleVariables(terragruntOptions)
```
brikis98, line 358: Sanity check: is it possible that we're somehow concealing a bug by only generating the `.tfvars.json` file with the input vars we find? 

I guess one possibility is that in `terragrunt.hcl`, you have a typo: e.g., you set `instance_typ = "t3.micro"` instead of `instance_type = "t3.micro"`. Terragrunt wouldn't find an `instance_typ` variable in your module's inputs and therefore wouldn't include it in the `.tfvars.json` file. I guess this would be a bit confusing during debugging. That said, today, with env vars, you have the exact same problem, but it's less visible.

If we were to output a `.tfvars` file instead of `tfvars.json`, we could include a commented out section of variables that were in your inputs, but for which there were no matching input vars. That would make things a lot clearer...

cli/args.go
```diff
@@ -346,41 +350,93 @@ func parseMultiStringArg(args []string, argName string, defaultValue []string) (
...
-			return nil, err
-		}
+	fileContents, err := terragruntTFVarsFileContents(terragruntOptions, terragruntConfig, variables)
+	if err != nil {
+		return err
+	}
 
-		out[envVarName] = string(envVarValue)
+	fileName := filepath.Join(terragruntOptions.WorkingDir, TerragruntTFVarsFileName)
+	if err := ioutil.WriteFile(fileName, fileContents, os.FileMode(int(0600))); err != nil {
```
brikis98, line 371: Do we always override the previous file? Is it ever worth checking if the user has a file with such a name there already?

cli/args.go
```diff
@@ -346,41 +350,93 @@ func parseMultiStringArg(args []string, argName string, defaultValue []string) (
...
+		return errors.WithStackTrace(err)
 	}
 
-	return out, nil
+	terragruntOptions.Logger.Printf("Variables passed to terraform are located in \"%s\"", fileName)
+	terragruntOptions.Logger.Printf("Run this command to replicate how terraform was invoked:")
+	terragruntOptions.Logger.Printf("\tterraform apply \"%s\"", terragruntOptions.WorkingDir)
```
brikis98, line 377: It's not always going to be `terraform apply`, right? E.g., It could be `plan`, `destroy`, etc.

Moreover, due to `extra_arguments`, we pass other args along too... Perhaps we should log the full, exact command we're executing more clearly for users?

cli/args.go
```diff
@@ -346,41 +350,93 @@ func parseMultiStringArg(args []string, argName string, defaultValue []string) (
...
+				terragruntOptions.Logger,
+				"WARN: The variable %s was omitted from the debug file because the env var %s is already set.",
+				varName, nameAsEnvVar,
+			)
+		} else if !varIsDefined {
+			util.Debugf(
+				terragruntOptions.Logger,
+				"WARN: The variable %s was omitted because it is not defined in the terraform module.",
+				varName,
+			)
```
brikis98, line 416: If we generated a `.tfvars` instead of `.tfvars.json`, we could include a commented-out section that has this exact info.

docs/_docs/02_features/inputs.md
```diff
@@ -24,18 +24,30 @@ inputs = {
 }
 ```
 
-Whenever you run a Terragrunt command, Terragrunt will set any inputs you pass in as environment variables. For example, with the `terragrunt.hcl` file above, running `terragrunt apply` is roughly equivalent to:
+Whenever you run a Terragrunt command, Terragrunt will convert these inputs into [a tfvars json
```
brikis98, line 27:
```suggestion
Whenever you run a Terragrunt command, Terragrunt will convert these inputs into a [`.tfvars.json`
```

cli/args.go
yorinasub17: This is valid, but as you mentioned, we already have this problem, and unlike with env vars, you would spot that the `instance_type` field is missing from the json when you load it. I think the benefit is marginal (e.g., it is just as easy to gloss over comments).

I agree that we should generate `tfvars`, but in the interest of having something ship sooner, I think we should punt on this to the next iteration. Generating `hcl` with comments in a robust manner will be complex since AFAIK, you can't just output the cty; you need to manipulate the AST to inject comment tokens.

cli/args.go
yorinasub17: See above re:my thoughts on `tfvars` hcl.

cli/args.go
yorinasub17: Good point. Since we already log the terraform command with the args as part of `RunCommandWithOutput`, and since I don't have access to the full args passed in at this stage, I decided to remove these logs and update the runcommand args to include the working dir.

6814a3a

cli/args.go
yorinasub17: I think we should always override it to avoid the element of surprise for now. I added a warning log when terragrunt detects the file: 6814a3a

When we convert to hcl, I think we should switch to the same behavior as `overwrite_terragrunt` mode of `generate` blocks, which is:

- Inject a signature as a comment to the generated file.
- If file does not exist, generate it.
- If file exists and has signature, overwrite it.
- If file exists but no signature, error out.

cli/args.go
```diff
@@ -346,41 +350,100 @@ func parseMultiStringArg(args []string, argName string, defaultValue []string) (
...
 
-// Convert the given variables to a map of environment variables that will expose those variables to Terraform. The
-// keys will be of the format TF_VAR_xxx and the values will be converted to JSON, which Terraform knows how to read
-// natively.
-func toTerraformEnvVars(vars map[string]interface{}) (map[string]string, error) {
-	out := map[string]string{}
+// writeTFVarsFile will create a tfvars file that can be used to invoke the terraform module with the inputs generated
+// in terragrunt.
+func writeTFVarsFile(terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig) error {
+	terragruntOptions.Logger.Printf("Generating tfvars file")
```
brikis98, line 356: NIT: specify what file is being generated (`TerragruntTFVarsFileName`).

cli/args.go
brikis98: Roger. Could file issue to track generating a `.tfvars` in future.

cli/args.go
```diff
@@ -346,41 +350,100 @@ func parseMultiStringArg(args []string, argName string, defaultValue []string) (
...
-		}
+	fileContents, err := terragruntTFVarsFileContents(terragruntOptions, terragruntConfig, variables)
+	if err != nil {
+		return err
+	}
+
+	fileName := filepath.Join(terragruntOptions.WorkingDir, TerragruntTFVarsFileName)
+
+	// If the file already exists, log a warning indicating that we will overwrite it.
+	if util.FileExists(fileName) {
```
brikis98, line 373: Will this warning appear if you re-run a Terragrunt command and there's a previously generated file in there?

cli/args.go
brikis98: Roger

brikis98 (dismissed):
Great work Yori 👍

cli/args.go
yorinasub17: Unfortunately yes, as currently there is no way to distinguish between a file that is user created and a file that is terragrunt generated (since we don't have comments to inject a signature).

cli/args.go
yorinasub17: Done: f599a09

cli/args.go
brikis98: To reduce noise, should we only log it as a debug?

yorinasub17: Good point: done in c7c4b71

brikis98:
> > Do we delete the file after?
> 
> No, so that it can be debugged. However, we should consider if that is the right approach given the next bullet:
> 
> > What if the values in the `inputs` block are secrets? Writing them to disk may be a bad idea.
> 
> **** I don't have a solution for this. Perhaps this is what forces us to delete this file after each run?
> 

Hm, yea, this is a tough one. Some options to think through:

1. Could we use an in-memory-only file? This would be by top choice... I'm pretty sure we could make that work on *nix, but I don't know Windows well enough to see if there's a way to do it there. Example approaches:

    - [stdin](https://stackoverflow.com/a/39230472/483528)
    - [named pipe](https://unix.stackexchange.com/a/63933/215969)
    - [process substitution](https://stackoverflow.com/a/39230530/483528)
    - [tmpfs](https://unix.stackexchange.com/a/188540/215969)

1. We could delete the right after each run. Of course, this makes debugging harder, so we'd need some `--debug` flag to force the file to stick around. We'd have to be extra careful to ensure that Terragrunt deletes the file even if there's an error or panic. Not much we could do about a forced `kill` (e.g., via `CTRL+C`).

1. Neither of the above works well enough, so we go back to env vars, and only write the file when a `--debug` flag is passed.

yorinasub17:
I decided to go with option (2). Now the tfvars file is deleted with a `defer` call, which can be skipped with `--terragrunt-debug`.

> We could delete the file right after each run. Of course, this makes debugging harder, so we'd need some --debug flag to force the file to stick around. We'd have to be extra careful to ensure that Terragrunt deletes the file even if there's an error or panic. Not much we could do about a forced kill (e.g., via CTRL+C).

Looking at [the docs for panic](https://blog.golang.org/defer-panic-and-recover), it looks like all defer calls are executed normally (emphasis mine):

> anic is a built-in function that stops the ordinary flow of control and begins panicking. When the function F calls panic, execution of F stops, **any deferred functions in F are executed normally**, and then F returns to its caller. To the caller, F then behaves like a call to panic....

So I think as long as the defer is done before the call to `writeTFVarsFile` (which I am doing), we are guaranteed to run this cleanup routine, which should mean that we will properly clean up the file in the face of error and panic.

brikis98:
Hm, I still a bit nervous about potentially writing secrets to disk:

1. `terragrunt apply` could run for an hour... That's a long time to have plaintext secrets sitting on disk.
1. For a variety of reasons, the Terragrunt process could be killed before the file is cleaned up. We should prob have tests that validate we clean up the file correctly, including in the face of a panic, but `CTRL+C`, `kill`, and a dozen other things would leave behind files on disk even if our code works perfectly.

Is it worth spending 30 min looking for a way on Windows to pass a file purely in memory? I know it can be done on *unix...

yorinasub17:
> We should prob have tests that validate we clean up the file correctly, including in the face of a panic

I don't think this kind of testing is going to be very easy without introducing a potential avenue for bugs in terragrunt (I am nervous about injecting something in the pipeline to induce a panic intentionally)...

---

> Is it worth spending 30 min looking for a way on Windows to pass a file purely in memory? I know it can be done on *nix...

Ok will look into it, but it'll be a while before I'll have time available to set up my dev env on my windows boxes again.

---

Let's then just revert back to the original debug mode version (option 3) for now so that we can ship something useful to satisfy this need of terragrunt debugging, and we don't block that on my ability to dedicate free nights and weekends to work on this.

yorinasub17:
UPDATE: rebased on https://github.com/gruntwork-io/terragrunt/pull/1263 so that this PR will automatically become an extension of that branch when that is merged, so that we don't throw away the work here.

brikis98:
> > We should prob have tests that validate we clean up the file correctly, including in the face of a panic
> 
> I don't think this kind of testing is going to be very easy without introducing a potential avenue for bugs in terragrunt (I am nervous about injecting something in the pipeline to induce a panic intentionally)...
> 
> > Is it worth spending 30 min looking for a way on Windows to pass a file purely in memory? I know it can be done on *nix...
> 
> Ok will look into it, but it'll be a while before I'll have time available to set up my dev env on my windows boxes again.
> 
> Let's then just revert back to the original debug mode version (option 3) for now so that we can ship something useful to satisfy this need of terragrunt debugging, and we don't block that on my ability to dedicate free nights and weekends to work on this.

Roger

abeluck:
Just wanted to drop a comment here thanking you for considering security in this case.

If terragrunt started writing secrets to disk we'd have to immediately stop using it.  We've stringent security policies in place that disallow secrets being written to plaintext on disk (even though the workstations/servers are full disk encrypted).... even for a few seconds.

It is common for workstation backup software to run in a continuous mode that constantly backs up changes to disk as they happen. So writing secrets to disk could easily result in them being swept into a backup archive somewhere.

That said, I'm definitely interested in the original motivation of this issue.

lorengordon:
To manage normal values vs secret values, could we maybe use two different inputs blocks, or a function? One input block would always write a tfvars file, the other would always export ENVs. Or check if that value is wrapped in the function, and exclude it from the tfvars and instead export it as an env...

yorinasub17:
Closing as stale

PR #1276, opened 2020-07-31
Only parse terraform source during stack generation

### LINKED ISSUE #992: From 0.19.24 to 0.21.10, can't use dependency in before_hook execute ?
Hi there

I've been using this successfully in 0.19.24 

      before_hook "open_tunnel_through_bastion" {
        commands = ["plan", "apply", "show", "destroy"]
        execute  = ["screen", "-d", "-m", "ssh", "-L", "12345:${dependency.instance.outputs.this_db_instance_address}:${dependency.instance.outputs.this_db_instance_port}", dependency.bastion.outputs.hostname, "-p", "22000", "sleep", "60"]
      }

Got a laptop issue, reinstalled latest Terragrunt 0.21.10, I ended up with 

> [terragrunt] 2020/01/09 17:27:59 /home/remy/work/terraform-live/development/aws/eu-west-1/postgres/databases/swarm-accessd/terragrunt.hcl:11,16-17: Unsuitable value type; Unsuitable value: element 6: string required

Does that ring a bell ? 

(edited to fix code block)

yorinasub17:
Fixes #992

This updates the partial parsing step during module stack generation to only parse the terraform source and ignore the other blocks/attributes. This allows usage of `dependency` references in the sub blocks of the `terraform` block, but not the `source`.

### DIFF
--- a/config/config_partial.go
+++ b/config/config_partial.go
@@ -21,6 +21,7 @@ const (
 	DependenciesBlock PartialDecodeSectionType = iota
 	DependencyBlock
 	TerraformBlock
+	TerraformSource
 	TerragruntFlags
 	TerragruntVersionConstraints
 )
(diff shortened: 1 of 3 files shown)


yorinasub17:

gercograndia:
guys, sorry to bring this back to life again but I was trying to use a dependency into the source part, and it was working with a regular plan|apply, but not with a `run-all`.

So ended up here, noticed this has been fixed for everything in the `terraform` block, except for `source`.

I assume there is (or was) a good reason that this was not possible for the `source`, is this still valid?

Or is there an opportunity to allow for `source` now as well?

PR #1296, opened 2020-08-12
Add terragrunt-parallelism to documentation

yorinasub17:
It looks like this option is not included in our reference docs, so I added it in.

**NOTE: will be merging this in directly given content.**


brikis98:
LGTM, thx for adding!

PR #1310, opened 2020-08-23
Migrate RFC and upgrade guides

yorinasub17:
This PR removes the confusing `_docs` folder at the top level and rolls in the migration guides and RFC docs to the terragrunt website.

Note that I opted to remove the Terragrunt 0.12.x migration guide in the process given that it is too old.


brikis98 (approved):
That's a great idea 👍

brikis98:
> Screen shot of new nav
> 
> Imports

Looks great.

Super minor NIT: any way to get "Rfc" to show as "RFC?" Don't spend more than 3 min on it 😁

Issue #1103, opened 2020-03-26
Add option to only check top level dependency blocks


dmattia:
## Current behavior

If I have a module named MainModule with two dependency blocks, and those two dependencies each depend on a few other modules, and I run a `terragrunt plan` on MainModule, terragrunt will run a `terraform output -json` to check the state of all modules described in this scenario, not just on the two top level dependencies.

### It is slower

In my real world terragrunt config, we have hundreds of modules, some of which have over a dozen dependencies, each of which have their own. Even with provider caching, I still see cases where a `terragrunt plan` takes over ten minutes because it checks all of the state in series without parallelization.

### It goes against my intuition of the permissions model

This one is highly opinionated, so I only mean it as a suggestion. But say I have the following modules:

- DevOrg module, which creates an AWS organization for dev resources
- ProdOrg module, which creates an AWS organization for prod resources
- CommonEcr module, which creates a Docker repo on AWS Ecr with read access from DevOrg and ProdOrg
- EcsDevModule, which depends on that ECR module to get the url of the docker image to use

Running a plan on EcsDevModule would have a deep level dependency on being able to access the Prod env state, which I do no like. Instead, I would prefer if I only had to give access to CommonsOrg state.

yorinasub17:
That's a fair point. This happens because getting the output calls `RunTerragrunt` [here](https://github.com/gruntwork-io/terragrunt/blob/6c304f91690badae3c2d64bf5b2c98d38368621e/config/dependency.go#L354), and that function is currently parsing the target config in full.

The implementation to fix this is a bit involved though because it means we need to update the `RunTerragrunt` routine to do "just-in-time" config parsing. That is, it needs to index all the terraform commands and construct a mapping so that it only parses the relevant pieces of the config for each command. E.g., `output` may only need the `terraform` block, while `apply` needs all the `inputs`, unless it has a plan file (you can see how this can get complicated).

Note that we can't just parse out the `terraform` block for calling `output` and be done with it. The crux of the issue is parsing out the `dependency` blocks from the target config, and avoiding that requires an update to the configuration parsing mechanism. Right now, we parse `dependency` blocks before `terraform` blocks (see https://terragrunt.gruntwork.io/docs/getting-started/configuration/#configuration-parsing-order) because we want to support the usage of `dependency` in setting `hooks` and `extra_arguments` and the like. Updating that means we need some kind graph based parsing mechanism so that we only parse `dependency` blocks if it is used: a very involved refactor!

[...]

PR #1311, opened 2020-08-23
Dependency optimization

yorinasub17:
This is an attempt at partially resolving https://github.com/gruntwork-io/terragrunt/issues/1103 

I thought if there was a way we can optimize the remote state fetching to avoid recursively parsing all the dependencies, and realized that if the config follows a few conventions, this is possible.

- The remote state is managed using `remote_state` blocks.
- The `remote_state` block is in generate mode (the `generate` attribute is set).
- The `remote_state` block itself does not depend on any `dependency` outputs (`locals` and `include` are ok).

When these conditions are met, we are able to generate a terraform module that only contains the remote state backend configuration, which we can use to run `terraform output` to get the recorded outputs from the state without retrieving dependencies. You don't even need the full module source to do this!

Note that supporting arbitrary terragrunt configuration (e.g. those managing backends with `generate`) is difficult as there won't be a stable API we could define to account for all the cases. For these configurations, we will have to fall back to the old method until we are ready to relax these constraints further.

### DIFF
--- a/config/dependency.go
+++ b/config/dependency.go
@@ -395,16 +395,23 @@ func cloneTerragruntOptionsForDependencyOutput(terragruntOptions *options.Terrag
 // by directly pulling down the state file. Otherwise, terragrunt will fallback to running `terragrunt output` on the
 // target module.
 func getTerragruntOutputJson(terragruntOptions *options.TerragruntOptions, targetConfig string) ([]byte, error) {
+	// Make a copy of the terragruntOptions so that we can reuse the same execution environment, but in the context of
+	// the target config.
+	targetTGOptions, err := cloneTerragruntOptionsForDependencyOutput(terragruntOptions, targetConfig)
+	if err != nil {
+		return nil, err
+	}
+
 	// First, attempt to parse the `remote_state` blocks without parsing/getting dependency outputs. If this is
 	// possible, proceed to routine that fetches remote state directly. Otherwise, fallback to calling
 	// `terragrunt output` directly.
-	remoteStateTGConfig, err := PartialParseConfigFile(targetConfig, terragruntOptions, nil, []PartialDecodeSectionType{RemoteStateBlock})
+	remoteStateTGConfig, err := PartialParseConfigFile(targetConfig, targetTGOptions, nil, []PartialDecodeSectionType{RemoteStateBlock})
 	if err != nil || !canGetRemoteState(remoteStateTGConfig.RemoteState) {
 		terragruntOptions.Logger.Printf("WARNING: Could not parse remote_state block from target config %s", targetConfig)
 		terragruntOptions.Logger.Printf("WARNING: Falling back to terragrunt output.")
-		return runTerragruntOutputJson(terragruntOptions, targetConfig)
+		return runTerragruntOutputJson(targetTGOptions, targetConfig)
 	}
-	return getTerragruntOutputJsonFromRemoteState(terragruntOptions, targetConfig, remoteStateTGConfig.RemoteState)
+	return getTerragruntOutputJsonFromRemoteState(targetTGOptions, targetConfig, remoteStateTGConfig.RemoteState)
 }
 
 // canGetRemoteState returns true if the remote state block is not nil and has the generate block configured.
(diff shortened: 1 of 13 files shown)


yorinasub17:
>> Note that supporting arbitrary terragrunt configuration (e.g. those managing backends with generate) is difficult as there won't be a stable API we could define to account for all the cases.
> I'm probably missing something obvious. Can you state it differently?

What I mean is that in order to support a generic solution that works for any terragrunt configuration, we need to implement some complex parsing mechanism that supports just-in-time lookups to ensure that we only get the information that is absolutely necessary.

This is because we don't have a stable API for managing remote state blocks in terragrunt: there are multiple ways one can do this between:

- Using `remote_state` block.
- Using `generate` block to generate the backend config.
- Using `extra_arguments` to feed the specific backend config to `init` calls.
- Using `dependency` references in `generate` blocks to output backend config blocks for buckets managed by terraform.

We won't be able to replicate the optimizations here for each of these use cases, until we are ready to start enforcing stricter constraints on the config similar to what we stated here.

E.g., the next one I see being possible is if we introduce a way to flag `generate` blocks that are used to generate backend config (with a special attribute like `is_state_block = true`).

yorinasub17:
Thanks for review! I'm going to merge this in now and release so we can start benefitting!

config/dependency.go
```diff
@@ -382,28 +386,100 @@ func cloneTerragruntOptionsForDependencyOutput(terragruntOptions *options.Terrag
...
 }
 
-// runTerragruntOutputJson uses terragrunt running functions to extract the json output from the target config. Make a
-// copy of the terragruntOptions so that we can reuse the same execution environment.
-func runTerragruntOutputJson(terragruntOptions *options.TerragruntOptions, targetConfig string) ([]byte, error) {
-	targetOptions, err := cloneTerragruntOptionsForDependencyOutput(terragruntOptions, targetConfig)
+// Retrieve the outputs from the terraform state in the target configuration. This attempts to optimize the output
+// retrieval if the following conditions are true:
+// - State backends are managed with a `remote_state` block.
+// - The `remote_state` block is using the generator pattern (the `generate` attribute is set).
```
brikis98, line 392: If the user isn't setting the `generate` attribute, but we still have all the remote state config... Could we generate a `backend` block for them anyway, just for this workaround?

config/dependency.go
```diff
@@ -382,28 +386,100 @@ func cloneTerragruntOptionsForDependencyOutput(terragruntOptions *options.Terrag
 	return targetOptions, nil
 }
 
-// runTerragruntOutputJson uses terragrunt running functions to extract the json output from the target config. Make a
-// copy of the terragruntOptions so that we can reuse the same execution environment.
-func runTerragruntOutputJson(terragruntOptions *options.TerragruntOptions, targetConfig string) ([]byte, error) {
-	targetOptions, err := cloneTerragruntOptionsForDependencyOutput(terragruntOptions, targetConfig)
+// Retrieve the outputs from the terraform state in the target configuration. This attempts to optimize the output
+// retrieval if the following conditions are true:
+// - State backends are managed with a `remote_state` block.
```
brikis98, line 391: For backends not managed this way, what would we need to do? I guess we'd need to be able to (a) check out the source code, which is either in the current working dir, or checked out based on the `source` param, and (b) run any `generate` blocks. As long as the `source` param and `generate` blocks don't depend on other `dependency` blocks, could we add a special case here too, that checks out the code, runs the code generation, and then runs `output`?

config/dependency.go
```diff
@@ -382,28 +386,100 @@ func cloneTerragruntOptionsForDependencyOutput(terragruntOptions *options.Terrag
...
+	// Create working directory where we will run terraform in. Make sure it is cleaned up before the function returns.
+	tempWorkDir, err := ioutil.TempDir("", "")
+	if err != nil {
+		return nil, err
+	}
+	defer os.RemoveAll(tempWorkDir)
+
+	// Set the terraform working dir to the tempdir, and set stdout writer to ioutil.Discard so that output content is
+	// not logged.
+	targetTGOptions.WorkingDir = tempWorkDir
```
brikis98, line 443: NIT: if this function modifies `TerragruntOptions`, this is probably the function that should clone it.

config/dependency.go
```diff
@@ -382,28 +386,100 @@ func cloneTerragruntOptionsForDependencyOutput(terragruntOptions *options.Terrag
...
+	targetTGOptions.Writer = ioutil.Discard
```
brikis98, line 444: Will this result in errors being hidden?

brikis98 (commented):
This is super clever Yori! Nicely done 👍

config/dependency.go
yorinasub17: Nope because the `ErrWriter` is shared and all errors are writen to `stderr` so you will see the error messages if this command fails.

config/dependency.go
yorinasub17: I did it this way because I needed the cloned options in both the parent caller and here and thought it wasn't very memory efficient to be cloning multiple times, but I suppose you are right about how it should reclone here to avoid subtle bugs.

config/dependency.go
yorinasub17: oooo that is a good point!

yorinasub17: Addressed in https://github.com/gruntwork-io/terragrunt/pull/1317

config/dependency.go
yorinasub17: Addressed in https://github.com/gruntwork-io/terragrunt/pull/1317

Issue #1314, opened 2020-08-26
0.23.35 init fails where 0.23.34/0.23.31 succeed


thoellrich:
I just upgraded to .35 this morning on ubuntu 18.04.4 LTS and attempted to `apply` a component of our infrastructure with the upgraded terragrunt. It fails during `init` where the previous version had no problems.

Please note that I'm using `TERRAGRUNT_IAM_ROLE` in the environment.

```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.4 LTS
Release:        18.04
Codename:       bionic
$ terraform version
Terraform v0.12.28

Your version of Terraform is out of date! The latest version
is 0.13.0. You can update by downloading from https://www.terraform.io/downloads.html
$ env | grep TERRA
TERRAGRUNT_IAM_ROLE=arn:aws:iam::xxx:role/yyy
$ rm -rf .terragrunt-cache && terragrunt0.23.35 init
[terragrunt] [/home/ubuntu/dev/terraform/xxx/us-east-2/uat/wave] 2020/08/26 14:57:31 Running command: terraform --version
[terragrunt] 2020/08/26 14:57:31 Terraform version: 0.12.28
[terragrunt] 2020/08/26 14:57:31 Reading Terragrunt config file at /home/ubuntu/dev/terraform/xxx/us-east-2/uat/wave/terragrunt.hcl
[terragrunt] [/home/ubuntu/dev/terraform/xxx/us-east-2/uat/route53] 2020/08/26 14:57:31 Generated file /tmp/690994747/auto_generated_backend.tf.
[terragrunt] [/home/ubuntu/dev/terraform/xxx/us-east-2/uat/route53] 2020/08/26 14:57:31 Running command: terraform init
[terragrunt] [/home/ubuntu/dev/terraform/xxx/us-east-2/uat/dynamodb] 2020/08/26 14:57:31 Generated file /tmp/017886757/auto_generated_backend.tf.
[terragrunt] [/home/ubuntu/dev/terraform/xxx/us-east-2/uat/dynamodb] 2020/08/26 14:57:31 Running command: terraform init
[terragrunt] [/home/ubuntu/dev/terraform/xxx/us-east-2/uat/global] 2020/08/26 14:57:31 Generated file /tmp/463338304/auto_generated_backend.tf.
[...]

PR #1315, opened 2020-08-26
Account for iam role configurations when fetching remote state in dependency optimization

yorinasub17:
Fixes https://github.com/gruntwork-io/terragrunt/issues/1314

This fixes a bug that was introduced in the recent dependency retrieval optimization, where it was not accounting for IAM role assume configurations.

NOTE: I didn't add tests because it doesn't look like we have tests for this functionality, but we should figure out a way to have a test. In the meantime, I am going to cut an alpha testing release so that the reporter can test this.

### DIFF
--- a/aws_helper/config.go
+++ b/aws_helper/config.go
@@ -165,3 +165,24 @@ func GetAWSUserID(terragruntOptions *options.TerragruntOptions) (string, error) 
 
 	return *identity.UserId, nil
 }
+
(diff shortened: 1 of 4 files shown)


brikis98 (approved):

yorinasub17:
Have confirmation that this indeed fixes the issue, so will go ahead to merge. Thanks for the review!

PR #1317, opened 2020-08-26
Follow up to dependency optimization

yorinasub17:
- Allow `remote_state` blocks without `generate` set by dynamically configuring `generate` just for the purpose of output fetching.

- Reclone terragrunt options in inner function to avoid global patching issues.

### DIFF
--- a/config/dependency.go
+++ b/config/dependency.go
@@ -415,9 +415,9 @@ func getTerragruntOutputJson(terragruntOptions *options.TerragruntOptions, targe
 	return getTerragruntOutputJsonFromRemoteState(targetTGOptions, targetConfig, remoteStateTGConfig.RemoteState, remoteStateTGConfig.IamRole)
 }
 
-// canGetRemoteState returns true if the remote state block is not nil
+// canGetRemoteState returns true if the remote state block is not nil and dependency optimization is not disabled
 func canGetRemoteState(remoteState *remote.RemoteState) bool {
-	return remoteState != nil
+	return remoteState != nil && !remoteState.DisableDependencyOptimization
 }
 
 // getTerragruntOutputJsonFromRemoteState will retrieve the outputs directly by using just the remote state block. This
--- a/config/dependency.go
+++ b/config/dependency.go
@@ -20,6 +20,7 @@ import (
 	"golang.org/x/sync/errgroup"
 
 	"github.com/gruntwork-io/terragrunt/aws_helper"
+	"github.com/gruntwork-io/terragrunt/codegen"
 	"github.com/gruntwork-io/terragrunt/errors"
 	"github.com/gruntwork-io/terragrunt/options"
 	"github.com/gruntwork-io/terragrunt/remote"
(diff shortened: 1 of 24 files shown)


yorinasub17:
There is a legitimate test failure in `TestDependencyOutputCachePathBug`. I'll need to dig in to figure out what caused this, but don't quite have the time at the moment. Will revisit later.

yorinasub17:
UPDATE: 86c5604

This switches to disabling the dependency optimization by default. https://github.com/gruntwork-io/terragrunt/issues/1318 worries me that this may silently break a bunch of workflows in the wild that depend on hooks and `extra_arguments` for `init`. Since it is a bit involved to refactor the code to allow setting these up in the dependency runtime, I decided that the lesser evil is to disable by default and make dependency optimization an opt-in feature. That way, at least it won't break existing workflows when naively upgrading to the latest version.

This does mean that users will need to set the flag to enable the optimization, but I hope that is relatively easy to do.

config/dependency.go
```diff
@@ -408,22 +408,16 @@ func getTerragruntOutputJson(terragruntOptions *options.TerragruntOptions, targe
...
-		targetTGOptions.IamRole = remoteStateTGConfig.IamRole
-	}
-	return getTerragruntOutputJsonFromRemoteState(targetTGOptions, targetConfig, remoteStateTGConfig.RemoteState)
+	terragruntOptions.Logger.Printf("Dependency optimization is enabled and available.")
+	terragruntOptions.Logger.Printf("Fetching dependency by pulling down remote state directly.")
+	return getTerragruntOutputJsonFromRemoteState(targetTGOptions, targetConfig, remoteStateTGConfig.RemoteState, remoteStateTGConfig.IamRole)
 }
 
-// canGetRemoteState returns true if the remote state block is not nil and has the generate block configured.
+// canGetRemoteState returns true if the remote state block is not nil and user did not request to disable optimization
```
yorinasub17, line 418:
```suggestion
// canGetRemoteState returns true if the remote state block is not nil and user requested to enable optimization
```

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -184,6 +184,9 @@ The `remote_state` block supports the following arguments:
   have support in Terragrunt to be automatically created if the storage does not exist. Currently `s3` and `gcs` are the
   two backends with support for automatic creation. Defaults to `false`.
 
+- `enable_dependency_optimization` (attribute): When `true`, attempt to optimize dependency fetching for terragrunt
```
brikis98, line 187: A bit hard to see this in the PR... Is this flag on the `remote_state` block or a global flag? If it's the former, do you need to put it on the `remote_state` block of the current module or *every module that is a dependency* of the current module?

brikis98: Talked with Yori live. Since most users likely have `remote_state` defined once, centrally, and `include`-ed everywhere else, this is effectively global.

brikis98:
Yori and I talked live and we're guessing that the type of issues mentioned in https://github.com/gruntwork-io/terragrunt/issues/1318 are probably fairly uncommon. They only affect:

1. Users of `dependency` blocks...
1. Who are also using `remote_state` blocks...
1. Who are also using `extra_arguments` or `hooks`...
1. And those `extra_arguments` or `hooks` affect the behavior of `init` in such a way that the `output` command behaves differently if `extra_arguments` or `hooks` were not executed.

I suspect this is a small minority of users. Given that everyone else benefits from this optimization, we're going to default to having it on, but offer a flag to shut it off in the rare cases it causes issues.

brikis98 (approved):

yorinasub17:
Thanks for review! Build passed so going to merge and release this now.

Issue #1313, opened 2020-08-25
terraform init : backend re initialization is failing  


irfanjs:
Hi,
i am using terraform + terragrunt to deploy the infra. 

here is the directory structure i have . 

> resource-group/
> service-bus/
> terragrunt.hcl

i am using remote backend to store the state files . here is the configuration.

```
remote_state {
    backend = "azurerm"
    config = {
        key = "${path_relative_to_include()}/terraform.tfstate"
        resource_group_name = "ABC-QA-RG"
        storage_account_name = "strgacct"
        container_name = "tfstate"
        subscription_id = "sadsadsads45435c59"
    }
}

inputs = {
  location = "East US"
  subs_id  = "fsdfdsfdsfdsf-sdfdsfdsf-6-4250b797bc59"
  tenant_id = "sdfdsfdsf67657657-546546"
  
}

```

the issue is , i have just moved the development environment to some other machine and copy/paste the above folder structure as it is . 
now, in the parent folder , when i run the command `terragrunt plan-all ` it fails , saying that , you need to initialize the terraform. 
as per doc : https://www.terraform.io/docs/backends/config.html , it says, just run the command` terraform init `and everything will be initialized. 

As per my understanding , i expect terraform / terragrunt to get the state of infra from the remote backend and start the execution without error . is that correct  ? 

please suggest / help what needs to be done to resolve the error .

irfanjs:
please suggest . this is blocker .

yorinasub17:
`terragrunt` should run `init` automatically on a completely clean clone. Did you by chance copy over the `.terragrunt-cache` as well, and if so, does it work if you remove that dir?

ahlinc:
I found that the **v0.23.34** version works fine and doesn't run `terraform init` every time on every dependency when I call `terragrunt apply-all`. All newer versions starting from **v0.23.35** seems are broken.

yorinasub17:
> doesn't run terraform init every time on every dependency

This is part of the dependency optimization feature and it is `init`-ing in a temporary folder so that terragrunt can pull down the remote state information as opposed to running through the standard path of parsing the entire config (including recursive dependencies) and redownloading sources, etc.

If this is not desirable, you can disable this functionality starting with `v0.23.36` by setting `disable_dependency_optimization = true` on the `remote_state` block. You can read more about this in the updated docs on the [dependency block](https://terragrunt.gruntwork.io/docs/reference/config-blocks-and-attributes/#dependency).

[...]

PR #1321, opened 2020-08-28
Run terraform init with flags to avoid downloading plugins

yorinasub17:
Addresses https://github.com/gruntwork-io/terragrunt/issues/1313#issuecomment-682226842

It turns out that `terraform init` will still download all the providers even though there is no source code because the info is stored in the state file. However, it is not necessary to initialize with the plugins in order to run `terraform output`: the only thing necessary is the backend configuration.

As such, this PR introduces a further optimization that avoids the provider downloads using CLI flags passed to the `init` call. Note that this causes the command to fail:

```
Initializing the backend...

Successfully configured the backend "s3"! Terraform will automatically
use this backend unless the backend configuration changes.

Missing required providers.

The following provider constraints are not met by the currently-installed
provider plugins:

* aws (any version)
* null (any version)
* template (any version)

Terraform can automatically download and install plugins to meet the given
constraints, but this step was skipped due to the use of -get-plugins=false
and/or -plugin-dir on the command line.

If automatic installation is not possible or desirable in your environment,
you may manually install plugins by downloading a suitable distribution package
and placing the plugin's executable file in one of the directories given in
by -plugin-dir on the command line, or in the following directory if custom
plugin directories are not set:
    terraform.d/plugins/darwin_amd64


Error: missing provider "null"



Error: missing provider "aws"



Error: missing provider "template"
```

At this time, there is no way to just setup the backend using `init`, so in this PR, I opted for ignoring the `init` error. To avoid confusing the logs with this expected errors, the stderr output is also now suppressed for the init call. Note that if there was any issue with the backend configuration, the output call will fail and you will see the error there. I also added the option to get the stderr logs to show up by setting `TG_LOG=debug`.

### DIFF
--- a/config/dependency.go
+++ b/config/dependency.go
@@ -476,10 +476,11 @@ func getTerragruntOutputJsonFromRemoteState(
 	}
 
 	// The working directory is now set up to interact with the state, so pull it down to get the json output.
-	// First run init, then output.
-	if err := shell.RunTerraformCommand(targetTGOptions, "init"); err != nil {
-		return nil, err
-	}
+
+	// First run init to setup the backend configuration so that we can run output.
+	runTerraformInitForDependencyOutput(targetTGOptions, tempWorkDir, targetConfig)
+
+	// Now that the backend is initialized, run terraform output to get the data and return it.
 	out, err := shell.RunTerraformCommandWithOutput(targetTGOptions, "output", "-json")
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


brikis98 (approved):
The fix looks good, but note that `-get-plugins=false` may be broken in TF 0.13: https://github.com/hashicorp/terraform/issues/25813

yorinasub17:
Thanks for review. That is good to know, but it looks like they will try to restore it somehow: https://github.com/hashicorp/terraform/issues/25813#issuecomment-675728123

I think this won't affect functionality because we swallow `init` errors, but it will affect runtimes, in which case you can always opt out. I also think https://github.com/gruntwork-io/terragrunt/issues/1322 is the proper fix here (if one wants optimum speed).

---

Will merge and release this now!

ahmad-hamade:
Great job. well done 🥇
Running plan-all for more than 100 modules has increased the process a lot!!

PR #1326, opened 2020-09-01
Hacky attempt at AWS Provider patching to help users work around Terraform bug

### LINKED ISSUE #1306: Add reference docs for terragrunt specific commands
In the reference section of terragrunt.gruntwork.io, we should add a section that lists all the terragrunt commands:

- `apply-all`
- `plan-all` (including limitations: see below)
- `destroy-all`
- etc

This should also list the limitations of `plan-all`, as described in https://github.com/gruntwork-io/terragrunt/issues/720#issuecomment-497888756 and https://github.com/gruntwork-io/terragrunt/issues/1127#issuecomment-611353504

brikis98:
This PR adds an `aws-provider-patch` command that can be used to override attributes in nested `provider` blocks. This is an attempt at a hacky workaround for [a Terraform bug](https://github.com/hashicorp/terraform/issues/13018) where `import` does not work if you are using any modules that have `provider` blocks nested in within them. With this PR, users could run:

```bash
terragrunt aws-provider-patch --terragrunt-override-attr region=eu-west-1
```

And Terragrunt will:

1. Run `terraform init` to download the code for all your modules into `.terraform/modules`.
1. Scan all the Terraform code in `.terraform/modules`, find AWS `provider` blocks, and hard-code the `region` param
   to `eu-west-1` for each one. 

Once you do this, you'll hopefully be able to run `import` on that module. After that, you can delete the modified `.terraform/modules` and go back to normal.

Note that I wanted to document this command somewhere, so I updated the CLI docs to include all Terragrunt commands, and not just options. Therefore, as a side effect, this PR also fixes #1306.

### DIFF
--- a/cli/aws_provider_patch.go
+++ b/cli/aws_provider_patch.go
@@ -131,6 +131,10 @@ func findAllTerraformFilesInModules(terragruntOptions *options.TerragruntOptions
 // any dynamic values in nested provider blocks are not handled correctly when you call 'terraform import', so by
 // temporarily hard-coding them, we can allow 'import' to work.
 func patchAwsProviderInTerraformCode(terraformCode string, terraformFilePath string, attributesToOverride map[string]string) (string, bool, error) {
+	if len(attributesToOverride) == 0 {
+		return terraformCode, false, nil
+	}
+
 	hclFile, err := hclwrite.ParseConfig([]byte(terraformCode), terraformFilePath, hcl.InitialPos)
 	if err != nil {
 		return "", false, errors.WithStackTrace(err)
@@ -140,25 +144,27 @@ func patchAwsProviderInTerraformCode(terraformCode string, terraformFilePath str
 
 	for _, block := range hclFile.Body().Blocks() {
 		tokens := block.BuildTokens(nil)
-		if len(tokens) > 4 {
-			// The tokens should be:
-			//   - provider
-			//   - "
-			//   - aws
-			//   - "
-			maybeProvider := tokens[0]
-			maybeAws := tokens[2]
 
-			if maybeProvider.Type == hclsyntax.TokenIdent && string(maybeProvider.Bytes) == "provider" &&
-				maybeAws.Type == hclsyntax.TokenQuotedLit && string(maybeAws.Bytes) == "aws" {
+		for index, token := range tokens {
+			if token.Type == hclsyntax.TokenIdent && string(token.Bytes) == "provider" {
+				// We're looking for 4 tokens in a row that look like this:
+				//   - provider
+				//   - "
+				//   - aws
+				//   - "
+				if len(tokens) - index > 2 {
+					maybeAws := tokens[index + 2]
 
-				codeWasUpdated = len(attributesToOverride) > 0
-				for key, value := range attributesToOverride {
-					block.Body().SetAttributeValue(key, cty.StringVal(value))
+					if maybeAws.Type == hclsyntax.TokenQuotedLit && string(maybeAws.Bytes) == "aws" {
+						for key, value := range attributesToOverride {
+							block.Body().SetAttributeValue(key, cty.StringVal(value))
+						}
+
+						codeWasUpdated = true
+					}
 				}
 			}
 		}
-
 	}
 
 	return string(hclFile.Bytes()), codeWasUpdated, nil
--- a/cli/aws_provider_patch.go
+++ b/cli/aws_provider_patch.go
@@ -152,8 +152,8 @@ func patchAwsProviderInTerraformCode(terraformCode string, terraformFilePath str
 				//   - "
 				//   - aws
 				//   - "
-				if len(tokens) - index > 2 {
-					maybeAws := tokens[index + 2]
+				if len(tokens)-index > 2 {
+					maybeAws := tokens[index+2]
 
 					if maybeAws.Type == hclsyntax.TokenQuotedLit && string(maybeAws.Bytes) == "aws" {
 						for key, value := range attributesToOverride {
--- a/cli/aws_provider_patch.go
+++ b/cli/aws_provider_patch.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"encoding/json"
 	"fmt"
 	"github.com/gruntwork-io/terragrunt/errors"
 	"github.com/gruntwork-io/terragrunt/options"
@@ -10,6 +11,8 @@ import (
 	"github.com/hashicorp/hcl2/hclwrite"
 	"github.com/mattn/go-zglob"
 	"github.com/zclconf/go-cty/cty"
+	"io/ioutil"
+	"path/filepath"
 )
 
 func applyAwsProviderPatch(terragruntOptions *options.TerragruntOptions) error {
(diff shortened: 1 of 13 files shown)


brikis98:
I got the patching code working, but there's a bit of an issue: Terraform downloads modules into `.terraform/modules`, but it downloads the _entire_ repo into that folder, and not just a single module... So the patching code either:

1. Updates every file in the entire repo, which is wasteful, and potentially results in parsing files that perhaps shouldn't be parsed.
1. We have to parse `.terraform/modules/modules.json` to know exactly in which folders to look... But I'm not sure if that's a stable API we can rely on.

cli/aws_provider_patch.go
```diff
@@ -0,0 +1,131 @@
...
+
+	return nil
+}
+
+// findAllTerraformFiles returns all Terraform source files within the modules being used by this Terragrunt
+// configuration. To be more specific, it only returns the source files downloaded for module "xxx" { ... } blocks into
+// the .terraform/modules folder; it does NOT return Terraform files for the top-level (AKA "root") module.
+//
+// NOTE: this method only supports *.tf files right now. Terraform code defined in *.json files is not currently
+// supported.d
```
yorinasub17, line 67:
```suggestion
// supported.
```

cli/aws_provider_patch.go
```diff
@@ -0,0 +1,131 @@
...
+func findAllTerraformFilesInModules(terragruntOptions *options.TerragruntOptions) ([]string, error) {
+	modulesPath := util.JoinPath(terragruntOptions.DataDir(), "modules")
+
+	if !util.FileExists(modulesPath) {
+		return nil, nil
+	}
+
+	// Ideally, we'd use a builin Go library like filepath.Glob here, but per https://github.com/golang/go/issues/11862,
```
yorinasub17:
```suggestion
	// Ideally, we'd use a builtin Go library like filepath.Glob here, but per https://github.com/golang/go/issues/11862,
```

cli/aws_provider_patch.go
```diff
@@ -0,0 +1,131 @@
...
+	hclFile, err := hclwrite.ParseConfig([]byte(terraformCode), terraformFilePath, hcl.InitialPos)
+	if err != nil {
+		return "", false, errors.WithStackTrace(err)
+	}
+
+	codeWasUpdated := false
+
+	for _, block := range hclFile.Body().Blocks() {
+		tokens := block.BuildTokens(nil)
+		if len(tokens) > 4 {
```
yorinasub17: If you switch to the newer HCL2 library (`"github.com/hashicorp/hcl/v2"`), you should have access to `block.Type()` and `block.Labels`, which should simplify this to:

```
if block.Type() == "provider" && len(block.Labels()) == 1 && block.Labels()[0] == "aws" {
```

cli/aws_provider_patch.go
```diff
@@ -0,0 +1,179 @@
...
+// the .terraform/modules folder; it does NOT return Terraform files for the top-level (AKA "root") module.
+//
+// NOTE: this method only supports *.tf files right now. Terraform code defined in *.json files is not currently
+// supported.d
+func findAllTerraformFilesInModules(terragruntOptions *options.TerragruntOptions) ([]string, error) {
+	// Terraform downloads modules into the .terraform/modules folder. Unfortunately, it downloads not only the module
+	// into that folder, but the entire repo it's in, which can contain lots of other unrelated code we probably don't
+	// want to touch. To find the paths to the actual modules, we read the modules.json file in that folder, which is
+	// a manifest file Terraform uses to track where the modules are within each repo. Note that this is an internal
+	// API, so the way we parse/read this modules.json file may break in future Terraform versions.
```
yorinasub17, line 73: Add link to https://github.com/hashicorp/terraform/blob/master/internal/modsdir/manifest.go as a reference (Note that we can't import that directly because it is labeled as internal).

cli/aws_provider_patch.go
brikis98: Whhhhhaaaat? Will try now, thanks! 😁

brikis98: Well goddamn. It worked. Nice one Yori! https://github.com/gruntwork-io/terragrunt/pull/1326/commits/bdd6eed5b941e31a365774ca0bf40170f4c92f90

cli/aws_provider_patch.go
brikis98: Good idea: https://github.com/gruntwork-io/terragrunt/pull/1326/commits/8fcc5060861ec81ccf754f81fb9c79248629fa7f

cli/aws_provider_patch.go
```diff
@@ -0,0 +1,131 @@
+package cli
+
+import (
+	"fmt"
+	"github.com/gruntwork-io/terragrunt/errors"
+	"github.com/gruntwork-io/terragrunt/options"
+	"github.com/gruntwork-io/terragrunt/util"
+	"github.com/hashicorp/hcl2/hcl"
+	"github.com/hashicorp/hcl2/hcl/hclsyntax"
+	"github.com/hashicorp/hcl2/hclwrite"
```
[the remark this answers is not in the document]
brikis98: Fixed: https://github.com/gruntwork-io/terragrunt/pull/1326/commits/bdd6eed5b941e31a365774ca0bf40170f4c92f90

cli/aws_provider_patch.go
brikis98: Fixed: https://github.com/gruntwork-io/terragrunt/pull/1326/commits/0db564ace59f9ea140363172ee614d694e5ed824

cli/aws_provider_patch.go
brikis98: Fixed: https://github.com/gruntwork-io/terragrunt/pull/1326/commits/0db564ace59f9ea140363172ee614d694e5ed824

brikis98:
Update: I decided to bite the bullet and parse the `.terraform/modules/modules.json` format after all. Not sure of a better approach.

Note that I wanted to document this command somewhere, so I updated the CLI docs to include all Terragrunt commands, and not just options. Therefore, as a side effect, this PR also fixes #1306. I've updated the description of this PR as well, and I believe it's now ready for review.

test/integration_test.go
```diff
@@ -1214,6 +1215,18 @@ func TestAutoRetryApplyAllDependentModuleRetries(t *testing.T) {
...
 
+func TestAwsProviderPatch(t *testing.T) {
+	t.Parallel()
+
+	stderr := new(bytes.Buffer)
+	rootPath := copyEnvironment(t, TEST_FIXTURE_AWS_PROVIDER_PATCH)
+	modulePath := util.JoinPath(rootPath, TEST_FIXTURE_AWS_PROVIDER_PATCH)
+	err := runTerragruntCommand(t, fmt.Sprintf("terragrunt aws-provider-patch --terragrunt-override-attr region=eu-west-1 --terragrunt-working-dir %s", modulePath), os.Stdout, stderr)
+
+	assert.Nil(t, err)
```
yorinasub17, line 1226: NIT: Use `assert.NoError`. The difference is that when there is an error, this will render the error object so you can see the error message (as opposed to some weird pointer).

yorinasub17 (dismissed):
LGTM!

test/integration_test.go
brikis98: Good call! https://github.com/gruntwork-io/terragrunt/pull/1326/commits/0fc46e35e5eb595dfedf5ddbd8c1b944653b22f7

brikis98:
Can I haz one more review plz? 😄

brikis98:

PR #1327, opened 2020-09-01
Deprecate "lock_table" attribute from use in the remote S3 backend configuration

kwilczynski:
The "lock_table" attribute of the remote S3 backend configuration has been deprecated in Terraform before release 0.13 where it has been retired. The current name of the attribute is "dynamodb_table".

Terragrunt currently supports both of the attributes, but should the "lock_table" one be used, then it would be passed as-is either when generating a remote S3 backend block when templating Terraform or as a command-line argument. Using older attribute would not be a problem with Terraform before release 0.13, where the old attribute is still present albeit deprecated, but for anyone who uses the latest release, this would now become a problem as Terraform would terminate with an error.

This change deprecates the use of the "lock_table" in Terragrunt for remote S3 backend configuration and adds facilities to notify the user about the attribute being deprecated. Additionally, the use of "lock_table" will be normalised so that the "dynamodb_table" will be used instead in the resulting remote S3 backend block or as a command-line argument.

### DIFF
--- a/remote/remote_state_s3.go
+++ b/remote/remote_state_s3.go
@@ -19,6 +19,10 @@ import (
 	"github.com/mitchellh/mapstructure"
 )
 
+const (
+	lockTableDeprecationMessage = "Remote state configuration 'lock_table' attribute is deprecated; use 'dynamodb_table' instead."
+)
+
 /*
  * We use this construct to separate the two config keys 's3_bucket_tags' and 'dynamodb_table_tags'
  * from the others, as they are specific to the s3 backend, but only used by terragrunt to tag
@@ -61,7 +65,7 @@ type RemoteStateConfigS3 struct {
 	RoleArn          string `mapstructure:"role_arn"`
 	ExternalID       string `mapstructure:"external_id"`
 	SessionName      string `mapstructure:"session_name"`
-	LockTable        string `mapstructure:"lock_table"`
+	LockTable        string `mapstructure:"lock_table"` // Deprecated in Terraform version 0.13 or newer.
 	DynamoDBTable    string `mapstructure:"dynamodb_table"`
 	CredsFilename    string `mapstructure:"shared_credentials_file"`
 	S3ForcePathStyle bool   `mapstructure:"force_path_style"`
@@ -82,8 +86,9 @@ func (c *ExtendedRemoteStateConfigS3) GetAwsSessionConfig() *aws_helper.AwsSessi
 	}
 }
 
-// The DynamoDB lock table name used to be called lock_table, but has since been renamed to dynamodb_table, and the old
-// name deprecated. To maintain backwards compatibility, we support both names.
+// The DynamoDB lock table attribute used to be called "lock_table", but has since been renamed to "dynamodb_table", and
+// the old attribute name deprecated. The old attribute name has been eventually removed from Terraform starting with
+// release 0.13. To maintain a backwards compatibility, we support both names.
 func (s3Config *RemoteStateConfigS3) GetLockTableName() string {
 	if s3Config.DynamoDBTable != "" {
 		return s3Config.DynamoDBTable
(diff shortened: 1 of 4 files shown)


kwilczynski:
Hi @yorinasub17,

I was able to successfully test this with latest releases of Terraform 0.12 and 0.13, and in both cases it works fine when Terragrunt is being used with existing states or after an initial `terragrunt init` has been run (provided that the `terragrunt.hcl` had a proper remote S3 bucket configuration in place). In other words, for existing users and new users there will be no issue as long as the same release of Terraform is being used.

I also tested this change using either "lock_table" or "dynamodb_table" interchangeably with an existing state, and it does work as intended.

Where it does fall short is when user upgrades their Terraform from release 0.12 to 0.13 (where the "lock_table" has been already retired and will not be included in the state as a attribute with `NULL` value), as then running either vanilla Terraform or Terragrunt would result in Terraform complaining about an unsupported (old?) property being present in the state file, as per:

Screenshot 2020-09-01 at 17 16 58

This would be true following the upgrade regardless whether Terragrunt was used or not. This is quite unfortunate and not the best user experience - which is why I am wondering, whether there would be a way to facilitate this "migration" on behalf of the user? Terragrunt can read and parse the Terraform state file without issues (see [terraform_state_file.go](https://github.com/gruntwork-io/terragrunt/blob/master/remote/terraform_state_file.go)), but it leaves - and rightfully so, as this can become a maintenance burden very quickly - state management wholly to Terraform itself.

What do you think could be done to handle this? Should we _try_ to manage this in the existing state from Terragrunt?

kwilczynski:
I looked into few options like e.g., prompting the user about old attribute being present while a new release of Terraform has been detected (a version constraint check of `>= 0.13.0`), detecting that existing backend has old property present, etc., but perhaps keeping it simple stupid and allowing it to fail so that the user would run `terraform init -reconfigure` at their convenience - especially since it would upgrade their local and remote state making it incompatible with release **0.12** (as there are new fields added that older version of Terraform does not understand).

@yorinasub17, let me know if you think this is acceptable, or if you think we rather ought to do something better here. :^)

remote/remote_state_s3.go
```diff
@@ -82,8 +86,9 @@ func (c *ExtendedRemoteStateConfigS3) GetAwsSessionConfig() *aws_helper.AwsSessi
 	}
 }
 
-// The DynamoDB lock table name used to be called lock_table, but has since been renamed to dynamodb_table, and the old
-// name deprecated. To maintain backwards compatibility, we support both names.
+// The DynamoDB lock table attribute used to be called "lock_table", but has since been renamed to "dynamodb_table", and
+// the old attribute name deprecated. The old attribute name has been eventually removed from Terraform starting with
+// release 0.13. To maintain a backwards compatibility, we support both names.
```
yorinasub17, line 91:
```suggestion
// release 0.13. To maintain backwards compatibility, we support both names.
```

remote/remote_state_s3.go
```diff
@@ -186,6 +191,17 @@ func configValuesEqual(config map[string]interface{}, existingBackend *Terraform
 		}
 	}
 
+	// Nowadays it only makes sense to set the "dynamodb_table" attribute as it has
+	// been supported in Terraform since the release of version 0.10. The deprecated
+	// "lock_table" attribute it's either set to NULL in the state file or missing
```
yorinasub17, line 196:
```suggestion
	// "lock_table" attribute is either set to NULL in the state file or missing
```

yorinasub17 (commented):
Code LGTM! Apologies for the delay in reviewing this as I had been neck deep in other areas.

I'm going to kick off the build now, and if the test resolves, we can merge this in after addressing the two documentation nits.

> let me know if you think this is acceptable

I think what you have is reasonable. I agree that automatically resolving this would be a maintenance burden, especially if it was a one time thing for one version.

Deferring to the user to self resolve seems reasonable to do here.

yorinasub17 (approved):

yorinasub17:
Nevermind I was able to commit the suggestions myself so I can go ahead and merge this in. Thanks again for your contribution and very detailed report!

kwilczynski:
Hi @yorinasub17,

> Code LGTM! Apologies for the delay in reviewing this as I had been neck deep in other areas.
[...]

No problem! I can imagine you are very busy maintaining all the tools, etc. :) Thank you so much!

[...]
> > let me know if you think this is acceptable
> 
> I think what you have is reasonable. I agree that automatically resolving this would be a maintenance burden, especially if it was a one time thing for one version.
> 
> Deferring to the user to self resolve seems reasonable to do here.

Hopefully users upgrading to Terraform 0.13 or newer from their 0.12 state will have to do it only once.

Thank you again for help! You rock!

remote/remote_state_s3.go
kwilczynski: Doh! Apologies!

PR #1334, opened 2020-09-10
Make Terragrunt recover from missing module and provider error

kwilczynski:
This change enables Terragrunt to detect errors that otherwise required the user to run "terraform init" and attempt a recovery by running Terraform automatically.

When a Terraform module or provider version changes or when a module or provider is missing and Terragrunt is invoked with an action other than "init" (e.g., plan, apply, etc.), then the Terraform run that has been started by Terragrunt will terminate with an error asking the user to run "terraform init" to reinitialize to fetch the missing runtime dependencies.

Terragrunt should detect these type of missing module and provider errors and automatically re-run "terraform init" to resolve them and retry the previous command automatically.

Signed-off-by: Krzysztof Wilczyński <kw@linux.com>

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -384,7 +384,27 @@ func RunTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 		}
 	}
 
-	return runTerragruntWithConfig(terragruntOptions, terragruntConfig, false)
+	isInitCommand := util.FirstArg(terragruntOptions.TerraformCliArgs) == CMD_INIT
+
+	output, err := runTerragruntWithConfig(terragruntOptions, terragruntConfig, false)
+
+	// If running Terraform resulted in an error that matches one of the init errors,
+	// then run the required interim init, and then re-run previous command as-is,
+	// otherwise return error without making any attempt to recover from it or
+	// retry anything.
+	if err != nil && isErrorRequiringInit(output, err, terragruntOptions) && !isInitCommand {
+		terragruntOptions.Logger.Println("Detected an error that requires Terraform to be re-initialized; running 'terraform init' now...")
+
+		// Run the "terraform init" command.
+		if err := runTerraformInit(terragruntOptions, terragruntConfig, nil); err != nil {
+			return err
+		}
+
+		// If the init succeeded, re-run previous command again.
+		_, err = runTerragruntWithConfig(terragruntOptions, terragruntConfig, false)
+	}
+
+	return err
 }
 
 // Check the version constraints of both terragrunt and terraform. Note that as a side effect this will set the
@@ -509,7 +529,7 @@ func shouldRunHook(hook config.Hook, terragruntOptions *options.TerragruntOption
 
 // Runs terraform with the given options and CLI args.
 // This will forward all the args and extra_arguments directly to Terraform.
-func runTerragruntWithConfig(terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig, allowSourceDownload bool) error {
+func runTerragruntWithConfig(terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig, allowSourceDownload bool) (string, error) {
 
 	// Add extra_arguments to the command
 	if terragruntConfig.Terraform != nil && terragruntConfig.Terraform.ExtraArgs != nil && len(terragruntConfig.Terraform.ExtraArgs) > 0 {
@@ -520,48 +540,51 @@ func runTerragruntWithConfig(terragruntOptions *options.TerragruntOptions, terra
 	}
 
 	if err := setTerragruntInputsAsEnvVars(terragruntOptions, terragruntConfig); err != nil {
-		return err
+		return "", err
 	}
 
 	if util.FirstArg(terragruntOptions.TerraformCliArgs) == CMD_INIT {
 		if err := prepareInitCommand(terragruntOptions, terragruntConfig, allowSourceDownload); err != nil {
-			return err
+			return "", err
 		}
 	} else {
 		if err := prepareNonInitCommand(terragruntOptions, terragruntConfig); err != nil {
-			return err
+			return "", err
 		}
 	}
 
 	// Now that we've run 'init' and have all the source code locally, we can finally run the patch command
 	if shouldApplyAwsProviderPatch(terragruntOptions) {
-		return applyAwsProviderPatch(terragruntOptions)
+		return "", applyAwsProviderPatch(terragruntOptions)
 	}
 
 	if err := checkProtectedModule(terragruntOptions, terragruntConfig); err != nil {
-		return err
+		return "", err
 	}
 
-	return runActionWithHooks("terraform", terragruntOptions, terragruntConfig, func() error {
+	return runActionWithHooks("terraform", terragruntOptions, terragruntConfig, func() (string, error) {
 		return runTerraformWithRetry(terragruntOptions)
 	})
 }
 
 // Run the given action function surrounded by hooks. That is, run the before hooks first, then, if there were no
 // errors, run the action, and finally, run the after hooks. Return any errors hit from the hooks or action.
-func runActionWithHooks(description string, terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig, action func() error) error {
+func runActionWithHooks(description string, terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig, action func() (string, error)) (string, error) {
 	beforeHookErrors := processHooks(terragruntConfig.Terraform.GetBeforeHooks(), terragruntOptions)
 
-	var actionErrors error
+	var (
+		actionErrors error
+		actionStderr string
+	)
 	if beforeHookErrors == nil {
-		actionErrors = action()
+		actionStderr, actionErrors = action()
 	} else {
 		terragruntOptions.Logger.Printf("Errors encountered running before_hooks. Not running '%s'.", description)
 	}
 
 	postHookErrors := processHooks(terragruntConfig.Terraform.GetAfterHooks(), terragruntOptions, beforeHookErrors, actionErrors)
 
-	return errors.NewMultiError(beforeHookErrors, actionErrors, postHookErrors)
+	return actionStderr, errors.NewMultiError(beforeHookErrors, actionErrors, postHookErrors)
 }
 
 // The Terragrunt configuration can contain a set of inputs to pass to Terraform as environment variables. This method
@@ -585,22 +608,24 @@ func setTerragruntInputsAsEnvVars(terragruntOptions *options.TerragruntOptions, 
 	return nil
 }
 
-func runTerraformWithRetry(terragruntOptions *options.TerragruntOptions) error {
+func runTerraformWithRetry(terragruntOptions *options.TerragruntOptions) (string, error) {
 	// Retry the command configurable time with sleep in between
 	for i := 0; i < terragruntOptions.MaxRetryAttempts; i++ {
 		if out, tferr := shell.RunTerraformCommandWithOutput(terragruntOptions, terragruntOptions.TerraformCliArgs...); tferr != nil {
-			if out != nil && isRetryable(out.Stderr, tferr, terragruntOptions) {
-				terragruntOptions.Logger.Printf("Encountered an error eligible for retrying. Sleeping %v before retrying.\n", terragruntOptions.Sleep)
-				time.Sleep(terragruntOptions.Sleep)
-			} else {
-				return tferr
+			if out != nil {
+				if isRetryable(out.Stderr, tferr, terragruntOptions) {
+					terragruntOptions.Logger.Printf("Encountered an error eligible for retrying. Sleeping %v before retrying.\n", terragruntOptions.Sleep)
+					time.Sleep(terragruntOptions.Sleep)
+				} else {
+					return out.Stderr, tferr
+				}
 			}
-		} else {
-			return nil
+			return "", tferr
 		}
+		return "", nil
 	}
 
-	return errors.WithStackTrace(MaxRetriesExceeded{terragruntOptions})
+	return "", errors.WithStackTrace(MaxRetriesExceeded{terragruntOptions})
 }
 
 // Prepare for running 'terraform init' by initializing remote state storage and adding backend configuration arguments
@@ -742,7 +767,10 @@ func runTerraformInit(terragruntOptions *options.TerragruntOptions, terragruntCo
 		return err
 	}
 
-	return runTerragruntWithConfig(initOptions, terragruntConfig, terraformSource != nil)
+	if _, err = runTerragruntWithConfig(initOptions, terragruntConfig, terraformSource != nil); err != nil {
+		return err
+	}
+	return nil
 }
 
 func prepareInitOptions(terragruntOptions *options.TerragruntOptions, terraformSource *TerraformSource) (*options.TerragruntOptions, error) {
@@ -901,6 +929,14 @@ func isRetryable(tfoutput string, tferr error, terragruntOptions *options.Terrag
 		return false
 	}
 	return util.MatchesAny(terragruntOptions.RetryableErrors, tfoutput)
+}
+
+// isErrorRequiringInit checks whether there was an error and we should attempt to re-run init.
+func isErrorRequiringInit(tfoutput string, tferr error, terragruntOptions *options.TerragruntOptions) bool {
+	if !terragruntOptions.AutoRetry || !terragruntOptions.AutoInit || tferr == nil {
+		return false
+	}
+	return util.MatchesAny(terragruntOptions.ErrorsRequiringInit, tfoutput)
 }
 
 // Custom error types
--- a/cli/cli_app_test.go
+++ b/cli/cli_app_test.go
@@ -1,6 +1,8 @@
 package cli
 
 import (
+	"errors"
+	"fmt"
 	"testing"
 
 	"github.com/gruntwork-io/terragrunt/config"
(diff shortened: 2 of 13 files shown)


docs/_docs/02_features/auto-init.md
```diff
@@ -23,7 +23,7 @@ When Auto-Init is enabled (the default), terragrunt will automatically call [`te
 
 As [mentioned]({{site.baseurl}}/docs/features/keep-your-cli-flags-dry/#extra_arguments-for-init), `extra_arguments` can be configured to allow customization of the `terraform init` command.
 
-Note that there might be cases where terragrunt does not properly detect that `terraform init` needs be called. In this case, terraform would fail. Running `terragrunt init` again corrects this situation.
+Terragrunt will attempt to detect several common errors to automatically handle the cases where the `terraform init` would need to be called (for example a module version has changed and a new version has to be downloaded, etc.) and then run it as appropriate. In a case when Terragrunt detected that `terraform init` has to be run it will then re-run the original command (e.g., `plan`, `apply`, etc.) once `terraform init` completes successfully. Note that even that Terragrunt will try to recover from most errors there might be cases where Terragrunt does not properly detect that `terraform init` needs be called. In this case, Terraform would fail. Running `terragrunt init` again corrects this situation.
```
yorinasub17, line 26:
```suggestion
Terragrunt will attempt to detect several common errors to automatically handle the cases where `terraform init` would need to be called (for example a module version has changed and a new version has to be downloaded, etc.) and then run it as appropriate. In cases where Terragrunt detected that `terraform init` has to be run it will then re-run the original command (e.g., `plan`, `apply`, etc.) once `terraform init` completes successfully. Note that even though Terragrunt will try to recover from most errors there might be cases where Terragrunt does not properly detect that `terraform init` needs to be called. In this case, Terraform would fail. Running `terragrunt init` again corrects this situation.
```

brikis98, line 26: Super minor NIT:

```suggestion
Terragrunt will attempt to detect several common errors to automatically handle the cases where the `terraform init` would need to be called (for example a module version has changed and a new version has to be downloaded, etc.) and then run it as appropriate. In a case when Terragrunt detected that `terraform init` has to be run it will then re-run the original command (e.g., `plan`, `apply`, etc.) once `terraform init` completes successfully. Note that even though Terragrunt will try to recover from most errors there might be cases where Terragrunt does not properly detect that `terraform init` needs be called. In this case, Terraform would fail. Running `terragrunt init` again corrects this situation.
```

yorinasub17 (commented):
Docs change makes sense to me!

brikis98 (commented):
Doc changes LGTM!

brikis98:
hahah, lol, race condition with Yori and Jim detected.

docs/_docs/02_features/auto-init.md
kwilczynski: Thank you!

kwilczynski: Thank you!

cli/cli_app.go
```diff
@@ -542,26 +542,44 @@ func runTerragruntWithConfig(terragruntOptions *options.TerragruntOptions, terra
...
-	return runActionWithHooks("terraform", terragruntOptions, terragruntConfig, func() error {
+	err, output := runActionWithHooks("terraform", terragruntOptions, terragruntConfig, func() (error, string) {
 		return runTerraformWithRetry(terragruntOptions)
 	})
+	if err != nil && isErrorRequiringInit(output, err, terragruntOptions) {
+		initOptions := terragruntOptions.Clone(terragruntOptions.TerragruntConfigPath)
+		initOptions.TerraformCommand = "init"
+		initOptions.TerraformCliArgs = []string{"init"}
+		initOptions.AutoInit = false
+		if err := runTerragruntWithConfig(initOptions, terragruntConfig, allowSourceDownload); err != nil {
```
brikis98, line 553: Can this get into an infinite loop? Since `runTerragruntWithConfig` is calling itself, if `init` fails again and again due to some permanent issue, won't this keep running forever?

cli/cli_app.go
```diff
@@ -542,26 +542,44 @@ func runTerragruntWithConfig(terragruntOptions *options.TerragruntOptions, terra
...
+			return err
+		}
+
+		err, _ = runActionWithHooks("terraform", terragruntOptions, terragruntConfig, func() (error, string) {
+			return runTerraformWithRetry(terragruntOptions)
+		})
```
brikis98, line 559: I'm a bit worried about this copy/pasted block. If we change something in the main one, we may forget to change something in this retry. Any way to structure the code to avoid the duplication?

cli/cli_app.go
```diff
@@ -542,26 +542,44 @@ func runTerragruntWithConfig(terragruntOptions *options.TerragruntOptions, terra
...
 	beforeHookErrors := processHooks(terragruntConfig.Terraform.GetBeforeHooks(), terragruntOptions)
 
-	var actionErrors error
+	var (
+		actionErrors error
+		actionOutput string
+	)
 	if beforeHookErrors == nil {
-		actionErrors = action()
+		actionErrors, actionOutput = action()
```
brikis98, line 575: Does `actionOutput` contain `stdout`? Or `stderr`? Or both? Might want to make that clearer via the var name.

options/auto_retry_options.go
```diff
@@ -18,3 +18,11 @@ var RETRYABLE_ERRORS = []string{
 	"NoSuchBucket: The specified bucket does not exist",
 	"(?s).*Error creating SSM parameter: TooManyUpdates:.*",
 }
+
+// List of recurring transient errors encountered when calling terraform
+// If one of these are encountered, we will re-run the init command
+var ERRORS_REQUIRING_INIT = []string{
+	"(?s).*Module version requirements have changed.*",
+	"(?s).*Could not satisfy plugin requirements.*",
+	"(?s).*Error accessing remote module registry.*",
```
brikis98, line 27: What about:

```
 * provider.external: new or changed plugin executable
```

I used to get this all the time. Not sure if that error message changed?

And:

```
Please run "terraform init".
```

This is part of the "Could not satisfy plugin requirements" message you have above, but perhaps adding it too is a good way to catch other similar errors, where Terraform clearly knows what needs to be done, but for some reason chooses not to do it?

cli/cli_app.go
brikis98: Yea, thinking about this more, I think this needs to use the max retry and timeout logic...

brikis98 (commented):
Looks like good progress 👍 

A few integration tests will def be important to check this actually works!

cli/cli_app.go
kwilczynski: Good point. It contains standard error output. I will update the name! Thank you!

options/auto_retry_options.go
kwilczynski: There are more error messages to be added to both the `RETRYABLE_ERRORS ` and `ERRORS_REQUIRING_INIT `. I will take care about it. The one I added here are mainly those I was able to use when resting (reproducing) the problem locally. Update is coming shortly.

kwilczynski:
Hi @brikis98, thank you for the review thus far! Much appreciated.

There is another use-case I need to test this solution against - a more complex structure with nested sub-directories (when someone has complex infrastructure and their topology contains of a lot of different states being managed). Why? Because I need to verify how passing the `terragruntOptions.TerragruntConfigPath` to `terragruntOptions.Clone()` will influence execution in a more complex scenario.

cli/cli_app.go
kwilczynski: Let me see if I can make this better. Just to clarify - the "main one" is the first invocation of `runActionWithHooks` above this one that you are referring to?

cli/cli_app.go
kwilczynski: Done.

options/auto_retry_options.go
kwilczynski: I added a number of new error messages to match against - hopefully these would cover most of the common errors.

cli/cli_app.go
kwilczynski: This is a very good point. Although, in this case we have an exit condition here:

https://github.com/gruntwork-io/terragrunt/blob/94b962bb85b58bee0019c0718c0987bd92179cda/cli/cli_app.go#L526-L534

The first moment we got an error when running `terraform init`, then we would exit. This is because if the command is `init` and the `prepareInitCommand()` runs and returns an error, then this error would bubble up the stack and then we would exit.

Interestingly, I need to re-factor this, as we _might_ want to retry `terraform init` itself, which we don't at the moment.

kwilczynski: I am going to retire this idea of running `runTerragruntWithConfig()` recursively. It's unnecessarily complex and hard to test.

cli/cli_app.go
kwilczynski: This is going to change.

kwilczynski:
Rebased against master.

kwilczynski:
Hi @brikis98 and @yorinasub17,

Apologies for a bit of commotion here.

This is a different approach - a simpler one. Without having to change or refactor too much.

Rather than calling `runTerragruntWithConfig()` recursively, which is also hard to work with (as per @brikis98's comment) and test, simply detect the need to run `terraform init` within `RunTerragrunt()` then set everything up and run it.

Should `terraform init` succeeds, then re-run the previous command as-is without changing any options, should it fail, then the Terragrunt run fails too (since errors propagation is straightforward in this case).

This approach also enabled support for normal retry mechanism should there be any issues that the `RETRYABLE_ERRORS` list would normally cover - this works for any command that was passed and when the extra interim `terraform init` is required.

There is also an option of wrapping things into `util.DoWithRetry()`, so that a whole flow of _a command_ then _terraform init_ then _a command_ (again) could be retried on errors at any point. Albeit, when this is done, then it looks slightly awkward from the user perspective as messages would suggest there were some errors at some point, etc. I wasn't fun of this approach.

Tests to follow.

Let me know what do you think?

cli/cli_app.go
```diff
@@ -384,7 +384,36 @@ func RunTerragrunt(terragruntOptions *options.TerragruntOptions) error {
...
+		initOptions.Writer = initOptions.ErrWriter
+		initOptions.TerraformCommand = "init"
+		initOptions.TerraformCliArgs = []string{"init"}
+		initOptions.WorkingDir = terragruntOptions.WorkingDir
+		initOptions.AutoInit = false
+
+		// Run the "terraform init" command.
+		if err, _ = runTerragruntWithConfig(initOptions, terragruntConfig, false); err != nil {
+			return err
+		}
```
yorinasub17, line 406: Can we actually use `runTerraformInit` here? That should simplify this routine.

Alternatively, you can use `prepareInitOptions` to get the `terragruntOptions` that works with `init`.

cli/cli_app.go
```diff
@@ -384,7 +384,36 @@ func RunTerragrunt(terragruntOptions *options.TerragruntOptions) error {
...
 	}
 
-	return runTerragruntWithConfig(terragruntOptions, terragruntConfig, false)
+	err, output := runTerragruntWithConfig(terragruntOptions, terragruntConfig, false)
+
+	// If running Terraform resulted in an error that matches one of the init errors,
+	// then run the required interim init, and then re-run previous command as-is,
+	// otherwise return error without making any attempt to recover from it or
+	// retry anything.
+	if err != nil && isErrorRequiringInit(output, err, terragruntOptions) {
```
yorinasub17, line 393: I think this routine should only be run if the command is not init. Any errors that `init` should retry on should be captured in the auto-retry feature instead of here.

yorinasub17 (commented):
The new implementation looks better! Had a suggestion for further improvement as well.

I think this is ready for some integration tests.

cli/cli_app.go
kwilczynski: I can use `runTerraformInit()`. I gather it's OK to pass `nil` as the last argument (of type `TerraformSource`)?

cli/cli_app.go
kwilczynski: By the "auto-retry" feature, do you mean the `runTerraformWithRetry()`?

yorinasub17: Yup that's correct!

cli/cli_app.go
yorinasub17: Yup that should be ok.

cli/cli_app.go
kwilczynski: OK. In which case this will work as intended - all the `terraform init` -related error messages are going to be matched against and taken care of in the `runTerraformWithRetry()`.

cli/cli_app.go
kwilczynski: No worries. Would you mind if I clean `runTerraformInit()` a bit? It calls internally `prepareInitOptions()` which takes `TerraformSource` as second argument and then does not do anything with it, and it also returns an error which is always `nil`. As per:

https://github.com/gruntwork-io/terragrunt/blob/94b962bb85b58bee0019c0718c0987bd92179cda/cli/cli_app.go#L748-L759

brikis98 (commented):
The new approach does look cleaner. A few integration tests will def be necessary to check this works.

cli/cli_app.go
```diff
@@ -384,7 +384,31 @@ func RunTerragrunt(terragruntOptions *options.TerragruntOptions) error {
...
+
+		// If the init succeeded, re-run previous command again.
+		if err, _ = runTerragruntWithConfig(terragruntOptions, terragruntConfig, false); err != nil {
+			return err
+		}
+	} else if err != nil {
+		return err
+	}
+
+	return nil
```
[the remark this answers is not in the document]
kwilczynski: Done. Thank you!

cli/cli_app.go
```diff
@@ -384,7 +384,31 @@ func RunTerragrunt(terragruntOptions *options.TerragruntOptions) error {
...
+
+	// If running Terraform resulted in an error that matches one of the init errors,
+	// then run the required interim init, and then re-run previous command as-is,
+	// otherwise return error without making any attempt to recover from it or
+	// retry anything.
+	if err != nil && isErrorRequiringInit(output, err, terragruntOptions) && !isInitCommand {
+		terragruntOptions.Logger.Println("Detected an error that requires Terraform to be re-initialized; running 'terraform init' now...")
+
+		// Run the "terraform init" command.
+		if err = runTerraformInit(terragruntOptions, terragruntConfig, nil); err != nil {
```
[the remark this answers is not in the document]
kwilczynski: Done.

cli/cli_app.go
```diff
@@ -384,7 +384,31 @@ func RunTerragrunt(terragruntOptions *options.TerragruntOptions) error {
...
+			return err
+		}
+
+		// If the init succeeded, re-run previous command again.
+		if err, _ = runTerragruntWithConfig(terragruntOptions, terragruntConfig, false); err != nil {
```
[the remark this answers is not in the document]
kwilczynski: Done.

cli/cli_app.go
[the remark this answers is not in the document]
kwilczynski: Actually. We can't do that. Perhaps different refactor to make this look a little better.

Since if the initial `runTerragruntWithConfig()` returns an error, then it will be set for the remained of the runtime, thus in the end Terragrunt would report an issue even though we recovered (by running `terraform init` successfully):

```bash
[terragrunt] 2020/09/11 19:53:59 Hit multiple errors:
exit status 1
```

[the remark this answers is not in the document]
kwilczynski: Re-factored.

kwilczynski:
Rebased against master.

kwilczynski:
Hi @yorinasub17, is there anything here left to be done?

cli/cli_app.go
```diff
@@ -585,22 +608,24 @@ func setTerragruntInputsAsEnvVars(terragruntOptions *options.TerragruntOptions,
...
+func runTerraformWithRetry(terragruntOptions *options.TerragruntOptions) (error, string) {
 	// Retry the command configurable time with sleep in between
 	for i := 0; i < terragruntOptions.MaxRetryAttempts; i++ {
 		if out, tferr := shell.RunTerraformCommandWithOutput(terragruntOptions, terragruntOptions.TerraformCliArgs...); tferr != nil {
-			if out != nil && isRetryable(out.Stderr, tferr, terragruntOptions) {
-				terragruntOptions.Logger.Printf("Encountered an error eligible for retrying. Sleeping %v before retrying.\n", terragruntOptions.Sleep)
-				time.Sleep(terragruntOptions.Sleep)
-			} else {
-				return tferr
+			if out != nil {
```
brikis98, line 615: Is this change necessary? Or is it a left over due to earlier attempts to do the `init` retries here?

cli/cli_app_test.go
```diff
@@ -132,6 +134,162 @@ func TestTerragruntHandlesCatastrophicTerraformFailure(t *testing.T) {
...
+				ErrorsRequiringInit: []string{
+					"(?s).*some-error.*",
+					"(?s).*a-test-error.*",
+					"(?s).*other-error.*",
+				},
+			},
+			matches: true,
+		},
+		{
+			description: "Matches a multi-line error with ANSI colors from the provided list using regular expression",
```
brikis98, line 269: Great unit tests 👍

test/integration_test.go
```diff
@@ -1198,6 +1199,67 @@ func TestAutoRetryEnvVarWithRecoverableError(t *testing.T) {
...
+
+	stdOut.Reset()
+	stdErr.Reset()
+
+	assert.NoError(t, copyTerragruntFiles(paths["new"]))
+
+	// Step 2: Run "terragrunt apply" against a module that uses different (new) version of the provider.
+	// This will cause initial "apply" to fail, and then recovery for known init errors will run "terraform init"
+	// so that new provider version will be downloaded. Then "apply" will be run again.
+	err = runTerragruntCommand(t, fmt.Sprintf("terragrunt apply -no-color --auto-approve --terragrunt-non-interactive --terragrunt-working-dir %s", modulePath), &stdOut, &stdErr)
```
brikis98, line 1243:
```suggestion
	err = runTerragruntCommand(t, fmt.Sprintf("terragrunt apply -no-color -auto-approve --terragrunt-non-interactive --terragrunt-working-dir %s", modulePath), &stdOut, &stdErr)
```

test/integration_test.go
```diff
@@ -1198,6 +1199,67 @@ func TestAutoRetryEnvVarWithRecoverableError(t *testing.T) {
...
+	assert.Nil(t, err)
```
brikis98, line 1244:
```suggestion
	assert.NoError(t, err)
```

test/integration_test.go
```diff
@@ -1198,6 +1199,67 @@ func TestAutoRetryEnvVarWithRecoverableError(t *testing.T) {
...
+		err    error
+		stdOut bytes.Buffer
+		stdErr bytes.Buffer
+	)
+
+	assert.NoError(t, copyTerragruntFiles(paths["old"]))
+
+	// Step 1: Initialize Terraform and download first version (old) of the provider defined in the module.
+	err = runTerragruntCommand(t, fmt.Sprintf("terragrunt init --terragrunt-non-interactive --terragrunt-working-dir %s", modulePath), &stdOut, &stdErr)
+	assert.Nil(t, err)
```
brikis98, line 1232:
```suggestion
	assert.NoError(t, err)
```

test/integration_test.go
```diff
@@ -1198,6 +1199,67 @@ func TestAutoRetryEnvVarWithRecoverableError(t *testing.T) {
...
+		return nil
+	}
+
+	var (
+		err    error
+		stdOut bytes.Buffer
+		stdErr bytes.Buffer
+	)
+
+	assert.NoError(t, copyTerragruntFiles(paths["old"]))
```
brikis98, line 1228: `assert.XXX` will mark the test as failed, but allow it to keep running. `require.XXX` will mark the test failed and stop the test at that point. In some of the checks here, such as this one, you probably want `require.XXX` instead of `assert.XXX`, as any subsequent output would just be confusing/misleading.

test/integration_test.go
```diff
@@ -1198,6 +1199,67 @@ func TestAutoRetryEnvVarWithRecoverableError(t *testing.T) {
...
+	stdOut.Reset()
+	stdErr.Reset()
+
+	assert.NoError(t, copyTerragruntFiles(paths["invalid"]))
+
+	// Step 3: Run "terragrunt apply" against a module that uses an unavailable (invalid) version of the provider.
+	// This will cause initial "apply" to fail, and then recovery for known init errors will run "terraform init"
+	// that is going to fail causing the entire Terragrunt run to be interrupted with an error.
+	err = runTerragruntCommand(t, fmt.Sprintf("terragrunt apply -no-color --auto-approve --terragrunt-non-interactive --terragrunt-working-dir %s", modulePath), &stdOut, &stdErr)
+	assert.NotNil(t, err)
```
brikis98, line 1257:
```suggestion
	assert.Error(t, err)
```

test/integration_test.go
```diff
@@ -1198,6 +1199,67 @@ func TestAutoRetryEnvVarWithRecoverableError(t *testing.T) {
 	assert.NotContains(t, out.String(), "Apply complete!")
 }
 
+func TestAutomaticInitAfterRecoverableError(t *testing.T) {
```
brikis98, line 1202: Nice test 👍

brikis98 (commented):
Thanks @kwilczynski! I added a few more minor issues to fix, and then this is good to go.

cli/cli_app.go
```diff
@@ -509,7 +529,7 @@ func shouldRunHook(hook config.Hook, terragruntOptions *options.TerragruntOption
 
 // Runs terraform with the given options and CLI args.
 // This will forward all the args and extra_arguments directly to Terraform.
-func runTerragruntWithConfig(terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig, allowSourceDownload bool) error {
+func runTerragruntWithConfig(terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig, allowSourceDownload bool) (error, string) {
```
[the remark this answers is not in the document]
kwilczynski: Doh. Indeed. Thank you.

cli/cli_app.go
kwilczynski: Yes. I had an integration test failing without this. The standard error wasn't set if there was no error resulting in nil dereference.

kwilczynski:
Hi @yorinasub17 and @brikis98,

It appears that either Terraform release 0.13+ has a potential problem or the behaviour has changed at some point as per:

```bash
$ cat > main.tf
provider "null" {
    version = "2.1.1"
}
^D
$ terraform init

Initializing the backend...

- Finding hashicorp/null versions matching "2.1.1"...
- Installing hashicorp/null v2.1.1...
- Installed hashicorp/null v2.1.1 (signed by HashiCorp)

You may now begin working with Terraform. Try running "terraform plan" to see
any changes that are required for your infrastructure. All Terraform commands
should now work.

If you ever set or change modules or backend configuration for Terraform,
rerun this command to reinitialize your working directory. If you forget, other
commands will detect it and remind you to do so if necessary.

$ cat > main.tf
provider "null" {
    version = "9.9.9"
}
^D

$ terraform apply -auto-approve
Apply complete! Resources: 0 added, 0 changed, 0 destroyed.

$ terraform -version
Terraform v0.13.0
+ provider registry.terraform.io/hashicorp/null v2.1.1
```

This is not _quite_ what I would expect, and it breaks integration test, sadly.

brikis98:
Weird. That sounds like a bug.  Perhaps file one in the Terraform repo?

Also, for the integration test, what about switching the `source` URL of a `module` instead of the `provider` version?

eahrend:
Hey, I'm having this issue with updating to terraform 0.13. I'm just going into all my subdirectories with terragrunt.hcl files and re-init'ing them, which seems to work, albeit a bit time consuming. Having a init-all --force command that I can run from my root directory that would force a re-init would be useful.

Issue #1344, opened 2020-09-14
New "dependency" optimisation feature breaks "tfenv"


Fran-ELS-AMS:
The new `dependency` optimisation introduced in `terragrunt` version `0.23.35` breaks `tfenv` (at least on a Mac) when it's used with a `.terraform-version` file in the `terragrunt` root.

I think this is caused by the introduction of a temporary folder [here](https://github.com/gruntwork-io/terragrunt/pull/1311/files#diff-a8f51f0ba302e08284801441eab2c6e1R435-R443)

`terragrunt` changes its working directory to the temporary folder, and when it runs a `terraform` command from there `tfenv` cannot find the `.terraform-version` file that is in the project root, so it will use the system-wide version, which will often be either unset or a different version

This causes issues like this (`tfenv` system-wide version unset):

```
terragrunt plan
tg plan
[terragrunt] [</terragrunt/path>] 2020/09/14 11:35:11 Running command: terraform --version
[terragrunt] 2020/09/14 11:35:13 Terraform version: 0.12.29

...

[terragrunt] [</some/terragrunt/dependency/path] 2020/09/14 11:35:19 Running command: terraform output -json
Version could not be resolved (set by /usr/local/Cellar/tfenv/2.0.0/version or tfenv use <version>)
Version could not be resolved (set by /usr/local/Cellar/tfenv/2.0.0/version or tfenv use <version>)
Version could not be resolved (set by /usr/local/Cellar/tfenv/2.0.0/version or tfenv use <version>)
Version could not be resolved (set by /usr/local/Cellar/tfenv/2.0.0/version or tfenv use <version>)
Version could not be resolved (set by /usr/local/Cellar/tfenv/2.0.0/version or tfenv use <version>)
Version could not be resolved (set by /usr/local/Cellar/tfenv/2.0.0/version or tfenv use <version>)
Version could not be resolved (set by /usr/local/Cellar/tfenv/2.0.0/version or tfenv use <version>)
[terragrunt] 2020/09/14 11:35:22 exit status 1
```

or this (`tfenv` system-wide version different and incompatible):

```
terragrunt plan
tg plan
[terragrunt] [</terragrunt/path>] 2020/09/14 11:35:11 Running command: terraform --version
[terragrunt] 2020/09/14 11:35:13 Terraform version: 0.12.29

...

[terragrunt] [</some/terragrunt/dependency/path] 2020/09/14 11:35:19 Running command: terraform output -json
[...]

PR #1346, opened 2020-09-14
Use the terragrunt download dir for generating the temporary workspace for dependency optimization

yorinasub17:
Partially resolves https://github.com/gruntwork-io/terragrunt/issues/1344

This works for `tfenv` under the default settings because the `terragrunt-cache` will be in the same directory as the terragrunt config, and thus will be in the right spot in the folder tree of `tfenv` terraform versions.

Note that this will NOT work if the download dir is configured to somewhere outside of the tree, but currently that flow breaks `tfenv` even under normal usage (because we don't copy hidden files and folders, so `.terraform-version` does not get copied over).

### DIFF
--- a/config/dependency.go
+++ b/config/dependency.go
@@ -437,12 +437,18 @@ func getTerragruntOutputJsonFromRemoteState(
 ) ([]byte, error) {
 	util.Debugf(terragruntOptions.Logger, "Detected remote state block with generate config. Resolving dependency by pulling remote state.")
 
-	// Create working directory where we will run terraform in. Make sure it is cleaned up before the function returns.
-	tempWorkDir, err := ioutil.TempDir("", "")
+	// Create working directory where we will run terraform in. We will create the temporary directory in the download
+	// directory for consistency with other file generation capabilities of terragrunt. Make sure it is cleaned up
+	// before the function returns.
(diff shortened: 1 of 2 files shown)


brikis98 (approved):
LGTM

Fran-ELS-AMS:
As I also reported in #1344:

> It seems to be working perfectly.
> 
> I've tested with:
> - the system-wide `terraform` version for `tfenv` **unset**
> - the system-wide `terraform` version for `tfenv` **set to a version that's incompatible with the current version of `terragrunt` (`0.12.26`)**
> 
> Before each test I made sure to remove all `.terragrunt-cache` folders with:
> ```
> find . -type d -name ".terragrunt-cache" -prune -exec rm -rf {} \;
> ```
> Both tests were successful.
> 
> I then re-run both tests without removing the `.terragrunt-cache` folders, just in case, to test both cases.
> 
> Both tests were successful again.
> 
> The temporary directories, containing only the `backend.tf` file, get created (one each) in the `.terragrunt-cache` folder of each dependency for which `terragrunt` runs `terraform output -json` and it almost immediately gets removed, leaving behind an empty `.terragrunt-cache` folder.

LGTM

yorinasub17:
Thanks for review! Will merge and release now.

PR #1347, opened 2020-09-15
Corrects documentation of --terragrunt-check

owsygavin:
The documentation was wrong here - this parameter makes the command return 1 if the command finds problems, and 0 if it does not.

I've also reworded it for clarity.

Thanks,
- Gavin


brikis98 (approved):
Good catch! Thanks for the fix. Merging now.

PR #1357, opened 2020-09-17
Update repo to use TF 0.13

brikis98:
1. Update CI to use latest Docker image, which has TF 0.13 installed.
1. Add pre-commit hooks to run `terraform fmt` and `go fmt`. 
1. Fix formatting on all `terraform fmt` issues.
1. Add `-auto-approve` to all the integration tests. How on earth did it work without this before???
1. Remove a trailing space from the `command` passed to `runTerragrunt` in a bunch of tests in `integration_test.go`. Under the hood, `runTerragrunt` calls `strings.Split` to split the `command` into a bunch of CLI args, and with the trailing space, it was passing it as an explicit arg... Terraform, in turn, was interpreting this as the folder path where it should run. Of course, a folder called space (`" "`) doesn't exist, so you get a super confusing test failure about `stat: no such file or directory` with no more information.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,8 +20,7 @@ jobs:
             # Oct 26, 2019: Install the last known working version of pre-commit. Also, we have to pin the version of
             # transitive dependencies that are being pulled in which released new versions that are no longer compatible
             # with any python < 3.6.
-            pyenv global 3.5.2
-            pip install pre-commit==1.21.0 cfgv==2.0.1 zipp==1.1.0
+            pip3 install pre-commit==1.21.0 cfgv==2.0.1 zipp==1.1.0
             pre-commit install
             pre-commit run --all-files
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ jobs:
     <<: *defaults
     steps:
       - checkout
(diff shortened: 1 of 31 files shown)


brikis98:
Mentioned the space issue in https://github.com/hashicorp/terraform/issues/15932#issuecomment-694423725.

brikis98:
I'm super confused how the tests in `integration_test.go` were working before... Or if they were working at all... But all tests are passing now and this is ready for review!

feliperfmarques:
Thanks for this PR 👏 👏 👏

.circleci/config.yml
```diff
@@ -16,6 +13,16 @@ jobs:
     steps:
       - checkout
       - run: gruntwork-install --binary-name 'terratest_log_parser' --repo 'https://github.com/gruntwork-io/terratest' --tag 'v0.13.8'
```
yorinasub17, line 15: NIT: While we are here, we should bump this to `v0.29.0`.

yorinasub17 (approved):
LGTM!

.circleci/config.yml
brikis98: Done: https://github.com/gruntwork-io/terragrunt/pull/1357/commits/c2185a629981636ae9238d18cf9516ac9d2ad3b7

brikis98:
One more review plz?

yorinasub17 (approved):

brikis98:

PR #1365, opened 2020-09-21
Add regression test to confirm vars from env var in apply-all works

yorinasub17:
There was confusion from one of our users regarding whether or not setting tf vars via env vars works with `apply-all`, so I created a regression test for it. AFAIK, none of the existing tests tests this combination.


brikis98 (approved):
LGTM!

yorinasub17:
Thanks for review! Will merge this in now.

PR #1386, opened 2020-10-08
Simplifying expression

jtyr:
This PR is removing the variable interpolation and trailing comma from the expression which improves its readability.


brikis98 (approved):
Good idea, thanks!

PR #1391, opened 2020-10-15
Update aws-provider-patch to support nested attributes

brikis98:
I have come across yet another `import` bug with Terraform. The [original one](https://github.com/hashicorp/terraform/issues/13018) prompted us to add the `aws-provider-patch` command. The [new one](https://github.com/hashicorp/terraform/issues/26211), which is the motivation for this PR, requires that the `aws-provider-patch` command can override not only top-level attributes in the `provider` block, but also attributes nested within blocks within the `provider` block, such as a `role_arn` within an `assume_role` block. This PR allows you to specify these nested attributes using a "dot" notation: e.g., `--terragrunt-override-attr assume_role.role_arn=""`.

### DIFF
--- a/cli/aws_provider_patch.go
+++ b/cli/aws_provider_patch.go
@@ -12,6 +12,7 @@ import (
 	"github.com/zclconf/go-cty/cty"
 	"io/ioutil"
 	"path/filepath"
+	"strings"
 )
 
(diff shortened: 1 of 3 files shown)


ina-stoyanova (approved):
Looks very good. I like the explanations - make it very easy to understand what's going on

brikis98:

PR #1399, opened 2020-10-22
configValuesEqual should have no side effects

yorinasub17:
I have no idea how this was working before, but @ina-stoyanova and @zackproser found a bug in terragrunt where we were stripping out the terragrunt only config vals before parsing out the extended s3 config object.

Digging in, I found out that this was because we were stripping that out in the `configValuesEqual` function using `delete`. I fixed this by removing the side effect from the equals function.

NOTE: I attempted to create a regression test for this, but I had difficulty reproducing the behavior in a test environment. I was able to repro this behavior, but it is not exactly clear when it happens because it relies on `.terragrunt-cache` already having been init with an s3 bucket AND it is creating a new s3 bucket.

### DIFF
--- a/remote/remote_state_s3.go
+++ b/remote/remote_state_s3.go
@@ -117,6 +117,17 @@ func (s3Initializer S3Initializer) NeedsInitialization(remoteState *RemoteState,
 		return false, nil
 	}
 
+	// Nowadays it only makes sense to set the "dynamodb_table" attribute as it has
+	// been supported in Terraform since the release of version 0.10. The deprecated
+	// "lock_table" attribute is either set to NULL in the state file or missing
+	// from it altogether. Display a deprecation warning when the "lock_table"
+	// attribute is being used.
+	if util.KindOf(remoteState.Config["lock_table"]) == reflect.String && remoteState.Config["lock_table"] != "" {
+		terragruntOptions.Logger.Printf("%s\n", lockTableDeprecationMessage)
+		remoteState.Config["dynamodb_table"] = remoteState.Config["lock_table"]
+		delete(remoteState.Config, "lock_table")
+	}
+
 	if !configValuesEqual(remoteState.Config, existingBackend, terragruntOptions) {
 		return true, nil
 	}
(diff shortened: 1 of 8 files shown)


ina-stoyanova (dismissed):
Hey @yorinasub17 thanks for the PR! It was suspicious behaviour, so your fix is super helpful & on time - unblocks me for moving forward! 

Awesome work :)

yorinasub17:
Thanks for review!

I decided to add in the regression test anyway, even though it doesn't properly demonstrate the broken behavior, because at least it gives me confident that the code did not depend on the manipulation in `configValuesEqual` for the happy path (and thus this refactoring is legit).

... So needs another review please!

remote/remote_state_s3.go
```diff
@@ -117,6 +117,17 @@ func (s3Initializer S3Initializer) NeedsInitialization(remoteState *RemoteState,
...
 
+	// Nowadays it only makes sense to set the "dynamodb_table" attribute as it has
+	// been supported in Terraform since the release of version 0.10. The deprecated
+	// "lock_table" attribute is either set to NULL in the state file or missing
+	// from it altogether. Display a deprecation warning when the "lock_table"
+	// attribute is being used.
+	if util.KindOf(remoteState.Config["lock_table"]) == reflect.String && remoteState.Config["lock_table"] != "" {
+		terragruntOptions.Logger.Printf("%s\n", lockTableDeprecationMessage)
+		remoteState.Config["dynamodb_table"] = remoteState.Config["lock_table"]
+		delete(remoteState.Config, "lock_table")
```
ina-stoyanova, line 128: just to understand better, this change means now we're going to look at the `remoteState.config`

ina-stoyanova (approved):

ina-stoyanova:
Looks good, Yori!

zackproser (approved):
LGTM :+1:

remote/remote_state_s3.go
yorinasub17: Actually, this is doing the exact same thing as before within `configValuesEqual`, because go maps pass by reference. I am just moving the routine outside of the function to:

- make the map manipulation on `remoteState.Config` more explicit.
- ensure that the `configValuesEqual` function is side effect free.

yorinasub17:

remote/remote_state_s3.go
```diff
@@ -193,23 +204,16 @@ func configValuesEqual(config map[string]interface{}, existingBackend *Terraform
...
-	// attribute is being used.
-	if util.KindOf(config["lock_table"]) == reflect.String && config["lock_table"] != "" {
-		terragruntOptions.Logger.Printf("%s\n", lockTableDeprecationMessage)
-		config["dynamodb_table"] = config["lock_table"]
-		delete(config, "lock_table")
-	}
-
-	// Delete custom S3 and DynamoDB tags that are only used in Terragrunt config and not in Terraform's backend
-	for _, key := range terragruntOnlyConfigs {
-		delete(config, key)
```
brikis98, line 209: That does seem like a bad side effect to have in `configValuesEqual`, but as a sanity check, are we removing or ignoring these Terragrunt-specific configs somewhere else now? That is, we shouldn't be calling `terraform init -backend-config=xxx=yyy` where `xxx` is some Terragrunt-only flag.

yorinasub17: Yes: https://github.com/gruntwork-io/terragrunt/blob/master/remote/remote_state_s3.go#L271

Note that the generator calls this function to get the stripped down config: https://github.com/gruntwork-io/terragrunt/blob/master/remote/remote_state.go#L181

brikis98: Roger, thx for confirming!

Issue #1333, opened 2020-09-09
Access Logging setup causes logging loop.


tony-harverson-moonpig:
After #645 being merged, Terragrunt is setting up Access_Logging on terraform state buckets, logging access to the prefix TFStateLogs/ in the same bucket.  This causes an access logging loop causing there to be lots of logs. It's also against Amazon's recommendations:

https://aws.amazon.com/premiumsupport/knowledge-center/s3-server-access-logs-same-bucket/
https://docs.aws.amazon.com/AmazonS3/latest/dev/ServerLogs.html#server-access-logging-overview

I'd suggest the "skip_bucket_accesslogging" option should default to false on bucket creation.

If I PR'ed that change, would it be likely to be merged?

brikis98:
So you're saying that if you send access logs for bucket X to bucket X, AWS gets in an infinite loop where it seems the access logs themselves as a new "access" and logs that too, and so on, infinitely?

tony-harverson-moonpig:
According to Amazon, That's what happens.  

Replication recipe:
Spin up a new bucket by running a terragrunt apply on any minimal terraform stack.
list the bucket in the aws console.  No TFStateLogs will be present initially (not sure why this is, maybe there's a setup delay or something).

Inspect the bucket in an hour's time.  We're seeing logs being continually written to that bucket, despite to additional access.

brikis98:
Yikes 😮  Thanks for reporting.

I think then we'd want to change the logic to use a separate bucket for access logging. Perhaps we pick some name pattern by default (`xxx-access-logs` where `xxx` is the original bucket name?) and allow it to be overridden via config. A PR to fix this is VERY welcome.

tony-harverson-moonpig:
With derived bucket names, there's an edge case around the lengths of bucket names (63 characters max), so long bucket names could result in illegally long access bucket names (which would assumedly fail creation), or would need to be truncated before adding the logging suffix (which would result in possibly non-unique names).  

It might be better to have a single logs bucket with a predictable name, and a unique prefix (say /TFStateLogs/{tfstate-bucket-name}/) for each.

brikis98:
Ah, that makes sense. In that case, let's make access logging off by default, and allow it to be turned with one config parameter, that also requires you to set the access logs bucket name in another param.

tony-harverson-moonpig:
I'l try to get a PR in for that in the next couple of days.

brikis98:
Thank you!

[...]

PR #1400, opened 2020-10-23
Fix for s3 TF state bucket logs loop - issue 1333

ina-stoyanova:
This PR is aiming to fix this issue: https://github.com/gruntwork-io/terragrunt/issues/1333

- [x] Fix related bug (pre-existing) to this issue - [more details can be found here.](https://github.com/gruntwork-io/terragrunt/pull/1399) Thanks to @yorinasub17 & @zackproser 
- [x] Decide what to do with `skip_bucket_access_logging` - do we deprecate or fix forward & replace with `enable_bucket_access_logging`)
    - see below in comments 
- [x] Implement `enable_bucket_access_logging` as bool to flip logic around - e.g. if not provided, by default the logging will be turned off for the bucket. _**Thanks to @infraredgirl**_
- [x] implement `enable_bucket_access_logging` as a string - if provided, we enable access logging for the bucket & use the string param as a name for the bucket; if not provided - disable access logging by default _**Thanks to @infraredgirl**_
- [x] implement create bucket for access logs if name is provided, but bucket does not exist yet
- ~[x] refactor for reausability the s3 create & set access methods~
- [x] refactor to avoid duplication `createS3BucketIfNecessary` & `CreateLogsS3BucketIfNecessary` - potentially we can pass function call as a param?
- [x] update/ add to automated tests
- [x] update logging so it's clear that a setting has been turned on:
      - `EnableVersioningForS3Bucket` (e.g. enabling..... -> enabled.....)
      - `EnableSSEForS3BucketWide`
      - `EnableAccessLoggingForS3BucketWide`
      - `EnablePublicAccessBlockingForS3Bucket`
      - `configureBucketAccessLoggingAcl`


ina-stoyanova:
[last-terragrunt-apply-output.txt](https://github.com/gruntwork-io/terragrunt/files/5430691/last-terragrunt-apply-output.txt)
☝️ this is the output of the code ran at the last commit state. 

All the logs are expected for the following terragrunt config:
```
# stage/terragrunt.hcl
remote_state {

  backend = "s3"
  generate = {
    path      = "backend.tf"
  }

  config = {
    bucket = "ina-1"
    accesslogging_bucket_name = "ina-1-logs"
    key = "${path_relative_to_include()}/terraform.tfstate"
    region         = "eu-west-2"
    encrypt        = true
    dynamodb_table = "ina-1"
  }
}
```

ina-stoyanova:
Next steps to pick up:
- confirm we can still see the same behaviour as the logs above show
- confirm if param is not provided that S3 bucket for logs is not created
- confirm if we need to sanitise input for name of s3 bucket
- confirm what happens to `skip...` parameter
- refactor duplication between createS3 bucket methods (for logs, and for standard state bucket)

ina-stoyanova:
As part of this PR, the default behaviour of terragrunt's functionality to create the TF State S3 bucket will change. 

**- previous implementation:** create TF State S3 bucket with server access logging enabled & set to the origin (same) S3 bucket. _For more details, please read through https://github.com/gruntwork-io/terragrunt/issues/1333, demonstrating why this is undesirable behaviour._  

**- new implementation with this PR"** create TF State S3 bucket with server access logging disabled. Terragrunt users will be able to enable server access logging for the TF State S3 bucket by providing a name for the desired logs target bucket. This can be done using the `accesslogging_bucket_name <string>` parameter in the `terragrunt.hcl` configuration of their implementation. 

In more detail:
- The `skip_bucket_accesslogging` terragrunt option is being deprecated. It's no longer going to take effect on the S3 bucket for TF state creation and logging configuration. 
- A new terragrunt configuration option is added: `accesslogging_bucket_name <string>`
    - states in the `terragrunt.hcl` configuration file
         - not present -> server access logging disabled by default for TF State S3 bucket
         - empty -> server access logging disabled by default for TF State S3 bucket
         - present (valid string) -> the server access logging for the TF State S3 bucket will be enabled; the target logging bucket will be the S3 bucket with name as specified by the configuration option value `accesslogging_bucket_name` 
              - if bucket with such name does not exist, the user will be prompted to create the S3 bucket.


Deprecating `skip_bucket_accesslogging`:

_We've got 3 options identified as to how to proceed. After chatting to @brikis98 we've decided it's best to go with a combination of **1** & **2**. Option 3 is the most ideal scenario one, and feels like it's outside of the scope of this PR currently. As a way to see what's the engagement/appetite for option 3, here's an issue raised on the Terragrunt repo. _

1. basic option - remove the flag & update the docs & release notes 
2. we detect the param - show error & provide guidance 
3. warn people of existing bug & point them to fix it/or we automatically fix it for them (existing buckets)

ina-stoyanova:
Tests have been updated very little. The reasoning is, the important thing is that the terraform state is initialised using s3 backend. I couldn't find any other examples of the other config options being tested. 

For the `accesslogging_bucket_name` config option, it feels like the effort to automate the testing of it will be more than the value of it would actually bring. I'm happy to take guidance if anyone feels strongly we should have the tests cover this explicitly. If that's the direction we take, I'd suggest doing it separately, so we can also test the rest of the properties alongside with this one, which would increase the test confidence in what the `terraform init` will affect if run with terragrunt. 

I have tested manually the following cases:
- not providing `accesslogging_bucket_name` => ✅
- providing the value `accesslogging_bucket_name` => ✅ 
- providing incorrect S3 name for the bucket => ✅ 
- providing `accesslogging_bucket_name` as incorrect type of input => ✅ 
- keeping the `skip_bucket_accesslogging` => ✅ 
- providing value (bool) for the `skip_bucket_accesslogging` => ✅

remote/remote_state_s3.go
```diff
@@ -341,32 +343,20 @@ func validateS3Config(extendedConfig *ExtendedRemoteStateConfigS3, terragruntOpt
...
-			sleepBetweenRetries := 10 * time.Second
-
-			return util.DoWithRetry(description, maxRetries, sleepBetweenRetries, terragruntOptions.Logger, func() error {
-				return CreateS3BucketWithVersioningSSEncryptionAndAccessLogging(s3Client, config, terragruntOptions)
-			})
-		}
-	}
-
-	return nil
+func createS3BucketWithRetry(s3Client *s3.S3, config *ExtendedRemoteStateConfigS3, terragruntOptions *options.TerragruntOptions) error {
```
yorinasub17, line 346: This change of moving `DoesS3BucketExist` check to `CreateS3Bucket` makes me nervous because it is a fairly significant change in behavior.

Before, terragrunt never attempted to "migrate" an existing bucket to match the configurations and defaults that you have set. This means that even if we update bucket parameters across releases, and if the user upgrades terragrunt versions, the s3 bucket remained untouched. This is somewhat surprising and confusing behavior for folks who are used to terraform, but has a few advantages for us. Off the top of my head:

- It is simpler to maintain, as we didn't really have to worry about backwards compatibility in the s3 bucket creation code: we always started from scratch with a new bucket.

- It doesn't have the behavior of modifying the state bucket during a `plan` or `validate` stage, which can be surprising to a user attempting to do a "read-only" task (NOTE: creating the bucket also happens during these stages, which is arguably just as surprising, but in the create case there is an interactive prompt to confirm the creation that softens the impact).

There is potential that this change can break a lot of existing users, and we also should think through the implications of applying some of the updates that this will do. From the looks of it, this change appears to be orthogonal to the access logging change, so I think this specific update should be in a separate PR where we can discuss the implications of this update.

remote/remote_state_s3.go
```diff
@@ -435,9 +420,19 @@ func CreateS3BucketWithVersioningSSEncryptionAndAccessLogging(s3Client *s3.S3, c
 	}
 
 	if config.SkipBucketAccessLogging {
-		terragruntOptions.Logger.Printf("Access Logging is disabled for the remote state AWS S3 bucket %s using 'skip_bucket_accesslogging' config.", config.remoteStateConfigS3.Bucket)
-	} else if err := EnableAccessLoggingForS3BucketWide(s3Client, &config.remoteStateConfigS3, terragruntOptions); err != nil {
-		return err
+		terragruntOptions.Logger.Printf("___WARNING___: The terragrunt configuration option 'skip_bucket_accesslogging' is now deprecated. To enable access logging for bucket %s, please provide property `accesslogging_bucket_name` in the terragrunt config file. For more details, please refer to the Terragrunt documentation.", config.remoteStateConfigS3.Bucket)
```
yorinasub17, line 423:
```suggestion
		terragruntOptions.Logger.Printf("___WARNING___: The terragrunt configuration option 'skip_bucket_accesslogging' is now deprecated. Access logging for the state bucket %s is disabled by default. To enable access logging for bucket %s, please provide property `accesslogging_bucket_name` in the terragrunt config file. For more details, please refer to the Terragrunt documentation.", config.remoteStateConfigS3.Bucket, config.remoteStateConfigS3.Bucket)
```

remote/remote_state_s3.go
```diff
@@ -435,9 +420,19 @@ func CreateS3BucketWithVersioningSSEncryptionAndAccessLogging(s3Client *s3.S3, c
...
+	}
+
+	if config.AccessLoggingBucketName != "" {
+		if err := CreateS3Bucket(s3Client, aws.String(config.AccessLoggingBucketName), terragruntOptions); err != nil {
+			terragruntOptions.Logger.Printf("Access Logging is disabled for the remote state AWS S3 bucket for bucket %s", config.remoteStateConfigS3.Bucket)
```
yorinasub17, line 428: This message needs to be updated.

remote/remote_state_s3.go
```diff
@@ -435,9 +420,19 @@ func CreateS3BucketWithVersioningSSEncryptionAndAccessLogging(s3Client *s3.S3, c
 	}
 
 	if config.SkipBucketAccessLogging {
-		terragruntOptions.Logger.Printf("Access Logging is disabled for the remote state AWS S3 bucket %s using 'skip_bucket_accesslogging' config.", config.remoteStateConfigS3.Bucket)
-	} else if err := EnableAccessLoggingForS3BucketWide(s3Client, &config.remoteStateConfigS3, terragruntOptions); err != nil {
-		return err
+		terragruntOptions.Logger.Printf("___WARNING___: The terragrunt configuration option 'skip_bucket_accesslogging' is now deprecated. To enable access logging for bucket %s, please provide property `accesslogging_bucket_name` in the terragrunt config file. For more details, please refer to the Terragrunt documentation.", config.remoteStateConfigS3.Bucket)
+	}
+
+	if config.AccessLoggingBucketName != "" {
```
yorinasub17, line 426: There should be a log message here indicating that terragrunt is going to create the access logging bucket and enable it.

yorinasub17 (commented):
The updated implementation LGTM. The deprecation policy we've implemented here makes sense and is approachable.

I did have a concern about one of the changes that was implemented here, which could use a second pair of eyes to confirm if my theory is correct and if my concerns are valid. See the comment below.

Regarding testing, it would be good to have a sanity check regression integration test that makes sure terragrunt sets up access logging to the right bucket. It doesn't have to be comprehensive/complicated where it validates if access logs actually get generated, but an existence and config sanity check to make sure the API calls are correct and happen in the right order would be useful.

remote/remote_state_s3.go
ina-stoyanova: Thank @yorinasub17 This is really useful insight. Here's some of my thoughts, however, based on the concerns you've got, it might be easier to chat through this on a call so I can understand better _(I suspect I'm missing some context & understanding)_

- the `CreateS3Bucket` func has only been called from 2 places, one of which is the new introduction for creating logs bucket, and the other one being `CreateS3BucketWithVersioningSSEncryptionAndAccessLogging`
- the `CreateS3BucketWithVersioningSSEncryptionAndAccessLogging` func is also only called from 1 more place - the `createS3BucketWithRetry` method and following the chain of functioncalls, only when `func (remoteState *RemoteState) Initialize(...)` is called.
- based on the above 2 observations, I'm not following exactly where the concern that this behaviour will impact existing users will come from. As far as I see, this behaviour of `DoesS3BucketExist` has been pushed closer to the actual s3 client API call to create the bucket, but it doesn't ultimately change the behaviour we execute when calling into `terragrunt apply | plan | validate`
- we do get the prompt every time to create the bucket, even if `terragrunt validate` has been called - the user should still be prompted to see if they want the S3 bucket created or not

ina-stoyanova: I agree that this change can be extracted into its own separate PR, so I'll raise a separate one for it, so we can have an appropriate discussion.

remote/remote_state_s3.go
ina-stoyanova: Hm, I'm unsure what you meant here

remote/remote_state_s3.go
ina-stoyanova: did you mean something like in the method being called (See lines 654-656). I think it's a fair point though, as it will read easier if the logging as all done at the same level.

remote/remote_state_s3.go
ina-stoyanova: ahh. ignore the above ☝️ I _think_ I understand your train of thought: the error message is about the creation of the bucket, not the access logging 👍 Let me know if my understadning is correct.

remote/remote_state_s3.go
yorinasub17: The concern I have is that `CreateS3BucketWithVersioningSSEncryptionAndAccessLogging` will do additional mutating actions to the bucket after `CreateS3Bucket` is called, specifically:

- `EnableEnforcedTLSAccesstoS3Bucket`
- `EnablePublicAccessBlockingForS3Bucket`
- `TagS3Bucket`
- `EnableVersioningForS3Bucket`
- `EnableSSEForS3BucketWide`
- `EnableAccessLoggingForS3BucketWide`

Before, this whole routine was skipped when the bucket already existed, meaning that we did not attempt to mutate that bucket to enable these features. Now, we are only skipping the create routine, meaning that we will attempt to do these mutations on that existing bucket, which may have been created before all of this code existed. My main concern here is that we may add further mutations in this routine (like we are doing now in this PR), and we will need to make sure that those mutations work even if the bucket was created in older versions of terragrunt, potentially as far back as `0.18`.

ina-stoyanova: ahaa! got it - yeah, totally missed this! Ok, just about to raise the separate PR for this.

remote/remote_state_s3.go
yorinasub17: Yup that is correct. E.g., I would switch this to `("Error creating access logging bucket: %s", err)`.

remote/remote_state_s3.go
yorinasub17: No I mean that since this is doing multiple steps, it would help to have a log message here indicating the two steps it is doing (creating access logging bucket and then hooking it up). I think that is slightly better in terms of debuggability.

It's not a big deal though so if you disagree that is helpful, feel free to ignore (hence labeled as `nit`).

ina-stoyanova: ah I see! that's a good suggestion indeed! will get it in there :)

ina-stoyanova:
I've not also gone ahead and added a new integration test as suggested @yorinasub17. This is still [WIP] as I needed to touch `terratest` as such functionality wasn't there yet. [Please see this Draft PR. ](https://github.com/gruntwork-io/terratest/pull/689)

I'm worrying that by touching terratest as part of this, it might be a bit of a scope creep, but the changes seem quite quick anyway. What's your take on this?

yorinasub17:
> What's your take on this?

You can always start by having the function as a private function in the repo that will use it (in this case `terragrunt`), and then add a comment or file a TODO to port it to terratest. Even better if you start the PR in terratest (like you did here) so that we are less likely to forget it.

The main goal here is to disassociate merging this in from the terratest PR so that they can be released at different times. That way, if you end up in a massive yak shave to generalize the function or we have extended conversations about whether or not the function makes sense in terratest, it doesn't block this PR.

ina-stoyanova:
@yorinasub17 Have a look at the issue I've created [here](https://github.com/gruntwork-io/terragrunt/issues/1404) for the purposes of the discussion in [your comment ](https://github.com/gruntwork-io/terragrunt/pull/1400#discussion_r513017599)

remote/remote_state_s3.go
ina-stoyanova: > This change of moving `DoesS3BucketExist` check to `CreateS3Bucket` makes me nervous because it is a fairly significant change in behavior.
> 
> Before, terragrunt never attempted to "migrate" an existing bucket to match the configurations and defaults that you have set. This means that even if we update bucket parameters across releases, and if the user upgrades terragrunt versions, the s3 bucket remained untouched. This is somewhat surprising and confusing behavior for folks who are used to terraform, but has a few advantages for us. Off the top of my head:
> 
> * It is simpler to maintain, as we didn't really have to worry about backwards compatibility in the s3 bucket creation code: we always started from scratch with a new bucket.
> * It doesn't have the behavior of modifying the state bucket during a `plan` or `validate` stage, which can be surprising to a user attempting to do a "read-only" task (NOTE: creating the bucket also happens during these stages, which is arguably just as surprising, but in the create case there is an interactive prompt to confirm the creation that softens the impact).
> 
> There is potential that this change can break a lot of existing users, and we also should think through the implications of applying some of the updates that this will do. From the looks of it, this change appears to be orthogonal to the access logging change, so I think this specific update should be in a separate PR where we can discuss the implications of this update.



> here is to disassociate merging this in

So, I just pushed the last bit of changes i've added to this repo, but so far the test is not proving anything, as it points to  the not-implemented function in Terratest yet. 

Do you suggest that I implement a temp function using the s3client here, or is leaving it as a commented `TODO` like you've pointed out enough?

ina-stoyanova:
> > What's your take on this?
> 
> You can always start by having the function as a private function in the repo that will use it (in this case `terragrunt`), and then add a comment or file a TODO to port it to terratest. Even better if you start the PR in terratest (like you did here) so that we are less likely to forget it.
> 
> The main goal here is to disassociate merging this in from the terratest PR so that they can be released at different times. That way, if you end up in a massive yak shave to generalize the function or we have extended conversations about whether or not the function makes sense in terratest, it doesn't block this PR.

Roger that! Thanks for the support!

So, I just pushed the last bit of changes I've added to this repo, but so far the test is not proving anything, as it points to the not yet implemented function in Terratest.

Do you suggest that I implement a temp function using the s3client in terragrunt & get the test passing, or is leaving it as a commented TODO like you've pointed out enough?

remote/remote_state_s3.go
```diff
@@ -435,14 +433,45 @@ func CreateS3BucketWithVersioningSSEncryptionAndAccessLogging(s3Client *s3.S3, c
...
-	} else if err := EnableAccessLoggingForS3BucketWide(s3Client, &config.remoteStateConfigS3, terragruntOptions); err != nil {
-		return err
+		terragruntOptions.Logger.Printf("___WARNING___: The terragrunt configuration option 'skip_bucket_accesslogging' is now deprecated. Access logging for the state bucket %s is disabled by default. To enable access logging for bucket %s, please provide property `accesslogging_bucket_name` in the terragrunt config file. For more details, please refer to the Terragrunt documentation.", config.remoteStateConfigS3.Bucket, config.remoteStateConfigS3.Bucket)
+	}
+
+	if config.AccessLoggingBucketName != "" {
+		terragruntOptions.Logger.Printf("Enabling bucket-wide Access Logging on AWS S3 bucket %s - using as TargetBucket %s", config.remoteStateConfigS3.Bucket, config.AccessLoggingBucketName)
+
+		if err := CreateLogsS3BucketIfNecessary(s3Client, aws.String(config.AccessLoggingBucketName), terragruntOptions); err != nil {
+			terragruntOptions.Logger.Printf("Could not create logs bucket for AWS S3 bucket %s", config.remoteStateConfigS3.Bucket)
```
yorinasub17, line 443: NIT: Consider including the access logging bucket name in the log message.

remote/remote_state_s3.go
```diff
@@ -435,14 +433,45 @@ func CreateS3BucketWithVersioningSSEncryptionAndAccessLogging(s3Client *s3.S3, c
...
+			return err
+		}
+
+		if err := EnableAccessLoggingForS3BucketWide(s3Client, &config.remoteStateConfigS3, terragruntOptions, config.AccessLoggingBucketName); err != nil {
+			return err
```
yorinasub17, line 448: NIT: consider adding a log message here, similar to above for the create error.

test/integration_test.go
```diff
@@ -595,6 +595,45 @@ func TestTerragruntHonorsS3RemoteStateSkipFlagsRegression(t *testing.T) {
...
+func TestTerragruntSetsAccessLoggingForTfSTateS3BuckeToADifferentBucket(t *testing.T) {
+	t.Parallel()
+
+	examplePath := filepath.Join(TEST_FIXTURE_REGRESSIONS, "accesslogging-bucket-name")
+	cleanupTerraformFolder(t, examplePath)
+	// Also clean up generated backend file
+	removeFile(t, filepath.Join(examplePath, "backend.tf"))
+
+	s3BucketName := fmt.Sprintf("terragrunt-test-bucket-%s", strings.ToLower(uniqueId()))
+	s3BucketLogsName := fmt.Sprintf("terragrunt-test-bucket-%s-tf-state-logs", strings.ToLower(uniqueId()))
```
yorinasub17, line 609: NIT:
```suggestion
	s3BucketLogsName := fmt.Sprintf("%s-tf-state-logs", s3BucketName)
```

yorinasub17 (commented):
Updates LGTM! Found a few more minor nits, but otherwise the changes makes sense to me!

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -238,8 +238,7 @@ For the `s3` backend, the following additional properties are supported in the `
 - `dynamodb_table` - (Optional) The name of a DynamoDB table to use for state locking and consistency. The table must have a primary key named LockID. If not present, locking will be disabled.
 - `skip_bucket_versioning`: When `true`, the S3 bucket that is created to store the state will not be versioned.
 - `skip_bucket_ssencryption`: When `true`, the S3 bucket that is created to store the state will not be configured with server-side encryption.
-- `skip_bucket_accesslogging`: When `true`, the S3 bucket that is created to store the state will not be configured with
-  access logging.
+- `skip_bucket_accesslogging`: _DEPRECATED_ If provided, will be ignored. A log warning will be issues in the console output to notify the user.
```
brikis98, line 241: NIT:

```suggestion
- `skip_bucket_accesslogging`: _DEPRECATED_ If provided, will be ignored. A log warning will be issued in the console output to notify the user.
```

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -248,6 +247,7 @@ For the `s3` backend, the following additional properties are supported in the `
 - `disable_aws_client_checksums`: When `true`, disable computing and checking checksums on the request and response,
   such as the CRC32 check for DynamoDB. This can be used to workaround
   https://github.com/gruntwork-io/terragrunt/issues/1059.
+- `accesslogging_bucket_name`: (Optional) When provided as a valid `string`, the S3 bucket that is created to store the terraform state will create an additional S3 bucket as a `target bucket` for the logs output. If not provided, or sting is empty or invalid S3 bucket name, then server access logging for the S3 bucket storing the terraform state will be disabled.
```
brikis98, line 250:
```suggestion
- `accesslogging_bucket_name`: (Optional) When provided as a valid `string`, create an S3 bucket with this name to store the access logs for the S3 bucket used to store Terraform state. If not provided, or string is empty or invalid S3 bucket name, then server access logging for the S3 bucket storing the terraform state will be disabled.
```

Also, shouldn't empty string or invalid bucket name or any other invalid input produce an error, rather than just silently disabling access logging?

remote/remote_state_s3.go
```diff
@@ -387,21 +389,17 @@ func checkIfVersioningEnabled(s3Client *s3.S3, config *RemoteStateConfigS3, terr
...
-			terragruntOptions.Logger.Printf("Looks like someone is creating bucket %s at the same time. Will not attempt to create it again.", config.remoteStateConfigS3.Bucket)
+		if isBucketAlreadyOwnedByYouError(err) {
+			terragruntOptions.Logger.Printf("Looks like you're already creating bucket %s at the same time. Will not attempt to create it again.", config.remoteStateConfigS3.Bucket)
 			return WaitUntilS3BucketExists(s3Client, &config.remoteStateConfigS3, terragruntOptions)
 		}
 
 		return err
 	}
 
-	if err := WaitUntilS3BucketExists(s3Client, &config.remoteStateConfigS3, terragruntOptions); err != nil {
```
brikis98, line 401: Apologies if I missed the discussion around this, but why remove the wait?

remote/remote_state_s3.go
```diff
@@ -435,14 +433,45 @@ func CreateS3BucketWithVersioningSSEncryptionAndAccessLogging(s3Client *s3.S3, c
...
+		shouldCreateBucket, err := shell.PromptUserForYesNo(prompt, terragruntOptions)
+		if err != nil {
+			return err
+		}
+
+		if shouldCreateBucket {
+			err := CreateS3Bucket(s3Client, logsBucketName, terragruntOptions)
+			if err != nil {
+				return err
+			}
```
brikis98, line 469: NIT:

```suggestion
			return CreateS3Bucket(s3Client, logsBucketName, terragruntOptions)
```

test/integration_test.go
```diff
@@ -595,6 +595,47 @@ func TestTerragruntHonorsS3RemoteStateSkipFlagsRegression(t *testing.T) {
 	assert.NotEqual(t, versioningStatus, "Enabled")
 }
 
+// Regression test to ensure that accesslogging_bucket_name is taken into account
+// & sets the TargetLock bucket to a new S3 bucket, different from the origin S3 bucket
+func TestTerragruntSetsAccessLoggingForTfSTateS3BuckeToADifferentBucket(t *testing.T) {
+	t.Parallel()
+	t.Skip("Skipping this test untul PR Terratest-689 has been merged in")
```
brikis98, line 602: Looks like that Terratest PR is still in draft state. Is that intentional? Also, you might want to put a TODO above this to make reduce the odds of us forgetting about this!

brikis98 (commented):
Very close! Just some minor comments.

remote/remote_state_s3.go
ina-stoyanova: Good spot, Jim! I meant to comment on this, and confirm this is a duplicated piece of logic. Apologies for forgetting to do so earlier. 

I've removed it thinking it was a duplicate. See commit https://github.com/gruntwork-io/terragrunt/pull/1400/commits/e77cde2ecef9eb2f576c6c6e825d7d10edafb157

docs/_docs/04_reference/config-blocks-and-attributes.md
ina-stoyanova: So any invalid input for an S3 bucket name will produce an error from the S3 client: `InvalidBucketName: The specified bucket is not valid.`.

I've decided to leave the case where "empty" because, I didn't know how to differentiate between `""` and `nil` for strings - they seem to be treated the same coming from the config. 

One option is as you've suggested to treat these two cases separately (_maybe with checking the length of the string passed_), and/or update the logs - and be specific why the logging was disabled.

test/integration_test.go
ina-stoyanova: The terratest PR was failing the tests, so wanted to see why they were failing. Will wrap it up as well, once I'm happy the tests are going through

remote/remote_state_s3.go
brikis98: Why do you think it's duplicated?

ina-stoyanova: _Actually, after re-reading it for the 5th time, I think I was massively misreading what the flow of the logic is 🤦 ._

_What I was thinking:_ It's already handling the error & waiting for the bucket to exist.= in the guard block (line 395-401)

_What's I think is actually happening. P;ease correct me if I'm wrong in this :) _ It's only surpressing the 'OwnedByYouError' so it doesn't exit prematurely & waits instead. But if there's no error thrownm we still want to make sure the bucket has been created fine - hence the second `WaitUntilS3...`

brikis98: Yea, the check on 395-401 is an extra check for concurrent bucket creation: e.g., if you run `terragrunt apply-all`, it will execute on a bunch of modules in parallel, all of which might try to create the exact same bucket at the same time.

The check `WaitUntilS3BucketExists` I highlighted in this comment ensures that _after_ the bucket has been created, we wait long enough that we can "see" it in the AWS APIs. This is to work around the fact that AWS APIs are async and eventually consistent. Without this check, the code would move to the following parts, which try to modify the S3 bucket, and potentially hit errors because the bucket information hasn't propagated yet.

ina-stoyanova: you're absolutely right - thanks for spotting & explaining 🎈 
*Reverting this commit now*

ina-stoyanova:
Some new commits here. Another review would be very welcome 👀 🙂

brikis98 (approved):
LGTM! 

Note: I'm not sure if you're waiting to merge the Terratest thing first or merging this and coming back to update the Terratest version later?

ina-stoyanova:
Thank you @brikis98 🎉 

I think the reasoning we agreed on with @yorinasub17 was to not necessarily wait on the terratest, as that might be a yak shave, as it's turning out. I'll release this first thing on Monday, and then keep as a priority to close the terratest PR too.

PR #1405, opened 2020-10-29
Rename the gostructToCty function in config_as_cty.go to goTypeToCty

### LINKED ISSUE #1384: Rename the gostructToCty function in config_as_cty.go
As per [this comment](https://github.com/gruntwork-io/terragrunt/pull/1232/files#r444288437) the [`gostructToCty`](https://github.com/gruntwork-io/terragrunt/blob/master/config/config_as_cty.go#L184) function has a misleading name, because it works not only for structs but also for other data types, such as arrays. It should be renamed to better reflect its true use.

VishalCR7:
Fixes #1384 

Used [gocty doc](https://github.com/zclconf/go-cty/blob/main/docs/gocty.md#implied-cty-type-of-a-go-value) to gain some context regarding implied type. Updated the comment to reflect the same.

### DIFF
--- a/config/config_as_cty.go
+++ b/config/config_as_cty.go
@@ -43,7 +43,7 @@ func terragruntConfigAsCty(config *TerragruntConfig) (cty.Value, error) {
 		output["remote_state"] = remoteStateCty
 	}
 
-	dependenciesCty, err := gostructToCty(config.Dependencies)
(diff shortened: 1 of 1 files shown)


brikis98 (approved):
Thanks! Code changes LGTM. I'll kick off tests now.

brikis98:

PR #1411, opened 2020-11-02
fix image path for quick-start

ina-stoyanova:
Change from path `docs` to `documentation` to be able to find the `promote-immutable-Terraform-code-across-envs.png` image asset


brikis98 (approved):
Thx for the fix!

ina-stoyanova:
Thanks for reviewing! Merging now :)

Issue #1253, opened 2020-07-15
-all commands are finding terragrunt.hcl files within child dir .terragrunt-cache directories


tgourley01:
I've seen this behavior several times and I can gather more info but first wanted to ask -
(windows 10, terragrunt version v0.23.30)

If I have a nested directory structure of terragrunt modules and try and apply-all from the top level, and forever reason it fails.
If I cd into the failing module and run terragrunt from there I get a .terragrunt-cache dir in that child dir.
after resolving the issue I should be able to cd back to the top and apply-all again to deploy any remaining modules which depended on the failed one.

What I see is that apply-all shows me all the expected modules AND one it found in the clild .terragrunt-cache dir created above.

Am I alone in seeing this issue? I have encountered this quite a few times.

This affects all -all commands, here is a snip from destroy-all (in D:/live/live/project/us-west-1/xxx/demo) showing a module being found in a cache dir -

  => Module D:/live/live/project/us-west-1/xxx/demo/infra/network/peering (excluded: false, dependencies: [D:/live/live/project/us-west-1/xxx/demo/infra/network/vpc_compute, D:/live/live/project/us-west-1/xxx/demo/infra/network/vpc_docdb])
  => Module D:/live/live/project/us-west-1/xxx/demo/infra/network/vpc_compute (excluded: false, dependencies: [])
  => Module D:/live/live/project/us-west-1/xxx/demo/infra/network/**vpc_compute/.terragrunt-cache/9XMpZZzIpP7iixD598IFNEhphH4/Rvcr1Xqh4QpJ60TT9YEtkizwh6A** (excluded: false, dependencies: [])
  => Module D:/live/live/project/us-west-1/xxx/demo/infra/network/vpc_docdb (excluded: false, dependencies: [])

Anyway, before I gather a bunch of other info, is there something obvious I'm overlooking?

Thanks,
Tim

brikis98:
[...]

PR #1424, opened 2020-11-13
Skip the cache dir entirely when finding configs

awiddersheim:
Rather than searching each path that might exist below the
`.terragrunt-cache` directory, skip it entirely speeding up the process.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -277,6 +277,11 @@ func FindConfigFilesInPath(rootPath string, terragruntOptions *options.Terragrun
 			return err
 		}
 
+		// Skip the Terragrunt cache dir entirely
+		if info.IsDir() && info.Name() == options.TerragruntCacheDir {
+			return filepath.SkipDir
+		}
+
 		isTerragruntModule, err := containsTerragruntModule(path, info, terragruntOptions)
 		if err != nil {
 			return err


awiddersheim:
This is semi-related to https://github.com/gruntwork-io/terragrunt/issues/1253.

awiddersheim:
Tests https://gist.github.com/awiddersheim/989354ab66d4f11fd0ffb7983ee20aeb.

config/config.go
```diff
@@ -277,6 +277,11 @@ func FindConfigFilesInPath(rootPath string, terragruntOptions *options.Terragrun
 			return err
 		}
 
+		// Skip the Terragrunt cache dir entirely
+		if info.IsDir() && info.Name() == options.TerragruntCacheDir {
```
brikis98, line 281: Does `.Name()` return the folder name? Or the entire folder path? Because an equality check would only work for the former.

brikis98 (commented):
Thanks for the PR!

Were you able to find a simple test case that reproduces the issue off finding `terragrunt.hcl` files in `.terragrunt-cache` dirs? If we had a repro, we could add a test case, which presumably would initially fail, and check that this PR fixes that failure.

config/config.go
awiddersheim: Just the folder name.

awiddersheim:
@brikis98 There is actually already some logic and tests to skip the `.terragrunt-cache` directories.

https://github.com/gruntwork-io/terragrunt/blob/master/config/config.go#L303-L306

The problem is that `*-all` commands still traverse every file in cache directories looking for configs and be quite slow instead of just skipping the directories entirely and not looking at everything beneath them. I thought about restructuring things a bit to remove the redundancy I'm adding but I'm not familiar with the code base enough to do that reshuffling confidently so open to suggestions.

awiddersheim:
Yes, that is correct. It is certainly noticeable and measurable with my setup where I'm running on a laptop and have a lot of files in those cache directories. Obviously, it will vary based on any number of factors when it comes to people's setups.

brikis98 (approved):
OK, kicking off tests now.

brikis98:
Oh, wait, could you pull in  the latest from `master` first?

awiddersheim:
@brikis98 Yup! Rebased.

brikis98:
OK, running tests now.

brikis98 (approved):
Tests passed! Merging now.

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.26.4

awiddersheim:
Thanks!

PR #1426, opened 2020-11-13
update for least priv minimal iam access to create an s3 backend

loreleimccollum-work:
The docs for IAM access required for creating the S3 backend were missing a few permissions
Updated the minimal policy to include what was needed to successfully create the backend


brikis98 (approved):
LGTM, thank you!

PR #1428, opened 2020-11-13
Fix Terraform version check with CLI env vars

awiddersheim:
Fix passing command line arguments to Terraform using the `TF_CLI_ARGS`
environment variables[1]. Need to exclude those variables from the
version check.

[1]: https://www.terraform.io/docs/commands/environment-variables.html#tf_cli_args-and-tf_cli_args_name

### DIFF
--- a/cli/version_check.go
+++ b/cli/version_check.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"regexp"
+	"strings"
 
 	"github.com/gruntwork-io/terragrunt/errors"
 	"github.com/gruntwork-io/terragrunt/options"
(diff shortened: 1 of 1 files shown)


cli/version_check.go
```diff
@@ -25,6 +26,13 @@ func PopulateTerraformVersion(terragruntOptions *options.TerragruntOptions) erro
 	terragruntOptionsCopy.Writer = ioutil.Discard
 	terragruntOptionsCopy.ErrWriter = ioutil.Discard
 
+	// Remove any TF_CLI_ARGS before version checking
```
brikis98, line 29: NIT: could you expand this comment to explain why we're removing these (i.e., these are args to be appended to the "real" command and not the `--version` check) and add a link to the `TF_CLI_ARGS` docs?

brikis98 (commented):
Thanks for the PR! This looks like a good fix. Just one suggestion and then I can kick off tests.

cli/version_check.go
awiddersheim: Sure! Updated.

brikis98:
Great, thank you! One more thing, could you pull in the latest from `master`? I can kick off tests after that.

awiddersheim:
@brikis98 Sure! Rebased.

brikis98:

awiddersheim:
@brikis98 Looks like I didn't `fmt` my last commits. I updated. Can you kick them off again?

brikis98:
Yup!

brikis98 (approved):
Tests passed! Merging now.

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.26.3

awiddersheim:
Thanks!

PR #1429, opened 2020-11-13
fix: Set Terraform plan file as last argument in apply-all

### LINKED ISSUE #1271: terragrunt apply-all specific plan fails

```sh
# Generated command

...
```

amontalban:
This PR should fix #1271 (The position of the Terraform plan file in the `apply-all` command).

I'm having the same problem as @ibacalu when running Terragrunt with:

```
  extra_arguments "save_plan" {
    commands = [
      "plan"
    ]

    arguments = [
      "-out=${get_terragrunt_dir()}/${trimspace(run_cmd("basename", "${get_terragrunt_dir()}"))}.tfplan"
    ]
  }

  extra_arguments "load_plan" {
    commands = [
      "apply"
    ]

    arguments = [
      "${get_terragrunt_dir()}/${trimspace(run_cmd("basename", "${get_terragrunt_dir()}"))}.tfplan"
    ]
  }
```

And then running `plan-all` and `apply-all` (I am aware that this is an antipattern as described by @yorinasub17 but I think it needs to be fixed anyway).

Thanks!

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of CLI arguments so that any file argument is now always placed at the end of the command, ensuring correct argument order for Terraform commands.

### DIFF
--- a/options/options.go
+++ b/options/options.go
@@ -276,9 +276,27 @@ func (terragruntOptions *TerragruntOptions) InsertTerraformCliArgs(argsToInsert 
 	// Options must be inserted after command but before the other args
 	// command is either 1 word or 2 words
 	var args []string
+	var lastArg string
+
 	args = append(args, terragruntOptions.TerraformCliArgs[:commandLength]...)
-	args = append(args, argsToInsert...)
+	// We need to iterate over the argsToInsert to find an arg that is a file.
+	// If we found one, we treat it as the Terraform plan file and we then set it
+	// as lastArg to append it as last argument.
+	for _, arg := range argsToInsert {
+		if util.IsFile(arg) {
+			lastArg = arg
+		} else {
+			args = append(args, arg)
+		}
+	}
+
 	args = append(args, terragruntOptions.TerraformCliArgs[commandLength:]...)
+
+	// Append lastArg which is a Terraform plan file
+	if lastArg != "" {
+		args = append(args, lastArg)
+	}
+
 	terragruntOptions.TerraformCliArgs = args
 }
 


options/options.go
```diff
@@ -276,9 +276,27 @@ func (terragruntOptions *TerragruntOptions) InsertTerraformCliArgs(argsToInsert
 	// Options must be inserted after command but before the other args
 	// command is either 1 word or 2 words
 	var args []string
+	var lastArg string
+
 	args = append(args, terragruntOptions.TerraformCliArgs[:commandLength]...)
-	args = append(args, argsToInsert...)
+	// We need to iterate over the argsToInsert to find an arg that is a file.
+	// If we found one, we treat it as the Terraform plan file and we then set it
+	// as lastArg to append it as last argument.
```
brikis98, line 284: Hm, I'm not sure this is a safe assumption to make. That is, it's possible to have a different arg passed to Terraform that is a file—or worse yet, happens to match the name of a valid file—but isn't a plan file. Here's an example:

```bash
terraform apply \ 
  -state foo.tfstate \
  -var foo=bar
```

If there's a `foo.tfstate` already, the code above will move that parameter to the end of the command, as if it's a plan file.

brikis98 (commented):
Thanks for the PR! I suspect the current approach will not work correctly. To know for sure, this PR will need tests to validate the behavior. Check out `test/integration_test.go` and the fixtures it uses to see how to structure such a test.

options/options.go
amontalban: Thank you @brikis98, I haven't considered that use case. I will write a test as well as validating the current ones work fine and come back to you.

yhakbar:
This pull request has gone very stale. Sorry we never got to it. That's on us. Closing this PR, as I don't think it's needed or will get updated.

PR #1435, opened 2020-11-20
Fix CircleCI Contexts

zackproser:
This pull request was programmatically opened by the multi-repo-updater program. It should be adding the 'Gruntwork Admin' context to any Workflows -> Jobs nodes and should also be leaving the rest of the .circleci/config.yml file alone. 

 This PR was opened so that all our repositories' .circleci/config.yml files can be converted to use the same CircleCI context, which will make rotating secrets much easier in the future.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,9 +109,5 @@ workflows:
             branches:
               only: master
     jobs:
-      - unit_test:
-          context:
-            - Gruntwork Admin
-      - integration_test:
-          context:
-            - Gruntwork Admin
+      - unit_test
+      - integration_test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,6 @@
 defaults: &defaults
   docker:
     - image: 087285199408.dkr.ecr.us-east-1.amazonaws.com/circle-ci-test-image-base:tf13
-
 version: 2
 jobs:
   # We're running unit tests separately from integration tests - with no parallelization.
@@ -34,7 +33,6 @@ jobs:
           path: logs
       - store_test_results:
           path: logs
-
   integration_test:
     <<: *defaults
     steps:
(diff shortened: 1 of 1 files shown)


lorengordon:
> This pull request was programmatically opened by the multi-repo-updater program

Ok, intrigued... What is the multi-repo-updater program?

.circleci/config.yml
```diff
@@ -108,5 +109,9 @@ workflows:
...
               only: master
     jobs:
-      - unit_test
-      - integration_test
+      - unit_test:
+          context:
+            - Gruntwork Admin
+      - integration_test:
+          context:
+            - Gruntwork Admin
```
brikis98, line 117: Hm, can nightly jobs use contexts? I don't think they can. See https://gruntwork-io.slack.com/archives/CLYE0JYS0/p1606143137141900?thread_ts=1602848098.207000&cid=CLYE0JYS0 for discussion.

brikis98:
> > This pull request was programmatically opened by the multi-repo-updater program
> 
> Ok, intrigued... What is the multi-repo-updater program?

We wrote a script that can:

1. Check out all of our repos (or a specified subset).
1. Make some changes in that repo. Currently, we are using it to update CircleCi config files in all the repos, but we'll hopefully expand that to do other tasks later.
1. If there were changes, commit them to a branch, and open a PR.

If we can polish this code up a bit, we may release it more broadly.

lorengordon:
@brikis98 very interesting! we definitely have a similar use case. maintaining CI configs over hundreds of repos is a struggle!

> If we can polish this code up a bit, we may release it more broadly.

brikis98:
> @brikis98 very interesting! we definitely have a similar use case. maintaining CI configs over hundreds of repos is a struggle!

It is indeed a struggle! We'll announce when this tool is widely available. May take some time, as right now, it's in "internal prototype" status, but it helps to know others are interested!

.circleci/config.yml
zackproser, line 117:
```suggestion
      - unit_test
      - integration_test
```

zackproser: Thanks, caught up on that discussion + docs, and repaired this PR to exclude the nightly jobs from the context addition.

brikis98 (approved):
LGTM!

zackproser:
> LGTM!

Thanks for reviewing!

PR #1439, opened 2020-11-24
Fix nested mocks

yorinasub17:
### Problem

When retrieving dependencies, terragrunt would clone the `TerragruntOptions` object and replace the `TerraformCommand` to `output` so that we can run `terraform output` in the terragrunt context. This breaks the mock outputs feature because the mock outputs feature depended on knowing the original terraform command to decide when to return mock outputs. When you have deep dependencies (a dependency config that also has a dependency within), the nested dependency sees `output` as the terraform command, as opposed to the original one (e.g., `validate`). This makes configurations confusing because you need to add `output` to the `mock_outputs_allowed_terraform_commands`.

### Solution

We now track what the original terraform command. This will be the top level command that was run from the CLI. The PR works by only setting the `OriginalTerraformCommand` once at the beginning, and flowing that through everytime we clone the options.

### DIFF
--- a/options/options.go
+++ b/options/options.go
@@ -51,6 +51,8 @@ type TerragruntOptions struct {
 	// TerraformCommand to `output` to run `terraform output`, which loses the context of the original command that was
 	// run to fetch the dependency. This is a problem when mock_outputs is configured and we only allow mocks to be
 	// returned on specific commands.
+	// NOTE: For `xxx-all` commands, this will be set to the Terraform command, which would be `xxx`. For example,
+	// if you run `apply-all` (which is a terragrunt command), this variable will be set to `apply`.
 	OriginalTerraformCommand string
 
 	// Version of terraform (obtained by running 'terraform version')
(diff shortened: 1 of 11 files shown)


options/options.go
```diff
@@ -46,6 +46,13 @@ type TerragruntOptions struct {
 	// Current Terraform command being executed by Terragrunt
 	TerraformCommand string
 
+	// Original Terraform command being executed by Terragrunt. Used to track command evolution as terragrunt chains
```
brikis98, line 49: Will this be set to `apply-all` all the way through? Or switch to `apply` at some point?

options/options.go
```diff
@@ -46,6 +46,13 @@ type TerragruntOptions struct {
...
+	// different commands together. For example, when retrieving dependencies, terragrunt will change the
+	// TerraformCommand to `output` to run `terraform output`, which loses the context of the original command that was
+	// run to fetch the dependency. This is a problem when mock_outputs is configured and we only allow mocks to be
+	// returned on specific commands.
+	OriginalTerraformCommand string
```
brikis98, line 54: This value probably needs to be set in the `NewTerragruntOptions` method?

brikis98 (commented):
Clever solution!

options/options.go
yorinasub17: Good question! Added comment to clarify what happens for xxx-all commands.

options/options.go
yorinasub17: Updated! 956796e

yorinasub17:
Thanks for review! Addressed comments and this is ready for another round of review.

brikis98 (approved):
LGTM!

yorinasub17:
Thanks for review! Merging and releasing this now.

Issue #1366, opened 2020-09-22
 Only one level of includes is allowed


nedaair:
# 1. This is my project structure.

~~~ hcl
terragrunt-project 
  - enviroment
     - ec2
        - terragrunt.hcl
     - s3
        - terragrunt.hcl
     - terragrunt.hcl
  - module
     - ec2
        - main.tf
        - outputs.tf
        - variables.tf
~~~

# 2. This is root terragrunt.hcl
 
~~~hcl
generate "provider" {
  path      = "provider.tf"
  if_exists = "overwrite_terragrunt"
  contents  = <<EOF
provider "aws" {
  region = "ap-northeast-2"
}
EOF
}

remote_state {
  backend = "s3"
  config = {
    encrypt        = true
    bucket         = "terragrunt-test"
    key            = "${path_relative_to_include()}/terraform.tfstate"
    region         = "ap-northeast-2"
    dynamodb_table = "terragrunt-locks"
  }
  generate = {
    path      = "backend.tf"
    if_exists = "overwrite_terragrunt"
  }
}
~~~
  
# 3. This is terragrunt.hcl in ec2 folder

~~~ hcl
terraform {
  source = "../../module/ec2"
}

inputs = {
  subnet_id = "subnet-2d11c946"
}

# Include all settings from the root terragrunt.hcl file
include {
  path = find_in_parent_folders()
}
~~~

---

I called "terragrunt apply-all" under the environment folder.
Initially, it worked well and created an EC2 instance. However, the second execution results in the error as follows.

~~~
Error processing module at 'C:/Users/abc/PycharmProjects/terragrunt-study/enviroment/ec2/.terragrunt-cache/atTOD5LBQHAAgxKMQddh6R4j
XUE/mW46bbJv_XuxaHj-jku8hlgoUlQ/terragrunt.hcl'. How this module was found: Terragrunt config file found in a subdirectory of C:/Users/abc/PycharmProjects/terragrun
t-study/enviroment. Underlying error: C:/Users/abc/PycharmProjects/terragrunt-study/enviroment/ec2/.terragrunt-cache/atTOD5LBQHAAgxKMQddh6R4jXUE/mW46bbJv_XuxaHj-jku
8hlgoUlQ/terragrunt.hcl includes C:/Users/abc/PycharmProjects/terragrunt-study/enviroment/ec2/terragrunt.hcl, which itself includes C:/Users/abc/PycharmProj
[...]

PR #1440, opened 2020-11-24
fix: use normalized separator

xmclark:
# Description

in #1366, we see that terragrunt does not correctly run xxx-all commands on windows due to an unaccounted .terragrunt-cache directory. This is because terragrunt is confusing slash separators in file paths. 

[There is exactly one failing unit test when run on windows in the cli module](https://github.com/gruntwork-io/terragrunt/blob/master/cli/hclfmt_test.go#L57). Debugging the tests reveals that the library[ `zglob` automatically normalizes paths to "/" (unix slashes)](https://github.com/mattn/go-zglob/blob/master/zglob.go#L30). The test works as expected on linux and mac because it has the same os.PathSeparator. This fails on windows though because os.PathSeparator is "\". In the debugger, one can see that the list contains exactly 1 element (exactly one result that matches entire path).

image

# Change

This PR updates the call to `util.ListContainsElement` to the hard-coded "/" separator. This resolves the issue in #1366 for windows users. It can be verified by running tests in the `cli` module locally on a windows host.

### DIFF
--- a/cli/hclfmt.go
+++ b/cli/hclfmt.go
@@ -34,6 +34,7 @@ func runHCLFmt(terragruntOptions *options.TerragruntOptions) error {
 	}
 
 	terragruntOptions.Logger.Printf("Formatting terragrunt.hcl files from the directory tree %s.", terragruntOptions.WorkingDir)
+	// zglob normalizes paths to "/"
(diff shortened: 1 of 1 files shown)


yorinasub17 (commented):
Thanks for the PR! This makes sense to me. I'll kick off the build now.

yorinasub17 (approved):
Build passed so will go ahead to merge this in!

PR #1442, opened 2020-11-25
Delete nightly job

zackproser:
Per discussions, we've decided to stop using "nightly" or triggered jobs as they are incompatible with circleci contexts.


brikis98 (approved):
LGTM!

PR #1459, opened 2020-12-04
Update Terragrunt to work with Terraform 0.14

### LINKED ISSUE #1423: terragrunt always re-runs `terraform init` with terraform 0.14.0-rc1
It seems terragrunt fails to detect when initialization is not necessary when using terraform 0.14.0-rc1. It always runs `terraform init` no matter what, and it takes a lot of time especially when dependencies are involved.

I also tried `terragrunt apply --terragrunt-no-auto-init`, and here's what happens:

```
$ terragrunt apply --terragrunt-no-auto-init
...
[terragrunt] 2020/11/13 01:29:18 Cannot continue because init is needed, but Auto-Init is disabled.  You must run 'terragrunt init' manually.
[terragrunt] 2020/11/13 01:29:18 Unable to determine underlying exit code, so Terragrunt will exit with error code 1
```

### LINKED ISSUE #1454: Copy Provider Dependency Lock file to `teragrunt.hcl` directory
Terraform 0.14 adds support for [Provider Dependency Lock file](https://www.terraform.io/upgrade-guides/0-14.html#provider-dependency-lock-file).

This is a feature request for Terragrunt to work with this new lock file.

- If Terragrunt notices that the lock file does not exist, it should copy it from the `.terragrunt-cache` directory to the directory that Terragrunt was invoked from (i.e. the one with `terragrunt.hcl`).
- If the file already exists, there's nothing to do because Terragrunt will copy the file into `.terragrunt-cache` and Terraform will work with it.

infraredgirl:
1. Add support for copying lock files as necessary. Fixes #1454.
1. Fix issue where Terragrunt would re-init every time you ran it. Fixes #1423.

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -549,6 +549,7 @@ func runTerragruntWithConfig(terragruntOptions *options.TerragruntOptions, terra
 
 	return runActionWithHooks("terraform", terragruntOptions, terragruntConfig, func() error {
 		return runTerraformWithRetry(terragruntOptions)
+		//TODO: Make sure lock file gets copied back to where you ran terragrunt from
 	})
 }
 
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -558,6 +558,13 @@ func runTerragruntWithConfig(originalTerragruntOptions *options.TerragruntOption
 
 		var lockFileError error
 		if util.FirstArg(terragruntOptions.TerraformCliArgs) == CMD_INIT {
+			// Copy the lock file from the Terragrunt working dir (e.g., .terragrunt-cache/xxx/<some-module>) to the
+			// user's working dir (e.g., /live/stage/vpc). That way, the lock file will end up right next to the user's
+			// terragrunt.hcl and can be checked into version control. Note that in the past, Terragrunt allowed the
+			// user's working dir to be different than the directory where the terragrunt.hcl file lived, so just in
+			// case, we are using the user's working dir here, rather than just looking at the parent dir of the
+			// terragrunt.hcl. However, the default value for the user's working dir, set in options.go, IS just the
+			// parent dir of terragrunt.hcl, so these will likely always be the same.
 			lockFileError = copyLockFile(terragruntOptions.WorkingDir, originalTerragruntOptions.WorkingDir, terragruntOptions.Logger)
 		}
 
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"log"
 	"regexp"
 	"strings"
 	"time"
@@ -538,7 +539,7 @@ func runTerragruntWithConfig(originalTerragruntOptions *options.TerragruntOption
 			return err
 		}
 	} else {
-		if err := prepareNonInitCommand(terragruntOptions, terragruntConfig); err != nil {
+		if err := prepareNonInitCommand(originalTerragruntOptions, terragruntOptions, terragruntConfig); err != nil {
 			return err
 		}
 	}
@@ -557,7 +558,7 @@ func runTerragruntWithConfig(originalTerragruntOptions *options.TerragruntOption
 
 		var lockFileError error
 		if util.FirstArg(terragruntOptions.TerraformCliArgs) == CMD_INIT {
-			lockFileError = copyLockFile(terragruntOptions.WorkingDir, originalTerragruntOptions.WorkingDir)
+			lockFileError = copyLockFile(terragruntOptions.WorkingDir, originalTerragruntOptions.WorkingDir, terragruntOptions.Logger)
 		}
 
 		return errors.NewMultiError(runTerraformError, lockFileError)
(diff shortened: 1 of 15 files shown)


brikis98:
@infraredgirl We need to remember to look into https://github.com/gruntwork-io/terragrunt/issues/1423 also  as part of supporting 0.14!

brikis98:
Tests passed! This is now ready for review!

ina-stoyanova (dismissed):
Reviewed together in TicketJam. Looks good!

cli/cli_app.go
```diff
@@ -548,10 +554,31 @@ func runTerragruntWithConfig(terragruntOptions *options.TerragruntOptions, terra
 	}
 
 	return runActionWithHooks("terraform", terragruntOptions, terragruntConfig, func() error {
-		return runTerraformWithRetry(terragruntOptions)
+		runTerraformError := runTerraformWithRetry(terragruntOptions)
+
+		var lockFileError error
+		if util.FirstArg(terragruntOptions.TerraformCliArgs) == CMD_INIT {
+			lockFileError = copyLockFile(terragruntOptions.WorkingDir, originalTerragruntOptions.WorkingDir, terragruntOptions.Logger)
```
yorinasub17, line 561: This could use a comment because it took me a bit to follow what's going on, or improvement.

Am I correct in understanding that what we are trying to accomplish here is to get the lock file to be copied to where the terragrunt config is? And this works because of https://github.com/gruntwork-io/terragrunt/blob/master/options/options.go#L207 ?

If that's correct, I feel like this makes more sense if we use the directory where `terragruntOptions.TerragruntConfigPath` is. I think the direct implementation would make this easier to follow in terms of what we are trying to accomplish here, and we can avoid passing around the `originalTerragruntOptions`.

yorinasub17 (commented):
Had one question about something that looked confusing, but other than that LGTM!

cli/cli_app.go
brikis98: Hm, I could've sworn that the "working dir" could be different in the past from the dir where the `terragrunt.hcl` file was... E.g., You could run `cd /foo && terragrunt apply --terragrunt-config /bar/terragrunt.hcl` and  working dir used to be set to `/foo` even though `terragrunt.hcl` was in `/bar`... But looking at `terragrunt.Options`, at least, that doesn't seem to be the case. Am I just mis-remembering? Or did that change at some point?

yorinasub17: > But looking at terragrunt.Options, at least, that doesn't seem to be the case. Am I just mis-remembering? Or did that change at some point?

Honestly speaking, I am not sure. This is one of those areas that is extremely difficult to follow in the terragrunt codebase, especially when you take into account `dependency`, `xxx-all`, etc.

So all the more reason that we either:

- Need a comment here with the true intent of where we want the lock file to eventually be.
- Update the code, if possible, to be more clear/direct about where we place the lock file.

> You could run cd /foo && terragrunt apply --terragrunt-config /bar/terragrunt.hcl and working dir used to be set to /foo even though terragrunt.hcl was in /bar...

Even if that were the case, is that where we want the lock file? It seems like we want it where the terragrunt config is, right?

brikis98: > Even if that were the case, is that where we want the lock file (in `/foo`)? It seems like we want it where the terragrunt config is (in `/bar`), right? Since we only copy things out of where the `terragrunt.hcl` file is located?

You know, I'm not 100% sure. I could see arguments for either one. I also realize we need to document the behavior, not just in a comment as you said, but in the official docs. I won't have time to get to this today, so I'll have to give this more thought and come back and wrap it up later. Thx for the great questions.

brikis98: Alright, added a comment explaining this behavior in https://github.com/gruntwork-io/terragrunt/pull/1459/commits/31830b5aa13806267bea785d157290914fd105f7 and a whole new page in the documentation that describes how Terragrunt handles lock files in https://github.com/gruntwork-io/terragrunt/pull/1459/commits/c15eefbb1dc8eebf336c334e9d78f011c4798ff9.

cli/cli_app.go
```diff
@@ -514,7 +517,10 @@ func shouldRunHook(hook config.Hook, terragruntOptions *options.TerragruntOption
 
 // Runs terraform with the given options and CLI args.
 // This will forward all the args and extra_arguments directly to Terraform.
-func runTerragruntWithConfig(terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig, allowSourceDownload bool) error {
+
+// This function takes in the "original" terragrunt options has the unmodified 'WorkingDir' from before downloading the code from the source URL,
```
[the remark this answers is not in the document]
brikis98: Good catch. Fixed in https://github.com/gruntwork-io/terragrunt/pull/1459/commits/314257f45a1f37929043296233fc396759f1061a.

marinalimeira (approved):

brikis98:
Thank you for the review! Merging now.

brikis98:
Released in: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.27.0

PR #1494, opened 2021-01-18
Add TF14 to compatibility table

brikis98:
We updated Terragrunt to work with Terraform 0.14 in https://github.com/gruntwork-io/terragrunt/pull/1459, but forgot to update the version compatibility table.


ina-stoyanova (approved):
Well spotted!

brikis98:

Issue #1485, opened 2021-01-08
get_aws_account_id Fails For Localstack


jonwinton:
Hi!

I'm using [Localstack](https://github.com/localstack/localstack) for local development testing and I'm running into issues where helper function calls to AWS fail at initialization time. Here's an example of the error:

```
Error in function call; Call to function "get_aws_account_id" failed: NoCredentialProviders: no valid providers in chain. Deprecated.
        For verbose messaging see aws.Config.CredentialsChainVerboseErrors.
```

If I remove these calls and hardcode the Localstack account id everything works properly and I can grab the account id from a Terraform `data` source (`data "aws_caller_identity" "current" {}`).

It looks like the built-in functions (like `get_aws_account_id`) are all evaluated using local credentials before any Terraform is evaluated, which means we can't set [a custom endpoint](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html#custom-endpoint) for for the session being used to grab values from AWS. [Here's the call to get the current aws account id](https://github.com/gruntwork-io/terragrunt/blob/master/aws_helper/config.go#L121).

I can't find any environment valiables to configure which endpoint to use for the AWS SDK, so it seems like this needs to be defined as a Terragrunt specific env var/runtime flag  that can be passed to the function. I'm happy to try and get a PR together, but I wanted to check that this would be good to implement.

Let me know if I missed something, and thanks in advance!

brikis98:
You are correct! Terragrunt can't parse the Terraform code to find custom endpoints you've defined, as it may need to process something like `get_aws_account_id` _before_ it knows what Terraform code to even parse. So, the only option is to expose the ability to override SDK endpoints directly in Terragrunt itself. I don't see any way within the AWS Go SDK to override endpoints globally, so we'll have to do it when each config/client is being initialized. PR to support this is welcome.

jonwinton:
@brikis98 thanks for confirming, I'll try and get a PR together soon 👍

PR #1497, opened 2021-01-18
Support Custom STS Endpoint For Built-In Functions

jonwinton:
Aiming to fix https://github.com/gruntwork-io/terragrunt/issues/1485

Terragrunt can't pull a custom STS endpoint from Terraform configuration because the built-in helper functions (like `get_aws_account_id`) run before the Terraform code is parsed. Since the AWS Go SDK doesn't provide a way to override endpoints with any environment variables, we have to define a custom endpoint when creating a session.

This PR adds a config flag (`--terragrunt-sts-endpoint`) that is passed to the call to STS for the helper functions. I wasn't quite sure what to name the flag, so I went with something similar to `--terragrunt-iam-role`. Happy to update/change anything as necessary.

Thanks!

### DIFF
--- a/aws_helper/config.go
+++ b/aws_helper/config.go
@@ -30,24 +30,7 @@ type AwsSessionConfig struct {
 // Returns an AWS session object for the given config region (required), profile name (optional), and IAM role to assume
 // (optional), ensuring that the credentials are available
 func CreateAwsSession(config *AwsSessionConfig, terragruntOptions *options.TerragruntOptions) (*session.Session, error) {
-	defaultResolver := endpoints.DefaultResolver()
-	customResolverFn := func(service, region string, optFns ...func(*endpoints.Options)) (endpoints.ResolvedEndpoint, error) {
-		if service == "s3" && config.CustomS3Endpoint != "" {
-			return endpoints.ResolvedEndpoint{
-				URL:           config.CustomS3Endpoint,
-				SigningRegion: config.Region,
-			}, nil
-		}
-
-		if service == "sts" && config.CustomStsEndpoint != "" {
-			return endpoints.ResolvedEndpoint{
-				URL:           config.CustomStsEndpoint,
-				SigningRegion: region,
-			}, nil
-		}
-
-		return defaultResolver.EndpointFor(service, region, optFns...)
-	}
+	customResolverFn := createCustomResolver(config)
 
 	var awsConfig = aws.Config{
 		Region:                  aws.String(config.Region),
@@ -96,6 +79,30 @@ func CreateAwsSession(config *AwsSessionConfig, terragruntOptions *options.Terra
 	}
 
 	return sess, nil
+}
+
+// Create a custom endpoint resolver; useful for connecting to custom endpoints for AWS services
+func createCustomResolver(config *AwsSessionConfig) func(string, string, ...func(*endpoints.Options)) (endpoints.ResolvedEndpoint, error) {
+	defaultResolver := endpoints.DefaultResolver()
+	customResolverFn := func(service, region string, optFns ...func(*endpoints.Options)) (endpoints.ResolvedEndpoint, error) {
+		if service == "s3" && config.CustomS3Endpoint != "" {
+			return endpoints.ResolvedEndpoint{
+				URL:           config.CustomS3Endpoint,
+				SigningRegion: config.Region,
+			}, nil
+		}
+
+		if service == "sts" && config.CustomStsEndpoint != "" {
+			return endpoints.ResolvedEndpoint{
+				URL:           config.CustomStsEndpoint,
+				SigningRegion: region,
+			}, nil
+		}
+
+		return defaultResolver.EndpointFor(service, region, optFns...)
+	}
+
+	return customResolverFn
 }
 
 // Make API calls to AWS to assume the IAM role specified and return the temporary AWS credentials to use that role
--- a/aws_helper/config.go
+++ b/aws_helper/config.go
@@ -17,6 +17,7 @@ import (
 type AwsSessionConfig struct {
 	Region                  string
 	CustomS3Endpoint        string
+	CustomStsEndpoint       string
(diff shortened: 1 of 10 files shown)


gflarity:
Can we get this reviewed? We'd love to use this and don't want to maintain a fork.

brikis98 (commented):
Thanks for the PR! The overall logic LGTM; I only had a couple minor NITs to look at.

Is it possible to add a test for this logic?

jonwinton:
Sorry for the delay, I'll try and address everything in the next day or two. Thanks for the review!

jonwinton:
@brikis98 I realized that a majority of this logic was being replicated in the `CreateAwsSession` method in that same file, so I moved the session creation to leverage what was already there. Here's the commit with the change: https://github.com/gruntwork-io/terragrunt/pull/1497/commits/23d1290ad728bdb19e9f15564a8c3508c4e67350

I'm wondering if you think this is an ok approach? I was thinking we could keep all the AWS session logic confined to a single place so there's less to reason about. 

Regarding the questions you asked before:

> Should this entire block only happen if terragruntOptions.StsEndpoint != ""?
Yep! We want the SDK to use the default endpoint for STS unless we're explicitly opting into a new endpoint. 

> This logic should probably be extracted into a NewSession method.
Using `CreateAwsSession` hopefully resolves this!

> Is it possible to add a test for this logic?
I'll look at adding testing for all of this if you think the approach is sound!

brikis98 (commented):
Ah, yes, this seems like the right approach! Of course, I _could_ be wrong, so a test to validate this is working would be very helpful 👍

aws_helper/config.go
```diff
@@ -119,7 +127,9 @@ func AssumeIamRole(iamRoleArn string) (*sts.Credentials, error) {
 
 // Return the AWS caller identity associated with the current set of credentials
 func GetAWSCallerIdentity(terragruntOptions *options.TerragruntOptions) (sts.GetCallerIdentityOutput, error) {
-	sess, err := session.NewSession()
+	sess, err := CreateAwsSession(&AwsSessionConfig{
+		CustomStsEndpoint: terragruntOptions.StsEndpoint,
```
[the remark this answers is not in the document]
jonwinton: Nope! This session is only being created to call STS (GetCallerIdentity function), so we only need to pass in the endpoint for STS

jonwinton:
@brikis98 sorry for the delay, but I think I've it to a point for review. I ran into some issues trying to test the entire `CreateAwsSession` function because it requires credentials for AWS, so I moved the custom resolver into its own function so we could test that logic alone ([see this commit](https://github.com/gruntwork-io/terragrunt/pull/1497/commits/5b5edf626e8cad44bfd7f04a03e46c6d70f147fd))

Hopefully this approach works for you, but happy to incorporate any feedback. Thanks!

jonwinton:
@brikis98 bumping this! We're trying to get a local Terragrunt workflow going and are hoping to see if this can get merged soon! Appreciate all the help!

aws_helper/config_test.go
```diff
@@ -0,0 +1,86 @@
...
+			defaultS3Endpoint := fmt.Sprintf("https://s3.%s.amazonaws.com", testCase.region)
+			// STS does not have region-specific endpoints by default
+			// https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html
+			defaultStsEndpoint := "https://sts.amazonaws.com"
+
+			// Grab the endpoint for S3 that our custom resolver is returning
+			resolvedS3Endpoint, err := resolver("s3", testCase.region)
+			if err != nil {
+				t.Fatalf("Failed to resolve endpoint for S3: %v", err)
+			}
```
brikis98, line 61: This will produce better log output in case of error:

```suggestion
			require.NoError(t, err)
```

aws_helper/config_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+			// Grab the endpoint for STS that our custom resolver is returning
+			resolvedStsEndpoint, err := resolver("sts", testCase.region)
+			if err != nil {
+				t.Fatalf("Failed to resolve endpoint for STS: %v", err)
+			}
```
brikis98, line 67:
```suggestion
			require.NoError(t, err)
```

cli/cli_app.go
```diff
@@ -214,6 +216,7 @@ GLOBAL OPTIONS:
    terragrunt-source                            Download Terraform configurations from the specified source into a temporary folder, and run Terraform in that temporary folder.
    terragrunt-source-update                     Delete the contents of the temporary folder to clear out any old, cached source code before downloading new source code into it.
    terragrunt-iam-role                          Assume the specified IAM role before executing Terraform. Can also be set via the TERRAGRUNT_IAM_ROLE environment variable.
+   terragrunt-sts-endpoint                  	Use a custom endpoint for Terragrunt calls to STS for built-in helper functions. Can also be set by TERRAGRUNT_STS_ENDPOINT environment variable.
```
brikis98, line 219: NIT: indentation looks off on the description?

brikis98 (commented):
Sorry for the delay. Super buried these days. Just a few NITs, and I can kick off tests!

jonwinton:
@brikis98 no worries at all, your time on this is appreciated!

The changes in you called out are made, so hopefully it's good to go!

lopezm1:
Bump, @jonwinton and @krisdevopsbot appreciate your initial work on this.

Running into this myself, was wondering if either of you found an alternative solution or could we maybe pick this back up? I could help if needed.

lopezm1:
@jonwinton Going to spend this week reviving your PR.

Will re-create the PR and assume this one is dead. Really appreciate your initial work on it.

jonwinton:
@lopezm1 just saw these comments, but sounds good!

kharann:
@lopezm1 did you manage to proceed with this? running into this myself :)

Issue #1502, opened 2021-01-21
--terragrunt-debug dosen't work since v0.27.0


dmitriy-kharchenko:
Hello folks!

In terragrunt 0.27.1 doesn't work --terragrunt-debug option. Files terragrunt-debug.* aren't created.

dmitriy-kharchenko:
And with `terragrunt apply` doesn't work too.

yorinasub17:
Ahhh yes I confirmed it is a regression. Will look into this today.

yorinasub17:
Fixed in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.27.2 (binaries should show up shortly)

PR #1503, opened 2021-01-21
Fix debug mode

yorinasub17:
Fixes https://github.com/gruntwork-io/terragrunt/issues/1502, where the `Debug` flag wasn't being propagated across the `Clone` call.


yorinasub17:
Thanks for review! Merging this now.

Issue #1437, opened 2020-11-24
Add option to specify target prefix when setting access logging on terraform state bucket


dsecik:
In addition to the `accesslogging_bucket_name` configuration option introduced in #1400 allowing to specify target bucket for s3 access logs, it would be handy to have an ability to specify also the target prefix to override the default `targetPrefix := "TFStateLogs/"` This will add more flexibility to access logging configuration, for example to be able to send access logs from multiple terraform state buckets to single logging bucket and not mix them together.

brikis98:
Seems worthwhile. PR to add this is welcome!

ina-stoyanova:
I'll pick this up this week!

ina-stoyanova:
Closing this issue. The feature has been implemented.

PR #1507, opened 2021-01-22
Add TargetPrefix as config input to access bucket logging

ina-stoyanova:
This PR is in response to [this issue. ](https://github.com/gruntwork-io/terragrunt/issues/1437)

Here's the TODOs:
- [x] add new variable for the `TFStateLogs`
- [x] update code when creating a new logs bucket for the TF state bucket logs
- [x] update tests
- [x] test
- [x] update official/any docs


ina-stoyanova:
I've tested this today - but came across this error. I can't quite figure out what I need to do to overcome it, but am sure I've seen it before. Any help would be appreciated!

Screenshot 2021-02-16 at 17 39 21

Otherwise, the tests pass, however, if this manual test doesn't - I'm even more confused where's the issue.

remote/remote_state_s3.go
```diff
@@ -443,12 +445,16 @@ func CreateS3BucketWithVersioningSSEncryptionAndAccessLogging(s3Client *s3.S3, c
 	if config.AccessLoggingBucketName != "" {
 		terragruntOptions.Logger.Printf("Enabling bucket-wide Access Logging on AWS S3 bucket %s - using as TargetBucket %s", config.remoteStateConfigS3.Bucket, config.AccessLoggingBucketName)
 
+		if config.AccessLoggingTargetPrefix != "" {
+			terragruntOptions.Logger.Printf("Setting Target Prefix for Access Logging to %s", config.AccessLoggingTargetPrefix)
```
yorinasub17, line 449: Use `Debugf` given conversation from https://github.com/gruntwork-io/terragrunt/pull/1541

remote/remote_state_s3.go
```diff
@@ -674,20 +680,19 @@ func EnableSSEForS3BucketWide(s3Client *s3.S3, config *RemoteStateConfigS3, terr
...
 // Enable bucket-wide Access Logging for the AWS S3 bucket specified in the given config
-func EnableAccessLoggingForS3BucketWide(s3Client *s3.S3, config *RemoteStateConfigS3, terragruntOptions *options.TerragruntOptions, logsBucket string) error {
+func EnableAccessLoggingForS3BucketWide(s3Client *s3.S3, config *RemoteStateConfigS3, terragruntOptions *options.TerragruntOptions, logsBucket string, logsBucketPrefix string) error {
 	if err := configureBucketAccessLoggingAcl(s3Client, aws.String(logsBucket), terragruntOptions); err != nil {
 		return errors.WithStackTrace(err)
 	}
 
-	targetPrefix := "TFStateLogs/"
-	terragruntOptions.Logger.Printf("Putting bucket logging on S3 bucket %s with TargetBucket %s and TargetPrefix %s", config.Bucket, logsBucket, targetPrefix)
+	terragruntOptions.Logger.Printf("Putting bucket logging on S3 bucket %s with TargetBucket %s and TargetPrefix %s", config.Bucket, logsBucket, logsBucketPrefix)
```
yorinasub17, line 688: Use `Debugf` for same reason as above.

yorinasub17 (commented):
Implementation LGTM. For the test failure, my guess is that you ran it once before you updated `terragruntOnlyConfigs`. In that case, terragrunt will generate the `backend.tf` file with the new setting included, and then won't regenerate because it exists. Does it work if you remove that file to force regeneration?

ina-stoyanova:
> Implementation LGTM. For the test failure, my guess is that you ran it once before you updated terragruntOnlyConfigs. In that case, terragrunt will generate the backend.tf file with the new setting included, and then won't regenerate because it exists. Does it work if you remove that file to force regeneration?

Unfortunately, still no luck.

ina-stoyanova:
This unfortunately has slipped my radar the last few weeks! I am looking into closing it now.

ina-stoyanova:
I've tested this and I believe it's ready to go!

remote/remote_state_s3.go
```diff
@@ -443,12 +445,16 @@ func CreateS3BucketWithVersioningSSEncryptionAndAccessLogging(s3Client *s3.S3, c
 	if config.AccessLoggingBucketName != "" {
 		terragruntOptions.Logger.Debugf("Enabling bucket-wide Access Logging on AWS S3 bucket %s - using as TargetBucket %s", config.remoteStateConfigS3.Bucket, config.AccessLoggingBucketName)
 
+		if config.AccessLoggingTargetPrefix != "" {
+			terragruntOptions.Logger.Debugf("Setting Target Prefix for Access Logging to %s", config.AccessLoggingTargetPrefix)
+		}
```
brikis98, line 450: NIT: Do we need this log statement here? There's already one in `EnableAccessLoggingForS3BucketWide`.

remote/remote_state_s3.go
```diff
@@ -42,6 +42,7 @@ type ExtendedRemoteStateConfigS3 struct {
 	EnableLockTableSSEncryption bool              `mapstructure:"enable_lock_table_ssencryption"`
 	DisableAWSClientChecksums   bool              `mapstructure:"disable_aws_client_checksums"`
 	AccessLoggingBucketName     string            `mapstructure:"accesslogging_bucket_name"`
+	AccessLoggingTargetPrefix   string            `mapstructure:"accesslogging_target_prefix"`
```
brikis98, line 45: I believe the original default was `TFStateLogs/`. Is the default now empty string?

remote/remote_state_s3.go
ina-stoyanova: It's definitely not essential - happy to remove.

remote/remote_state_s3.go
ina-stoyanova: Yep, you're right that it had a default value - I've now decided to leave it explicitly to the user's control - so it is a bt more intuitive - if there's no value set, then they don't get a prefix set - and I've considered this (so far) as a potential decision for the user.

_I'd be grateful to see your take on this 🙂  Do you think it's better if we have the default value?_

remote/remote_state_s3.go
brikis98: Since we're already logging it, prob no need to duplicate the logic here.

remote/remote_state_s3.go
brikis98: It's a trade-off, so it's a judgment call. Keeping the prefix would be backwards compatible, so all future buckets are created the way all previous ones were, which I would guess is a bit more preferable to not having a default at all in the future...

remote/remote_state_s3.go
```diff
@@ -674,20 +676,23 @@ func EnableSSEForS3BucketWide(s3Client *s3.S3, config *RemoteStateConfigS3, terr
...
+func EnableAccessLoggingForS3BucketWide(s3Client *s3.S3, config *RemoteStateConfigS3, terragruntOptions *options.TerragruntOptions, logsBucket string, logsBucketPrefix string) error {
 	if err := configureBucketAccessLoggingAcl(s3Client, aws.String(logsBucket), terragruntOptions); err != nil {
 		return errors.WithStackTrace(err)
 	}
 
-	targetPrefix := "TFStateLogs/"
-	terragruntOptions.Logger.Debugf("Putting bucket logging on S3 bucket %s with TargetBucket %s and TargetPrefix %s", config.Bucket, logsBucket, targetPrefix)
+	if logsBucketPrefix == "" {
+		logsBucketPrefix = "TFStatelogs/"
+	}
```
brikis98, line 686: OK, a few sanity checks here:

1. Does implementing it this way mean you can't set the value to an empty string? If so, we may need to set this default at parse time, when we can tell the difference between "user set the value explicitly to an empty string" or "user didn't set this value at all."
1. NIT: Don't you have a constant declared for this value already?
1. NIT: extra whitespace before the closing curly brace?

ina-stoyanova: Hey Jim - you're absolutely right! I went and chose the most straightforward way to implement it, but all of the points you've made are valid. 
1. I actually misunderstood your first comment, in this case, apologies! As before, we weren't giving the user any choice, I assumed that's what you meant by achieving backwards compatibility! _So, totally agree - let's put it at parse time of the config - I think the right place is `parseExtendedS3Config`_
2. Wasn't one just yet - but it's worth it at this time!
3. Hm, my IDE doesn't show it and `go fmt` ran fine. I'll make sure I double-check again before pushing!

remote/remote_state_s3.go
```diff
@@ -21,7 +21,8 @@ import (
 )
 
 const (
-	lockTableDeprecationMessage = "Remote state configuration 'lock_table' attribute is deprecated; use 'dynamodb_table' instead."
+	lockTableDeprecationMessage       = "Remote state configuration 'lock_table' attribute is deprecated; use 'dynamodb_table' instead."
+	s3BucketAccessLoggingTargetPrefix = "TFStatelogs"
```
brikis98, line 25: Is a trailing slash needed here?

brikis98: Also, NIT, but consider naming the var `defaultS3BucketAccessLoggingTargetPrefix`

remote/remote_state_s3.go
```diff
@@ -313,6 +316,10 @@ func parseExtendedS3Config(config map[string]interface{}) (*ExtendedRemoteStateC
 		return nil, errors.WithStackTrace(err)
 	}
 
+	if config["accesslogging_target_prefix"] == "" {
```
brikis98, line 319: Is empty string here the same as not being set at all?

Or do you actually want:

```go
prefix, prefixIsSet := config["accesslogging_target_prefix"]
if !prefixIsSet {
  // If no prefix is set, we set a default one
  config["accesslogging_target_prefix"] = s3BucketAccessLoggingTargetPrefix
}
```

ina-stoyanova: ah shit - thought I marked this as WIP, as I was testing the different cases - so this is still WIP - very much thanks for the pointers, Jim!

ina-stoyanova:
Cool @brikis98 @robmorgan I've now followed your guidance on how to check if it has been provided. In reality, checking for `xxx == ""` was also behaving the same at the parsing level, as it would render the map element as null instead of empty string, but from a reader's point - how it is now, seems to be better! (Let me know if you do think otherwise).

Also, I've added an extra test to check specifically the case with using the default `TargetPrefix` value. The test cases covered now are:
 - when a value is provided (the test expects a value of `logs/`, but it is the same scenario if it was an empty string)
 - when the attribute is not provided at all, that it sets TargetPrefix to the default value
 
 I've also updated the docs to reflect this. 
 
 More feedback welcome! 🙂

brikis98 (approved):
LGTM!

ina-stoyanova:
Thanks for the reviews, merging now! :)

Issue #432, opened 2018-03-07
Option for less verbose output


andrewhertog:
Is there an option that will hide a lot of the extra information that terragrunt outputs? While this information is nice to see, it would be nice to have a clean output that would just show us the raw output from terraform.

brikis98:
Not currently, but there should be. I'd welcome a PR to move Terragrunt to a proper logging system with INFO/DEBUG/etc and a way to set the log level via CLI. We've been meaning to move it over to [gruntwork-cli](https://github.com/gruntwork-io/gruntwork-cli) as the base, which has that sort of logging built-in, but haven't had time. A PR for that is very welcome.

In the meantime, a workaround is to note that Terragrunt logs everything to `stderr`, so you could do `terragrunt apply 2>/dev/null` to suppress it.

ezh:
solution for bash only:
```sh
terragrunt apply 2> >(grep -v "\[terragrunt]" >&2)
```
STDOUT and STDERR lines are preserved except [terragrunt] ones

josegonzalez:
One side-effect of this is that `terragrunt output ip` has a bunch of intermediate output, so you necessarily need to redirect stderr to ensure you're only getting what you want.

brikis98:
Terragrunt logs everything to stderr, so `terragrunt out <var>` should work just fine. If not, please open a separate bug with a way to repro.

josegonzalez:
Sorry, I meant that the regex above will not catch all lines from terragrunt - specifically, if you have an un-initialized module, some lines will be missing the prefix. Apologies!

I'd assume that the terraform call _also_ outputs to stderr in some instances, so redirecting all stderr wouldn't be a great solution in those cases.

brikis98:
> Sorry, I meant that the regex above will not catch all lines from terragrunt - specifically, if you have an un-initialized module, some lines will be missing the prefix. Apologies!

Ah, I gotcha, thanks for clarifying.

> I'd assume that the terraform call also outputs to stderr in some instances, so redirecting all stderr wouldn't be a great solution in those cases.

Yea, Terraform also logs to stderr. However, command where you are explicitly asking something to be printed such as `terraform output` write the requested info to stdout.

voiski:
use the [scenery](https://github.com/dmlittle/scenery) for this, it will filter out the terragrunt output and you will also win colors =D
> --no-color will remove any colors if you want to remove any
```
terragrunt ... | scenery
```

spaszek:
[...]

PR #1510, opened 2021-01-23
Option for less verbose output

amnk:
This moves majority of Terragrunt's output to DEBUG level, and sets
default output mode to ERROR. The idea is that in default run user should
see much less output.

Closes: gruntwork-io/terragrunt#432


cli/cli_app.go
```diff
@@ -4,7 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"log"
+	log "github.com/sirupsen/logrus"
```
brikis98, line 7: NIT: any reason to `import` this under the name `log` instead of `logrus`? It took me a while to figure out that `log` was really `logrus`, so perhaps it'll be confusing for others too?

cli/debug.go
```diff
@@ -31,8 +31,8 @@ func writeTerragruntDebugFile(terragruntOptions *options.TerragruntOptions, terr
 	if err != nil {
 		return err
 	}
-	util.Debugf(terragruntOptions.Logger, "The following variables were detected in the terraform module:")
-	util.Debugf(terragruntOptions.Logger, "%v", variables)
+	terragruntOptions.Logger.Debugf("The following variables were detected in the terraform module:")
```
brikis98, line 34: Nice job replacing `util.Debugf` too!

codegen/generate.go
```diff
@@ -87,24 +87,24 @@ func WriteToFile(logger *log.Logger, basePath string, config GenerateConfig) err
 	if err := ioutil.WriteFile(targetPath, []byte(contentsToWrite), 0644); err != nil {
 		return errors.WithStackTrace(err)
 	}
-	logger.Printf("Generated file %s.", targetPath)
+	terragruntOptions.Logger.Infof("Generated file %s.", targetPath)
```
brikis98, line 90: Could probably be a `Debugf`.

codegen/generate.go
```diff
@@ -114,11 +114,11 @@ func shouldContinueWithFileExists(logger *log.Logger, path string, ifExists Gene
 			return false, err
 		}
 		if !wasGenerated {
-			logger.Printf("ERROR: The file path %s already exists and was not generated by terragrunt.", path)
+			terragruntOptions.Logger.Debugf("ERROR: The file path %s already exists and was not generated by terragrunt.", path)
```
brikis98, line 117: Should be `Errorf`

config/config_helpers.go
```diff
@@ -210,9 +210,9 @@ func runCommand(args []string, include *IncludeConfig, terragruntOptions *option
 	}
 
 	if suppressOutput {
-		terragruntOptions.Logger.Printf("run_cmd output: [REDACTED]")
+		terragruntOptions.Logger.Infof("run_cmd output: [REDACTED]")
 	} else {
-		terragruntOptions.Logger.Printf("run_cmd output: [%s]", cmdOutput.Stdout)
+		terragruntOptions.Logger.Infof("run_cmd output: [%s]", cmdOutput.Stdout)
```
brikis98, line 215: I think these should both be `Debugf`

config/locals.go
```diff
@@ -56,15 +56,15 @@ func evaluateLocalsBlock(
...
 		return nil, nil
 	}
 
-	util.Debugf(terragruntOptions.Logger, "Found locals block: evaluating the expressions.")
+	terragruntOptions.Logger.Debugf("Found locals block: evaluating the expressions.")
 
 	locals, diags := decodeLocalsBlock(localsBlock)
 	if diags.HasErrors() {
-		terragruntOptions.Logger.Printf("Encountered error while decoding locals block into name expression pairs.")
+		terragruntOptions.Logger.Debugf("Encountered error while decoding locals block into name expression pairs.")
```
brikis98, line 67: `Errorf`

config/locals.go
```diff
@@ -90,15 +90,15 @@ func evaluateLocalsBlock(
 			diagsWriter,
 		)
 		if err != nil {
-			terragruntOptions.Logger.Printf("Encountered error while evaluating locals.")
+			terragruntOptions.Logger.Debugf("Encountered error while evaluating locals.")
```
brikis98, line 93: `Errorf`

config/locals.go
```diff
@@ -90,15 +90,15 @@ func evaluateLocalsBlock(
...
 			return nil, err
 		}
 	}
 	if len(locals) > 0 {
 		// This is an error because we couldn't evaluate all locals
-		terragruntOptions.Logger.Printf("Not all locals could be evaluated:")
+		terragruntOptions.Logger.Debugf("Not all locals could be evaluated:")
 		for _, local := range locals {
-			terragruntOptions.Logger.Printf("\t- %s", local.Name)
+			terragruntOptions.Logger.Debugf("\t- %s", local.Name)
```
brikis98, line 101: `Errorf`

config/locals.go
```diff
@@ -135,7 +135,7 @@ func attemptEvaluateLocals(
 
 	evaluatedLocalsAsCty, err := convertValuesMapToCtyVal(evaluatedLocals)
 	if err != nil {
-		terragruntOptions.Logger.Printf("Could not convert evaluated locals to the execution context to evaluate additional locals")
+		terragruntOptions.Logger.Debugf("Could not convert evaluated locals to the execution context to evaluate additional locals")
```
brikis98, line 138: `Errorf`

configstack/running_module.go
```diff
@@ -211,20 +211,20 @@ func (module *runningModule) runModuleWhenReady(semaphore chan struct{}) {
...
 			if module.Module.TerragruntOptions.IgnoreDependencyErrors {
-				module.Module.TerragruntOptions.Logger.Printf("Dependency %s of module %s just finished with an error. Module %s will have to return an error too. However, because of --terragrunt-ignore-dependency-errors, module %s will run anyway.", doneDependency.Module.Path, module.Module.Path, module.Module.Path, module.Module.Path)
+				module.Module.TerragruntOptions.Logger.Debugf("Dependency %s of module %s just finished with an error. Module %s will have to return an error too. However, because of --terragrunt-ignore-dependency-errors, module %s will run anyway.", doneDependency.Module.Path, module.Module.Path, module.Module.Path, module.Module.Path)
 			} else {
-				module.Module.TerragruntOptions.Logger.Printf("Dependency %s of module %s just finished with an error. Module %s will have to return an error too.", doneDependency.Module.Path, module.Module.Path, module.Module.Path)
+				module.Module.TerragruntOptions.Logger.Debugf("Dependency %s of module %s just finished with an error. Module %s will have to return an error too.", doneDependency.Module.Path, module.Module.Path, module.Module.Path)
```
brikis98, line 223: `Errorf`

main.go
```diff
@@ -17,10 +17,6 @@ var VERSION string
 
 // The main entrypoint for Terragrunt
 func main() {
-	// Log the terragrunt version in debug mode. This helps with debugging issues and ensuring a specific version of
-	// terragrunt used.
-	util.Debugf(util.CreateLogger(""), "Terragrunt Version: %s", VERSION)
```
brikis98, line 22: Looks like this used to log the Terragrunt version in debug mode during boot. We shouldn't lose that behavior.

remote/remote_state.go
```diff
@@ -108,13 +107,13 @@ func (remoteState *RemoteState) NeedsInit(terragruntOptions *options.TerragruntO
...
-		terragruntOptions.Logger.Printf("Backend type has changed from %s to %s", existingBackend.Type, remoteState.Backend)
+		terragruntOptions.Logger.Infof("Backend type has changed from %s to %s", existingBackend.Type, remoteState.Backend)
 		return true
 	}
 
 	if !terraformStateConfigEqual(existingBackend.Config, remoteState.Config) {
-		terragruntOptions.Logger.Printf("Backend config has changed (Set environment variable TG_LOG=debug to have terragrunt log the changes)")
-		util.Debugf(terragruntOptions.Logger, "Changed from %s to %s", existingBackend.Config, remoteState.Config)
+		terragruntOptions.Logger.Infof("Backend config has changed (Set environment variable TG_LOG=debug to have terragrunt log the changes)")
+		terragruntOptions.Logger.Debugf("Changed from %s to %s", existingBackend.Config, remoteState.Config)
```
brikis98, line 116: I'd remove the first statement and only have the second

remote/remote_state_gcs.go
```diff
@@ -145,8 +145,8 @@ func gcsConfigValuesEqual(config map[string]interface{}, existingBackend *Terraf
 	}
 
 	if !terraformStateConfigEqual(existingBackend.Config, comparisonConfig) {
-		terragruntOptions.Logger.Printf("Backend config has changed (Set environment variable TG_LOG=debug to have terragrunt log the changes)")
-		util.Debugf(terragruntOptions.Logger, "Changed from %s to %s", existingBackend.Config, config)
+		terragruntOptions.Logger.Infof("Backend config has changed (Set environment variable TG_LOG=debug to have terragrunt log the changes)")
+		terragruntOptions.Logger.Debugf("Changed from %s to %s", existingBackend.Config, config)
```
brikis98, line 149: Same here

remote/remote_state_s3.go
```diff
@@ -216,8 +216,8 @@ func configValuesEqual(config map[string]interface{}, existingBackend *Terraform
 	}
 
 	if !terraformStateConfigEqual(existingBackend.Config, terraformConfig) {
-		terragruntOptions.Logger.Printf("Backend config has changed (Set environment variable TG_LOG=debug to have terragrunt log the changes)")
-		util.Debugf(terragruntOptions.Logger, "Changed from %s to %s", existingBackend.Config, config)
+		terragruntOptions.Logger.Info("Backend config has changed (Set environment variable TG_LOG=debug to have terragrunt log the changes)")
+		terragruntOptions.Logger.Debugf("Changed from %s to %s", existingBackend.Config, config)
```
brikis98, line 220: And here

shell/run_shell_cmd.go
```diff
@@ -57,9 +57,9 @@ func RunShellCommandWithOutput(
 	command string,
 	args ...string,
 ) (*CmdOutput, error) {
-	terragruntOptions.Logger.Printf("Running command: %s %s", command, strings.Join(args, " "))
+	terragruntOptions.Logger.Infof("Running command: %s %s", command, strings.Join(args, " "))
 	if suppressStdout {
-		terragruntOptions.Logger.Printf("Command output will be suppressed.")
+		terragruntOptions.Logger.Info("Command output will be suppressed.")
```
brikis98, line 62: `Debugf`

util/logger.go
```diff
@@ -3,47 +3,55 @@ package util
...
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclparse"
+	log "github.com/sirupsen/logrus"
 	"golang.org/x/crypto/ssh/terminal"
 )
 
-// Create a logger with the given prefix
-func CreateLogger(prefix string) *log.Logger {
-	return CreateLoggerWithWriter(os.Stderr, prefix)
+// CreateLogger creates a logger. If debug is set, we use ErrorLevel to enable verbose output, otherwise - only errors are shown
```
brikis98, line 15: Why not have this method take in the log level param instead of a bool flag? Same for the one below?

util/logger.go
```diff
@@ -3,47 +3,55 @@ package util
...
+	logger := CreateLogger(debug)
+	var fields log.Fields
 	if prefix != "" {
-		prefix = fmt.Sprintf("[%s] ", prefix)
+		prefix = fmt.Sprintf("[%s]", prefix)
+		fields = log.Fields{"prefix": prefix}
+	} else {
+		fields = log.Fields{}
 	}
-	return log.New(writer, fmt.Sprintf("[terragrunt] %s", prefix), log.LstdFlags)
```
brikis98, line 26: Did we lose the `[terragrunt]` prefix?

util/logger.go
```diff
@@ -3,47 +3,55 @@ package util
...
+// CreateLoggerWithWriter Create a logger around the given output stream and prefix
+func CreateLogEntryWithWriter(writer io.Writer, prefix string, debug bool) *log.Entry {
+	logger := CreateLogEntry(prefix, debug)
+	logger.Logger.SetOutput(writer)
+	return logger
 }
 
 // ColorLogf
-func ColorLogf(logger *log.Logger, colorCode *color.Color, fmtString string, fmtArgs ...interface{}) {
+func ColorLogf(logger *log.Entry, colorCode *color.Color, fmtString string, fmtArgs ...interface{}) {
```
brikis98, line 47: Where is this used?

brikis98 (commented):
Nice! I left some detailed thoughts on log levels to use, but also a higher level thought:

1. We should accept a `--log-level` param now.
1. The old debug param should be deprecated, but still supported for backwards compatibility.
1. Please update the docs!

cli/cli_app.go
amnk: Importing `logrus` as `log` allows accessing default logging object. Ans since `logrus` can be used as a drop-in-replacement for `log`, it generally causes no issues. But I will change this to usual imports to avoid confusion.

util/logger.go
amnk: https://github.com/gruntwork-io/terragrunt/blob/master/main.go#L42

amnk: I was not sure about if it is save to remove it, so decided to keep it. But `logrus` by itself does use color output (when a TTY is attached), so maybe we can remove it?

cli/cli_app.go
brikis98: Roger!

util/logger.go
brikis98: I think you can remove it and use `Errorf` in the place it was called from.

cli/cli_app.go
```diff
@@ -125,6 +127,11 @@ var TERRAFORM_COMMANDS_THAT_DO_NOT_NEED_INIT = []string{
 	"graph-dependencies",
 }
 
+// DEPRECATED_ARGUMENTS is a map of deprecated arguments to the argument that replace them.
+var DEPRECATED_ARGUMENTS = map[string]string{
```
brikis98, line 131: Nice addition! 👍

main.go
```diff
@@ -34,9 +30,9 @@ func checkForErrorsAndExit(err error) {
 	if err == nil {
 		os.Exit(0)
 	} else {
-		logger := util.CreateLogger("")
+		logger := util.CreateLogEntry("", "debug")
```
brikis98, line 33: Does this configure this logger to debug log level as its default? Would be nice for the error handling to support `--terragrunt-log-level` too...

main.go
```diff
@@ -34,9 +30,9 @@ func checkForErrorsAndExit(err error) {
...
 		if os.Getenv("TERRAGRUNT_DEBUG") != "" {
-			logger.Println(errors.PrintErrorWithStackTrace(err))
+			logger.Errorln(errors.PrintErrorWithStackTrace(err))
```
brikis98, line 35: If we supported `--terragrunt-log-level`, then we could use `Errorf` on the basic error message (as in the `else` block below) and `Debugf` for the stack trace.

options/options.go
```diff
@@ -70,8 +73,11 @@ type TerragruntOptions struct {
 	// The working directory in which to run Terraform
 	WorkingDir string
 
-	// The logger to use for all logging
-	Logger *log.Logger
+	// Basic log entry
+	Logger *logrus.Entry
+
+	// Log level
+	LogLevel string
```
brikis98, line 80: Does logrus not define an enum with log levels?

options/options.go
```diff
@@ -219,7 +226,7 @@ func NewTerragruntOptionsForTest(terragruntConfigPath string) (*TerragruntOption
 	opts, err := NewTerragruntOptions(terragruntConfigPath)
 
 	if err != nil {
-		logger := util.CreateLogger("")
+		logger := util.CreateLogEntry("", "warn")
```
brikis98, line 229: Use `DEFAULT_LOG_LEVEL`?

test/integration_test.go
```diff
@@ -1043,7 +1043,7 @@ func TestCustomLockFile(t *testing.T) {
 
 	source := "../custom-lock-file-module"
 	downloadDir := util.JoinPath(TEST_FIXTURE_CUSTOM_LOCK_FILE, TERRAGRUNT_CACHE)
-	result, err := cli.ProcessTerraformSource(source, downloadDir, TEST_FIXTURE_CUSTOM_LOCK_FILE, util.CreateLogger(""))
+	result, err := cli.ProcessTerraformSource(source, downloadDir, TEST_FIXTURE_CUSTOM_LOCK_FILE, util.CreateLogEntry("", "warn"))
```
brikis98, line 1046: Same here about `DEFAULT_LOG_LEVEL`. In fact, the log level prob shouldn't be hard-coded anywhere, but fall back to this default var.

util/logger.go
```diff
@@ -3,47 +3,60 @@ package util
...
-// Create a logger with the given prefix
-func CreateLogger(prefix string) *log.Logger {
-	return CreateLoggerWithWriter(os.Stderr, prefix)
+// CreateLogger creates a logger. If debug is set, we use ErrorLevel to enable verbose output, otherwise - only errors are shown
+func CreateLogger(lvl string) *logrus.Logger {
+	logger := logrus.New()
+	level, err := logrus.ParseLevel(lvl)
+	if err != nil {
+		fmt.Println(err.Error())
+		os.Exit(1)
```
brikis98, line 21: We try very hard to avoid calling `panic` or `os.Exit` in pretty much all of our code (see our [Go style guide](https://gruntwork.io/guides/style%20guides/golang-style-guide/#error-handling)), as that makes the code easier to test, re-use, etc. This should probably return an error instead.

brikis98 (commented):
OK, this is very close! One more place to update in the docs: https://terragrunt.gruntwork.io/docs/reference/cli-options/#cli-options.

main.go
amnk: @brikis98 the problem here is that `checkForErrorsAndExit()` is in `main` package, and is executed when default logger is not yet available. So changing this will require refactoring in itself - either it is possible to reuse some of `urfave/cli` functionality, or something move this bit to `cli/cli_app`. What do you think?

brikis98: Moving this error handling into `cli` and making logging more consistent seems like a good idea to me!

cli/args.go
```diff
@@ -290,11 +299,24 @@ func filterTerragruntArgs(args []string) []string {
...
+	}
+	return optionName
+}
+
 // Find a boolean argument (e.g. --foo) of the given name in the given list of arguments. If it's present, return true.
 // If it isn't, return defaultValue.
 func parseBooleanArg(args []string, argName string, defaultValue bool) bool {
 	for _, arg := range args {
 		if arg == fmt.Sprintf("--%s", argName) {
+			argName = isDeprecatedOption(argName)
```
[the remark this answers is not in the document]
amnk: here we need to run `isDeprecatedOption()` to give us deprecation warning. Variable itself is not used.

main.go
amnk: @brikis98 I ended up solving this differently - by introducing a global `LogEntry` object, that gets initiated early. That allows us to log things, and use `--terragrunt-log-level`

cli/args.go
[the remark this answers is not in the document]
brikis98: If the returned value isn't used, we shouldn't set it on the calling side. Otherwise, it makes it _seem_ like it's used, and may confuse maintainers (like me!) if that's intentional or not.

main.go
brikis98: Is this still missing?

main.go
```diff
@@ -34,18 +28,17 @@ func checkForErrorsAndExit(err error) {
...
 		os.Exit(0)
 	} else {
-		logger := util.CreateLogger("")
 		if os.Getenv("TERRAGRUNT_DEBUG") != "" {
-			logger.Println(errors.PrintErrorWithStackTrace(err))
+			util.LogEntry.Debugf(errors.PrintErrorWithStackTrace(err))
 		} else {
 			// Log error in red so that it is highlighted
-			util.ColorLogf(logger, color.New(color.FgRed), err.Error())
+			util.LogEntry.Errorf(err.Error())
```
brikis98, line 35: I'd get rid of the if-else, and leave the `Debugf` and then `Errorf`. The right data will show up based on what log level you have set.

options/options.go
```diff
@@ -28,6 +28,9 @@ const DEFAULT_PARALLELISM = math.MaxInt32
 // TERRAFORM_DEFAULT_PATH just takes terraform from the path
 const TERRAFORM_DEFAULT_PATH = "terraform"
 
+// DEFAULT_LOG_LEVEL defines default log level for Terragrunt
+const DEFAULT_LOG_LEVEL = util.DEFAULT_LOG_LEVEL
```
brikis98, line 32: NIT: Why create an alias rather than use `util.DEFAULT_LOG_LEVEL` everywhere?

remote/remote_state_gcs.go
```diff
@@ -145,8 +145,7 @@ func gcsConfigValuesEqual(config map[string]interface{}, existingBackend *Terraf
 	}
 
 	if !terraformStateConfigEqual(existingBackend.Config, comparisonConfig) {
-		terragruntOptions.Logger.Printf("Backend config has changed (Set environment variable TG_LOG=debug to have terragrunt log the changes)")
-		util.Debugf(terragruntOptions.Logger, "Changed from %s to %s", existingBackend.Config, config)
+		terragruntOptions.Logger.Debugf("Changed from %s to %s", existingBackend.Config, config)
```
brikis98, line 148:
```suggestion
		terragruntOptions.Logger.Debugf("Backend config has changed from %s to %s", existingBackend.Config, config)
```

remote/remote_state_s3.go
```diff
@@ -216,8 +216,7 @@ func configValuesEqual(config map[string]interface{}, existingBackend *Terraform
 	}
 
 	if !terraformStateConfigEqual(existingBackend.Config, terraformConfig) {
-		terragruntOptions.Logger.Printf("Backend config has changed (Set environment variable TG_LOG=debug to have terragrunt log the changes)")
-		util.Debugf(terragruntOptions.Logger, "Changed from %s to %s", existingBackend.Config, config)
+		terragruntOptions.Logger.Debugf("Changed from %s to %s", existingBackend.Config, config)
```
brikis98, line 219:
```suggestion
		terragruntOptions.Logger.Debugf("Backend config has changed from %s to %s", existingBackend.Config, config)
```

util/logger.go
```diff
@@ -3,47 +3,67 @@ package util
...
+		prefix = fmt.Sprintf("[terragrunt] %s", prefix)
 	}
+	logger := CreateLogEntry(prefix, level)
+	logger.Logger.SetOutput(writer)
+	return logger
 }
 
 // ColorLogf
-func ColorLogf(logger *log.Logger, colorCode *color.Color, fmtString string, fmtArgs ...interface{}) {
+func ColorLogf(logger *logrus.Entry, colorCode *color.Color, fmtString string, fmtArgs ...interface{}) {
```
brikis98, line 59: This can probably be removed, as mentioned in a comment earlier, and replaced with `Errorf`.

main.go
amnk: re terragrunt version: I initially moved it to later stage. Now with global logger it makes sense to move it back.

options/options.go
amnk: Yeah, I was not sure about it. I did the alias because all our options are in `options` package. So like we are setting `DEFAULT_PARALLELISM = math.MaxInt32` based on `math` module, here we set defaults based on `util` package. We can use `util` though, but then options would become scattered across different packages. What do you think?

main.go
amnk: I initially moved it to later stage (because we had no logger here). Now with global logger it makes sense to move it back.

cli/args.go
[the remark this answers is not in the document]
amnk: I reworked the `isDeprecatedOption()`, so it would be more obvious.

options/options.go
brikis98: I don't feel strongly either way, so OK to leave as is.

cli/args.go
```diff
@@ -290,11 +307,25 @@ func filterTerragruntArgs(args []string) []string {
 	return out
 }
 
+// isDeprecatedOption checks if provided option is deprecated, and returns its substitution with the check
+// if option is not deprecated - we are returning same value
+func isDeprecatedOption(optionName string) (string, bool) {
+	newOption, deprecated := DEPRECATED_ARGUMENTS[optionName]
```
brikis98, line 313: Ah, OK, returning both values makes that a lot clearer. Nice 👍

docs/_docs/04_reference/cli-options.md
```diff
@@ -502,10 +502,28 @@ When passed in, limit the number of modules that are run concurrently to this nu
 
 **CLI Arg**: `--terragrunt-debug`
 
+**WARNING: Deprecated**. Use `--terragrunt-log-level` instead
+
 When passed in, run Terragrunt in debug mode where the vars that are passed into Terragrunt are recorded as a
 tfvars.json file.
 
 
+### terragrunt-log
```
brikis98, line 511: Doh!

```suggestion
### terragrunt-log-level
```

brikis98 (commented):
Alright, one typo left, and I'll kick off tests!

docs/_docs/04_reference/cli-options.md
amnk: oh my bad :((((

brikis98: No worries at all!

brikis98 (dismissed):
LGTM! I'll kick off tests.

brikis98:
Looks like the precommit hooks failed. Could you run `go fmt` on your code?

brikis98:
Running tests again!

cli/cli_app.go
```diff
@@ -309,7 +309,7 @@ func RunTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 	}
 
 	if terragruntConfig.Skip {
-		terragruntOptions.Logger.Infof("Skipping terragrunt module %s due to skip = true.",
+		terragruntOptions.Logger.Warnf("Skipping terragrunt module %s due to skip = true.",
```
brikis98, line 312: Why change this one to a warning?

cli/args.go
```diff
@@ -39,6 +39,9 @@ func ParseTerragruntOptions(cliContext *cli.Context) (*options.TerragruntOptions
 // and look for the ones we need, but in the future, we should change to a different CLI library to avoid this
 // limitation.
 func parseTerragruntOptionsFromArgs(terragruntVersion string, args []string, writer, errWriter io.Writer) (*options.TerragruntOptions, error) {
+
+	// We need to set proper writer early on
+	util.GlobalFallbackLogEntry.Logger.SetOutput(errWriter)
```
brikis98, line 44: I think we want the logrus stuff to default to stderr as soon as you create it. That way, even if there is an error in `main.go`, before we get here, it still logs to `stderr`.

test/integration_test.go
```diff
@@ -231,8 +231,7 @@ func TestTerragruntInitHookNoSourceNoBackend(t *testing.T) {
 		t.Errorf("Did not expect to get error: %s", err.Error())
 	}
 
-	// `init` hook should execute only once (2 occurrences due to the echo and its output)
-	assert.Equal(t, 2, strings.Count(output, "AFTER_INIT_ONLY_ONCE"), "Hooks on init command executed more than once")
+	assert.Equal(t, 1, strings.Count(output, "AFTER_INIT_ONLY_ONCE"), "Hooks on init command executed more than once")
```
brikis98, line 234: Why change this down to 1?

cli/args.go
amnk: Okay, will push a change shortly

cli/cli_app.go
amnk: @brikis98 because [TestApplySkipTrue() expects](https://github.com/gruntwork-io/terragrunt/blob/master/test/integration_test.go#L2086) this output with default log level (which is "warn" in this PR).

test/integration_test.go
amnk: Because this is what the [test expects](https://github.com/gruntwork-io/terragrunt/blob/master/test/fixture-hooks/init-once/no-source-no-backend/terragrunt.hcl#L10)

cli/cli_app.go
brikis98: Ah, fair enough

brikis98: Actually, we should make that more explicit. You can keep this log statement an `Infof`, but in the `TestApplySkipTrue`, on [this line](https://github.com/gruntwork-io/terragrunt/blob/master/test/integration_test.go#L2078), set `--terragrunt-log-level info`. Good integration test of logging behavior too!

cli/args.go
brikis98: Thx!

test/integration_test.go
amnk: I think seeing it twice was some kind of side effect, which got fixed with this transition.

brikis98: Right, but why was the test originally looking for 2? I'm prob just missing some context, and the comment that was there about "echo and its output" isn't making sense to me.

amnk: @brikis98 so on `master` the output is this:
```
    integration_test.go:3114: [TestTerragruntInitHookNoSourceNoBackend] [terragrunt] [/var/folders/96/29zyl8fn0w10t_3y3cc7hl_r0000gn/T/terragrunt-test251996998/fixture-hooks/init-once/no-source-no-backend] 2021/01/28 16:09:45 Detected 2 Hooks
    integration_test.go:3114: [TestTerragruntInitHookNoSourceNoBackend] [terragrunt] [/var/folders/96/29zyl8fn0w10t_3y3cc7hl_r0000gn/T/terragrunt-test251996998/fixture-hooks/init-once/no-source-no-backend] 2021/01/28 16:09:45 Executing hook: after_init
    integration_test.go:3114: [TestTerragruntInitHookNoSourceNoBackend] [terragrunt] [/var/folders/96/29zyl8fn0w10t_3y3cc7hl_r0000gn/T/terragrunt-test251996998/fixture-hooks/init-once/no-source-no-backend] 2021/01/28 16:09:45 Running command: echo AFTER_INIT_ONLY_ONCE
    integration_test.go:3114: [TestTerragruntInitHookNoSourceNoBackend] AFTER_INIT_ONLY_ONCE
    integration_test.go:3114: [TestTerragruntInitHookNoSourceNoBackend] [terragrunt] 2021/01/28 16:09:45 Running command: terraform apply -auto-approve
    integration_test.go:3114: [TestTerragruntInitHookNoSourceNoBackend] [terragrunt] 2021/01/28 16:09:47 Detected 2 Hooks
    integration_test.go:3114: [TestTerragruntInitHookNoSourceNoBackend]
```
First we see `Running command: echo AFTER_INIT_ONLY_ONCE`, and then actual output `AFTER_INIT_ONLY_ONCE`. With this PR `Running command: ...` is not shown by default (will be seeing with debug level). So that is why we see exactly one `AFTER_INIT_ONLY_ONCE` now

amnk: We can switch `Running command:` to be shown by default (move it from Infof to Warnf), and then the test will be failing with 1 and passing with 2.

brikis98: Ohhhh, I get it! OK, I think the logging improvement here is better, so it's fine to update the test to 1. I just wanted to be sure we understood _why_ that was changing so we could be confident we weren't covering up a bug. I think we're good on this one, thx for digging in 👍

brikis98:
Could you pull in the latest from `master`?

brikis98 (approved):
Woot! Tests are passing.

I'm going to wait for https://github.com/gruntwork-io/terragrunt/pull/1521 to merge, as that fixes the build, so we'll be able to release this.

yorinasub17:
https://github.com/gruntwork-io/terragrunt/pull/1521/ is merged now! Going to merge this in so I can work on the merge conflict in https://github.com/gruntwork-io/terragrunt/pull/1520 . That said, I don't quite have time to look into the details of this branch at this moment, so will defer the release to @brikis98 (otherwise I don't know what to put in the release notes).

brikis98:
Thanks Yori! I'll release this shortly.

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.28.0

PR #1519, opened 2021-01-29
Fix formatting bug in aws-provider-patch

yorinasub17:
The code that is outputted by `aws-provider-patch` formats complex expressions in a way that causes a syntax error in terraform. Specifically, it was formatting the following:

**before**
```
resource "null_resource" "complex_expression" {
  triggers = (
    var.secondary_aws_region == "us-east-1"
    ? { default = "True" }
    : {}
  )
}
```

**after**
```
resource "null_resource" "complex_expression" {
  triggers = (
    var.secondary_aws_region == "us-east-1"
    ? { default = "True" }
    : {}

) }
```

leading to the error

```
Error: Missing newline after argument

  on .terraform/modules/example_module/test/fixture-aws-provider-patch/example-module/main.tf line 33, in resource "null_resource" "complex_expression":
  28:   triggers = (
  29:     var.secondary_aws_region == "us-east-1"
  30:     ? { default = "True" }
  31:     : {}
  33: ) }
```

This PR fixes that by bumping the hcl library version.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 defaults: &defaults
   docker:
-    - image: 087285199408.dkr.ecr.us-east-1.amazonaws.com/circle-ci-test-image-base:tf14
+    - image: 087285199408.dkr.ecr.us-east-1.amazonaws.com/circle-ci-test-image-base:tf14.4
 version: 2
 jobs:
   # We're running unit tests separately from integration tests - with no parallelization.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,7 +58,9 @@ jobs:
     <<: *defaults
     steps:
       - checkout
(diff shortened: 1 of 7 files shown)


brikis98 (approved):
LGTM!

yorinasub17:
Thanks for review! Merging and releasing now!

brikis98:
@yorinasub17 I think the release build for this PR failed. As a result, the binaries didn't get published, so now other builds that depend on them are failing. Would you mind taking a look?

yorinasub17:
> I think the release build for this PR failed. As a result, the binaries didn't get published, so now other builds that depend on them are failing. Would you mind taking a look?

Fixed! I manually did the deploy steps to do the binaries, due to an unrecoverable failure in the deploy step: https://github.com/gruntwork-io/terragrunt/pull/1521

brikis98:
Thank you Yori!

PR #1520, opened 2021-01-29
Introduce run-all to replace xxx-all

yorinasub17:
Implements https://github.com/gruntwork-io/terragrunt/issues/492


cli/cli_app.go
```diff
@@ -878,81 +884,92 @@ func remoteStateNeedsInit(remoteState *remote.RemoteState, terragruntOptions *op
...
-func applyAll(terragruntOptions *options.TerragruntOptions) error {
-	stack, err := configstack.FindStackInSubfolders(terragruntOptions)
-	if err != nil {
-		return err
+	var prompt string
+	switch terragruntOptions.TerraformCommand {
+	case "apply":
+		prompt = "Are you sure you want to run 'terragrunt apply' in each folder of the stack described above?"
+	case "destroy":
+		prompt = "WARNING: Are you sure you want to run `terragrunt destroy` in each folder of the stack described above? There is no undo!"
```
brikis98, line 901: Any reason we only prompt for `apply` and `destroy`? Is it because they might result in changes? If so, what about `import`, `taint`, `untaint`, and `state`? Also, should some command explicitly be blocked from `run-all`? E.g., `console`, `force-unlock`, `login`, `logout`, and a bunch of others don't make sense to combine with `run-all`. Or, alternatively, we could have a set of allowed commands, and block all others.

cli/cli_app.go
```diff
@@ -878,81 +884,92 @@ func remoteStateNeedsInit(remoteState *remote.RemoteState, terragruntOptions *op
...
+	opts.TerraformCommand = "validate"
+	opts.OriginalTerraformCommand = "validate"
+	opts.TerraformCliArgs = append([]string{"validate"}, opts.TerraformCliArgs...)
+	logCommandAllDeprecationWarning(opts, "validate-all")
+	return runAll(opts)
+}
 
-	terragruntOptions.Logger.Printf("%s", stack.String())
-	return stack.Validate(terragruntOptions)
+func logCommandAllDeprecationWarning(terragruntOptions *options.TerragruntOptions, command string) {
```
brikis98, line 970: See also: https://github.com/gruntwork-io/terragrunt/pull/1510/files#diff-43e9ba6ef3ee5016f1ec4acc37037441ca46ffcbf0221417035e48c4393bfbc9R313... We'll want to go with one of these prob.

configstack/stack.go
```diff
@@ -34,19 +33,31 @@ func (stack *Stack) Graph(terragruntOptions *options.TerragruntOptions) {
 	WriteDot(terragruntOptions.Writer, terragruntOptions, stack.Modules)
 }
 
-// Plan execute plan in the given stack in their specified order.
-func (stack *Stack) Plan(terragruntOptions *options.TerragruntOptions) error {
-	stack.setTerraformCommand([]string{"plan"})
+func (stack *Stack) Run(terragruntOptions *options.TerragruntOptions) error {
+	stackCmd := terragruntOptions.TerraformCommand
+	// For apply and destroy, run in non-interactive mode to avoid comingling stdin across multiple concurrent runs.
```
brikis98, line 38: Why only apply and destroy? Couldn't there be `stdin` for, say, `plan`? E.g., If you're missing a variable?

brikis98 (commented):
Very nice work Yori!

cli/cli_app.go
yorinasub17: Good point. I opted to specify the list of commands to explicitly disallow, since I think it is easier to maintain a disallow list than it is to maintain an allow list.

4ba280428beb919a44b8f83607ae1c14496c6c6c

cli/cli_app.go
yorinasub17: Good catch! Consolidated into a more functional deprecation handler: 8fb85adcdf9fcc1061b393903448e71b8f1732aa

configstack/stack.go
yorinasub17: Ah yup that is true. Fixed to use the list of commands tracked in `config/config_helpers.go`: 509598898b8b5897367a11b2a5569877c91e0934

cli/cli_app.go
```diff
@@ -249,25 +284,26 @@ func runApp(cliContext *cli.Context) (finalErr error) {
...
 // checkDeprecated checks if the given command is deprecated.  If so: prints a message and returns the new command.
-func checkDeprecated(command string, terragruntOptions *options.TerragruntOptions) string {
-	newCommand, deprecated := DEPRECATED_COMMANDS[command]
+func checkDeprecated(command string, terragruntOptions *options.TerragruntOptions) (*options.TerragruntOptions, string) {
+	deprecationHandler, deprecated := deprecatedCommands[command]
 	if deprecated {
-		terragruntOptions.Logger.Printf("%v is deprecated; running %v instead.\n", command, newCommand)
-		return newCommand
+		newOptions, newCommand, newCommandFriendly := deprecationHandler(terragruntOptions)
+		terragruntOptions.Logger.Printf("%s is deprecated; running %s instead.\n", command, newCommandFriendly)
```
brikis98, line 296:
```suggestion
		terragruntOptions.Logger.Printf("'%s' is deprecated. Running '%s' instead. Please update your workflows to use '%s', as '%s' may be removed in the future!\n", command, newCommandFriendly, newCommandFriendly, command)
```

cli/cli_app.go
brikis98: Nice 👍 

One more sanity check: we now disallow a few commands... Are there any others beyond `apply` and `destroy` that merit a prompt from Terragrunt?

cli/cli_app.go
```diff
@@ -88,18 +82,63 @@ const CMD_TERRAGRUNT_READ_CONFIG = "terragrunt-read-config"
...
+// The following commands are DEPRECATED
+const (
+	CMD_SPIN_UP      = "spin-up"
+	CMD_TEAR_DOWN    = "tear-down"
+	CMD_PLAN_ALL     = "plan-all"
+	CMD_APPLY_ALL    = "apply-all"
+	CMD_DESTROY_ALL  = "destroy-all"
+	CMD_OUTPUT_ALL   = "output-all"
+	CMD_VALIDATE_ALL = "validate-all"
+)
```
brikis98, line 129: NIT: we have so many constants declared at the top of this file now... Might merit moving them to a `constants.go`. Could be done in a separate PR too.

brikis98 (dismissed):
Nice! Just a few NITs for you to resolve or ignore as you choose.

cli/cli_app.go
yorinasub17: Added `state` to prompt: 7f4d53a . I think that's the only command left?

cli/cli_app.go
yorinasub17: Will defer to another PR. I've been meaning to also update this to adhere to go style by avoiding all caps as well, but then it would lead to too many changes that conflate the PR.

yorinasub17:
@brikis98 Ok addressed comments and sanity checks, and rebased + resolved merge conflicts. I think this is ready for a final review!

yorinasub17:
UPDATE: Fixed broken build in bc4eba9 . Not sure why this works on `master`, but in any case, this fixes a bug from https://github.com/gruntwork-io/terragrunt/pull/1510 where the `LogLevel` is set to `PanicLevel` when the `TerragruntOptions` struct is `Clone`-d.

brikis98 (approved):
LGTM! Nice work Yori 👍

PR #1521, opened 2021-02-01
Fix deploy step

yorinasub17:
Testing with https://github.com/gruntwork-io/terragrunt/releases/tag/v0.0.0-20210201

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,7 +65,9 @@ jobs:
     <<: *defaults
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


yorinasub17:
FYI deploy test was successful (pasting screenshot so I can delete release):

Release_v0_0_0-20210201_·_gruntwork-io_terragrunt

brikis98 (approved):
Ah, yea, I hit that too with `repo-copier`. We'll have to do something similar for every Go build with the newer versions of Go. Thx for fixing 👍

yorinasub17:

Issue #1524, opened 2021-02-03
Creating new S3 backend is only possible with --terragrunt-log-level info


wiegandf:
I was wondering why my `terragrunt plan` didn't do anything for a module that should be stored in a new remote backend (s3).

It turned out that terragrunt 0.28 hides info logs from you - but in those logs you're asked if you want to create a new backend:
```
INFO[0001] Remote state S3 bucket xyz does not exist or you don't have permissions to access it. Would you like Terragrunt to create it? (y/n)
```
I was only able to see this with --terragrunt-log-level info.

It should only affect terragrunt >= 0.28.0.

tkrille:
This also happens, when doing a `terragrunt run-all apply`. The confirmation message is not shown, because it is logged on `INFO` level. After setting `--terragrunt-log-level info` it is shown:

```
INFO[0000] Are you sure you want to run 'terragrunt apply' in each folder of the stack described above? (y/n)
```

I think this message should be logged at `WARN` leven or even bypass the logger entirely.

brikis98:
Oooh, good catch. We'll get this fixed.

brikis98:
Please give https://github.com/gruntwork-io/terragrunt/releases/tag/v0.28.2 a shot!

tkrille:
I can see the confirmation message now without setting the log level. So this works. Thank you very much! :)

But I've noticed this: The message says:

```
Are you sure you want to run 'terragrunt apply' in each folder of the stack described above? (y/n)
```

But there is no stack described above. Because it is logged at `INFO` and not shown with default settings :)

brikis98:
Let's track that further in https://github.com/gruntwork-io/terragrunt/issues/1530

PR #1526, opened 2021-02-03
Write prompts to ErrWriter instead of log

amnk:
Also added small test to integration suit.

Closes: gruntwork-io/terragrunt#1524

### DIFF
--- a/shell/prompt.go
+++ b/shell/prompt.go
@@ -11,6 +11,10 @@ import (
 
 // Prompt the user for text in the CLI. Returns the text entered by the user.
 func PromptUserForInput(prompt string, terragruntOptions *options.TerragruntOptions) (string, error) {
+	// We are writing directly to ErrWriter so the prompt is always visible
+	// no matter what logLevel is configured.
+	//
+	// See https://github.com/gruntwork-io/terragrunt/issues/1524 for additional context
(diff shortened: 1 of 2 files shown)


shell/prompt.go
```diff
@@ -11,10 +11,16 @@ import (
 
 // Prompt the user for text in the CLI. Returns the text entered by the user.
 func PromptUserForInput(prompt string, terragruntOptions *options.TerragruntOptions) (string, error) {
-	terragruntOptions.Logger.Print(prompt)
-
+	n, err := terragruntOptions.ErrWriter.Write([]byte(prompt))
```
brikis98, line 14: NIT: could you add a comment that we should NOT be using a logger here, as this prompt should always be visible, regardless of log level? And perhaps cross-link to the relevant issue?

brikis98 (commented):
Thanks for the fix! Just one minor NIT and then I can kick off tests.

brikis98:
Thx! I'll kick off tests now.

brikis98 (approved):
Tests passed, merging now, thanks!

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.28.2

Issue #1322, opened 2020-08-28
Dependency optimization init is too slow when there are no nested dependencies


yorinasub17:
Since `v0.23.35`, we introduced dependency fetching optimization where it attempts to only handle a single level of dependencies instead of recursively fetching. This depends on running `init` to configure the backend in a working directory to run `output` on.

However, this ignores the cache, and so this is much slower for folks that had only one layer of dependency before. We can further optimize the dependency fetching if we leverage the cache. That is, if the terragrunt cache already exists for the dependency (the module is properly initialized), we should be able to immediately run `terraform output` there.

yorinasub17:
This optimization was implemented and released as part of https://github.com/gruntwork-io/terragrunt/releases/tag/v0.28.3

PR #1533, opened 2021-02-05
Faster dependency outputs

yorinasub17:
Fixes https://github.com/gruntwork-io/terragrunt/issues/1322

This implements a further optimization to dependency output fetching, where if terragrunt detects that the target module is already initialized, the dependency fetcher will directly run `terraform output` instead of running `init`. This skips a relatively long operation, where terraform needs to pull down the providers and modules.

Note that to implement this, I needed to refactor the `TerraformSource` struct so that it is available in the `config` package without a circular dependency between `config` and `cli`. Everything in `cli/tfsource` are functions that were straight copied from `cli`, with a few exceptions where the functions were converted to methods.

(I was compelled to implement this because I noticed this `init` operation is SUPER SLOW for the account-baseline module, which we have dependency links to in the ref arch)

### DIFF
--- a/cli/tfsource/types.go
+++ b/cli/tfsource/types.go
@@ -1,0 +1,214 @@
+package tfsource
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/url"
+	"regexp"
+	"strings"
+
+	"github.com/hashicorp/go-getter"
+	urlhelper "github.com/hashicorp/go-getter/helper/url"
+	"github.com/sirupsen/logrus"
+
+	"github.com/gruntwork-io/terragrunt/errors"
+	"github.com/gruntwork-io/terragrunt/util"
+)
+
+var forcedRegexp = regexp.MustCompile(`^([A-Za-z0-9]+)::(.+)$`)
+
+// This struct represents information about Terraform source code that needs to be downloaded
+type TerraformSource struct {
+	// A canonical version of RawSource, in URL format
+	CanonicalSourceURL *url.URL
+
+	// The folder where we should download the source to
+	DownloadDir string
+
+	// The folder in DownloadDir that should be used as the working directory for Terraform
+	WorkingDir string
+
+	// The path to a file in DownloadDir that stores the version number of the code
+	VersionFile string
+}
+
+func (src *TerraformSource) String() string {
+	return fmt.Sprintf("TerraformSource{CanonicalSourceURL = %v, DownloadDir = %v, WorkingDir = %v, VersionFile = %v}", src.CanonicalSourceURL, src.DownloadDir, src.WorkingDir, src.VersionFile)
+}
+
+// Encode a version number for the given source. When calculating a version number, we simply take the query
+// string of the source URL, calculate its sha1, and base 64 encode it. For remote URLs (e.g. Git URLs), this is
+// based on the assumption that the scheme/host/path of the URL (e.g. git::github.com/foo/bar) identifies the module
+// name and the query string (e.g. ?ref=v0.0.3) identifies the version. For local file paths, there is no query string,
+// so the same file path (/foo/bar) is always considered the same version. See also the encodeSourceName and
+// ProcessTerraformSource methods.
+func (terraformSource TerraformSource) EncodeSourceVersion() string {
+	return util.EncodeBase64Sha1(terraformSource.CanonicalSourceURL.Query().Encode())
+}
+
+// Write a file into the DownloadDir that contains the version number of this source code. The version number is
+// calculated using the EncodeSourceVersion method.
+func (terraformSource TerraformSource) WriteVersionFile() error {
+	version := terraformSource.EncodeSourceVersion()
+	return errors.WithStackTrace(ioutil.WriteFile(terraformSource.VersionFile, []byte(version), 0640))
+}
+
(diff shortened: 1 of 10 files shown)


cli/tfsource/types.go
```diff
@@ -0,0 +1,214 @@
...
+
+	// The path to a file in DownloadDir that stores the version number of the code
+	VersionFile string
+}
+
+func (src *TerraformSource) String() string {
+	return fmt.Sprintf("TerraformSource{CanonicalSourceURL = %v, DownloadDir = %v, WorkingDir = %v, VersionFile = %v}", src.CanonicalSourceURL, src.DownloadDir, src.WorkingDir, src.VersionFile)
+}
+
+// Encode a version number for the given source. When calculating a version number, we simply take the query
```
brikis98, line 39: Per your inspiration Yori 😄 

```suggestion
// Encode a version number for the given source. When calculating a version number, we take the query
```

config/dependency.go
```diff
@@ -418,6 +429,59 @@ func canGetRemoteState(remoteState *remote.RemoteState) bool {
...
+func terragruntAlreadyInit(terragruntOptions *options.TerragruntOptions, configPath string) (bool, string, error) {
+	// We need to first determine the working directory where the terraform source should be located. This is dependent
+	// on the source field of the terraform block in the config.
+	terraformBlockTGConfig, err := PartialParseConfigFile(configPath, terragruntOptions, nil, []PartialDecodeSectionType{TerraformSource})
+	if err != nil {
+		return false, "", err
+	}
+	var workingDir string
+	sourceUrl := GetTerraformSourceUrl(terragruntOptions, terraformBlockTGConfig)
+	if sourceUrl == "" || sourceUrl == "." {
```
brikis98, line 444: When would it be set to "."?

config/dependency.go
```diff
@@ -418,6 +429,59 @@ func canGetRemoteState(remoteState *remote.RemoteState) bool {
...
+		}
+	} else {
+		terraformSource, err := tfsource.NewTerraformSource(sourceUrl, terragruntOptions.DownloadDir, terragruntOptions.WorkingDir, terragruntOptions.Logger)
+		if err != nil {
+			return false, "", err
+		}
+		// We're only interested in the computed working dir.
+		workingDir = terraformSource.WorkingDir
+	}
+	// Terragrunt is already init-ed if the terraform state dir (.terraform) exists in the working dir.
```
brikis98, line 460: Does this cause problems if some `ref` parameter changed? Or a `provider` version changed? That is, if the user has a `.terraform` folder in that folder, but still needs to force `init` to run, is there a way to do that?

brikis98 (commented):
Nice improvement!

config/dependency.go
yorinasub17: In this case, the working dir would be different, and thus it won't detect that the module is already `init`-ed. This is because the [working dir depends on the download dir which depends on the ref](https://github.com/gruntwork-io/terragrunt/blob/yori-faster-outputs/cli/tfsource/types.go#L118).

config/dependency.go
yorinasub17: This is more defensive coding for certain user behavior, where one might decide that it is better to be explicit instead of implicit.

cli/tfsource/types.go
yorinasub17: Ah good catch! I was blindly copy pasting so didn't really comb through these. 👍

config/dependency.go
yorinasub17: Added comment clarifying this: 92e9001

config/dependency.go
brikis98: Roger

config/dependency.go
brikis98: Right... But what about the intermittent errors where Terraform fails to download a provider? The `.terraform` folder exists, and `ref` did not change, but yet, you need to run `init` again. Is there a way to force that? I ask because, sadly, these intermittent errors happen quite often, esp with `xxx-all` commands.

yorinasub17: Ah gotcha. I see what you are getting at. Let me do a few experiments to see if we can autodetect this, but I think an easy workaround is for someone to run `terragrunt run-all init`.

yorinasub17: UPDATE: it looks like you can still retrieve outputs even if you enter a state where you need to reinit. Here's what I tried to see this behavior:

- Created a module with a single resource:
```
provider "null" {
  version = "~> 2.0"
}

resource "null_resource" "foo" {}

output "foo" {
  value = null_resource.foo.id
}
```

- Ran `terraform init && terraform apply`. Verified outputs can be retrieved with `terraform output`
- Updated the `version` field for the `null` provider. Verified outputs can be retrieved with `terraform output`.
- Added a module block, and confirmed `plan` fails (`Error: Module not installed`). However, outputs were retrievable even without init with `terraform output`.

brikis98: Roger, thanks for checking! That seems good enough to me. If some issue does arise, then you're right, `run-all init` is prob good enough. Thx!

brikis98 (approved):
LGTM!

yorinasub17:
Thanks for review! Going to merge this and release.

PR #1547, opened 2021-02-12
Add ability to force skip auto-init

yorinasub17:
This is a workaround for https://github.com/gruntwork-io/terragrunt/issues/1545

Sometimes, we need to be able to run a terraform command with partial initialization (in this case, only the state needs to be initialized and nothing else). However, terragrunt attempts to detect full initialization of modules and refuse to run terraform commands even when the `--terragrunt-no-auto-init` flag is passed in. This is problematic for the issue linked above as now there is no way to run the required upgrade command to finish initialization.

To workaround this, this PR introduces another flag `--terragrunt-force-skip-init` that will bypass the error message.

### DIFF
--- a/cli/args.go
+++ b/cli/args.go
@@ -152,6 +152,7 @@ func parseTerragruntOptionsFromArgs(terragruntVersion string, args []string, wri
 
 	opts.TerraformPath = filepath.ToSlash(terraformPath)
 	opts.AutoInit = !parseBooleanArg(args, OPT_TERRAGRUNT_NO_AUTO_INIT, os.Getenv("TERRAGRUNT_AUTO_INIT") == "false")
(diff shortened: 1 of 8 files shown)


brikis98 (commented):
I'm sure I'm missing something obvious, but why doesn't the existing `terragrunt-no-auto-init` skip auto init?

yorinasub17:
> I'm sure I'm missing something obvious, but why doesn't the existing terragrunt-no-auto-init skip auto init?

Are you suggesting that we change the `terragrunt-no-auto-init` to avoid this error message in the first place?

yorinasub17:
Ok implemented the version where `no-auto-init` doesn't error out (https://github.com/gruntwork-io/terragrunt/pull/1549). I used a separate PR + branch in case I misunderstood and we want to come back to this flavor. I'll close this PR for now (but we can reopen if we want this version).

PR #1549, opened 2021-02-12
Remove InitNeededButDisabled error and instead log a warning

yorinasub17:
Alternative implementation of https://github.com/gruntwork-io/terragrunt/pull/1547, where we simply bypass the disabled init error.


brikis98 (approved):
LGTM!

yorinasub17:
Thanks for review! Merging now!

PR #1550, opened 2021-02-13
follow symlinks for path_relative_to_include

### LINKED ISSUE #543: path_relative_to_include and symlinks
We have the following kind of setup:

```
dev-account/us-west-2/env-dev/terraform.tfvars
env/dev -> dev-account/us-west-2/env-dev
```

The `env/dev` symlink is just a convenience for manually navigating the directory structure. The terragrunt config specifies a remote state path using `path_relative_to_include()`. This evaluates differently depending on whether you've navigated to the directory directly or via the symlink. Is this desirable behavior, or should the symlink be expanded?

Patricol:
Closes #543

My use case involves wanting to prefix terragrunt dirs with numbers to make the stage order obvious, generate state keys based on path to ensure uniqueness, and use a fair number of `terraform_remote_state` objects.

Reordering or inserting stages is a time-consuming process involving moving many different state files, adjusting `terraform_remote_state` objects, and renaming folders.

Also, `cd`ing isn't as frictionless as it could be; with the extra often 0-padded prefixes.

Creating numbered symlinks to the terragrunt dirs (but having terragrunt ignore those symlinks) would be an ideal solution.

Things to consider:
- I think `path_relative_to_include` should take a boolean argument if possible; defaulting to the old behavior but allowing the use of this behavior.
- Should this be done for `path_relative_from_include` too?
- Haven't yet tested. Especially if not implementing the bool toggle; need to see how this behaves when a relative include path is valid from one but not both of the real and symlinked directories.
- Is it alright that this runs `filepath.Clean` on the path too?

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of symbolic links in configuration paths to ensure more accurate path resolution.

### DIFF
--- a/config/config_helpers.go
+++ b/config/config_helpers.go
@@ -301,6 +301,7 @@ func pathRelativeToInclude(include *IncludeConfig, terragruntOptions *options.Te
 
 	includePath := filepath.Dir(include.Path)
 	currentPath := filepath.Dir(terragruntOptions.TerragruntConfigPath)
+	currentPath = filepath.EvalSymlinks(currentPath)
 
 	if !filepath.IsAbs(includePath) {
 		includePath = util.JoinPath(currentPath, includePath)


Patricol:
@brikis98 @yorinasub17 

Happy to work on this more if I get confirmation that the 'help wanted' tag on #543 is still accurate. (Maybe after >2 years the change is no longer desired?)

brikis98:
@Patricol Sorry for the delay. We are super, super buried right now. The PR is very much appreciated and we'll review it as soon as we can!

brikis98:
Thanks for the PR and your patience as we got around to reviewing it!

> Things to consider:
> 
> * I think `path_relative_to_include` should take a boolean argument if possible; defaulting to the old behavior but allowing the use of this behavior.

Unless you know of a specific use case where the current approach would break as a result, I'd rather avoid extra flags / branches.

> * Should this be done for `path_relative_from_include` too?

Yes, probably.

> * Is it alright that this runs `filepath.Clean` on the path too?

Sure.

Note that for symlinks to work, we may also need to solve https://github.com/gruntwork-io/terragrunt/issues/1611.

PR #1563, opened 2021-02-25
Fix dead link in multiple aws accounts docs

paulvandenburg:
The link to AWS docs is now 404.
The corrected link seems to most closely resemble the intended target.
Other options to consider:
https://aws.amazon.com/organizations/getting-started/best-practices/
https://docs.aws.amazon.com/controltower/latest/userguide/aws-multi-account-landing-zone.html

### DIFF
--- a/docs/_docs/02_features/work-with-multiple-aws-accounts.md
+++ b/docs/_docs/02_features/work-with-multiple-aws-accounts.md
@@ -14,7 +14,7 @@ nav_title_link: /docs/
 
 ### Motivation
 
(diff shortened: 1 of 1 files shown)


brikis98 (commented):
Ah, good catch! I think the other link you found might be a better replacement: https://aws.amazon.com/organizations/getting-started/best-practices/.

paulvandenburg:
@brikis98 Replaced the link with the best practice version.

brikis98 (approved):
Thank you!

PR #1541, opened 2021-02-10
Change default logging level to Info

amnk:
Recent PR (#1510) changed logging, and also introduced default logging
level. While it seemed like a good idea, it also introduced a lot of
confusion, because in logrus `Println()` is an alias to `Info()`
loglevel - so lots of messages were lost.

This change restores previous behavior in logging output.

Related: gruntwork-io/terragrunt#1529
Related: gruntwork-io/terragrunt#1530
Related: gruntwork-io/terragrunt#1531
Related: gruntwork-io/terragrunt#1532

PR #1568, opened 2021-02-28
Fix empty outputs

amnk:
If stack run finished without errors, `summarizePlanAllErrors()`
receives empty buffer and outputs empty line. This change ensures that
only non-empty outputs are getting logged.

Related: #1541

### DIFF
--- a/configstack/stack.go
+++ b/configstack/stack.go
@@ -81,6 +81,12 @@ func (stack *Stack) Run(terragruntOptions *options.TerragruntOptions) error {
 func (stack *Stack) summarizePlanAllErrors(terragruntOptions *options.TerragruntOptions, errorStreams []bytes.Buffer) {
 	for i, errorStream := range errorStreams {
 		output := errorStream.String()
+
+		if len(output) == 0 {
+			// We get empty buffer if stack execution completed without errors, so skip that to avoid logging too much
+			continue
+		}
+
 		terragruntOptions.Logger.Infoln(output)
 		if strings.Contains(output, "Error running plan:") {
 			if strings.Contains(output, ": Resource 'data.terraform_remote_state.") {


amnk:
This problem was initially reported in https://github.com/gruntwork-io/terragrunt/pull/1541#issuecomment-785725213

brikis98 (commented):
@amnk I'm not sure I entirely follow. The outstanding work from the comments in #1541 was:

> > The following log lines appear hundreds of times within the output and I believe are not necessary with the normal log level:
> > ```
> > => Module /Users/xxx/Desktop/Repositories/xxx/xxx/eu-central-1/prod/xxx/s3-bucket/ml (excluded: true, dependencies: [])
> > ```
> 
> Hm, I believe these are necessary when we are interactively prompt the user if they want to run their `run-all` command on all these modules. @amnk These probably need to be bundled into the prompt message show to users, so that if `--non-interactive` is set, they are only logged at the `debug` level, but otherwise, show up as epxected. Not sure why they would appear hundreds of times though?

Does this PR handle this?

> 
> > ```
> > INFO[0072]                                               prefix=[/Users/xxx/Desktop/Repositories/xxx]
> > ```
> 
> Not sure what this is. @amnk Could you investigate?

Or is this PR related to these weird empty lines?

Or, another way to approach it, is perhaps to share what the log output looked like before and after this PR? Thanks!

amnk:
@brikis98 this PR fixes empty lines only, as it's a bug. First part (with Module output which is apparently excessive) will require some more work.

configstack/stack.go
```diff
@@ -81,18 +81,23 @@ func (stack *Stack) Run(terragruntOptions *options.TerragruntOptions) error {
...
-		terragruntOptions.Logger.Infoln(output)
-		if strings.Contains(output, "Error running plan:") {
-			if strings.Contains(output, ": Resource 'data.terraform_remote_state.") {
-				var dependenciesMsg string
-				if len(stack.Modules[i].Dependencies) > 0 {
-					dependenciesMsg = fmt.Sprintf(" contains dependencies to %v and", stack.Modules[i].Config.Dependencies.Paths)
+
+		// We receive empty buffer if stack execution completed without errors
+		// But since we still want to capture full output if error occured - we check length of resulting string here
+		if len(output) > 0 {
```
brikis98, line 87: NIT: might be a tiny bit clearer to invert this. Something like:

```go
if len(output) == 0 {
  // If there were no errors, we'll get an empty buffer, so skip that to avoid logging lots of extra newlines
  continue
}

brikis98 (commented):
Understood, thanks!

configstack/stack.go
amnk: @brikis98 yeah, your version looks cleaner 👍

brikis98 (approved):
LGTM! I'll kick off tests now.

brikis98:
Tests passed. Merging now.

PR #1572, opened 2021-03-04
Introduce validate-inputs, which can be used to check for variable alignment

yorinasub17:
Refer to [the docs](https://github.com/gruntwork-io/terragrunt/blob/39b67b52ecde9b49d66cb43820f1f60cd27fb31b/docs/_docs/04_reference/cli-options.md#terragrunt-input-info) for info on this new command.


cli/args.go
```diff
@@ -220,23 +220,9 @@ func filterTerraformExtraArgs(terragruntOptions *options.TerragruntOptions, terr
...
-					// It is possible that many files resolve to the same path, so we remove duplicates.
-					if arg.OptionalVarFiles != nil {
-						for _, file := range util.RemoveDuplicatesFromListKeepLast(*arg.OptionalVarFiles) {
-							if util.FileExists(file) {
-								out = append(out, fmt.Sprintf("-var-file=%s", file))
-							} else {
-								terragruntOptions.Logger.Debugf("Skipping var-file %s as it does not exist", file)
-							}
-						}
+					varFiles := arg.GetVarFiles(terragruntOptions.Logger)
```
yorinasub17, line 223: This is a refactor to have a common function that can be used to extract all the var files that terragrunt should set.

config/config.go
```diff
@@ -487,59 +513,6 @@ func decodeAsTerragruntConfigFile(
...
-		return file, nil
-	}
-
-	file, parseDiagnostics := parser.ParseHCL([]byte(hcl), filename)
-	if parseDiagnostics != nil && parseDiagnostics.HasErrors() {
-		return nil, parseDiagnostics
-	}
-
-	return file, nil
-}
```
yorinasub17, line 541: These functions were refactored out into `hcl_parser.go` so all the HCL parsing related functions were consolidated.

config/hcl_parser.go
```diff
@@ -0,0 +1,115 @@
...
+
+	decodeDiagnostics := gohcl.DecodeBody(file.Body, evalContext, out)
+	if decodeDiagnostics != nil && decodeDiagnostics.HasErrors() {
+		return decodeDiagnostics
+	}
+
+	return nil
+}
+
+// ParseAndDecodeVarFile uses the HCL2 parser to parse the given varfile string into an HCL file body, and then decode it
```
yorinasub17, line 68: This function is a straight copy from terratest for now. In a near-future PR, I plan on refactoring the cty helpers into `gruntwork-cli` so we can reuse it.

test/integration_test.go
```diff
@@ -124,97 +124,12 @@ const (
 	TERRAFORM_STATE                                         = "terraform.tfstate"
```
yorinasub17, line 124: The tests that were removed here were refactored out given how big this file is.

test/integration_serial_test.go
```diff
@@ -0,0 +1,209 @@
...
+func TestPriorityOrderOfArgument(t *testing.T) {
+	out := new(bytes.Buffer)
+	injectedValue := "Injected-directly-by-argument"
+	runTerragruntRedirectOutput(t, fmt.Sprintf("terragrunt apply -auto-approve -var extra_var=%s --terragrunt-non-interactive --terragrunt-working-dir %s", injectedValue, TEST_FIXTURE_EXTRA_ARGS_PATH), out, os.Stderr)
+	t.Log(out.String())
+	// And the result value for test should be the injected variable since the injected arguments are injected before the suplied parameters,
+	// so our override of extra_var should be the last argument.
+	assert.Contains(t, out.String(), fmt.Sprintf(`test = "%s"`, injectedValue))
+}
+
```
yorinasub17, line 194: The two tests after this line are new. Everything before are from `integration_test.go` that needed to run in serial.

test/integration_serial_test.go
```diff
@@ -0,0 +1,209 @@
+package test
```
yorinasub17, line 1: NOTE: scroll to next comment in this file that marks which test is new. Everything inbetween are from `integration_test.go`

test/integration_debug_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/gruntwork-io/terragrunt/util"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	TERRAGRUNT_DEBUG_FILE = "terragrunt-debug.tfvars.json"
+)
+
+func TestDebugGeneratedInputs(t *testing.T) {
```
yorinasub17, line 23: This test is migrated from `integration_test.go`. No need to review.

test/integration_debug_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// Also make sure the undefined variable is not included in the json file
+	debugJsonContents, err := ioutil.ReadFile(debugFile)
+	require.NoError(t, err)
+	var data map[string]interface{}
+	require.NoError(t, json.Unmarshal(debugJsonContents, &data))
+	_, isDefined := data["undefined_var"]
+	assert.False(t, isDefined)
+}
+
```
yorinasub17, line 64: Everything below this line in this file is new.

config/config.go
```diff
@@ -487,59 +513,6 @@ func decodeAsTerragruntConfigFile(
 	return &terragruntConfig, nil
 }
```
yorinasub17, line 515: These functions were refactored out into hcl_parser.go so all the HCL parsing related functions were consolidated.

cli/input_info.go
```diff
@@ -0,0 +1,188 @@
...
+
+	// Unused variables are those that are passed in by terragrunt, but is not defined in terraform.
+	unusedVars := []string{}
+	for _, varName := range allInputs {
+		if !util.ListContainsElement(allVars, varName) {
+			unusedVars = append(unusedVars, varName)
+		}
+	}
+
+	// Missing variables are those that are required by the terraform config, but is not defined in terragrunt.
```
zackproser, line 36:
```suggestion
	// Missing variables are those that are required by the terraform config, but not defined in terragrunt.
```

cli/input_info.go
```diff
@@ -0,0 +1,188 @@
...
+// - inputs blocks.
+func getDefinedTerragruntInputs(terragruntOptions *options.TerragruntOptions, workingConfig *config.TerragruntConfig) ([]string, error) {
+	envVarTFVars := getTerraformInputNamesFromEnvVar(terragruntOptions, workingConfig)
+	inputsTFVars := getTerraformInputNamesFromConfig(workingConfig)
+	varFileTFVars, err := getTerraformInputNamesFromVarFiles(terragruntOptions, workingConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	// Dedup the input vars. We use a map as a set to accomplish this.
```
zackproser, line 86:
```suggestion
	// Dedupe the input vars. We use a map as a set to accomplish this.
```

cli/cli_app.go
```diff
@@ -81,6 +81,7 @@ const CMD_INIT_FROM_MODULE = "init-from-module"
 const CMD_PROVIDERS = "providers"
 const CMD_LOCK = "lock"
 const CMD_TERRAGRUNT_INFO = "terragrunt-info"
+const CMD_TERRAGRUNT_VALIDATE_INPUTS = "validate-inputs"
```
brikis98, line 84: Should it be `terragrunt-validate-inputs`? We prefix just about all Terragrunt-specific commands with `terragrunt-`... And used to even have logic to filter those args out of args lists based on that prefix.

cli/validate_inputs.go
```diff
@@ -0,0 +1,188 @@
...
+		return err
+	}
+	allVars := append(required, optional...)
+
+	allInputs, err := getDefinedTerragruntInputs(terragruntOptions, workingConfig)
+	if err != nil {
+		return err
+	}
+
+	// Unused variables are those that are passed in by terragrunt, but is not defined in terraform.
```
brikis98, line 28:
```suggestion
	// Unused variables are those that are passed in by terragrunt, but are not defined in terraform.
```

cli/validate_inputs.go
```diff
@@ -0,0 +1,188 @@
...
+		for _, varName := range unusedVars {
+			terragruntOptions.Logger.Warnf("\t- %s", varName)
+		}
+		terragruntOptions.Logger.Warn("")
+	} else {
+		terragruntOptions.Logger.Info("All variables passed in by terragrunt are in use.")
+	}
+
+	if len(missingVars) > 0 {
+		terragruntOptions.Logger.Warn("The following required inputs are missing:\n")
```
brikis98, line 56: Shouldn't these be errors?

cli/validate_inputs.go
```diff
@@ -0,0 +1,188 @@
...
+	// Return an error when there are misaligned inputs.
+	if len(unusedVars) > 0 || len(missingVars) > 0 {
+		return fmt.Errorf("Terragrunt configuration has misaligned inputs")
+	}
+	return nil
+}
+
+// getDefinedTerragruntInputs will return a list of names of all variables that are configured by terragrunt to be
+// passed into terraform. Terragrunt can pass in inputs from:
+// - var files defined on terraform.extra_arguments blocks.
```
brikis98, line 74: Hm, I think there's are several categories of variables missing here: 

- `.tfvars`, `.tfvars.json`, `.auto.tfvars`, and `*.auto.tfvars.json` files in the working directory.
- `.tfvars`, `.tfvars.json`, `.auto.tfvars`, and `*.auto.tfvars.json` files in the module code.
- Var files passed in via `-var-file` CLI args (not `extra_args`, but normal CLI args passed to the `terragrunt` command).
- Variables passed in via `-var` CLI args in `extra_args`.
- Variables passed in via `-var` CLI args (not `extra_args`, but normal CLI args passed to the `terragrunt` command).

There may be others too, which makes me worry about maintaining this functionality... Is there any way to run `terraform` to tell us which vars got passed in? E.g., Run some command with no resources, only the input vars?

brikis98 (commented):
Nice addition! 

Didn't you have a prev hackday project that could auto-generate a `terragrunt.hcl` file, with placeholders for all input vars in an `inputs` block, for a given module too?

cli/validate_inputs.go
yorinasub17: > Is there any way to run terraform to tell us which vars got passed in?

Nope there's nothing that will tell us this. Otherwise, I would have used that here :)

cli/cli_app.go
yorinasub17: > We prefix just about all Terragrunt-specific commands with terragrunt-...

This is not true:

```
COMMANDS:
   run-all              Run a terraform command against a 'stack' by running the specified command in each subfolder. E.g., to run 'terragrunt apply' in each subfolder, use 'terragrunt run-all apply'.
   terragrunt-info      Emits limited terragrunt state on stdout and exits
   graph-dependencies   Prints the terragrunt dependency graph to stdout
   hclfmt               Recursively find terragrunt.hcl files and rewrite them into a canonical format.
   aws-provider-patch   Overwrite settings on nested AWS providers to work around a Terraform bug (issue #1430
```

Note that only `terragrunt-info` is prefixed with terragrunt.

> And used to even have logic to filter those args out of args lists based on that prefix.

This logic is only for option flags, not commands (e.g., `--terragrunt-config` and `--terragrunt-check`).

cli/validate_inputs.go
yorinasub17: > Hm, I think there's are several categories of variables missing here:

I went for the 80-20 rule here. I think what I have covers 80% of the use cases (possibly more), for 20% of the effort.

Out of those, the configured `-var` CLI args are the only one in that list IMO that probably is worth supporting given that that is a static configuration. I added that in: 6632f4c (I also added docs in that commit to clarify which variable sources are checked in the command.)

For the others, I think it is ok if we punt supporting them until someone requests it:

- I think it's an anti pattern to be using automatically loaded tfvars in a terragrunt project. It has an element of surprise, and breaks the rule where you can't use `terragrunt.hcl` as a root (that is, looking at `terragrunt.hcl` won't tell you the full picture anymore).

- I avoided the CLI args based variable manipulation (those that are passed in via normal CLI args) because they are dynamic across each run. I view this more of a linter for the terragrunt configuration, and it feels wrong to me for a linter to change its results based on dynamic CLI args.*

\* I was on the fence about including the `TF_VAR` env vars as well for the same reason, but since we have this logic deeply baked in to the variable setting logic of terragrunt, I decided to just include it.

cli/validate_inputs.go
yorinasub17: Updated to be error logs: a276745

cli/cli_app.go
brikis98: Ah, doh, you're right, I got the flags confused with commands.

cli/validate_inputs.go
brikis98: > > Is there any way to run terraform to tell us which vars got passed in?
> 
> Nope there's nothing that will tell us this. Otherwise, I would have used that here :)

One option to consider:

- Extract solely the variable declarations from the Terraform code into a tmp folder.
- Run `terraform plan -out=tmp.plan` in that folder.
- Run `terraform show -json tmp.plan`, passing all variables the way Terragrunt would normally do it. I believe this will give you the values Terraform found for every variable?

brikis98: Ah, perhaps the `plan` command will fail if some of the variables are not set though... Damn, so close.

brikis98: > > Hm, I think there's are several categories of variables missing here:
> 
> I went for the 80-20 rule here. I think what I have covers 80% of the use cases (possibly more), for 20% of the effort.

If we already have all the logic to handle `-var` and `-var-files` in `extra_arguments`, is it that much more to find the same passed via CLI commands and to check for the `.tfvars` files?

I'm pushing on this because I have the impression that it's not 80% of use cases... `.tfvars` files in particular seem very common in all Terraform usage, including Terragrunt, in no small part because earlier versions of Terragrunt had their configuration within `.tfvars` files; also, we didn't used to have `read_terragrunt_config`, `yamldecode`, etc, so reliance on `.tfvars` files was very common!

brikis98: Ohhh... One more thought on using Terraform to validate the args:

- As of 0.12.x, if you set a variable _in a `.tfvars` file_ (not an env var) that isn't defined in a variable block, Terraform would give an error.
- In all versions, if you didn't set a required variable, Terraform would give an error.

So I wonder if this new command should write all Terragrunt variables to a temporary `.tfvars` file, run `plan -json` or `validate -json` with that, delete the temporary file, and report the results?

yorinasub17: > If we already have all the logic to handle -var and -var-files in extra_arguments, is it that much more to find the same passed via CLI commands and to check for the .tfvars files?

Ok I will add this in, because you are right: this is probably trivial to add now. That said...

> I'm pushing on this because I have the impression that it's not 80% of use cases... .tfvars files in particular seem very common in all Terraform usage, including Terragrunt, in no small part because earlier versions of Terragrunt had their configuration within .tfvars files; also, we didn't used to have read_terragrunt_config, yamldecode, etc, so reliance on .tfvars files was very common!

I might be missing something, but I thought the tfvars usage that we've been recommending and showing in examples are all based on `extra_arguments`, and not dynamically passed args via the CLI for each run. I thought one of the benefits of terragrunt was that you **didn't** need to set those variables each time you ran `apply`/`plan`, so that it was all captured as code.

E.g., the old way (prior to `yamldecode`) of doing things used `extra_arguments` which this supports already: https://github.com/gruntwork-io/terragrunt-infrastructure-live-example/blob/7cdce0f96c580fc7808b191c5ed3fab463d81958/prod/terraform.tfvars#L19

> So I wonder if this new command should write all Terragrunt variables to a temporary .tfvars file, run plan -json or validate -json with that, delete the temporary file, and report the results?

I don't think this works. `validate` doesn't check variables (from the help text, bold mine):

> Validate runs checks that verify whether a configuration is syntactically valid and internally consistent, **regardless of any provided variables** or existing state.

And there is no `-json` option for `plan`, and AFAIK `out` doesn't produce an output if there is an error.

brikis98: > I might be missing something, but I thought the tfvars usage that we've been recommending and showing in examples are all based on `extra_arguments`, and not dynamically passed args via the CLI for each run. I thought one of the benefits of terragrunt was that you **didn't** need to set those variables each time you ran `apply`/`plan`, so that it was all captured as code.

Terragrunt deliberately copies things in pwd to the working dir, including `.tfvars` files. That was specifically so that `terraform.tfvars`, which is what we used to use for Terragrunt's config, would be copied too, and any variables set within it would "just work." I suspect there a non-trivial number of users relying on that copying behavior still.

> > So I wonder if this new command should write all Terragrunt variables to a temporary .tfvars file, run plan -json or validate -json with that, delete the temporary file, and report the results?
> 
> I don't think this works. `validate` doesn't check variables (from the help text, bold mine):
> 
> > Validate runs checks that verify whether a configuration is syntactically valid and internally consistent, **regardless of any provided variables** or existing state.
> 
> And there is no `-json` option for `plan`, and AFAIK `out` doesn't produce an output if there is an error.

Ah, bummer.

yorinasub17: I just did a check with `plan -out` and `validate`:

- For plan, when there is an unused var, it only shows a warning in the log, and that var doesn't show up in the plan file:

```
Warning: Value for undeclared variable

The root module does not declare a variable named "unused_input" but a value
was found in file "terraform.tfvars". To use this value, add a "variable"
block to the configuration.

Using a variables file to set an undeclared variable is deprecated and will
become an error in a future release. If you wish to provide certain "global"
settings to all configurations in your organization, use TF_VAR_...
environment variables to set these instead.


Warning: Value for undeclared variable

The root module does not declare a variable named "foo" but a value was found
in file "terraform.tfvars". To use this value, add a "variable" block to the
configuration.

Using a variables file to set an undeclared variable is deprecated and will
become an error in a future release. If you wish to provide certain "global"
settings to all configurations in your organization, use TF_VAR_...
environment variables to set these instead.
```

- When there is a missing required var, this becomes an error log and no plan file is generated:

```
Error: No value for required variable

  on main.tf line 1:
   1: variable "input" {}

The root module input variable "input" is not set, and has no default value.
Use a -var or -var-file command line argument to provide a value for this
variable.
```

- Validate doesn't fail with either error case

---

So in theory we can run plan in the mode you specified to handle this. There are a few downsides to this though, one of them being major:

- [major] Writing secrets to disk: https://github.com/gruntwork-io/terragrunt/pull/1267#issuecomment-667530771

- In the current implementation, the variables values you provide can be mocks even if you are using data sources, because it doesn't actually run any of the resources. With the `plan` approach, you can get extraneous error messages resulting from unresolvable data sources.

brikis98: Yea... If we were *nix only, we could probably generate purely in-memory `.tfvars` files, and pipe those to Terraform... But not sure there's portable way to do that. Sadly, prob not worth it.

yorinasub17: Ok added in autoloaded tfvars logic: ed37094

cli/validate_inputs.go
```diff
@@ -0,0 +1,317 @@
...
+// variables from extra_arguments blocks to see if there are any TF_VAR environment variables that set terraform
+// variables. This will return the list of names of variables that are set in this way by the given terragrunt
+// configuration.
+func getTerraformInputNamesFromEnvVar(terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig) []string {
+	envVars := terragruntOptions.Env
+
+	// Make sure to check if there are configured env vars in the parsed terragrunt config.
+	if terragruntConfig.Terraform != nil {
+		for _, arg := range terragruntConfig.Terraform.ExtraArgs {
+			// only focus on plan args
```
brikis98, line 135: Sorry, why only plan args?

cli/validate_inputs.go
```diff
@@ -0,0 +1,317 @@
...
+	if err != nil {
+		return inputNames, err
+	}
+
+	if terragruntConfig.Terraform != nil {
+		for _, arg := range terragruntConfig.Terraform.ExtraArgs {
+			// only focus on plan args
+			if util.ListContainsElement(arg.Commands, "plan") {
+				if arg.Arguments != nil {
+					vars, rawVarFiles, err := getVarFlagsFromArgList(*arg.Arguments)
```
brikis98, line 195: Is this different than what `getTerraformInputNamesFromVarFiles` is doing?

docs/_docs/04_reference/cli-options.md
```diff
@@ -217,6 +218,51 @@ Might produce output such as:
...
+
+- `TF_VAR` environment variables defined on `terraform.extra_arguments` blocks.
+
+- `TF_VAR` environment variables defined in the environment.
+
+Be aware that other ways to pass variables to `terraform` are not checked by this command.
+
+This command will exit with an error if terragrunt detects any unused inputs or undefined required inputs.
+
+Example:
```
brikis98, line 249: NIT: Put this example higher up, right after the intro paragraph, and all the details about what variables it checks below that.

brikis98 (commented):
Almost there! Just a few minor sanity checks.

cli/validate_inputs.go
yorinasub17: I originally had it checking only `plan` because I thought it would simplify the code, but looking back at it now, I guess it doesn't matter to go through all the args. 90% of the time this check would be moot anyway (since most people probably configure var args using `get_terraform_commands_that_need_vars`.

yorinasub17: Removed check: 295ec58eae93847070e42b16aa44c6b767d47416

cli/validate_inputs.go
yorinasub17: Yes this is different. `getTerraformInputNamesFromVarFiles` is extracting var files from `required_var_files` and `optional_var_files`, while this is extracting it from the raw args list.

Updated comments to clarify this: 70edaf281643441cbfa489b023c9b6cf1d078122

docs/_docs/04_reference/cli-options.md
yorinasub17: Done: 6c29b20530a3b8034fa426b1a0985deb3061733d

yorinasub17:
Thanks for review! Will merge and release!

PR #1575, opened 2021-03-06
add getTerragruntSource helper function

suhussai:


brikis98 (commented):
Thanks for the PR! The code changes look good overall, but could you please:

1. Describe the use case for this
2. Update the docs

Thanks!

suhussai:
Hi brikis98, thanks for reviewing!

Regarding your questions, I added the new function to the list of internal functions in the documentation and provided some information on what the function will return. I'm not sure what the use case for this might be. I'll ask in the original issue for more clarification.

config/config_helpers.go
```diff
@@ -115,6 +115,7 @@ func CreateTerragruntEvalContext(
 		"get_terraform_commands_that_need_input":       wrapStaticValueToStringSliceAsFuncImpl(TERRAFORM_COMMANDS_NEED_INPUT),
 		"get_terraform_commands_that_need_parallelism": wrapStaticValueToStringSliceAsFuncImpl(TERRAFORM_COMMANDS_NEED_PARALLELISM),
 		"sops_decrypt_file":                            wrapStringSliceToStringAsFuncImpl(sopsDecryptFile, extensions.Include, terragruntOptions),
+		"get_terragrunt_source":                        wrapVoidToStringAsFuncImpl(getTerragruntSource, extensions.Include, terragruntOptions),
```
yorinasub17, line 118: Perhaps we should rename this to `get_terragrunt_source_cli_flag` to make clear that this only returns values passed through with `--terragrunt-source`?

docs/_docs/04_reference/built-in-functions.md
```diff
@@ -610,3 +612,7 @@ inputs = merge(
   }
 )
 ```
+
+## get\_terragrunt\_source
+
+`get_terragrunt_source()` returns the location of the Terraform files. This location can be provided via CLI `--terragrunt-source`, an environment variable `TERRAGRUNT_SOURCE` or in the `terragrunt.hcl` file. Calling this function will return the value of that field.
```
yorinasub17, line 618:
```suggestion
`get_terragrunt_source()` returns the value passed in via the CLI `--terragrunt-source` or an environment variable `TERRAGRUNT_SOURCE`. Note that this will return an empty string when either of those values are not provided.

This is useful for constructing before and after hooks, or TF flags that only apply to local development (e.g., setting up debug flags, or adjusting the `iam_role` parameter).
```

yorinasub17:
@brikis98 

> Describe the use case for this

The main use case for this is to adjust the terragrunt configuration depending on whether its local development or not (`--terragrunt-source` was passed in). This came up in one of our support tickets, where a customer wanted to run `terraform fmt` as a before hook whenever you ran against local source during development. I can see other use cases for this as well:

- Setting debug logging when local dev.
- Adjusting the kubernetes provider configuration so that it targets minikube instead of real clusters.
- Providing special mocks pulled in from the local dev source (e.g., something like `mock_outputs = jsondecode(file("${get_terragrunt_source_cli_arg()}/dependency_mocks/vpc.json"))`).

brikis98:
> @brikis98
> 
> > Describe the use case for this
> 
> The main use case for this is to adjust the terragrunt configuration depending on whether its local development or not (`--terragrunt-source` was passed in). This came up in one of our support tickets, where a customer wanted to run `terraform fmt` as a before hook whenever you ran against local source during development. I can see other use cases for this as well:
> 
> * Setting debug logging when local dev.
> * Adjusting the kubernetes provider configuration so that it targets minikube instead of real clusters.
> * Providing special mocks pulled in from the local dev source (e.g., something like `mock_outputs = jsondecode(file("${get_terragrunt_source_cli_arg()}/dependency_mocks/vpc.json"))`).

Roger, thx for the explanation! May be worth adding to the docs!

suhussai:
Added use cases to the docs for `get_terragrunt_source_cli_flag()`.

Let me know if there's anything else I can add.

yorinasub17 (approved):
Updates LGTM! Will kick off the build momentarily.

yorinasub17:
Build passed, so going to merge this in! Thanks for your contribution!

PR #1576, opened 2021-03-07
feat: add get_terragrunt_git_dir built-in function

rymdo:
See [docs](https://github.com/rymdo/terragrunt/blob/899ed9c17006b91f50fbfaa03f48a0bbf1761177/docs/_docs/04_reference/built-in-functions.md#get_terragrunt_git_dir) for info

### DIFF
--- a/config/config_helpers.go
+++ b/config/config_helpers.go
@@ -156,9 +156,9 @@ func getTerragruntDir(include *IncludeConfig, terragruntOptions *options.Terragr
 
 // Return the directory in the git repoisitory where the Terragrunt configuration file lives
 func getTerragruntGitDir(include *IncludeConfig, terragruntOptions *options.TerragruntOptions) (string, error) {
-	args := []string {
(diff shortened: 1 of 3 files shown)


config/config_helpers.go
```diff
@@ -153,6 +154,27 @@ func getTerragruntDir(include *IncludeConfig, terragruntOptions *options.Terragr
 	return filepath.ToSlash(filepath.Dir(terragruntConfigFileAbsPath)), nil
 }
 
+// Return the directory in the git repoisitory where the Terragrunt configuration file lives
+func getTerragruntGitDir(include *IncludeConfig, terragruntOptions *options.TerragruntOptions) (string, error) {
+	args := []string{
+		"/bin/bash",
```
brikis98, line 160: Hm, this won't work on Windows...

brikis98 (commented):
Thanks for the PR! Could you describe the use case for this a bit more?

PR #1580, opened 2021-03-10
doc: contributing: fix broken link to circleci

marco-m-pix4d:
The only change is the circleci URL, but I also reformatted the lines to follow as much as possible the existing line length.


brikis98 (approved):
Thank you for the fix!

Issue #1045, opened 2020-02-13
Built-in function to retrieve terragrunt cache directory path?


sleungcy:
I am working on a project that can potentially have 100+ terragrunt modules.

I am looking into a way to modify or add to the terraform cached files before terraform plan, such that I can dry up the boilerplate that is in every module.

```
  before_hook "plan" {
    commands = ["plan"]
    execute  = ["/bin/bash", "${local.tg_hcls_root}/scripts/copy_common_tf.sh", local.tg_hcls_root, "."]
  }
```

Above is my first attempt, but I would prefer to have a more specfic "get-terragrunt-cache-dir" built-in function instead of using ".", what do you think?

yorinasub17:
This is one of the solutions to address https://github.com/gruntwork-io/terragrunt/issues/749, and has been generally considered as a usable function.

However, note that the implementation is non-trivial (I had attempted to implement it once) as you don't know the terragrunt cache directory unless you know the terraform module path, which means you need to:

- Do a partial parsing of the config to extract the terraform source.
- Compute the terragrunt cache dir, refactoring the code to ensure the same routine is used both within the helper and in terragrunt runtime; something that is annoying due to all the private functions and package organization.

yorinasub17:
FYI as mentioned in https://github.com/gruntwork-io/terragrunt/issues/749#issuecomment-586051840, https://github.com/gruntwork-io/terragrunt/pull/1050 might be an alternative.

lorengordon:
Looking for this also. In terragrunt <= 0.18, when the hooks ran in the cache directory, we could specify commands to run against files _in the remote source_ using a local/relative path. I.e. terragrunt would retrieve the remote source, change into the cache directory, and run hooks. So we could specify something like, `pip install requirements.txt && python render.py`, where both requirements.txt and render.py were maintained in the remote module. That no longer works, because as mentioned in #749 the hooks no longer operate out of the cache directory.

PR #1584, opened 2021-03-12
RDD: introduce ability to call terragrunt commands in before or after hooks

yorinasub17:
This is a README proposal of a feature to allow terragrunt to run terragrunt commands in a before or after hook. Opened early to get feedback.

## Use case

Our multiregion modules almost always require you to run `aws-provider-patch` prior to any `destroy` call, making the `destroy` experience very unfriendly. With this feature, we can work around this issue by registering the `aws-provider-patch` call as a `destroy` before hook so that it always runs anytime someone wants to destroy the multi region module.

This feature is necessary because, as the README update indicates, before hooks run in the terraform working directory, which means `terragrunt` commands fail. So you can't set `execute = ["terragrunt", "aws-provider-patch"]` to address this.

### DIFF
--- a/docs/_docs/04_reference/config-blocks-and-attributes.md
+++ b/docs/_docs/04_reference/config-blocks-and-attributes.md
@@ -52,12 +52,16 @@ The `terraform` block supports the following arguments:
 
 - `before_hook` (block): Nested blocks used to specify command hooks that should be run before `terraform` is called.
   Hooks run from the directory with the terraform module, except for hooks related to `terragrunt-read-config` and
-  `init-from-module`. These hooks run in the terragrunt configuration directory (the directory where `terragrunt.hcl`
-  lives).
+  `init-from-module`, and if `execute_terragrunt` is set. These hooks run in the terragrunt configuration directory (the
+  directory where `terragrunt.hcl` lives).
   Supports the following arguments:
     - `commands` (required) : A list of `terraform` sub commands for which the hook should run before.
     - `execute` (required) : A list of command and arguments that should be run as the hook. For example, if `execute` is set as
-      `["echo", "Foo"]`, the command `echo Foo` will be run.
+      `["echo", "Foo"]`, the command `echo Foo` will be run. Mutually exclusive with `execute_terragrunt`.
+    - `execute_terragrunt` (required) : A list of terragrunt command and arguments that should be run as the hook. Only
+      supports terragrunt specific commands other than `run-all` (e.g., `aws-provider-patch`). For example, if
+      `execute_terragrunt` is set as `["aws-provider-patch", "--terragrunt-override-attr", "region=us-east-2"]`,
+      terragrunt will run the `aws-provider-patch` command. Mutually exclusive with `execute`.
     - `run_on_error` (optional) : If set to true, this hook will run even if a previous hook hit an error, or in the
       case of "after" hooks, if the Terraform command hit an error. Default is false.
 


brikis98:
> Our multiregion modules almost always require you to run `aws-provider-patch` prior to any `destroy` call, making the `destroy` experience very unfriendly. With this feature, we can work around this issue by registering the `aws-provider-patch` call as a `destroy` before hook so that it always runs anytime someone wants to destroy the multi region module.

Could you remind me why? I vaguely remember seeing some discussion of this, but am struggling to find it.

> This feature is necessary because, as the README update indicates, before hooks run in the terraform working directory, which means `terragrunt` commands fail. So you can't set `execute = ["terragrunt", "aws-provider-patch"]` to address this.

Would a simpler alternative be to add support for a `working_directory` param to hooks? And a helper to return the `.terragrunt-cache` folder?

yorinasub17:
> Could you remind me why? I vaguely remember seeing some discussion of this, but am struggling to find it.

I don't exactly know why yet as I haven't had the time to dig into it. But I have seen it happen in practice 100% of the time when tearing down the architecture catalog based reference architecture.

I believe it has to do with the fact that we are nesting providers within modules that expose the parameters as variables (which we have to do to implement multiregion support), and terraform isn't really optimized for this use case (in fact, the docs recommend specifically against this pattern, where the provider definitions should always be at the top level module).

> Would a simpler alternative be to add support for a working_directory param to hooks?

I thought about this, but I was worried if this might introduce a "shoot yourself in the foot" situation as the terragrunt config gets parsed another time in a separate subprocess. The recursive terragrunt config parsing scares me, vs having internal logic in terragrunt to avoid another parse and rerouting the commands instead.

It is certainly going to be easier to maintain though.

> And a helper to return the .terragrunt-cache folder?

Note that this is non-trivial: https://github.com/gruntwork-io/terragrunt/issues/1045#issuecomment-585545515

brikis98:
> I don't exactly know why yet as I haven't had the time to dig into it. But I have seen it happen in practice 100% of the time when tearing down the architecture catalog based reference architecture.
> 
> I believe it has to do with the fact that we are nesting providers within modules that expose the parameters as variables (which we have to do to implement multiregion support), and terraform isn't really optimized for this use case (in fact, the docs recommend specifically against this pattern, where the provider definitions should always be at the top level module).

Hm, alright, sounds identical to the `import` bug... I didn't realize it affected `destroy` too. That's a bummer.

> > Would a simpler alternative be to add support for a working_directory param to hooks?
> 
> I thought about this, but I was worried if this might introduce a "shoot yourself in the foot" situation as the terragrunt config gets parsed another time in a separate subprocess. The recursive terragrunt config parsing scares me, vs having internal logic in terragrunt to avoid another parse and rerouting the commands instead.

Hm, worth trying with a real example to see what happens?

> > And a helper to return the .terragrunt-cache folder?
> 
> Note that this is non-trivial: [#1045 (comment)](https://github.com/gruntwork-io/terragrunt/issues/1045#issuecomment-585545515)

Is it comparable in time to refactoring the code to allow a hook that calls itself? Because if it is, it seems like we get more payoff doing this working and `working_directory`—two general purpose features that will pay off many times—than adding a "call yourself" special case...

yorinasub17:
> Is it comparable in time to refactoring the code to allow a hook that calls itself? Because if it is, it seems like we get more payoff doing this working and working_directory—two general purpose features that will pay off many times—than adding a "call yourself" special case...

FWIW, we don't need to implement this to address this usecase as we want to change the working directory to `get_terragrunt_dir()`, not the terragrunt-cache dir (which is the default).

But to answer your question, it's hard to say, but my guess is that it's going to be easier to refactor the hook call than the terragrunt cache dir function. I have a sense of how to implement this already (otherwise I wouldn't write this RDD), vs the terragrunt cache dir function I already attempted once and couldn't quite come up with a good way to implement it.

---

Given the above, I'll implement the `working_dir` function, try using that, and go from there. Even if it doesn't work out, it seems very useful to have anyway and it is low effort.

Thanks for the sanity checks!

brikis98:
> > Is it comparable in time to refactoring the code to allow a hook that calls itself? Because if it is, it seems like we get more payoff doing this working and working_directory—two general purpose features that will pay off many times—than adding a "call yourself" special case...
> 
> FWIW, we don't need to implement this to address this usecase as we want to change the working directory to `get_terragrunt_dir()`, not the terragrunt-cache dir (which is the default).
> 
> But to answer your question, it's hard to say, but my guess is that it's going to be easier to refactor the hook call than the terragrunt cache dir function. I have a sense of how to implement this already (otherwise I wouldn't write this RDD), vs the terragrunt cache dir function I already attempted once and couldn't quite come up with a good way to implement it.
> 
> Given the above, I'll implement the `working_dir` function, try using that, and go from there. Even if it doesn't work out, it seems very useful to have anyway and it is low effort.
> 
> Thanks for the sanity checks!

Roger on all points!

Issue #1551, opened 2021-02-17
Unexpected dependency behavior between stacks


cciotti:
terragrunt version: v0.28.2

I have two stacks, stackA and stackB and stackB declares a dependency to stackA. I tried to run `run-all apply` at the stackB level thinking it would see the dependency and run stackA. It did not. I get the following error.

> stackA/terragrunt.hcl is a dependency of stackB/terragrunt.hcl but detected no outputs. Either the target module has not been applied yet, or the module has no outputs. If this is expected, set the skip_outputs flag to true on the dependency block.

When I apply stackA first, stackB works just fine. My confusion is from the docs. I read [here](https://terragrunt.gruntwork.io/docs/features/execute-terraform-commands-on-multiple-modules-at-once/#dependencies-between-modules) the following.

> Once you’ve specified the dependencies in each terragrunt.hcl file, when you run the terragrunt run-all apply or terragrunt run-all destroy, Terragrunt will ensure that the dependencies are applied or destroyed, respectively, in the correct order. 

This is not happening for me. For reference, my dependencies are declared like this in stackB's terragrunt.hcl:

```
dependency "stackA" {
  config_path = "../stackA"
}

dependencies {
  paths = ["../stackA"]
}
```

Did I misread the docs? Is there something I need to configure?

Thank you.

yorinasub17:
`run-all` only runs the modules within the current directory tree. By default, it will warn you about external dependencies and interactively ask you if you want to include them or not. When you use `--terragrunt-non-interactive`, the default behavior is to exclude these external dependencies.

If you want to automatically run all external dependencies as well regardless of the directory tree, you can use the [--terragrunt-include-external-dependencies](https://terragrunt.gruntwork.io/docs/reference/cli-options/#terragrunt-include-external-dependencies) CLI option.

cciotti:
@yorinasub17 Thank you for clarifying. The doc I referenced was not clear on the cli option.

yorinasub17:
Closing this as the original question has been answered. If there are any follow ups, please open a new ticket with updated context. Thanks!

PR #1593, opened 2021-03-17
Clarify non-interactive will not include external dependencies

yorinasub17:
Addresses https://github.com/gruntwork-io/terragrunt/issues/1551

See https://github.com/gruntwork-io/terragrunt/blob/master/configstack/module.go#L383 for reference in code where this is happening.


infraredgirl (approved):

yorinasub17:

PR #1612, opened 2021-03-26
Use a single logger instance

boekkooi-fresh:
By using a single logger instance we avoid adding log entries to the logs written by `Stack.summarizePlanAllErrors`.

## Summary by CodeRabbit

- **Refactor**
  - Updated how logging context is handled, now extending the existing logger with additional information instead of creating a new instance.
  - Removed an unused logging utility function, streamlining internal logging behavior.

No user-facing features or visible changes in functionality.


brikis98:
Thanks for the PR! Could you provide a bit more context on what this does? What's the issue you're trying to fix? What does the log output look like now and what will it look like with this change?

yhakbar:
This has gone stale. I'm sorry about that @boekkooi-fresh .

We've done quite a bit of work on our logging, and we hope this is resolved.

Issue #1607, opened 2021-03-25
`validate-inputs` and `default = null` variables


lawliet89:
This isn't really a bug report but more of a question.

Consider the following variables:

```hcl
variable "cpu_limit" {
  description = "CPU limit for thing"
  default     = null
  type        = string
}

variable "memory_request" {
  description = "Memory Request for thing"
  default     = null
  type        = string
}
```

I am not sure if this is intended by Terraform core, but I can run plan/apply without providing these values ever. And `null` is meaningful in the usage of these variables and I would like to default them to `null`.

When I run `terragrunt validate-inputs`, terragrunt reports that these required inputs are missing. Is this an intended design, or simply a behavior inherited from using Terraform core code?

yorinasub17:
Yup this is something I ran into as well, and is a bug. I will be looking to fix this sometime this week or next.

lawliet89:
Hey @yorinasub17 thanks for confirming. I noticed that when I publish modules to Terraform Registry, it detects `default = null` as required variables too. Thus my question. But if it can be fixed in Terrragrunt, that's great!

yorinasub17:
Released as https://github.com/gruntwork-io/terragrunt/releases/tag/v0.28.18

PR #1613, opened 2021-03-26
Fix validate-inputs to support null defaults

yorinasub17:
Fixes https://github.com/gruntwork-io/terragrunt/issues/1607

Required updating `terraform-config-inspect` to latest, which has `Required` attribute on variables.


brikis98 (approved):
LGTM!

yorinasub17:
Thanks for review! Will merge this in.

Issue #1615, opened 2021-03-28
Add built-in function to `get_aws_account_alias` 


jones2026:
Accounts are often provisioned with several items named using the account alias and it would be helpful to have this helper be able to fetch this value for the terragrunt config. 

Another use case is naming your var files something like `my-account-name.tfvars` which helps identify which account/environment this file should be used for and is more human friendly than an account number. If terragrunt could look this up you could have it dynamically load the correct var file based on the current account the user is logged into.

Below is an example of using this name as the mechanism to load the correct `.tfvars` file:

``` hcl
terraform {
  extra_arguments "account_vars" {
    commands  = get_terraform_commands_that_need_vars()
    # With the get_aws_account_alias() function, you can load the correct var file for this account
    arguments = ["-var-file=${get_aws_account_alias()}.tfvars"]
  }
}
```
I am hoping to work on a PR for this soon and would love any comments and/or suggestions!

brikis98:
Thx for filing this issue and the PR. 

Taking a step back, I'm a little worried that adding more and more AWS-specific helpers will end up bloating the Terragrunt API, making it harder to use, and making it harder to maintain. The generic functions related to Terragrunt/Terratest (e.g., `get_terraform_commands_that_need_vars()` and `get_terragrunt_parent_dir()`) make sense to add, but the cloud-specific ones, such as `get_aws_account_id()`, may have been a mistake. They are certainly useful, but there is an infinite list of possibly useful such functions, and if we wanted to support similar functionality for every cloud (AWS, GCP, Azure), the complexity of using and maintaining the API might get out of hand quickly.

I'm tempted to put a cap on it now, and not add any new cloud-specific functions. Perhaps in the future, we even deprecate the existing ones. I believe [`run_cmd`](https://terragrunt.gruntwork.io/docs/reference/built-in-functions/#run_cmd) provides a reasonable workaround for these use cases, especially once we add memoization, as we're discussing in https://github.com/gruntwork-io/terragrunt/pull/1554. E.g.,

```hcl
locals {
  aws_account_id    = run_cmd("aws", "sts", "get-caller-identity", "--output", "text", "--query", "Account")
  aws_account_alias = run_cmd("aws", "iam", "list-account-aliases", "--output", "text", "--query", "AccountAliases")
}
```

That said, I'm open to other opinions. @yorinasub17 and @infraredgirl I'd value your input on this too!

jones2026:
[...]

PR #1616, opened 2021-03-28
feat: add built-in function to `get_aws_account_alias`

jones2026:
This is the PR for issue #1615 to add support for looking up the current account alias for AWS. I did not see any other unit tests for the other AWS built-in functions, but would be glad to figure out how to add some in for this if that would help.


jones2026:
closing this based on issue #1615 discussion.

Issue #1625, opened 2021-04-01
"Could not parse remote_state block from target config" when expressing dependency as of v0.28.0


jtackaberry:
I have a test project with the following structure:

```
├── lab
│   ├── nsg
│   │   └── terragrunt.hcl
│   ├── tfstate
│   │   └── terragrunt.hcl
│   ├── terragrunt.hcl
│   └── tfvars.hcl
└── modules
    ├── nsg
    │   └── main.tf
    └── tfstate
        └── main.tf
```

The project just creates an Azure blog storage container to hold tfstate, and an Azure Network Security Group, both in an existing subscription and resource group.

The nsg config declares a dependency on the tfstate config, to ensure appropriate ordering with run-all.  `lab/nsg/terragrunt.hcl` contains:

```
dependency "tfstate" {
  config_path = "${get_terragrunt_dir()}/../tfstate"
}
```

However, with this stanza, I get the following warning on running terragrunt commands against the nsg config:

```
WARN[0000] Could not parse remote_state block from target config /home/jtackaberry/tmp/tgtest/lab/tfstate/terragrunt.hcl
WARN[0000] Falling back to terragrunt output.
```

Indeed, this file doesn't contain a `remote_state` block.  That's defined in `lab/terragrunt.hcl`, which is included by the `nsg` config (and would be included in general by all non-`tfstate` config).

Near as I can tell, this warning is purely cosmetic -- everything seems to function otherwise.  It also didn't occur prior to v0.28.0.  The warning goes away if I remove the dependency block quoted above, but then of course `terragrunt run-all apply` wouldn't about the dependency and could apply the config in the wrong order.

Is this a bug, or am I expressing the dependency in the wrong way?

I am including the full `tgtest` project as an attachment here.  Thanks!

[tgtest.tar.gz](https://github.com/gruntwork-io/terragrunt/files/6240194/tgtest.tar.gz)

yorinasub17:
Apologies for the confusion here!

Based on what I see, it does look like you have everything configured correctly.

That warning is indeed benign and safe to ignore. It is indicating that the default dependency fetching optimization can not be done because terragrunt didn't detect the `remote_state` block. We initially set it at warning level so it is more visible, but looking back at it now, it probably should be info logs.

You can refer to the section "Can I speed up dependency fetching?" in [the `dependency` block documentation for more details](https://terragrunt.gruntwork.io/docs/reference/config-blocks-and-attributes/#dependency).

[...]

PR #1628, opened 2021-04-02
Switch log level of terragrunt dependency fetching informational messages

yorinasub17:
Related to https://github.com/gruntwork-io/terragrunt/issues/1625

This updates the informational messages related to dependency fetching optimizations to info log level from warning. I initially made these warning to make it more visible, but in hindsight these should have been `info` level because they are purely informational.

Warning indicates that something might be amiss and the user needs to make changes. In this case, there is probably nothing wrong here because there are many valid use cases of setting up terragrunt without `remote_state` blocks (e.g., if your project uses remote state that is not backed by `s3` or `gcs`). Thus info log level seems more appropriate for these messages.


brikis98 (approved):
Should it even be `Info`? Or `Debug`? That is, what percentage of users need to see this message at all, by default?

yorinasub17:
> Should it even be Info? Or Debug? That is, what percentage of users need to see this message at all, by default?

Good point - updated to use debug: 6d40576

brikis98 (approved):
LGTM!

yorinasub17:

Issue #1630, opened 2021-04-02
Make max retry attempts and sleep interval configurable


andreykaipov:
I saw from [your docs](https://terragrunt.gruntwork.io/docs/features/auto-retry/) this is something you'd be open to:

> Future upgrades to terragrunt may include the ability to configure max retries and retry intervals in the terragrunt config (PRs welcome!).

My corp network is absolute dogshit so even three retries is not enough to save our infrastructure, so I'd like to help!

I'll submit a PR later today but figured I'd open up this issue too for tracking purposes.

andreykaipov:
I was thinking of just adding two more top-level fields, e.g.:

```hcl
max_retry_attempts = 5
retry_sleep_interval = 10
retryable_errors = [
  "blah",
]
```

That feels like the most straight-forward approach, but lemme know if I should consider something else!

brikis98:
Thanks!

brikis98:
Fixed in #1632 and released in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.28.19.

PR #1632, opened 2021-04-02
Add configurable retries

andreykaipov:
Addresses https://github.com/gruntwork-io/terragrunt/issues/1630. Please lemme know what y'all think!

This PR adds two new options to configure our retries. For example, the following will try up to 10 times with 60 seconds in between each retry:

```hcl
retry_max_attempts = 10
retry_sleep_interval_sec = 60
retryable_errors = [
  "My own little error",
  "Another one of my errors",
]
```

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -391,20 +391,18 @@ func RunTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 		terragruntOptions.RetryableErrors = terragruntConfig.RetryableErrors
 	}
 
-	if val := terragruntConfig.RetryMaxAttempts; val != nil {
-		if *val < 1 {
-			terragruntOptions.Logger.Warnf("%s", "Cannot have less than 1 max retry, so we're using 1.")
-			*val = 1
+	if terragruntConfig.RetryMaxAttempts != nil {
+		if *terragruntConfig.RetryMaxAttempts < 1 {
+			return fmt.Errorf("Cannot have less than 1 max retry, but you specified %d", *terragruntConfig.RetryMaxAttempts)
 		}
-		terragruntOptions.RetryMaxAttempts = *val
+		terragruntOptions.RetryMaxAttempts = *terragruntConfig.RetryMaxAttempts
 	}
 
-	if val := terragruntConfig.RetrySleepInterval; val != nil {
-		if *val < 0 {
-			terragruntOptions.Logger.Warnf("%s", "Cannot sleep for less than 0 seconds, so we're using 0.")
-			*val = 0
+	if terragruntConfig.RetrySleepIntervalSec != nil {
+		if *terragruntConfig.RetrySleepIntervalSec < 0 {
+			return fmt.Errorf("Cannot sleep for less than 0 seconds, but you specified %d", *terragruntConfig.RetrySleepIntervalSec)
 		}
-		terragruntOptions.RetrySleepInterval = time.Duration(*terragruntConfig.RetrySleepInterval) * time.Second
+		terragruntOptions.RetrySleepIntervalSec = time.Duration(*terragruntConfig.RetrySleepIntervalSec) * time.Second
 	}
 
 	updatedTerragruntOptions := terragruntOptions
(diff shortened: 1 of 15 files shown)


cli/cli_app.go
```diff
@@ -391,6 +391,22 @@ func RunTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 		terragruntOptions.RetryableErrors = terragruntConfig.RetryableErrors
 	}
 
+	if val := terragruntConfig.RetryMaxAttempts; val != nil {
```
brikis98, line 394:
```suggestion
	if terragruntConfig.RetryMaxAttempts != nil {
```

cli/cli_app.go
```diff
@@ -391,6 +391,22 @@ func RunTerragrunt(terragruntOptions *options.TerragruntOptions) error {
...
+		if *val < 1 {
+			terragruntOptions.Logger.Warnf("%s", "Cannot have less than 1 max retry, so we're using 1.")
+			*val = 1
+		}
```
brikis98, line 398: This should be an error returned to the user.

cli/cli_app.go
```diff
@@ -391,6 +391,22 @@ func RunTerragrunt(terragruntOptions *options.TerragruntOptions) error {
...
+		terragruntOptions.RetryMaxAttempts = *val
+	}
+
+	if val := terragruntConfig.RetrySleepInterval; val != nil {
```
brikis98, line 402:
```suggestion
	if terragruntConfig.RetrySleepInterval != nil {
```

cli/cli_app.go
```diff
@@ -391,6 +391,22 @@ func RunTerragrunt(terragruntOptions *options.TerragruntOptions) error {
...
+		if *val < 0 {
+			terragruntOptions.Logger.Warnf("%s", "Cannot sleep for less than 0 seconds, so we're using 0.")
+			*val = 0
```
brikis98, line 405: This should be an error returned to the user.

config/config.go
```diff
@@ -67,6 +69,8 @@ type terragruntConfigFile struct {
 	TerragruntDependencies      []Dependency              `hcl:"dependency,block"`
 	GenerateBlocks              []terragruntGenerateBlock `hcl:"generate,block"`
 	RetryableErrors             []string                  `hcl:"retryable_errors,optional"`
+	RetryMaxAttempts            *int                      `hcl:"retry_max_attempts,optional"`
+	RetrySleepInterval          *int                      `hcl:"retry_sleep_interval,optional"`
```
brikis98, line 73: NIT: should it be `retry_sleep_interval_sec` to make the unit clear from the name?

docs/_docs/02_features/auto-retry.md
```diff
@@ -46,6 +44,12 @@ retryable_errors = [
 ]
 ```
 
-Future upgrades to `terragrunt` may include the ability to configure max retries and retry intervals in the `terragrunt` config (PRs welcome\!).
+By default, `auto-retry` tries a maximum of three times to re-run a command, pausing for five seconds between each retry, at which point it will deem the error as not transient, and accept the `terraform` failure.
+However, we can override these defaults too. For example, the following retries up to five times, with 60 seconds in between each retry:
```
brikis98, line 48:
```suggestion
However, you can override these defaults. For example, the following retries up to five times, with 60 seconds in between each retry:
```

test/integration_test.go
```diff
@@ -105,6 +105,7 @@ const (
 	TEST_FIXTURE_AUTO_RETRY_CUSTOM_ERRORS                   = "fixture-auto-retry/custom-errors"
 	TEST_FIXTURE_AUTO_RETRY_CUSTOM_ERRORS_NOT_SET           = "fixture-auto-retry/custom-errors-not-set"
 	TEST_FIXTURE_AUTO_RETRY_APPLY_ALL_RETRIES               = "fixture-auto-retry/apply-all"
+	TEST_FIXTURE_AUTO_RETRY_CONFIGURABLE_RETRIES            = "fixture-auto-retry/configurable-retries"
```
brikis98, line 108: Did you forget to check this in?

brikis98 (commented):
Thanks for the PR!

cli/cli_app.go
andreykaipov: I do this assignment to reduce the verbosity in the conditional since I have to check if the value is less than 1 to return the error to the user. I could remove it, but the conditional would look a bit more messy. Are you sure?

test/integration_test.go
andreykaipov: Woops you're right! I checked this in now along with another test for the error cases.

test/integration_test.go
```diff
@@ -105,6 +105,9 @@ const (
...
+	TEST_FIXTURE_AUTO_RETRY_CONFIGURABLE_RETRIES_ERROR_1    = "fixture-auto-retry/configurable-retries-incorrect-retry-attempts"
+	TEST_FIXTURE_AUTO_RETRY_CONFIGURABLE_RETRIES_ERROR_2    = "fixture-auto-retry/configurable-retries-incorrect-sleep-interval"
```
andreykaipov, line 110: I didn't want to make this a really long name and offset every other constant, but lemme know if y'all would like for the name to be more descriptive.

brikis98: Roger

brikis98 (approved):

brikis98:
Tests passed! Merging now.

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.28.19

andreykaipov:
Oh wow thank you so much for the quick release!!!

Issue #1301, opened 2020-08-18
read_terragrunt_config doesn't respect relative paths


dudicoco:
I have a dependency defined as follows:

```
dependency "eks-cluster" {
  config_path = "../../../devops/eks-ocean/eks-cluster"

  skip_outputs = true
}
```

When reading the dependency with `read_terragrunt_config`, the config path is relative to the file with the dependency block. I would expect it to be relative to the file reading the config.

yorinasub17:
This can go either way as far as understandability goes. I find it more intuitive that everything is relative to the file and not the calling context. I also understand that it might feel more intuitive if it was based on the original source dir. I also had folks ask if it can always be relative to the working dir. In this case, we chose to implement it this way as it is more intuitive if the logic is self contained in the config (you should be able to understand exactly what happens just by looking at the single config file).

---

Side note: there is a major bug with using dependencies in `read_terragrunt_config` and it is currently not recommended to do this. See https://github.com/gruntwork-io/terragrunt/issues/1128

dudicoco:
Thanks @yorinasub17. Maybe it's best to specify that it's currently not recommended to use in the docs?

yorinasub17:
Closing this as the original question has been answered. If there are any follow ups, please open a new ticket with updated context. Thanks!

Side note: we now have `dependency` sharing with deep merge in `include`, which should be a working alternative for this use case. That has support for the different `include` related functions, which should make it easier to build the relative paths.

PR #1637, opened 2021-04-09
Add new get_original_terragrunt_dir() helper

brikis98:
`get_original_terragrunt_dir()` returns the directory where the original Terragrunt configuration file (by default 
`terragrunt.hcl`) lives. This is primarily useful when one Terragrunt config is being read from another: e.g., if 
`/terraform-code/terragrunt.hcl` calls `read_terragrunt_config("/foo/bar.hcl")`, and within `bar.hcl`, you call 
`get_original_terragrunt_dir()`, you'll get back `/terraform-code`.

@yorinasub17 This was part of my hack day project. I'll post more info on this in Slack shortly.

**Note**: this PR originally was a one-liner that changed the behavior of `get_terragrunt_dir()`, but I decided that led to some unintuitive file path behavior in files read via `read_terragrunt_config()`, and that a separate, explicit `get_original_terragrunt_dir()` function handled this requirement more clearly.

### DIFF
--- a/cli/args.go
+++ b/cli/args.go
@@ -139,6 +139,8 @@ func parseTerragruntOptionsFromArgs(terragruntVersion string, args []string, wri
 		return nil, err
 	}
 
+	opts.OriginalTerragruntConfigPath = opts.TerragruntConfigPath
+
 	debug := parseBooleanArg(args, OPT_TERRAGRUNT_DEBUG, false)
 	if debug {
 		opts.Debug = true
--- a/config/config_helpers.go
+++ b/config/config_helpers.go
@@ -104,6 +104,7 @@ func CreateTerragruntEvalContext(
 		"read_terragrunt_config":                       readTerragruntConfigAsFuncImpl(terragruntOptions),
 		"get_platform":                                 wrapVoidToStringAsFuncImpl(getPlatform, extensions.Include, terragruntOptions),
 		"get_terragrunt_dir":                           wrapVoidToStringAsFuncImpl(getTerragruntDir, extensions.Include, terragruntOptions),
+		"get_original_terragrunt_dir":                  wrapVoidToStringAsFuncImpl(getOriginalTerragruntDir, extensions.Include, terragruntOptions),
 		"get_terraform_command":                        wrapVoidToStringAsFuncImpl(getTerraformCommand, extensions.Include, terragruntOptions),
 		"get_terraform_cli_args":                       wrapVoidToStringSliceAsFuncImpl(getTerraformCliArgs, extensions.Include, terragruntOptions),
 		"get_parent_terragrunt_dir":                    wrapVoidToStringAsFuncImpl(getParentTerragruntDir, extensions.Include, terragruntOptions),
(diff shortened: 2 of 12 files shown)


ghostinushanka:
@brikis98 Hi.
As a matter of unbelievable coincidence I was looking into the possibility to introduce the `get_child_terragrunt_dir()` literally for the past hour or so, saw that there was a context switching in `readTerragruntConfig` and decided to poke around the repo.

Can I suggest to unify the naming to `get_parent_terragrunt_dir()` and `get_child_terragrunt_dir()`, due to the fact that in documentation for `include` relation is explained in these terms?

yorinasub17 (approved):
LGTM!

brikis98:
> @brikis98 Hi.
> As a matter of unbelievable coincidence I was looking into the possibility to introduce the `get_child_terragrunt_dir()` literally for the past hour or so, saw that there was a context switching in `readTerragruntConfig` and decided to poke around the repo.
> 
> Can I suggest to unify the naming to `get_parent_terragrunt_dir()` and `get_child_terragrunt_dir()`, due to the fact that in documentation for `include` relation is explained in these terms?

Hm, I don't think this has anything to do with `include`... The change in this PR is solely related to usage with `read_terragrunt_config()`. Or am I misunderstanding you?

ghostinushanka:
Documentation is quite confusing for when you intend to both `include` and `read_terragrunt_config` same file from the module.
I was researching a way to create "global level variables", that is - "locals" in the "main" Terragrunt file that behave consistently when that file is "imported" both ways.

brikis98:
Ah, I gotcha. Well, FWIW, `get_original_terragrunt_dir` should behave as expected with `include` as well.

ghostinushanka:
Here's a specific example.
`terragrunt.hcl` (main)
```
locals {
  # will contain e.g. "tenants/company_name/dev/aws/aurora/auth"
  relative_path = trimprefix(
    get_child_terragrunt_dir(),
    get_parent_terragrunt_dir()
  )

  # will contain e.g. ["tenants", "company_name", "dev", "aurora", "auth"]
  rpl = split("/", local.relative_path)

  path_struct = [
    "tenant_home_dir",
    "tenant_name",
    "tenant_environment",
    "deployment_target",
    "tenant_module",
    "tenant_module_purpose",
  ]

  # will contain e.g.
  #  {
  #    "tenant_home_dir":       "tenants",
  #    "tenant_name":           "company_name",
  #    "tenant_environment":    "dev",
  #    "deployment_target":     "aws",
  #    "tenant_module":         "aurora",
  #    "tenant_module_purpose": "auth"
  #  }
  tenant_config = { for i in range(length(local.path_struct))
    : local.path_struct[i]
    => i < length(local.rpl) ? local.rpl[i] : ""
  }

# [...]
# global generate blocks, remote state etc
```

`terragrunt.hcl` (module)
```
include {
}

locals {
  main = read_terragrunt_config(find_in_parent_folders())["locals"]
}
```
This will allow to reuse e.g. `tenant_config` from main in the module without having to redo the whole operation again and will make sure that the contents will be the same in both the "main" and "module"

yorinasub17 (dismissed):
Update LGTM! Minor nit: consider adding a regression test for the `run-all` case.

brikis98:
> Update LGTM! Minor nit: consider adding a regression test for the `run-all` case.

Added: https://github.com/gruntwork-io/terragrunt/pull/1637/commits/c6fdac1c4ef19c38c8fe3feddf19cebe8b925f84

brikis98:
I also updated the way we `Clone` for dependencies, and added tests for same: https://github.com/gruntwork-io/terragrunt/pull/1637/commits/671ca7f593af1b4bc69f745b98d0394b3e963f85

yorinasub17 (approved):
Updates LGTM!

brikis98:

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.28.21

dudicoco:
@brikis98 thank you for the added functionality.

Does this resolve issues with relative paths when using `read_terragrunt_config`? See https://github.com/gruntwork-io/terragrunt/issues/1301.

Also, @yorinasub17 stated in the linked issue that there is a major bug with using dependencies in `read_terragrunt_config`, is that still the case?

brikis98:
> Does this resolve issues with relative paths when using `read_terragrunt_config`? See #1301.

The default behavior of `read_terragrunt_config` is unchanged, but you can use this new `get_original_terragrunt_dir()` as a way to change the behavior.

> Also, @yorinasub17 stated in the linked issue that there is a major bug with using dependencies in `read_terragrunt_config`, is that still the case?

That is also unchanged.

dudicoco:
Thanks @brikis98.

I've tested `get_original_terragrunt_dir` together with `read_terragrunt_config` and it did solve the relative paths issue.

Are there any plans on fixing the bug when using dependencies in `read_terragrunt_config`? 
The docs specify that `read_terragrunt_config`  can be used with dependency blocks, if that is not the case should the docs be adjusted accordingly?

brikis98:
Most likely plan is to implement `imports` at some point, but I'm not sure when yet: https://terragrunt.gruntwork.io/docs/rfc/imports/

dudicoco:
@brikis98 since it seems like `imports` won't happen any time soon (understandable due to limited time/resources), perhaps it would be possible to fix the issue with the current `read_terragrunt_config` feature?
Adding that fix would allow us to make the `terragrunt.hcl` files totally DRY across the stack.

brikis98:
It's hard to know without doing the work, but our current guess is that fixing the issue in `read_terragrunt_config` may actually be harder than implementing `imports`!

Issue #1636, opened 2021-04-08
run_cmd helper not receiving AWS_* env variables when using iam_role property


lmontanag:
Hi Everyone, hope everything is going well

I am using:

```hcl
iam_role = "arn:aws:iam::111111111111:role/iam_target_role"
```

As per documentation [here](https://terragrunt.gruntwork.io/docs/features/work-with-multiple-aws-accounts/)

So terragrunt will call `sts assume-role` before running terraform commands and that is working nice.  

I have a shell script that use `aws-cli` to get a value from aws ssm (parameter store) and I am using [`run_cmd`](https://terragrunt.gruntwork.io/docs/reference/built-in-functions/#run_cmd) to run that in terragrunt like:

```hcl
inputs = {
  certificate_body  = run_cmd("--terragrunt-quiet", "./get-aws-ssm-value.sh", "foo")
}

```

But I got the error: `An error occurred (ParameterNotFound) when calling the GetParameter operation:`

I have been trying to troubleshoot and it seems that for `run_cmd` the environment variables defined after assuming IAM role [here](https://github.com/gruntwork-io/terragrunt/blob/master/aws_helper/config.go#L188-L191)  are not passed as environment variables so `aws-cli` does not have valid AWS credentials.

not sure if I may be missing something, any help/clue, will be much appreciated 

thanks 👍🏼

brikis98:
Ah, yea, the docs aren't clear on this. I think currently, `iam_role` is only used when calling your actual Terraform command (e.g., `terraform apply`) and not anything else. That said, it would be more consistent for `iam_role` to affect everything. A PR to make this change is welcome.

lmontanag:
Closing this issue as it have been fixed , see comment here 

https://github.com/gruntwork-io/terragrunt/pull/1638#issuecomment-948714616

thanks so much for this 👍🏼

PR #1638, opened 2021-04-11
iam_role assume role before running helper functions.

lmontanag:
Hi Everyone, 

This one is related to https://github.com/gruntwork-io/terragrunt/issues/1636 , so the AWS credentials are available as environment variables for helper functions, for example, `run_cmd`.

thanks for your time and feedback 

👍🏼

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -346,6 +346,18 @@ func RunTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 		return err
 	}
 
+	if terragruntOptions.IamRole == "" {
+		terragruntFlags, err := config.PartialParseConfigFile(terragruntOptions.TerragruntConfigPath, terragruntOptions, nil, []config.PartialDecodeSectionType{config.TerragruntFlags})
+		if err != nil {
+			return err
+		}
+		terragruntOptions.IamRole = terragruntFlags.IamRole
+	}
+
+	if err := aws_helper.AssumeRoleAndUpdateEnvIfNecessary(terragruntOptions); err != nil {
+		return err
+	}
+
 	terragruntConfig, err := config.ReadTerragruntConfig(terragruntOptions)
 	if err != nil {
 		return err
@@ -364,14 +376,6 @@ func RunTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 			terragruntOptions.TerragruntConfigPath,
 		)
 		return nil
-	}
-
-	if terragruntOptions.IamRole == "" {
-		terragruntOptions.IamRole = terragruntConfig.IamRole
-	}
-
-	if err := aws_helper.AssumeRoleAndUpdateEnvIfNecessary(terragruntOptions); err != nil {
-		return err
 	}
 
 	// get the default download dir


cli/cli_app.go
```diff
@@ -346,6 +346,18 @@ func RunTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 		return err
 	}
 
+	if terragruntOptions.IamRole == "" {
+		terragruntFlags, err := config.PartialParseConfigFile(terragruntOptions.TerragruntConfigPath, terragruntOptions, nil, []config.PartialDecodeSectionType{config.TerragruntFlags})
```
brikis98, line 350: I'm having trouble thinking through the implications of doing a partial parsing here... E.g.,

1. What if the IAM role ARN to assume is read from a `dependency` block. Will that work?
1. What about an IAM role ARN read via `read_terragrunt_config`?
1. What about an `iam_role` param in an `include` block? Will that work?

@yorinasub17 Would appreciate another set of eyes on this.

brikis98 (commented):
Thanks for the PR!

We'll def want at least a basic regression test for this. See `integration_test.go` for how we write these.

cli/cli_app.go
lmontanag: Thanks for your feedback @brikis98 , really appreciate it 👍🏼 , I was trying to get something for the regression tests on this, but seems there is no test for the `iam_role` functionality, according to  this @yorinasub17 PR [here](https://github.com/gruntwork-io/terragrunt/pull/1315) 
> NOTE: I didn't add tests because it doesn't look like we have tests for this functionality, but we should figure out a way to have a test. In the meantime, I am going to cut an alpha testing release so that the reporter can test this.

The idea was to assume the aws iam role as soon as possible and affect everything, so the AWS credentials will be available for other subcommands for instance `run_cmd`, I don't have the broad `terragrunt` picture and flow as clear as you may have guys , so any clue / feedback will be much apreaciated 

more info here https://github.com/gruntwork-io/terragrunt/issues/1636

lmontanag: Hi @yorinasub17 , sorry to bother you, but I was wondering if you could give me some hints/clues,  about how could I implement regression tests for this scenario,  I have seen the `integration_test.go` file but could not come up with a good approach for that. 

Any idea will be much appreciated.

thanks for your time 👍🏼

lmontanag: Hi @yorinasub17 , no problem, thanks so much for your time and ideas!, I have added the regression tests aligned with your recommendations, could you please have a look 👍🏼  

```bash
go test  -v -run ^TestTerragruntIamRoleProperty$
```

not sure though about case 1:

> What if the IAM role ARN to assume is read from a dependency block. Will that work?

any idea about how to tackle that one?

yorinasub17:
This has recently been implemented by another PR, so will close. Thanks for the contribution and apologies we didn't review this in a timely fashion.

lmontanag:
Thanks so much for this @yorinasub17 👍🏼  , really apreciate all the work you are doing with terragrunt. have a nice week

PR #1639, opened 2021-04-12
Allow generate and remote_state to be set as attrs

brikis98:
This allows us to configure code generation and remote state dynamically. See the documentation updates for examples.


config/config.go
```diff
@@ -777,7 +826,26 @@ func convertToTerragruntConfig(
 		terragruntConfig.IamRole = *terragruntConfigFromFile.IamRole
 	}
 
-	for _, block := range terragruntConfigFromFile.GenerateBlocks {
+	generateBlocks := []terragruntGenerateBlock{}
+	generateBlocks = append(generateBlocks, terragruntConfigFromFile.GenerateBlocks...)
+
+	if terragruntConfigFromFile.GenerateAttrs != nil {
+		remoteStateMap, err := parseCtyValueToMap(*terragruntConfigFromFile.GenerateAttrs)
```
yorinasub17, line 833:
```suggestion
		generateMap, err := parseCtyValueToMap(*terragruntConfigFromFile.GenerateAttrs)
```

config/config.go
```diff
@@ -777,7 +826,26 @@ func convertToTerragruntConfig(
...
+		if err != nil {
+			return nil, err
+		}
+
+		for name, block := range remoteStateMap {
```
yorinasub17, line 838:
```suggestion
		for name, block := range generateMap {
```

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -225,6 +225,34 @@ The `remote_state` block supports the following arguments:
...
+  remote_state = {
+    backend = "s3"
+    config = {
+      bucket = "mybucket"
+      key    = "path/to/my/key"
+      region = "us-east-1"
+    }
+  }
+}
+```
```
yorinasub17, line 242: Side note: It isn't strictly necessary to set this as a local, as `read_terragrunt_config` will expose the `remote_state` block under the `remote_state` attribute.

yorinasub17 (commented):
LGTM! Only had some style nit comments

config/config.go
brikis98: Whoops, good catch. Fixed in https://github.com/gruntwork-io/terragrunt/pull/1639/commits/b6f61c8300610841e5f4ae4b466f5d76490c0374.

config/config.go
brikis98: Whoops, good catch. Fixed in https://github.com/gruntwork-io/terragrunt/pull/1639/commits/b6f61c8300610841e5f4ae4b466f5d76490c0374.

docs/_docs/04_reference/config-blocks-and-attributes.md
brikis98: Oh, interesting!

What about `generate` blocks? Is there a way to access those from what `read_terragrunt_config` returns?

yorinasub17: Generate is also available under `generate` attribute as a map. See https://github.com/gruntwork-io/terragrunt/blob/master/test/integration_test.go#L3043

yorinasub17 (approved):

docs/_docs/04_reference/config-blocks-and-attributes.md
brikis98: Oh, very handy. OK, docs updated: https://github.com/gruntwork-io/terragrunt/pull/1639/commits/675fc22446876104c2ccab49523c1c6e4020d10c.

brikis98:
Thx for the reviews! Could I get one more ship it?

yorinasub17 (approved):
LGTM! I didn't think the native formats were better, but then I realized you get type checking that way so it's good! 👍

brikis98:
> LGTM! I didn't think the native formats were better, but then I realized you get type checking that way so it's good! 👍

Yea, it's both cleaner to read, and safer. Nice call.

Thx for the reviews. Merging now.

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.28.21

Issue #1640, opened 2021-04-12
Terragrunt evaluating all submodules for locals with `--terragrunt-strict-include`


celestialorb:
Not sure if Terragrunt needs to do this or if this is intended design, but it seems that when using `run-all` in combination with `--terragrunt-strict-include` and `--terragrunt-include-dir` Terragrunt will still iterate over every submodule to evaluate locals. This, in my case, resulted in Terragrunt downloading more providers than expected. While this isn't exactly a problem, I found it to be quite confusing.

My setup is that I have a module (A) that I wanted to run `init` on via `run-all init` with `--terragrunt-strict-include` and the appropriate `--terragrunt-include-dir` flag, and another module elsewhere (B) that reads in an extra HCL file via `read_terragrunt_config` into a local variable; and this HCL file includes a dependency to another module (C). Module (A) has no dependencies, but in order to get the outputs of module C for the locals of module B, Terragrunt has to initialize module C, causing it to pull down the providers required for it. Thus Terragrunt downloads the extra providers for module C during the initial locals evaluation of a `run-all` command.

The behavior I would've expected here is that there would be no evaluation of locals for any modules outside of those specified by the `--terragrunt-include-dir` flags.

Is this intended / necessary, or could this be considered a bug?

celestialorb:
I have crafted a basic PR to address this, let me know if the approach taken in it is viable, or if there is another approach / option that should be considered here.

Issue #1600, opened 2021-03-20
`--terragrunt-strict-include` is not strict when no include dirs are specified


celestialorb:
Correct me if this is the intended design for `strict-include`, but I would've assumed that if I had specified `--terragrunt-strict-include` and no accompanying `--terragrunt-include-dir` flags that it would not run on any modules and would return successfully.

I've found that when I run with `--terragrunt-strict-include` and no `--terragrunt-include-dir` flags that Terragrunt acts as if the `--terragrunt-strict-include` flag does not exist and proceeds to operate over all modules under the working directory.

In my problematic command I'm only specifying `--terragrunt-working-dir` and `--terragrunt-strict-include`.

Example:
`terragrunt run-all validate --terragrunt-working-dir <account> --terragrunt-strict-include`

brikis98:
[`--terragrunt-strict-include`](https://terragrunt.gruntwork.io/docs/reference/cli-options/#terragrunt-strict-include) only modifies the behavior of `--terragrunt-include-dir`. If you don't specify the latter, I believe the former has no effect.

celestialorb:
The Terragrunt documentation seems to indicate that it would be the other way around, or at least that's how I interpret it:
> When passed in, only modules under the directories passed in with `–terragrunt-include-dir` will be included.

Could we consider changing the behavior of the flag?

brikis98:
Changing the behavior to what?

celestialorb:
To not run on any directories when `--terragrunt-strict-include` is supplied without any `--terragrunt-include-dir` flags.

brikis98:
On re-reading the description, I think you're right; that behavior is probably the more expected one. We're buried at the moment, so if someone has the time to make this update, we'd very much welcome the PR!

celestialorb:
I can take a stab at it this weekend, seems like it'd be fairly straight-forward from a quick glance at the codebase. :)

celestialorb:
Basic PR to address this is ready for review.

brikis98:
Fixed in #1631 and released in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.28.20.

Issue #1033, opened 2020-02-05
--terragrunt-exclude-dir not working?


jcalvento:
I have the following structure
```
root
|- commons (should not read this directly)
     | - *.tf
     | - *.hcl
|-dev
     | - parent.hcl
     | - commons
|-test
     | - parent.hcl
     | - commons
```
I'd like to run a plan-all on the root folder but when I try to exclude commons dir it reads it anyway
`terragrunt plan-all --terragrunt-exclude-dir ./commons`
I'm getting 
```
 Error processing module at '.../root/commons/dynamodb/items/configs/terragrunt.hcl'. How this module was found: Terragrunt config file found in a subdirectory of .../root/. Underlying error: .../root/commons/dynamodb/items/configs/terragrunt.hcl:9,10-33: Error in function call; Call to function "find_in_parent_folders" failed: ParentFileNotFound: Could not find a terragrunt.hcl in any of the parent folders of .../root/commons/dynamodb/items/configs/terragrunt.hcl. Cause: Traversed all the way to the root.., and 1 other diagnostic(s)
```

yorinasub17:
Can you share the full log output up to the interactive prompt asking if you would like to run this on a list of modules?

jcalvento:
@yorinasub17 Sorry for the late response but that's all the logs I have. I'm not getting any kind of confirmation to run this

yorinasub17:
Ah ok so it is failing in the partial parsing of the configuration. Because the include and exclude functionality still processes the dependencies, terragrunt needs to read just enough of the configuration in ALL config that it finds in the current working directory to build the tree, before proceeding to process include and excludes. So you need to have a complete, valid config for those components even for those directories that are excluded.

As far as what components need to be valid, see https://terragrunt.gruntwork.io/docs/getting-started/configuration/#configuration-parsing-order, specifically the section about `-all`.

lawliet89:
I'm also facing the same issue.

Perhaps there should be a new option to ask Terragrunt to simply ignore some directories when finding modules to run?

yorinasub17:
> Perhaps there should be a new option to ask Terragrunt to simply ignore some directories when finding modules to run?

I am weary of adding additional config options here. I think real solution here would be to fix the processing so that `terragrunt` processes includes and excludes prior to reading the config. We're a bit buried to implement this, but a PR to fix that implementation would be much appreciated!

yorinasub17:
[...]

PR #1644, opened 2021-04-17
Restrict Locals Evaluation with Include Dirs

celestialorb:
This PR is an attempt to address #1640.

We have a Terragrunt project that has a rather large number of modules (close to the order of 100) and we've noticed that Terragrunt is very slow to start when using the `run-all` command. This was discovered to be due to the fact that Terragrunt is running over all modules in the working directory and evaluating their locals, which was causing a few modules to be initialized during this phase as some locals were/are referencing dependencies.

It was discovered that this behavior is the same even when specifying `--terragrunt-strict-include`. It is my belief that Terragrunt shouldn't need to evaluate locals of any modules outside of those specified in the `--terragrunt-include-dir` flags (and subsequent module dependencies).

This PR attempts to resolve this by changing the implementation of `FindConfigFIlesInPath` under the `config` package, which originally walks over the entire working directory looking for Terragrunt modules. The implementation has been changed to only walk over the directories specified by `--terragrunt-include-dir` flags. Two tests have been added to the test suite to test these changes.

I have verified that all tests pass under the `config` package with these changes, and I have also tested the build with these changes on our large Terragrunt project and can see significant startup time improvement.


brikis98:
@yorinasub17 I think you may be best to review this one.

celestialorb:
Believe I have found a small issue with the code in this PR, will take a look a bit later and see if I can address it. Reverting this PR to draft/WIP status for now.

yorinasub17:
This implementation makes sense to me, and would also provide a workaround for https://github.com/gruntwork-io/terragrunt/issues/1033

celestialorb:
@yorinasub17 This is ready for a formal review again. I was right about the small issue I mentioned previously in that this original implementation did not account for globs in the `--terragrunt-include-dir` flags. I have added another small test case and re-implemented `FindInConfigFiles` with the use of `zglob.Glob` to account for this.

A small oddity I encountered is that the ordering of the file paths returned by `zglob.Glob` didn't seem to be determinate, so I added a sort function to the end of `FindInConfigFiles`, which meant that I had to reorder the expected results of a few test cases. I'm assuming the ordering of the file paths returned by `FindInConfigFiles` has no effect and thus does not matter.

config/config.go
```diff
@@ -351,29 +354,60 @@ func GetDefaultConfigPath(workingDir string) string {
...
-		}
+		directories = append(directories, matches...)
+	}
 
-		isTerragruntModule, err := containsTerragruntModule(path, info, terragruntOptions)
-		if err != nil {
-			return err
-		}
+	// TODO: since the terragrunt include directories can be specified as globs, we need to expand
+	// those globs before walking over them.
```
yorinasub17, line 374: This TODO is now addressed in the for loop above right? If so, we should remove this.

yorinasub17 (approved):
Implementation LGTM! Only had one nit about a stale comment. Once that is addressed, and if the build passes, we can merge this in!

yorinasub17:
Ah looks like the build failed for `TestIncludeDirsStrict` (in `test` folder). Here are the test logs:

```
%~> go test -v -run TestIncludeDirsStrict$ .                                                                                                                                                                    [0]
=== RUN   TestIncludeDirsStrict
=== PAUSE TestIncludeDirsStrict
=== CONT  TestIncludeDirsStrict
    integration_test.go:3251: [TestIncludeDirsStrict] Full contents of validate-all stdout:
    integration_test.go:3254: [TestIncludeDirsStrict]
    integration_test.go:3251: [TestIncludeDirsStrict] Full contents of validate-all stderr:
    integration_test.go:3254: [TestIncludeDirsStrict]
    integration_test.go:4089:
                Error Trace:    integration_test.go:4089
                                                        integration_test.go:1964
                Error:          Received unexpected error:
                                Could not find any subfolders with Terragrunt configuration files
                Test:           TestIncludeDirsStrict
--- FAIL: TestIncludeDirsStrict (0.00s)
FAIL
FAIL    github.com/gruntwork-io/terragrunt/test 2.781s
FAIL
```

config/config.go
celestialorb: Ah yes, forgot to take that out. Good catch.

celestialorb:
> Ah looks like the build failed for `TestIncludeDirsStrict` (in `test` folder). Here are the test logs:
> 
> ```
> %~> go test -v -run TestIncludeDirsStrict$ .                                                                                                                                                                    [0]
> === RUN   TestIncludeDirsStrict
> === PAUSE TestIncludeDirsStrict
> === CONT  TestIncludeDirsStrict
>     integration_test.go:3251: [TestIncludeDirsStrict] Full contents of validate-all stdout:
>     integration_test.go:3254: [TestIncludeDirsStrict]
>     integration_test.go:3251: [TestIncludeDirsStrict] Full contents of validate-all stderr:
>     integration_test.go:3254: [TestIncludeDirsStrict]
>     integration_test.go:4089:
>                 Error Trace:    integration_test.go:4089
>                                                         integration_test.go:1964
>                 Error:          Received unexpected error:
>                                 Could not find any subfolders with Terragrunt configuration files
>                 Test:           TestIncludeDirsStrict
> --- FAIL: TestIncludeDirsStrict (0.00s)
> FAIL
> FAIL    github.com/gruntwork-io/terragrunt/test 2.781s
> FAIL
> ```

Seems this is stemming from this line:
```
includedModulesWithNone := runValidateAllWithIncludeAndGetIncludedModules(t, testPath, []string{}, true)
```

which is running a validate all with strict include and no given modules, and is thus expecting no modules to be returned...however that test is also checking to ensure that Terragrunt does not throw any errors; which it will in this case now since it's not detecting any Terragrunt configurations in `FindInConfigFiles` anymore.

I think the best course of action here might be to change that error to warning since it should be expected to not find any Terragrunt configuration files with strict include on and no include directories specified.

yorinasub17:
Hmm I think the error message is actually correct. I would want terragrunt to give me a non-zero exit code when it finds no terragrunt.hcl config that it should run on.

Can you actually just remove that test condition? I think using strict-include without include-dir should be considered an error given the intention of strict-include, so we shouldn't be testing that condition.

celestialorb:
> Hmm I think the error message is actually correct. I would want terragrunt to give me a non-zero exit code when it finds no terragrunt.hcl config that it should run on.
> 
> Can you actually just remove that test condition? I think using strict-include without include-dir should be considered an error given the intention of strict-include, so we shouldn't be testing that condition.

What about cases where a script or CI system might generate a Terragrunt command with strict include based off of a set of changed modules? That set might be empty and could generate and run a Terragrunt command with `--terragrunt-strict-include` and no `--terragrunt-include-dir` flags.

yorinasub17:
Personally, I think the CI system should handle the case where there are no directories and not let terragrunt figure it out. That's a really simple array count check in bash, or whatever language it's in.

In fact, I would argue that terragrunt should really be erroring much earlier with a clear error message if `--terragrunt-strict-include` was passed in without `--terragrunt-include-dir`.

As a user, I would want terragrunt to complain to me that `--terragrunt-strict-include` expects `--terragrunt-include-dir` and that it's undefined behavior if you pass in `--terragrunt-strict-include` without the other.

celestialorb:
> Personally, I think the CI system or script should handle the case where there are no directories and not let terragrunt figure it out. That's a really simple array count check in bash, or whatever language it's in.
> 
> In fact, I would argue that terragrunt should really be erroring much earlier with a clear error message if `--terragrunt-strict-include` was passed in without `--terragrunt-include-dir`.
> 
> As a user, I would want terragrunt to complain to me that `--terragrunt-strict-include` expects `--terragrunt-include-dir` and that it's undefined behavior if you pass in `--terragrunt-strict-include` without the other.

You might be interested in taking a look at #1600 then.

yorinasub17:
Ah you got me. I missed that, and if @brikis98 made that decision, then I'm fine going forward with this.

I would still argue that this should be handled early in the pipeline though, as a special check in https://github.com/gruntwork-io/terragrunt/blob/f99b3fc2fb3ac7e4e5b5046fa126c80f15a4bc87/configstack/stack.go#L114 . That is probably the cleanest, and would also avoid parsing a whole bunch of stuff.

rdettai:
I am also interested to see this merged, is there a specific reason why it is stuck for over a year?

celestialorb:
> I am also interested to see this merged, is there a specific reason why it is stuck for over a year?

Shortly after this my team ditched Terragrunt in favor of plain Terraform, and thus I lost motivation to follow through on this.

mfulgo:
@yorinasub17 - If one wanted to get this merged, at this point is it just a matter of resolving conflicts and making sure the existing tests pass?

yhakbar:
> @ yorinasub17 - If one wanted to get this merged, at this point is it just a matter of resolving conflicts and making sure the existing tests pass?

Please make sure that you don't tag anyone that isn't currently a [CODEOWNER](https://github.com/gruntwork-io/terragrunt/blob/main/CODEOWNERS). I want to make sure we aren't bugging anyone that isn't currently responsible for maintaining the project.

@mfulgo Please do resolve conflicts and make sure tests pass. After that, please evaluate whether this pull request is necessary, given that it was opened quite a while ago.

I've moved the pull request to in-draft, if you can convince the author to do these things, please encourage them to mark the PR as ready for review afterwards too (or do all this in a new pull request you start, etc).

PR #1651, opened 2021-04-22
Update to Terraform 0.15

brikis98:
This PR updates the Terragrunt build to run with Terraform 0.15.

### DIFF
--- a/cli/args_test.go
+++ b/cli/args_test.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 	"testing"
 
 	"strings"
@@ -373,18 +374,19 @@ func TestTerraformHelp(t *testing.T) {
 		args     []string
 		expected string
 	}{
-		{[]string{"terragrunt", "plan", "--help"}, "Usage: terraform plan"},
-		{[]string{"terragrunt", "apply", "-help"}, "Usage: terraform apply"},
-		{[]string{"terragrunt", "apply", "-h"}, "Usage: terraform apply"},
+		{[]string{"terragrunt", "plan", "--help"}, "Usage: terraform .* plan"},
+		{[]string{"terragrunt", "apply", "-help"}, "Usage: terraform .* apply"},
+		{[]string{"terragrunt", "apply", "-h"}, "Usage: terraform .* apply"},
 	}
 
 	for _, testCase := range testCases {
 		err := app.Run(testCase.args)
-
 		require.NoError(t, err)
-		if !strings.Contains(output.String(), testCase.expected) {
-			t.Errorf("expected output to include help text; got stdout: %v.", output.String())
-		}
+
+		expectedRegex, err := regexp.Compile(testCase.expected)
+		require.NoError(t, err)
+
+		assert.Regexp(t, expectedRegex, output.String())
 	}
 }
 
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -522,44 +522,6 @@ func TestTerragruntWorksWithLocalTerraformVersion(t *testing.T) {
 	validateDynamoDBTableExistsAndIsTagged(t, TERRAFORM_REMOTE_STATE_S3_REGION, lockTableName, expectedDynamoDBTableTags)
 }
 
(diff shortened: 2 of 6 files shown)


zackproser (approved):
LGTM pending tests!

cli/args_test.go
```diff
@@ -373,18 +374,19 @@ func TestTerraformHelp(t *testing.T) {
 		args     []string
 		expected string
 	}{
-		{[]string{"terragrunt", "plan", "--help"}, "Usage: terraform plan"},
-		{[]string{"terragrunt", "apply", "-help"}, "Usage: terraform apply"},
-		{[]string{"terragrunt", "apply", "-h"}, "Usage: terraform apply"},
+		{[]string{"terragrunt", "plan", "--help"}, "Usage: terraform .* plan"},
+		{[]string{"terragrunt", "apply", "-help"}, "Usage: terraform .* apply"},
+		{[]string{"terragrunt", "apply", "-h"}, "Usage: terraform .* apply"},
```
brikis98, line 379: Terraform's help text is a bit different now, so had to update this test accordingly.

test/integration_test.go
```diff
@@ -522,44 +522,6 @@ func TestTerragruntWorksWithLocalTerraformVersion(t *testing.T) {
 	validateDynamoDBTableExistsAndIsTagged(t, TERRAFORM_REMOTE_STATE_S3_REGION, lockTableName, expectedDynamoDBTableTags)
 }
 
-// Regression test to ensure that skip flags associated with the S3 bucket autocreation feature. At one point,
-// terragrunt was stripping the flags in a way that it was no longer being honored.
-// This bug only happens if terragrunt calls remotestate.NeedsInitialization on the s3 initializer, so we need to setup
-// a situation where terragrunt calls this. To do this, we first setup the module with local backend (so that
-// moduleNeedsInit is false), before switching to the remote backend.
-func TestTerragruntHonorsS3RemoteStateSkipFlagsRegression(t *testing.T) {
```
brikis98, line 530: This test runs Terragrunt first with a local backend and then with an S3 backend. Unfortunately, with TF 0.15, when you switch to the S3 backend, it shows an interactive prompt asking if you want to migrate state over from the local backend. This interactive prompt now causes the test to fail. I tried switching the second Terragrunt run from `validate` to `apply -input=false`, but the interactive prompt still came up.

Not sure how this worked before 0.15, but as per @yorinasub17's [comment here](https://github.com/gruntwork-io/terragrunt/pull/1399#issuecomment-714774114), this test doesn't quite test what we are looking for anyway, and give the complexity of working around an interactive prompt, I opted to remove it.

test/integration_test.go
```diff
@@ -586,13 +546,7 @@ func TestTerragruntSetsAccessLoggingForTfSTateS3BuckeToADifferentBucketWithGiven
 		lockTableName,
 		"remote_terragrunt.hcl",
 	)
-	localBackendTerragruntConfigPath := filepath.Join(examplePath, "local_terragrunt.hcl")
-
-	// Pass 1 with local backend. This should only download and setup the terragrunt-cache. Note that the first pass has
-	// to be an apply so that the state file is created. Otherwise, terragrunt short circuits the problematic routine.
-	runTerragrunt(t, fmt.Sprintf("terragrunt apply -auto-approve --terragrunt-non-interactive --terragrunt-config %s --terragrunt-working-dir %s", localBackendTerragruntConfigPath, examplePath))
```
brikis98, line 593: This test case and the one below look like the copy/pasted the logic from `TestTerragruntHonorsS3RemoteStateSkipFlagsRegression` around running first with a local backend and then with an S3 backend, so they hit the same interactive prompt. However, I don't think these two test cases needed this two-step process, so I updated them to _only_ run with the S3 backend, which seems to work.

brikis98:
Woot, tests are passing!

zackproser (approved):
LGTM

brikis98:

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.29.0

lorengordon:
@brikis98 seems like the binaries should have dropped by now... did the CI fail when building them again?

brikis98:
Gah, yea, there's a timing issue in one test that fails now and then. Just re-ran the build, binaries should show up shortly. Thanks for the heads up!

PR #1663, opened 2021-04-30
Honor the shared config in the default session

yorinasub17:
The default session that we construct in terragrunt does not configure the shared config enabled bit, so it ignores `~/.aws/config`. This breaks the functions that use this path.


brikis98 (approved):
LGTM

yorinasub17:

Issue #1655, opened 2021-04-23
Consider publishing a terragrunt container to dockerhub?


lorengordon:
Would you consider hosting a Dockerfile and publishing a terragrunt container to dockerhub? I could probably do an initial PR for a Dockerfile, but someone from Gruntwork would need to setup the Dockerhub builds...

(There are lots of services for building and publishing containers, but the easiest I've found is just to use the [Dockerhub automated builds](https://docs.docker.com/docker-hub/builds/), configured to trigger on branch updates to create a "latest" image and on tags to created tagged images.)

brikis98:
In the future, yes, definitely. But right now, we're too buried to do the work to set up the CI auth, publishing, etc.

lorengordon:
I can open a PR for a Dockerfile at least. And I'll post a screenshot of how I setup Dockerhub builds. It was super easy, maybe 10 minutes of work.

I would suggest including *only* terragrunt, and let the user provide their own terraform (perhaps with a multi-stage build, for example). Otherwise, the image would have to tie a specific terraform version to the terragrunt container, or use a tool like tfenv to auto-install/select a specific version....

lorengordon:
Here's the dockerhub build settings I've used for another project. It builds and publishes the `latest` image on any push to the master branch, and version-tagged images whenever a GitHub tag is created.

![image](https://user-images.githubusercontent.com/8457307/116925664-50cc0680-ac0e-11eb-99dd-c564afde1c8e.png)

yhakbar:
We already publish a docker image that includes an installation of Terragrunt, but we'd like to change the image in the future so that it is actually just a from scratch image with Terragrunt installed.

https://hub.docker.com/r/gruntwork/terragrunt

We can't prioritize this right now, but I'll mark this as preserved so we revisit this in the future. It shouldn't be much work.

PR #1665, opened 2021-05-03
Adds a Dockerfile that creates a terragrunt image

lorengordon:
Related to #1655 

## Summary by CodeRabbit

- **New Features**
  - Added automated building and verification of a Docker container for the application, triggered on new version tags.
  - Introduced a Dockerfile to package the application as a container image.

- **Chores**
  - Updated CI workflows to include container build and version verification steps.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+FROM golang:1.16.3-alpine as builder
+
+RUN apk add --update --no-cache make git
+
+WORKDIR /go/src/terragrunt
+
+# Download modules in a separate layer so docker caches it, to reduce build time
+# when modules have not changed.
+COPY go.mod .
+COPY go.sum .
+RUN go mod download -x
+
+COPY . .
+
+RUN make build
+
+###
+
+FROM alpine:latest
+
+COPY --from=builder /go/src/terragrunt/terragrunt /usr/local/bin/
+
+ENTRYPOINT terragrunt
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,12 +12,12 @@ RUN go mod download -x
 
 COPY . .
 
-RUN make build
+RUN CGO_ENABLED=0 make build
 
 ###
 
-FROM alpine:latest
+FROM scratch
 
-COPY --from=builder /go/src/terragrunt/terragrunt /usr/local/bin/
+COPY --from=builder /go/src/terragrunt/terragrunt /app/
 
-ENTRYPOINT ["terragrunt"]
+ENTRYPOINT ["/app/terragrunt"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,4 +20,4 @@ FROM alpine:latest
 
 COPY --from=builder /go/src/terragrunt/terragrunt /usr/local/bin/
 
-ENTRYPOINT terragrunt
+ENTRYPOINT ["terragrunt"]
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,6 +60,19 @@ jobs:
     steps:
       - checkout
       - run: build-go-binaries --app-name terragrunt --dest-path bin --ld-flags "-X main.VERSION=$CIRCLE_TAG"
+  build_container:
+    <<: *defaults
+    steps:
+      - checkout
+      # Build the container
+      - run: docker build -t terragrunt-test -f Dockerfile .
+      # Run the container, outputting the terragrunt version and checking that it is correct
+      - run:
+          command: |
+            set -xe
+            vtag_maybe_extra=$(git describe --tags --abbrev=12 --dirty --broken)
+            docker run --rm terragrunt-test --version
+            docker run --rm terragrunt-test --version | grep $vtag_maybe_extra
   deploy:
     <<: *defaults
     steps:
@@ -86,6 +99,16 @@ workflows:
             - Gruntwork Admin
             - Gruntwork GCP
       - build:
+          requires:
+            - unit_test
+            - integration_test
+          filters:
+            tags:
+              only: /^v.*/
+          context:
+            - Gruntwork Admin
+            - Gruntwork GCP
+      - build_container:
           requires:
             - unit_test
             - integration_test


lorengordon:
fwiw, demonstrating it builds...

```
$ docker build -t gruntworkio/terragrunt -f Dockerfile .
[+] Building 153.1s (13/13) FINISHED
 => [internal] load build definition from Dockerfile
=> => transferring dockerfile: 353B
=> [internal] load .dockerignore
=> => transferring context: 2B
=> [internal] load metadata for docker.io/library/alpine:latest
=> [internal] load metadata for docker.io/library/golang:1.16.3-alpine
=> CACHED [stage-1 1/2] FROM docker.io/library/alpine:latest@sha256:69e70a79f2d41ab5d637de98c1e0b055206ba40a8145e7bddb55ccc04e13cf8f
=> CACHED [builder 1/5] FROM docker.io/library/golang:1.16.3-alpine@sha256:49c07aa83790aca732250c2258b5912659df31b6bfa2ab428661bc66833769e1
=> [internal] load build context
=> => transferring context: 109.75kB
=> [builder 2/5] RUN apk add --update --no-cache make git
=> [builder 3/5] WORKDIR /go/src/terragrunt
=> [builder 4/5] COPY . .
=> [builder 5/5] RUN make build
=> [stage-1 2/2] COPY --from=builder /go/src/terragrunt/terragrunt /usr/local/bin/
=> exporting to image
=> => exporting layers
=> => writing image sha256:26a8ca8ea8a6d6f55d916b6a9d7a4e57b099a45f4380bc7512131d74317fc6b1
=> => naming to docker.io/gruntworkio/terragrunt
```

and runs:

```
$ docker run --rm gruntworkio/terragrunt
DESCRIPTION:
   terragrunt - Terragrunt is a thin wrapper for Terraform that provides extra tools for working with multiple
   Terraform modules, remote state, and locking. For documentation, see https://github.com/gruntwork-io/terragrunt/.

USAGE:
   terragrunt <COMMAND>
...
VERSION:
   v0.16.5-1087-g7f1cff4aafa6-dirty

AUTHOR(S):
   Gruntwork <www.gruntwork.io>
```

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+# Download modules in a separate layer so docker caches it, to reduce build time
+# when modules have not changed.
+COPY go.mod .
+COPY go.sum .
+RUN go mod download -x
+
+COPY . .
+
+RUN make build
```
brikis98, line 15: This will end up with an empty version number in the binary. It needs to be set via:

```
-ldflags "-X main.VERSION=<VERSION>"
```

But it's not clear what version number we'd be able to use?

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+###
+
+FROM alpine:latest
```
brikis98, line 19: NIT: couldn't this be `scratch`? I think a standalone Go binary doesn't need anything else...

brikis98 (commented):
Thanks for the PR!

I could configure a Docker Hub build as you [mentioned here](https://github.com/gruntwork-io/terragrunt/issues/1655#issuecomment-831494034), but there are a few gotchas:

1. How do we set the version number in the binary? See my comment in the PR itself.
1. How do we test the `Dockerfile` _before_ merging/releasing? I'd rather not merge something and later find out the Docker Hub build failed.
1. Docker Hub has introduced a bunch of [limits](https://www.docker.com/increase-rate-limits) recently, and seems to be heading in the direction of adding still more. Terragrunt _might_ be popular enough to trip this limits... Not sure how many people use it via Docker though.

Dockerfile
lorengordon: The version is not empty! That's one reason I used the `make build` target. The version is determined by the command `git describe --tags --abbrev=12 --dirty --broken`. When the checkout matches a git tag, the value is just the git tag. When the checkout is not a tag, you get the last tag plus a git ref marker.

If you go with the dockerhub build service, dockerhub clones the repo and has all the tags, so this evaluates correctly. I'd expect any other container build service would also.

```
$ git describe --tags --abbrev=12 --dirty --broken
v0.29.2-2-gc90d385e21b8
$ git tag v0.29.3-testing
$ git describe --tags --abbrev=12 --dirty --broken
v0.29.3-testing
```

Dockerfile
lorengordon: Perhaps? Using `alpine:latest` and the multistage build, it's only a 40MB image. Might be over-optimizing... I'll give it a shot though.

lorengordon:
> How do we set the version number in the binary? See my comment in the PR itself.

I addressed that inline. The version is being set!

> How do we test the Dockerfile before merging/releasing? I'd rather not merge something and later find out the Docker Hub build failed.

What I usually do is build the image in CI and then run the container, exercising some suite of tests. How thorough I want to be with the image generally depends on how confident I am in the unit tests I run more directly. Sometimes I just will run the container passing a basic flag, like `--version`, to ensure the binary did build and can be executed.

> Docker Hub has introduced a bunch of limits recently, and seems to be heading in the direction of adding still more. Terragrunt might be popular enough to trip this limits... Not sure how many people use it via Docker though.

I haven't seen any DockerHub limits on building or pushing images for public projects. There are rate limits on users [_pulling_ images](https://docs.docker.com/docker-hub/download-rate-limit/), and it is the user's responsibility to login to dockerhub to avoid the rate limit.

There is also a plan to expire images that have not been used in 6 months, for accounts on the free tier. That seems pretty reasonable, though. If the image is being used, it remains available. If it's not used for 6 months, it seems likely it is not needed (and a user in desperate need could build it themselves at that point).

Dockerfile
lorengordon: For example:

```
$ docker run --rm gruntworkio/terragrunt --version
terragrunt version v0.29.2-3-gdbe93918437f-dirty
```

Dockerfile
lorengordon: The scratch base image worked, just needed to use `CGO_ENABLED=0`. Image size is 34MB.

```
$ docker run --rm gruntworkio/terragrunt --version
terragrunt version v0.29.2-4-g4deae14c6db3-dirty
```

lorengordon:
@brikis98 I added a step to the circle-ci config to build the container as an example of how to validate the docker build in CI, but I have no ability to run it, of course. I also have never used circle-ci, so I may have guessed wrong about how to do this! 😬 

> What I usually do is build the image in CI and then run the container, exercising some suite of tests. How thorough I want to be with the image generally depends on how confident I am in the unit tests or integration tests I run more directly. Sometimes I just will run the container passing a basic flag, like --version, to ensure the binary did build and can be executed.

angeloskaltsikis:
Hey folks,
We are thinking of using Terragrunt as a Kubernetes Cronjob (to get alerts for Drift like mentioned in this [discussion](https://github.com/gruntwork-io/terragrunt/issues/1166#issuecomment-742532815)).
As a result, we are going to prefer if we could utilize any official Terragrunt Docker image.
@brikis98 Do you have any other concerns or this could proceed? 🙏🏽

Matthew-Beckett:
Following on from @angeloskaltsikis question, what is blocking progression on this PR? I am happy to pick up getting this merged into branch so that [alpine/terragrunt](https://hub.docker.com/r/alpine/terragrunt) at Dockerhub with over 10 million downloads can be deprecated.

PR #1673, opened 2021-05-12
Add support for STS duration option

### LINKED ISSUE #495: terragrunt's --assume-role option and sts token expiration in terraform (default one hour duration)
Hi, 

Not a new subject but I regularly get caught by the default 1-hour sts token duration retrieved through the aws sts assume-role method called by terragrunt when running with the `--terragrunt-iam-role` option.

As I implement AWS multi-account setup, dedicating one account to aws iam users management and other accounts to segregate aws resources, i need to run terragrunt with the  `--terragrunt-iam-role`  option to jump in the right environments with temporary tokens.

The problem occurs when running very long terraform tasks such as restoring an rds aurora cluster from a backup which can easily take several hours.

I'm under the impression that terragrunt calls the assume-role method without specifying the `duration-seconds` param which enables to extend the default 1-hour duration up to 12 hours.

Then, passed 1 hour, terraform at some point fails with this error:
```
ExpiredToken: The provided token has expired
```

What would you think of letting terragrunt users specify the duration of sts tokens fetched by the `--terragrunt-iam-role` option to avoid breaching the token validity period?

---

**EDIT**: reviewing the code and the aws go provider, it seems that the default duration is set to 15 minutes, and I already am able to run with terragrunt terraform tasks longer than 15 minutes, so there might be something else already renewing the token, right?

---

**EDIT2**: instead of using the `--terragrunt-iam-role` option i manually endorsed the role using the `aws sts assume-role` function, setting the token duration to 4 hours, and all went flawlessly when using the terragrunt's option failed each time. So i think it validates my proposal :-)

thehunt33r:
This PR adds the possibility to specify a specific STS Session duration when assuming a role via Terragrunt.

Resolves #495

Test output: https://gist.github.com/thehunt33r/895f31f880142e9587dd6afd5937dc7c

From what I gather on the tests, it's only GCP/GCS stuff which fails (due to me not having a GCP account), which I haven't touched


aws_helper/config.go
```diff
@@ -124,7 +124,7 @@ func CreateAwsSession(config *AwsSessionConfig, terragruntOptions *options.Terra
 }
 
 // Make API calls to AWS to assume the IAM role specified and return the temporary AWS credentials to use that role
-func AssumeIamRole(iamRoleArn string) (*sts.Credentials, error) {
+func AssumeIamRole(iamRoleArn string, duration int64) (*sts.Credentials, error) {
```
brikis98, line 127: NIT: Could you use a name that makes the units clearer? E.g., `durationSeconds`?

aws_helper/config.go
```diff
@@ -202,8 +203,8 @@ func AssumeRoleAndUpdateEnvIfNecessary(terragruntOptions *options.TerragruntOpti
 		return nil
 	}
 
-	terragruntOptions.Logger.Debugf("Assuming IAM role %s", terragruntOptions.IamRole)
-	creds, err := AssumeIamRole(terragruntOptions.IamRole)
+	terragruntOptions.Logger.Debugf("Assuming IAM role %s with duration %d", terragruntOptions.IamRole, terragruntOptions.StsDuration)
```
brikis98, line 206:
```suggestion
	terragruntOptions.Logger.Debugf("Assuming IAM role %s with duration %d seconds", terragruntOptions.IamRole, terragruntOptions.StsDuration)
```

cli/args.go
```diff
@@ -110,6 +110,12 @@ func parseTerragruntOptionsFromArgs(terragruntVersion string, args []string, wri
 		return nil, err
 	}
 
+	envValue, envProvided := os.LookupEnv("TERRAGRUNT_STS_DURATION")
+	stsDuration, err := parseIntArg(args, OPT_TERRAGRUNT_STS_DURATION, envValue, envProvided, 3600)
```
brikis98, line 114: NIT: extract `3600` into a constant with a clear name (e.g., `default_iam_role_duration`) and comment (e.g., `# one hour`).

cli/cli_app.go
```diff
@@ -217,6 +219,7 @@ GLOBAL OPTIONS:
    terragrunt-source                            Download Terraform configurations from the specified source into a temporary folder, and run Terraform in that temporary folder.
    terragrunt-source-update                     Delete the contents of the temporary folder to clear out any old, cached source code before downloading new source code into it.
    terragrunt-iam-role                          Assume the specified IAM role before executing Terraform. Can also be set via the TERRAGRUNT_IAM_ROLE environment variable.
+   terragrunt-sts-duration                      Session duration for STS session. Can also be set via the TERRAGRUNT_STS_DURATION environment variable.
```
brikis98, line 222: NIT: I'm not sure "STS" is going to be as recognizable to many users as "IAM Role." Perhaps `--terragrunt-iam-role-duration` would be a clearer name? This applies to the CLI arg, env var, etc.

cli/cli_app.go
```diff
@@ -370,6 +373,11 @@ func RunTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 		terragruntOptions.IamRole = terragruntConfig.IamRole
 	}
 
+	// replace default sts duration if set in config
+	if terragruntOptions.StsDuration == 3600 && terragruntConfig.StsDuration != nil {
```
brikis98, line 377: Why are you checking for exactly equality to 3600? Do you mean `!=`? Any reason not to always set the param, as long as it's not `nil`?

config/config_as_cty.go
```diff
@@ -79,6 +79,15 @@ func terragruntConfigAsCty(config *TerragruntConfig) (cty.Value, error) {
 		output["retryable_errors"] = retryableCty
 	}
 
+	stsDurationCty, err := goTypeToCty(config.StsDuration)
+	if err != nil {
+		return cty.NilVal, err
+	}
+	fmt.Println(stsDurationCty)
```
brikis98, line 86: Left over debugging?

docs/_docs/04_reference/cli-options.md
```diff
@@ -381,28 +381,44 @@ This should allow you to run `import` on the module and work around those Terraf
...
-- [terragrunt-strict-include](#terragrunt-strict-include)
-- [terragrunt-ignore-dependency-order](#terragrunt-ignore-dependency-order)
-- [terragrunt-ignore-external-dependencies](#terragrunt-ignore-external-dependencies)
-- [terragrunt-include-external-dependencies](#terragrunt-include-external-dependencies)
-- [terragrunt-parallelism](#terragrunt-parallelism)
-- [terragrunt-debug](#terragrunt-debug)
-- [terragrunt-check](#terragrunt-check)
-- [terragrunt-hclfmt-file](#terragrunt-hclfmt-file)
-- [terragrunt-override-attr](#terragrunt-override-attr)
+- [CLI commands](#cli-commands)
```
brikis98, line 384: This is quite a large diff. Was your branch out of date or something?

docs/_docs/04_reference/cli-options.md
```diff
@@ -525,6 +541,14 @@ When passed in, the `*-all` commands continue processing components even if a de
 Assume the specified IAM role ARN before running Terraform or AWS commands. This is a convenient way to use Terragrunt
 and Terraform with multiple AWS accounts.
 
+### terragrunt-sts-duration
+
+**CLI Arg**: `--terragrunt-sts-duration`<br/>
+**Environment Variable**: `TERRAGRUNT_STS_DURATION`<br/>
+**Requires an argument**: `--terragrunt-sts-duration 3600`
+
+Uses the specified duration as the session duration of the STS credentials assumed for the role defined in `--terragrunt-iam-role`
```
brikis98, line 550: NIT: specify the units for the duration (seconds).

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -724,6 +733,19 @@ Example:
 iam_role = "arn:aws:iam::ACCOUNT_ID:role/ROLE_NAME"
 ```
 
+### sts_duration
+
+The `sts_duration` attribute can be used to specify an IAM role that Terragrunt should assume prior to invoking Terraform.
```
brikis98, line 738: NIT: specify the units for the duration (seconds).

options/options.go
```diff
@@ -100,6 +100,9 @@ type TerragruntOptions struct {
 	// The ARN of an IAM Role to assume before running Terraform
 	IamRole string
 
+	// Duration of the STS Session
+	StsDuration int64
```
brikis98, line 104: I believe there are `New` and `Clone` methods for `Options` that need to be updated with this param.

brikis98 (commented):
Thanks for the PR! And for updating the docs + tests 👍

docs/_docs/04_reference/cli-options.md
thehunt33r: Looks like my markdown linter went a bit haywire here. My branch wasn't that out of date (just 1 commit behind HEAD)! I'll fix that :)

config/config_as_cty.go
thehunt33r: 😅  Oops sorry. Fixed.

aws_helper/config.go
thehunt33r: Renamed it to `sessionDurationSeconds` which is clearer in terms of what the duration is for, as well as the unit. Thanks for the feedback.

options/options.go
thehunt33r: I've updated the Clone and New method and added a default const for the default StsDuration of 3600.

cli/cli_app.go
thehunt33r: I agree with the fact that `terragrunt-sts-duration` might be a bit short, however `terragrunt-iam-role-duration` seems "wrong" (i.e it's not really the duration of the IAM role, but of the AssumeRole session)

Do you think `terragrunt-iam-assume-role-duration` / `iam_assume_role_duration` / `TERRAGRUNT_IAM_ASSUME_ROLE_DURATION` might be better?

cli/cli_app.go
thehunt33r: The intent here was to check whether `StsDuration` had been overwritten (different from the default of 3600) in `terragruntOptions` (i.e. via CLI or ENV arguments). If it hasn't been overwritten, and `StsDuration` has been defined in `terragruntConfig`, then we should use the `StsDuration` defined in `terragruntConfig`.

I think this is a similar approach to the `DownloadDir` assignment on line 395:397.

thehunt33r:
Hey @brikis98 

Thank you so much for your review! I've addressed most of your comments.

cli/cli_app.go
brikis98: That works for me!

cli/cli_app.go
brikis98: Ah, OK. Extracting the 3600 into a constant will help make that clear.

thehunt33r:
Hey @brikis98 , thanks again for your review. 

I've renamed the argument as agreed, as well as extracted the `3600` into a var.

Finally I've re-run the tests and updated the gist. 

Would you like me to clean up the history before we can consider merging?

cli/cli_app.go
```diff
@@ -372,6 +375,12 @@ func RunTerragrunt(terragruntOptions *options.TerragruntOptions) error {
 		terragruntOptions.IamRole = terragruntConfig.IamRole
 	}
 
+	// replace default sts duration if set in config
+	defaultIamAssumeRoleDuration := 3600 // Default duration for IAM Assume role session is 3600s (1hr)
+	if terragruntOptions.IamAssumeRoleDuration == int64(defaultIamAssumeRoleDuration) && terragruntConfig.IamAssumeRoleDuration != nil {
```
brikis98, line 380: NIT: remove `defaultIamAssumeRoleDuration` and use `options.DEFAULT_IAM_ASSUME_ROLE_DURATION` here instead.

cli/args.go
```diff
@@ -115,6 +115,13 @@ func parseTerragruntOptionsFromArgs(terragruntVersion string, args []string, wri
 		return nil, err
 	}
 
+	envValue, envProvided := os.LookupEnv("TERRAGRUNT_IAM_ASSUME_ROLE_DURATION")
+	defaultIamAssumeRoleDuration := 3600 // Default STS sessions are 1 hour long (3600 seconds)
+	IamAssumeRoleDuration, err := parseIntArg(args, OPT_TERRAGRUNT_IAM_ASSUME_ROLE_DURATION, envValue, envProvided, defaultIamAssumeRoleDuration)
```
brikis98, line 120: NIT: remove `defaultIamAssumeRoleDuration` and use `options.DEFAULT_IAM_ASSUME_ROLE_DURATION` here instead.

brikis98 (commented):
Very close! Just a couple minor tweaks, and I can then kick off tests. Thank you!

config/config.go
```diff
@@ -85,6 +86,7 @@ type terragruntConfigFile struct {
 	PreventDestroy         *bool               `hcl:"prevent_destroy,attr"`
 	Skip                   *bool               `hcl:"skip,attr"`
 	IamRole                *string             `hcl:"iam_role,attr"`
+	IamAssumeRoleDuration            *int64              `hcl:"iam_assume_role_duration,attr"`
```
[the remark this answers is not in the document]
thehunt33r: oops... sorry my editor was acting up there. Fixed!

cli/cli_app.go
thehunt33r: Good catch! I hadn't noticed the options package was already imported/used for the purpose of setting consts across the whole project. I've updated this now :)

thehunt33r:
Here we go @brikis98; I've updated my PR with your requested changes. Thanks again for the review!

brikis98 (approved):

brikis98:
Tests passed! Merging now.

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.29.4

thehunt33r:
Thank you so much @brikis98!

I'll be waiting for the binaries and starting to use it ASAP :D

ct-hbk:
@brikis98 @thehunt33r - thank you guys! I am starting to use it ASAP too

Issue #1138, opened 2020-04-15
Add support for `--terragrunt-source-map` as a more consistent, predictable way to override source URLs


yorinasub17:
Terragrunt's `--terragrunt-source` parameter, which is designed to make it easy to override `source` URLs in `terragrunt.hcl`, has a couple issues:

1. If you run `terragrunt apply --terragrunt-source /foo/bar` in folder `yyy`, as of now, the *only* thing it does is override the `source` URL in `yyy/terragrunt.hcl` to the exact value `/foo/bar`. However, if you run `terragrunt apply-all --terragrunt-source /foo/bar` in folder `yyy`, it overrides the `source` value of each child module within `yyy/sub/path` to something like `/foo/bar/sub/path`. This inconsistency can be confusing and only gets worse when working with `read_terragrunt_config`, where we have to parse another `terragrunt.hcl` and decide whether to override its `source` URL or not.
1. `--terragrunt-source` assumes all your modules are coming from the *same* root repo. You won't be able to do an `apply-all` if you're pulling from multiple repos (e.g., your own infra modules, Terraform Registry, Gruntwork repos, etc). 

Perhaps a better solution to this is to support some sort of mapping:

```
terragrunt apply --terragrunt-source-map source:dest
```

The `--terragrunt-source-map` param would replace any `source` URL that starts with `source` to instead start with `dest`. 

For example:

```
terragrunt apply --terragrunt-source-map github.com/org/modules.git:/local/path/to/modules
```

The above would replace `source = "github.com/org/modules.git//xxx"` with `/local/path/to/modules//xxx` regardless of whether you were running `apply`, or `apply-all`, or using a `dependency`. This would give us a consistent behavior across all commands _and_ support mapping multiple repos by specifying the param multiple times:

```
terragrunt apply \
  --terragrunt-source-map github.com/org/modules.git:/local/path/to/modules \
  --terragrunt-source-map github.com/org/another-repo.git:/local/path/to/another-repo
```

We could also consider allowing users to specify this mapping in a file:

```
terragrunt apply --terragrunt-source-map mapping.hcl
```

Where `mapping.hcl` contains:

```hcl
mappings = {
  "github.com/org/modules.git"      = "/local/path/to/modules"
  "github.com/org/another-repo.git" = "/local/path/to/another-repo"  
}
```

See https://github.com/gruntwork-io/terragrunt/pull/1134#discussion_r407167726 for context for this issue.

yorinasub17:
[...]

PR #1674, opened 2021-05-14
Implement --terragrunt-source-map

yorinasub17:
Fixes https://github.com/gruntwork-io/terragrunt/issues/1138

This implements the `--terragrunt-source-map` feature described in the ticket linked above. Note the following differences with the issue description:

- `--terragrunt-source-map` uses standard unix CLI kv definition. That is, it uses `--terragrunt-source-map source=dest` instead of `--terragrunt-source-map source:dest`.
- Does not implement file based configuration. The file based configuration can be implemented as an improvement in a future PR if there is demand.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -2,14 +2,16 @@ package config
 
 import (
 	"fmt"
-	"github.com/mitchellh/mapstructure"
+	"net/url"
 	"os"
 	"path/filepath"
 	"reflect"
 	"strings"
 
+	"github.com/hashicorp/go-getter"
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclparse"
+	"github.com/mitchellh/mapstructure"
 	"github.com/sirupsen/logrus"
 	"github.com/zclconf/go-cty/cty"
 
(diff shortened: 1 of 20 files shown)


config/config.go
```diff
@@ -317,14 +319,79 @@ func (conf *TerraformExtraArguments) GetVarFiles(logger *logrus.Entry) []string
...
+	moduleUrlParsed, err := url.Parse(moduleUrl)
+	if err != nil {
+		return source, err
+	}
+	moduleUrlParsed.RawQuery = ""
+	moduleUrlQuery := moduleUrlParsed.String()
+
+	// Check if there is an entry to replace the URL portion in the map. Return the source as is if there is no entry in
+	// the map.
+	sourcePath, hasKey := sourceMap[moduleUrlQuery]
```
brikis98, line 378: Just to make sure I understand, we're doing an exact match? E.g., If I specify `source = "git::ssh://git@github.com/gruntwork-io/foo.git//bar"` in one place and `source = "ssh://git@github.com/gruntwork-io/foo.git//bar" (same URL, diff prefix) in another place, then `--terragrunt-source-map git::ssh://git@github.com/gruntwork-io/foo.git=/path/to/local-modules` will only match the former?

docs/_docs/04_reference/cli-options.md
```diff
@@ -501,6 +502,29 @@ for all of your Terraform modules, and for each module processed by the `xxx-all
...
 
+### terragrunt-source-map
+
+**CLI Arg**: `--terragrunt-source-map`<br/>
+**Requires an argument**: `--terragrunt-source-map git::ssh://github.com=/path/to/local-terraform-code`
+
+Can be supplied multiple times: `--terragrunt-source-map source1=dest1 --terragrunt-source-map source2=dest2`
+
+The `--terragrunt-source-map source=dest` param replaces any `source` URL (including those defined in `dependency`
+blocks) that has root `source` with `dest`.
```
brikis98, line 513: When you say in `dependency` block, do you mean the `config_path`? Or a `source` URL _within_ `dependency`?

brikis98 (commented):
Very nice!

Just a few questions in the PR. Also, two more here: 

1. Does it make sense to deprecate `--terragrunt-source`? 
1. Should we support an env var like `TERRAGRUNT_SOURCE_MAP`?

config/config.go
yorinasub17: Yes this is doing an exact match. I wanted to make this more fuzzy, but I thought that can introduce a lot of edge cases.

E.g., I wanted it to match URLs without specifying the protocol, but the url parser in go doesn't really parse `git::ssh://` correctly (it splits it as `git` for protocol, and URL becomes `:ssh://git@github.com`).

So for now, I went with the easier to understand and maintain version of the implementation.

yorinasub17: Added docs clarifying this: 44de255ddc0b249f8cf7584dab8d25e6505c9742

docs/_docs/04_reference/cli-options.md
yorinasub17: Source URL within dependency. Updated docs to clarify this: 44de255ddc0b249f8cf7584dab8d25e6505c9742

yorinasub17:
> Does it make sense to deprecate --terragrunt-source?

Let's wait a little bit for `--terragrunt-source-map` to get some "in the wild" usage. I am worried there are a few corner cases that are covered by `--terragrunt-source` that `--terragrunt-source-map` doesn't cover.

> Should we support an env var like TERRAGRUNT_SOURCE_MAP?

What would be the syntax for specifying multiple source mappings via env var?

config/config.go
brikis98: Roger!

docs/_docs/04_reference/cli-options.md
brikis98: Roger, thanks!

brikis98:
> > Does it make sense to deprecate --terragrunt-source?
> 
> Let's wait a little bit for `--terragrunt-source-map` to get some "in the wild" usage. I am worried there are a few corner cases that are covered by `--terragrunt-source` that `--terragrunt-source-map` doesn't cover.

That's a good point.

> 
> > Should we support an env var like TERRAGRUNT_SOURCE_MAP?
> 
> What would be the syntax for specifying multiple source mappings via env var?

Delimiter separated list? The delimiter could be space or semicolon or something else not commonly found in URLs.

brikis98 (approved):
Impl LGTM! Feel free to make the call on env var support (or add in another PR or file issue to track).

yorinasub17:
Thanks for review! I'll go ahead and merge this in + release it, and then open a new PR for the env var fix.

Issue #1704, opened 2021-06-07
yamldecode cannot parse string started with "0b"


andreykaipov:
This was an issue with Terraform in https://github.com/hashicorp/terraform/issues/23770.

It was fixed in https://github.com/hashicorp/terraform/pull/25273 by updating the `zclconf/go-cty-yaml` library from v1.0.1 to v1.0.2.

Here's the different behavior between Terraform and Terragrunt:

```console
❯ terraform console
> yamldecode("key: 0ba")
{
  "key" = "0ba"
}
> exit

❯ cat terragrunt.hcl
locals {
  test = yamldecode("key: 0ba")
}

❯ terragrunt version
Error: Error in function call

  on /tmp/test/terragrunt.hcl line 2, in locals:
   2:   test = yamldecode("key: 0ba")

Call to function "yamldecode" failed: cannot parse "0ba" as tag:yaml.org,2002:int.

ERRO[0000] Encountered error while evaluating locals.
ERRO[0000] /tmp/test/terragrunt.hcl:2,10-21: Error in function call; Call to function "yamldecode" failed: cannot parse "0ba" as tag:yaml.org,2002:int.
ERRO[0000] Unable to determine underlying exit code, so Terragrunt will exit with error code 1
```

Versions:

```console
❯ terragrunt --version
terragrunt version v0.29.4

❯ terraform --version
Terraform v0.15.4
on linux_amd64
```

PR #1706, opened 2021-06-07
go.mod: update go-cty-yaml; update yamldecode regressions

andreykaipov:
Fixes https://github.com/gruntwork-io/terragrunt/issues/1704

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,7 +30,7 @@ require (
 	github.com/hashicorp/hcl/v2 v2.9.1
 	github.com/hashicorp/terraform v0.12.24
 	github.com/hashicorp/terraform-config-inspect v0.0.0-20210318070130-9a80970d6b34
-	github.com/hashicorp/vault v1.7.0
+	github.com/hashicorp/vault v1.7.0 // indirect
 	github.com/kr/pretty v0.2.1 // indirect
 	github.com/lib/pq v1.8.0 // indirect
 	github.com/mattn/go-colorable v0.1.7 // indirect
@@ -48,6 +48,7 @@ require (
 	github.com/ulikunitz/xz v0.5.7 // indirect
 	github.com/urfave/cli v1.22.3
 	github.com/zclconf/go-cty v1.8.1
+	github.com/zclconf/go-cty-yaml v1.0.2 // indirect
 	go.mozilla.org/sops/v3 v3.7.0
 	golang.org/x/crypto v0.0.0-20210220033148-5ea612d1eb83
 	golang.org/x/oauth2 v0.0.0-20200902213428-5d25da1a8d43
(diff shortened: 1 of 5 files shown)


yorinasub17:
Thanks for your contribution! However, I think the go mod file was botched somehow because I am getting the following compile error when I checkout this PR:

```
# github.com/gruntwork-io/terragrunt/test
../../../../../pkg/mod/go.mozilla.org/sops/v3@v3.7.0/hcvault/keysource.go:16:2: ambiguous import: found package github.com/hashicorp/vault/api in multiple modules:
        github.com/hashicorp/vault v0.10.4 (/Users/yoriy/go/pkg/mod/github.com/hashicorp/vault@v0.10.4/api)
        github.com/hashicorp/vault/api v1.0.5-0.20210210214158-405eced08457 (/Users/yoriy/go/pkg/mod/github.com/hashicorp/vault/api@v1.0.5-0.20210210214158-405eced08457)
FAIL    github.com/gruntwork-io/terragrunt/test [setup failed]
```

Perhaps it's better to update the internal terraform version to latest, which is where we are getting the terraform defined helper functions from?

andreykaipov:
That's weird. The `go.mod` file didn't run for me even before updating `go-cty-yaml` so I think it's something on my part...

> Perhaps it's better to update the internal terraform version to latest, which is where we are getting the terraform defined helper functions from?

Would that fix it? Reading through https://github.com/hashicorp/terraform/issues/23770, it looks like it was fixed in Terraform in v0.13, so I imagine the fix would have been included in Terragrunt by now.

yorinasub17:
We don't update the internal link with terraform frequently, and only opt to update it as needed. So if you look at `go.mod`, you will see that we are linking in `v0.12.24`.

andreykaipov:
Ah I see! I tried updating it to the latest v0.15.5 but I'm getting even stranger module import errors. I'm kind of worried what other implications that might have though.

I realized my mistake in the original commit though. I didn't realize y'all exclude the vendor directory so I shouldn't have tidied. Would you be okay with only a change to the underlying library?

yorinasub17:
> I realized my mistake in the original commit though. I didn't realize y'all exclude the vendor directory so I shouldn't have tidied. Would you be okay with only a change to the underlying library?

That's fine! I just confirmed the updated go.mod works for me now, so we can move forward with this PR! I'll kick off the internal build, and if it passes, we can merge this in.

yorinasub17 (approved):

yorinasub17:
Merged and released! Thanks for your contribution! https://github.com/gruntwork-io/terragrunt/releases/tag/v0.29.10

(binaries will show up shortly)

andreykaipov:
Thank you so much! I really appreciate how quick the team cuts releases! 🙏

PR #1714, opened 2021-06-11
Fix data type issue with aws-provider-patch

### LINKED ISSUE #1709: aws-provider-patch data types
The `aws-provider-patch` command converts list inputs to a string, causing problems when the values are replaced.

E.g, `terragrunt aws-provider-patch --terragrunt-override-attr allowed_account_ids=["0000012345636"]` sets the `allowed_account_ids` to `"[0000012345636]"`.

yorinasub17:
Fixes #1709 

`aws-provider-patch` blindly patches the provider attributes to string type, but the underlying type for the provider might be different (as is the case with `allowed_account_ids`). To support this, I updated the `aws-provider-patch` to use json inputs instead.

Note that this does result in a backward incompatible API as you now need to provide `"` for the json to resolve correctly. I updated the docs accordingly to highlight this.

### DIFF
--- a/cli/aws_provider_patch.go
+++ b/cli/aws_provider_patch.go
@@ -10,7 +10,7 @@ import (
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclwrite"
 	"github.com/mattn/go-zglob"
-	"github.com/zclconf/go-cty/cty"
+	ctyjson "github.com/zclconf/go-cty/cty/json"
 
 	"github.com/gruntwork-io/terragrunt/errors"
 	"github.com/gruntwork-io/terragrunt/options"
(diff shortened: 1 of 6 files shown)


cli/aws_provider_patch.go
```diff
@@ -230,15 +233,31 @@ func patchAwsProviderInTerraformCode(terraformCode string, terraformFilePath str
...
 	if body == nil || body.GetAttribute(attr) == nil {
 		// We didn't find an existing block or attribute, so there's nothing to override
-		return false
+		return false, nil
+	}
+
+	// The cty library requires concrete types, but since the value is user provided, we don't have a way to know the
+	// underlying type. Additionally, the provider block themselves don't give us the typing information either unless
+	// we maintain a mapping of all possible provider configurations (which is unmaintainable). To handle this, we
+	// assume the user provided input is json, and convert to cty that way.
```
brikis98, line 248: Clever solution!

brikis98 (approved):
Very nice 👍

yorinasub17:
Thanks for review! Going to merge this in and cut a backward incompatible release.

PR #1715, opened 2021-06-11
Better error message for aws-provider-patch

yorinasub17:
This is a follow up to https://github.com/gruntwork-io/terragrunt/pull/1714.

After testing, I realized that the error message is super confusing when you run the command the old way without quoting:

```
ERRO[0007] invalid character 'e' looking for beginning of value
ERRO[0007] Unable to determine underlying exit code, so Terragrunt will exit with error code 1
```

This PR patches that to make it easier to follow and explain what is going on:

```
ERRO[0001] Could not determine underlying type of JSON string eu-west-1. This usually happens when the JSON string is malformed, or if the value is not properly quoted (e.g., "eu-west-1"). Underlying error: invalid character 'e' looking for beginning of value
ERRO[0001] Unable to determine underlying exit code, so Terragrunt will exit with error code 1
```

### DIFF
--- a/cli/aws_provider_patch.go
+++ b/cli/aws_provider_patch.go
@@ -249,11 +249,15 @@ func overrideAttributeInBlock(block *hclwrite.Block, key string, value string) (
 	valueBytes := []byte(value)
 	ctyType, err := ctyjson.ImpliedType(valueBytes)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


brikis98 (approved):
LGTM!

yorinasub17:
Thanks for review! Will merge and release.

Issue #1566, opened 2021-02-26
Implement imports RFC


yorinasub17:
https://terragrunt.gruntwork.io/docs/rfc/imports/

NOTE: In the RFC, we mention implementing `import` as a separate block. To expedite development, we recommend implementing it as part of `include` by enhancing its capability in backwards compatible ways.

Recommended phased implementation (expectation is that each of bullets is a separate PR):

1. Expose the underlying config values under a reference var. E.g., being able to say include.foo.locals.
     - This would replace `read_terragrunt_config`, and has the advantage of providing a way to solve #1128 because we have more control over when parsing happens during a block decode, than a function eval.

1. Add the ability to configure the way included config merges. Right now `include` does an automatic shallow merge. In addition to shallow merge, user should have the ability to specify (a) no merge (full replacement for `read_terragrunt_config`), and (b) deep merge.

1. Add the ability to include multiple levels. Note that a potential yak shave is that this may break the include related functions, like `get_parent_terragrunt_folder` and `path_relative_to_include`. We will need to respecify how these work when multiple levels are included.

Shocktrooper:
I see that this ticket has been marked as high priority and there are only 10 of those in total at the time of this comment so I believe it is one of the next few in the queue but do you have a general timeframe for this being implemented by chance @yorinasub17 ?

brikis98:
`high-priority` are tickets we'll try to prioritize internally at Gruntwork. We do not have a timeline on them.

yorinasub17:
First feature in the list is now implemented in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.30.4

yorinasub17:
Deep merge with cross include dependency reference support is now available: https://github.com/gruntwork-io/terragrunt/releases/tag/v0.31.3

yorinasub17:
Multiple (**single level**) `include` blocks is now supported starting https://github.com/gruntwork-io/terragrunt/releases/tag/v0.32.0

This feature should allow significantly more DRY `terragrunt` configs even without multiple `include` levels. Refer to the updated [use case documentation](https://terragrunt.gruntwork.io/docs/features/keep-your-terragrunt-architecture-dry/) for more details.

chrisurf:
@yorinasub17 what is the current status on adding the ability to include multiple levels? THX

yorinasub17:
[...]

PR #1716, opened 2021-06-14
Expose include config attributes

yorinasub17:
This is the first of several PRs for implementing the Imports RFC, as described in https://github.com/gruntwork-io/terragrunt/issues/1566

This PR is a minimal, backward compatible change that exposes included config attributes in the current config if the special `expose` attribute is toggled to `true`. When `expose` is set to `true`, users can reference the values from a parent included config using `include`, e.g. `include.locals.foo`.

Some caveats:
- In the interest of backward compatibility, the `expose` toggle defaults to `false` - in the future, once the full feature set is implemented, we can decide if we should default to `true` instead.
- Dependencies are not automatically fetched at the moment. This is a future feature (the ticket has been updated accordingly).
- Currently only one include block can exist per terragrunt config (this is why the include vars are available directly under `include` and not `include.LABEL`). This will change in the future.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -534,7 +534,12 @@ func ParseConfigFile(filename string, terragruntOptions *options.TerragruntOptio
 //      - dependency
 // 5. Merge the included config with the parsed config. Note that all the config data is mergable except for `locals`
 //    blocks, which are only scoped to be available within the defining config.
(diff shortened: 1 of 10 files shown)


brikis98 (approved):
LGTM! I like the idea of implementing `import` incrementally 👍

yorinasub17:

PR #1722, opened 2021-06-17
Fix go modules from vault changes

yorinasub17:
Vault reorganized their modules, which caused some problems with the imports. This fixes that.


zackproser (approved):
LGTM

yorinasub17:
Thanks for review! Merging.

PR #1724, opened 2021-06-17
Introduce concept of merge strategy for includes

yorinasub17:
This adds a new attribute `merge_strategy` to `include` blocks, allowing you to specify an `include` block without merging the config in. This can be used in conjunction with the new `expose` attribute to simulate `read_terragrunt_config` for certain use cases. Note that due to the limitation of single include blocks, it is not a full replacement for `read_terragrunt_config` (yet).

This PR also refactors the merge functionality of the config into a new `include.go` file. This is in preparation for the next PR after this one, which will introduce deep merge strategy.


config/config.go
```diff
@@ -203,6 +204,29 @@ func (cfg *IncludeConfig) GetExpose() bool {
 	return *cfg.Expose
 }
 
+func (cfg *IncludeConfig) GetMergeStrategy() (MergeStrategyType, error) {
```
brikis98, line 207: Is this necessary because you can't set the type of `MergeStrategy` to `*MergeStrategyType` in `IncludeConfig`? I guess the HCL parser doesn't know how to parse a string into that type?

brikis98 (approved):
LGTM!

config/config.go
yorinasub17: Yup... unfortunately.

yorinasub17:
Thanks for review! Will merge now.

PR #1725, opened 2021-06-18
Add Terraform Validate test

zackproser:
Add Terraform validation test. This test will scan the entire repo for Terraform modules and run terraform init and terraform validate on each.


yorinasub17:
We probably don't need this here, since all the terraform modules in this repo are test fixtures.

PR #1726, opened 2021-06-18
Bump test terraform to tf 1.0

yorinasub17:
This bumps the test terraform bin to 1.0. I also went ahead and bumped the internal pulled in terraform version to 0.15.3. Note that that specific version is unique, as it is the last version that exposes a lot of the internal helper functions. After that version, terraform hides all the internal helpers to `internal` package, which has special meaning in go: it can not be imported outside the project using it. So they are not accessible.

Longer term, we will need to figure out an alternative strategy for pulling in terraform helpers, but for now I think this would suffice.


remote/remote_state_gcs.go
```diff
@@ -14,7 +14,6 @@ import (
 	"github.com/gruntwork-io/terragrunt/options"
 	"github.com/gruntwork-io/terragrunt/shell"
 	"github.com/gruntwork-io/terragrunt/util"
-	"github.com/hashicorp/terraform/helper/pathorcontents"
```
yorinasub17, line 17: This package is no longer available, so we need an alternate function which I created down below.

bwhaley (approved):
LGTM!

> Longer term, we will need to figure out an alternative strategy for pulling in terraform helpers, but for now I think this would suffice.

That is unfortunate. Hopefully the internal functions don't start shifting around more frequently.

yorinasub17:
Thanks for review. Will merge this in now!

Issue #1727, opened 2021-06-18
include and local lookup fails in locals block


stevie-:
Terragrunt Version 0.30.6
Terraform 0.14.11

When we try to evaluate `include` and `local` on some line value is null.

pseudo test code

`test/terragrunt.hcl`  (parent)
```hcl
locals {
  environment = "dev"
}
```

`test/child/terragrunt.hcl` (child)

```hcl
include {
  path   = find_in_parent_folders()
  expose = true
}

locals {
  datalake_account_id         = "test"
  works1 = "${include.locals.environment}_"
  works2 = "${local.datalake_account_id}_"
  fails = "${include.locals.environment}_${local.datalake_account_id}"
}
```

```bash
❯ cd test/child
❯ terragrunt plan --terragrunt-debug  --terragrunt-log-level trace
DEBU[0000] Found locals block: evaluating the expressions.
DEBU[0000] Found locals block: evaluating the expressions.
DEBU[0000] Evaluated 1 locals (remaining 0): environment
Error: Attempt to get attribute from null value

  on PATHTO/test/child/terragrunt.hcl line 10, in locals:
  10:   fails = "${include.locals.environment}_${local.datalake_account_id}"

This value is null, so it does not have any attributes.

ERRO[0000] Encountered error while evaluating locals.
ERRO[0000] PATHTO/test/child/terragrunt.hcl:10,49-69: Attempt to get attribute from null value; This value is null, so it does not have any attributes.
ERRO[0000] Unable to determine underlying exit code, so Terragrunt will exit with error code 1
```

Works if used in `inputs` block.

yorinasub17:
Fixed in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.30.7 Thanks again for the detailed bug report and repro!

PR #1723, opened 2021-06-17
Fix include locals

yorinasub17:
fixes https://github.com/gruntwork-io/terragrunt/issues/1721

PR #1728, opened 2021-06-18
Should be continue, not return

yorinasub17:
Fixes https://github.com/gruntwork-io/terragrunt/issues/1727

https://github.com/gruntwork-io/terragrunt/pull/1723 was a 🤦 fix because it didn't take into account the fact that an expression can contain more than one variable...


zackproser (approved):
LGTM

yorinasub17:
Thanks for review! For posterity, the plan files being removed in this PR were checked in accidentally and was causing some test issues, so removed.

Merging now!

PR #1739, opened 2021-06-30
fix: Fix arguments order for planfile

### LINKED ISSUE #1271: terragrunt apply-all specific plan fails

PertsevRoman:
This issue should fix #1271 

This MR fixes args order when planfile is specified in extra_args for apply


yorinasub17:
Thanks for your contribution! However, just removing `-input=false` and `-auto-approve` isn't sufficient, because you will quickly run into the UX problem of having to handle interactive user input on multiple concurrent prompts.

In fact, this is a broken experience because all the prompts share a single stdin, so typing yes once will actually feed to *all* the stacks that are running. This means that, as a user, you might be approving a plan that you haven't seen yet.

Note also that the outputs from the plans for the interactive prompt will be incomprehensible because outputs from each concurrently running terraform action will be interleaved (since they all share stdout and stderr).

Given that, we can't really accept this as is because the feature is unusable unless parallelism is 1, but that is also not very practical for sufficiently large projects. To move forward, we would need a way to handle concurrent interactive prompts.

PertsevRoman:
Thank you @yorinasub17 for your answer!

So, the issue is promt handling. I was thinking that it can make sense to add the flags manually once we use suggested option.
I'll try to play with args order instead of disabling implicit flags.

PR #1740, opened 2021-07-01
fix: Check planfile and put it in the end of an arguments list

### LINKED ISSUE #1271: terragrunt apply-all specific plan fails

PertsevRoman:
This issue should fix #1271

This MR fixes args order when planfile is specified in extra_args for apply

### DIFF
--- a/options/options.go
+++ b/options/options.go
@@ -304,7 +304,14 @@ func (terragruntOptions *TerragruntOptions) Clone(terragruntConfigPath string) *
 
 // Check if argument is planfile TODO check file format
 func checkIfPlanFile(arg string) bool {
-	return util.IsFile(arg)
+	isFile := util.IsFile(arg)
+
+	if isFile {
+		ext := filepath.Ext(arg)
+		return ext == ".tfplan"
+	}
+
+	return false
 }
 
 // Extract planfile from arguments list
--- a/options/options.go
+++ b/options/options.go
@@ -315,19 +315,19 @@ func checkIfPlanfile(arg string) bool {
 
 // Extract planfile from arguments list
 func extractPlanFile(argsToInsert []string) (*string, []string) {
-	planfile := ""
-	filteredArgs := []string{}
+	planFile := ""
+	var filteredArgs []string
 
 	for _, arg := range argsToInsert {
 		if checkIfPlanfile(arg) {
-			planfile = arg
+			planFile = arg
 		} else {
 			filteredArgs = append(filteredArgs, arg)
 		}
 	}
 
-	if planfile != "" {
-		return &planfile, filteredArgs
+	if planFile != "" {
+		return &planFile, filteredArgs
 	}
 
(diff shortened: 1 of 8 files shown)


alexey-eshmetov (dismissed):
👍

options/options.go
```diff
@@ -302,8 +302,40 @@ func (terragruntOptions *TerragruntOptions) Clone(terragruntConfigPath string) *
...
+// Check if argument is planfile TODO check file format
+func checkIfPlanfile(arg string) bool {
+	stat, err := os.Stat(arg)
+
+	if err != nil {
+		return false
+	}
+
+	return !stat.IsDir()
+}
```
yorinasub17, line 314: You can use [util.IsFile](https://github.com/gruntwork-io/terragrunt/blob/master/util/file.go#L130) for checking if file.

options/options.go
```diff
@@ -302,8 +302,40 @@ func (terragruntOptions *TerragruntOptions) Clone(terragruntConfigPath string) *
...
+	planFile := ""
+	var filteredArgs []string
+
+	for _, arg := range argsToInsert {
+		if checkIfPlanfile(arg) {
+			planFile = arg
+		} else {
+			filteredArgs = append(filteredArgs, arg)
+		}
+	}
```
yorinasub17, line 327: Wouldn't this also extract out var files?

yorinasub17 (commented):
Thanks for your contribution! I had two comments regarding the implementation. Once those are addressed, I can kick off a build.

That said, this implementation (where you need to introspect if the arg is a plan file) makes me wonder if there should actually be a dedicated `plan_file` configuration in `terragrunt.hcl` or CLI (`--terragrunt-plan-file`).

options/options.go
PertsevRoman: Hi @yorinasub17 ,
fixed

options/options.go
PertsevRoman: @yorinasub17 , var files args are passed in `-var-file=filename` so `util.IsFile()` shouldn't return false

PertsevRoman: Also var files can't be specified with planfile

PertsevRoman: I've added tests which checks argument lists order

yorinasub17: You can pass in var files as `terragrunt -var-file filename` too, which I believe would fall into this case, right?

PertsevRoman: No, it'll fail with `flag provided but not defined: -var-file` error. Anyway, I'll special check for tfvars files

yorinasub17 (commented):
Updates LGTM! Will kick off a build now.

yorinasub17 (approved):
Regression test passed, so will merge this in. Thanks again for your contribution!

PertsevRoman:
@yorinasub17 thank you for help! Will be happy to join if there is some roadmap

PR #1744, opened 2021-07-08
Provider generation functions

PertsevRoman:
Implements #1745

This MR introduces few functions for providers generation


yorinasub17:
Thanks for the contribution! At a high level, I understand the benefits of such a construct, but I worry about having to maintain a struct that maps to the fields of the AWS provider. This means that we need to hand maintain the correlation between the `aws` provider config and the struct, which increases the overhead in terragrunt. We also now need to be version aware of the provider, as I imagine the structs could render out deprecated fields as the provider updates. It also opens us up to have to maintain the same set of code + structs for every provider out there (e.g., why only `aws`? why not also `kubernetes` and `google` and `azurerm` for that matter?), which can get unwieldly.

Is there an alternative approach where we can provide similar things, but without having to maintain a struct internally? E.g., is there a way we can get the schema from the provider source code, and use that for marshalling/unmarshalling? Or, if your concern is the use of heredoc syntax to generate the provider block, why not use json encoding?

```hcl
generate "provider" {
  path      = "provider.tf.json"
  if_exists = "overwrite"
  contents = jsonencode({
    provider = {
      aws = {
        region = local.default_region
        version = "= 2.3.1"
        allowed_account_ids = [local.account_id]
      }
    }
  })
}
```

PertsevRoman:
Hi @yorinasub17 ,
thanks for your feedback!

I'm starting to agree with approach you suggested on during this functionality development when I'm figuring out hcl syntax.
I think yes, it's better to choose "agnostic" approach and allow user deal with provider configuration in freestyle manner.

So, how could we name this function ? `hclencode` ?
This function will transform one HCL input to another HCL representation. That's would be interesting if HCL is able to express this. :)

PertsevRoman:
I'm worrying if we could use smth like this
```hcl
provider = {
  block = true
  block_name = "aws"

  params = {
    region = local.default_region
    version = "= 2.3.1"
    allowed_account_ids = [local.account_id]

    assume_role = {
      block = true
      params = {
          role_arn = local.role_arn
      }
    }
  }
}
```
because I'm not sure if previous example allows to represent blocks. Otherwise this example requires to remember some specific rules to configure root and nested blocks.

PertsevRoman:
> E.g., is there a way we can get the schema from the provider source code, and use that for marshalling/unmarshalling?

Honestly, don't think so. I reviewed AWS provider internals, and they use simple flat structures for representation, without any direct HCL integration/binding.

yorinasub17:
> because I'm not sure if previous example allows to represent blocks. Otherwise this example requires to remember some specific rules to configure root and nested blocks.

This rides on terraform json encoding, which does support nested blocks (although admittedly, it's verbose). See [the official docs on this](https://www.terraform.io/docs/language/syntax/json.html#nested-block-mapping).

Since json formatting of terraform code is accepted for machine generation, I wonder if we should just stick to that here? I'm not sure we can come up with an alternative format that will be familiar enough to the community. The only alternative I can think of is directly writing the provider blocks in the `generate` block, but that is not going to be easy to implement. E.g., something like:

```


generate "provider" {
  path      = "provider.tf.json"
  if_exists = "overwrite"
  contents {
    provider "aws" {
        region = local.default_region
        version = "= 2.3.1"
        allowed_account_ids = [local.account_id]
    }
  }
}
```

PertsevRoman:
> > because I'm not sure if previous example allows to represent blocks. Otherwise this example requires to remember some specific rules to configure root and nested blocks.
> 
> This rides on terraform json encoding, which does support nested blocks (although admittedly, it's verbose). See [the official docs on this](https://www.terraform.io/docs/language/syntax/json.html#nested-block-mapping).
> 
> Since json formatting of terraform code is accepted for machine generation and well documented, I wonder if we should just stick to that here? I'm not sure we can come up with an alternative format that will be familiar enough to the community. The only alternative I can think of is directly writing the provider blocks in the `generate` block, but that is not going to be easy to implement. E.g., something like:
> 
> ```
> 
> 
> generate "provider" {
>   path      = "provider.tf.json"
>   if_exists = "overwrite"
>   contents {
>     provider "aws" {
>       region = local.default_region
>       version = "= 2.3.1"
>       allowed_account_ids = [local.account_id]
>     }
>   }
> }
> ```

Looks perfect. Yes, it can require more complicated solution. I'll start to implement this approach. If I don't get succeed then I will implement file template based rendreing linke in terraform

PertsevRoman:
@yorinasub17 Just finished `templatefile` function implementation. So, it was stolen from native terraform implementation with few little changes. Also could we just use terrafom native implementation without any changes ? It's pretty simple to import

yorinasub17:
> Just finished templatefile function implementation. So, it was stolen from native terraform implementation with few little changes. Also could we just use terrafom native implementation without any changes ? It's pretty simple to import

Ah this is already imported and available. See https://terragrunt.gruntwork.io/docs/reference/built-in-functions/#terraform-built-in-functions, which is handled by https://github.com/gruntwork-io/terragrunt/blob/master/config/config_helpers.go#L133-L135

PertsevRoman:
Hmm, seems we have everything we need to get rid of EOF :)
So, then there is no necessary to implement something over current functionality. @yorinasub17 sorry I've taken your time

yorinasub17:
No problem! Glad we have something that works for you!

Issue #1603, opened 2021-03-23
Terragrunt configuration has misaligned inputs


lmayorga1980:
**Terraform Version:** `0.14.8`
**Terragrunt Version:** `v0.28.15`

I get an error when trying to validate inputs

`terragrunt validate-inputs`

```
ERRO[0000] Terragrunt configuration has misaligned inputs 
ERRO[0000] Unable to determine underlying exit code, so Terragrunt will exit with error code 1 
```

**variables.tf**

```
variable "machine_type" {
}

variable "image_family" {
}

variable "image_project" {
}

variable "instance_name" {
}

variable "admin_username" {
}

variable "admin_password" {
}

variable "instance_count" {
  default = 1
}

variable "region" {
  
}
```

**terragrunt.hcl**

```
inputs = {
  instance_count = 2
  image_family   = "windows-2019"
  image_project  = "windows-cloud"
  machine_type   = "e2-standard-2"
  instance_name  = "gcloud-instance"
  admin_username = "myadmin"
  admin_password = "somepassword"
  region         = "us-east1"
}

```

brikis98:
Could you share the full log output? Also, please turn up the [logging level](https://terragrunt.gruntwork.io/docs/reference/cli-options/#terragrunt-log-level).

lmayorga1980:
Is interesting that this command fails while the other commands like

`terragrunt validate` and `terragrunt plan` work normally.

```
DEBU[0000] Found locals block: evaluating the expressions. 
DEBU[0000] Evaluated 1 locals (remaining 0): region     
DEBU[0000] Running command: terraform --version          prefix=[/Users/user1/repos/tf-gcloud/dev/us-east1] 
DEBU[0000] Terraform version: 0.14.8                    
DEBU[0000] Reading Terragrunt config file at /Users/user1/repos/tf-gcloud/dev/us-east1/terragrunt.hcl 
DEBU[0000] Found locals block: evaluating the expressions. 
DEBU[0000] Evaluated 1 locals (remaining 0): region     
DEBU[0000] Downloading Terraform configurations from file:///Users/user1/repos/tf-gcloud/modules into /Users/user1/repos/tf-gcloud/dev/us-east1/.terragrunt-cache/UwDU4KmcvYMwQpkXTB8TGycN1pI/cQSWxX-efoUW-6aaeb--TtWy8Hw 
[...]

PR #1757, opened 2021-07-26
Introduce --terragrunt-strict flag for validate-inputs

zackproser:
Fixes https://github.com/gruntwork-io/terragrunt/issues/1603

These changes introduce the `--terragrunt-strict-validate` flag which enables the behavior outlined [here](https://github.com/gruntwork-io/terragrunt/issues/1603#issuecomment-805944075). 

By default, terragrunt's `validate-inputs` command runs in relaxed mode. In relaxed mode, `validate-inputs` only returns an error if there are required inputs that are missing. 

By passing `--terragrunt-strict-validate`, strict mode is enabled. When strict mode is enabled, Terragrunt will return an error if any required inputs are missing, OR if any variables that are passed in a Terragrunt configuration are NOT used in the underlying Terraform configuration.

### DIFF
--- a/cli/args.go
+++ b/cli/args.go
@@ -149,6 +149,8 @@ func parseTerragruntOptionsFromArgs(terragruntVersion string, args []string, wri
 		return nil, err
 	}
 
+	strictMode := parseBooleanArg(args, OPT_TERRAGRUNT_STRICT, false)
+
 	opts, err := options.NewTerragruntOptions(filepath.ToSlash(terragruntConfigPath))
 	if err != nil {
 		return nil, err
@@ -177,6 +179,7 @@ func parseTerragruntOptionsFromArgs(terragruntVersion string, args []string, wri
 	opts.WorkingDir = filepath.ToSlash(workingDir)
 	opts.DownloadDir = filepath.ToSlash(downloadDir)
 	opts.LogLevel = loggingLevel
+	opts.Strict = strictMode
 	opts.Logger = util.CreateLogEntry("", loggingLevel)
 	opts.Logger.Logger.SetOutput(errWriter)
 	opts.RunTerragrunt = RunTerragrunt
--- a/cli/args.go
+++ b/cli/args.go
@@ -149,7 +149,7 @@ func parseTerragruntOptionsFromArgs(terragruntVersion string, args []string, wri
 		return nil, err
 	}
 
-	strictMode := parseBooleanArg(args, OPT_TERRAGRUNT_STRICT, false)
+	validateStrictMode := parseBooleanArg(args, OPT_TERRAGRUNT_STRICT_VALIDATE, false)
 
 	opts, err := options.NewTerragruntOptions(filepath.ToSlash(terragruntConfigPath))
 	if err != nil {
@@ -179,7 +179,7 @@ func parseTerragruntOptionsFromArgs(terragruntVersion string, args []string, wri
 	opts.WorkingDir = filepath.ToSlash(workingDir)
 	opts.DownloadDir = filepath.ToSlash(downloadDir)
 	opts.LogLevel = loggingLevel
-	opts.Strict = strictMode
+	opts.Strict = validateStrictMode
 	opts.Logger = util.CreateLogEntry("", loggingLevel)
 	opts.Logger.Logger.SetOutput(errWriter)
 	opts.RunTerragrunt = RunTerragrunt
--- a/cli/args.go
+++ b/cli/args.go
@@ -179,7 +179,7 @@ func parseTerragruntOptionsFromArgs(terragruntVersion string, args []string, wri
 	opts.WorkingDir = filepath.ToSlash(workingDir)
 	opts.DownloadDir = filepath.ToSlash(downloadDir)
 	opts.LogLevel = loggingLevel
-	opts.Strict = validateStrictMode
+	opts.ValidateStrict = validateStrictMode
 	opts.Logger = util.CreateLogEntry("", loggingLevel)
 	opts.Logger.Logger.SetOutput(errWriter)
 	opts.RunTerragrunt = RunTerragrunt
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -50,7 +50,7 @@ const OPT_TERRAGRUNT_HCLFMT_FILE = "terragrunt-hclfmt-file"
 const OPT_TERRAGRUNT_DEBUG = "terragrunt-debug"
 const OPT_TERRAGRUNT_OVERRIDE_ATTR = "terragrunt-override-attr"
 const OPT_TERRAGRUNT_LOGLEVEL = "terragrunt-log-level"
-const OPT_TERRAGRUNT_STRICT = "terragrunt-strict"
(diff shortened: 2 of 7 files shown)


cli/cli_app.go
```diff
@@ -235,6 +237,7 @@ GLOBAL OPTIONS:
    terragrunt-override-attr                     A key=value attribute to override in a provider block as part of the aws-provider-patch command. May be specified multiple times.
    terragrunt-debug                             Write terragrunt-debug.tfvars to working folder to help root-cause issues.
    terragrunt-log-level                         Sets the logging level for Terragrunt. Supported levels: panic, fatal, error, warn (default), info, debug, trace.
+   terragrunt-strict                            Sets strict mode for the validate-inputs command. By default, strict mode is off. When this flag is passed, strict mode is turned on. When strict mode is turned off, the validate-inputs command will only return an error if required inputs are missing from all input sources (env vars, var files, etc). When strict mode is turned on, an error will be returned if required inputs are missing OR if unused variables are passed to Terragrunt.
```
yorinasub17, line 240: I could use a second opinion here from @brikis98, but looking at this now, I wonder if this should be named something more specific to `validate-inputs`? E.g., `terragrunt-strict-validate`.

Otherwise, it's confusing as you might mix this up with the existing [--terragrunt-strict-include flag](https://terragrunt.gruntwork.io/docs/reference/cli-options/#terragrunt-strict-include).

cli/validate_inputs.go
```diff
@@ -53,7 +53,7 @@ func validateTerragruntInputs(terragruntOptions *options.TerragruntOptions, work
 		}
 		terragruntOptions.Logger.Warn("")
 	} else {
-		terragruntOptions.Logger.Info("All variables passed in by terragrunt are in use.")
+		terragruntOptions.Logger.Info(fmt.Sprintf("All variables passed in by terragrunt are in use. Strict mode enabled: %t", terragruntOptions.Strict))
```
yorinasub17, line 56: NIT: strict mode info should be debug logs, not info.

cli/validate_inputs.go
```diff
@@ -63,12 +63,14 @@ func validateTerragruntInputs(terragruntOptions *options.TerragruntOptions, work
...
 
-	// Return an error when there are misaligned inputs.
-	if len(unusedVars) > 0 || len(missingVars) > 0 {
-		return fmt.Errorf("Terragrunt configuration has misaligned inputs")
+	// Return an error when there are misaligned inputs. Terragrunt strict mode defaults to false. When it is false,
+	// an error will only be returned if required inputs are missing. When strict mode is true, an error will be
+	// returned if required inputs are missing OR if any unused variables are passed
+	if len(missingVars) > 0 || len(unusedVars) > 0 && terragruntOptions.Strict {
+		return fmt.Errorf("Terragrunt configuration has misaligned inputs. Strict mode enabled: %t", terragruntOptions.Strict)
 	}
```
yorinasub17, line 74: NIT: in non strict mode, it should emit a warning log at the end for unused variables.

yorinasub17 (commented):
Implementation LGTM! Just had a few nits about logging, and a concern about the UX around flag name.

cli/cli_app.go
zackproser: Makes sense - the same thought crossed my mind as well.

cli/validate_inputs.go
zackproser: Done in 139b820

cli/validate_inputs.go
zackproser: Is this separate from the warning message that is logged on line 50?

yorinasub17: Yea. The idea here is that most people will/might only look at the last message, so we want to give the overall summary at the end.

yorinasub17: Something like `"Terragrunt configuration has misaligned inputs, but running in relaxed mode so ignoring."`

zackproser: Ah roger, thanks :+1:

cli/cli_app.go
brikis98: Agreed! `--terragrunt-strict-validate` seems reasonable.

zackproser: Changed to `--terragrunt-validate-strict` in 9fc55d6

cli/validate_inputs.go
zackproser: Done in 03bd7fb

cli/args.go
```diff
@@ -149,6 +149,8 @@ func parseTerragruntOptionsFromArgs(terragruntVersion string, args []string, wri
 		return nil, err
 	}
 
+	strictMode := parseBooleanArg(args, OPT_TERRAGRUNT_STRICT, false)
```
yorinasub17, line 152: Can you also rename the vars + attributes used in golang for this to also mention `STRICT_VALIDATE`/`validateStrictMode`? That way we won't get confused when working with the code.

cli/validate_inputs.go
```diff
@@ -63,13 +64,19 @@ func validateTerragruntInputs(terragruntOptions *options.TerragruntOptions, work
...
-	// Return an error when there are misaligned inputs.
-	if len(unusedVars) > 0 || len(missingVars) > 0 {
-		return fmt.Errorf("Terragrunt configuration has misaligned inputs")
+	// Return an error when there are misaligned inputs. Terragrunt strict mode defaults to false. When it is false,
+	// an error will only be returned if required inputs are missing. When strict mode is true, an error will be
+	// returned if required inputs are missing OR if any unused variables are passed
+	if len(missingVars) > 0 || len(unusedVars) > 0 && terragruntOptions.Strict {
+		return fmt.Errorf(fmt.Sprintf("Terragrunt configuration has misaligned inputs. Strict mode enabled: %t.", terragruntOptions.Strict))
+	} else if len(unusedVars) > 0 {
+		terragruntOptions.Logger.Info("Terragrunt configuration has misaligned inputs, but running in relaxed mode so ignoring.")
```
yorinasub17, line 77: NIT: should be warning level.
```suggestion
		terragruntOptions.Logger.Warn("Terragrunt configuration has misaligned inputs, but running in relaxed mode so ignoring.")
```

yorinasub17 (commented):
Updates almost LGTM! Just one more var name change for maintainability and I think this will be good to go!

cli/args.go
zackproser: Roger! - done in e3d0b82

yorinasub17: Just took a look at e3d0b829b35d5bea8057328fdf3e5fe2a7300536. Don't forget to rename `opts.Strict` too!

zackproser: Thanks! Renamed in 2f59cd7

yorinasub17 (commented):
Updates LGTM! The test failed, but I'm not sure if it is legit. I just kicked off another build, and if it fails, I'll investigate to see if it is related to this change.

yorinasub17 (approved):
Yup build was a transient failure - it passed on retry. This is good to go!

zackproser:
Thanks for the reviews Yori!

PR #1759, opened 2021-07-29
Deep merge strategy

yorinasub17:
This implements deep merge strategy for `include`. Refer to [the updated docs](https://github.com/gruntwork-io/terragrunt/pull/1759/files#diff-eac47411aff1ec957598f585714c7199a1c1c31a710ef472868bae5abef38cf9) for more info on what this means.

### DIFF
--- a/config/include.go
+++ b/config/include.go
@@ -4,13 +4,16 @@ import (
 	"fmt"
 	"path/filepath"
 
+	"github.com/imdario/mergo"
+	"github.com/zclconf/go-cty/cty"
+
 	"github.com/gruntwork-io/terragrunt/errors"
 	"github.com/gruntwork-io/terragrunt/options"
 	"github.com/gruntwork-io/terragrunt/util"
 )
 
 // Parse the config of the given include, if one is specified
-func parseIncludedConfig(includedConfig *IncludeConfig, terragruntOptions *options.TerragruntOptions) (*TerragruntConfig, error) {
+func parseIncludedConfig(includedConfig *IncludeConfig, terragruntOptions *options.TerragruntOptions, dependencyOutputs *cty.Value) (*TerragruntConfig, error) {
 	if includedConfig.Path == "" {
 		return nil, errors.WithStackTrace(IncludedConfigMissingPath(terragruntOptions.TerragruntConfigPath))
 	}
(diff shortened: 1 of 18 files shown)


config/include.go
```diff
@@ -133,6 +225,174 @@ func (targetConfig *TerragruntConfig) Merge(sourceConfig *TerragruntConfig, terr
...
 
+// DeepMerge performs a deep merge of the given sourceConfig into the targetConfig. Deep merge is defined as follows:
+// - For simple types, the source overrides the target.
+// - For lists, the two attribute lists are combined together in concatenation.
+// - For maps, the two maps are combined together recursively. That is, if the map keys overlap, then a deep merge is
+//   performed on the map value.
+// - Note that some structs are not deep mergeable due to an implementation detail. This will change in the future. The
+//   following structs have this limitation:
+//     - remote_state
+//     - generate
```
brikis98, line 236: Out of curiosity, what is the limitation?

config/include.go
```diff
@@ -133,6 +225,174 @@ func (targetConfig *TerragruntConfig) Merge(sourceConfig *TerragruntConfig, terr
...
+		if targetConfig.Dependencies != nil {
+			resultModuleDependencies.Paths = targetConfig.Dependencies.Paths
+		}
+		resultModuleDependencies.Paths = append(resultModuleDependencies.Paths, sourceConfig.Dependencies.Paths...)
+		targetConfig.Dependencies = resultModuleDependencies
+	}
+
+	if sourceConfig.RetryableErrors != nil {
+		targetConfig.RetryableErrors = append(targetConfig.RetryableErrors, sourceConfig.RetryableErrors...)
+	}
```
brikis98, line 295: Does this work OK if `targetConfig.RetryableErrors` is `nil`?

config/include.go
```diff
@@ -133,6 +225,174 @@ func (targetConfig *TerragruntConfig) Merge(sourceConfig *TerragruntConfig, terr
...
+		mergedInputs, err := deepMergeInputs(sourceConfig.Inputs, targetConfig.Inputs)
+		if err != nil {
+			return err
+		}
+		targetConfig.Inputs = mergedInputs
+	}
+
+	// MAINTAINER'S NOTE: The following structs cannot be deep merged due to an implementation detail (they do not
+	// support nil attributes, so we can't determine if an attribute was intentionally set, or was defaulted from
+	// unspecified - this is especially problematic for bool attributes).
```
brikis98, line 322: Ah, I see... So in the future, we'd need to update these to support `nil`?

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -393,6 +394,63 @@ inputs = {
 }
 ```
 
+**What is deep merge?**
```
brikis98, line 397: NIT: consider adding an example showing two `terragrunt.hcl` files and what a merged output would look like.

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -393,6 +394,63 @@ inputs = {
...
+
+When the `merge_strategy` for the `include` block is set to `deep`, Terragrunt will perform a deep merge of the included
+config. For Terragrunt config, deep merge is defined as follows:
+
+- For simple types, the source overrides the target.
```
brikis98, line 402: NIT: What is source and what is target? Would help to clarify, perhaps with the example mentioned above.

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -393,6 +394,63 @@ inputs = {
...
+definition). The following blocks have this limitation:
+    - `remote_state`
+    - `generate`
+
+Similarly, the `locals` block is deliberately omitted from the merge operation by design. That is, you will not be able
+to access parent config `locals` in the child config, and vice versa in a merge. However, you can access the parent
+locals in child config if you use the `expose` feature.
+
+Finally, `dependency` blocks have special treatment. When doing a `deep` merge, `dependency` blocks from **both** child
+and parent config are accessible in **both** places. For example, consider the following setup:
```
brikis98, line 420: Interesting. Why?

brikis98 (approved):
Nice! For the most part, LGTM. Just had some NITs and a couple questions.

config/include.go
yorinasub17: Yes. Append handles `nil` correctly. See https://play.golang.org/p/5RRq6Owj_sS

config/include.go
yorinasub17: I'm going to assume you got the answer from the maintainer's note down below, so will mark as resolved.

config/include.go
yorinasub17: Yea because we can't tell if a `false` bool attribute was intentionally set in the code in the child `terragrunt.hcl` config, or if it was because it was unspecified. If unspecified, then we want to inherit from the parent, but if it was intentionally set, then it needs to override the parent. Since the resulting behavior is different, it's important that we can differentiate the two cases.

I decided to avoid addressing this for now in this PR since it was already getting big, but we will definitely want to address this as it feels limiting to not be able to deep merge an important block like `remote_state`.

docs/_docs/04_reference/config-blocks-and-attributes.md
yorinasub17: I don't really have a good answer to this other than that it naturally happened when I coded the deep merge. I actually realized this fact when I started testing it.

I remember chasing the reasoning when I realized it during implementation, but it's been about 2 months since I coded this up so I have forgotten it now...

yorinasub17:
Thanks for review! I'll merge now, and then open a follow up PR with the docs nit fixes.

docs/_docs/04_reference/config-blocks-and-attributes.md
yorinasub17: Added example: https://github.com/gruntwork-io/terragrunt/pull/1764

docs/_docs/04_reference/config-blocks-and-attributes.md
yorinasub17: Switched to child and parent language: https://github.com/gruntwork-io/terragrunt/pull/1764

PR #1763, opened 2021-07-30
Caching of run_cmd invocations

denis256:
Updated `DecodeBaseBlocks` to cache results of `evaluateLocalsBlock`, as cache key is used checksum calculated form HCL file content

Included changes:
 * map for caching of calculated locals, as key is used md5 from file content;
 * test for verification if single execution of "run_cmd" is executed only once;
 * test for verification if multiple invocations of "run_cmd" are handled correctly;

Example:

```
locals {
  foo = run_cmd("echo", uuid())
}

```

Before:
```
$ terragrunt init
d5cf86eb-3894-14a9-d17d-7cf08e5c3c9b
dc24b9ee-6acb-7f02-8f48-500fc4c10406
ERRO[0000] Did not find any Terraform files (*.tf) in /raid1/projects-work/g/terragrunt-denis/test1
```

After:
```
$ terragrunt init
3207b829-b30f-e8bc-77a6-60b8581f4c41
ERRO[0000] Did not find any Terraform files (*.tf) in /raid1/projects-work/g/terragrunt-denis/test1
```

Fix for: #1427


yorinasub17 (commented):
Ah so you took the approach of caching the locals? I think this works for most use cases, but there are a few cases where this won't catch. E.g., I think there are places where users will use `run_cmd` in the `terraform` block and I think the same issue exists there.

It's probably ok to go with this approach, but it needs to be documented (e.g., if you use `run_cmd` outside of `locals` blocks, there is a possibility that it could be called twice). Can you update the following docs pages to highlight this behavior?

- https://github.com/gruntwork-io/terragrunt/blob/master/docs/_docs/02_features/locals.md
- `run_cmd` function docs: https://github.com/gruntwork-io/terragrunt/blob/master/docs/_docs/04_reference/built-in-functions.md
- `locals` block reference docs: https://github.com/gruntwork-io/terragrunt/blob/master/docs/_docs/04_reference/config-blocks-and-attributes.md

Outside of that, this is good to go!

brikis98:
> Ah so you took the approach of caching the locals? I think this works for most use cases, but there are a few cases where this won't catch. E.g., I think there are places where users will use `run_cmd` in the `terraform` block and I think the same issue exists there.

Ah, interesting... Could we cache these other blocks too? Or just cache everything on a file-by-file basis?

yorinasub17:
> Could we cache these other blocks too? Or just cache everything on a file-by-file basis?

It's hard to cache those blocks from a technical perspective because they're parsed directly by the HCL parser unlike the `locals` blocks. It's also hard to cache everything because we take multiple partial passes through the config. That's part of why my original suggestion was caching at the function level.

brikis98:
> > Could we cache these other blocks too? Or just cache everything on a file-by-file basis?
> 
> It's hard to cache those blocks from a technical perspective because they're parsed directly by the HCL parser unlike the `locals` blocks. It's also hard to cache everything because we take multiple partial passes through the config. That's part of why my original suggestion was caching at the function level.

Understood. @denis256 do you want to give caching at the function level a shot? I'm curious if that helps avoid the limitation where `run_cmd` outside of `locals` may still be executed multiple times.

denis256:
Updated PR to cache invocations of `run_cmd` in `config_helpers.go` - as cache key is used path and command arguments.

Example behavior:

```
locals {
  uuid = run_cmd("echo", uuid())
  uuid2 = run_cmd("echo", uuid())
  potato = run_cmd("echo", "potato")
  potato2 = run_cmd("echo", "potato")
  carrot = run_cmd("echo", "carrot")
}

```

Output:
```
bd3499c9-cff7-e552-fbd7-29d90bb5c6d7
9efc7d00-c67c-6093-162d-0b380426d322
potato
carrot
```
Invocation of `run_cmd("echo", "potato")` got cached and printed only once, however invcations of `run_cmd("echo", uuid())` miss cache each time because "echo" is invoked each time with random UUID.

brikis98:
> Updated PR to cache invocations of `run_cmd` in `config_helpers.go` - as cache key is used path and command arguments.

Thanks!

> Example behavior:
> 
> ```
> locals {
>   uuid = run_cmd("echo", uuid())
>   uuid2 = run_cmd("echo", uuid())
>   potato = run_cmd("echo", "potato")
>   potato2 = run_cmd("echo", "potato")
>   carrot = run_cmd("echo", "carrot")
> }
> 
> inputs = {
>   fileName = run_cmd("echo", "carrot")
> }
> ```
> 
> Output:
> 
> ```
> bd3499c9-cff7-e552-fbd7-29d90bb5c6d7
> 9efc7d00-c67c-6093-162d-0b380426d322
> potato
> carrot
> ```
> 
> Invocations of `run_cmd("echo", "potato") / run_cmd("echo", "carrot")` got cached and printed only once, however invcations of `run_cmd("echo", uuid())` miss cache, and executed each time - because "echo" was invoked with random UUID.

Ohh, interesting. So I guess this is happening because the args that you're caching are _after_ parsing of HCL... So any native Terraform functions, such as `uuid()` have already been cached? 

If that's the case, then I'm afraid this won't really fix the issue, as we'll still have unexpected behavior. Consider, say:

```hcl
locals {
  foo = run_cmd("bash", "-c", "echo ${uuid()} >> foo.txt")
}
```

This code will be executed twice, with two different values appended to `foo.txt`, whereas the user would expect only one.

Hmph, well, it's good you had those automated tests in place to catch that.

So, thinking out loud:

1. Is there any way to cache the contents _before_ Terraform execution?
2. If not, are there any other approaches we should consider?

denis256:
In the current implementation, with cache of locals(config_partial.go) and  `rum_cmd` (config_helpers.go) it will be executed only once - added test case for this.

```
$ wc -l foo.txt
0 foo.txt

$ cat cat terragrunt.hcl
locals {
  foo = run_cmd("bash", "-c", "echo ${uuid()} >> foo.txt")
}

$ terragrunt init

$ wc -l foo.txt
1 foo.txt

```

brikis98:
(Sorry for the delay; I'm OOO on vacation; will respond as soon as I can!)

config/config_helpers.go
```diff
@@ -242,6 +250,19 @@ func runCommand(args []string, include *IncludeConfig, terragruntOptions *option
 
 	currentPath := filepath.Dir(terragruntOptions.TerragruntConfigPath)
 
+	// caching key based on path and arguments
```
brikis98, line 253: NIT: Extract the cache look up logic into a reusable function. The point of the function is less about code reuse and more about helping maintainers understand why we cache in the first place. To make that super clear, add a comment to the function to explain that we use a cache to ensure that any given `run_cmd` is only executed once, even if we have multiple parsing passes through this code, and include a link to https://github.com/gruntwork-io/terragrunt/issues/1427 for further info.

config/config_helpers.go
```diff
@@ -255,6 +276,10 @@ func runCommand(args []string, include *IncludeConfig, terragruntOptions *option
 		terragruntOptions.Logger.Debugf("run_cmd output: [%s]", value)
 	}
 
+	runCommandCacheMutex.Lock()
+	runCommandCache[cacheKey] = value
+	runCommandCacheMutex.Unlock()
```
brikis98, line 281: Similarly, extract the cache storage logic into a function too, with a similar comment.

config/config_partial.go
```diff
@@ -115,6 +123,13 @@ func DecodeBaseBlocks(
 		return nil, nil, TrackInclude{}, err
 	}
 
+	md5Sum := md5.Sum(hclFile.Bytes)
```
brikis98, line 126: Use the lookup function here

config/config_partial.go
```diff
@@ -132,6 +147,10 @@ func DecodeBaseBlocks(
 		return nil, nil, trackInclude, err
 	}
 
+	localsCacheMutex.Lock()
+	localsCache[cacheKey] = localsAsCty
+	localsCacheMutex.Unlock()
```
brikis98, line 152: Use the storage function here

config/config_partial.go
```diff
@@ -87,6 +89,12 @@ type terragruntRemoteState struct {
 	Remain      hcl.Body               `hcl:",remain"`
 }
 
+// Map with cached evaluated locals to avoid re-evaluation
+var localsCache = map[string]cty.Value{}
```
brikis98, line 93: Sanity check, if we have a `run_cmd` cache, why do we need a separate `locals` cache?

brikis98 (commented):
OK, just came back and had a chance to look through this. I think I misunderstood the example with `uuid()` before; for some reason, I had read it that a single `run_cmd` with `uuid()` would execute twice, but it's actually that two separate `run_cmd` calls with similar arguments will execute twice if those arguments involve `uuid()` (or anything else that produces different values after HCL parsing/processing) or once if those arguments are more static. Ideally, we'd be consistent across the board, but it seems like a minor issue, and the fixes in this PR are a vast improvement over what we have today! Also, the test cases look great, thanks for adding those.

I added a few final NITs and then I can kick off tests on this PR!

denis256:
Updated PR with:
 * extracted cache in separated structure with access functions and basic tests - extracted as a separated structure since it can be re-used in future;
 * set caching only of `run_cmd` invocations - will get more consistent behaviour but sill calculate `locals` multiple times;
 
Behaviour after changes:

```
terragrunt.hcl
locals {
  uuid = run_cmd("echo", "uuid1",  uuid())
  uuid2 = run_cmd("echo", "uuid2", uuid())
  potato = run_cmd("echo", "potato")
  potato2 = run_cmd("echo", "potato")
}
inputs = {
  potato3 = run_cmd("echo", "potato")
}

$ terragrunt init

uuid1 a22581be-2bac-199a-7251-80b7590ec753
uuid2 a24b58bc-8a96-dc21-96bf-64492abbf277
potato
uuid1 388f5660-f1b8-629e-1b8b-6a3702c6e705
uuid2 3196da16-d105-f5fe-7318-f3e4ac8e4d2a
ERRO[0000] Did not find any Terraform files (*.tf)

```

brikis98 (commented):
> * extracted cache in separated structure with access functions and basic tests - extracted as a separated structure since it can be re-used in future;

Nice. That makes the intent clearer and may have flushed out a potential bug.

> * set caching only of `run_cmd` invocations - will get more consistent behavior but sill calculate `locals` multiple times;
> 
> Behavior after changes, without locals cache:
> 
> ```
> $ cat terragrunt.hcl
> locals {
>   uuid = run_cmd("echo", "uuid1",  uuid())
>   uuid2 = run_cmd("echo", "uuid2", uuid())
>   uuid3 = run_cmd("echo", "uuid3", uuid())
>   potato = run_cmd("echo", "potato")
>   potato2 = run_cmd("echo", "potato")
> }
> inputs = {
>   potato3 = run_cmd("echo", "potato")
>   uuid3 = run_cmd("echo", "uuid3", uuid())
>   uuid4 = run_cmd("echo", "uuid4", uuid())
> }
> 
> $ terragrunt init
> 
> uuid1 b48379e1-924d-2403-8789-c72d50be964c
> uuid1 9f3a8398-b11f-5314-7783-dad176ee487d
> uuid2 2d65972b-3fa9-181f-64fe-dcd574d944d0
> uuid3 e345de60-9cfa-0455-79b7-af0d053a15a5
> potato
> uuid3 7f90a4ed-96e3-1dd8-5fee-91b8c8e07650
> uuid2 8638fe79-c589-bebd-2a2a-3e6b96f7fc34
> uuid3 310d0447-f0a6-3f67-efda-e6b1521fa1fb
> uuid4 f8e80cc6-1892-8db7-bd63-6089fef00c01
> 
> ERRO[0000] Did not find any Terraform files (*.tf)
> ```
> 
> Notes:
> 
> * `run_cmd("echo", "potato")`  - got cached and was executed only once even if was referenced in locals and inputs;
> * `uuid1-3` executed multiple times because of arguments uuid argument;
> * `uuid3` - got evaluated 3 times: 2 times because of locals and once in inputs;
> * `uuid4` - only once since it is referenced only in `inputs`;
> 
> Not sure if this is what was expected - results of `run_cmd` are cached in any block, but `locals` are still evaluated multiple times.

Why does `uuid1` appear twice?

denis256:
> Why does uuid1 appear twice?
Without caching of locals, all locals are evaluated twice, and in the case of UUID it misses cache for `run_cmd`

Evaluation in code:
```
config/config.go:537
config/config_partial.go:170
```

brikis98:
> > Why does uuid1 appear twice?
> 
> Without caching of locals, all locals are evaluated twice, and in the case of UUID it misses cache for `run_cmd`
> 
> Evaluation in code:
> 
> ```
> PartialParseConfigString() config/config_partial.go:170
> ParseConfigString() config/config.go:537
> ```

Ah, right. Man, there are several things at play here, and its easy to forget about one or the other:

1. Everything in `locals` is evaluated twice based on how Terragrunt currently does parsing.
1. This doesn't matter for the vast majority of the code, as it's idempotent.
1. However, `run_cmd` is an exception, as it might not be idempotent (e.g., may have side effects).
1. To avoid executing `run_cmd` calls multiple times, we cache them, based on the args passed to `run_cmd`.
1. This way, after the first execution, subsequent ones are fetched from the cache.
1. However, the args passed to `run_cmd` are evaluated _before_ caching, so in the case of function calls like `uuid()`, where you get a different value on each evaluation, `run_cmd` is still going to be executed multiple times.

Does that capture it correctly and clearly? If so, we may want to add the description to the code and the docs. In the docs, we show the example you have after this description.

brikis98 (dismissed):
LGTM, thank you!

I'll kick off tests now.

brikis98:
Looks like pre-commit checks failed:

```
goimports................................................................Failed
- hook id: goimports
- files were modified by this hook

config/cache_test.go
```

Please run `goimports`!

denis256:
Updated PR, maybe will be good to have `goimports` as an option in makefile

brikis98:
Re-running tests!

brikis98 (approved):
OK, all tests passed, except one, which I believe is due to an unrelated issue flushed out by having a slash in the branch name. I filed https://github.com/gruntwork-io/terragrunt/issues/1778 to track that problem. It has nothing to do with this fix, so I'm going to merge, thanks!

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.31.6

PR #1767, opened 2021-08-12
Terraform Module Registry support

yorinasub17:
Fixes https://github.com/gruntwork-io/terragrunt/issues/311

This PR adds a new getter protocol to allow downloading Terraform modules from any Terraform Registry, including private ones.

Example source URL: `tfr://registry.terraform.io/terraform-aws-modules/vpc/aws?version=3.3.0`

Note that there are a few design decisions that were made here that are worth commentating on:

- Ideally we would be able to reuse code from Terraform. However, `terraform` has moved to packaging all its libraries under `internal` so that you can't use them as a library outside of Terraform. To respect the direction and spirit of the Terraform team, I opted for not doing anything funky to workaround the limitation (like copying those files in here). This means that we would need to roll our own registry client, but luckily, this wasn't too hard as shown here.

- I set the new `tfr` package to `internal` to respect the `terraform` team's decision to avoid increasing the surface area of terraform registry clients.

- To keep the implementation simple, I opted for introducing a new protocol string to the source URL. This cleans up the implementation by only requiring to implement a custom `Getter` and loading it into `go-getter` without mucking with the Terragrunt config to introduce another field for module versions (like how the Terraform module source parameter works), or implementing complex detector interfaces. This risks deviating from familiar APIs, but I think the maintenance overhead of supporting the Terraform interface is too costly.

- Similarly, I opted for using an environment variable provided API token for authenticating to private module registries. This is unconventional (not something Terraform supports), which again risks deviating from familiar APIs of Terraform, but the maintenance overhead of duplicating the logic without importing didn't seem worth it for this initial version. I think the implementation is modular enough that we can add this in later if we deem it necessary.

**IMPORTANT**: For testing, I created a new null resource module and published it to [the registry](https://registry.terraform.io/modules/yorinasub17/terragrunt-registry-test/null/latest). For speed of iteration/implementation, I created [this repo under my account](https://github.com/yorinasub17/terraform-null-terragrunt-registry-test), but we should migrate this to the `gruntwork-io` org. I will need help from @brikis98 to complete this.


lorengordon:
Neat idea. Makes me wonder if go-getter would accept this upstream?

yorinasub17:
TODO: Investigate if `localterraform.com` needs special support (https://www.terraform.io/docs/cloud/registry/using.html#the-generic-module-hostname-localterraform-com-)

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -36,9 +36,19 @@ before or after calling Terraform.
...
+  for the Terraform registry** (see below note), including local file paths, Git URLs, and Git URLS with `ref`
+  parameters. Terragrunt will download all the code in the repo (i.e. the part before the double-slash `//`) so that
+  relative paths work correctly between modules in that repo.
+    - NOTE: the `source` parameter can be configured to pull Terraform modules from any Terraform module registry using
+      the `tfr` protocol. The `tfr` protocol expects URLs to be provided in the format
+      `tfr://REGISTRY_HOST/MODULE_SOURCE?version=VERSION`. For example, to pull the `terraform-aws-modules/vpc/aws`
+      module from the public Terraform registry, you can use the following as the source parameter:
+      `tfr://registry.terraform.io/terraform-aws-modules/vpc/aws?version=3.3.0`. If you wish to access a private module
+      registry (e.g., [Terraform Cloud/Enterprise](https://www.terraform.io/docs/cloud/registry/index.html)), you can
+      provide the authentication to Terragrunt as an environment variable with the key `TG_TF_REGISTRY_TOKEN`. This
```
brikis98, line 49: Terraform works off a [token in a config file](https://www.terraform.io/docs/cli/config/config-file.html#credentials). Is there any way for us to reuse this existing mechanism, so you don't have to set a separate env var?

internal/tfr/doc.go
```diff
@@ -0,0 +1,7 @@
+// Package tfr contains functions and routines for interacting with terraform registries using the Module Registry
+// Protocol documented in the official terraform docs:
+// https://www.terraform.io/docs/internals/module-registry-protocol.html
+//
+// MAINTAINER'S NOTE: This package is intentionally set to internal as we do not want to support usage of this outside
```
brikis98, line 5: NIT: prob worth mentioning the context from the PR description here: namely, that Terraform's own code is in an `internal` package, so to respect their design, we aren't using it directly, and similarly putting our code in an `internal` package.

internal/tfr/getter.go
```diff
@@ -0,0 +1,259 @@
...
+
+// Get is the main routine to fetch the module contents specified at the given URL and download it to the dstPath.
+// This routine assumes that the srcURL points to the Terraform registry URL, with the Path configured to the module
+// path encoded as `:namespace/:name/:system` as expected by the Terraform registry. Note that the URL query parameter
+// must have the `version` key to specify what version to download.
+func (tfrGetter *TerraformRegistryGetter) Get(dstPath string, srcURL *url.URL) error {
+	ctx := tfrGetter.Context()
+
+	registryDomain := srcURL.Host
+	queryValues := srcURL.Query()
```
brikis98, line 86: Any chance this is `nil`?

internal/tfr/getter.go
```diff
@@ -0,0 +1,259 @@
...
+		Scheme: "https",
+		Host:   domain,
+		Path:   serviceDiscoveryPath,
+	}
+	bodyData, _, err := httpGETAndGetResponse(ctx, sdURL)
+	if err != nil {
+		return "", err
+	}
+
+	var respJSON map[string]interface{}
```
brikis98, line 199: Any reason not to parse into a more well-defined, typed struct?

internal/tfr/getter.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+// Get is the main routine to fetch the module contents specified at the given URL and download it to the dstPath.
+// This routine assumes that the srcURL points to the Terraform registry URL, with the Path configured to the module
+// path encoded as `:namespace/:name/:system` as expected by the Terraform registry. Note that the URL query parameter
+// must have the `version` key to specify what version to download.
+func (tfrGetter *TerraformRegistryGetter) Get(dstPath string, srcURL *url.URL) error {
+	ctx := tfrGetter.Context()
+
+	registryDomain := srcURL.Host
```
brikis98, line 85: NIT: worth defaulting the Host to `app.terraform.io` if not explicitly specified in `source`? Would allow for the pretty short URLs like `hashicorp/consul/aws`.

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -36,9 +36,19 @@ before or after calling Terraform.
...
-  (i.e. the part before the double-slash `//`) so that relative paths work correctly between modules in that repo.
+  [module source](https://www.terraform.io/docs/modules/sources.html) parameter for Terraform `module` blocks **except
+  for the Terraform registry** (see below note), including local file paths, Git URLs, and Git URLS with `ref`
+  parameters. Terragrunt will download all the code in the repo (i.e. the part before the double-slash `//`) so that
+  relative paths work correctly between modules in that repo.
+    - NOTE: the `source` parameter can be configured to pull Terraform modules from any Terraform module registry using
+      the `tfr` protocol. The `tfr` protocol expects URLs to be provided in the format
+      `tfr://REGISTRY_HOST/MODULE_SOURCE?version=VERSION`. For example, to pull the `terraform-aws-modules/vpc/aws`
+      module from the public Terraform registry, you can use the following as the source parameter:
+      `tfr://registry.terraform.io/terraform-aws-modules/vpc/aws?version=3.3.0`. If you wish to access a private module
```
brikis98, line 47: NIT:

This seems a little hidden to document a potentially important feature.

Perhaps one of the quick start guides should be updated to use Registry URLs? One benefit of this is the user could start trying / learning Terragrunt without having to create their own modules.

brikis98 (commented):
Thanks for putting this together Yori! 

I want to make sure I understand a few points:

1. The go-getter library doesn't natively support Terraform Registry URLs?
1. Is the custom `tfr` protocol so `go-getter` can _detect_ that the user is trying to us a Registry URL? And in that way, handle that URL with our own `Getter` implementation? The alternative would be to allow users to specify Registry URLs exactly as you would with [Registry `source` URLs in `.tf` code](https://www.terraform.io/docs/language/modules/sources.html#terraform-registry): e.g., `source = "hashicorp/consul/aws"`. Is that just trickier to implement?

docs/_docs/04_reference/config-blocks-and-attributes.md
yorinasub17: Please read the PR description

> Similarly, I opted for using an environment variable provided API token for authenticating to private module registries. This is unconventional (not something Terraform supports), which again risks deviating from familiar APIs of Terraform, but the maintenance overhead of duplicating the logic without importing didn't seem worth it for this initial version. I think the implementation is modular enough that we can add this in later if we deem it necessary.

yorinasub17:
> The go-getter library doesn't natively support Terraform Registry URLs?

Nope it does not. The terraform registry logic in `terraform` source is all implemented outside of `go-getter`.

> Is the custom tfr protocol so go-getter can detect that the user is trying to us a Registry URL? And in that way, handle that URL with our own Getter implementation? The alternative would be to allow users to specify Registry URLs exactly as you would with Registry source URLs in .tf code: e.g., source = "hashicorp/consul/aws". Is that just trickier to implement?

Please read the PR description:

> To keep the implementation simple, I opted for introducing a new protocol string to the source URL. This cleans up the implementation by only requiring to implement a custom Getter and loading it into go-getter without mucking with the Terragrunt config to introduce another field for module versions (like how the Terraform module source parameter works), or implementing complex detector interfaces. This risks deviating from familiar APIs, but I think the maintenance overhead of supporting the Terraform interface is too costly.

internal/tfr/getter.go
yorinasub17: Nope that's not possible given that the return type is object and not pointer:

```
func (u *URL) Query() Values
```

internal/tfr/getter.go
yorinasub17: Ah originally I had the misconception that golang didn't support partial json parsing, but it looks like it does, so went ahead and created a struct: 
b51591c

internal/tfr/getter.go
yorinasub17: Ah that's a good idea. Done: 0ad5c1e

internal/tfr/doc.go
yorinasub17: Done: 95d8610

docs/_docs/04_reference/config-blocks-and-attributes.md
yorinasub17: Took a stab at this: fdbf47f

docs/_docs/04_reference/config-blocks-and-attributes.md
brikis98: Hm, alright. We should prob file a bug to track this and add a comment to the code linking to the bug.

internal/tfr/getter.go
brikis98: Roger

internal/tfr/getter.go
brikis98: Nice! We should update the docs accordingly!

docs/_docs/04_reference/config-blocks-and-attributes.md
brikis98: Nice 👍

brikis98:
> > The go-getter library doesn't natively support Terraform Registry URLs?
> 
> Nope it does not. The terraform registry logic in `terraform` source is all implemented outside of `go-getter`.
> 
> > Is the custom tfr protocol so go-getter can detect that the user is trying to us a Registry URL? And in that way, handle that URL with our own Getter implementation? The alternative would be to allow users to specify Registry URLs exactly as you would with Registry source URLs in .tf code: e.g., source = "hashicorp/consul/aws". Is that just trickier to implement?
> 
> Please read the PR description:
> 
> > To keep the implementation simple, I opted for introducing a new protocol string to the source URL. This cleans up the implementation by only requiring to implement a custom Getter and loading it into go-getter without mucking with the Terragrunt config to introduce another field for module versions (like how the Terraform module source parameter works), or implementing complex detector interfaces. This risks deviating from familiar APIs, but I think the maintenance overhead of supporting the Terraform interface is too costly.
> 
> To expand a bit, the key related point here is **complex detector interfaces**. If we try to support something like `terraform`, then we need to implement the logic to determine if the url is a Terraform Registry URL or a local file path in go-getter's detector subsystem. This would take more work and code, which felt like we could cut out for the initial release of this feature.

Got it, that's what I was trying to understand. May be worth filing a bug for this with this context so we can track it for possible future improvements.

internal/tfr/getter.go
brikis98: Specifically, these docs: https://github.com/gruntwork-io/terragrunt/pull/1767/files#diff-eac47411aff1ec957598f585714c7199a1c1c31a710ef472868bae5abef38cf9R43.

docs/_docs/04_reference/config-blocks-and-attributes.md
yorinasub17: Done:

- https://github.com/gruntwork-io/terragrunt/issues/1771
- 55a1670

internal/tfr/getter.go
yorinasub17: Done: d3531d9

yorinasub17:
> Got it, that's what I was trying to understand. May be worth filing a bug for this with this context so we can track it for possible future improvements.

Done:

- https://github.com/gruntwork-io/terragrunt/issues/1772
- 55a1670

brikis98 (approved):
LGTM! Thanks for getting this done Yori 👍

yorinasub17:
Thanks for review! Merging this in now!

yorinasub17:
@lorengordon oops I thought I responded but I didn't

> Makes me wonder if go-getter would accept this upstream?

Yea it's definitely worth a try, although given the trend in the `terraform` repo, I'm not too hopeful. I was planning on opening an issue on the `go-getter` repo after we get a few cycles of testing in the wild to make sure this approach actually works in real world use cases before bothering them.

antonbabenko:
@yorinasub17 Thanks a lot for this improvement! 🎉

lorengordon:
@yorinasub17 Thanks for responding, sounds good to me! The go-getter project has been moving pretty quick lately, I feel like (I subscribe to all activity on the project), adding features and releases. So I am somewhat hopeful.

giacomoaluigi:
@yorinasub17 great job! Very very useful, thanks a lot!

PR #1775, opened 2021-08-18
Add note in docs clarifying which modules one should be using with terragrunt

yorinasub17:
After seeing https://github.com/gruntwork-io/terragrunt/issues/1774, I realized that it is probably worth clarifying the intent of terragrunt somewhere. For lack of a better place, I placed the note in the reference section.

### DIFF
--- a/docs/_docs/01_getting-started/quick-start.md
+++ b/docs/_docs/01_getting-started/quick-start.md
@@ -49,6 +49,17 @@ terraform {
   source = "tfr:///terraform-aws-modules/vpc/aws?version=3.5.0"
 }
 
+# Indicate what region to deploy the resources into
+generate "provider" {
+  path = "provider.tf"
+  if_exists = "overwrite_terragrunt"
+  contents = <<EOF
+provider "aws" {
+  region = "us-east-1"
+}
+EOF
+}
+
 # Indicate the input values to use for the variables of the module.
 inputs = {
   name = "my-vpc"
@@ -81,6 +92,9 @@ tfr://REGISTRY_DOMAIN/MODULE?version=VERSION
 
 Note that you can omit the `REGISTRY_DOMAIN` to default to the Public Terraform Registry.
 
+The `generate` block is used to inject the provider configuration into the active Terraform module. This can be used to
+customize how Terraform interacts with the cloud APIs, including configuring authentication parameters.
+
 The `inputs` block is used to indicate what variable values should be passed to terraform. This is equivalent to having
 the contents of the map in a tfvars file and passing that to terraform.
 
@@ -89,6 +103,29 @@ documentation](https://terragrunt.gruntwork.io/docs/reference/config-blocks-and-
 sources that terragrunt supports.
 
 You can deploy this example by copy pasting it into a folder and running `terragrunt apply`.
+
+
+## A note about using modules from the registry
+
+The key design of Terragrunt is to act as a preprocessor to convert **shared modules** in the registry into a **root
+module**. In Terraform, modules can be loosely categorized into two types:
+
+* **Root Module**: A Terraform module that is designed for running `terraform init` and the other workflow commands
+  (`apply`, `plan`, etc). This is the entrypoint module for deploying your infrastructure. Root modules are identified
+  by the presence of key blocks that setup configuration about how Terraform behaves, like `backend` blocks (for
+  configuring state) and `provider` blocks (for configuring how Terraform interacts with the cloud APIs).
+* **Shared Module**: A Terraform module that is designed to be included in other Terraform modules through `module`
+  blocks. These modules are missing many of the key blocks that are required for running the workflow commands of
+  terraform.
+
+Terragrunt started off with features that help directly deploy **Root Modules**, but over the years have implemented
+many features that allow you to turn **Shared Modules** into **Root Modules**  by injecting the key configuration
+blocks that are necessary for Terraform modules to act as **Root Modules**.
+
+Modules on the Terraform Registry are primarily designed to be used as **Shared Modules**. That is, you won't be able to
+`git clone` the underlying repository and run `terraform init` or `apply` directly on the module without modification.
+When using modules in the registry, it helps to think about what blocks and resources are necessary to operate the
+module, and translating those into Terragrunt blocks that generate them.
 
 
 ## Key features
--- a/docs/_docs/01_getting-started/quick-start.md
+++ b/docs/_docs/01_getting-started/quick-start.md
@@ -104,58 +104,8 @@ sources that terragrunt supports.
 
 You can deploy this example by copy pasting it into a folder and running `terragrunt apply`.
 
-
-## A note about using modules from the registry
-
-The key design of Terragrunt is to act as a preprocessor to convert **shared service modules** in the registry into a **root
-module**. In Terraform, modules can be loosely categorized into two types:
-
-* **Root Module**: A Terraform module that is designed for running `terraform init` and the other workflow commands
-  (`apply`, `plan`, etc). This is the entrypoint module for deploying your infrastructure. Root modules are identified
-  by the presence of key blocks that setup configuration about how Terraform behaves, like `backend` blocks (for
-  configuring state) and `provider` blocks (for configuring how Terraform interacts with the cloud APIs).
-* **Shared Module**: A Terraform module that is designed to be included in other Terraform modules through `module`
-  blocks. These modules are missing many of the key blocks that are required for running the workflow commands of
-  terraform.
-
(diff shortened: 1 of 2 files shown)


antonbabenko:
The clarification in this PR is clear but I think there can be also an example of how to use submodules with the short syntax introduced recently. This works well for me:

    source = "tfr:///terraform-aws-modules/iam/aws//modules/iam-policy?version=4.2.0"

brikis98 (commented):
Nice 👍 

We might want to clarify terminology, as per https://github.com/gruntwork-io/terragrunt/issues/1774#issuecomment-901834611, to make this a bit clearer.

lorengordon:
I don't think this is entirely accurate. Any nested module with a readme should be considered public and directly consumable by the user:

>Nested modules. Nested modules should exist under the modules/ subdirectory. Any nested module with a README.md is considered usable by an external user. If a README doesn't exist, it is considered for internal use only.

From: https://www.terraform.io/docs/language/modules/develop/structure.html

yorinasub17:
Thanks for the comments everyone! All those points make sense. Will work on clarifying these points!

> I don't think this is entirely accurate. Any nested module with a readme should be considered public and directly consumable by the user:

I mentioned this in https://github.com/gruntwork-io/terragrunt/issues/1774#issuecomment-901980069, but it is not entirely contradictory. I think the terminology clarification proposed by @brikis98 should hopefully make this distinction more accurate.

yorinasub17:
Updated the docs given the latest information from the thread. Hopefully this would be satisfactory to all parties!

docs/_docs/01_getting-started/quick-start.md
```diff
@@ -91,6 +105,29 @@ sources that terragrunt supports.
...
+The key design of Terragrunt is to act as a preprocessor to convert **shared modules** in the registry into a **root
+module**. In Terraform, modules can be loosely categorized into two types:
+
+* **Root Module**: A Terraform module that is designed for running `terraform init` and the other workflow commands
+  (`apply`, `plan`, etc). This is the entrypoint module for deploying your infrastructure. Root modules are identified
+  by the presence of key blocks that setup configuration about how Terraform behaves, like `backend` blocks (for
+  configuring state) and `provider` blocks (for configuring how Terraform interacts with the cloud APIs).
+* **Shared Module**: A Terraform module that is designed to be included in other Terraform modules through `module`
+  blocks. These modules are missing many of the key blocks that are required for running the workflow commands of
+  terraform.
```
brikis98, line 119: I think this misses some important nuance. See my comment here: https://github.com/gruntwork-io/terragrunt/issues/1774#issuecomment-904475528

brikis98 (commented):
Thanks Yori. Big +1 on making this clearer. However, see my latest comments on an attempt to try to capture some important context / nuance.

docs/_docs/01_getting-started/quick-start.md
brikis98: To me, the current docs read as if they are encouraging users to turn "shared modules" into "root modules," but isn't that the very problem we're trying to avoid in #1774? Isn't that behavior precisely what is causing some module maintainers headaches? In that sense, I worry these docs would indeed make things worse.

On the other hand, explaining that Terragrunt is designed to deploy only root modules, but that includes partially configured root modules, takes a little more work, but is more accurate, and may reduce the problems in #1774.

yorinasub17: > To me, the new docs in this PR read as if they are encouraging users to turn "shared modules" into "root modules," but isn't that the very problem we're trying to avoid in #1774? Isn't that behavior precisely what is causing some module maintainers headaches? In that sense, I worry these docs would indeed make things worse.
> On the other hand, explaining that Terragrunt is designed to deploy only root modules, but that includes partially configured root modules, takes a little more work, but is more accurate, and may reduce the problems in #1774.

Something about explaining the service module distinction here was bothering me, but I couldn't quite point out why. After sleeping on it, I think I figured out what was bothering me: this distinction actually puts the burden on module maintainers, and not us, Terragrunt.

Taking a step back, the real problem we are trying to solve here is that Terragrunt limitations are reported as bugs in the registry modules that the module maintainers need to fix, and not something that Terragrunt should address.

Currently, there is no clear cut technical reason why something is a service module from Terragrunt's perspective, like the definition Terraform has for root vs shared. There are technical limitations that prevent Terragrunt from deploying certain types of modules, but that doesn't mean implementing those would suddenly turn a module into a service module as we defined it. AFAIK, there are no service modules in the Terraform public registry. Even our GCP modules, which technically support direct deployment from Terragrunt, are not service modules under our definition.

Without clear technical definitions, it puts a lot of burden on module maintainers to make that distinction. The only distinction they can make is tagging their modules as such (service vs library). I worry that this burden would actually exacerbate the exact problem we are trying to solve, by sending the message that it's the responsibility of the module maintainer to support Terragrunt, and the maintainer is somehow explicitly making the choice not to support Terragrunt when it doesn't work.

That is the exact mentality I want to shift here. I want to push the blame on to Terragrunt, not module maintainers.

Given that, by saying that Terragrunt is trying to convert shared modules on the registry into root modules, we at least make it clear that when it doesn't work, it's Terragrunt's fault, NOT the module maintainers.

brikis98: I agree that we don't have a clear technical definition of what Terragrunt is designed for, and the result is that people either hit a Terragrunt limitation, or try to use it for the wrong thing, and then blame the module maintainer, which we absolutely don't want. So I feel that it's important to do the following:

1. Provide a clear technical definition of what Terragrunt is designed to do. IMO, the "Terragrunt is a canonical Terraform wrapper script for deploying partially configured root modules" as I explained in https://github.com/gruntwork-io/terragrunt/issues/1774#issuecomment-904475528 is the right thing to aim for here, rather than trying to tell folks that Terragrunt is designed to turn _any_ shared module into a root module, which it's not.
1. List out the known limitations of Terragrunt. E.g., Talk about limits of what sorts of partially configured root modules we can deploy, the need to set `type` on input vars, `run-all` limitations, etc. 
1. Explain what to do if you hit a new limitation. Namely, file a bug in this repo, rather than bothering module maintainers. We can even be more explicit here: if the module works with pure Terraform, but not with Terragrunt, that's Terragrunt's fault, and not the module maintainers.

WDYT?

yorinasub17: I still disagree strongly (namely that, whether we like it or not, Terragrunt IS a tool to convert any shared module to root module), but I also don't have the energy to continue this discussion and question myself whether it's on the verge of bike shedding, so I am willing to disagree and commit here.

Did my best to satisfy myself and hopefully your recommendations in (1), (2), and (3) with e304ad6

brikis98: We should probably chat live to make sure we're on the same page here. 

One note: these doc updates are in the quick start guide... But that level of detail doesn't feel particularly quick. Perhaps we should move this new content to a different doc, and update the quick start guide with a "Heads up, not all Registry modules can be deployed with Terragrunt, see `(link to new doc)` for details."

yorinasub17: Good point! Done in 647c546

brikis98 (approved):
Changes LGTM, thanks Yori!

Also, Yori and I talked live, and there's a tricky balance here:

1. On the one hand, Terragrunt's explicit goal is NOT to allow you to deploy _any_ module as a root module. The idea is that you use it to deploy modules that are designed to be root modules, but are partially configured.
1. On the other hand, "partially configured root module" doesn't have a very crisp technical definition, there's no enforcement of what does or doesn't fit that definition in the Terraform or Terragrunt world, and Terragrunt has some known limitations (e.g., around input vars without types or output vars without sensitive) that we don't have a fix for yet.
1. And, of course, users will do just about anything that's possible with your tools/APIs, rather than what's intended.

So here, we're updating the docs to explain (1), but (2) may still lead to issues. Our hope is that flushing out (1) and our stance here will lead to users complaining to us, the Terragrunt maintainers, when they hit issues with (2), rather than the module maintainers, who shouldn't be bothered with that stuff. But due to (3), there are no guarantees. I think this PR is a net improvement, but more work will be necessary before this is totally resolved.

lorengordon:
>So here, we're updating the docs to explain (1), but (2) may still lead to issues. Our hope is that flushing out (1) and our stance here will lead to users complaining to us, the Terragrunt maintainers, when they hit issues with (2), rather than the module maintainers, who shouldn't be bothered with that stuff. But due to (3), there are no guarantees. I think this PR is a net improvement, but more work will be necessary before this is totally resolved.

That's a really good summary!

yorinasub17:
Thanks for review! Going to merge this in given the summary.

PR #1780, opened 2021-08-23
Move TestTerragruntParallelism to serial

yorinasub17:
This PR attempts to stabilize the `TestTerragruntParallelism` test by converting it to a serial test. The precise timing requirements of the test make it sensitive to system load, so it makes sense to try to run it by itself without anything else running on the system.


brikis98 (approved):
Good approach 👍

yorinasub17:
Thanks for review! Merging it in now.

PR #1787, opened 2021-08-26
fix(registry): Add base url if module download url is relative

### LINKED ISSUE #1786: Third Party Terraform Registry provider issues with tfr protocol
Problem:
We're using Gitlab as our Terraform registry and I wanted to use the added support for registries from @yorinasub17.

I stumpled upon the issue, that gitlab returns it's download URL in the X-Terraform-GET Header as a relative path.
The tfr protocol should support relative paths like terraform does [here]( https://github.com/hashicorp/terraform/blob/daad109067931a4b5e5056877ded4c25b9dc686d/internal/registry/client.go#L185).

Regards,
Hown3d

hown3d:
Closes #1786.

### DIFF
--- a/internal/tfr/getter.go
+++ b/internal/tfr/getter.go
@@ -121,7 +121,13 @@ func (tfrGetter *TerraformRegistryGetter) Get(dstPath string, srcURL *url.URL) e
 		Host:   registryDomain,
 		Path:   moduleFullPath,
 	}
-	downloadURL, err := getDownloadURLFromRegistry(ctx, moduleURL)
+
+	terraformGet, err := getTerraformGetHeader(ctx, moduleURL)
+	if err != nil {
+		return err
+	}
+
+	downloadURL, err := getDownloadURLFromHeader(ctx, moduleURL, terraformGet)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


Hown3d:
Any updates on this?

brikis98 (commented):
Thank you for figuring out the issue and submitting a PR!

The code changes LGTM, but is there any way to add some sort of test around them? I worry this behavior could get lost in refactors without some sort of test in place.

One idea is to refactor `getDownloadURLFromRegistry` into two methods, one that does the HTTP GET and one that extracts the URL from the response, and perhaps add a unit test for the latter?

Hown3d:
Sounds good, submitted a commit regarding this!

brikis98 (approved):
LGTM, thanks! I'll kick off tests now.

hown3d:
Seems like a version conflict, go-getter refactored with version 1.5.8 the redact method via this commit:
https://github.com/hashicorp/go-getter/commit/854150ffed2dc250662096b4309b3510a13e0574

brikis98:
Ah, that was fixed in https://github.com/gruntwork-io/terragrunt/pull/1799. Could you pull in the latest from `master`?

hown3d:
> Ah, that was fixed in #1799. Could you pull in the latest from `master`?

Done.

brikis98:
Thanks! I'll re-run tests.

brikis98 (approved):
Tests passed! Merging now.

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.31.9

PR #1802, opened 2021-09-09
Allow setting log level with env var

yorinasub17:
This PR updates the log level routine to allow configuring the terragrunt log level using the env var `TERRAGRUNT_LOG_LEVEL`.

Note that the env var configured log level has special properties beyond the `--terragrunt-log-level` CLI arg, where it allows configuring the terragrunt log level for the global fallback (because it doesn't depend on parsing the CLI args). This is very useful for debugging panic errors where we need the stack trace, because the stack trace is only available in debug mode on the global logger.

### DIFF
--- a/cli/args.go
+++ b/cli/args.go
@@ -138,7 +138,7 @@ func parseTerragruntOptionsFromArgs(terragruntVersion string, args []string, wri
 	strictInclude := parseBooleanArg(args, OPT_TERRAGRUNT_STRICT_INCLUDE, false)
 
 	// Those correspond to logrus levels
-	logLevel, err := parseStringArg(args, OPT_TERRAGRUNT_LOGLEVEL, util.DEFAULT_LOG_LEVEL.String())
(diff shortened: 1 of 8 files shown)


brikis98 (dismissed):
LGTM

denis256:
I'm a little bit concern that there is no test that will verify that `TERRAGRUNT_LOG_LEVEL` is applied, if in future code will be changed that `TERRAGRUNT_LOG_LEVEL` will be ignored and log level set each time to `defaultLogLevel` - tests will still pass

yorinasub17:
I'm not sure how useful that test would be, since it would be more or less a tautology with the const definition of `logLevelEnvVar`. That is, more likely than not, we would modify that const, and then the test will fail, and just update the test to match the value changed to that.

yorinasub17:
Ok tests added! Need another review.

brikis98 (approved):

yorinasub17:

PR #1804, opened 2021-09-10
Multiple include blocks

yorinasub17:
This is the implementation of the feature described in https://github.com/gruntwork-io/terragrunt/pull/1803

This is backward compatible for the most part, except for the expose feature, where you must now index into the label - any prior references to exposed `include` must be updated to `include[""]`. E.g., if the user had:

```hcl
include {
  expose = true
}

inputs = {
  region = include.locals.region
}
```
the `include.locals.region` reference must be updated to `include[""].locals.region`.
```hcl
include {
  expose = true
}

inputs = {
  region = include[""].locals.region
}
```


config/config_helpers.go
```diff
@@ -354,13 +358,28 @@ func findInParentFolders(params []string, include *IncludeConfig, terragruntOpti
 }
 
 // Return the relative path between the included Terragrunt configuration file and the current Terragrunt configuration
-// file
-func pathRelativeToInclude(include *IncludeConfig, terragruntOptions *options.TerragruntOptions) (string, error) {
-	if include == nil {
+// file. Name param is required and used to lookup the relevant import block when called in a child config with multiple
+// import blocks.
+func pathRelativeToInclude(params []string, trackInclude *TrackInclude, terragruntOptions *options.TerragruntOptions) (string, error) {
```
brikis98, line 363: Were you able to get all the relative path helpers to work with multiple `include`?

config/include.go
```diff
@@ -470,3 +501,203 @@ func mergeHooks(terragruntOptions *options.TerragruntOptions, childHooks []Hook,
...
+// {
+//   "include": [
+//     {"path": "foo"},
+//     {
+//       "labeled": {"path": "bar"}
+//     }
+//   ]
+// }
+//
+// For simplicity of implementation, we focus on handling Case 1 and 2, and ignore Case 3. If we see Case 3, we will
```
brikis98, line 606: This comment block is very helpful 👍

config/include.go
```diff
@@ -470,3 +501,203 @@ func mergeHooks(terragruntOptions *options.TerragruntOptions, childHooks []Hook,
...
+		if jsonIsIncludeBlock(typed) {
+			return updateSingleBareIncludeInParsedJSON(parsed, typed)
+		}
+		return nil, false, nil
+	}
+
+	return nil, false, errors.WithStackTrace(IncludeIsNotABlockErr{parsed: includeBlock})
+}
+
+// updateBareIncludeInParsedJSON replaces the include attribute into a block with the label "" in the json. Note that we
```
brikis98, line 662: NIT: is `""` the best name we can use here? I find it a bit hard to read, and suspect for anyone without all the context, it might be even trickier. Some ideas to think through:

1. Perhaps instead of an empty string, we pick a slightly more readable name? E.g., We could go with `"default"` if we make that name "special" and don't allow it to be used for includes names by users. Or we could go with something very unlikely to conflict, such as `"__default__"`. 
1. What if we leave the old syntax as it was: that is, `include.locals.region` would _only_ work with a single, unlabeled `include` block. If you are using labeled `include` blocks, you instead look things up via `includes["label"].locals.region`. That way, it's fully backwards compatible, and there's no awkward empty string to use as a label.

config/include.go
```diff
@@ -470,3 +501,203 @@ func mergeHooks(terragruntOptions *options.TerragruntOptions, childHooks []Hook,
...
+// - The 'path' attribute is a string
+func jsonIsIncludeBlock(jsonData interface{}) bool {
+	typed, isMap := jsonData.(map[string]interface{})
+	if isMap {
+		pathAttr, hasPath := typed["path"]
+		if hasPath {
+			_, pathIsString := pathAttr.(string)
+			return pathIsString
+		}
+		return false
```
brikis98, line 684: NIT: not strictly necessary; could fall through.

docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
```diff
@@ -0,0 +1,345 @@
...
+  assume_role {
+    role_arn = "arn:aws:iam::0123456789:role/terragrunt"
+  }
+}
+EOF
+}
+```
+
+You can then include this in each of your **child** `terragrunt.hcl` files using the `include` block for each
+infrastructure module you need to deploy:
```
brikis98, line 55: Should our default recommendation now be to always include a label on `include`? It's probably good to capture intent and it makes it easier to extend into multiple `include` blocks without introducing a new syntax. In the reference docs, we could keep a mention of the old unlabeled `include` block for backwards compatibility, but label it as deprecated.

docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
```diff
@@ -0,0 +1,345 @@
...
+include "env" {
+  path = "${get_terragrunt_dir()}/../../_env/app.hcl"
+}
+
+inputs = {
+  env = "qa"
+}
+```
+
+### Using read\_terragrunt\_config to DRY parent configurations
```
brikis98, line 205: I mentioned this in the other PR, but we _probably_ want to encourage putting things that you typically change on an env-by-env basis in the env itself so it works better with CI / CD; and putting things that you would change globally (albeit rarely) into an `_env` folder. So for example, env name, instance type, number of instances, `ref` (version) are all things you may want to change in one env, but not another, and you may want CI / CD to be able to detect it based on the file you changed, so it makes more sense to define those params directly in `stage/app/terragrunt.hcl` for stage and `prod/app/terragrunt.hcl` for prod. For everything else, you can put it in `_env`, under the assumption that it changes very rarely, and you can somehow manually trigger CI / CD as necessary for that change.

That does make me think of an alternative approach, which might be nonsense, but I figured it's worth thinking through: would it make sense to version the `terragrunt.hcl` files in `_env`? Ignoring implementation details, imagine for a second that all the `_env` stuff is defined in `infrastructure-modules`; inside the `terragrunt.hcl` files in `_env`, it uses relative paths within `infrastructure-modules`; and the whole repo is versioned as usual. Then, in `infrastructure-live`, you `include` not from a `path` but from a versioned `source` parameter. That allows you to use different versions of `_env` in each actual env, and promote changes from one env to another. In that case, you _could_ define environment-specific values in `_env` as well (e.g., instance type, instance count, etc), as you'd bump the version number in `stage` or `prod` to pick up the new configs, and CI / CD would pick that up.

So, in effect, `infra-live` would become VERY minimal: just a bunch of `terragrunt.hcl` marker files that use `include` to pull in the actual configs from versioned URLs.

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -377,19 +377,26 @@ more about the inheritance properties of Terragrunt in the [Filling in remote st
...
 "Keep your remote state configuration DRY" use case overview.
 
-The `include` block supports the following arguments:
+You can have more than one `include` block, but each one must have a unique label. Note that a bare `include` block with
+no label (`include {}`) is a short hand for an `include` block that uses the label `""` (equivalent to `include "" {}`).
 
+`include` blocks support the following arguments:
+
+- `name` (label): You can define multiple `include` blocks in a single terragrunt config. Each include block
+  must be labeled with a unique name to differentiate it from the other includes. E.g., if you had a block `include
```
brikis98, line 386: I'd be tempted to make labeled `include` blocks the default / recommended option and only mention the unlabeled one as a deprecated pattern kept around for backward compat. Again, the goal is to have one, simple, recommended path to follow.

brikis98 (commented):
Nice work Yori 👍

config/config_helpers.go
brikis98: Oh, also, does this support multiple _levels_ of `include`? I understand that `foo/terragrunt.hcl` can have two `include` blocks now... But what if `foo/terragrunt.hcl` has an `include` pointing to `bar/terragrunt.hcl`, and `bar/terragrunt.hcl` itself has another `include` pointing to `baz/terragrunt.hcl`. Does that work? And if so, do we detect cycles? Either way, we should document explicitly what is/isn't supported.

yorinasub17: > Were you able to get all the relative path helpers to work with multiple include?

Yes. They also now accept an optional parameter to indicate which include the relative helper should apply to, which is required when there are multiple include blocks.

> Oh, also, does this support multiple levels of include?

Not yet. This still limits to a single level. I updated the docs to clarify this: 9d3a702

config/include.go
yorinasub17: Went with option (2): b7f61ec 

I like that because it is backward compatible.

config/include.go
yorinasub17: Ah good point. Removed: b29b8cb

docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
yorinasub17: > That does make me think of an alternative approach, which might be nonsense, but I figured it's worth thinking through: would it make sense to version the terragrunt.hcl files in _env? Ignoring implementation details, imagine for a second that all the _env stuff is defined in infrastructure-modules; inside the terragrunt.hcl files in _env, it uses relative paths within infrastructure-modules; and the whole repo is versioned as usual. Then, in infrastructure-live, you include not from a path but from a versioned source parameter. That allows you to use different versions of _env in each actual env, and promote changes from one env to another. In that case, you could define environment-specific values in _env as well (e.g., instance type, instance count, etc), as you'd bump the version number in stage or prod to pick up the new configs, and CI / CD would pick that up.

Yes I had this exact thought! This actually has another advantage of better handling variable updates across module versions. E.g., in the current setup, there is a limitation where the shared inputs are in a static file shared across `dev`, `stage`, and `prod`. What if a version bump removes a variable, or there is a new required variable, or an old variable type changed? Ideally, you would make the change in the shared variables in `_env`, but then that would break `stage` and `prod`.

yorinasub17: > I mentioned this in the other PR, but we probably want to encourage putting things that you typically change on an env-by-env basis in the env itself so it works better with CI / CD; and putting things that you would change globally (albeit rarely) into an _env folder. So for example, env name, instance type, number of instances, ref (version) are all things you may want to change in one env, but not another, and you may want CI / CD to be able to detect it based on the file you changed, so it makes more sense to define those params directly in stage/app/terragrunt.hcl for stage and prod/app/terragrunt.hcl for prod. For everything else, you can put it in _env, under the assumption that it changes very rarely, and you can somehow manually trigger CI / CD as necessary for that change.

Good point. I updated the docs to use `expose` instead of `read_terragrunt_config` for the module version, and updated `read_terragrunt_config` docs to be something more generic: c034ca9c79d780ff8aff08fadb710f76ed777dba

docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
yorinasub17: Good idea. Done: 9ecd02333edc6cbac0a5327a54883be9d6b4d47e

brikis98 (approved):
This is great work Yori 👍

docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
brikis98: > > That does make me think of an alternative approach, which might be nonsense, but I figured it's worth thinking through: would it make sense to version the terragrunt.hcl files in _env? Ignoring implementation details, imagine for a second that all the _env stuff is defined in infrastructure-modules; inside the terragrunt.hcl files in _env, it uses relative paths within infrastructure-modules; and the whole repo is versioned as usual. Then, in infrastructure-live, you include not from a path but from a versioned source parameter. That allows you to use different versions of _env in each actual env, and promote changes from one env to another. In that case, you could define environment-specific values in _env as well (e.g., instance type, instance count, etc), as you'd bump the version number in stage or prod to pick up the new configs, and CI / CD would pick that up.
> 
> Yes I had this exact thought! This actually has another advantage of better handling variable updates across module versions. E.g., in the current setup, there is a limitation where the shared inputs are in a static file shared across `dev`, `stage`, and `prod`. What if a version bump removes a variable, or there is a new required variable, or an old variable type changed? Ideally, you would make the change in the shared variables in `_env`, but then that would break `stage` and `prod`.

We should def experiment with this as the way to make the Ref Arch super DRY, versioned, etc.

yorinasub17:
Thanks for review! Will merge this in, cut a new release as backward compatible (given how much code change there is to the `include` machinery), and then open a follow up PR to update that function docs.

config/include.go
yorinasub17: > NIT: Does the comment on this function need to be updated then? Could be done in a follow-up PR.

Ah on second look, the docs don't need to update because this is still true. Internally, bare include blocks are still labeled with `""`. The special logic is in the expose feature, where single bare include blocks are injected to the top level of `include` var. So the comment here is accurate, where we are still converting bare include blocks to include with `""` as label, and the special casing happens later downstream (which I commented here: https://github.com/gruntwork-io/terragrunt/blob/f02beeb7497faf7c030864b43fd56e5f5d879b04/config/cty_helpers.go#L193-L198)

brikis98: Ah, roger!

Issue #667, opened 2019-02-28
iam_role introduced in v0.17.2 does not have behave like --terragrunt-iam-role cli flag


conzy:
The documentation says these are equivalent but they are not. It seems like when using the `--terragrunt-iam-role` cli flag or exporting the `TERRAGRUNT_IAM_ROLE` env var the role assumption happens early in the terragrunt flow and sets up my remote state backend correctly. This is the expected behaviour

`aws-vault exec bastion -- terragrunt plan --terragrunt-iam-role arn:aws:iam::1111111111:role/admin`

In this case the bastion account ID is 00000000 and the target account ID is 1111111111

This all works as expected

However when I set the iam_role in the terragrunt.tfvars file like this:

```
terragrunt = {
  iam_role = "arn:aws:iam::1111111111:role/admin"
  terraform {
    source = "git@github.com:foo/foo-modules.git//terragrunt_iam_bastion?ref=v0.8.26"
  }

  # Include all settings from the root terraform.tfvars file
  include = {
    path = "${find_in_parent_folders()}"
  }
}

aws_region    = "eu-west-1"
account_alias = "foo-bastion"
```

The role assumption seems to happen later in the terragrunt flow. SO terragrunt thinks I am configuring a new remote state backend. e.g when calling

`aws-vault exec bastion -- terragrunt plan`

It wants to reconfigure the remote state backend to be that of the bastion account

```
Backend config has changed from 
map[bucket:foo-11111111111-state dynamodb_table:terraform-locks encrypt:%!s(bool=true) key:organizations/bastion/global/iam/terraform.tfstate region:eu-west-1] to 
map[region:eu-west-1 dynamodb_table:terraform-locks encrypt:%!s(bool=true) bucket:foo-0000000000-state key:organizations/bastion/global/iam/terraform.tfstate]
```

The remote state backend configuration in my root terraform.tfvars looks like this:

```
terragrunt = {
  remote_state {
    backend = "s3"
    config {
      encrypt        = true
      bucket         = "foo-${get_aws_account_id()}-state"
      key               = "${path_relative_to_include()}/terraform.tfstate"
      region         = "eu-west-1"
      dynamodb_table = "terraform-locks"
    }
  }
```

Let me know if you need any additional info

conzy:
[...]

PR #1807, opened 2021-09-14
#667 iam_role parsing before evaluation of other HCL blocks

denis256:
Updated HCL parsing first to extract `iam_role` and use obtained value in the evaluation of HCL blocks

Example:
```
$ aws sts get-caller-identity
{
    "UserId": "AAA",
    "Account": "BBB",
    "Arn": "arn:aws:iam::YYYY:/app0"
}

$ cat terragrunt.hcl
iam_role = "arn:aws:iam::ZZZZ:user/app666"

locals {
  var = run_cmd("echo", "get_aws_account_id", "${get_aws_account_id()}")
}
```
Before:
```
$ terragrunt init
....
get_aws_account_id YYYY
....
```
After this PR:
```
$ terragrunt init
...
get_aws_account_id ZZZZ
...
```

Fix for issue: https://github.com/gruntwork-io/terragrunt/issues/667


config/config_partial.go
```diff
@@ -99,6 +99,16 @@ func DecodeBaseBlocks(
...
+	// first decode of IamRole which can be referenced in other blocks
+	contextExtensions := EvalContextExtensions{}
+	flags := terragruntFlags{}
+	err := decodeHcl(hclFile, filename, &flags, terragruntOptions, contextExtensions)
+	if err != nil {
+		return nil, nil, TrackInclude{}, err
+	}
+	if flags.IamRole != nil {
+		terragruntOptions.IamRole = *flags.IamRole
+	}
```
yorinasub17, line 111: I think this should check if `terragruntOptions.IamRole` is already set or not, since we should prefer to use the CLI passed in option over the local configuration.

config/config_partial.go
```diff
@@ -99,6 +99,16 @@ func DecodeBaseBlocks(
 	filename string,
 	includeFromChild *IncludeConfig,
 ) (*cty.Value, *terragruntInclude, TrackInclude, error) {
+	// first decode of IamRole which can be referenced in other blocks
+	contextExtensions := EvalContextExtensions{}
+	flags := terragruntFlags{}
+	err := decodeHcl(hclFile, filename, &flags, terragruntOptions, contextExtensions)
+	if err != nil {
+		return nil, nil, TrackInclude{}, err
+	}
```
yorinasub17, line 108: Doing this parsing here means that we won't honor `iam_role` set on parent configurations that are `include`d. You'll need to implement a special version of `PartialParseConfigFile` that bypasses locals and only parses `include` + `iam_role` to handle this.

yorinasub17 (commented):
This looks like a good start! I think the general idea of parsing `iam_role` early in the process before others makes sense, but there are a few gotchas/edge cases that this implementation doesn't handle. See the comments below for more details.

config/config_partial.go
denis256: Make sense, will look in this dirrection

denis256:
Behaviour with current changes:

```
$ cat terragrunt.hcl
remote_state {
  backend = "local"
  generate = {
    path      = "${get_aws_account_id()}.txt"
    if_exists = "overwrite"
  }
  config = {
    path = "terraform.tfstate"
  }
}

include "issue667" {
  path = "../issue-667-include/vars.hcl"
}
$ cat ../issue-667-include/vars.hcl
iam_role = "arn:aws:iam::AAAA:role/terragrunttest"

$ aws sts get-caller-identity
{
    "UserId": "X",
    "Account": "Y",
    "Arn": "arn:aws:iam::BBBB:user/terragrunttest2"
}

$ terragrunt init
....

$ ls AAAA.txt 
AAAA.txt
```

So `get_aws_account_id()` resolved with value from `iam_role` included from other file

config/config.go
```diff
@@ -545,6 +545,17 @@ func ParseConfigFile(filename string, terragruntOptions *options.TerragruntOptio
 		return nil, err
 	}
 
+	// If IamRole wasn't provided, and if it is config file, try first to decode IamRole
```
brikis98, line 548: NIT: link back to this issue/PR for more context.

config/config.go
```diff
@@ -545,6 +545,17 @@ func ParseConfigFile(filename string, terragruntOptions *options.TerragruntOptio
...
+	if include == nil && terragruntOptions.IamRole == "" {
+		iamConfig, err := PartialParseConfigString(configString, terragruntOptions, include, filename, []PartialDecodeSectionType{TerragruntFlags})
```
brikis98, line 550: What happens if the `iam_role` param depends on a `local` var? Or a `dependency` block? Or anything else that hasn't been parsed yet? In other words, do we need to update the docs on parse order to call out any new limitations here?

test/integration_test.go
```diff
@@ -3621,6 +3622,32 @@ func TestTerragruntVersionConstraints(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	cleanupTerraformFolder(t, TEST_FIXTURE_READ_IAM_ROLE)
+
+	// Execution outputs to be verified
+	stdout := bytes.Buffer{}
+	stderr := bytes.Buffer{}
+
+	// Invoke terragrunt and verify used IAM role
+	err = runTerragruntCommand(t, fmt.Sprintf("terragrunt init --terragrunt-working-dir %s", TEST_FIXTURE_READ_IAM_ROLE), &stdout, &stderr)
```
brikis98, line 3640: Will this actually try to assume the fake IAM role and lead to a permissions error?

denis256: Yes, and it will fail to create state file with IAM role ID

denis256:
Moved to draft since this implementation caused more integration tests to fail

denis256:
My local issue, the same number of tests don't pass with or without this change

denis256:
In the current implementation, it adds a +1 evaluation of HCL code at the beginning, fetching from result `TerragruntFlags / IamRole`  and doing another revaluation.

It will work in cases of remote state definition but may generate some unexpected output in cases like:
```
iam_role = "arn:aws:iam::${local.id}:user/terragrunt"
locals {
  id = "6666"
  var = run_cmd("echo", "get_aws_account_id", "${get_aws_account_id()}")
}
```
It will print:
```
get_aws_account_id AAAA
get_aws_account_id 6666
```
`AAAA` from the first evaluation when `iam_role` was undefined and was loaded default value, and `6666` from the second evaluation when `iam_role` was defined

brikis98:
@yorinasub17 If you're still digging into Terragrunt issues, could you review this one?

config/config.go
```diff
@@ -545,6 +545,18 @@ func ParseConfigFile(filename string, terragruntOptions *options.TerragruntOptio
...
+	// https://github.com/gruntwork-io/terragrunt/issues/667
+	if include == nil && terragruntOptions.IamRole == "" {
+		iamConfig, err := PartialParseConfigString(configString, terragruntOptions, include, filename, []PartialDecodeSectionType{TerragruntFlags})
+		if err != nil {
+			return nil, err
+		}
+		if iamConfig.IamRole != "" {
+			terragruntOptions.IamRole = iamConfig.IamRole
+		}
+	}
```
yorinasub17, line 558: We probably want this logic in `ParseConfigString`, since there are more entrypoints to that function than `ParseConfigFile`.

config/config.go
```diff
@@ -545,6 +545,18 @@ func ParseConfigFile(filename string, terragruntOptions *options.TerragruntOptio
 		return nil, err
 	}
 
+	// Initial evaluation of configuration to load flags like IamRole which will be used for final parsing
+	// https://github.com/gruntwork-io/terragrunt/issues/667
+	if include == nil && terragruntOptions.IamRole == "" {
```
yorinasub17, line 550: I think you need to omit the `include == nil` check here, as this means it won't handle a config where the `iam_role` is defined on the child config that has an `include` defined.

config/config.go
yorinasub17: Also, put this in a separate function to make it easier to extend in the future.

yorinasub17: Oh do we also need to parse and set [iam_assume_role_duration](https://terragrunt.gruntwork.io/docs/reference/config-blocks-and-attributes/#iam_assume_role_duration) here?

yorinasub17 (commented):
Updates LGTM! I think we are very close. Just a few more nits to address and I think we can get this going.

---

The side effect of additional parsing of locals is a potential cause for concern, but I think it would actually work out given that I think most configurations would limit the amount of side effects in locals. What is more of a concern is the backward incompatible nature of this feature.

Currently, before this PR, `get_aws_account_id` will always return the value based on the original credentials. Now, this will change to the value after `iam_role` is assumed. In most use cases, this is probably not a huge issue and could be worked around.

I think where this gets super tricky is if users had a config like the following:

```
iam_role = "arn:aws:iam::${get_aws_account_id()}:role/terragrunttest"
```

Currently as implemented, what would happen is that in the initial parse for `iam_role`, `iam_role` would resolve `get_account_id` based on the original credentials, but when it goes through to do the full parse, `get_account_id` would now be based on the assumed `iam_role`. That means that potentially, the `iam_role` could change depending on which stage of parsing it is in.

This probably isn't a big deal for functions like `get_account_id`, where the value doesn't change even after assuming the role, but what about the following config?

```
# Where 333333333 is another account
iam_role = "arn:aws:iam::333333333:role/from-${get_aws_account_id()}"
```

In this case, `get_aws_account_id()` will return `333333333` in the full parse, which means it will assume a different IAM role when it runs terraform.

That said, I think this is fine for now because I feel like the above case will be rare in practice. But it makes me wonder if this is a fundamentally flawed feature, and whether we should consider separating configuration of cloud authentication out of the main `terragrunt.hcl` so that it can be parsed without side effects.

@brikis98 any thoughts on the above analysis of the implications here?

config/config.go
```diff
@@ -659,6 +653,23 @@ func ParseConfigString(
 	return config, nil
 }
 
+// extractIamRole - extract IAM role details from Terragrunt flags block
+func extractIamRole(configString string, terragruntOptions *options.TerragruntOptions, filename string) error {
```
yorinasub17, line 657: NIT: Rename to `setIAMRole`, to indicate that this function has side effects (modifying attribute on `terragruntOptions`.

yorinasub17 (commented):
Updates LGTM! Just had one nit on the function name. I'll kick off a regression build.

Note that before merge, I'd like to make sure @brikis98 has a chance to sanity check the backward incompatible implications.

yorinasub17 (approved):
Build passed and last update LGTM, so only thing remaining is sanity check from Jim!

docs/_docs/04_reference/built-in-functions.md
```diff
@@ -591,19 +591,22 @@ Output:
...
+uuid2 289ff371-8021-54c6-2254-72de9d11392a
+uuid3 baa19863-1d99-e0ef-11f2-ede830d1c58a
 carrot
 ```
 **Notes:**
   * Output contains only once `carrot` and `potato`, because other invocations got cached, caching works for all sections
-  * Output contains twice `uuid1` and `uuid2` because during HCL evaluation each `run_cmd` in `locals` section is evaluated twice, and value is cached under different key since `uuid()` add random value in key
-  * Output contains three times `uuid3` - 2 prints because `uuid3` was declared in `locals`, once because it is declared in `inputs`
+  * Output contains multiple times `uuid1` and `uuid2` because during HCL evaluation each `run_cmd` in `locals` is evaluated multiple times and random argument generated from `uuid()` save cached value under different key each time
+  * Output contains multiple times `uuid3`, +1 more output comparing to `uuid1` and `uuid2` - because `uuid3` is declared in locals and inputs which add one more evaluation
```
brikis98, line 609: NIT: are these changes related to this PR?

brikis98:
> The side effect of additional parsing of locals is a potential cause for concern, but I think it would actually work out given that I think most configurations would limit the amount of side effects in locals. What is more of a concern is the backward incompatible nature of this feature.
> 
> Currently, before this PR, `get_aws_account_id` will always return the value based on the original credentials. Now, this will change to the value after `iam_role` is assumed. In most use cases, this is probably not a huge issue and could be worked around.
> 
> I think where this gets super tricky is if users had a config like the following:
> 
> ```
> iam_role = "arn:aws:iam::${get_aws_account_id()}:role/terragrunttest"
> ```
> 
> Currently as implemented, what would happen is that in the initial parse for `iam_role`, `iam_role` would resolve `get_account_id` based on the original credentials, but when it goes through to do the full parse, `get_account_id` would now be based on the assumed `iam_role`. That means that potentially, the `iam_role` could change depending on which stage of parsing it is in.
> 
> This probably isn't a big deal for functions like `get_account_id`, where the value doesn't change even after assuming the role, but what about the following config?
> 
> ```
> # Where 333333333 is another account
> iam_role = "arn:aws:iam::333333333:role/from-${get_aws_account_id()}"
> ```
> 
> In this case, `get_aws_account_id()` will return `333333333` in the full parse, which means it will assume a different IAM role when it runs terraform.

Oof, you're right, this does feel messy. 

> But it makes me wonder if this is a fundamentally flawed feature, and whether we should consider separating configuration of cloud authentication out of the main `terragrunt.hcl` so that it can be parsed without side effects.

Yea, I've been having lots of second thoughts about:

- iam_role
- iam_assume_role_duration
- get_aws_account_id()
- get_aws_caller_identity_arn()
- get_aws_caller_identity_user_id()

And for that matter, the way we've implemented Auto Init for S3 buckets, DynamoDB tables, etc.

They all feel like partial implementations that don't fully solve the problem; they also feel too cloud-specific and are a pain to maintain.

I wonder if what we really need is:

1. Either remove all auth from Terragrunt and require that it's handled externally. Or, alternatively, have first-class auth functionality, but handled separately than the rest of the config. This would need a design.
1. Either remove all "data fetching" from Terragrunt and require that it's handled externally. Or, alternatively, have first-class data source support: conceptually, I imagine a separate Terraform module with a bunch of `data` sources and output variables, we run `apply` on that first, and then expose it to your config similar to a `dependency` block; you can then use all that fetched data in your config, and we run `apply` on the module you have in `source`.
1. Either remove all "Auto Init" for remote state for Terragrunt and require that it's handled externally. Or, alternatively, have first-class remote state support: conceptually, I imagine a separate Terraform module that creates your remote state resources (e.g., S3 bucket, DynamoDB table) and provides info about them via output variables, we run `apply` on that first, and then expose it to your config similar to a `dependency` block; you can then use that data in your config, and we run `apply` on the module you have in `source`.

But these are all larger changes, prob not super relevant to this PR.

For this PR, the key question, I suppose, is how many users combine the `iam_role` config with the various data functions:

- get_aws_account_id()
- get_aws_caller_identity_arn()
- get_aws_caller_identity_user_id()

My guess is that 99% of the time, they are used separately, and the intention is (a) assume the specified IAM role and (b) fetch the requested data while authenticated as that IAM role. In the <1% of use cases where the two are combined together, the result will likely be unpredictable. Not sure if our partial parse can detect that and show a warning? If not, then perhaps we document it, merge this PR, and file issues to do the larger clean up mentioned above?

docs/_docs/04_reference/built-in-functions.md
yorinasub17: Yes. It's the result of the change in the parsing order.

yorinasub17:
That makes sense.

> If not, then perhaps we document it, merge this PR, and file issues to do the larger clean up mentioned above?

Let's move forward with documenting this. @denis256 can you update the docs for the following functions and mention the caveat that the value will change depending on the parsing stage?

- get_aws_account_id
- get_aws_caller_identity_arn
- get_aws_caller_identity_user_id

The source is here: https://github.com/gruntwork-io/terragrunt/blob/master/docs/_docs/04_reference/built-in-functions.md

Separately, I'll go ahead and file an issue capturing the above point.

yorinasub17 (approved):
Docs updates LGTM! I'm going to merge this in now given our discussion, and release as backward incompatible so we can better highlight the behavior change.

Issue #1808, opened 2021-09-14
Terragrunt should support marking Sensitive outputs as sensitive to support transforming shared modules to root modules


yorinasub17:
Originally reported as https://github.com/gruntwork-io/terragrunt/issues/1774

This allows supporting deployment of modules that don't have sensitive derived `output` marked as `sensitive`.

yorinasub17:
RFC with design proposal for how to implement this: https://github.com/gruntwork-io/terragrunt/pull/1809

PR #1809, opened 2021-09-14
RFC: transform blocks for handling terragrunt limitations

yorinasub17:
This a proposed design for addressing https://github.com/gruntwork-io/terragrunt/issues/1808. I wanted to write this down before it hits brain rot and I forget about it.

[Easy reading link](https://github.com/gruntwork-io/terragrunt/blob/yori-rfc-inputoutputmod/docs/_docs/05_rfc/variable_output_modification.md).

Note that I think this feature does indeed belong in Terragrunt. As the discussion in https://github.com/gruntwork-io/terragrunt/issues/1774 revealed, the vast majority of modules in the registry do not support Terragrunt, nor do the module maintainers want to (**it goes without saying that that is perfectly ok**).

However, this does not mean that the modules in the registry are NOT service modules. Use cases of the module by users may show that there are modules in the registry that actually can be deployed directly as a service module, despite not being labeled as such, but have the `variable` and `output` limitations that terragrunt currently doesn't support. For those modules, it is fairly painful for users to have to wrap and repeat all the variables and outputs of the underlying module.

The feature proposed in this RFC should help support this, and I believe we can implement this with relatively little effort.

cc @lorengordon since you indicated being curious about how we could address this problem.

### DIFF
--- a/docs/_docs/05_rfc/variable_output_modification.md
+++ b/docs/_docs/05_rfc/variable_output_modification.md
@@ -89,7 +89,7 @@ output "length_my_list" {
 Using Terragrunt with this module will run into the following issue:
 
 - Because the output `my_password_hashed` is not marked as sensitive, terraform will error out.
-- `my_list` is missing the type definition, so the input from `terragrunt` will be interpretted as a string. This means
+- `my_list` is missing the type definition, so the input from `terragrunt` will be interpreted as a string. This means
   that the output `length_my_list` will be the string length, and not the list length.
 
 We will need to transform these variables and outputs. We will introduce the `transform` block to handle this. The
--- a/docs/_docs/05_rfc/variable_output_modification.md
+++ b/docs/_docs/05_rfc/variable_output_modification.md
@@ -1,0 +1,146 @@
+---
+layout: collection-browser-doc
+title: variable and output modifications
+category: rfc
+categories_url: rfc
+excerpt: variable and output modifications
+tags: ["rfc", "contributing", "community"]
+order: 505
+nav_title: Documentation
+nav_title_link: /docs/
+---
+
+# variable and output modifications
+
+**STATUS**: In proposal
+
+
+## Background
+
+Terragrunt over the years has evolved to adapt to deploying shared modules from any source as root modules by injecting
+various blocks and terraform code to support the deployment. In Terraform, modules can be loosely categorized into two types:
+
+* **Root Module**: A Terraform module that is designed for running `terraform init` and the other workflow commands
+  (`apply`, `plan`, etc). This is the entrypoint module for deploying your infrastructure. Root modules are identified
+  by the presence of key blocks that setup configuration about how Terraform behaves, like `backend` blocks (for
+  configuring state) and `provider` blocks (for configuring how Terraform interacts with the cloud APIs).
+* **Shared Module**: A Terraform module that is designed to be included in other Terraform modules through `module`
+  blocks. These modules are missing many of the key blocks that are required for running the workflow commands of
+  terraform.
+
+Note that Terragrunt is not designed to deploy any **Shared Module**. That is, modules that are necessary to be composed
+with other modules should not really be deployed with Terragrunt. Terragrunt further distinguishes shared modules
+between **service modules** and **modules**:
+
+* **Shared Service Module**: A Terraform module that is designed to be standalone and applied directly. These modules
+  are not root modules in that they are still missing the key blocks like `backend` and `provider`, but aside from that
+  do not need any additional configuration or composition to deploy. For example, the
+  [terraform-aws-modules/vpc](https://registry.terraform.io/modules/terraform-aws-modules/vpc/aws/latest) module can be
+  deployed by itself without composing with other modules or resources.
+* **Shared Module**: A Terraform module that is designed to be composed with other modules. That is, these modules must
+  be embedded in another Terraform module and combined with other resources or modules. For example, the
+  [consul-security-group-rules
+  module](https://registry.terraform.io/modules/hashicorp/consul/aws/latest/submodules/consul-security-group-rules)
+
+At its core, Terragrunt is designed to add the necessary ingredients to convert **Shared Service Modules** into **Root
+Modules** so that they can be deployed with `terraform apply`. Features like `generate` and `remote_state` support this
+transition by injecting the necessary blocks that support directly invoking `terraform`.
+
+Note that the distinction between **Shared Service Modules** and **Shared Modules** is subtle, and oftentimes is not a
+clear cut technical difference. However, there are technical limitations in the current Terragrunt implementation that
+prevents deploying certain shared modules:
+
+- Every complex input must have a `type` associated with it. Otherwise, Terraform will interpret the input that
+  Terragrunt passes through as `string`. This includes `list` and `map`.
+- Derived sensitive outputs must be marked as `sensitive`. Refer to the [terraform tutorial on sensitive
+  variables](https://learn.hashicorp.com/tutorials/terraform/sensitive-variables#reference-sensitive-variables) for more
+  information on this requirement.
+
+Terraform only enforces these restrictions on the **Root Module**. This means that there are some shared modules on the
+registry that Terragrunt can not deploy. Note that the lack of these parameters may not by itself indicate that they are
+**Shared Modules**. That is, these modules may be **Shared Service Modules** by design, but because they are
+only designed for use with Terraform, they may not set the `type` or `sensitive` flags on the `variable` and `output`
+blocks, preventing usage as a transformed root module unless those inputs and outputs are configured.
+
+
+## Proposed solution
+
+To handle this, this RFC proposes a new block: `transform`. Here is an example:
+
+Consider a module that has the following:
+
+```hcl
+variable "my_password" {
+  sensitive = true
+}
+
+variable "my_list" {}
+
+# NOTE: this must be marked as sensitive since it is derived from a sensitive variable
+output "my_password_hashed" {
+  value = base64sha256(var.my_password)
+}
+
+output "length_my_list" {
+  value = length(var.my_list)
+}
+```
+
+Using Terragrunt with this module will run into the following issue:
+
+- Because the output `my_password_hashed` is not marked as sensitive, terraform will error out.
+- `my_list` is missing the type definition, so the input from `terragrunt` will be interpretted as a string. This means
+  that the output `length_my_list` will be the string length, and not the list length.
+
+We will need to transform these variables and outputs. We will introduce the `transform` block to handle this. The
+following `terragrunt.hcl` configuration indicates the necessary transformations to `variable` and `output` to support
+deployment:
+
+```hcl
+transform {
+  # Sub blocks are 'variable' or 'output', to indicate what terraform block needs to be transformed. Next, the label
+  # should match with the corresponding variable label defined in the terraform module.
+  # Each subblock matches the underlying terraform block, and under the hood, the attributes and blocks are merged into
+  # the terraform module.
+  # For example, in this first variable subblock, the 'type' attribute is merged into the upstream terraform module
+  # before terragrunt invokes terraform.
+  variable "my_list" {
+    type = list(string)
+  }
+  output "my_password_hashed" {
+    sensitive = true
+  }
+}
+```
+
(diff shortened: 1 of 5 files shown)


docs/_docs/05_rfc/variable_output_modification.md
```diff
@@ -0,0 +1,152 @@
...
+- Because the output `my_password_hashed` is not marked as sensitive, terraform will error out.
+- `my_list` is missing the type definition, so the input from `terragrunt` will be interpretted as a string. This means
+  that the output `length_my_list` will be the string length, and not the list length.
+
+We will need to transform these variables and outputs. We will introduce the `transform` block to handle this. The
+following `terragrunt.hcl` configuration indicates the necessary transformations to `variable` and `output` to support
+deployment:
+
+```hcl
+transform {
```
lorengordon, line 100: Would this be an attribute of the `terraform` block, since it would be specifically related to the `source` argument?

docs/_docs/05_rfc/variable_output_modification.md
```diff
@@ -0,0 +1,152 @@
...
+  # should match with the corresponding variable label defined in the terraform module.
+  # Each subblock matches the underlying terraform block, and under the hood, the attributes and blocks are merged into
+  # the terraform module.
+  # For example, in this first variable subblock, the 'type' attribute is merged into the upstream terraform module
+  # before terragrunt invokes terraform.
+  variable "my_list" {
+    type = list(string)
+  }
+  output "my_password_hashed" {
+    sensitive = true
```
lorengordon, line 111: How would collisions be handled? E.g. say `sensitive = true` already existed but the user specifically wanted to override it with `sensitive = false`?

docs/_docs/05_rfc/variable_output_modification.md
lorengordon: Theoretically, this could also rewrite resources? 🤯

docs/_docs/05_rfc/variable_output_modification.md
```diff
@@ -0,0 +1,152 @@
...
+  `my_password_hashed` output will have `sensitive = true` attribute set.
+- Terragrunt invokes `terraform apply` on the modified module.
+
+In this way, Terragrunt can convert the underlying shared module into a root module that can be deployed directly.
+Under the hood, this transformation is implemented in the same way that the `aws-provider-patch` command works. This
+should be feasible by doing the following:
+
+- Scan all `.tf` files in the directory.
+- For each file found, parse using the `hclwrite` parser.
+- Walk the AST, looking for `variable` or `output` blocks that match the `transform` sub blocks.
```
lorengordon, line 134: Theoretically, this could also rewrite resources? 🤯

brikis98 (commented):
Thanks for putting together the RFC. A few thoughts:

1. Does this RFC encourage anti-patterns? For example, it seems like not specifying the `type` on an input variable, especially when the type is _not_ a `string`, is borderline a bug. I understand it works in Terraform, but it feels like a leftover from the early days, when there weren't explicit types, and like something we'd want to discourage, rather than support via a first-class feature. Similarly, what is the use case for not putting `sensitive` on output variables that contain sensitive data, even in a shared module? Perhaps there is some legitimate reason for doing that, but it's not obvious to me what that is, and I don't know if adding a first-class feature to work around that is a good direction. _Update:_ thinking about this more, is this just to avoid having to put `sensitive = true` all over the place, and instead allowing Terraform to track sensitive values internally, and you only need the explicit label if that output is in a root module and therefore, could be written to a log file?
1. Thinking more through the points above, is the primary goal to make it possible to use legacy modules (i.e., those that are out of date with the latest Terraform functionality) or those that a user might not have access to change?
1. Could overwriting the source code result in the changes accidentally getting written back to version control? In particular, IIRC, if you had a `source` URL pointing to a local file path, we (or more accurately, go-getter) used to use a symlink, so if we modify the code in `.terragrunt-cache`, it will actually modify it in the original source folder. Not sure if that still happens.
1. Does this create maintenance headaches? E.g., Let's say you add `transform` blocks for a few variables, and then at some point, the maintainer of the module changes those variables: e.g., they add `type` to some of them, rename others, etc. Will it be obvious what's going on with these transform blocks? Is this debuggable?

docs/_docs/05_rfc/variable_output_modification.md
yorinasub17: I kind of like it being a separate block, given that the `terraform` block is more about how terragrunt calls terraform, and transforming the code is a different operation. In that regard, it is more similar to `generate` than `terraform`.

docs/_docs/05_rfc/variable_output_modification.md
yorinasub17: In my mind, this was more like a shallow merge operation, where each block shallow merges up, with the "default" being what's in the terraform code. Updated to clarify this: f2d8e44

docs/_docs/05_rfc/variable_output_modification.md
yorinasub17: Yes in theory, although I think that would be fairly complex. I think at that point, forking the underlying module is a better approach.

yorinasub17:
> Update: thinking about this more, is this just to avoid having to put sensitive = true all over the place, and instead allowing Terraform to track sensitive values internally, and you only need the explicit label if that output is in a root module and therefore, could be written to a log file?

Yes, this is my understanding. I think the issue is that it's hard to see this in shared modules, and thus know when you have to mark outputs as sensitive or not, since terraform doesn't really give you tools to do this (e.g., `validate` doesn't reveal this), as it is only necessary to do in the root module.

> Does this RFC encourage anti-patterns?

I'm not sure this RFC by itself would encourage anti-patterns. Ultimately, many module developers would much rather focus on supporting the core terraform use case rather than additionally Terragrunt, and these issues are arguably only a problem if you use Terragrunt to deploy those modules (since it turns the shared module where these aren't issues, into a root module where they are). I also think this is a maintenance headache if you have direct access to the module itself. It seems much easier to maintain to have this logic directly in the module, than in terragrunt.hcl.

So it really is an escape hatch for cases where you can't modify the module.

That said, you do bring up a good point that these should really only be used for modules on the public registry, because otherwise it is desirable to mark `input` with `type` and `output` with `sensitive` in the module itself. Perhaps we can address this by restricting the block to only work when the `source` is `tfr`?

> Thinking more through the points above, is the primary goal to make it possible to use legacy modules (i.e., those that are out of date with the latest Terraform functionality) or those that a user might not have access to change?

I was primarily thinking about the latter. More specifically, about modules on the public registry (which relates to the point above about restricting to `tfr` sources).

> Could overwriting the source code result in the changes accidentally getting written back to version control? In particular, IIRC, if you had a source URL pointing to a local file path, we (or more accurately, go-getter) used to use a symlink, so if we modify the code in .terragrunt-cache, it will actually modify it in the original source folder. Not sure if that still happens.

Ah yes that is true... Another reason to do the `tfr` restriction...

> Does this create maintenance headaches? E.g., Let's say you add transform blocks for a few variables, and then at some point, the maintainer of the module changes those variables: e.g., they add type to some of them, rename others, etc. Will it be obvious what's going on with these transform blocks? Is this debuggable?

This is a good point. In terms of debuggability, I think we can add a feature to `validate-inputs` (or a new command) that will write out the transformations that terragrunt will make. E.g., a command that writes out:

```
INFO: the following transformations will be made
- variable "existing-variable"
    - `type=string` => `type=list(string)`
- variable "var-with-no-type"
    - NIL => `type=map(string)`

WARN: the following transform blocks did not apply to this module
- variable "non-existant-variable"
- output "non-existant-output"
```

---

Given the above, what are your thoughts on restricting this to `tfr` source only?

lorengordon:
>Given the above, what are your thoughts on restricting this to tfr source only?

Please no! We have a use case where we fork every module so we "own" it, but that's just so we can trust the source and the tags we've already inspected (managing supply chain risk)! We prefer not to modify the forks because that is a maintenance headache, and instead inspect upstream changes and update the forks as needed. Being able to modify the source on the fly when needed would be handy!

docs/_docs/05_rfc/variable_output_modification.md
lorengordon: Hmm, and that makes me wonder what kind of terragrunt config/feature might it be to support multiple terraform blocks...

yorinasub17:
Ah that makes sense @lorengordon . Hmm I'll have to think through what mechanisms we can put in place in terragrunt to support both concerns.

Off the top of my head, a compromise is to make it a soft restriction that is overrideable with config/cli arg. Something that basically indicates "I know what I am doing" to terragrunt. E.g., maybe by default it is restricted to `tfr`, but then there could be an undocumented but supported `allow_protocols` attribute on the `transform` block that allows you to override that restriction for use cases like the one you have where you know the module source is an unmodified fork.

I know this can be annoying to you, but at least it makes it explicit that terragrunt doesn't encourage usage of the `transform` block with sources that you actually have control over.

lorengordon:
>There could be an undocumented but supported `allow_source_protocols` attribute on the `transform` block that allows you to override that restriction for use cases like the one you have where you know the module source is an unmodified fork.

That would be fine. We certainly have no problem managing config. But I would suggest documenting it and putting a warning in the doc. Default value could be `allow_source_protocols = ['tfr']`.

dmattia:
If my understanding is correct, some submodules cannot be used as terragrunt root modules, but all modules that work as terragrunt root modules could be used as submodules in a terraform root module. Would it then make sense to make it so that terragrunt does not download the `terraform.source` parameter, but just references it via a slim module definition with a single `module` call?

A bit more in depth...

Currently, if I use module `foo` from the module registry, I see that `.terragrunt-cache/abc/def/main.tf` would have the `main.tf` file from the `foo` module, as defined remotely, just with extra files like `remote_state_terragrunt.tf` in the same directory to wrap the module into a root module.

Would things break if instead, terragrunt created a `.terragrunt-cache/abc/def/main.tf` file that was not downloaded from any remote source, but had something like:

```terraform
module "terragrunt_root" {
  source = "terraform-aws-modules/foo/aws"

  inputs = "resolved from terragrunt pre-processing"
}

output "foo_output_1" { ...generated for each output in the foo module... }
```

There would need to be some path mapping done for local modules, but my (probably naive) understanding is that this would make the terragrunt/terraform compatibility stronger, as all modules could be supported this way.

As far as the `output` generation goes, I've used https://github.com/hashicorp/terraform-config-inspect in `terragrunt-atlantis-config` and it was pretty easy to use, and looks like it would make this generation pretty straightforward.

dmattia:
@yorinasub17 Thank you for that explanation!

I think some of these may have solutions (like `terragrunt state mv` modifying the args before passing to `terraform state mv`, and maybe we still do download the `terraform.source` module, but put it under `.terragrunt-cache/abc/def/terragrunt_managed_submodule/`, which would allow for keeping `generate` support), but the state one is certainly tricky, as overriding that sounds problematic from a maintenance POV, even if terraform v1.0.0 promises more stability.

Overall I like this proposal, and think it's great that it's all backwards compatible. I do have one question though: Where would the line be drawn between using `transform` blocks and adding `override.tf` files in the directory next to the `terragrunt.hcl` file? I have a few spots in our codebase where we use a remote `terraform.source`, and then customize the module via override files

yorinasub17:
> Where would the line be drawn between using transform blocks and adding override.tf files in the directory next to the terragrunt.hcl file?

The line would be whether you need to __add__ blocks to the existing module or __modify__ blocks. Some blocks in terraform support combining (e.g., the `terraform` block), but most blocks can't be merged together in `terraform`. For example, having two `variable` blocks with the same label will throw an error in terraform.

In essence, the `transform` block would primarily be used as a monkey patching escape hatch to modify existing blocks in the target module (hence the name `transform` as opposed to `generate`).

docs/_docs/05_rfc/variable_output_modification.md
```diff
@@ -0,0 +1,168 @@
...
+
+output "length_my_list" {
+  value = length(var.my_list)
+}
+```
+
+Using Terragrunt with this module will run into the following issue:
+
+- Because the output `my_password_hashed` is not marked as sensitive, terraform will error out.
+- `my_list` is missing the type definition, so the input from `terragrunt` will be interpretted as a string. This means
```
zackproser, line 92:
```suggestion
- `my_list` is missing the type definition, so the input from `terragrunt` will be interpreted as a string. This means
```

yorinasub17:
Closing as stale

Issue #1499, opened 2021-01-20
Support dependency without outputs


damon-atkins:
terragrunt_v0.26.7 calls dependency with  terraform output -json which response with
```
{}
```

Terragrunts messages are
```
[terragrunt] [dir] 2021/01/20 18:14:13 Running command: terraform output -json
[terragrunt] 2021/01/20 18:14:37 xxxxxxx/terragrunt.hcl is a dependency of  yyyyy/terragrunt.hcl but detected no outputs. Either the target module has not been applied yet, or the module has no outputs. If this is expected, set the skip_outputs flag to true on the dependency block.
[terragrunt] 2021/01/20 18:14:37 Unable to determine underlying exit code, so Terragrunt will exit with error code 1
```
Consider `{}` a valid response of no `output` however error as its expected to have output without waiting 30 seconds.

brikis98:
Any chance you could share the code so we have some way to repro this?

yorinasub17:
After looking at this again, I think the issue here is that terragrunt `dependency` always assumes there is some output provided by the module, since that is the point of the explicit dependency (to string outputs to the dependent module).

I think the feature request here is to support this semantic of `dependency` where you only want to add it for the purpose of deployment ordering with `xxx-all` commands. I updated the issue to reflect this.

For now, the workaround is to use the `dependencies` block for these modules instead, which is only used for the purpose of dependency ordering as opposed to dependency passing.

yorinasub17:
Another workaround: if you know you are not using any outputs from `dependency` block, you can also set `skip_outputs = true`. With `skip_outputs = true`, terragrunt would still include the dependency in the graph, but will skip pulling out the outputs with `terragrunt output`, thereby bypassing the check for empty output.

damon-atkins:
It seems when terraform outputs  `{}` only "terragrunt appears to timeout waiting for the output" instead of moving on. I am surprised its hitting the timeout limit.    Maybe its looking  for
 ```
{
 }
```
When their  is no output variables in json

**Comment by damon-atkins** [comment]
It **could** also be that terraform has changed its behavior when their are no outputs and you request json
```
$ terraform output

Warning: No outputs found

The state file either has no outputs defined, or all the defined outputs are
empty. Please define an output in your configuration with the `output` keyword
and run `terraform refresh` for it to become available. If you are using
interpolation, please verify the interpolated value is not empty. You can use
the `terraform console` command to assist.

$ terraform output -json
{}
```

[...]

PR #1811, opened 2021-09-16
#1499 Added logic to skip outputs verification in case of module initialization

denis256:
Added logic to skip outputs verification in case of module initialization, if the dependency fail to generate outputs it is handled as error code from terraform invocation

Example: https://github.com/denis256/terragrunt-test-1499

Before this change:
```
$ terragrunt run-all init

ERRO[0011] Module /work/app has finished with an error: /work/module1/terragrunt.hcl is a dependency of /work/app/terragrunt.hcl but detected no outputs. Either the target module has not been applied yet, or the module has no outputs. If this is expected, set the skip_outputs flag to true on the dependency block.  prefix=[/work/app] 
ERRO[0011] 1 error occurred:
        * /work/module1/terragrunt.hcl is a dependency of /work/app/terragrunt.hcl but detected no outputs. Either the target module has not been applied yet, or the module has no outputs. If this is expected, set the skip_outputs flag to true on the dependency block.
 
ERRO[0011] Unable to determine underlying exit code, so Terragrunt will exit with error code 1 
```

After this change:
```
$ terragrunt run-all init
...
...
$ terragrunt run-all apply
Apply complete!
```

Fix for missing outputs from https://github.com/gruntwork-io/terragrunt/issues/1499

### DIFF
--- a/config/dependency.go
+++ b/config/dependency.go
@@ -309,7 +309,11 @@ func getTerragruntOutputIfAppliedElseConfiguredDefault(dependencyConfig Dependen
 		if err != nil {
 			return nil, err
 		}
-
+		// Skip output variables verification in case of initialization
+		// https://github.com/gruntwork-io/terragrunt/issues/1499
+		if terragruntOptions.AutoInit && isEmpty {
+			return outputVal, err
+		}
 		if !isEmpty && dependencyConfig.shouldMergeMockOutputsWithState(terragruntOptions) {
 			stateOutputsMap, err := parseCtyValueToMap(*outputVal)
 			if err != nil {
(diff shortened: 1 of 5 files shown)


infraredgirl:
Thanks for the PR! The code LGTM and I tested and verified that it fixes the bug. I'll kick the tests off.

infraredgirl:
Looks like `terraform fmt` needs to be run on the newly added fixture files. Everything else looks good!

denis256:
Updated formatting for test fixture, wondering if this can be somehow verified locally

infraredgirl:
Excellent, thanks! Re-running the tests now.

You can run `pre-commit` locally to verify, e.g. `pre-commit run --all-files`.

infraredgirl (approved):
Tests passed, so I'm going to approve and merge!

infraredgirl:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.32.2

Issue #1818, opened 2021-09-20
Enhance evaluation of maps members within locals


daniel-anova:
When using the following code example:

```hcl
locals {
  values = {
    this = "this"

    // error: "Could not evaluate all locals in block."
    name     = local.values.this
    location = "eastus"
  }
}

terraform {
  source  = "tfr://registry.terraform.io/bcochofel/resource-group/azurerm?version=1.4.1"
}

generate "provider" {
  path      = "provider.tf"
  if_exists = "overwrite_terragrunt"
  contents  = <<EOF
provider "azurerm" {
  features {}
}
EOF
}

inputs = local.values
```

it produces the error "Could not evaluate all locals in block" when referencing a sub value of a map within the map itself. Moving that value outside the map works normally.

Could evaluation of maps within locals be evaluated in the same more "robust" way locals are evaluated to allow the above to work without error?

Alternatively could erros be clearer in pointing which block is failing to evaluate and why.

yorinasub17:
I'm not sure this is feasible given the parsing tools we have from the `hcl` library and the current approach. AFAIK, to implement this, we will need to overhaul the parsing mechanism to use a lazy evaluation approach like `terraform`, which is a major transition (not to mention more difficult to maintain).

I think we will consider a community PR to implement this, but will most likely not implement it ourselves.

> Alternatively could erros be clearer in pointing which block is failing to evaluate and why.

Good point. Opened PR to enhance error message: https://github.com/gruntwork-io/terragrunt/pull/1819

Now the following will be logged for the following config:

_config_
```hcl
locals {
  hello = var.foo
  values = {
    this        = "this"
    name     = local.values.this
    location = "eastus"
  }
}
```

_error log_
```
ERRO[0000] Not all locals could be evaluated:
ERRO[0000]      - hello (Can't evalute expression at terragrunt.hcl:2,11-18 because of non-local variable reference (var is not defined))
ERRO[0000]      - values (Can't evalute expression at terragrunt.hcl:3,12-9,4 because local reference 'values' is not evaluated)
ERRO[0000] Could not evaluate all locals in block.
```

PR #1819, opened 2021-09-20
Show reason for why a local could not be evaluated

yorinasub17:
Partially addresses https://github.com/gruntwork-io/terragrunt/issues/1818 by enhancing the error message.


config/locals.go
```diff
@@ -192,49 +195,68 @@ func attemptEvaluateLocals(
...
+		if rootName == "include" {
 			continue
 		}
 
 		// We can't evaluate any variable other than `local`
-		if var_.RootName() != "local" {
-			return false
+		if rootName != "local" {
+			reason := fmt.Sprintf(
+				"Can't evalute expression at %s because of non-local variable reference (%s is not defined)",
```
brikis98, line 228: NIT: might be clearer to say something like:

```suggestion
				"Can't evaluate expression at %s: you can only reference other local variables here, but it looks like you're referencing something else (%s is not defined)",
```

Also, fixes a typo in the word "evalute".

config/locals.go
```diff
@@ -192,49 +195,68 @@ func attemptEvaluateLocals(
...
+			)
+			return false, reason
 		}
 
 		// If we can't get any local name, we can't evaluate it.
 		localName := getLocalName(terragruntOptions, var_)
 		if localName == "" {
-			return false
+			reason := fmt.Sprintf(
+				"Can't evalute expression at %s because local var name can not be determined",
```
brikis98, line 239:
```suggestion
				"Can't evaluate expression at %s because local var name can not be determined",
```

brikis98: Should the error message show the expression that can't be evaluated and/or the variable name?

config/locals.go
```diff
@@ -192,49 +195,68 @@ func attemptEvaluateLocals(
...
+			)
+			return false, reason
 		}
 
 		// If the referenced local isn't evaluated, we can't evaluate this expression.
 		_, hasEvaluated := evaluatedLocals[localName]
 		if !hasEvaluated {
-			return false
+			reason := fmt.Sprintf(
+				"Can't evalute expression at %s because local reference '%s' is not evaluated",
```
brikis98, line 249:
```suggestion
				"Can't evaluate expression at %s because local reference '%s' is not evaluated",
```

brikis98: What does "is not evaluated" mean here? That is, _why_ is that local reference not evaluated?

brikis98 (commented):
Kudos on better error messages 👍 

Key thing to think about is, will a user seeing that message understand what it means, and know what to do to fix the issue?

config/locals.go
yorinasub17: That's what I want, but the code to take the AST representation and turn it into string is unfortunately very messy. I couldn't find a public function in the `hcl` library to do that, so we will basically have to implement it from scratch, which means walking the AST and switch casing the tokens. That didn't seem worth implementing here.

PR #1821, opened 2021-09-23
render-json command

yorinasub17:
This introduces the `render-json` command, which will render out a json representation of the full, interpretted `terragrunt.hcl` config (that is, with all the includes merged, dependencies resolved/interpolated, function calls executed, etc).

The primary use case of this would be to run [OPA](https://github.com/open-policy-agent/opa) on the rendered json to enforce policies on `terragrunt.hcl` (since it doesn't support running against HCL inputs).

As an alternative, one could use [hcl2json](https://github.com/tmccombs/hcl2json), but this won't interpret the blocks, which makes it harder to write policies. E.g., if you want to enforce restrictions on the module source for the final example in https://terragrunt.gruntwork.io/docs/features/keep-your-terragrunt-architecture-dry/, you need to be able to see the rendered `source` that involves `include` and `local` interpolation. A naive hcl to json translation would make the `terraform.source` attribute look like:

```
{
  "terraform": {
    "source": "${include.env.locals.source_base_url}?ref=v0.2.0"
  }
}
```

which won't be possible to match against OPA rules. So a terragrunt native feature like `render-json` proposed here which outputs the fully interpreted terragrunt config to run OPA rules against is a lot more practical.

As a bonus, this tool could be a nice way to debug `terragrunt.hcl` as well, especially for those that have multiple `include` and `read_terragrunt_config` calls.

### DIFF
--- a/cli/args.go
+++ b/cli/args.go
@@ -210,6 +210,10 @@ func parseTerragruntOptionsFromArgs(terragruntVersion string, args []string, wri
 	opts.Check = parseBooleanArg(args, optTerragruntCheck, os.Getenv("TERRAGRUNT_CHECK") == "true")
 	opts.HclFile = filepath.ToSlash(terragruntHclFilePath)
 	opts.AwsProviderPatchOverrides = awsProviderPatchOverrides
+	opts.JSONOut, err = parseStringArg(args, optTerragruntJSONOut, "terragrunt_rendered.json")
+	if err != nil {
+		return nil, err
+	}
 
 	return opts, nil
 }
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -52,6 +52,7 @@ const (
 	optTerragruntOverrideAttr                = "terragrunt-override-attr"
 	optTerragruntLogLevel                    = "terragrunt-log-level"
 	optTerragruntStrictValidate              = "terragrunt-strict-validate"
+	optTerragruntJSONOut                     = "terragrunt-json-out"
 )
 
 var allTerragruntBooleanOpts = []string{
@@ -83,6 +84,7 @@ var allTerragruntStringOpts = []string{
 	optTerragruntOverrideAttr,
 	optTerragruntLogLevel,
 	optTerragruntStrictValidate,
+	optTerragruntJSONOut,
 }
 
 const CMD_INIT = "init"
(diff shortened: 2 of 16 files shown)


cli/args.go
```diff
@@ -210,6 +210,10 @@ func parseTerragruntOptionsFromArgs(terragruntVersion string, args []string, wri
 	opts.Check = parseBooleanArg(args, OPT_TERRAGRUNT_CHECK, os.Getenv("TERRAGRUNT_CHECK") == "true")
 	opts.HclFile = filepath.ToSlash(terragruntHclFilePath)
 	opts.AwsProviderPatchOverrides = awsProviderPatchOverrides
+	opts.JSONOut, err = parseStringArg(args, OPT_TERRAGRUNT_JSON_OUT, "terragrunt_rendered.json")
```
brikis98, line 213: Super NIT, but would `JsonOut` be the more Go idiomatic var name?

cli/cli_app.go
```diff
@@ -211,6 +214,7 @@ COMMANDS:
    graph-dependencies    Prints the terragrunt dependency graph to stdout
    hclfmt                Recursively find hcl files and rewrite them into a canonical format.
    aws-provider-patch    Overwrite settings on nested AWS providers to work around a Terraform bug (issue #13018)
+   render-json           Render the final interpretted terragrunt config (including running functions and merging included configs) as json. This is useful for enforcing policies using static analysis tools like Open Policy Agent, or for debugging your terragrunt config.
```
brikis98, line 217:
```suggestion
   render-json           Render the final terragrunt config, with all variables, includes, and functions resolved, as json. This is useful for enforcing policies using static analysis tools like Open Policy Agent, or for debugging your terragrunt config.
```

brikis98 (approved):
Ah, very nice! 👍

cli/args.go
yorinasub17: Go recommends using [consistent casing for acronyms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms), and since JSON is an acronym (JavaScript Object Notation), I think this is more idiomatic.

yorinasub17:
Accepted your suggestion so need another review!

cli/args.go
brikis98: Roger

brikis98 (dismissed):
LGTM!

yorinasub17:
Sorry I need one more review cycle due to a merge conflict with https://github.com/gruntwork-io/terragrunt/pull/1833 - had to rename the parameter to use `optTerragrunt` instead of `OPT_TERRAGRUNT`.

brikis98 (approved):
LGTM

yorinasub17:
Thanks for review. Merging now!

Issue #299, opened 2017-09-27
Destroy command should automatically warn about dependencies


brikis98:
Terragrunt allows you to specify modules you depend via [`dependency`](https://terragrunt.gruntwork.io/docs/reference/config-blocks-and-attributes/#dependency) or via . When a user runs `terraform destroy`, we should automatically find modules that depend on the current module, and warn the user that destroying the current one may cause issues for the others. 

I believe this would work as follows:

1. Use `git rev-parse --show-toplevel` to find the root folder in the Git repo. If it fails, either because this isn't a Git repo or `git` isn't installed, see if the current module has an `include` block, and use the path within as the root folder. If there's on `include` block, perhaps log a warning that no checks can be done.

1. Go into the root folder and find all Terragrunt modules within that folder.

1. Scan the `dependency` and `dependencies` block of each of those modules.

1. If any modules list the current module as a dependency, warn the user that deleting the current module may cause problems for those other modules and prompt for a yes/no confirmation to proceed. Skip this confirmation if the user specifies the `--terragrunt-non-interactive` flag.

cstavro:
In non-interactive mode I'd like to see this fail with a non-zero exit code unless you provide another switch that explicitly permits the forced behaviour.

brikis98:
Hm, that would be quite backwards incompatible with current behavior, but it's probably the correct thing to do from a technical perspective... Not sure the best way to go on that one.

brikis98:
@denis256 could you look into this one?

yorinasub17:
This was implemented in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.33.0

PR #1823, opened 2021-09-24
#299 Added confirmation before destroy with list of dependent modules

denis256:
Updated terragrunt flow to show a list of dependent modules and confirmation of action before destroy
Example (https://github.com/denis256/terragrunt-test-1815) :

app1, app2 -> service-discovery-core -> vpc
```
$ tree
.
.
├── applications
│   └── gateway
│       ├── app1
│       │   ├── main.tf
│       │   └── terragrunt.hcl
│       └── app2
│           ├── main.tf
│           └── terragrunt.hcl
├── README.md
├── shared
│   ├── service-discovery-core
│   │   ├── main.tf
│   │   └── terragrunt.hcl
│   └── vpc
│       ├── main.tf
│       ├── terragrunt.hcl
└── terragrunt.hcl

$ cd shared/service-discovery-core
$ terragrunt destroy
Detected dependent modules:
/raid1/tests/terragrunt-test-1815/applications/gateway/app1
/raid1/tests/terragrunt-test-1815/applications/gateway/app2
WARNING: Are you sure you want to continue? (y/n) y
...
Destroy complete!

...

$ cd shared/vpc
$ terragrunt destroy
Detected dependent modules:
/raid1/tests/terragrunt-test-1815/shared/service-discovery-core
WARNING: Are you sure you want to continue? (y/n)
...


$ cd applications/gateway/app2
$ terragrunt destroy
...
Do you really want to destroy all resources?
  Terraform will destroy all your managed infrastructure, as shown above.
  There is no undo. Only 'yes' will be accepted to confirm.

  Enter a value: 

```

Particularities of implementation:
* `--terragrunt-non-interactive` automatically approve the action, to be consistent with other commands
* detection of root directory from where to find dependencies is based on `git rev-parse --show-toplevel`, if it fails `include` directive is used

Fix for the issue: https://github.com/gruntwork-io/terragrunt/issues/299

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -665,6 +665,29 @@ func runTerragruntWithConfig(originalTerragruntOptions *options.TerragruntOption
 	}
 
 	return runActionWithHooks("terraform", terragruntOptions, terragruntConfig, func() error {
+		if terragruntOptions.TerraformCommand == "destroy" {
+			modules, err := configstack.FindWhereWorkingDirIsIncluded(terragruntOptions)
+			if err != nil {
+				terragruntOptions.Logger.Warnf("Failed to detect where module is used %v", err)
+			}
+			if len(modules) != 0 {
+				terragruntOptions.ErrWriter.Write([]byte("Detected dependent modules:\n"))
+				for _, module := range modules {
+					terragruntOptions.ErrWriter.Write([]byte(fmt.Sprintf("%s\n", module.Path)))
+				}
+
+				var prompt string
+				prompt = "WARNING: Are you sure you want to run `terragrunt destroy` ? Destroying current module may cause issues for dependent modules"
+				shouldRun, err := shell.PromptUserForYesNo(prompt, terragruntOptions)
+				if err != nil {
+					return err
+				}
+				if shouldRun == false {
+					return nil
+				}
+			}
+		}
+
 		runTerraformError := runTerraformWithRetry(terragruntOptions)
 
 		var lockFileError error
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -666,26 +666,9 @@ func runTerragruntWithConfig(originalTerragruntOptions *options.TerragruntOption
 
 	return runActionWithHooks("terraform", terragruntOptions, terragruntConfig, func() error {
 		if terragruntOptions.TerraformCommand == "destroy" {
-
-			modules, err := configstack.FindWhereWorkingDirIsIncluded(terragruntOptions, terragruntConfig)
-			if err != nil {
-				terragruntOptions.Logger.Warnf("Failed to detect where module is used %v", err)
-			}
-			if len(modules) != 0 {
-				terragruntOptions.ErrWriter.Write([]byte("Detected dependent modules:\n"))
-				for _, module := range modules {
-					terragruntOptions.ErrWriter.Write([]byte(fmt.Sprintf("%s\n", module.Path)))
-				}
-
-				var prompt string
-				prompt = "WARNING: Are you sure you want to run `terragrunt destroy` ? Destroying current module may cause issues for dependent modules"
-				shouldRun, err := shell.PromptUserForYesNo(prompt, terragruntOptions)
-				if err != nil {
-					return err
-				}
-				if shouldRun == false {
-					return nil
-				}
+			allowDestroy := warnDependentModulesBeforeDestroy(terragruntOptions, terragruntConfig)
+			if !allowDestroy {
+				return nil
 			}
 		}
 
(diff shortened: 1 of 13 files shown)


shell/prompt.go
```diff
@@ -55,3 +55,16 @@ func PromptUserForYesNo(prompt string, terragruntOptions *options.TerragruntOpti
...
+	if err != nil {
+		terragruntOptions.Logger.Error(err)
+		return errors.WithStackTrace(err)
+	}
+	if n != len(prompt) {
+		terragruntOptions.Logger.Errorln("Failed to write data")
+		return errors.WithStackTrace(err)
+	}
+	return nil
+}
```
yorinasub17, line 70: Is this function needed? Can't this simplify to `terragruntOptions.ErrWriter.Write([]byte(prompt))`?

The reason I think so is that the second check for `n != len(prompt)` is dead code because of the spec of `io.Writer.Write` (from https://github.com/golang/go/blob/1537f14/src/io/io.go#L87-L94):

> // Write must return a non-nil error if it returns n < len(p)

yorinasub17 (commented):
Mostly LGTM! I had a bunch of comments around code organization, but the core logic makes sense to me.

We should definitely document this behavior somewhere though. Not sure where the best place is, but probably as a subsection of https://terragrunt.gruntwork.io/docs/features/execute-terraform-commands-on-multiple-modules-at-once/ after `Dependencies between modules` makes the most sense for now.

cli/cli_app.go
```diff
@@ -683,6 +692,29 @@ func runTerragruntWithConfig(originalTerragruntOptions *options.TerragruntOption
...
+func confirmActionWithDependentModules(terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig) bool {
+	modules, err := configstack.FindWhereWorkingDirIsIncluded(terragruntOptions, terragruntConfig)
+	if err != nil {
+		terragruntOptions.Logger.Warnf("Failed to detect where module is used %v", err)
+		return true
+	}
+	if len(modules) != 0 {
+		terragruntOptions.ErrWriter.Write([]byte("Detected dependent modules:\n"))
+		for _, module := range modules {
+			terragruntOptions.ErrWriter.Write([]byte(fmt.Sprintf("%s\n", module.Path)))
```
yorinasub17, line 705: We should do error handling here:
```suggestion
		if _, err := terragruntOptions.ErrWriter.Write([]byte("Detected dependent modules:\n")); err != nil {
			terragruntOptions.Logger.Error(err)
			return false
		}
		for _, module := range modules {
			if _, err := terragruntOptions.ErrWriter.Write([]byte(fmt.Sprintf("%s\n", module.Path))); err != nil {
				terragruntOptions.Logger.Error(err)
				return false
			}
```

config/include.go
```diff
@@ -264,6 +264,8 @@ func (targetConfig *TerragruntConfig) Merge(sourceConfig *TerragruntConfig, terr
 	if sourceConfig.Inputs != nil {
 		targetConfig.Inputs = mergeInputs(sourceConfig.Inputs, targetConfig.Inputs)
 	}
+
+	targetConfig.ProcessedIncludes = sourceConfig.ProcessedIncludes
```
yorinasub17, line 268: I think this can be removed. By moving the setting of the `ProcessedIncludes` to where it is now, it should be set after all the merge operations have been handled. Same with the equivalent setting for `DeepMerge`.

docs/_docs/02_features/execute-terraform-commands-on-multiple-modules-at-once.md
```diff
@@ -306,6 +306,8 @@ In the example above it'll generate this graph
 Note that this graph shows the dependency relationship in the direction of the arrow (top down), however terragrunt will run the action
 in reverse order (bottom up)
 
+**Note:** During execution of `destroy` command, Terragrunt will try to find all dependent modules and show a confirmation prompt with a list of all detected dependencies, because once resources will be destroyed, any commands on dependent modules will fail with missing dependencies. For example, if will be executed `destroy` on `redis`, will be asked to confirm action because `backend-app` depends on `redis`.
```
yorinasub17, line 309:
```suggestion
**Note:** During execution of `destroy` command, Terragrunt will try to find all dependent modules and show a confirmation prompt with a list of all detected dependencies, because once resources will be destroyed, any commands on dependent modules will fail with missing dependencies. For example, if `destroy` was called on the `redis` module, you will be asked to confirm the action because `backend-app` depends on `redis`. You can avoid the prompt by using `--terragrunt-non-interactive`.
```

yorinasub17 (commented):
Updates LGTM! Just a few more minor nits. Separately, I'll kick off a build so you get one test cycle.

yorinasub17:
Ah build failed, because the `ProcessedIncludes` need to be tracked like `IsPartial` in the test. Should be able to fix by updating [this line](https://github.com/gruntwork-io/terragrunt/blob/master/config/config_as_cty_test.go#L202):

```
case "IsPartial", "ProcessedIncludes":
```

config/include.go
```diff
@@ -367,6 +367,8 @@ func (targetConfig *TerragruntConfig) DeepMerge(sourceConfig *TerragruntConfig,
 	for key, val := range sourceConfig.GenerateConfigs {
 		targetConfig.GenerateConfigs[key] = val
 	}
+
+	targetConfig.ProcessedIncludes = sourceConfig.ProcessedIncludes
```
yorinasub17, line 371: Also remove this line, for the same reason it was removed from the `Merge` function.

yorinasub17 (commented):
One last nit and I think this is good to merge!

yorinasub17 (approved):
This is good now! Thanks for the contribution. Will merge this in and cut a release!

dudicoco:
Hi @denis256 and @yorinasub17, thank you for this feature!

There are a couple of issues with this new addition:
1. The destroy plan output no longer shows when running `terragrunt destroy` and there are existing dependent modules, it will only show output for `Detected dependent modules: ...`
2. The dependent modules detection does not run when running `terragrunt run-all plan -destroy`

Issue #1829, opened 2021-09-28
No binaries for v0.32.5 release


Chuxel:
We have some automated processes that get the latest version of terragrunt for use in developer environments that are bombing at the moment because of this issue.

siunb:
Same here

yorinasub17:
Ah apologies for the error here. The binaries are available now! Also opened a PR for a more permanent fix here: https://github.com/gruntwork-io/terragrunt/pull/1830

siunb:
Thanks so much @yorinasub17

PR #1830, opened 2021-09-28
Don't link build->deploy with test

yorinasub17:
We have a few flaky tests that fail occasionally. This isn't much of a concern during PR stage, but it causes issues during releases where it stops the binary from being built + uploaded.

This PR updates the workflow so that the `build` and `deploy` steps happen regardless of test failure.

Permanent fix for https://github.com/gruntwork-io/terragrunt/issues/1829

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ jobs:
           shell: powershell.exe
           no_output_timeout: 45m
           command: |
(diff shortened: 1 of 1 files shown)


brikis98 (approved):
LGTM!

yorinasub17:
Thanks for review! Merging this in now - hopefully this stabilizes our asset release pipeline.

PR #1837, opened 2021-09-29
Revert #1811 to restore useful error message when dependency output is missing

yorinasub17:
#1811 implements logic where we always allow empty outputs unless `auto-init` is disabled. Auto-init shouldn't matter for whether or not empty outputs are allowed. I think the original intention of the PR was to only allow empty outputs when `terragrunt init` is run.

While I can fix this `if` logic to implement the original intention, I don't think that logic is correct either. This is because the intention of the `dependency` block is to allow direct references to the dependency output. Consider the following simplified example:

```
dependency "dep" {
  config_path = "../dep"
}

inputs = {
  input = dependency.dep.outputs.id
}
```

With the logic in #1811, this returns a confusing error message because terragrunt can't resolve `dependency.dep.outputs.id`. The original error message before #1811, was more helpful (specifically, the part about if this is intended, they should set `skip_outputs = true` on the `dependency` block):

```
xxxxxxx/terragrunt.hcl is a dependency of  yyyyy/terragrunt.hcl but detected no outputs. Either the target
module has not been applied yet, or the module has no outputs. If this is expected, set the skip_outputs
flag to true on the dependency block.
```

If one wants to conditionally skip outputs based on the command, then the solution is to use `mock_outputs` and include those commands in the `mock_outputs_allowed_terraform_commands`. So for example, if the intention is to skip outputs just for `init` (which is what #1811 tried to implement), they can do:

```
dependency "dep" {
  config_path = "../dep"
  mock_outputs = {
    id = "foo"
  }
  mock_outputs_allowed_terraform_commands = ["init"]
}

inputs = {
  input = dependency.dep.outputs.id
}
```

as covered in https://terragrunt.gruntwork.io/docs/features/execute-terraform-commands-on-multiple-modules-at-once/#unapplied-dependency-and-mock-outputs.

So this PR reverts #1811.

---

In regards to the issue about timing brought up in #1499, please read https://github.com/gruntwork-io/terragrunt/issues/1499#issuecomment-930414538

### DIFF
--- a/config/dependency.go
+++ b/config/dependency.go
@@ -345,12 +345,9 @@ func getTerragruntOutputIfAppliedElseConfiguredDefault(dependencyConfig Dependen
 		return dependencyConfig.MockOutputs, nil
 	}
 
-	// In case of init, and empty output, skip generating error about missing outputs
-	// https://github.com/gruntwork-io/terragrunt/issues/1499
-	if dependencyConfig.shouldGetOutputs() && terragruntOptions.AutoInit {
-		return nil, nil
-	}
-
+	// At this point, we expect outputs to exist because there is a `dependency` block without skip_outputs = true, and
+	// returning mocks is not allowed. So return a useful error message indicating that we expected outputs, but they
+	// did not exist.
 	err := TerragruntOutputTargetNoOutputs{
 		targetConfig:  targetConfig,
 		currentConfig: currentConfig,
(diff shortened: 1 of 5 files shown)


yorinasub17:
@denis256 FYI

denis256 (approved):

brikis98 (approved):
Oh, I forgot we had `skip_outputs`. LGTM!

yorinasub17:
Thanks for review. Merging now.

PR #1858, opened 2021-10-11
Fix issues with IAM role configuration after merging multiple features

yorinasub17:
We recently merged multiple features and fixes that ended up interacting poorly. Specifically, the feature to expose IAM role session name did not take into account the issues around assuming the IAM role for the AWS related functions. To address this, this PR:

- Replaces the individual IAM related option strings/ints into a single struct to make some of the logic easier.
- Refactors places that reference the IAM role from various places to use the new `IAMRoleOptions` struct to ensure all the information is set.
- Track all the IAM related options, not just the role arn.

In the process, the following bugs were identified and fixed:

- There was a bug in the logic for `--terragrunt-assume-role-session-name`, where the default session name was always used when the CLI option was not passed in due to the session name being set on the options struct with the default.
- There was a bug in the logic for assuming IAM roles in the remote state initialization, where it did not use the session duration or the session name that was set on the CLI.
- There was a bug in the logic for assuming IAM roles for the terragrunt internal helper functions, where it did not use the session duration or the session name from the config or CLI.
- There was a bug in the logic for dependency, where it did not use the session duration or the session name from the config or CLI.

### DIFF
--- a/aws_helper/config.go
+++ b/aws_helper/config.go
@@ -96,6 +96,8 @@ func getSTSCredentialsFromIAMRoleOptions(sess *session.Session, iamRoleOptions o
 	optFns = append(optFns, func(p *stscreds.AssumeRoleProvider) {
 		if iamRoleOptions.AssumeRoleDuration > 0 {
 			p.Duration = time.Second * time.Duration(iamRoleOptions.AssumeRoleDuration)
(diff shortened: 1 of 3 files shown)


yorinasub17:

PR #1859, opened 2021-10-11
Update CODEOWNERS

rhoboat:


yorinasub17:
Merging this in so that the proper folks get pinged.

Issue #1863, opened 2021-10-13
Deprecate, remove, and replace `dependencies` block


yorinasub17:
With the introduction of `dependency` blocks, the functionality of `dependencies` is less useful as it is an implicit list of configurations. To avoid confusion, we should deprecate and remove the `dependencies` block.

Before removing however, there is one use case for `dependencies` block where it allows a concise representation of multiple dependencies. We should consider replacing the `dependencies` block by adding support for `for_each` to `dependency` blocks. E.g.:

```
dependency "all_deps" {
  for_each    = ["../vpc", "../mysql", "../redis"]
  config_path = each.value
}
```

## Motivation

Currently, we use the old `dependencies` block for the `configstack` graph representation. This was done to speed up development of the `dependency` blocks, but has led to a few headaches, the latest one being https://github.com/gruntwork-io/terragrunt/issues/1852. We should take this opportunity to remove the `dependencies` block and consolidate the dependency handling logic around the internal representation of `dependency` blocks.

tjstansell:
I just ran across this ticket while investigating long startup times for terragrunt.  If I understand correctly, the two would provide the same functional and performance impact:
```
dependencies {
  paths = ["../vpc", "../mysql", "../redis"]
}

dependency {
  for_each     = ["../vpc", "../mysql", "../redis"]
  config_path  = each.value
  skip_outputs = true
}

Issue #1852, opened 2021-10-10
graph-dependencies does not follow include dependencies


dmikalova:
I have been using `include` blocks to DRY repeated dependencies, however graph-dependencies does not follow these transitive dependencies. Example configs:

example/terragrunt.hcl:
```
dependency "container_registry" {
  config_path = "../../../dmikalova/container-registry/"
}

include "kubernetes" {
  path = find_in_parent_folders("kubernetes.hcl")
}

include "terraform" {
  path = find_in_parent_folders("terraform.hcl")
}

inputs = {
  container_registry_info = dependency.container_registry.outputs.info
}

terraform {
  source = "git@gitlab.com:e91e63/terraform-digitalocean-kubernetes.git//modules/service-account-container-registry/"
}
```
kubernetes.hcl:
```
dependency "kubernetes" {
  config_path = "${get_parent_terragrunt_dir()}/kubernetes/"
}

generate "kubernetes" {
  contents  = file(find_in_parent_folders("terraform/providers/kubernetes.tf"))
  if_exists = "overwrite"
  path      = "provider-kubernetes.tf"
}

inputs = {
  k8s_info = dependency.kubernetes.outputs.info
}
```

When graphing these, I would expect `example` to be a dependent on `kubernetes`, but `graph-dependencies` shows that it is only dependent on `container-registry`.

dmikalova:
Thanks for the quick fix, the graph now looks as expected!

PR #1864, opened 2021-10-13
Fix bug with dependency blocks in multi include

yorinasub17:
https://github.com/gruntwork-io/terragrunt/issues/1852 revealed a bug where the dependency graph constructed with `run-all` incorrectly ignored some included dependencies. This is due to the following implementation details that mixed poorly:

- We were using the old `Dependencies` list (equivalent to the `dependencies` block) to track dependencies in the stack graph. This meant that in the graph building parsing stage, we would convert the list of `dependency` blocks into `dependencies` block.
- `dependencies` block were shallow merged. This meant that later includes would override the `dependencies` list, instead of concatenating the lists.

This wasn't really an issue with single includes, as it was rare to pack dependencies in the `include` config. However, this issue started to surface with the introduction of multiple includes taking the pattern of fragmenting dependency blocks to DRY. This caused a single dependency in the child to override the combined dependency list from the parents!

To fix this, we special case the shallow merge logic to deep merge the `dependencies` list. This is unfortunately **backward incompatible** for configurations that use `dependencies` blocks. However, with multi-includes, I think this can be worked around. That is, the shallow merge of `dependencies` blocks is useful to nullify the `dependencies` in the parent config for some of the children. You can mimic this behavior with multi include by breaking out just the `dependencies` block into a separate config that is conditionally included. That is, all children that don't override the `dependencies` block include the config that defines it, while the child that is overriding the `dependencies` block does not.

It is more verbose, but I think I'd rather focus on improving the UX for users that use `dependency` blocks than those that use `dependencies`.

I think the "right" fix for this, by the way, is https://github.com/gruntwork-io/terragrunt/issues/1863, but that will take time to implement, so this PR is the intermediary, less good but good enough solution.

### DIFF
--- a/docs/_docs/04_reference/config-blocks-and-attributes.md
+++ b/docs/_docs/04_reference/config-blocks-and-attributes.md
@@ -17,6 +17,9 @@ where `terragrunt.hcl` is mentioned you can always use `terragrunt.hcl.json` ins
 
 The following is a reference of all the supported blocks and attributes in the configuration file:
 
+  - [Blocks](#blocks)
+  - [Attributes](#attributes)
+
 ## Blocks
 
 - [terraform](#terraform)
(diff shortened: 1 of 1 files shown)


yorinasub17:
Thanks for the review! Will merge and release this as backward incompatible release.

> BTW, in the future, should we deprecate and eventually remove dependencies? Seems like dependency blocks are a superset?

Hah had this exact thought in mind! See this issue I filed yesterday: https://github.com/gruntwork-io/terragrunt/issues/1863

Issue #514, opened 2018-07-06
terragrunt state pull throws an "unexpected end of JSON input" error


cswingler:
I'm not sure what's going on here, but if you run:

```
% terragrunt init
<snip>
Successfully configured the backend "s3"! Terraform will automatically
use this backend unless the backend configuration changes.
<snip>
Initializing provider plugins...

Terraform has been successfully initialized!
% terragrunt state pull > terraform.tfstate
[terragrunt] [/Users/chrisswingler/src/infra_terraform] 2018/07/06 12:03:12 Running command: terraform --version
[terragrunt] 2018/07/06 12:03:13 Reading Terragrunt config file at /Users/chrisswingler/Sprout/src/infra_terraform/terraform.tfvars
[terragrunt] 2018/07/06 12:03:13 unexpected end of JSON input
[terragrunt] 2018/07/06 12:03:13 Unable to determine underlying exit code, so Terragrunt will exit with error code 1
```

Terragrunt pukes on something not being valid JSON. I don't know _what_ it's puking on, though.  `terraform state pull` works fine. 

Debug output:

```
➜  infra_terraform [default] git:(terragrunt_k8s) ✗  TERRAGRUNT_DEBUG=trye terragrunt state pull > terraform.tfstate
[terragrunt] [/Users/chrisswingler/Sprout/src/infra_terraform] 2018/07/06 13:16:51 Running command: terraform --version
[terragrunt] 2018/07/06 13:16:51 Reading Terragrunt config file at /Users/chrisswingler/Sprout/src/infra_terraform/terraform.tfvars
[terragrunt] 2018/07/06 13:16:51 *json.SyntaxError unexpected end of JSON input
/private/tmp/terragrunt-20180703-977-jbnzes/terragrunt-0.15.2/src/github.com/gruntwork-io/terragrunt/remote/terraform_state_file.go:79 (0x142fbbb)
/private/tmp/terragrunt-20180703-977-jbnzes/terragrunt-0.15.2/src/github.com/gruntwork-io/terragrunt/remote/terraform_state_file.go:71 (0x142faed)
/private/tmp/terragrunt-20180703-977-jbnzes/terragrunt-0.15.2/src/github.com/gruntwork-io/terragrunt/remote/terraform_state_file.go:56 (0x142f925)
[...]

PR #1885, opened 2021-10-22
feat: Prefer remote tfstate over local tfstate, if present

andreykaipov:
This addresses issue #514:

Upon a local to remote tfstate migration, Terraform will leave an empty
`terraform.tfstate` file in the working directory. Subsequent
invocations on the state should then refer to the `terraform.tfstate`
file in the data directory (`.terraform`) which points to the remote
state. By prefering the data directory when parsing the tfstate,
Terragrunt behaves similarly to Terraform, and no longer errors out on
the empty `terraform.tfstate` file.

---

Reproduction of issue #514:

```console
❯ echo 'resource "null_resource" "blah" {}' > main.tf
❯ terraform init
❯ terraform apply -auto-approve
```

Now we must generate and migrate to a remote backend with Terragrunt. I use an HTTP backend:

```hcl
remote_state {
  generate = {
    path      = "backend.tf"
    if_exists = "overwrite"
  }

  backend = "http"

  config = {
    username = get_env("TF_BACKEND_USERNAME")
    password = get_env("TF_BACKEND_PASSWORD")
    address  = "https://tf.kaipov.com/demo/terragrunt.bug.test"
  }
}
```

Upon a `terragrunt init`, a yes to the prompt to migrate state, and a `terragrunt plan`, we'll get that `unexpected end of JSON input` error because of the empty `terraform.tfstate` file.

---

I don't expect this change to break anything since the new behavior is in line with how Terraform behaves, i.e. completely igoring any local `terraform.tfstate` files if a `.terraform/terraform.tfstate` is present. Even if we were to replace the empty `terraform.tfstate` file with a valid state, Terraform would pay no attention it, so it too prefers remote states over local!

### DIFF
--- a/remote/terraform_state_file.go
+++ b/remote/terraform_state_file.go
@@ -53,10 +53,10 @@ func ParseTerraformStateFileFromLocation(backend string, config map[string]inter
 	stateFile, ok := config["path"].(string)
 	if backend == "local" && ok && util.FileExists(stateFile) {
 		return ParseTerraformStateFile(stateFile)
-	} else if util.FileExists(util.JoinPath(workingDir, DEFAULT_PATH_TO_LOCAL_STATE_FILE)) {
-		return ParseTerraformStateFile(util.JoinPath(workingDir, DEFAULT_PATH_TO_LOCAL_STATE_FILE))
 	} else if util.FileExists(util.JoinPath(dataDir, DEFAULT_PATH_TO_REMOTE_STATE_FILE)) {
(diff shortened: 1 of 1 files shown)


yorinasub17 (approved):
This makes sense to me! Thanks for the contribution! I'll kick off a sanity check build and if that passes, we can merge this in.

yorinasub17:
Build passed, so will merge this in! Thanks again for the contribution!

PR #1886, opened 2021-10-22
Demote log to debug level because it is more informative than warning

yorinasub17:
Originally I set this log a warn level, but then now I realize it should actually be `debug` level because this is more of an informational log for the end user. This information is only a warning when the user is first developing the `terragrunt.hcl`, but once it's setup and the user accepts the partial parsing, this log is completely use less. As such, it is more of a debug level log, where you only want it when something doesn't work.


brikis98 (approved):
LGTM!

yorinasub17:

PR #1893, opened 2021-10-27
--terragrunt-modules-that-include filter for run-all

yorinasub17:
## Motivation

With the introduction of multiple include blocks, it is becoming increasingly common to setup terragrunt projects that have shared configurations for specific components. However, this introduces a problem for infrastructure CI/CD pipelines, where you want to only run against the modules that changed.

Consider our example in [terragrunt-infrastructure-live-example](https://github.com/gruntwork-io/terragrunt-infrastructure-live-example). The common configurations related to `mysql` are extracted in the file `_envcommon/mysql.hcl`. If we were bumping the module version for the `qa` environment by editing the source reference in [non-prod/us-east-1/qa/mysql/terragrunt.hcl](https://github.com/gruntwork-io/terragrunt-infrastructure-live-example/blob/master/non-prod/us-east-1/qa/mysql/terragrunt.hcl#L10), the pipeline is straightforward - we just run `terragrunt plan` and `terragrunt apply` in the folder of the file that changed.

However, what should we do if the common configuration gets updated? E.g., bumping the instance type: https://github.com/gruntwork-io/terragrunt-infrastructure-live-example/blob/master/_envcommon/mysql.hcl#L40

It isn't immediately obvious what the desired behavior is in this case. For example, for something like changing the instance type of the database, the desired behavior is probably to run `terragrunt plan` and `terragrunt apply` progressively across the environments. That is, start with `qa`, then `stage`, and then `prod`. On the other hand, for a change like updating the environment variables for an application, you can get away with running `terragrunt plan` and `terragrunt apply` in parallel across all environments.

This PR seeks to provide a basic building block that can be used for designing a pipeline for both use cases.

## Solution

In this PR, we introduce a flag for `run-all` (`--terragrunt-modules-that-include`) which will filter out all modules in a stack that does not include at least one of the files listed in the arg.

With such a filter, you can implement the two flows:

For the promotion flow, you can have your pipeline run each environment in succession:
```
terragrunt run-all plan --terragrunt-modules-that-include ../_envcommon/mysql.hcl --terragrunt-working-dir non-prod/qa
# if that succeeds
terragrunt run-all plan --terragrunt-modules-that-include ../_envcommon/mysql.hcl --terragrunt-working-dir non-prod/stage
# ... and so on ...
```

For the parallel flow, you can call `run-all` from the root once:
```
terragrunt run-all plan --terragrunt-modules-that-include _envcommon/mysql.hcl
```

Refer to the docs update in this PR for more details on the use case and how this feature can help implement the pipeline.

### DIFF
--- a/docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
+++ b/docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
@@ -442,3 +442,23 @@ _env/app.hcl`. This will:
   `_env/app.hcl`).
 
 Thereby allowing you to only touch those modules that need to be updated by the code change.
+
+Alternatively, you can implement a promotion workflow if you have multiple environments that depend on the
+`_env/app.hcl` configuration. In the above example, suppose you wanted to progressively roll out the changes through the
+environments, `qa`, `stage`, and `prod` in order. In this case, you can use `--terragrunt-working-dir` to scope down the
+updates from the common file:
+
+```
+# Roll out the change to the qa environment first
+terragrunt run-all plan --terragrunt-modules-that-include _env/app.hcl --terragrunt-working-dir qa
+terragrunt run-all apply --terragrunt-modules-that-include _env/app.hcl --terragrunt-working-dir qa
+# If the apply succeeds to qa, move on to the stage environment
+terragrunt run-all plan --terragrunt-modules-that-include _env/app.hcl --terragrunt-working-dir stage
+terragrunt run-all apply --terragrunt-modules-that-include _env/app.hcl --terragrunt-working-dir stage
+# And finally, prod.
+terragrunt run-all plan --terragrunt-modules-that-include _env/app.hcl --terragrunt-working-dir prod
+terragrunt run-all apply --terragrunt-modules-that-include _env/app.hcl --terragrunt-working-dir prod
+```
+
+This allows you to have flexibility in how changes are rolled out. For example, you can add extra validation stages
+inbetween the roll out to each environment, or add in manual approval between the stages.
--- a/docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
+++ b/docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
@@ -462,3 +462,9 @@ terragrunt run-all apply --terragrunt-modules-that-include _env/app.hcl --terrag
 
 This allows you to have flexibility in how changes are rolled out. For example, you can add extra validation stages
 inbetween the roll out to each environment, or add in manual approval between the stages.
+
+**NOTE**: If you identify an issue with rolling out the change in a downstream environment, and want to abort, you will
+need to make sure that that environment uses the older version of the common configuration. This is because the common
+configuration is now partially rolled out, where some environments need to use the new updated common configuration,
+while other environments need the old one. The best way to handle this situation is to create a new copy of the common
+configuration at the old version and have the environments that depend on the older version point to that version.
--- a/docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
+++ b/docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
@@ -20,6 +20,8 @@ nav_title_link: /docs/
 
   - [Using read\_terragrunt\_config to DRY parent configurations](#using-read_terragrunt_config-to-dry-parent-configurations)
 
+  - [Considerations for CI/CD pipelines](#considerations-for-ci-cd-pipelines)
+
 
 ### Motivation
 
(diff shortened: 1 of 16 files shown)


docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
```diff
@@ -409,3 +411,34 @@ terraform {
...
+were updated, you can run `terragrunt plan` or `terragrunt apply` by passing in the updated file with
+`--terragrunt-config`.
+
+However, if you use `include` or `read_terragrunt_config`, then a single file change may need to be reflected on
+multiple files that were not touched at all in the commit. In our previous example, when a configuration is updated in
+the `_env/app.hcl` file, we need to apply the change to all the modules that `include` that common environment
+configuration.
+
+Terragrunt currently does not have any features for supporting this use case when `read_terragrunt_config` is
+used. However, for `include` blocks, you can use the
```
brikis98, line 432: Should we have something for `read_terragrunt_config`? Should `--terragrunt-modules-that-include` be something like `--terragrunt-modules-that-reference` instead, and support all possible ways to include or reference some other file?

docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
```diff
@@ -409,3 +411,34 @@ terraform {
...
+[--terragrunt-modules-that-include]({{site.baseurl}}/docs/reference/cli-options/#terragrunt-modules-that-include) CLI
+option for the `run-all` command.
+
+In the previous example, your CI/CD pipeline can run `terragrunt run-all plan --terragrunt-modules-that-include
+_env/app.hcl`. This will:
+
+- Recursively find all Terragrunt modules in the current directory tree.
+- Filter out any modules that don't include `_env/app.hcl` so that they won't be touched.
+- Run `plan` on any modules remaining (which will be the set of modules in the current tree that include
```
brikis98, line 441: How does `apply` work in this case? In particular:

1. If multiple modules in a single env include something that was modified: e.g., you have some `_env/db.hcl` and you use that 5 times in the `dev` environment to deploy 5 DBs. Will / should our CI / CD pipeline try to `apply` them all in parallel? I'm guessing the answer is yes.
2. If multiple modules across different envs include this thing: e.g., the `_env/db.hcl` is used in `dev`, `stage`, and `prod`. Will / should our CI / CD pipeline try to `apply` them all in parallel? I'm guessing the answer here should be no. But then, is our pipeline able to run `plan` in one env, prompt for approval, if approved, run `apply`, then automatically repeat the process with each subsequent env?

docs/_docs/04_reference/cli-options.md
```diff
@@ -735,6 +736,8 @@ When passed it, sets logging level for terragrunt. All supported levels are:
 
 **CLI Arg**: `--terragrunt-check`<br/>
 **Environment Variable**: `TERRAGRUNT_CHECK` (set to `true`)
+**Commands**:
+- [hclfmt](#hclfmt)
```
brikis98, line 740: Is this related to this PR?

docs/_docs/04_reference/cli-options.md
```diff
@@ -762,5 +767,65 @@ block by specifying `<BLOCK>.<ATTR>`, where `<BLOCK>` is the block name: e.g., `
...
+- [plan-all (DEPRECATED: use run-all)](#plan-all-deprecated-use-run-all)
+- [apply-all (DEPRECATED: use run-all)](#apply-all-deprecated-use-run-all)
+- [output-all (DEPRECATED: use run-all)](#output-all-deprecated-use-run-all)
+- [destroy-all (DEPRECATED: use run-all)](#destroy-all-deprecated-use-run-all)
+- [validate-all (DEPRECATED: use run-all)](#validate-all-deprecated-use-run-all)
+
+When passed in, `run-all` will only run the command against Terragrunt modules that include the specified file.
+
+This applies to the set of modules that are identified based on all the existing criteria for deciding which modules to
+include. For example, consider the following folder structure:
```
brikis98, line 791: NIT: You show the file structure here, but not the contents of the files, so it's not obvious what uses `include` on what if you're not already familiar with Terragrunt.

brikis98 (dismissed):
Overall, code changes LGTM! Just had some questions / NITs that can be answered separately from this PR.

docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
yorinasub17: Ideally yes, but it's non-trivial to implement this using the tools that the `hcl` library gives us, since function calls can't be parsed out. That is, there is no way to say "give me all the calls to `read_terragrunt_config`," like there is a way to say "give me all the `include` blocks."

docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
yorinasub17: I mentioned this in the PR description, but realized I didn't do it for the docs so expanded on it: 9c4dc14b3fae49d361e230796c3890420e353dec

Note that I think this is still flawed, in that the commit is a one way street. That is, if you abort after deploying to `stage`, you are now in an awkward state where you have to roll this change out when you want to touch `prod` since the updates are already merged for `qa` and `stage`.

I think the real solution here is going to be figuring out how to implement remote `include`, although that has its own can of worms as we discussed.

docs/_docs/04_reference/cli-options.md
yorinasub17: Yes, and it comes from the format change in https://github.com/gruntwork-io/terragrunt/pull/1893/files/fe655623b8730eeb428e6588fe1f2a48b557767e#diff-0ff83c0cbafb0d058b828ba1777066af4bc6eee4f726cecc0e0351b8f15ee65eR776

docs/_docs/04_reference/cli-options.md
yorinasub17: Makes sense! Added in example: 8eb213b4567d56ed6f2fdd8dfb807414a280a2ae

docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
brikis98: I'm going to toss out an idea that may be dumb, but just in case: since we implement the `read_terragrunt_config` helper, could it maintain an internal cache (map) of (a) which files were read in and (b) which files were doing the reading? If so, could that cache be used to do the `--terragrunt-modules-that-include` style filtering?

docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
brikis98: Ah, yea, you're right. This would require a rollback if the deploy failed part way through. I think the best we can do for now is document the need for rollback, both in the Terragrunt docs, and our own pipeline docs. In the future, as you said, remote `include` is prob the proper solution.

brikis98 (dismissed):
Two comments left, neither blocking, so giving it a "ship it" if you want to merge/release this and come back to the other items later.

docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
yorinasub17: I think that would work, but I am a bit hesitant with that approach primarily because it is essentially relying on a global variable that gets updated with functions that have side effects. It feels... messy...

I filed https://github.com/gruntwork-io/terragrunt/issues/1910 so we don't lose track of it.

docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
yorinasub17: Good point about documenting. What do you think about adcafb8f3b8656794301dec00c90355c94f3bdf6?

docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
brikis98: Roger, thx!

docs/_docs/02_features/keep-your-terragrunt-architecture-dry.md
brikis98: LGTM

brikis98 (approved):
LGTM

yorinasub17:
Thanks for review and working through this! I'm going to merge this in now.

Issue #1901, opened 2021-11-03
Autoinit no longer works


yorinasub17:
Gruntwork team is running into issues with autoinit. Specifically, the case where the module source ref being updated to a new ref is no longer being handled by autoinit.

## Repro:

- Create a `terragrunt.hcl` that references a module by ref.
- Run `terragrunt apply`
- Update `ref` to a new tag.
- Run `terragrunt apply`

### Expected:

Terragrunt runs `init` as part of auto init on the second apply

### Actual:

Terragrunt proceeds to call `terraform` without `init`, leading to a module source change error.

denis256:
I will take a look at this one

yorinasub17:
Fixed in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.35.7

PR #1903, opened 2021-11-04
#1901 Added logic to run init when new source is downloaded

denis256:
Updated flow to create `.terragrunt-init-required` in `WorkingDir` when a new source is downloaded, which will flag that `init` is required to be executed, if `init` is successfully executed  - `.terragrunt-init-required` is removed

```
# terragrunt.hcl
terraform {
  source = "git::git@github.com:denis256/terraform-test-module.git//modules/test-file?ref=v0.0.4"
  //source = "git::git@github.com:denis256/terraform-test-module.git//modules/test-file?ref=v0.0.1"
}
```

Before:
```
$ terragrunt apply --terragrunt-log-level debug --terragrunt-debug
...
DEBU[0001] Running command: terraform init
DEBU[0004] Running command: terraform apply 
...

# comment change commented source
$ terragrunt init --terragrunt-log-level debug --terragrunt-debug
...
DEBU[0001] Running command: terraform apply
...
```

With this change:
```
$ terragrunt init --terragrunt-log-level debug --terragrunt-debug
DEBU[0002] Running command: terraform init 
DEBU[0009] Running command: terraform apply
...
# change commented source
$ terragrunt init --terragrunt-log-level debug --terragrunt-debug
DEBU[0002] Running command: terraform init 
DEBU[0009] Running command: terraform apply
...
```

https://github.com/gruntwork-io/terragrunt/issues/1901

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"os"
 	"regexp"
 	"strings"
 	"time"
@@ -981,7 +982,13 @@ func runTerraformInit(originalTerragruntOptions *options.TerragruntOptions, terr
 		return err
 	}
 
-	return runTerragruntWithConfig(originalTerragruntOptions, initOptions, terragruntConfig, terraformSource != nil)
+	err = runTerragruntWithConfig(originalTerragruntOptions, initOptions, terragruntConfig, terraformSource != nil)
+
+	if err == nil {
+		initExecuted := util.JoinPath(terragruntOptions.WorkingDir, MODULE_NEED_INIT)
+		return os.Remove(initExecuted)
+	}
+	return err
 }
 
 func prepareInitOptions(terragruntOptions *options.TerragruntOptions, terraformSource *tfsource.TerraformSource) (*options.TerragruntOptions, error) {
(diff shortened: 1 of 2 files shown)


cli/download_source.go
```diff
@@ -18,6 +18,7 @@ import (
 
 // manifest for files copied from terragrunt module folder (i.e., the folder that contains the current terragrunt.hcl)
 const MODULE_MANIFEST_NAME = ".terragrunt-module-manifest"
+const MODULE_NEED_INIT = ".terragrunt-init-required"
```
yorinasub17, line 21: Style NIT: We've been switching to canonical golang consts of using `camelCase` instead of `ALL_CAPS`. Can you switch this to camelCase?

yorinasub17: Also, there should be a new line between this const and the `MODULE_MANIFEST_NAME` const. Otherwise, as is, the comment looks like it applies to both consts which isn't correct.

cli/cli_app.go
```diff
@@ -981,7 +982,15 @@ func runTerraformInit(originalTerragruntOptions *options.TerragruntOptions, terr
...
-	return runTerragruntWithConfig(originalTerragruntOptions, initOptions, terragruntConfig, terraformSource != nil)
+	err = runTerragruntWithConfig(originalTerragruntOptions, initOptions, terragruntConfig, terraformSource != nil)
+
+	if err == nil {
+		moduleNeedInit := util.JoinPath(terragruntOptions.WorkingDir, MODULE_NEED_INIT)
+		if util.FileExists(moduleNeedInit) {
+			return os.Remove(moduleNeedInit)
+		}
+	}
+	return err
```
yorinasub17, line 993: NIT: It's more canonical golang to flip the logic to check for err. I see why you did this, but it's actually harder to understand the `err = nil` condition, as the `return err` throws you off.

yorinasub17 (commented):
The overall idea LGTM! Mostly style nits.

That said, I think we need the following two updates prior to merge:

- The comments for these functions need to be updated to include info about the init file.
- The auto-init docs need to be updated to include info about the file (https://terragrunt.gruntwork.io/docs/features/auto-init/)

yorinasub17 (approved):
Updates LGTM! Kicked off build, and if it passes, we can merge this in.

yorinasub17:
Build passed, so will go ahead to merge this in!

Issue #1900, opened 2021-11-03
Overriding config_path causes errors with run-all


dudicoco:
Hi,

I have a common hcl file as follows:
```hcl
dependency "vpc_requester" {
  config_path = "empty"

  mock_outputs = {
    vpc_id                  = "(known after apply-all)"
  }
}

dependency "vpc_accepter" {
  config_path = "empty"

  mock_outputs = {
    vpc_id                  = "(known after apply-all)"
  }
}

inputs = {
  ...
}
```

I would like to override the config_path parameter within my child `terragrunt.hcl`:
```hcl
terraform {
  source = "git@github.com:xxxxxxxxxx//vpc-peering?ref=vpc-peering_v1.0.3"
}

prevent_destroy = true

dependency "vpc_requester" {
  config_path = "${get_parent_terragrunt_dir("root")}/path/to/file"
}

dependency "vpc_accepter" {
  config_path = "${get_parent_terragrunt_dir("root")}/path/to/other/file"

  mock_outputs = {
    vpc_region = "ap-southeast-2"
  }
}

include "root" {
  path           = find_in_parent_folders()
  merge_strategy = "deep"
}

include "common" {
  path           = "${dirname(find_in_parent_folders())}/_common/vpc-peering.hcl"
  merge_strategy = "deep"
}
```

When running `terragrunt plan` from within the child `terragrunt.hcl` dir it works as expected.
However, when running with `terragrunt run-all plan` it fails:
```sh
$ terragrunt run-all plan --terragrunt-strict-include --terragrunt-include-dir "dy-account/ap-southeast-2/_global/devops/vpc-peering/my-peering/"
ERRO[0013] Error processing module at '/my-account/ap-southeast-2/_global/devops/vpc-peering/my-peering/empty/terragrunt.hcl'. How this module was found: dependency of module at '/my-account/ap-southeast-2/_global/devops/vpc-peering/my-peering'. Underlying error: Error reading file at path /my-account/ap-southeast-2/_global/devops/vpc-peering/my-peering/empty/terragrunt.hcl: no such file or directory
ERRO[0013] Unable to determine underlying exit code, so Terragrunt will exit with error code 1
```

Terragrunt adds `empty` to the config path instead of completely overriding it.

yorinasub17:
This is indeed a bug - deep merge shouldn't concat the strings.

That said, I think you can workaround this by setting `config_path` to empty string (`""`). We have a test case that exercises this path:

[...]

PR #1906, opened 2021-11-07
#1900 Updated deep merge strategy to compare modules by path

denis256:
Updated deep merge strategy to compare by module names from "TerragruntDependencies" and use paths from modules which doesn't exist in "source"

Example:
https://github.com/denis256/terragrunt-tests/tree/master/deep-merge-fix

```
.
├── accepter
│   ├── main.tf
│   └── terragrunt.hcl
├── common.hcl
├── module
│   ├── main.tf
│   └── terragrunt.hcl
└── requester
    ├── main.tf
    └── terragrunt.hcl
```
Before:
```
$ cd module
$ terragrunt  run-all init
ERRO[0000] Error processing module at '/raid1/projects-work/g/test-repos/terragrunt-tests/deep-merge-fix/module/empty/terragrunt.hcl'. How this module was found: dependency of module at '/raid1/projects-work/g/test-repos/terragrunt-tests/deep-merge-fix/module'. Underlying error: Error reading file at path /raid1/projects-work/g/test-repos/terragrunt-tests/deep-merge-fix/module/empty/terragrunt.hcl: open /raid1/projects-work/g/test-repos/terragrunt-tests/deep-merge-fix/module/empty/terragrunt.hcl: no such file or directory 
```

After:
```
$ cd module
$ terragrunt  run-all init
Module /raid1/projects-work/g/test-repos/terragrunt-tests/deep-merge-fix/module depends on module /raid1/projects-work/g/test-repos/terragrunt-tests/deep-merge-fix/requester, which is an external dependency outside of the current working directory. Should Terragrunt run this external dependency? Warning, if you say 'yes', Terragrunt will make changes in /raid1/projects-work/g/test-repos/terragrunt-tests/deep-merge-fix/requester as well! (y/n) 

```

Fix for: https://github.com/gruntwork-io/terragrunt/issues/1900

### DIFF
--- a/config/include.go
+++ b/config/include.go
@@ -391,9 +391,10 @@ func (targetConfig *TerragruntConfig) DeepMerge(sourceConfig *TerragruntConfig, 
 	if sourceConfig.Dependencies != nil {
 		resultModuleDependencies := &ModuleDependencies{}
 		if targetConfig.Dependencies != nil {
+			// take in result dependencies only paths which aren't defined in source
+			// Fix for issue: https://github.com/gruntwork-io/terragrunt/issues/1900
 			targetPathMap := fetchDependencyPaths(targetConfig)
 			sourcePathMap := fetchDependencyPaths(sourceConfig)
-			// take in result dependencies only paths which aren't defined in source
 			for key, value := range targetPathMap {
 				_, found := sourcePathMap[key]
 				if !found {
--- a/config/include.go
+++ b/config/include.go
@@ -401,6 +401,20 @@ func (targetConfig *TerragruntConfig) DeepMerge(sourceConfig *TerragruntConfig, 
 					resultModuleDependencies.Paths = append(resultModuleDependencies.Paths, value)
 				}
 			}
+			// copy target paths which are defined only in Dependencies and not in TerragruntDependencies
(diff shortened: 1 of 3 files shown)


config/include.go
```diff
@@ -387,11 +387,20 @@ func (targetConfig *TerragruntConfig) DeepMerge(sourceConfig *TerragruntConfig,
...
+			// take in result dependencies only paths which aren't defined in source
+			// Fix for issue: https://github.com/gruntwork-io/terragrunt/issues/1900
+			targetPathMap := fetchDependencyPaths(targetConfig)
+			sourcePathMap := fetchDependencyPaths(sourceConfig)
+			for key, value := range targetPathMap {
+				_, found := sourcePathMap[key]
+				if !found {
+					resultModuleDependencies.Paths = append(resultModuleDependencies.Paths, value)
+				}
+			}
```
yorinasub17, line 403: Ah this logic works if there are no `dependencies` blocks, but since we still support that, this would break it.

You would need to somehow take into account paths that are in `targetConfig.Dependencies.Paths` which are not associated with any item in `TerragruntDependencies`.

yorinasub17 (commented):
I think the approach generally LGTM, but there is an edge condition that is not considered in the current implementation.

yorinasub17 (approved):
Build passed and the updated logic makes sense to me! Going to merge this in.

Issue #1895, opened 2021-10-29
Using TFR protocol sources with third-party registry providers reporting absolute urls is not working


spr-mweber3:
Hey there, we're using the Terraform Module Registry from Spacelift and want to use the added support for third-party module registries introduced with version [0.31.5](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.31.5). With version [0.31.9](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.31.9) Terragrunt supports Terraform registries, such as GitLab, that return relative paths. Not all third party providers do that.

Note how the .well-known/terraform.json reports absolute full URLs rather than relative paths - https://spacelift.io/.well-known/terraform.json.

This leads to the following error for us:

```
ERRO[0002] 1 error occurred:
	* error downloading 'tfr://spacelift.io/spring-media/maps-terraform-aws-elastic-ip?version=0.1.1': Error downloading module from https://spacelift.io/https:/app.spacelift.io/registry/modules/v1/spring-media/maps-terraform-aws-elastic-ip/0.1.1/download: error recieving HTTP data

ERRO[0002] Unable to determine underlying exit code, so Terragrunt will exit with error code 1
```

The url
> https://spacelift.io/https:/app.spacelift.io/registry/modules/v1/spring-media/maps-terraform-aws-elastic-ip/0.1.1/download

is constructed incorrectly.

The problem seems to originate from [here](https://github.com/gruntwork-io/terragrunt/pull/1767/files#diff-2775841193536f9771363fbc0d5e76f8bf3ba22b7b0ac049ece09a529370584dR131-R141).

So the fix in https://github.com/gruntwork-io/terragrunt/pull/1787 might have fixed one problem but introduced a new one.

ademspr:
I Have the same issue, using gitlab terraform registry.

denis256:
Hi @spr-mweber3, which URL is expected to be generated?
url: `https://app.spacelift.io/registry/modules/v1/spring-media/maps-terraform-aws-elastic-ip/0.1.1/download`  return 404 and no `x-terraform-get` header

spr-mweber3:
Hey @denis256, since the module is private the 404 is expected. The url to be constructed is `https://app.spacelift.io/registry/modules/v1/spring-media/maps-terraform-aws-elastic-ip/0.1.1/download` as you assumed correctly.

yorinasub17:
[...]

PR #1907, opened 2021-11-07
#1895 Updated tfr handling to support full paths

denis256:
Updated handling of TFR paths to concatenate strings and avoid using `path.Join` which for urls generate broken paths like:
```
path.Join("http://foo", "bar") => http:/foo/bar
```

https://github.com/gruntwork-io/terragrunt/issues/1895

### DIFF
--- a/internal/tfr/getter.go
+++ b/internal/tfr/getter.go
@@ -115,7 +115,8 @@ func (tfrGetter *TerraformRegistryGetter) Get(dstPath string, srcURL *url.URL) e
 		return err
 	}
 
-	moduleFullPath := path.Join(moduleRegistryBasePath, modulePath, version, "download")
+	moduleFullPath := fmt.Sprintf("%s/%s/%s/download", moduleRegistryBasePath, modulePath, version)
+		//path.Join(moduleRegistryBasePath, modulePath, version, "download")
 	moduleURL := url.URL{
 		Scheme: "https",
 		Host:   registryDomain,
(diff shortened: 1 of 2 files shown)


internal/tfr/getter.go
```diff
@@ -283,3 +281,25 @@ func httpGETAndGetResponse(ctx context.Context, getURL url.URL) ([]byte, *http.H
...
+
+// buildRequestUrl - create url to download module using moduleRegistryBasePath
+func buildRequestUrl(registryDomain string, moduleRegistryBasePath string, modulePath string, version string) (*url.URL, error) {
+	moduleRegistryBasePath = strings.TrimSuffix(moduleRegistryBasePath, "/")
+	modulePath = strings.TrimSuffix(modulePath, "/")
+	modulePath = strings.TrimPrefix(modulePath, "/")
+
+	moduleFullPath := fmt.Sprintf("%s/%s/%s/download", moduleRegistryBasePath, modulePath, version)
+
+	if strings.HasPrefix(moduleFullPath, "https") {
```
yorinasub17, line 293: Since `url.Parse` will successfully parse a path, what do you think of the following alternative?:

```
moduleURL, err := url.Parse(moduleFullPath)
if err != nil {
    return nil, err
}
if moduleURL.Scheme != "" {
    return moduleURL, nil
}
return &url.URL{Scheme: "https", Host: registryDomain, Path: moduleFullPath}, nil
```

I'm not sure if the module registry would return anything that is not `https`, so might be moot, but it would at least handle cases where the registry basepath has a URL that starts with https (e.g., `httpsmodules` or something like that).

yorinasub17 (commented):
Had one minor nit on the check logic, but otherwise LGTM!

yorinasub17 (approved):
Updates LGTM! Kicked off a build, and if it passes, we can merge this in!

yorinasub17:
Build passed, so will go ahead to merge this in!

Issue #1912, opened 2021-11-11
Raise minimum go version as code cannot compile with latest go 1.13


maunzCache:
I just set up the terragrunt repository locally as i have to investigate a regression in the parsing stage with the latest versions. I don't have that much of a knowledge regarding go yet so sorry if i overlooked something here.

Assuming that the file go.mod declares the projects dependencies, this has to be fixed. As defined in line 3 "go 1.13" can no longer be used to compile the project as minimum version.

The reason is that the go-getter in version 1.5.7 uses the url.Redacted() function from the Core which was introduced in 1.15 of go. There is however a backwards compability version of Redacted() for at least go 1.14

So here is the std output

```
$ go version
go version go1.13.15 linux/amd64

$ make build
set -xe ;\
vtag_maybe_extra=$(git describe --tags --abbrev=12 --dirty --broken) ;\
go build -o terragrunt -ldflags "-X main.VERSION=${vtag_maybe_extra}" .
+ git describe --tags --abbrev=12 --dirty --broken
+ vtag_maybe_extra=v0.35.8-1-gaf07ea65b640
+ go build -o terragrunt -ldflags -X main.VERSION=v0.35.8-1-gaf07ea65b640 .
# github.com/hashicorp/go-getter
../../../../go/pkg/mod/github.com/hashicorp/go-getter@v1.5.7/client.go:301:52: u.Redacted undefined (type *"net/url".URL has no field or method Redacted)
make: *** [Makefile:34: terragrunt] Error 2
```

My proposal is to raise the minimum version. I cannot recommend one version but i think 1.15 is a good place to start.

Also the contribution manual is sparse when explaining how to properly set up the local environment. I'd hope to find a recommended compiling version for terragrunt there.

denis256:
Hi, I think should be used 1.16, same as used in CI/CD:
https://github.com/gruntwork-io/terragrunt/blob/master/.circleci/config.yml#L7

yorinasub17:
This is now merged into master.

PR #1913, opened 2021-11-11
#1912 Go 1.16 in go.mod

denis256:
Updated `go.mod` to declare version go 1.16, same go version used in CI/CD

#1912


yorinasub17 (approved):
Ah good catch! LGTM!

Issue #1891, opened 2021-10-26
Destroy executes top-most terragrunt.hcl


riy:
Terraform Version: `1.0.9`
Terragrunt Version: `0.34.4`

I have a problem where running...
`terragrunt destroy --terragrunt-working-dir ./environments/hub/eu-central-1/network`
... leads to this error message:
```shell
$ terragrunt destroy --terragrunt-working-dir ./environments/
Error: Error in function call

  on environments/terragrunt.hcl line 9, in locals:
   9:   aws_accounts_vars = read_terragrunt_config(find_in_parent_folders("region.hcl"))

Call to function "find_in_parent_folders" failed: ParentFileNotFound: Could not find a region.hcl in any of the parent folders of environments/terragrunt.hcl. Cause: Traversed all the way to the root..

ERRO[0000] Encountered error while evaluating locals.
ERRO[0000] environments/terragrunt.hcl:9,46-69: Error in function call; Call to function "find_in_parent_folders" failed: ParentFileNotFound: Could not find a aws_accounts.hcl in any of the parent folders of environments/terragrunt.hcl. Cause: Traversed all the way to the root..
ERRO[0000] Unable to determine underlying exit code, so Terragrunt will exit with error code 1
```

This is the same error message that is shown if I make this call:
`terragrunt destroy --terragrunt-working-dir ./environments`

So to me the call `terragrunt destroy --terragrunt-working-dir ./environments/hub/eu-central-1/network` looks like it will work on `./environments/hub/eu-central-1/network/terragrunt.hcl` (this works fine and destroys what I want Terraform to destroy) and then (this is my assumption, as the error message looks like it) it will execute `./environments/terragrunt.hcl`.
Obviously the `./environments/terragrunt.hcl` doesn't have any value by itself, it only makes sense if it's included by a `terragrunt.hcl` folder somewhere deep down. But for some reason it seems to get executed by itself as well.

The directory structure is like this:
```
└─ environments/
  └─ terragrunt.hcl
  └─ hub/
    └─ eu-central-1/
      └─ region.hcl
        └─ network/
          └─ terragrunt.hcl
```

My `./environments/hub/eu-central-1/network/terragrunt.hcl` looks like this:
```terraform
...
include "root" {
  path   = find_in_parent_folders()
  expose = true
}
...
```

My `./environments/terragrunt.hcl` looks like this:
```terraform
locals {
  ...
  # Automatically load region-level variables
[...]

PR #1915, opened 2021-11-12
#1891 Dependent modules detection update

denis256:
Updated FindWhereWorkingDirIsIncluded to return only detected dependent modules, in case of failures user will be asked to confirm the action in any case.

Fix for: https://github.com/gruntwork-io/terragrunt/issues/1891

### DIFF
--- a/configstack/module.go
+++ b/configstack/module.go
@@ -557,7 +557,7 @@ func getSortedKeys(modules map[string]*TerraformModule) []string {
 // 1. Find root git top level directory and build list of modules
 // 2. Iterate over includes from terragruntOptions if git top level directory detection failed
 // 3. Filter found module only items which has in dependencies working directory
-func FindWhereWorkingDirIsIncluded(terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig) ([]*TerraformModule, error) {
+func FindWhereWorkingDirIsIncluded(terragruntOptions *options.TerragruntOptions, terragruntConfig *config.TerragruntConfig) []*TerraformModule {
 	var pathsToCheck []string
 	var matchedModulesMap = make(map[string]*TerraformModule)
 	var gitTopLevelDir = ""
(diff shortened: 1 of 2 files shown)


configstack/module.go
```diff
@@ -577,7 +577,8 @@ func FindWhereWorkingDirIsIncluded(terragruntOptions *options.TerragruntOptions,
 		dir = dir + filepath.FromSlash("/")
 		cfgOptions, err := options.NewTerragruntOptions(dir)
 		if err != nil {
-			return nil, err
+			terragruntOptions.Logger.Debugf("Failed to build terragrunt options from %s %v", dir, err)
```
brikis98, line 580: This seems like a real error that should probably be returned?

configstack/module.go
```diff
@@ -586,7 +587,10 @@ func FindWhereWorkingDirIsIncluded(terragruntOptions *options.TerragruntOptions,
 		}
 		stack, err := FindStackInSubfolders(cfgOptions)
 		if err != nil {
-			return nil, err
+			// loggign error as debug since in some cases stack building may fail because parent files can be designed
+			// to work with relative paths from downstream modules
+			terragruntOptions.Logger.Debugf("Failed to build module stack %v", err)
```
brikis98, line 592: Is this the real fix for this issue? Where we _don't_ return an error in this case?

configstack/module.go
denis256: Since building list of all modules that use current module may not work in some cases(usage of relative paths) approach is to log as debug and ask user to confirm only detected dependencies

https://github.com/gruntwork-io/terragrunt/pull/1889

https://github.com/gruntwork-io/terragrunt/issues/299

brikis98 (approved):
Understood! I'll kick off tests now.

brikis98:
Tests passed! Merging now.

brikis98:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.35.12

PR #1917, opened 2021-11-17
Clarify dependency settings.

rhoboat:
I'd like some help with wording here...

### DIFF
--- a/docs/_docs/04_reference/config-blocks-and-attributes.md
+++ b/docs/_docs/04_reference/config-blocks-and-attributes.md
@@ -838,13 +838,16 @@ The `dependency` block supports the following arguments:
   as a dependency in this configuration.
 - `skip_outputs` (attribute): When `true`, skip calling `terragrunt output` when processing this dependency. If
   `mock_outputs` is configured, set `outputs` to the value of `mock_outputs`. Otherwise, `outputs` will be set to an
-  empty map.
+  empty map. Put another way, setting `skip_outputs` means "use mocks all the time if `mock_outputs` are set."
 - `mock_outputs` (attribute): A map of arbitrary key value pairs to use as the `outputs` attribute when no outputs are
-  available from the target module, or if `skip_outputs` is `true`.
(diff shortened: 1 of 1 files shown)


docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -838,13 +838,16 @@ The `dependency` block supports the following arguments:
   as a dependency in this configuration.
 - `skip_outputs` (attribute): When `true`, skip calling `terragrunt output` when processing this dependency. If
   `mock_outputs` is configured, set `outputs` to the value of `mock_outputs`. Otherwise, `outputs` will be set to an
-  empty map.
+  empty map. Put another way, setting `skip_outputs` means "use mocks all the time if `mock_outputs` are set."
 - `mock_outputs` (attribute): A map of arbitrary key value pairs to use as the `outputs` attribute when no outputs are
-  available from the target module, or if `skip_outputs` is `true`.
+  available from the target module, or if `skip_outputs` is `true`. However, it's generally recommended not to set
+  `skip_outputs` if using `mock_outputs`, because `skip_outputs` means "use mocks all the time if they are set" whereas
+  `mock_outputs` means "use mocks only if real outputs are not available." Setting both can lead to unexpected behavior.
```
yorinasub17, line 845: I'd replace the last sentence with:
```suggestion
  `mock_outputs` means "use mocks only if real outputs are not available." Use `locals` instead when `skip_outputs = true`.
```

The behavior of using both is defined (always return mocks), but it can lead to verbosity in the configuration since you can replace it with locals.

yorinasub17 (commented):
Mostly LGTM! Had one minor correction.

docs/_docs/04_reference/config-blocks-and-attributes.md
rhoboat: Ah, I see. Thanks!

yorinasub17 (approved):
Updates LGTM!

rhoboat:
Thanks for the review! Merging nao. 🐱

Issue #1919, opened 2021-11-18
Add TerraformPath in dependency block to get output from different tfstate version


leonardobiffi:
When using the dependency on a resource that was created with a different terraform version, the error occurs when executing the `terraform output` command

```
Failed to load state: state snapshot was created by Terraform v1.0.6, which is newer than current v0.13.3; upgrade to Terraform v1.0.6 or greater to work with this state
```

The idea is to add an attribute to configure the terraform binary to be used for that dependency, like the example below

```
dependency "key_pair" {
  config_path       = "../test-keypair"
  terraform_path    = "/usr/local/bin/terraform/1.0.6/terraform"
  
  mock_outputs = {
    key_pair_name = "KEY"
  }
  mock_outputs_allowed_terraform_commands = ["validate", "plan"]
}
```

So when executing the `terraform output` command the terragrunt will use the binary configured in the dependency

PR #1920, opened 2021-11-18
add terraform_path in dependency

leonardobiffi:
Suggestion for Issue #1919

### DIFF
--- a/config/dependency.go
+++ b/config/dependency.go
@@ -36,6 +36,7 @@ type Dependency struct {
 	MockOutputs                         *cty.Value `hcl:"mock_outputs,attr" cty:"mock_outputs"`
 	MockOutputsAllowedTerraformCommands *[]string  `hcl:"mock_outputs_allowed_terraform_commands,attr" cty:"mock_outputs_allowed_terraform_commands"`
 	MockOutputsMergeWithState           *bool      `hcl:"mock_outputs_merge_with_state,attr" cty:"mock_outputs_merge_with_state"`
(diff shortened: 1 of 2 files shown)


yorinasub17:
Thanks for the contribution!

However, I'm weary of this setting living on the `dependency` block, because that means you have to set it on every child module that depends on it. The desired solution is for `terragrunt` to use the right `terraform` binary when [terraform_binary](https://terragrunt.gruntwork.io/docs/reference/config-blocks-and-attributes/#terraform_binary) is set on the dependency config (that is, `terraform_binary` is encoded in the `terragrunt.hcl` for `"../test-keypair"`). Does it work if you do that?

leonardobiffi:
Hi @yorinasub17 

I tried to add `terraform_binary` in`terragrunt.hcl` of dependency `"../test-keypair"`, but got the same error.
```
[terragrunt] [/builds/ZvsktM4n/0/sandbox/us-east-1/iac/dev/ec2-test-keypair] 2021/11/25 12:00:55 Running command: /usr/local/bin/terraform/0.13.3/terraform output -json
Failed to load state: state snapshot was created by Terraform v1.0.6, which is newer than current v0.13.3; upgrade to Terraform v1.0.6 or greater to work with this state
```

it seems that in the execution of the `terraform output` command the `terragruntOptions` of the current `terragrunt.hcl` is used and not the one of the dependency `"../test-keypair"`

yorinasub17:
Ah then that is actually a bug, as the intended behavior is to honor the `terraform_binary` of the dependency `terragrunt.hcl`. I'll investigate to see if I can get that fixed.

PR #1922, opened 2021-11-18
Support terragrunt args encoded as --OPTION_KEY=VALUE

yorinasub17:
Our internal pipeline guards only handle option args encoded as `--OPTION_KEY=VALUE`, which makes it impossible to pass in terragrunt args because it currently only supports encoding as `--OPTION_KEY VALUE`.

This PR updates terragrunt arg parsing to support the `--OPTION_KEY=VALUE` encoding.

### DIFF
--- a/cli/args.go
+++ b/cli/args.go
@@ -299,15 +299,14 @@ func filterTerragruntArgs(args []string) []string {
 	out := []string{}
 	for i := 0; i < len(args); i++ {
 		arg := args[i]
-		argWithoutPrefix := strings.TrimPrefix(arg, "--")
-
 		if util.ListContainsElement(MULTI_MODULE_COMMANDS, arg) {
 			// Skip multi-module commands entirely
 			continue
 		}
 
+		argWithoutPrefix := strings.TrimPrefix(arg, "--")
 		if util.ListContainsElement(allTerragruntStringOpts, argWithoutPrefix) {
-			// String flags have the argument and the value, so skip both
+			// String flags that directly match the arg have the argument and the value, so skip both
 			i = i + 1
 			continue
 		}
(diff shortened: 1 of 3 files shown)


zackproser:
Curious / sanity check: does this have any implications for the terragrunt validate / --terragrunt-strict behavior? I'm thinking not because it looks like your changes handle both styles of arg presentation / normalize them as needed.

yorinasub17:
> Curious / sanity check: does this have any implications for the terragrunt validate / --terragrunt-strict behavior? I'm thinking not because it looks like your changes handle both styles of arg presentation / normalize them as needed.

Nope no implications, as boolean args are handled in the same way as before. This only affects the string based args.

yorinasub17:
Thanks for review! I am reasonably confident this is backward compatible given the regression tests, so I'll go ahead and merge this in.

cli/args.go
```diff
@@ -360,15 +366,31 @@ func parseIAMRoleOptions(args []string) (*options.IAMRoleOptions, error) {
...
+func getOptionArgIfMatched(arg string, nextArg *string, optionName string) (string, bool, error) {
+	optionOnly := fmt.Sprintf("--%s", optionName)
+	if arg == optionOnly {
+		logIfDeprecatedOption(optionName)
+		if nextArg == nil {
+			return "", true, errors.WithStackTrace(ArgMissingValue(optionName))
+		}
+		return *nextArg, true, nil
+	} else if strings.HasPrefix(arg, optionOnly+"=") {
+		logIfDeprecatedOption(optionName)
```
rhoboat, line 382: NIT: consider moving the `logIfDeprecatedOption(optionName)` to above the conditional.

cli/args.go
```diff
@@ -360,15 +366,31 @@ func parseIAMRoleOptions(args []string) (*options.IAMRoleOptions, error) {
...
+		return strings.TrimPrefix(arg, fmt.Sprintf("--%s=", optionName)), true, nil
```
rhoboat, line 383: This will return `"", true, nil` when the `arg` is something like `--OPTION_KEY=` with no value after it, meaning no `nextArg`.

rhoboat: So it's an error to say `--OPTION_KEY` (with no value after) but it's not an error to say `--OPTION_KEY=` (with no value after).

cli/args.go
```diff
@@ -423,16 +442,15 @@ func parseMultiStringArg(args []string, argName string, defaultValue []string) (
...
-			} else {
-				return nil, errors.WithStackTrace(ArgMissingValue(argName))
-			}
+		var nextArg *string
+		if (i + 1) < len(args) {
+			nextArg = &args[i+1]
+		}
+		val, hasVal, err := getOptionArgIfMatched(arg, nextArg, argName)
+		if err != nil {
+			return nil, errors.WithStackTrace(ArgMissingValue(argName))
```
rhoboat, line 451: Why is this line not like `parseIntArg` line 428? `return nil, err`? I think that's the only error that's returned from `getOptionArgIfMatched`, so I would consider using `err` here.

cli/args.go
yorinasub17: Yes that is correct and intentional. Most CLI will interpret `--OPTION_KEY=` as setting the value to `""`, and that is the approach taken here. This is because of bashism, where `--OPTION_KEY=""` is not the same as `--OPTION_KEY='""'` due to automatic string concatenation (it reduces to `--OPTION_KEY=`).

cli/args.go
yorinasub17: Good catch - addressed in https://github.com/gruntwork-io/terragrunt/pull/1923

cli/args.go
rhoboat: Ah, interesting, thank you for that!

rhoboat: Right, I did some reading up on bash, and `""` will get chomped to nothing and `'""'` will become `""`.

rhoboat: So if someone actually means to pass in the non-empty string `""`, they should provide it quoted as in `'"'"`?

yorinasub17: > So if someone actually means to pass in the non-empty string "", they should provide it quoted as in '""'?

Yup that's right.

Issue #1921, opened 2021-11-18
Terragrunt auto-init always executes if no module is present


Ic3w0lf:
Hi,

in our project setup we often have the case that our `terragrunt modules` don't require any other modules. It seems that this breaks the `auto-init` behavior to **always** initialize the modules. 

I did some debugging and compared the outputs of a module which relies on an external module (e.g. [vpc](https://registry.terraform.io/modules/terraform-aws-modules/vpc/aws/latest)) with a plain module that has no `module` at all. If I see correctly this behavior is coming from the `modulesNeedInit()` function:

```
func modulesNeedInit(terragruntOptions *options.TerragruntOptions) (bool, error) {
        modulesPath := util.JoinPath(terragruntOptions.DataDir(), "modules")
	if util.FileExists(modulesPath) {                                          <----- Always false as no will be downloaded
		return false, nil
	}
	moduleNeedInit := util.JoinPath(terragruntOptions.WorkingDir, moduleInitRequiredFile)
	if util.FileExists(moduleNeedInit) {                                       <----- Will evaluate to true
		return true, nil
	}
        ....
}
```

I can fix this by just creating an empty `module` in all of my modules, but this seems kind of hacky and maybe not the best approach:
```
module "null" {
  source = "../null"  # empty folder
}
```

I first wanted to ask if this is desired behavior or a bug? It is slowing down the development for me a bit as for most of my modules `init` is run all the time (but not too much as its all cached anyway) :)

Terragrunt: v0.35.10
Terraform: 1.0.11

Let me know if I forgot any information!

Ic3w0lf:
I figured out a better workaround for this inspired by #1423 :)

```
after_hook "reinit_workaround" {
    commands = ["plan", "apply"]
    execute = ["/bin/sh", "-c", "mkdir -p .terraform/modules"]
}
```

Works like a charm for my case!

denis256:
hi @Ic3w0lf , can you share an example code where this issue appears

I tried to do with a small subproject:
https://github.com/denis256/terragrunt-tests/tree/master/terragrunt-init-each-time

Initialization happens, but on older versions(v0.32.0) too

Ic3w0lf:
Hi @denis256,

thank you for looking into the issue! I setup a small repo that reflects our development setup and was able to reproduce this issue: https://github.com/Ic3w0lf/terragrunt-test/tree/main/auto-init-issue

[...]

PR #1924, opened 2021-11-19
#1921 Version check before creating auto-init flag file

denis256:
Added check for source code version before creating file for auto-init, if terraform source is changed - create auto init file

Example runs:
Local dependency(https://github.com/Ic3w0lf/terragrunt-test/tree/main/auto-init-issue/accounts/prod/vpc):
```
$ terragrunt plan
Initializing the backend...

- Reusing previous version of hashicorp/local from the dependency lock file
- Installing hashicorp/local v2.1.0...
- Installed hashicorp/local v2.1.0 (signed by HashiCorp)

...

$ terragrunt plan
Terraform used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:
  + create

```

Changing terraform source path(https://github.com/denis256/terragrunt-tests/tree/master/terragrunt-init)
```
$ terragrunt plan
Initializing the backend...

- Reusing previous version of hashicorp/local from the dependency lock file
- Using previously-installed hashicorp/local v2.1.0

...
$ terragrunt plan
Terraform used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:
  + create
```
Change terraform source:
```
$ terragrunt plan
Initializing the backend...

- Reusing previous version of hashicorp/local from the dependency lock file
- Using previously-installed hashicorp/local v2.1.0

$ terragrunt plan
Terraform used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:
  + create

```

Fixes: https://github.com/gruntwork-io/terragrunt/issues/1921

### DIFF
--- a/cli/download_source.go
+++ b/cli/download_source.go
@@ -74,6 +74,16 @@ func downloadTerraformSourceIfNecessary(terraformSource *tfsource.TerraformSourc
 		return nil
 	}
 
+	var previousVersion = ""
+	// read previous source version
+	// https://github.com/gruntwork-io/terragrunt/issues/1921
+	if util.FileExists(terraformSource.VersionFile) {
+		previousVersion, err = readVersionFile(terraformSource)
+		if err != nil {
+			return err
+		}
+	}
+
 	// When downloading source, we need to process any hooks waiting on `init-from-module`. Therefore, we clone the
 	// options struct, set the command to the value the hooks are expecting, and run the download action surrounded by
 	// before and after hooks (if any).
@@ -95,12 +105,17 @@ func downloadTerraformSourceIfNecessary(terraformSource *tfsource.TerraformSourc
 		return err
 	}
 
-	initFile := util.JoinPath(terraformSource.WorkingDir, moduleInitRequiredFile)
-	f, createErr := os.Create(initFile)
-	if createErr != nil {
-		return createErr
+	currentVersion := terraformSource.EncodeSourceVersion()
+	// if source versions are different, create file to run init
+	// https://github.com/gruntwork-io/terragrunt/issues/1921
+	if previousVersion != currentVersion {
+		initFile := util.JoinPath(terraformSource.WorkingDir, moduleInitRequiredFile)
+		f, createErr := os.Create(initFile)
+		if createErr != nil {
+			return createErr
+		}
+		defer f.Close()
 	}
-	defer f.Close()
 
 	return nil
 }
--- a/cli/download_source_test.go
+++ b/cli/download_source_test.go
@@ -103,7 +103,7 @@ func TestDownloadTerraformSourceIfNecessaryLocalDirToAlreadyDownloadedDir(t *tes
 
 	copyFolder(t, "../test/fixture-download-source/hello-world-2", downloadDir)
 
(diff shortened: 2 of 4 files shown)


cli/download_source.go
```diff
@@ -95,12 +105,17 @@ func downloadTerraformSourceIfNecessary(terraformSource *tfsource.TerraformSourc
...
 	}
 
-	initFile := util.JoinPath(terraformSource.WorkingDir, moduleInitRequiredFile)
-	f, createErr := os.Create(initFile)
-	if createErr != nil {
-		return createErr
+	currentVersion := terraformSource.EncodeSourceVersion()
+	// if source versions are different, create file to run init
+	// https://github.com/gruntwork-io/terragrunt/issues/1921
+	if previousVersion != currentVersion {
```
brikis98, line 111: I'm not sure I follow the logic... Why is checking for `source`  equality the solution here, if the original problem was that if you didn't have `module` in your code, our `.modules` check was incorrect?

brikis98 (commented):
Thanks for the PR!

Could we add a regression test around this?

Also, NIT:

> Fix for: #1921

If you write this as "Fixes #1921" or "Closes #1921", GitHub auto-links the issue, and when this PR is merged, automatically marks the issue as closed.

cli/download_source.go
denis256: Creation of `module` directory is just a workaround to skip initialization, original issue occurs because "local" dependencies are copied each time to `.terragrunt-cache`, and to know when to do initialization or not - I added logic to verify checksum of terraform version.

rhoboat:
Sorry for the delay getting this reviewed. Plan is to review this week (today hopefuly).

rhoboat:
@denis256 Can you please update this PR to resolve conflicts?

denis256:
@rhoboat yup, synchronized with master branch

cli/download_source.go
rhoboat: The way I understand this logic now is:

Cases where it will not auto-init:

If there's no `modules` dir
and the terraformSources are the same
and there are no `module` blocks in the *.tf files

If there's a `modules` dir

Cases where it will auto-init:

If there's no `modules` dir
and the terraformSources are different

If there's no `modules` dir
and the terraformSources are different
but there are `module` blocks in the *.tf files

This change makes sense to me. In fact, I'm not fully understanding the current code as it is. Why does it always create the initFile?

denis256: This PR has dependencies with another issue: https://github.com/gruntwork-io/terragrunt/issues/1901
`initFile` was introduced to automatically (re)run terragrunt init, with this PR it will compare source code versions before creation of the `initFile` - otherwise, for local dependencies, it will do init each time

rhoboat (approved):
I'm inclined to go forward with this change. I think it's doing the right thing. But I'd love a review from either Jim or Yori.

cli/download_source.go
rhoboat: Ah I see.

rhoboat:
This again LGTM! With the added context of #1901 I understand a bit better why it was failing before. Go ahead and merge and release!

Issue #394, opened 2017-12-11
terragrunt not copying doted files (.something) from working dir in temp dir


ebarault:
this happens during stage labelled as
```
Copying files from my/workdir into /Users/ebarault/.terragrunt/qpryAuEVuRxDI-1s0Hw-004ArQc/WTycM2xCvnoeoxRFsIteSUwW1Zk
```

In `my/workdir` i have dotet files such `.config`, `.env` that are essential to my terraform task.

but when investigating after errors I noted that these files are not copied by terragrunt in the tmp dir `Users/ebarault/.terragrunt/qpryAuEVuRxDI-1s0Hw-004ArQc/WTycM2xCvnoeoxRFsIteSUwW1Zk`

is there a rational for this, or is this a bug?

i'm running terragrunt v0.13.23

brikis98:
Terragrunt intentionally skips hidden files and folders: https://github.com/gruntwork-io/terragrunt/pull/236. Ideas on how to include them, while still skipping "problematic" ones (e.g., `.terraform`, `.git`, etc) are welcome!

ebarault:
ah ah, the wonderful side effect ! This is deceptive.
this means one cannot trust the copy step as a carbon copy, so I should always pass the absolute path of any included src/assets/whatever ressources to my module.

on how to improve this : why not maintaining a black list of unwanted files instead of running a whitelist excluding all hidden files/folders ?

brikis98:
Unintended consequences :)

Right now, Terragrunt doesn't copy any hidden files or folders. I'd be fine with either a blacklist or whitelist approach, so long as you could configure it. Example:

```hcl
terragrunt = {
  terraform {
    source = "..."
    exclude_from_copy = [".terraform", ".git"]
  }
}
```

rgarrigue:
Maybe some kind of .ebaraultignore ? :trollface:  ;-)

ebarault:
please forgive this funny colleague of mine ;-)

ebarault:
I'll try to spare some time in the coming days to work on that one. Please could you refer where specifically in the code the hidden files/folders are ignored?

brikis98:
Method that copies files/folders: https://github.com/gruntwork-io/terragrunt/blob/master/util/file.go#L136

Terragrunt config as it's parsed from file: https://github.com/gruntwork-io/terragrunt/blob/master/config/config.go#L32

[...]

PR #1935, opened 2021-12-02
Implement feature to allow hidden files in local copy

yorinasub17:
Fixes https://github.com/gruntwork-io/terragrunt/issues/394

### DIFF
--- a/docs/_docs/04_reference/config-blocks-and-attributes.md
+++ b/docs/_docs/04_reference/config-blocks-and-attributes.md
@@ -63,6 +63,11 @@ The `terraform` block supports the following arguments:
       registry]({{site.baseurl}}/docs/getting-started/quick-start#a-note-about-using-modules-from-the-registry) for more
       information about using modules from the Terraform Registry with Terragrunt.
 
+- `include_in_copy` (attribute): When using local file path sources, this specifies a list of glob patterns (e.g.,
+  `*.txt`) that should always be copied. The path should be specified relative to the source directory. This is useful
+  to specify a list of hidden files that should be included in the source copy that Terragrunt makes (by default,
+  Terragrunt ignores all hidden files and folders in the Terraform source path).
+
 - `extra_arguments` (block): Nested blocks used to specify extra CLI arguments to pass to the `terraform` CLI. Learn more
   about its usage in the [Keep your CLI flags DRY]({{site.baseurl}}/docs/features/keep-your-cli-flags-dry/) use case overview. Supports
   the following arguments:
@@ -117,8 +122,7 @@ supported:
       while the working directory for hooks associated with `init` will be the terraform module.
 
 
-
-Example:
+Complete Example:
 
 ```hcl
 terraform {
@@ -194,6 +198,23 @@ terraform {
   }
 }
 ```
+
+Local File Path Example with allowed hidden files:
+
+```hcl
+terraform {
+  # Pull the terraform configuration from the local file system. Terragrunt will make a copy of the source folder in the
+  # Terragrunt working directory (typically `.terragrunt-cache`).
+  source = "../modules/networking/vpc"
+
+  # Always include the following file patterns in the Terragrunt copy.
+  include_in_copy = [
+    ".security_group_rules.json",
+    "*.yaml",
+  ]
+}
+```
+
 
 #### A note about using modules from the registry
 
--- a/docs/_docs/04_reference/config-blocks-and-attributes.md
+++ b/docs/_docs/04_reference/config-blocks-and-attributes.md
@@ -63,10 +63,17 @@ The `terraform` block supports the following arguments:
       registry]({{site.baseurl}}/docs/getting-started/quick-start#a-note-about-using-modules-from-the-registry) for more
       information about using modules from the Terraform Registry with Terragrunt.
 
-- `include_in_copy` (attribute): When using local file path sources, this specifies a list of glob patterns (e.g.,
-  `*.txt`) that should always be copied. The path should be specified relative to the source directory. This is useful
-  to specify a list of hidden files that should be included in the source copy that Terragrunt makes (by default,
-  Terragrunt ignores all hidden files and folders in the Terraform source path).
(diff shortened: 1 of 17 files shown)


test/integration_download_test.go
```diff
@@ -0,0 +1,444 @@
+package test
```
yorinasub17, line 1: NOTE: This file is a refactoring of `integration_test.go` to make that main test file smaller. The only new test introduced here is `TestLocalDownloadWithAllowedHiddenFiles` - all other test functions are a straight copy.

denis256 (dismissed):

yorinasub17:
Thanks for review! Prior to merge, I want one sanity check from @brikis98 on the terminology of the new feature: does `include_in_copy` make sense? I got that off [your comment for the inverse](https://github.com/gruntwork-io/terragrunt/issues/394#issuecomment-350758961). I can't come up with anything better, but the one thing that bothers me about it is that it isn't obvious that this configuration only exists for local file terraform sources - it doesn't apply to git sources.

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -63,6 +63,11 @@ The `terraform` block supports the following arguments:
       registry]({{site.baseurl}}/docs/getting-started/quick-start#a-note-about-using-modules-from-the-registry) for more
       information about using modules from the Terraform Registry with Terragrunt.
 
+- `include_in_copy` (attribute): When using local file path sources, this specifies a list of glob patterns (e.g.,
```
brikis98, line 66: Is this purely additive to Terragrunt's default behavior? Or does it allow you to have Terragrunt _not_ copy some items? I'm guessing by the name it's the former... But I worry that the minute we add this, users will also request the latter. I suppose we could add an `exclude_in_copy` feature in the future...

brikis98:
```suggestion
- `include_in_copy` (attribute): When using local file path sources, this specifies a list of glob patterns (e.g.,
```

Is this really the proper context on this feature? It might be worth adding some context like:

When you use the `source` param in your Terragrunt config and run `terragrunt <command>`, Terragrunt will download the code specified at `source` into a scratch folder (`.terragrunt-cache`, by default), copy the code in your current working directory into the same scratch folder, and then run `terraform <command>` in that scratch folder. By default, Terragrunt excludes hidden files and folders during the copy step. This feature allows you to specify glob patterns ...

yorinasub17: Sorry for the delay in responding.

> Is this purely additive to Terragrunt's default behavior? Or does it allow you to have Terragrunt not copy some items? I'm guessing by the name it's the former...

Yes this is purely additive.

> But I worry that the minute we add this, users will also request the latter. I suppose we could add an exclude_in_copy feature in the future...

I'm actually not as concerned about this, primarily because there is limited use case for exclusion in the context of `terragrunt`. I think in practice, it isn't practical to rely on terragrunt to exclude files using a lookup table, and instead should rely on other means (like moving the files to a different folder above levels, or use something like `git`). I think we can wait to add `exclude_in_copy` if someone asks for it.

> Is this really the proper context on this feature?

Ah good catch about mentioning the terragrunt working dir. Updated this section! 83c60b1d370bb4bcd056a90376825a2730c4687b

brikis98 (approved):
LGTM!

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -63,6 +63,18 @@ The `terraform` block supports the following arguments:
...
       information about using modules from the Terraform Registry with Terragrunt.
 
+- `include_in_copy` (attribute): A list of glob patterns (e.g., `*.txt`) that should always be copied into the Terraform
+  working directory. When you use the `source` param in your Terragrunt config and run `terragrunt <command>`,
+  Terragrunt will download the code specified at source into a scratch folder (`.terragrunt-cache`, by default), copy
+  the code in your current working directory into the same scratch folder, and then run `terraform <command>` in that
+  scratch folder. By default, Terragrunt excludes hidden files and folders during the copy step. This feature allows you
+  to specify glob patterns of files that should always be copied from the Terragrunt working directory. Additional notes:
+    - The path should be specified relative to the source directory.
+    - This list is also use when using a local file source (e.g., `source = "../modules/vpc"`). For example, if your
```
rhoboat, line 73:
```suggestion
    - This list is also used when using a local file source (e.g., `source = "../modules/vpc"`). For example, if your
```

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -63,6 +63,18 @@ The `terraform` block supports the following arguments:
       registry]({{site.baseurl}}/docs/getting-started/quick-start#a-note-about-using-modules-from-the-registry) for more
       information about using modules from the Terraform Registry with Terragrunt.
 
+- `include_in_copy` (attribute): A list of glob patterns (e.g., `*.txt`) that should always be copied into the Terraform
```
rhoboat, line 66: Something like this would be more clear:

```suggestion
- `include_in_copy` (attribute): A list of glob patterns (e.g., `["*.txt"]`) that should always be copied into the Terraform
```

yorinasub17:
Thanks for review! I'm going to merge this in now, and open a follow up PR with the suggested changes from @rhoboat (to avoid another review cycle).

docs/_docs/04_reference/config-blocks-and-attributes.md
yorinasub17: Done in https://github.com/gruntwork-io/terragrunt/pull/1941

docs/_docs/04_reference/config-blocks-and-attributes.md
yorinasub17: Done in https://github.com/gruntwork-io/terragrunt/pull/1941

PR #1936, opened 2021-12-03
Add merge strategies for mock outputs with state

### LINKED ISSUE #1934: Deep merge when using `mock_outputs_merge_with_state` in a `dependency` block
Following from https://github.com/gruntwork-io/terragrunt/pull/1765 , it's possible to merge the existing (applied) state into the mocks (state overwrites mocks). However, this is only a shallow merge.

Let's say you have a module that outputs a `map(map(string))`. An existing output of that module could be

```json
{
  "security_groups": {
    "sensitive": false,
    "type": [
      "object",
      {
        "first": [
          "object",
          {
            "arn": "string",
            "description": "string",
            "id": "string",
            "name": "string"
          }
        ]
      }
    ],
    "value": {
      "first": {
        "arn": "arn:aws:ec2:ap-southeast-1:000000000000:security-group/sg-aaaaaaaaaaaaaaaaa",
        "description": "My first security group",
        "id": "sg-aaaaaaaaaaaaaaaaa",
        "name": "my-first-sg"
      }
    }
  }
}
```

and let's say the usage of that module is like so:

```hcl
dependency "security_groups" {
  config_path = "../security_groups"

  mock_outputs_allowed_terraform_commands = ["validate", "plan", "state"]
  mock_outputs_merge_with_state = true
  mock_outputs = {
    security_groups = {
      first = { id = "sg-mock" }
    }
  }
}

inputs = {
  something = [
    dependency.security_groups.outputs.security_groups.first.id
  ]
}
```

If I were to change the inputs of the dependency (the security_groups module), which adds an additional value to the output map, so that it becomes

```diff
{
  "security_groups": {
    "sensitive": false,
    "type": [
      "object",
      {
        "first": [
          "object",
          {
            "arn": "string",
            "description": "string",
            "id": "string",
            "name": "string"
          }
        ],
+        "second": [
+          "object",
+          {
+            "arn": "string",
+            "description": "string",
+            "id": "string",
+            "name": "string"
+          }
+        ]
      }
    ],
    "value": {
      "first": {
        "arn": "arn:aws:ec2:ap-southeast-1:000000000000:security-group/sg-aaaaaaaaaaaaaaaaa",
        "description": "My first security group",
        "id": "sg-aaaaaaaaaaaaaaaaa",
        "name": "my-first-sg"
      },
+      "second": {
+        "arn": "arn:aws:ec2:ap-southeast-1:000000000000:security-group/sg-bbbbbbbbbbbbbbbbb",
+        "description": "My second security group",
+        "id": "sg-bbbbbbbbbbbbbbbbb",
+        "name": "my-second-sg"
+      }
    }
  }
}
```

and attempt to use it BEFORE the updated dependency has been applied

```diff
dependency "security_groups" {
  config_path = "../security_groups"

  mock_outputs_allowed_terraform_commands = ["validate", "plan", "state"]
  mock_outputs_merge_with_state = true
  mock_outputs = {
    security_groups = {
      first = { id = "sg-mock" },
+      second = { id = "sg-mock" }
    }
  }
}

inputs = {
  something = [
    dependency.security_groups.outputs.security_groups.first.id,
+    dependency.security_groups.outputs.security_groups.second.id
  ]
}
```

We'll receive the following error from terragrunt when trying to plan

```
Unsupported attribute; This object does not have an attribute named "second".
```

This is due to the logic in `getTerragruntOutputIfAppliedElseConfiguredDefault` only checking the output key and not supporting a deep merge of the output values https://github.com/gruntwork-io/terragrunt/blob/cc1927746e8883322a0b37394887db4d8d6810ae/config/dependency.go#L322-L327

Would you be open to a PR or discussion on adding support for deep merging the outputs into the mock?

jon-walton:
fixes #1934


config/dependency.go
```diff
@@ -61,7 +65,7 @@ func (targetDepConfig *Dependency) DeepMerge(sourceDepConfig Dependency) error {
 		if targetDepConfig.MockOutputs == nil {
 			targetDepConfig.MockOutputs = sourceDepConfig.MockOutputs
 		} else {
-			newMockOutputs, err := deepMergeCtyMaps(*targetDepConfig.MockOutputs, *sourceDepConfig.MockOutputs)
+			newMockOutputs, err := deepMergeCtyMaps(*targetDepConfig.MockOutputs, *sourceDepConfig.MockOutputs, mergo.WithAppendSlice, mergo.WithOverride)
```
jon-walton, line 68: I don't like exposing `mergo` at this level. Any preference? Having a `deepMergeCtyMaps` and `deepMergeOverrideCtyMaps` also felt wrong

yorinasub17: Hmm what's the rationale for having a different deep merge logic for the state merge? I think it would be confusing to have different deep merge behavior across the terragrunt code base.

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -846,8 +846,17 @@ The `dependency` block supports the following arguments:
 - `mock_outputs_allowed_terraform_commands` (attribute): A list of Terraform commands for which `mock_outputs` are
   allowed. If a command is used where `mock_outputs` is not allowed, and no outputs are available in the target module,
   Terragrunt will throw an error when processing this dependency.
-- `mock_outputs_merge_with_state` (attribute): When `true`, `mock_outputs` and the state outputs will be merged. That is,
-  the `mock_outputs` will be treated as defaults and the real state outputs will overwrite them if the keys clash.
+- `mock_outputs_merge_with_state` (attribute): DEPRECATED. Use mock_outputs_merge_strategy_with_state. When `true`,
```
yorinasub17, line 849:
```suggestion
- `mock_outputs_merge_with_state` (attribute): DEPRECATED. Use `mock_outputs_merge_strategy_with_state`. When `true`, 
```

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -846,8 +846,17 @@ The `dependency` block supports the following arguments:
...
   allowed. If a command is used where `mock_outputs` is not allowed, and no outputs are available in the target module,
   Terragrunt will throw an error when processing this dependency.
-- `mock_outputs_merge_with_state` (attribute): When `true`, `mock_outputs` and the state outputs will be merged. That is,
-  the `mock_outputs` will be treated as defaults and the real state outputs will overwrite them if the keys clash.
+- `mock_outputs_merge_with_state` (attribute): DEPRECATED. Use mock_outputs_merge_strategy_with_state. When `true`, 
+  `mock_outputs` and the state outputs will be merged. That is, the `mock_outputs` will be treated as defaults and the
+  real state outputs will overwrite them if the keys clash.
+- `mock_outputs_merge_strategy_with_state` (attribute): Specifies how any existing state should be merged into the 
+  mocks. Valid values are
+  - `no_merge` - any existing state will be used as is. If the dependency does not have an existing state (it hasn't
```
yorinasub17, line 854:
```suggestion
  - `no_merge` (default) - any existing state will be used as is. If the dependency does not have an existing state (it hasn't 
```

util/reflect.go
```diff
@@ -10,3 +13,34 @@ func KindOf(value interface{}) reflect.Kind {
 	}
 	return valueType.Kind()
 }
+
+func MustWalkTerraformOutput(value interface{}, path ...string) interface{} {
```
yorinasub17, line 17: Can you add docs for this function so it's clear what this is trying to do?

yorinasub17 (commented):
Thanks for your contribution and apologies for the long delay in reviewing this! The overall logic + implementation makes sense to me! The only thing that isn't clear is why you need to have two different deep merge strategies (the two different mergo calls) in the first place.

config/dependency.go
jon-walton: The issue I found with the existing deep merge options (`mergo.WithAppendSlice` and `mergo.WithOverride`) was a `run-all plan` against `fixture-get-output/mock-outputs-merge-strategy-with-state/merge-strategy-with-state-deep` resulted in the following plan for the child module

```
Changes to Outputs:
  ~ test_output1_from_parent               = "value1" -> "fake-output1"
  + test_output2_from_parent               = "fake-abc"
  ~ test_output_list_string                = [
        "a",
      + "fake-list-data",
    ]
  ~ test_output_map_map_string_from_parent = {
      ~ "map_root1"    = {
          ~ "map_root1_sub1" = "map_root1_sub1_value" -> "fake-map_root1_sub1"
        }
      + "not_in_state" = {
          + "abc" = "fake-abc"
        }
    }
```

which is incorrect for a number of reasons

- `test_output_list_string` has the mock appended to the real value
- `test_output_map_map_string_from_parent.map_root.map_root1_sub1` is being replaced from the correct output value to the mock

not using any merge options results in the correct plan

```
Changes to Outputs:
  + test_output2_from_parent               = "fake-abc"
  ~ test_output_map_map_string_from_parent = {
      + "not_in_state" = {
          + "abc" = "fake-abc"
        }
        # (1 unchanged element hidden)
    }
```

jon-walton:
Hi @yorinasub17 no worries about the time to review! I don't expect much in December 😄 

I've pushed updated docs, let me know if that does the job. Another thought on `MustWalkTerraformOutput`, I only needed it for my tests, i wouldn't be surprised if it panics on some cases, would you prefer it to be moved into the test folder instead of `reflect.go` so people don't rely on it?

cheers

config/dependency.go
jon-walton: hi @yorinasub17 , sorry to poke, but do you have any additional thoughts on this? thanks :)

jon-walton: @yorinasub17 Updated, thanks for the suggestion

jon-walton: Hi @yorinasub17 , sorry to poke again (and the delay. this fell off my radar 😢 )

I've made the changes, are we good to go? thanks

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -887,8 +887,17 @@ The `dependency` block supports the following arguments:
...
+  `mock_outputs` and the state outputs will be merged. That is, the `mock_outputs` will be treated as defaults and the
+  real state outputs will overwrite them if the keys clash.
+- `mock_outputs_merge_strategy_with_state` (attribute): Specifies how any existing state should be merged into the 
+  mocks. Valid values are
+  - `no_merge` (default) - any existing state will be used as is. If the dependency does not have an existing state (it 
+    hasn't been applied yet), then the mocks will be used
+  - `shallow` - the existing state will be shallow merged into the mocks. Mocks will only be used where the output does 
+    not already exist in the dependency's state
+  - `deep` - the existing state will be deeply merged into the mocks. If an output is a map, the mock key will be 
+    used where that key does not exist in the state. Lists will not be merged
```
yorinasub17, line 900: Can you call this `deep_map_only`, to differentiate with `deep` merge that appends the lists?

config/dependency.go
```diff
@@ -311,22 +329,18 @@ func getTerragruntOutputIfAppliedElseConfiguredDefault(dependencyConfig Dependen
...
-			mockOutputsMap, err := parseCtyValueToMap(*dependencyConfig.MockOutputs)
-			if err != nil {
-				return outputVal, err
+			mockMergeStrategy := dependencyConfig.getMockOutputsMergeStrategy()
+			switch mockMergeStrategy {
+			case NoMerge:
+				return outputVal, nil
+			case ShallowMerge:
+				return shallowMergeCtyMaps(*outputVal, *dependencyConfig.MockOutputs)
+			case DeepMerge:
```
yorinasub17, line 338: Can you introduce a new `DeepMergeMapOnly` value for the setting, as mentioned above?

yorinasub17 (commented):
Sorry for the delay! Mostly LGTM, but I think we should call this `deep_map_only` at the user config level to differentiate with the existing semantics of `deep` merging.

jon-walton:
hi @yorinasub17 , all comments should be resolved now :+1:

yorinasub17 (dismissed):
Updates LGTM! Thanks for working through this! Will kick off a regression build, and if that passes, we can merge this in.

yorinasub17:
Looks like one of the tests failed with the following error:

```
    integration_test.go:2404:
                Error Trace:    integration_test.go:2404
                Error:          Not equal:
                                expected: "value1"
                                actual  : "fake-output1"

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -value1
                                +fake-output1
                Test:           TestDependencyMockOutputMergeStrategyWithStateDeepMapOnly
    integration_test.go:2406:
                Error Trace:    integration_test.go:2406
                Error:          Not equal:
                                expected: "map_root1_sub1_value"
                                actual  : "fake-map_root1_sub1"

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -map_root1_sub1_value
                                +fake-map_root1_sub1
                Test:           TestDependencyMockOutputMergeStrategyWithStateDeepMapOnly
    integration_test.go:2408:
                Error Trace:    integration_test.go:2408
                Error:          Not equal:
                                expected: "a"
                                actual  : "fake-list-data"

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -a
                                +fake-list-data
                Test:           TestDependencyMockOutputMergeStrategyWithStateDeepMapOnly
```

jon-walton:
sorry, I missed the tfstate 😢

yorinasub17 (approved):
Build passed! Going to go ahead and merge this in now. Thanks for your contribution and patience to work through all the updates!

matt-land:
Thank you, this just saved our bacon.

PR #1946, opened 2021-12-13
fix terraformPath for dependency

### LINKED ISSUE #1919: Add TerraformPath in dependency block to get output from different tfstate version

leonardobiffi:
Resolve #1919

### DIFF
--- a/config/dependency.go
+++ b/config/dependency.go
@@ -474,7 +474,7 @@ func cloneTerragruntOptionsForDependencyOutput(terragruntOptions *options.Terrag
 	if err != nil {
 		return nil, err
 	}
-	if terragruntOptions.TerraformPath == options.TERRAFORM_DEFAULT_PATH && partialTerragruntConfig.TerraformBinary != "" {
+	if partialTerragruntConfig.TerraformBinary != "" {
 		targetOptions.TerraformPath = partialTerragruntConfig.TerraformBinary
 	}
 


leonardobiffi:
@yorinasub17 
I investigated the bug of not being able to use the `terraform_binary` value in the dependency file.
In my case, `TerraformPath` didn't have the value set in the dependency file because i run the command with the flag.

```
/usr/local/bin/terragrunt/0.25.1/terragrunt plan --terragrunt-non-interactive -out=tfplan.binary --terragrunt-tfpath /usr/local/bin/terraform/0.13.3/terraform
```

So the `terragruntOptions.TerraformPath` was `/usr/local/bin/terraform/0.13.3/terraform` , and in the if statement it was ignored

yorinasub17 (approved):
This makes sense to me! I'll kick off a regression build and if it passes, we can merge this in!

yorinasub17:
Build failed, but it's because of an out of data branch. Once I rebased, it passed so will go ahead and merge this in!

Issue #1944, opened 2021-12-13
--terragrunt-modules-that-include does not work


roehrijn:
Everytime I try to use `--terragrunt-modules-that-include`, `terragrunt` simply returns this and nothing is executed:

      [yyyy@XXXX:~/tg]$ terragrunt run-all plan --terragrunt-modules-that-include incl.hcl 
      INFO[0000] The stack at /home/yyyy/tg will be processed in the following order for command plan: 
      [yyyy@XXXX:~/tg]$ 

I can reproduce this with a very simple project: 
      .
      ├── first
      │   ├── main.tf
      │   └── terragrunt.hcl
      └── incl.hcl

Where the `terragrunt.hcl` file does nothing but include the `incl.hcl` (all other files are empty):

      include "incl" {
          path = "../incl.hcl"
      }

Even after turning on debug logging, I have no glue why terragrunt is constantly refusing to run my stack:

      yyyy@XXXX:~/tg]$ terragrunt run-all plan --terragrunt-modules-that-include incl.hcl --terragrunt-log-level debug
      DEBU[0000] Did not find any locals block: skipping evaluation.  prefix=[/home/yyyy/tg/first] 
      DEBU[0000] Did not find any locals block: skipping evaluation.  prefix=[/home/yyyy/tg/first] 
      DEBU[0000] [Partial] Included config ../incl.hcl has strategy shallow merge: merging config in (shallow).  prefix=[/home/yyyy/tg/first] 
      DEBU[0000] Setting download directory for module /home/yyyy/tg/first to /home/yyyy/tg/first/.terragrunt-cache 
      DEBU[0000] Stack at /home/yyyy/tg:
        => Module /home/yyyy/tg/first (excluded: true, assume applied: false, dependencies: []) 
      INFO[0000] The stack at /home/yyyy/tg will be processed in the following order for command plan: 
      [yyyy@XXXX:~/tg]$

It says the module `/home/yyyy/tg/first` was excluded but when I simply remove the `--terragrunt-modules-that-include` attribute from the command line, my stack is executed.

`terragrunt version v0.35.14`

denis256:
I will take a look at this one, looks like an issue with the detection of included files in
https://github.com/gruntwork-io/terragrunt/blob/master/configstack/module.go#L258

yorinasub17:
Will close this when the release is cut.

yorinasub17:
[...]

PR #1948, opened 2021-12-15
#1944 Updated logic to resolve full path on includes

denis256:
Updated module exclusion to first resolve path based on the module path, without this change, includes like bellow are ignored

```
app/terragrunt.hcl:

  include "incl" {
      path = "../parent.hcl"
  }

parent.hcl

locals {
    overwrite_var = run_cmd("echo", "parent hcl")
}

```

Fixes: https://github.com/gruntwork-io/terragrunt/issues/1944


yorinasub17 (approved):
Ah makes sense! Thanks for fixing this! Just kicked off a build and if that passes, can merge this in!

yorinasub17:
Build passed, so will merge this in!

Issue #1947, opened 2021-12-14
Sops decrypt raw file


sebastianmacarescu:
According to documentation for sops_decrypt_file: Only YAML and JSON formats are supported by sops_decrypt_file.
It would be really helpful to introduce a "raw" file type like here: https://github.com/carlpett/terraform-provider-sops

This will allow us to store any kind of secrets in our code and be able to push it to AWS SSM with terraform. (eg private ssh keys)

yorinasub17:
Fixed in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.35.16

PR #1956, opened 2021-12-17
#1947 Additional formats for sops_decrypt_file

denis256:
Updated `sops_decrypt_file` to support INI, ENV, and raw text files

Fixes: https://github.com/gruntwork-io/terragrunt/issues/1947


yorinasub17 (approved):
Build passed, so will merge this in! Thanks for fixing!

PR #1958, opened 2021-12-19
Fix typo in config-blocks-and-attributes docs

brad-alexander:


rhoboat (approved):
Thanks for the fix! LGTM.

PR #1967, opened 2022-01-05
Implement Error Hooks

### LINKED ISSUE #1522: Proposal: Error hooks
The idea behind error hooks is to execute arbitrary shell commands when a terraform command returns an error. 

Example syntax would be similar to `before_hook` and `after_hook` with required properties `commands` and `execute` and instead of `run_on_error` a different property `on_errors` would be required that would describe a list of errors on which should the hook run, described in the same manner as `retryable_errors`.

An example configuration that imports would run a shell script with `terraform import`:
```hcl
terraform {
  error_hook "import_resource" {
    commands  = ["apply"]
    execute   = ["./import_known_resource.sh"]
    on_errors = [
      "(?s).*A resource with the ID .* already exists.*",
    ]
  }
}
```

The idea behind this proposal is that there are some resources on terraform-providers that create an 'umbrella' resource which can create some child-resources on the cloud provider which are not being represented as a terraform resource. In some cases, the default child-resource configuration is ok, but when that resource needs to be modified, it becomes problematic.

One example of this would be an `azurerm` resource `azurerm_api_management` creates 2 `azurerm_api_management_product` instances which have one `azurerm_api_management_product_policy` each. The `product` can be accessed as `data` but the `product_policy` needs to be imported to state in order to be accessible from terraform.

smitthakkar96:
Fixes #1522


smitthakkar96:
@yorinasub17 @Ben10k please review

cli/cli_app.go
```diff
@@ -602,6 +602,41 @@ func shouldApplyAwsProviderPatch(terragruntOptions *options.TerragruntOptions) b
...
+	if len(hooks) == 0 || previousExecErrors.ErrorOrNil() == nil {
+		return nil
+	}
+
+	var errorsOccured *multierror.Error
+
+	terragruntOptions.Logger.Debugf("Detected %d error Hooks", len(hooks))
+
+	for _, curHook := range hooks {
+		if util.MatchesAny(curHook.OnErrors, previousExecErrors.GoString()) && util.ListContainsElement(curHook.Commands, terragruntOptions.TerraformCommand){
```
denis256, line 615: I think `pre-commit run --all-files` should be executed
```
TerraformCommand)<space>{
```

config/config.go
```diff
@@ -329,10 +341,24 @@ func (conf *TerraformConfig) GetAfterHooks() []Hook {
...
+	if conf == nil {
+		return nil
+	}
+
+	return conf.ErrorHooks
+}
+
 func (conf *TerraformConfig) ValidateHooks() error {
-	allHooks := append(conf.GetBeforeHooks(), conf.GetAfterHooks()...)
+	before_and_after_hooks := append(conf.GetBeforeHooks(), conf.GetAfterHooks()...)
```
denis256, line 353: `camelCase` is preferred over `snake_case`

test/fixture-hooks/before-after-and-on-error/terragrunt.hcl
```diff
@@ -0,0 +1,35 @@
...
+    run_on_error = true
+  }
+
+  after_hook "after_hook_2" {
+    commands = ["terragrunt-read-config"]
+    execute = ["echo", "AFTER_TERRAGRUNT_READ_CONFIG"]
+    run_on_error = true
+  }
+
+  error_hook "error_hook_1" {
```
denis256, line 30: Documentation should be updated to reference `error_hook`
https://github.com/gruntwork-io/terragrunt/blob/master/docs/_docs/04_reference/config-blocks-and-attributes.md

smitthakkar96:
@denis256 addressed your comments

shemul (dismissed):

docs/_docs/02_features/hooks.md
```diff
@@ -58,3 +58,19 @@ to Terraform.
...
 You can learn more about all the various configuration options supported in [the reference docs for the terraform
 block](/docs/reference/config-blocks-and-attributes/#terraform).
+
+## Error Hooks
+Think of error hooks like an exception handler, they allow you to specifiy a list of expressions that can be used to catch errors and run custom commands when that occur.
+
+Here is an example:
+``` hcl
+terraform {
+  error_hook "import_resource" {
```
denis256, line 68: I think `error_hook` should be referenced in the "Blocks" list like `after_hook` / `before_hook`, otherwise, users will not know that such block can be used in HCL

https://terragrunt.gruntwork.io/docs/reference/config-blocks-and-attributes/

https://github.com/gruntwork-io/terragrunt/blob/master/docs/_docs/04_reference/config-blocks-and-attributes.md

denis256 (approved):

docs/_docs/02_features/hooks.md
smitthakkar96: good point, let me update that

smitthakkar96: updated https://github.com/gruntwork-io/terragrunt/pull/1967/commits/a467f8c6f0ec3f49d9ebf8f6b4800502df71702b

denis256 (dismissed):

smitthakkar96:
@denis256 thanks for the approval, I see you are also the codeowner for this repo. Could please help merging this?

smitthakkar96:
@yorinasub17 @rhoboat @zackproser @brikis98 can we have one more reviewon this as suggested by @denis256 ?

rhoboat:
Reviewing!

docs/_docs/02_features/hooks.md
```diff
@@ -58,3 +58,19 @@ to Terraform.
 
 You can learn more about all the various configuration options supported in [the reference docs for the terraform
 block](/docs/reference/config-blocks-and-attributes/#terraform).
+
+## Error Hooks
+Think of error hooks like an exception handler, they allow you to specifiy a list of expressions that can be used to catch errors and run custom commands when that occur.
```
rhoboat, line 63: Just some grammatical suggestions here.

```suggestion
Think of error hooks like exception handlers. They allow you to specifiy a list of expressions that can be used to catch errors and run custom commands when those errors occur.
```

docs/_docs/02_features/hooks.md
```diff
@@ -1,6 +1,6 @@
 ---
 layout: collection-browser-doc
-title: Before and After Hooks
+title: Hooks
 category: features
 categories_url: features
 excerpt: Learn how to execute custom code before or after running Terraform.
```
rhoboat, line 6: We should update this too.

```suggestion
excerpt: Learn how to execute custom code before or after running Terraform, or when errors occur.
```

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -104,6 +104,7 @@ The `terraform` block supports the following arguments:
 - `after_hook` (block): Nested blocks used to specify command hooks that should be run after `terraform` is called.
   Hooks run from the terragrunt configuration directory (the directory where `terragrunt.hcl` lives). Supports the same
   arguments as `before_hook`.
+- `error_hook` (block): Needed blocks used to specify command hooks that run incase of error that matches one of the expression from the list of expressions specified in `on_errors` attribute.
```
rhoboat, line 107:
```suggestion
- `error_hook` (block): Nested blocks used to specify command hooks that run when an error is thrown. The
error must match one of the expressions listed in the `on_errors` attribute.
```

smitthakkar96:
Thanks @rhoboat for your suggestions. Just merged them :)

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -190,6 +191,14 @@ terraform {
...
   }
 
+  # Run incase of any error that occurs during apply or plan
+  error_hook "error_hook_1" {
+    commands  = ["apply", "plan"]
+    execute   = ["echo", "Error Hook executed"]
+    on_errors = [
+      ".*",
+    ]
+  }
```
rhoboat, line 202: Add an extra line here.

```suggestion
  }
  
```

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -190,6 +191,14 @@ terraform {
     run_on_error = true
   }
 
+  # Run incase of any error that occurs during apply or plan
```
rhoboat, line 195:
```suggestion
  # After an error occurs during apply or plan, run "echo Error Hook executed". This hook is configured so that it will run
  # after any error, with the ".*" expression.
```

config/config.go
```diff
@@ -329,10 +341,24 @@ func (conf *TerraformConfig) GetAfterHooks() []Hook {
...
+	beforeAndAfterHooks := append(conf.GetBeforeHooks(), conf.GetAfterHooks()...)
+
+	for _, curHook := range beforeAndAfterHooks {
+		if len(curHook.Execute) < 1 || curHook.Execute[0] == "" {
+			return InvalidArgError(fmt.Sprintf("Error with hook %s. Need at least one non-empty argument in 'execute'.", curHook.Name))
+		}
+	}
 
-	for _, curHook := range allHooks {
+	for _, curHook := range conf.GetErrorHooks() {
```
rhoboat, line 361: Hmm.. so we run the before and after hooks first, then we run the error hooks. We should mention this somewhere in the docs.

denis256 (dismissed):

config/config_as_cty_test.go
```diff
@@ -44,6 +44,14 @@ func TestTerragruntConfigAsCtyDrift(t *testing.T) {
...
 				},
 			},
+			ErrorHooks: []ErrorHook{
+				ErrorHook{
+					Name:     "init",
+					Commands: []string{"init"},
+					Execute:  []string{"true"},
+					OnErrors: []string{".*"},
+				},
+			},
```
rhoboat, line 54: Why is the ErrorHooks block not included under the Terraform one above, so that it would be nested at the same level as the Before/AfterHooks?

docs/_docs/02_features/hooks.md
```diff
@@ -1,9 +1,9 @@
 ---
 layout: collection-browser-doc
-title: Before and After Hooks
+title: Hooks
```
rhoboat, line 3: What about this instead?

```suggestion
title: Before, After, and Error Hooks
```

rhoboat (commented):
I have some questions.

Does it make sense to always process the error hooks after the other hooks? If you have an after hook that is specified to run on error, what happens? This hook will run first, then the error hook. Is that always the right sequencing? How do we know?

I left a few other nits and questions, but I do think this is a good feature!

config/config_as_cty_test.go
smitthakkar96: I didn't understand the comment, it is at same level as before/after hooks from what I can see

config/config.go
smitthakkar96: This is addressed https://github.com/gruntwork-io/terragrunt/pull/1967/commits/94037c72168be46d1cb7dbb47d687aea0933148c

smitthakkar96:
@rhoboat good question! Initially I was confused about the execution status I wasn't sure where should I start this dicussion but I think we can deprecate `run_on_error` from before/after hooks in favour of error hooks in future. Why? 
- Because it's very confusing (which one to use)
- Currently if there is no way for the downstream command/script that will be executed by the before/after hooks to know whether it is being executed because of an error or success. If others agree maybe we should add deprecation warning for `run_on_error`. WDYT?

rhoboat:
@smitthakkar96 This turned out to be a long response, so I've made sections. Bear with me! 🐻 

## Deprecating the `run_on_error`
For argument's sake: let's see what's the difference between an after_hook with run_on_error and an error_hook:

```
  after_hook "after_hook_1" {
    commands = ["apply", "plan"]
    execute = ["touch","after.out"]
    run_on_error = true
  }
```

The closest we can get to that with error_hook is

```
  error_hook "error_hook_1" {
    commands  = ["apply", "plan"]
    execute   = ["touch", "after.out"]
    on_errors = [".*"]
  }
```

Are these identical? No, because the after_hook will run regardless of whether there are errors, but the error_hook will only run if there are errors. Taken together, the error_hook is redundant. So I see the error_hook as a more specific way to handle the error-only case.

If you get rid of `run_on_error`, you will have to do:

```
  after_hook "after_hook_1" {
    commands = ["apply", "plan"]
    execute = ["touch","after.out"]
  }
```

plus

```
  error_hook "error_hook_1" {
    commands  = ["apply", "plan"]
    execute   = ["touch", "after.out"]
    on_errors = [".*"]
  }
```
Now, the after_hook makes sure it touches after.out only if there are no errors, and the error_hook makes sure it touches after.out if there are errors.

So I think we can't deprecate the `run_on_error` feature without making it harder to do the same thing. 

## Error/success
> Currently if there is no way for the downstream command/script that will be executed by the before/after hooks to know whether it is being executed because of an error or success.

That is true, however, this the main feature for `run_on_error`. It's for the set of cases where you don't want an error to interrupt execution of a particular hook. Is that still useful / needed? (I would also like to hear from people whether it is.) With the addition of `error_hook` we can handle specific error cases, so I do think this is an important feature.

## Execution order

I've been noodling on this a little more, and I think if we're going with error handling as other languages do it, the ordering should be different. I'll start by describing what I think the intent of each of the hooks is:

`after_hook` : run this thing if there's no errors
`after_hook` with `run_on_error` : always run this thing (like `always` and `finally` in typical exception handling parlance)
`error_hook` : run this thing only if there's some specific errors (or all errors with `".*"`) and it's meant to be used specifically to _handle_ the error. In the docs included for this PR, we're advising that you think of this like error handling.. That is, when you catch an error, you execute the catch block _before_ the finally/always blocks. Right?

Given that, I think the ideal execution order should be the following.

1. `before_hooks`
2. `after_hooks` without `run_on_error`
3. `error_hooks`
4. `after_hooks` with `run_on_error`

I also think this is doable from an implementation standpoint, but will not be as clean as you have it in the PR now.

### More intuitive naming?

The above feels intuitive from an execution standpoint, but it's hard to keep this in mind when writing terragrunt. It would be nice if we could do this, but make it easy to understand. Would naming help? Would it be any better if `run_on_error` for the `after_hook` was renamed to `finally` or `always`? I feel like `run_on_error` is clear, but `always = true` also reads better. `run_on_error` could be taken to mean "only run in case of error", which is wrong. `always` doesn't contain the word "error" but it is very clear that it runs "always" (whether or not there's an error).

But there's still the mental overhead of remembering that an `error_hook` which handles the specific error thrown will run before the `after_hook` in cases of error. Is this really as bad as I'm making it sound? WDYT?

smitthakkar96:
Hey thanks for putting a great deal of thoughts into the execution order. Here is what I think

> That is true, however, this the main feature for `run_on_error`. It's for the set of cases where you don't want an error to interrupt execution of a particular hook. Is that still useful / needed? (I would also like to hear from people whether it is.)

I just realised that I have a such use-case too. During plan regardless of error I want to continue the execution. My use-case was we have a utility that would move state files in S3 incase of renames and path changes (we have folders split per teams `squads/<squad_name>/module`). We rely on `git diff` to detect renames and move operations. When service responsibility changes the modules move from one squad folder to another. What we do with hooks is something like below:

``` hcl
  before_hook "before_plan" {
    commands     = ["plan"]
    execute      = [
      local.migrate_states_binary,
      "--yes",
    ]
    working_dir = local.repo_path
  }
  after_hook "after_plan" {
    commands     = ["plan"]
    execute      = [
      local.migrate_states_binary,
      "--dry-run",
      "--rollback",
    ]
    working_dir = local.repo_path
    run_on_error = true
  }
  before_hook "before_apply" {
    commands     = ["apply"]
    execute      = [
      local.migrate_states_binary,
      "--yes"
    ]
    working_dir = local.repo_path
  }
  error_hook "rollback_on_apply_failure" {
    commands     = ["apply"]
    execute      = [
      local.migrate_states_binary,
     "--rollback",
      "--yes"
    ]
    on_errors = [".*"]
  }
}
```
I guess above the example would look ugly without `run_on_error` where after plan we want to rollback the state changes incase developer doesn't decide to apply.


> The above feels intuitive from an execution standpoint, but it's hard to keep this in mind when writing terragrunt. It would be nice if we could do this, but make it easy to understand. Would naming help? Would it be any better if `run_on_error` for the `after_hook` was renamed to `finally` or `always`? I feel like `run_on_error` is clear, but `always = true` also reads better. `run_on_error` could be taken to mean "only run in case of error", which is wrong. `always` doesn't contain the word "error" but it is very clear that it runs "always" (whether or not there's an error).

I like the idea of renaming it to `always` however it will break backwards compatibility, not sure what is the guideline to do that in `gruntworks`, if changing name is hard maybe we can improve the documentation as a part of this PR explaining when to use `run_on_error`, when to use `error_hooks` and also document the order of execution. 

> Given that, I think the ideal execution order should be the following.

I like the execution order and I think if well documented there won't be mental overhead of remembering that an `error_hook` is executed before `after_hooks` with `run_on_error`. 

Should I work on improving docs and changing the execution order and we can open a separate issue (we can get some more feedback from community, maybe gruntworks can reach out to their customers internally too asking for feedback on naming) to rename `run_on_apply` to `always`?

config/config.go
yorinasub17, line 361: NIT: reduce this to the following for DRY:

```go
beforeAndAfterHooks := append(conf.GetBeforeHooks(), conf.GetAfterHooks()...)
allHooks := append(beforeAndAfterHooks, conf.GetErrorHooks()...)
for _, curHook := range allHooks {
```

docs/_docs/02_features/hooks.md
```diff
@@ -58,3 +58,19 @@ to Terraform.
 
 You can learn more about all the various configuration options supported in [the reference docs for the terraform
 block](/docs/reference/config-blocks-and-attributes/#terraform).
+
+## Error Hooks
+Think of error hooks like exception handlers. They allow you to specifiy a list of expressions that can be used to catch errors and run custom commands when those errors occur. Error hooks are executed after the before/after hooks.
```
yorinasub17, line 63: To be consistent:

```suggestion
*Error hooks* are a special type of after hook that act as exception handlers. They allow you to specify a list of expressions that can be used to catch errors and run custom commands when those errors occur. Error hooks are executed after the before/after hooks.
```

yorinasub17 (commented):
Thanks for the contribution! This mostly LGTM, but I found a few minor nits, and one bug: the new error hook blocks need to be handled in the respective merge functions for `include`. Specifically:

- https://github.com/gruntwork-io/terragrunt/blob/master/config/include.go#L306
- https://github.com/gruntwork-io/terragrunt/blob/master/config/include.go#L450

smitthakkar96:
@yorinasub17 thanks for your review, apart from nits and fixing the bug, should I also change execution order of hooks like @rhoboat suggested? What are your thoughts on execution order?

yorinasub17:
RE: execution order

I think ideally and what is most intuitive, is if the hooks are executed in the order that it is defined. That is, if you had the following:

```hcl
terraform {
  after_hook "always" {
    commands = ["plan"]
    execute = ["echo", "always"]
    run_on_error = true
  }
  after_hook "on_success" {
    commands = ["plan"]
    execute = ["echo", "success"]
  }
  error_hook "on_error" {
    commands = ["plan"]
    execute = ["echo", "error"]
    on_errors = [".*"]
  }
  after_hook "last" {
    commands = ["plan"]
    execute = ["echo", "last"]
    run_on_error = true
  }
}
```

On error, the output should be:

```
always
error
last
```

and on success, it should be:

```
always
success
last
```

This mimics how terraform `local-exec` provisioners work, and personally would be much easier to reason about because you don't have to know terragrunt internals of how hooks are processed.

This is the disadvantage of introducing a new block, as it makes it hard to know the order of the blocks when they are interleaved.

The only way I think we can implement this is if we use the approach terraform takes of using labels to identify the hook type - E.g.:

```hcl
terraform {
  hook "after" "always" {
    commands = ["plan"]
    execute = ["echo", "always"]
    run_on_error = true
  }
  hook "after" "on_success" {
    commands = ["plan"]
    execute = ["echo", "success"]
  }
  hook "error" "on_error" {
    commands = ["plan"]
    execute = ["echo", "error"]
    on_errors = [".*"]
  }
  hook "after" "last" {
    commands = ["plan"]
    execute = ["echo", "last"]
    run_on_error = true
  }
}
```

That said, this is a rather drastic backward incompatible change, but FWIW, terragrunt is still a pre-1.0 project and we haven't really identified the stable API yet. I think this is at least worth a consideration in an RFC discussion.

In any case, at this stage, I believe the best course of action is to avoid prescribing and mentioning an execution order and leaving that as undefined. That is, the basic execution order of after and error hooks running after the commands is defined and set, but in what order those individual hooks are run is not specified or guaranteed (similar to the unspecified order of iterating through a map in most languages). NOTE: we should document that it is undefined.

Then separately, we should open an RFC discussing what the actual execution order of hooks we want terragrunt to enforce and implement.

That way, we can consider the ideal UX of hook ordering without being tied to implementation details, and also we can be more systematic about the proposal rather than it being implemented/discussed on the fly in a PR.

rhoboat:
Thanks @yorinasub17 for further insights into the execution order! It makes sense to respect the lexical ordering (i.e., the order that the hooks appear in the code). I didn't even think of that. 

I agree with execution order being a separate issue we can tackle later. Let's not block this PR on that. I think it makes good enough sense as it is!

config/config_as_cty_test.go
rhoboat: @smitthakkar96 Did you make this change? Just wanted to give this thread a bump, since it might have gotten missed!

config/config.go
smitthakkar96: won't work, it will probably fail with error, as `ErrorHook` is a different type. The reason for creating different type of hook as `OnErrors` attribute doesn't need to be in before and after hooks. 

```
cannot use conf.GetErrorHooks() (type []ErrorHook) as type []Hook in append
```

I guess the instead of creating new type `ErrorHook` I will have to use the same type, only disadvantage to that will be OnErrors attribute won't be of any use for other two hooks (i.e before and after we can just add validation I guess).

yorinasub17: Ah that's right makes sense. You can keep it as is for now then!

smitthakkar96: I realised `mergeHooks` and `getIndexOfHookWithName` need to be duplicated too, any suggestions for better design/way to do it. Sorry I my relatively new to Golang

config/config_as_cty_test.go
smitthakkar96: Hey sorry I didn't get the comment, what am I missing here?

config/config.go
yorinasub17: It's out of scope of this PR, but the way to handle this is to define a function `validate` on the `Hook` and `ErrorHook` type, and then define an `interface`. Then, you can construct a list of the interface type, and call the `validate` function in the loop.

config/config_as_cty_test.go
rhoboat: Ah, I see. I was seeing things indented incorrectly due to the GitHub UI. Looking at the raw file, everything is formatted right. Closing this thread. Sorry for the noise!

rhoboat (approved):
LGTM!

smitthakkar96:
> bug: the new error hook blocks need to be handled in the respective merge functions for `include`. Specifically:
@yorinasub17 @rhoboat I was about to push a commit to above comment but PR got merged so I created a new PR to handle the merge part https://github.com/gruntwork-io/terragrunt/pull/1982

alikhil:
Hello!

Is it possible to pass catched error to hook now? I checked the code and it seems like it cannot

zackproser:
@alikhil could you please open a new issue and include all the relevant details so that we can better assist you? 

* Which version are you running?
* What command are you typing?
* What output are you expecting? 
* What output are you getting? 

Thank you!

PR #1990, opened 2022-01-20
Bump terraform version in CI to 1.1

yorinasub17:
So that we start testing against the latest terraform.


denis256 (approved):

yorinasub17:
Thanks for review! Will merge and release now.

PR #2000, opened 2022-01-27
Implement a mechanism for providing macro placeholders and substitutions

erhudy:
This implements a mechanism to allow Terragrunt to recognize and replace
specific macro placeholders in the Terraform command line with a
value specific to that instantiation of Terraform. The only implemented
macro in this commit is `::TERRAGRUNT_DIR::`, which is replaced with
the directory of the Terragrunt file.

The use case here is to allow Terragrunt to interact smoothly with our GitLab CI
processes and artifact archival by being able to control exactly where our generated
plans are created, instead of them being created deep inside `.terragrunt-cache`
somewhere.

### DIFF
--- a/cli/cli_app.go
+++ b/cli/cli_app.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"regexp"
 	"strings"
 	"sync"
@@ -63,6 +64,7 @@ const (
 	optTerragruntFetchDependencyOutputFromState = "terragrunt-fetch-dependency-output-from-state"
 	optTerragruntUsePartialParseConfigCache     = "terragrunt-use-partial-parse-config-cache"
 	optTerragruntOutputWithMetadata             = "with-metadata"
+	optTerragruntSubstituteMacros               = "terragrunt-substitute-macros"
 )
 
 var allTerragruntBooleanOpts = []string{
@@ -81,6 +83,7 @@ var allTerragruntBooleanOpts = []string{
 	optTerragruntFetchDependencyOutputFromState,
 	optTerragruntUsePartialParseConfigCache,
 	optTerragruntOutputWithMetadata,
+	optTerragruntSubstituteMacros,
 }
 var allTerragruntStringOpts = []string{
 	optTerragruntConfig,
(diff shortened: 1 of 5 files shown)


lorengordon:
I love this idea. I implemented something similar in a wrapper cli, to pattern match the terragrunt working directory and replace matches in the terraform command. Very handy for `state mv`, where the resource id includes an account/region/env-specific identifier (and those attributes are in the path to the working directory)...

erhudy:
Test results are here: https://gist.github.com/erhudy/a633e186ff37f6e5148d6b16245dca13

Most of the failures seem to be because I am running the test suite on an M1 Mac and certain provider/version combos are not available for that platform from the HashiCorp Registry.

cli/cli_app.go
```diff
@@ -278,6 +282,34 @@ var TERRAFORM_HELP_FLAGS = []string{
...
+	return map[string]func() string{
+		"::TERRAGRUNT_DIR::": func() string { return filepath.Dir(terragruntOptions.TerragruntConfigPath) },
+	}
+}
+
+// Takes terragruntOptions and mutates the args list such that all macros are replaced with their substitutions,
+// if SubstituteMacros is true.
+func performMacroSubstitutions(terragruntOptions *options.TerragruntOptions) error {
+	if terragruntOptions.SubstituteMacros {
+		newTerraformCliArgs := []string{}
```
denis256, line 296: can be `var newTerraformCliArgs []string`

cli/cli_app.go
```diff
@@ -278,6 +282,34 @@ var TERRAFORM_HELP_FLAGS = []string{
...
+func getMacroReplacements(terragruntOptions *options.TerragruntOptions) map[string]func() string {
+	return map[string]func() string{
+		"::TERRAGRUNT_DIR::": func() string { return filepath.Dir(terragruntOptions.TerragruntConfigPath) },
+	}
+}
+
+// Takes terragruntOptions and mutates the args list such that all macros are replaced with their substitutions,
+// if SubstituteMacros is true.
+func performMacroSubstitutions(terragruntOptions *options.TerragruntOptions) error {
+	if terragruntOptions.SubstituteMacros {
```
denis256, line 295: To reduce nested blocks, it can be

```
if !terragruntOptions.SubstituteMacros {
  return
}
```

cli/cli_app.go
```diff
@@ -278,6 +282,34 @@ var TERRAFORM_HELP_FLAGS = []string{
...
+		newTerraformCliArgs := []string{}
+		for _, arg := range terragruntOptions.TerraformCliArgs {
+			newArg := arg
+			for macro, replacementFunc := range getMacroReplacements(terragruntOptions) {
+				if strings.Contains(arg, macro) {
+					newArg = strings.ReplaceAll(arg, macro, replacementFunc())
```
denis256, line 301: Can happen situation when `newArg` will contain another macro, and it will not be replaced?

cli/cli_app.go
```diff
@@ -278,6 +282,34 @@ var TERRAFORM_HELP_FLAGS = []string{
...
+				}
+				newTerraformCliArgs = append(newTerraformCliArgs, newArg)
+			}
+		}
+		terragruntOptions.TerraformCliArgs = newTerraformCliArgs
+	}
+
+	// this doesn't currently do anything that could return an error, but maybe in the future it might
+	return nil
```
denis256, line 310: Not sure if we need to return an error since each time it returns nil

erhudy:
@denis256 Thanks for the feedback, made changes to address your comments.

lstn:
@denis256 any update on getting something like this merged in?

erhudy:
For the record, I've been using this in my company's internal version of Terragrunt since I made this PR (so over a year). If there is still interest in adding it, I'll fix the merge conflicts.

lorengordon:
I am still also interested :)

lstn:
Thinking a bit more about it, something that would help my usecase it would be nice if eventually the scope of this included defining custom macros through configuration instead of solely relying on hardcoded macros like here. Maybe that's a next step though :)

Issue #2003, opened 2022-01-31
Terragrunt cannot download module from 3rd party registry: 404


atikhono:
Hi, I'm facing an issue with tfr source when using a 3rd party Terraform registry implementation.
Module `/download` URL returns `X-Terraform-Get: ./archive.tar.gz`. The implementation in Terragrunt is not corrent: https://github.com/gruntwork-io/terragrunt/blob/master/internal/tfr/getter.go#L251. The comments in the code are good, but the code does not do what it's supposed to, ie does not prepend base module path for relative X-Terraform-Get values. This must be `fmt.Sprintf("%v://%v%v", url.Scheme, url.Host, path.Join(url.Path, "..", terraformGet))` for `./` and `/`.

yorinasub17:
Fixed in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.36.1. Binaries should be published shortly

PR #2005, opened 2022-01-31
Handle relative paths correctly when fetching tfr

yorinasub17:
Fixes https://github.com/gruntwork-io/terragrunt/issues/2003

This updates the relative path handling to use the same approach that terraform uses internally by relying on the `URL` struct to combine reference paths, instead of manually doing it.

### DIFF
--- a/internal/tfr/getter.go
+++ b/internal/tfr/getter.go
@@ -125,7 +125,7 @@ func (tfrGetter *TerraformRegistryGetter) Get(dstPath string, srcURL *url.URL) e
 		return err
 	}
 
-	downloadURL, err := getDownloadURLFromHeader(ctx, *moduleURL, terraformGet)
+	downloadURL, err := getDownloadURLFromHeader(*moduleURL, terraformGet)
 	if err != nil {
 		return err
 	}
@@ -242,7 +242,7 @@ func getTerraformGetHeader(ctx context.Context, url url.URL) (string, error) {
 
 // getDownloadURLFromHeader checks if the content of the X-Terraform-GET header contains the base url
 // and prepends it if not
-func getDownloadURLFromHeader(ctx context.Context, moduleURL url.URL, terraformGet string) (string, error) {
+func getDownloadURLFromHeader(moduleURL url.URL, terraformGet string) (string, error) {
 	// If url from X-Terrafrom-Get Header seems to be a relative url,
 	// append scheme and host from url used for getting the download url
 	// because third-party registry implementations may not "know" their own absolute URLs if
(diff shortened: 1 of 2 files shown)


internal/tfr/getter.go
```diff
@@ -242,13 +242,18 @@ func getTerraformGetHeader(ctx context.Context, url url.URL) (string, error) {
 
 // getDownloadURLFromHeader checks if the content of the X-Terraform-GET header contains the base url
 // and prepends it if not
-func getDownloadURLFromHeader(ctx context.Context, url url.URL, terraformGet string) (string, error) {
+func getDownloadURLFromHeader(ctx context.Context, moduleURL url.URL, terraformGet string) (string, error) {
```
denis256, line 245: I was wondering if `ctx` is required

internal/tfr/getter_test.go
```diff
@@ -47,16 +47,60 @@ func TestGetDownloadURLFromHeaderWithPrefixedURL(t *testing.T) {
...
+			expectedResult: "https://registry.terraform.io/modules/terraform-aws-modules/terraform-aws-vpc",
+		},
 	}
 
-	downloadURL, err := getDownloadURLFromHeader(context.Background(), testBaseURL, testTerraformGet)
-	require.NoError(t, err)
-	assert.Equal(t, "https://registry.terraform.io/terraform-aws-modules/terraform-aws-vpc", downloadURL)
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			downloadURL, err := getDownloadURLFromHeader(context.Background(), testCase.moduleURL, testCase.terraformGet)
```
denis256, line 99: Make sense to move `t.Parallel()` from function beginning test case function in range loop?

denis256 (approved):

internal/tfr/getter_test.go
yorinasub17: Since this is super fast (in memory), I wanted to avoid the complexity of using `t.Parallel` with the range variable. Introducing `t.Parallel` requires capturing `testCase` into the range variable, which requires some explanation (since it's confusing to see `testCase := testCase` to the uninitiated). I wanted to avoid all that explaining by not using it, since it's not necessary.

Hope that makes sense!

internal/tfr/getter.go
yorinasub17: Ah good point. Removed in 776f8de

yorinasub17: Also noticed that one of the tests could be consolidated into the new table driven test, so did that in the same commit.

denis256 (approved):

yorinasub17:

Issue #2020, opened 2022-02-22
Error: Inconsistent dependency lock file


MFAshby:
Note I'm doing something rather strange, this is almost certainly not a common use-case.

I have a system testing setup, where several (~10) copies of our system are booted to provide a 'pool' of instances that the test runner can use to run tests on. The system comprises several different applications with dependencies on one another. We have a big bash script that boots the whole lot in docker. I'm trying to replace a lot of bash script and docker with terraform / terragrunt configuration instead to make it more scaleable, maintainable, and flexible.

In order to keep the boot time for this setup reasonable, I need to start the different copies (referred to as 'stacks') in parallel. 

In order to achieve parallel boot, I'm running terragrunt multiple times in parallel, once per 'stack'. The state and data directories are kept separate from one another in order that they don't interfere. I have a terragrunt file that looks like this:
```terraform
# Containers isolated per stack, use separate data and state directories per-stack
locals {
  stack_suffix = get_env("TF_VAR_stack_suffix", "")
}
remote_state {
  backend = "local"
  config = {
    path = local.stack_suffix == "" ? ".state/terraform" : "./.${local.stack_suffix}.state/terraform"
  }
  # https://terragrunt.gruntwork.io/docs/reference/config-blocks-and-attributes/#dependency
  # You are not relying on before_hook, after_hook, or extra_arguments to the terraform init call. 
  # NOTE: terragrunt will not automatically detect this and you will need to explicitly opt out of the dependency 
  # optimization flag.
  disable_dependency_optimization = true
} 
terraform {
  extra_arguments "stack_workspace" {
    env_vars = {
      TF_DATA_DIR = local.stack_suffix == "" ? ".terraform" : ".${local.stack_suffix}.terraform"
    }
    commands = [
      "init",
      "apply",
      "refresh",
      "import",
      "plan",
      "taint",
      "untaint",
      "output"
    ]
  }
  # Multiple processes will read this lockfile simultaneously
  # so it's not safe to allow updates
  extra_arguments "lockfile_readonly" {
    commands = [
      "init"
    ]
    arguments = [
      "-lockfile=readonly",
    ]
  }
}
```

This _mostly_ works, however I'm occasionally hitting an error where the terraform dependency lock file appears to be inconsistent
```
16:14:38 Error: Inconsistent dependency lock file
```

[...]

PR #2021, opened 2022-02-22
Don't copy a file if it's the same source and destination.

MFAshby:
Should be safe, as the only effect of reading / writing back to the same file is to update the modified timestamp on the file.

Fixes https://github.com/gruntwork-io/terragrunt/issues/2020, because this file will no longer be modified unless necessary, meaning that multiple processes can read it safely.

### DIFF
--- a/util/file.go
+++ b/util/file.go
@@ -296,6 +296,10 @@ func TerragruntExcludes(path string) bool {
 
 // Copy a file from source to destination
 func CopyFile(source string, destination string) error {
+	if source == destination {
+		return nil
+	}
+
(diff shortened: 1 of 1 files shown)


MFAshby:
Unit tests for this package pass locally.
```
[martin@martin-intelz370 util]$ go test ./...
ok  	github.com/gruntwork-io/terragrunt/util	0.006s
```

denis256:
Suggested to use `terragrunt-working-dir` in https://github.com/gruntwork-io/terragrunt/issues/2020

Issue #2026, opened 2022-02-27
get_path_[to|from]_repo_root fail when specifying relative path for the Terragrunt working directory


andreykaipov:
Regarding the recent functions introduced in https://github.com/gruntwork-io/terragrunt/pull/1954. It's better demonstrated through example. Given the following directory structure:

```
a
├── b
│   ├── main.tf
│   └── terragrunt.hcl
└── terragrunt.hcl
```

With contents `a/terragrunt.hcl`:

```hcl
locals {
  _ = run_cmd("sh", "-c", <<EOF
    echo in parent: ${get_path_to_repo_root()}
    echo in parent: ${get_path_from_repo_root()}
  EOF
  )
}
```

And contents `a/b/terragrunt.hcl`:

```hcl
include "root" {
  path = find_in_parent_folders()
}
```

If we kick off the Terragrunt resolution from within `a`, we'll error:

```console
❯ cd a
❯ TERRAGRUNT_WORKING_DIR=b terragrunt version
...
ERRO[0000] Call to function "get_path_to_repo_root" failed: Rel: can't make /tmp/a relative to b.
...
ERRO[0000] Call to function "get_path_from_repo_root" failed: Rel: can't make b relative to /tmp/a.
```

It's easily solved by just specifying an absolute path for the working dir, but it felt a bit strange to have to:

```console
❯ TERRAGRUNT_WORKING_DIR=$PWD/b terragrunt version
in parent: ../
in parent: b
Terraform v1.1.6
on linux_amd64
```

We could also do something like `(cd b; terragrunt version)` to get it to work. I have no clue how this could be fixed permanently without making assumptions about a user's environment, but I figured I'd get this documented.

denis256:
Released in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.36.3

PR #1954, opened 2021-12-17
add repo path functions

tbell83:
This PR adds two functions, `get_path_to_repo_root()` and `get_path_from_repo_root()`, loosely linked to https://github.com/gruntwork-io/terragrunt/issues/305

Tests:
```shell
terragrunt/test on  add-repo-path-functions via 🐹 v1.17
❯ go test -v -run TestGetPathToRepoRoot
=== RUN   TestGetPathToRepoRoot
=== PAUSE TestGetPathToRepoRoot
=== CONT  TestGetPathToRepoRoot
    integration_test.go:3467: Copying fixture-get-path-to-repo-root to /var/folders/mg/dlfyyxbj4s32mwd8599bh3200000gq/T/terragrunt-test1054949920
runTerragruntVersionCommand after split
[terragrunt apply-all --terragrunt-non-interactive --terragrunt-working-dir /private/var/folders/mg/dlfyyxbj4s32mwd8599bh3200000gq/T/terragrunt-test1054949920/fixture-get-path-to-repo-root]
WARN[0000] 'apply-all' is deprecated. Running 'terragrunt run-all apply' instead. Please update your workflows to use 'terragrunt run-all apply', as 'apply-all' may be removed in the future!
INFO[0000] The stack at /private/var/folders/mg/dlfyyxbj4s32mwd8599bh3200000gq/T/terragrunt-test1054949920/fixture-get-path-to-repo-root will be processed in the following order for command apply:
Group 1
- Module /private/var/folders/mg/dlfyyxbj4s32mwd8599bh3200000gq/T/terragrunt-test1054949920/fixture-get-path-to-repo-root
  prefix=[/private/var/folders/mg/dlfyyxbj4s32mwd8599bh3200000gq/T/terragrunt-test1054949920/fixture-get-path-to-repo-root]

Initializing the backend...

Initializing provider plugins...

Terraform has been successfully initialized!

You may now begin working with Terraform. Try running "terraform plan" to see
any changes that are required for your infrastructure. All Terraform commands
should now work.

If you ever set or change modules or backend configuration for Terraform,
rerun this command to reinitialize your working directory. If you forget, other
commands will detect it and remind you to do so if necessary.

Changes to Outputs:
  + path_to_root = "../../"

You can apply this plan to save these new output values to the Terraform
state, without changing any real infrastructure.

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.

Outputs:

path_to_root = "../../"
runTerragruntVersionCommand after split
[...]

PR #2027, opened 2022-02-27
Add get_repo_root() built-in

andreykaipov:
A common pattern I've found myself repeating across our Terragrunt modules is referring to the repo root instead of boggling my head around relative paths.

For example, in the parent:
```hcl
locals {
  root = run_cmd("--terragrunt-quiet", "git", "rev-parse", "--show-toplevel")
}
```

In the child:
```hcl
include "root" {
  expose = true
  path   = find_in_parent_folders()
}

locals {
  something = "${include.root.locals.root}/some/config.yml"
}
```

While `get_parent_terragrunt_dir()` might be enough if our parent `terragrunt.hcl` is also at the root of the repo, as soon as one wants to nest their Terragrunt configuration inside of an `infra` directory or similar, it won't be enough.

I saw https://github.com/gruntwork-io/terragrunt/pull/1954 was merged in but it was only for relative paths (plus it has a few limitations documented in https://github.com/gruntwork-io/terragrunt/issues/2026). Having a built-in function to get the absolute path equivalent would be a nice to have. The above example would now look like:

In the parent:
```hcl
locals {}
```

In the child:
```hcl
include "root" {
  path   = find_in_parent_folders()
}

locals {
  something = "${get_repo_root()}/some/config.yml"
}
```

Please lemme know if y'all are open to this change! Thanks.

### DIFF
--- a/config/config_helpers.go
+++ b/config/config_helpers.go
@@ -123,6 +123,7 @@ func CreateTerragruntEvalContext(
 		"run_cmd":                                      wrapStringSliceToStringAsFuncImpl(runCommand, extensions.TrackInclude, terragruntOptions),
 		"read_terragrunt_config":                       readTerragruntConfigAsFuncImpl(terragruntOptions),
 		"get_platform":                                 wrapVoidToStringAsFuncImpl(getPlatform, extensions.TrackInclude, terragruntOptions),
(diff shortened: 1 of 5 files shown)


denis256 (approved):

tonerdo (approved):

tonerdo:
Thanks @andreykaipov!

PR #2030, opened 2022-03-03
Built-in functions list cleanup

denis256:
Removed empty list item in "Built-in functions list cleanup" that break spacing:

image


yorinasub17 (approved):
LGTM!

PR #2053, opened 2022-03-30
Update inputs.md typo

rhoboat:
Can someone please check that the language is correct here? Is it true that `TF_VAR_xxx` will override `inputs`, or is it true that `TF_VAR_xxx` will be overridden by `inputs`? The original language could have implied either one, and I want to fix this precisely because it's confusing.


denis256 (approved):

denis256:
Value from `TF_VAR_xxx` override value from `inputs`

```
terragrunt apply
...
3
...
```

```
TF_VAR_some_number=666 terragrunt apply
...
6
...
```

Full example in: https://github.com/denis256/terragrunt-tests/tree/master/var-override

rhoboat:
Thanks Denis! Merging now despite test failure, because that is unrelated.

PR #2066, opened 2022-04-12
Additional known transient error handling for "Client.Timeout exceeded while awaiting headers"

MFAshby:
We've seen this a number of times on our CI servers and some developers have encountered it locally. It passes on retry, so I believe it to be a transient error communicating with terraform's registry servers.


denis256 (approved):
:+1:

Issue #2062, opened 2022-04-07
TERRAGRUNT_TFPATH regression: no longer overrides dependency's terraform_binary


rhoboat:
### Description of the bug
A customer has been struggling (patiently) with a possible regression in our handling of terraform binaries using `TERRAGRUNT_TFPATH`.

They reported that they found this change in our release notes:

> Updated behavior of terraform_binary. Now terragrunt will always prefer the setting in the configuration. This works around the issue where you need mixed terraform binaries in your project, and terragrunt doesn’t know which one to use when fetching dependencies.

It seems very promising that this change might have introduced a regression.

This only seems to have an impact if you are using `dependency` in your `terragrunt.hcl`.

- TG module A - terragrunt.hcl calls out TG module B as dependency
- TG module B

They both set `terraform_binary`. When you invoke `terragrunt plan` on TG module A, the `TERRAGRUNT_TFPATH` overrides  but when the dependency B is evaluated, its `terraform_binary` setting is used, not `TERRAGRUNT_TFPATH`. The override is only at the top level!

### Expected behavior
Both TG module A and B should be affected by setting `TERRAGRUNT_TFPATH`, even though module B is a dependency in module A.

The customer reported that this was the case for them, until they upgraded Terragrunt from `v0.35.5` to `v0.36.6`. They suspect that it changed somewhere in between, probably at https://github.com/gruntwork-io/terragrunt/releases/tag/v0.35.15.

rhoboat:
Some help with reproducing. Start with some terragrunt.hcl, and also would help to use something like `tfenv`, so that you can manage multiple versions of terraform under the hood, and mess with the override.

# terragrunt.hcl
```hcl
terraform_binary             = "/usr/local/Cellar/tfenv/2.2.2/versions/0.15.3/terraform"
terraform_version_constraint = "0.15.3"

terraform {
  source = "./modules/s3-static-website"
}

inputs = {
  website_domain_name = "foo.com"
}
```

Then have another `terragrunt.hcl` which similarly sets `terraform_binary` as above, and add it as a dependency within the above `terragrunt.hcl`.

After that, in the terminal, set the TERRAGRUNT_TFPATH to `terraform`
$ export TERRAGRUNT_TFPATH=terraform
$ echo $TERRAGRUNT_TFPATH
terraform

Check your version of terragrunt as well:
```bash
$ terragrunt --version
terragrunt version v0.35.5
```

Then check to see what version of terraform gets used by terragrunt:
[...]

PR #2077, opened 2022-04-20
Clarify behavior of terragrunt-tfpath override with respect to dependencies

yorinasub17:
Related to #2062 

#2062 highlighted confusion about what is the desired and implemented behavior of the `TERRAGRUNT_TFPATH` env var, so this PR updates the docs to clarify what should be happening.


rhoboat (approved):

yorinasub17:
Thanks for review! Merging now without waiting for tests given that this is an update to the docs.

PR #2094, opened 2022-05-02
[RFC] custom_command block

tim775:
This RFC introduces a `custom_command` block which builds off the work done for `before/after_hook` to further improve support for user scripts and third party tools.

Here is [a link to the rendered markdown](https://github.com/infracost/terragrunt/blob/5dd6ab9c58cc18b15afe3cac80653ffbc137dd11/docs/_docs/05_rfc/custom_commands.md).

### DIFF
--- a/docs/_docs/05_rfc/custom_commands.md
+++ b/docs/_docs/05_rfc/custom_commands.md
@@ -1,0 +1,125 @@
+---
+layout: collection-browser-doc
+title: Custom Commands
+category: RFC
+categories_url: rfc
+excerpt: Ability to define custom commands
+tags: ["rfc", "contributing", "community"]
+order: 505
+nav_title: Documentation
+nav_title_link: /docs/
+---
+
+# Custom Commands
+
+**STATUS**: In proposal
+
+## Background
+
+Third party tools are available that perform static analysis of Terraform configurations to provide
+additional functionality such as cost estimation or security scanning.  Terragrunt currently supports execution
+of these tools in concert with a Terraform command using `before_hook` and `after_hook`, but there is no way to run commands without also running a Terraform command.
+
+## Proposed solution
+
+In order to allow seamless execution of third party tools, we propose adding a `custom_command` property to the `terraform` block.  Users could define a custom command that would execute similarly to a `before_hook` without requiring
+a Terraform command to be run.  The documentation entry would be:
+
+```
+ custom_command (block): Nested blocks used to specify custom
+  commands that can be run against terraform stacks.  Like
+  `before_hook`, custom commands run from the directory with
+  the terraform module.  Supports the following arguments:
+
+    * execute (required) : A list of command and arguments that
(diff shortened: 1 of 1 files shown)


yorinasub17 (commented):
Thanks for proposing this! On principle, I think it makes sense: there is a need to run commands in the terragrunt working dir (where the terraform code is defined).

However, I am not so sure about encoding this in the `terragrunt.hcl` config itself. My gut feeling suggests that it will be easier to implement, and much more flexible if this was implemented as a custom `terragrunt` command, e.g.:

```
terragrunt run-command --terragrunt-all -- infracost breakdown --path .
```

This has some disadvantages, but feels more intuitive and inline with terragrunt's current CLI interface. That is, you don't need to define an `apply` block to run `terragrunt apply` or `terragrunt run-all apply`.

What do you think?

---

I also have some additional questions:

- Is the command different for each module, or largely the same? It makes sense to encode it in the config if the commands are going to be different, but if they are largely the same, then I think the CLI interface is better.

- What happens if you call `terragrunt run-all infracost` and some of the `terragrunt` configs are missing the `custom_command` block for it?

- Are these commands conducive to `run-all` and interleaved outputs? E.g., there is a known issue right now where the `terragrunt run-all plan` output is really hard to read because `terragrunt` runs `plan` concurrently across all the modules.

tim775:
>  I am not so sure about encoding this in the terragrunt.hcl config itself. My gut feeling suggests that it will be easier to implement, and much more flexible if this was implemented as a custom terragrunt command,

That is an interesting idea and it does seem easier to just pass the command you want to run instead of defining it in the config.  I guess one advantage of the config file approach is that you would have access to the built-in fuctions, but I'm not sure how important that is.  

Aesthetically, I think it would be nicer to support this in `run-all` rather than with a `--terragrunt-all`.  So maybe something like:
```
terragrunt exec -- infracost breakdown --path .
terragrunt run-all exec -- infracost breakdown --path .
``` 
---
Regarding the additional questions, I would love to hear what the larger community thinks.

For the Infracost case the answer is that the commands are pretty much the same but we would use a follow up command to combine the individual outputs after everything completes.  

I was thinking we could write to a tmp files then use a glob to generate a combined output, something like:
```
$ terragrunt run-command --terragrunt-all -- infracost breakdown --path . --out-file $(mktemp /tmp/infratemp/out.json.XXXXX)
$ infracost output --path /tmp/infratemp/out.json.*
```

yorinasub17:
> Aesthetically, I think it would be nicer to support this in run-all rather than with a --terragrunt-all. So maybe something like:

Yea `run-all` would be good if we can make it work. I didn't suggest it at first though because `run-all` only supports `terraform` commands (e.g., `terragrunt run-all hclfmt` doesn't work) so was afraid of the complexities involved there. In any case, if it can be done in a clean way, I think that is the way to go!

tim775:
Closing since there doesn't seem to be much interest from the community.

Issue #2045, opened 2022-03-18
How to set error_hook expression string for resource exists issue


Simonzhaohui:
Hello,

I'm trying to use error_hook to catch resource exist issue. If I use `.*` in on_errors list, the error_hook can be triggered. But when I use other regex string, the error_hook can't be triggered, example `.*azurerm_resource_group.*`, `".*subscriptions.*resourceGroups.*test.*"`, etc.

How to set error_hook regex expression statement to catch resource exists issue ?

terragrunt.hcl:
```
terraform {
  source = ".//resource-group"

  error_hook "error_hook_1" {
    commands = ["apply"]
    execute  = ["echo", "resource group is exists"]
    on_errors = [
      ".*azurerm_resource_group.*", # how to set regex statement 
      ".*subscriptions.*resourceGroups.*test.*"
    ]
  }
}

inputs = {
  location = "Southeast Asia"
  name = "test"
}

```
main.tf:
```
provider "azurerm" {
  features {}
}
resource "azurerm_resource_group" "rg" {
  location = var.location
  name     = var.name
}
variable "location" {
  description = "Resource Group Location"
  type        = string
  default     = "Southeast Asia"
}

variable "name" {
  description = "Resource Group Name"
  type        = string
}
```
terragrunt apply command output:
```
terragrunt apply -auto-approve

Terraform used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # azurerm_resource_group.rg will be created
  + resource "azurerm_resource_group" "rg" {
      + id       = (known after apply)
      + location = "southeastasia"
      + name     = "test"
    }

Plan: 1 to add, 0 to change, 0 to destroy.
azurerm_resource_group.rg: Creating...
╷
│ Error: A resource with the ID "/subscriptions/.../resourceGroups/test" already exists - to be managed via Terraform this resource needs to be imported into the State. Please see the resource documentation for "azurerm_resource_group" for more information.
│
│   with azurerm_resource_group.rg,
│   on main.tf line 4, in resource "azurerm_resource_group" "rg":
│    4: resource "azurerm_resource_group" "rg" {
│
╵
ERRO[0035] 1 error occurred:
	* exit status 1
```

terragrunt version: v0.36.0
terraform version: v1.1.5 

**NOTES:**
 It seems the error_hook expression only matches ERROR messages `"1 error occurred: \n* exit status 1"`, not out Stderr messages.

[...]

PR #2101, opened 2022-05-11
#2045 Fixed pattern matching for error hooks 

denis256:
Included changes:
 * Updated handling of terraform execution to return error with execution outputs which later is used in pattern matching for error hooks
 * Updated integration tests to track that errors generated during the execution of terraform are handled by error hooks

Closes: https://github.com/gruntwork-io/terragrunt/issues/2045


yorinasub17 (approved):
LGTM!

PR #2105, opened 2022-05-12
Allow v0.36 to define disable_bucket_update to make it easier to transition to v0.37

yorinasub17:
__IMPORTANT: Merges into v036-dev, which is a snapshot of v0.36.11__

This PR is intended to be a patch release for `v0.36.x` that is intended to smooth the upgrade process from v0.36 to v0.37. Right now, users can't prepare for the v0.37 update by preconfiguring their terragrunt project with `disable_bucket_update = true`, because `terragrunt` 0.36.11 doesn't support this config (terragrunt crashes if that is set).

In this PR, we add the config variable but do nothing with it so that users can setup for the v0.37 update.

When this is merged to `v036-dev`, I'll go ahead and cut a patch release `v0.36.12` pointing to that branch, and then will delete the branches.


denis256 (approved):

yorinasub17:
Thanks for review! Will merge this in, and then cut 0.36.12

Issue #2104, opened 2022-05-12
Failure to parse S3 backend policy document


rickys-splunk:
#2063 seems to introduce the idea that terragrunt _must_ have management access to the configured backend s3 bucket. This suddenly breaks environments that generate plans with little-more-than-read-only roles and applies with elevated access.

Steps to reproduce:
- Manually create a bucket (or use a bucket that already exists)
- Create a role that has read-only access to that bucket
- Use that bucket for the s3 backend
- Try to run a `terragrunt plan`

I believe when the read only role cannot read/modify the S3 bucket, the s3 api does not return json, so the following error is observed:
```
time=2022-05-12T18:14:05Z level=debug msg=Initializing remote state for the s3 backend prefix=[REDACTED] 
time=2022-05-12T18:14:05Z level=debug msg=Checking if SSE is enabled for AWS S3 bucket REDACTED prefix=[REDACTED] 
time=2022-05-12T18:14:05Z level=debug msg=Checking if bucket REDACTED is have root access prefix=[REDACTED] 
time=2022-05-12T18:14:05Z level=error msg=json: cannot unmarshal string into Go struct field Statement.Statement.Resource of type []string
```

Since it's very common to have premade -- or widely shared buckets -- managing terraform state files, perhaps terragrunt should inform the user of configuration errors -- when available -- but not assume that it is in control of the backend bucket.

tjstansell:
The only issue is that `disable_bucket_update` doesn't seem to be a config option we can specify _before_ this breaking change ... so it has to be coordinated with the update to the terragrunt version.  We use a central docker image that has terragrunt built in for all of our projects ... so that coordination is kinda painful.  It would be nice if there was a more backward-compatible way to support this, even if there's an extra `v0.36.x` release that simply allows you to add the `disable_bucket_update` config option ahead of time ... so we could add that everywhere first, then update terragrunt in our docker image.

yorinasub17:
> It would be nice if there was a more backward-compatible way to support this, even if there's an extra v0.36.x release that simply allows you to add the disable_bucket_update config option ahead of time

This makes sense to me. I'll quickly hack an increment patch that will introduce this flag which does nothing for that version, but the presence should make upgrade life easier.

yorinasub17:
> Since it's very common to have premade -- or widely shared buckets -- managing terraform state files, perhaps terragrunt should inform the user of configuration errors -- when available -- but not assume that it is in control of the backend bucket.

[...]

PR #2106, opened 2022-05-13
Update remote state docs to recommend generate first

yorinasub17:
The comment in https://github.com/gruntwork-io/terragrunt/issues/2104 made me realize that our remote state documentation is a bit misleading. We default to recommending the `remote_state` block for setting up your remote state configuration, but in reality, that may not be the best option for all use cases.

For example, one could use `generate` blocks instead to manage the backend config, which works well just like `remote_state`; the main difference being that `terragrunt` does not try to manage the remote state objects when using `generate`, whereas the design of `remote_state` blocks is to have `terragrunt` manage your state buckets.

For this reason, this PR proposes to revamp the remote state docs to instead recommend:

- Using `generate` blocks to generate a `terraform` block containing the `backend` subblock.
- Recommend using `remote_state` block if users want to additionally have `terragrunt` manage the state resources, like automatically creating and updating them.

### DIFF
--- a/docs/_docs/02_features/keep-your-remote-state-configuration-dry.md
+++ b/docs/_docs/02_features/keep-your-remote-state-configuration-dry.md
@@ -16,9 +16,8 @@ nav_title_link: /docs/
 
   - [Filling in remote state settings with Terragrunt](#filling-in-remote-state-settings-with-terragrunt)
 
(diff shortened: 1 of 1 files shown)


denis256 (approved):

yorinasub17:

Issue #2113, opened 2022-05-18
new terragrunt feature to update remote states bucket raises an error


ebarault:
Hi,

We just upgraded to v0.37.1, and we have the following error :

```
DEBU[0002] Initializing remote state for the s3 backend  prefix=[xxxxxxxxxx] 
DEBU[0002] Checking if SSE is enabled for AWS S3 xxxxxxxxxxx  prefix=[xxxxxxxxxxx] 
DEBU[0003] Checking if bucket tfstates-gfx-infra-957585145944 is have root access  prefix=[xxxxxxxxxx] 
DEBU[0003] *json.UnmarshalTypeError json: cannot unmarshal array into Go struct field Statement.Statement.Action of type string
github.com/gruntwork-io/terragrunt/remote/remote_state_s3.go:861 (0x105245b21)
github.com/gruntwork-io/terragrunt/remote/remote_state_s3.go:522 (0x105243310)
github.com/gruntwork-io/terragrunt/remote/remote_state_s3.go:398 (0x105242aec)
github.com/gruntwork-io/terragrunt/remote/remote_state_s3.go:265 (0x105242150)
github.com/gruntwork-io/terragrunt/remote/remote_state.go:70 (0x10523e0a4)
github.com/gruntwork-io/terragrunt/cli/cli_app.go:884 (0x105449f20)
github.com/gruntwork-io/terragrunt/cli/cli_app.go:712 (0x105449038)
github.com/gruntwork-io/terragrunt/cli/cli_app.go:1019 (0x10544a914)
github.com/gruntwork-io/terragrunt/cli/cli_app.go:957 (0x10544a560)
github.com/gruntwork-io/terragrunt/cli/cli_app.go:716 (0x105449058)
github.com/gruntwork-io/terragrunt/cli/cli_app.go:516 (0x10544807c)
github.com/gruntwork-io/terragrunt/cli/cli_app.go:352 (0x105447810)
github.com/gruntwork-io/terragrunt/cli/cli_app.go:323 (0x105447434)
github.com/urfave/cli@v1.22.3/app.go:526 (0x10541bcac)
github.com/urfave/cli@v1.22.3/app.go:288 (0x105419a88)
github.com/gruntwork-io/terragrunt/main.go:25 (0x1054531a0)
runtime/proc.go:250 (0x104a12830)
runtime/asm_arm64.s:1263 (0x104a40db4) 
[...]

PR #2118, opened 2022-05-20
#2113 Improved parsing of AWS policy response

denis256:
Updated parsing of AWS policy statement to handle cases when fields "Action" and "Policy" are returned as arrays instead of strings

Closes: https://github.com/gruntwork-io/terragrunt/issues/2113

AWS reference:
https://docs.aws.amazon.com/IAM//latest/UserGuide/reference_policies_elements_action.html
https://docs.aws.amazon.com/IAM//latest/UserGuide/reference_policies_elements_resource.html

### DIFF
--- a/aws_helper/policy.go
+++ b/aws_helper/policy.go
@@ -2,18 +2,22 @@ package aws_helper
 
 import "encoding/json"
 
-// A representation of the polciy for AWS
+// Policy - representation of the policy for AWS
 type Policy struct {
 	Version   string      `json:"Version"`
 	Statement []Statement `json:"Statement"`
 }
 
+// Statement - AWS policy statement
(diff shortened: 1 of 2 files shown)


dancorne:
We're experiencing this issue with one of our AWS accounts as well, and I can confirm the linked PR solves it.

And a workaround is making the existing bucket policy use a list in the Resource if it's a string (this was the case in some of our RootAccess rules) eg. change the first line to the second
```
"Resource": "arn:aws:s3:::$BUCKET_NAME"

"Resource": ["arn:aws:s3:::$BUCKET_NAME", "arn:aws:s3:::$BUCKET_NAME/*"]
```

yorinasub17 (approved):
LGTM

Issue #2119, opened 2022-05-22
fetch dependency output from state directly


Ido-DY:
I have a few components that have many dependencies, when I run `terragrunt plan` it could take a few minutes to run.
I ran the plan with `--terragrunt-log-level debug` then I saw that for each dependency, Terragrunt is executing `terraform init`and `terraform output -json`.

I compared the output of `terraform output -json` with the state stored in my S3 bucket and it was just the same, then I made an experiment: for all of our components, I configured Terrraform to skip `init` and return the output from the bucket directly when called on a dependency with `output -json` by Terragrunt.

My experiment shows incredible results by performing the `plan` action up to 10X time faster while preserving just the exact result.

I saw in the code that the dependency output is fetched from the remote state by executing `terraform init && terraform output -json` : https://github.com/gruntwork-io/terragrunt/blob/020ad12980358b8fabb8b9d778e5bc15dd3b211f/config/dependency.go#L522

Now I wonder about a few things:

1. Have you considered fetching the output of dependencies directly from the state files (by an API call to the backend), instead of `terraform init && terraform output -json` ?
2. Will you accept such contribution ?
3. Do you see any risk with this approach ?

yorinasub17:
> by an API call to the backend
> Do you see any risk with this approach?

While I agree that this will most likely be the fastest approach, I worry that this may be brittle. Off the top of my head, I can see the following risks with this approach:

- What will we do if the project isn't using S3? E.g., the backend is in Terraform Cloud, or Azure?
- How do we keep track of all the version formats? E.g., what happens when Terraform updates the internal representation of the state object? When do we deprecate support for a particular state representation?
- Should we handle state locking? E.g., `terraform output` gets a read lock on the state file so that it doesn't try to read it out when an `apply` is actively happening on it. Should we mirror that behavior?

With that said, I think this is probably a good addition to have given that terraform output reading tends to be a looming performance concern. I think I would be open to a contribution that makes this, provided that:

- It is opt-in (that is, we don't default to this mechanism of extracting outputs)
- The state extracting logic is encapsulated in a way that we can extend this to GCS buckets and other backends in the future.
- There is a good, maintainable way to handle changing state representations.

lorengordon:
Does terragrunt cache the output of dependencies in any given run, or is it running init and output multiple times for the same dependency? Maybe a cache of the outputs for a given dependency would be a big improvement?

yorinasub17:
[...]

PR #2123, opened 2022-05-25
feat: fetch dependency output from state

Ido-DY:
In following to #2119


denis256:
Hi,
an integration test to track that this feature continue to work will be helpful, examples how to do it are in [integration_test.go](https://github.com/gruntwork-io/terragrunt/blob/master/test/integration_test.go#L716)

yorinasub17:
Thanks for the contribution! Will merge and release this now!

PR #2146, opened 2022-06-15
Fix backend generation when access and secret keys are supplied

jaysonsantos:
## Description

This should fix a bug that happens when you provide only access and secret keys and terragrunt tries to initialize the bucket, and it fails because it always fallbacks to env/config file from ~/.aws

## TODOs

- [x] Update the docs. (nothing changes)

## Release Notes (draft)

Allow terragrunt to be initialized with s3 backend using access and secret keys.

### Migration Guide

It should be backwards compatible

### DIFF
--- a/aws_helper/config.go
+++ b/aws_helper/config.go
@@ -27,6 +27,8 @@ type AwsSessionConfig struct {
 	DisableComputeChecksums bool
 	ExternalID              string
 	SessionName             string
+	AccessKey               *string
+	SecretKey               *string
 }
 
 // Returns an AWS session object for the given config region (required), profile name (optional), and IAM role to assume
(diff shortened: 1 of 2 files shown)


denis256:
Hi,
I'm not 100% sure if its a good idea to set different values for `access_key / secret_key` on the remote state block and have it in the source code since it may lead to leak of secrets in git and have less portable modules

If the approach is useful, then it should be documented in [config-blocks-and-attributes.md](https://github.com/gruntwork-io/terragrunt/blob/master/docs/_docs/04_reference/config-blocks-and-attributes.md) and also added tests that will validate if new values are still in use

denis256:
Hi,
I'm not 100% sure if its a good idea to set different values for `access_key / secret_key` on the remote state block and have it in the source code since it may lead to leak of secrets in git and have less portable modules

If the approach is useful, then it should be documented in [config-blocks-and-attributes.md](https://github.com/gruntwork-io/terragrunt/blob/master/docs/_docs/04_reference/config-blocks-and-attributes.md) and also added tests that will validate if new values are still in use

yhakbar:
Hey @jaysonsantos 

We're going through stale PRs and closing them out. If you would like to propose this functionality again in the future, please submit another pull request.

If possible, submit an issue first in order to make sure that the problem is well understood, and try to prioritize adding good documentation and testing with your feature so that we are confident in our review.

Thanks!

PR #2168, opened 2022-06-28
Use faster hashing approach when handling local module sources

### LINKED ISSUE #2154: Terragrunt hangs at 100% CPU for 5 minutes
Hi,

Recently (after upgrating both terraform & terragrunt to latest),  `terragrunt apply` hangs for 5 minutes with 100% CPU.

When doing htop, I've noticed the subprocesses:
1. `terraform init` that takes some seconds
2. `terraform outputs` that takes some seconds too
3. `terragrunt apply` hanging for some long minutes. Each subprocess taking 100% CPU one after another for 4 minutes.

I've tried to add `--terragrunt-debug` but it only generated the `terragrunt-debug.tfvars.json` after the 5 minutes and it's some super helpful.

What can I do?

Thanks!

Sinjo:
The current approach to hashing (sha256 of the file contents of each
module) is very slow on large module trees due to a combination of doing
lots of File I/O and spending lots of CPU time on running sha256
calculations.

This manifests as Terragrunt spending 5+ minutes at 100% CPU at the
beginning of every run, with no output.

This commit takes a faster, if slightly less perfect, approach to
determining if a local module has had its files modified. Rather than
hashing the entire contents of each file, it takes the `mtime` of the
file (obtained through a cheap `stat` call), appends that to the file
name, and hashes that combined string.

The downside of this approach is that if a file has been written to
since we last copied the module, we'll copy it again. Thankfully, that
is exceedingly unlikely to happen in the middle of a Terragrunt run.

---

Fixes #2154.

I've tested this by running `terragrunt plan` on some of our stacks and it seems to do the trick. I no longer get a really long startup phase where Terragrunt is at 100% CPU.

The one thing that's still on my mind is that I'm slightly confused by the original PR that added this. On my machine `terragrunt plan` takes pretty much the same amount of time with or without the optimisation to avoid copying source directories multiple times. I can only assume that there are other setups (maybe different hardware, or a different pattern of modules/stacks to the one we have) where that's not the case, but I can't reproduce the slowness when I get rid of the caching completely (by running pre-0.37.3 versions of Terragrunt).

That said, if you're 👍🏻 on this faster hashing, it definitely fixes the 100% CPU for 5 minutes problem for me, and I'm happy to run with it.


cli/tfsource/types.go
```diff
@@ -47,8 +48,31 @@ func (src *TerraformSource) String() string {
...
+
+			fileModified := info.ModTime().UnixMicro()
+			hashContents := fmt.Sprintf("%s:%d", path, fileModified)
+
+			sourceHash.Write([]byte(hashContents))
+
+			return nil
+		})
+
 		if err == nil {
```
denis256, line 73: I think will be helpful to log error, this will help to debug any possible access error

lorengordon: Based on #2171, yeah, it would be helpful to log a warning when there's an error hashing the directory contents...

Sinjo: @denis256 Done!

Sinjo: Right, I've fixed the logging code, so it should be good to go in its current form.

I do agree with you about bypassing the cache if we run into an error while validating it though. I'm going to see how tricky that would be to introduce.

lorengordon: Well you could just hash a random string, yeah? :p That would make it think it changed, and force it to update?

Sinjo: Okay, I've had a look around the codebase to see how this function gets used. We could probably bypass the cache if `EncodeSourceVersion` were to return an error, but it's not totally straightforward at all the callsites.

  - [This case](https://github.com/gruntwork-io/terragrunt/blob/7d143740c05ef9a5cc87414c5b22107bf3cb1425/cli/download_source.go#L112) and [this case](https://github.com/gruntwork-io/terragrunt/blob/7d143740c05ef9a5cc87414c5b22107bf3cb1425/cli/download_source.go#L147) seem reasonably easy to handle. We could have them assume the download folder didn't have an up-to-date copy of the source and copy it over again.
  - [This case](https://github.com/gruntwork-io/terragrunt/blob/7d143740c05ef9a5cc87414c5b22107bf3cb1425/cli/tfsource/types.go#L62) is slightly more painful. Returning an error would cause Terragrunt to exit without doing anything as the error would be propagated all the way back to `cli_app.go`. We could do something weird like write a random value into the file there, effectively signalling "we don't know what the hash was, so you're gonna have to copy this module again next time", but it feels somewhat janky.

The other thing that would need taking care of is some testing. I haven't looked at the test suite/testing libraries in use to see if we'd have an easy way of stubbing `EncodeSourceVersion` to return an error. I'm sure it's doable, but I know stubbing/mocking can be a bit of a pain in Go if you don't have an interface to act as a testing seam.

denis256: Ok, I think it should work fine with log + random hash

lorengordon: Also, if on error it's going to force-update the cache, which requires less from the user to "just work", then I think logging at the debug level is fine. I suggested a warning previously just because of what we were seeing in #2171, where a hash error may require user interaction to clean up their working dir, and so the visibility/noise of a warning was worth it.

Sinjo: Cool - I'll add that in.

I've got the day off tomorrow, so I'm not sure I'll look at this until Monday, but I'll add an extra commit to this PR for the new hashing behaviour then.

Sinjo: Cool - I'll add that in.

I've got the day off tomorrow, so I'm not sure I'll look at this until Monday, but I'll add an extra commit to this PR for the new hashing behaviour then.

Sinjo: @denis256 Alright, that's in there now as a second commit!

Testing the error case was a little tricky. Best I could come up with was making the containing directory have `0000` permissions so that the `stat` call failed and put us into our error handling code.

cli/download_source.go
```diff
@@ -144,7 +146,15 @@ func alreadyHaveLatestCode(terraformSource *tfsource.TerraformSource, terragrunt
 		return false, nil
 	}
 
-	currentVersion := terraformSource.EncodeSourceVersion()
+	currentVersion, err := terraformSource.EncodeSourceVersion()
+	// If we fail to calculate the source version (e.g. because walking the
+	// directory tree failed) use a random version instead, bypassing the cache.
+	if err != nil {
+		randomBytes := make([]byte, 32)
+		rand.Read(randomBytes)
```
denis256, line 154: Returned error not checked

cli/tfsource/types.go
```diff
@@ -45,21 +49,57 @@ func (src *TerraformSource) String() string {
...
-			return hash
+			hash := fmt.Sprintf("%x", sourceHash.Sum(nil))
+
+			return hash, nil
+		} else {
```
denis256, line 80: `else` I think is not required since previous block ends with `return`

cli/tfsource/types.go
```diff
@@ -45,21 +49,57 @@ func (src *TerraformSource) String() string {
...
 func (terraformSource TerraformSource) WriteVersionFile() error {
-	version := terraformSource.EncodeSourceVersion()
+	version, err := terraformSource.EncodeSourceVersion()
+	if err != nil {
+		// If we failed to calculate a SHA of the downloaded source, write a SHA of
+		// some random data into the version file.
+		//
+		// This ensures we attempt to redownload the source next time.
+		randomBytes := make([]byte, 32)
+		rand.Read(randomBytes)
```
denis256, line 99: Maybe make sense to extract as separate function and re-use it in `alreadyHaveLatestCode()`  and `WriteVersionFile()`

Sinjo:
@denis256 I've made all those changes. Can't believe I missed that `rand.Read` can fail 🤦🏻‍♂️ 

I also figured out a stronger test for the caching behaviour that includes writing a version file and checking we don't copy the local source again afterwards, and I've pushed it in a separate commit.

denis256 (approved):

Issue #2189, opened 2022-07-12
Bugs in render-json command?


brikis98:
It seems like the `render-json` command has a couple bugs:

1. It doesn't output any info about `dependency` blocks.
2. The `locals` map is `null` even when local vars are set.

I created a small repo to test with here: https://github.com/brikis98/render-json-bug-repro. To reproduce the issue, run `terragrunt render-json` in the root of that repo.

The output I get looks like this:

```json
{
  "dependencies": null,
  "download_dir": "",
  "generate": {
    "provider": {
      "comment_prefix": "# ",
      "contents": "# This is just a test",
      "disable_signature": false,
      "if_exists": "overwrite",
      "path": "provider.tf"
    }
  },
  "iam_assume_role_duration": null,
  "iam_assume_role_session_name": "",
  "iam_role": "",
  "inputs": {
    "another": "baz",
    "baz": "blah",
    "foo": "bar",
    "from_root": "Hi"
  },
  "locals": null,
  "retry_max_attempts": null,
  "retry_sleep_interval_sec": null,
  "retryable_errors": null,
  "skip": false,
  "terraform": {
    "after_hook": {},
    "before_hook": {},
    "error_hook": {},
    "extra_arguments": {},
    "include_in_copy": null,
    "source": "./foo"
  },
  "terraform_binary": "",
  "terraform_version_constraint": "",
  "terragrunt_version_constraint": ""
}
```

Even though the module has a `dependency` block and local variables defined, I don't see either in the output.

yorinasub17:
Fix released in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.38.5

lachlankrautz:
That release seems to introduce this issue not solve it. I'm getting `{"locals": null}` on `v0.38.5` and wasn't before..

yorinasub17:
@lachlankrautz do you have a minimum reproducible case we can use to debug the issue you are seeing? The latest version outputs the `locals` properly in the referenced repo from OP, and also in a few examples I have.

lachlankrautz:
Steps to reproduce just using the terragrunt-infrastructure-live-example

```shell
git clone git@github.com:gruntwork-io/terragrunt-infrastructure-live-example.git
cd terragrunt-infrastructure-live-example/prod/us-east-1/prod/mysql
terragrunt render-json --terragrunt-json-out /dev/stdout | jq -r '.locals'
```

Output
```shell
null
````

Output previous to this version:
````shell
[...]

PR #2191, opened 2022-07-12
Fix bug in render-json with dependency and locals

yorinasub17:
## Description

Fixes https://github.com/gruntwork-io/terragrunt/issues/2189

This fixes a bug in `render-json` where the `dependency` blocks were not always rendered out, as well as with `locals`. This happens when there are `include` blocks and `dependency` is set on the child. The reason for each is slightly different, although the core is the same - we were intentionally not merging these blocks in the merge operation.

### Why don't we merge locals?

We don't merge locals intentionally because we don't want to confuse the parsing engine and introduce a bug where Locals become Globals. Locals are intentionally meant to be a local variable that is only in scope for the current config. Thus, if we start merging them with `include`, then they can pollute the included config and become globals instead of locals.

However, because we don't merge them in the include, the Locals on the top level child config gets lost in the `render-json` command because we always returned the merged config, and we don't merge up the locals.

This is fixed in this PR by setting the `Locals` attribute on the merged config to the value of the top level config.

### Why don't we merge TerragruntDependencies (dependency blocks)?

Dependencies need to be available during the parsing step, and not after the full config is rendered. Because we handle `include` blocks at the end, we were deliberately ignoring dependency blocks because they were already merged before we get to the step.

However, because we don't merge through the dependency blocks, the attribute `TerragruntDependencies` wasn't accurate on the parsed config because we weren't updating this attribute when we actually merge the dependencies together.

Therefore, to fix this, in this PR I added the merge routine for `TerragruntDependencies` to the include handler.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Fixed a bug in `render-json` where `dependency` blocks and `locals` information were lost in the resulting rendered config.


denis256 (approved):

test/integration_debug_test.go
```diff
@@ -255,6 +277,99 @@ func TestRenderJSONConfig(t *testing.T) {
 	}
 }
 
+func TestRenderJSONConfigWithIncludesDependenciesAndLocals(t *testing.T) {
```
brikis98, line 280: Thanks for capturing that repro case in this test 👍

brikis98 (approved):
Thank you for the super quick turnaround @yorinasub17!

PR #2199, opened 2022-07-15
Add metadata to render-json output

### LINKED ISSUE #2196: Add support for outputting "metadata" in the render-json command
Something that would make `render-json` even more powerful for debugging is if the output included not only the fully-resolved values, but also other metadata, such as "breadcrumbs" that tell you where that value was found. This likely means changing the JSON output format, so we'd probably only want this with some flag like `--with-metadata`. 

For example, let's say this is the original `render-json` output (truncated to keep the example simple):

```json
{
  "generate": {
    "provider": {
      "comment_prefix": "# ",
      "contents": "# This is just a test",
      "disable_signature": false,
      "if_exists": "overwrite",
      "path": "provider.tf"
    }
  },
  "inputs": {
    "foo": "bar",
    "baz": "blah"
  },
  "terraform": {
    "source": "./foo"
  }
}
```

Here's what `render-json --with-metadata` might look like:

```json
{
  "generate": {
    "provider": {
      "value": {
        "comment_prefix": "# ",
        "contents": "# This is just a test",
        "disable_signature": false,
        "if_exists": "overwrite",
        "path": "provider.tf",
      },
      "metadata": {
        "found_in_file": "/live/dev/vpc/terragrunt.hcl"
      }
    }
  },
  "inputs": {
    "foo": {
      "value": "bar",
      "metadata": {
        "found_in_file": "/live/dev/vpc/terragrunt.hcl"
      }
    "baz":  {
      "value": "blah",
      "metadata": {
        "found_in_file": "/live/terragrunt.hcl"
      }
  },
  "terraform": {
    "source": {
      "value": "./foo",
      "metadata": {
        "found_in_file": "/live/_envcommon/vpc.hcl"
      }
    }
  }
}
```

In other words, every value because an object with the keys: 

* `value`: The original value goes here.
* `metadata`: Extra metadata goes in this object. Initially, the most useful metadata would be `found_in_file`, which gives you the file path where this value came from: you can see in the examples above, that may be a "child" config (`/live/dev/vpc/terragrunt.hcl`) or a parent config pulled in via `include` (`/live/terragrunt.hcl`, `/live/_envcommon/vpc.hcl`). In the future, we may also want to add metadata such as the name of the `include` that was used to pull in that file, the type of merging done, etc.

denis256:
## Description

Update `render-json` command to have argument `--with-metadata` that will include metadata for generated json

Fixes #2196.

## TODOs

- [x] Update the docs.

## Release Notes (draft)


brikis98 (commented):
Looks like a really good start to me! 👍

config/config.go
```diff
@@ -1015,6 +1047,70 @@ func configFileHasDependencyBlock(configPath string, terragruntOptions *options.
...
+	MetadataIamRole                     = "iam_role"
+	MetadataIamAssumeRoleDuration       = "iam_assume_role_duration"
+	MetadataIamAssumeRoleSessionName    = "iam_assume_role_session_name"
+	MetadataInputs                      = "inputs"
+	MetadataLocals                      = "locals"
+	MetadataGenerateConfigs             = "generate"
+	MetadataRetryableErrors             = "retryable_errors"
+	MetadataRetryMaxAttempts            = "retry_max_attempts"
+	MetadataRetrySleepIntervalSec       = "retry_sleep_interval_sec"
+)
```
brikis98, line 1070: Is there any way to avoid having to maintain this list by hand? It feels very redundant with the config definition and parsing, which already has too much duplication in it, and I'm worried that when we go to update one, we'll forget to update the other.

config/config.go
```diff
@@ -1015,6 +1047,70 @@ func configFileHasDependencyBlock(configPath string, terragruntOptions *options.
...
+
+// AddFieldMetadata persist "generic" metadata
+func (conf *TerragruntConfig) MetadataWithType(fieldType, fieldName string, m map[string]interface{}) {
```
brikis98, line 1073: NIT: The name `MetadataWithType` makes me think of _type constraints_: e.g., string, int, bool, etc. I don't think that's what's happening here... Is this more of there being a _fieldName_ (the name of a field, which is a constant) and _fieldId_ (a unique ID)? And so the method signature is more like `SetMetadataWithId(fieldName, fieldId string, ...)`.

config/config.go
```diff
@@ -1015,6 +1047,70 @@ func configFileHasDependencyBlock(configPath string, terragruntOptions *options.
...
+	if !found {
+		metadata = make(map[string]interface{})
+	}
+	for key, value := range m {
+		metadata[key] = value
+	}
+	conf.FieldsMetadata[field] = metadata
+}
+
+func (conf *TerragruntConfig) Metadata(fieldName string, m map[string]interface{}) {
```
brikis98, line 1090: NIT: I think it would be slightly clearer to name these `MetadataXXX` methods something like `SetMetadataXXX`. That way, you know it's a method to mutate the underlying object. Or perhaps even `SetFieldMetadata` since you have a `GetFieldMetadata` method.

config/config.go
```diff
@@ -1015,6 +1047,70 @@ func configFileHasDependencyBlock(configPath string, terragruntOptions *options.
...
+}
+
+func (conf *TerragruntConfig) GetMapFieldMetadata(fieldType, fieldName string) (map[string]interface{}, bool) {
+	if conf.FieldsMetadata == nil {
+		return nil, false
+	}
+	field := fmt.Sprintf("%s-%s", fieldType, fieldName)
+
+	value, found := conf.FieldsMetadata[field]
+	return value, found
```
brikis98, line 1105: NIT:
```suggestion
	return conf.FieldsMetadata[field]
```

config/config_partial.go
```diff
@@ -252,14 +259,26 @@ func PartialParseConfigString(
 				return nil, err
 			}
 			output.TerragruntDependencies = decoded.Dependencies
+			for _, item := range output.TerragruntDependencies {
+				output.MetadataWithType(MetadataDependency, item.Name, defaultMetadata)
+			}
+			//TODO: pass terragrunt dependencies
```
brikis98, line 265: ?

config/include.go
```diff
@@ -308,6 +308,7 @@ func (targetConfig *TerragruntConfig) Merge(sourceConfig *TerragruntConfig, terr
 
 	// Dependency blocks are shallow merged by name
 	targetConfig.TerragruntDependencies = mergeDependencyBlocks(targetConfig.TerragruntDependencies, sourceConfig.TerragruntDependencies)
+	//TODO: add dependencies
```
brikis98, line 311: ?

brikis98 (commented):
Overall, the approach looks good to me, and I mostly left minor NITs on style.

Do you have an example of the JSON output you could share for a Terragrunt config (especially one that uses a few `include` blocks)? Just curious how this is looking in the real world.

config/config.go
brikis98: Out of curiosity, before this PR, where were the names of the fields used in the `render-json` output coming from?

config/config_partial.go
denis256: Leftovers from my tests

config/include.go
denis256: Leftovers from my tests

config/config.go
denis256: In the current implementation, it is hardcoded during generation: https://github.com/gruntwork-io/terragrunt/blob/master/config/config_as_cty.go#L23

I was thinking to change this part too to use the same constants

config/config.go
denis256:
```
not enough arguments to return
        have (map[string]interface {})
        want (map[string]interface {}, bool)
make: *** [Makefile:34: terragrunt] Error 2
```

brikis98: Oh, Go, what a silly language 😁

brikis98 (approved):
LGTM, thank you!

PR #2246, opened 2022-08-19
#2245: Add option to skip state bucket public access blocking

### LINKED ISSUE #2245: S3 State bucket - put public access block denied
We are using a combination of account level S3 public access block, and an organization SCP that prevents the use of `PutBucketPublicAccessBlock`. 

This results in a (non-blocking, but mildly irritating) access denied error when creating new state buckets, and always having to skip bucket update for public access blocking when re-initialising a module to prevent access denied errors.

Selection_1343

I believe we can use `disable_bucket_update` for the second case, but for the first it seems like Terragrunt will always try to enable public access block on new buckets.

Is there a possibility of introducing a skip for public access blocking, similar to other bucket configurations?

gpdenny:
## Description

Fixes #2245.
Add S3 remote state option to skip setting BucketPublicAccessBlock.

## TODOs

- [x] Update the docs.

## Release Notes 
- Add S3 remote state option `skip_bucket_public_access_blocking` to skip setting BucketPublicAccessBlock


docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -387,6 +387,7 @@ For the `s3` backend, the following additional properties are supported in the `
 - `skip_bucket_accesslogging`: _DEPRECATED_ If provided, will be ignored. A log warning will be issued in the console output to notify the user.
 - `skip_bucket_root_access`: When `true`, the S3 bucket that is created will not be configured with bucket policies that allow access to the root AWS user.
 - `skip_bucket_enforced_tls`: When `true`, the S3 bucket that is created will not be configured with a bucket policy that enforces access to the bucket via a TLS connection.
+- `skip_bucket_public_access_blocking`: When `true`, the S3 bucket that is created will not have public access blocking enabled.
```
gpdenny, line 390: Not sure if this might need a "why", as it might not be immediately clear what this is useful for?

denis256:
Failing tests seems to be unrelated to PR changes

denis256 (approved):

remote/remote_state_s3.go
```diff
@@ -558,13 +562,15 @@ func checkIfS3BucketNeedsUpdate(s3Client *s3.S3, config *ExtendedRemoteStateConf
...
-		needUpdate = append(needUpdate, "Bucket Public Access Blocking")
+	if !config.SkipBucketPublicAccessBlocking {
+		enabled, err := checkIfS3PublicAccessBlockingEnabled(s3Client, &config.remoteStateConfigS3, terragruntOptions)
+		if err != nil {
+			return false, configBucket, err
+		}
+		if !enabled {
+			configBucket.PublicAccess = true
+			needUpdate = append(needUpdate, "Bucket Public Access Blocking")
+		}
```
rhoboat, line 573: I think this logic might be wrong?

docs/_docs/04_reference/config-blocks-and-attributes.md
rhoboat: I think this is poorly named. It's hard to understand. It would be clearer to call this "disable_public_access_blocking" or "enable_public_access". So when this variable is `true`, public access is allowed.

remote/remote_state_s3.go
denis256: Hi, which part? as I understand from code, value `PublicAccess` is used later in the decision to invoke EnablePublicAccessBlockingForS3Bucket 

https://github.com/gruntwork-io/terragrunt/blob/master/remote/remote_state_s3.go#L481

rhoboat: I just wanted to followup-- this logic is not wrong! It's just that the namings are misleading.

Issue #2267, opened 2022-09-06
S3 backend permissions: Missing debug output, silently ignored errors


apparentorder:
Cheers,

Terragrunt's handling of remote_state_s3 permission problems could be improved.

Especially for first-time users with pre-existing terraform state: We use a separate IAM Role for accessing terraform state files; this role had extremely limited permissions -- notably, it did *not* include `s3:GetBucketVersioning` and the other permissions that TG requires for its surprising checks.

In that situation, the first-time experience is a simple `AccessDenied` error, and even with debugging enabled, it gives absolutely no clue where it's coming from. It is, in fact, from a missing `s3:GetBucketVersioning` (see also issue #1312).

Further, TG will keep complaining that the bucket's encryption configuration does not match expectations, even if it already was configured correctly. It turns out that the existing bucket configuration was indeed correct, but the  `s3:GetEncryptionConfiguration` permission was missing. This error is actually recorded in the debug output, but does *not* cause an error for the failed AWS API call, as it should.
From reading the code, the same should be true for the checks on Root Access and Access Logging.

PR will follow.

denis256:
Improved error handling in release https://github.com/gruntwork-io/terragrunt/releases/tag/v0.51.2

PR #2268, opened 2022-09-06
improve s3 error handling (#2267)

apparentorder:
* add missing Logger.Debugf() in checkIfVersioningEnabled()

* in check*(), don't return nil when the AWS API call returned an error


yhakbar:
This was independently adjusted. Sorry we didn't get to this before the PR went stale, @apparentorder !

PR #2306, opened 2022-10-11
Adds regex to retry errors when backend initialization fails

### LINKED ISSUE #2305: Occasional failure to init backend when running against many stacks in parallel (e.g. run-all)
We've occasionally seen an error when initializing an S3 backend when using an EC2 instance profile (or Codebuild execution role) and running many stacks in parallel.

```
Error: error configuring S3 Backend: no valid credential sources for S3 Backend found.
Please see https://www.terraform.io/docs/language/settings/backends/s3.html
for more information about providing credentials.
Error: NoCredentialProviders: no valid providers in chain. Deprecated.
For verbose messaging see aws.Config.CredentialsChainVerboseErrors
```

The first "Error" is printed to stdout, so can't be captured in the current retry logic. But the second "Error" line goes to stderr and can be retried by Terragrunt.

What I suspect is happening is some kind of internal rate-limiting in the metadata service, such that the credential retrieval is failing. Instead of a HTTP 429 though, either Terraform or the SDK are just returning an error. Anyway, we've been using a `retryable_errors` config that includes this line for a week now and the error hasn't occurred since (when before it would happen several times a day).

```
  "(?s).*Error: NoCredentialProviders: no valid providers in chain.*",
```

lorengordon:
Adds regex to retry errors when backend initialization fails

Fixes #2305

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added regex to retry errors when backend initialization fails.

### DIFF
--- a/options/auto_retry_options.go
+++ b/options/auto_retry_options.go
@@ -21,4 +21,5 @@ var DEFAULT_RETRYABLE_ERRORS = []string{
 	"(?s).*ssh_exchange_identification.*Connection closed by remote host.*",
 	"(?s).*Client\\.Timeout exceeded while awaiting headers.*",
 	"(?s).*Could not download module.*The requested URL returned error: 429.*",
+	"(?s).*Error: NoCredentialProviders: no valid providers in chain.*",
 }


Etiene:
Thanks for submitting a PR! I'm not sure why this case should be in the default retries, though. Specially since, as you mentioned in your issue, you could resolve it by using a custom config. Have you tested what will happen when there's an actual credential error instead of the occasional rate-limiting problem?

yhakbar:
Hey @lorengordon ,

I'm closing this PR out as it has gone stale. It's not desirable to delay the error that a user would get when their credentials are in-fact valid by default. Terragrunt should try to exit as quickly as possible in those cases unless a user decides to opt-in to a workaround like this.

lorengordon:
It's unfortunate that the error message is so wildly incorrect. 😢

PR #2348, opened 2022-11-01
#2347 - add --terragrunt-global-cache support to run_cmd

### LINKED ISSUE #2347: ability to cache `run_cmd` ignoring current directory
I've been exploring slow startup times for terragrunt and one of the things I found was our use of `run_cmd` to look up an AWS account id from a name.  It queries AWS Organizations and caches that as a local .json output that subsequent runs will use, so only the first call hits AWS.  But even just this causes terragrunt to take an extra 20 seconds to startup just because of the overhead of `run_cmd` and the number of times it gets called.

terragrunt already has a cache for `run_cmd` but since the key is based on the cmd+directory, we do not get any benefit from that.  Perhaps if there was a way to flag the `run_cmd` output as not reliant on the directory, it could be cached using the same caching mechanism with an empty directory for the key.  This would let me then be able to load an account map with:

```hcl
locals {
    account_map = jsondecode(run_cmd_nodir("/usr/local/bin/get-account-map"))
}
```

and it would only run that command the first time, regardless of which directory the hcl is being parsed in.  Every subsequent one would just come out of the cache.

tjstansell:
## Description

Fixes #2347.

This adds a new `--terragrunt-global-cache` option to `run_cmd` that caches the command globally instead of relative to the directory it is run from.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added `--terragrunt-global-cache` option for `run_cmd` to cache output globally instead of per directory.

### DIFF
--- a/docs/_docs/04_reference/built-in-functions.md
+++ b/docs/_docs/04_reference/built-in-functions.md
@@ -636,7 +636,7 @@ remote_state {
 }
 ```
 
-If the command you are running has the potential to output sensitive values, you may wish to redact the output from appearing in the terminal. To do so, use the special `--terragrunt-quiet` argument which must be passed as the first argument to `run_cmd()`:
+If the command you are running has the potential to output sensitive values, you may wish to redact the output from appearing in the terminal. To do so, use the special `--terragrunt-quiet` argument which must be passed as one of the first arguments to `run_cmd()`:
 
 ``` hcl
 super_secret_value = run_cmd("--terragrunt-quiet", "./decrypt_secret.sh", "foo")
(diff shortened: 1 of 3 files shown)


docs/_docs/04_reference/built-in-functions.md
```diff
@@ -685,6 +685,12 @@ carrot
   * Output contains multiple times `uuid3`, +1 more output comparing to `uuid1` and `uuid2` - because `uuid3` is declared in locals and inputs which add one more evaluation
   * Output contains only once `uuid4` since it is declared only once in `inputs`, `inputs` is not evaluated twice
 
+You can modify this caching behavior to ignore the existing directory if you know the command you are running is not dependent on the current directory path.  To do so, use the special `--terragrunt-global-cache` argument which must be passed as one of the first arguments to `run_cmd()` (and can be combined with `--terragrunt-quiet` in any order):
```
Etiene, line 688: Thanks for the PR! Could you let me know how you tested this behavior? Looking at the code above it doesn't look like they can be combined in any order.

denis256 (approved):

PR #2367, opened 2022-11-20
#1717 Enable security layers for access logging buckets

### LINKED ISSUE #1717: Terragrunt should enable all additional security layers on access logging bucket
E.g., Enable encryption, enable blocking all public access, etc. [This logic here](https://github.com/gruntwork-io/terragrunt/blob/fd49e82778353b2fe5bc52e46ef2c5a5130374c9/remote/remote_state_s3.go#L453-L463) currently only creates the bucket, but does not, for example, call [`EnablePublicAccessBlockingForS3Bucket`](https://github.com/gruntwork-io/terragrunt/blob/fd49e82778353b2fe5bc52e46ef2c5a5130374c9/remote/remote_state_s3.go#L716) to block public access.

denis256:
## Description

Fixes #1717.

Update logic for access buckets creation to use same settings for SSE and TLS

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Updated access logs bucket configuration to configure encryption and TLS

### Migration Guide

Definitions of `remote_state` with `accesslogging_bucket_name` should also define `bucket_sse_algorithm = "AES256"`


brikis98 (dismissed):
Code changes LGTM, but a few sanity checks to do before merging:

Per the [AWS docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/enable-server-access-logging.html), "You can use [default bucket encryption](https://docs.aws.amazon.com/AmazonS3/latest/userguide/default-bucket-encryption.html) on the target bucket only if you use AES256 (SSE-S3). Default encryption with AWS KMS keys (SSE-KMS) is not supported." Is it possible for the user to accidentally use SSE-KMS here? If so, will they get a clear error message?

denis256:
Added validation of `bucket_sse_algorithm` if `accesslogging_bucket_name` is defined, if the configuration is invalid it will throw an error

brikis98 (approved):
LGTM!

PR #2493, opened 2023-03-20
Add --terragrunt-include-module-prefix option

### LINKED ISSUE #1194: Option to prefix *-all terraform output with working-dir label
Right now, all terraform output is unmodified, which makes a lot of sense for regular terraform commands because they are only working on one config at a time and the output can be captured/used for other purposes.

But with the *-all commands, the terraform output of parallel configs just blends together and it is very difficult to determine what line of output is associated with which config.

For example, running `validate-all`, I don't know which config succeeded and which failed.

```
Success! The configuration is valid.

Error: Missing required argument

  on main.tf line 39, in output "repos":
  39: output "repos" {

The argument "value" is required, but no definition was found.
```

I'm suggesting that, only for the `*-all` commands, prefix terraform output the same way the terragrunt output is handled, so we get something like this...

```
[configs/dev/foo] Success! The configuration is valid.

[configs/dev/bar] Error: Missing required argument
[configs/dev/bar] 
[configs/dev/bar]   on main.tf line 39, in output "repos":
[configs/dev/bar]   39: output "repos" {
[configs/dev/bar] 
[configs/dev/bar] The argument "value" is required, but no definition was found.
```

This way we can filter the log very easily to see all the messages for a given config.

maciasello:
## Description

Fixes #1194 .

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added `--terragrunt-include-module-prefix` flag to include module dir prefix in Terraform output

### DIFF
--- a/docs/_docs/04_reference/cli-options.md
+++ b/docs/_docs/04_reference/cli-options.md
@@ -501,6 +501,7 @@ prefix `--terragrunt-` (e.g., `--terragrunt-config`). The currently available op
 - [terragrunt-modules-that-include](#terragrunt-modules-that-include)
 - [terragrunt-fetch-dependency-output-from-state](#terragrunt-fetch-dependency-output-from-state)
 - [terragrunt-use-partial-parse-config-cache](#terragrunt-use-partial-parse-config-cache)
+- [terragrunt-include-module-prefix](#terragrunt-include-module-prefix)
 
 ### terragrunt-config
 
@@ -912,3 +913,10 @@ Currently only AWS S3 backend is supported.
 
 This flag can be used to drastically decrease time required for parsing Terragrunt files. The effect will only show if a lot of similar includes are expected such as the root terragrunt.hcl include.
 NOTE: This is an experimental feature, use with caution.
+
+### terragrunt-include-module-prefix
+
+**CLI Arg**: `--terragrunt-include-module-prefix`
+**Environment Variable**: `TERRAGRUNT_INCLUDE_MODULE_PREFIX` (set to `true`)
+
+When this flag is set output from Terraform sub-commands is prefixed with module path.
--- a/util/prefix-writer.go
+++ b/util/prefix-writer.go
@@ -1,0 +1,37 @@
+package util
+
+import (
+	"bytes"
+	"io"
+)
+
+func PrefixedWriter(writer io.Writer, prefix string) io.Writer {
+	return &prefixedWriter{writer: writer, prefix: prefix, beginningOfANewLine: true}
+}
+
+func (pf *prefixedWriter) Write(p []byte) (int, error) {
+	buf := bytes.Buffer{}
+
+	for _, b := range p {
+		if pf.beginningOfANewLine {
+			buf.WriteString(pf.prefix)
+			pf.beginningOfANewLine = false
+		}
+
+		buf.WriteByte(b)
+
+		pf.beginningOfANewLine = b == '\n'
+	}
+
+	n, err := pf.writer.Write(buf.Bytes())
+	if n > len(p) {
+		n = len(p)
+	}
+	return n, err
+}
+
+type prefixedWriter struct {
+	writer              io.Writer
+	prefix              string
+	beginningOfANewLine bool
+}
--- a/cli/args.go
+++ b/cli/args.go
@@ -156,6 +156,11 @@ func parseTerragruntOptionsFromArgs(terragruntVersion string, args []string, wri
 		opts.Debug = true
 	}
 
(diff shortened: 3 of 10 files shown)


maciasello:
This PR adds a flag to **enable** the prepending of prefixes.

Also it may be reasonable to trim the prefix to be relative to a common parent of the modules or to cwd (or other place) - please guide me what is the suggested approach. In logs `prefix` key is an absolute path - I started from that.

util/prefix-writer.go
```diff
@@ -0,0 +1,37 @@
...
+func (pf *prefixedWriter) Write(p []byte) (int, error) {
+	buf := bytes.Buffer{}
+
+	for _, b := range p {
+		if pf.beginningOfANewLine {
+			buf.WriteString(pf.prefix)
+			pf.beginningOfANewLine = false
+		}
+
+		buf.WriteByte(b)
```
denis256, line 21: I was wondering why output should be buffered and not directly written to `pf.writer`?

maciasello: I did not want to call the underlying `pf.writer` for each character as it may degrade performance.

This buffering does not cross single `.Write()` call span so it will not create a "typical" buffered i/o effect.

Still, it needs an additional memory and one copying of the input `[]byte`. If you feel there is more efficient way I would be more than happy to apply it.

docs/_docs/04_reference/cli-options.md
```diff
@@ -912,3 +913,10 @@ Currently only AWS S3 backend is supported.
 
 This flag can be used to drastically decrease time required for parsing Terragrunt files. The effect will only show if a lot of similar includes are expected such as the root terragrunt.hcl include.
 NOTE: This is an experimental feature, use with caution.
+
+### terragrunt-include-module-prefix
+
+**CLI Arg**: `--terragrunt-include-module-prefix`
+**Environment Variable**: `TERRAGRUNT_INCLUDE_MODULE_PREFIX` (set to `true`)
+
+When this flag is set output from Terraform sub-commands is prefixed with module name.
```
denis256, line 922: Actually, output include module path, not module name.

docs/_docs/04_reference/cli-options.md
```diff
@@ -912,3 +913,10 @@ Currently only AWS S3 backend is supported.
 
 This flag can be used to drastically decrease time required for parsing Terragrunt files. The effect will only show if a lot of similar includes are expected such as the root terragrunt.hcl include.
 NOTE: This is an experimental feature, use with caution.
+
+### terragrunt-include-module-prefix
+
+**CLI Arg**: `--terragrunt-include-module-prefix`
```
denis256, line 919: Noticed that `terragrunt-include-module-prefix` is not listed in the help output
```
$ terragrunt --help | grep terragrunt-include-module-prefix
$
```

docs/_docs/04_reference/cli-options.md
maciasello: Corrected the doc.

docs/_docs/04_reference/cli-options.md
maciasello: Included the info in help output. Is there any other place the help should be updated?

denis256:
```
[INFO] Initializing environment for https://github.com/gruntwork-io/pre-commit.
Terraform fmt............................................................Passed
goimports................................................................Failed
- hook id: goimports
- files were modified by this hook

util/prefix-writer_test.go
```

maciasello:
> ```
> [INFO] Initializing environment for https://github.com/gruntwork-io/pre-commit.
> Terraform fmt............................................................Passed
> goimports................................................................Failed
> - hook id: goimports
> - files were modified by this hook
> 
> util/prefix-writer_test.go
> ```

Fixed.

denis256 (approved):

lorengordon:
> Also it may be reasonable to trim the prefix to be relative to a common parent of the modules or to cwd (or other place) - please guide me what is the suggested approach. In logs `prefix` key is an absolute path - I started from that.

Thanks! Where did this end up? Is the prefix the fully qualified path, or a relative path of some kind? I didn't see an example in the changed files...

lorengordon:
Tested to find the answer myself, it is the absolute path...

dhirschfeld:
> *Tested to find the answer myself, it is the absolute path...*

Yeah, it's the absolute path:
```
[/home/sysop/code/github/dhirschfeld/azure-infra-test/terraform/dev/network]
```

That takes up a lot of screen real-estate. It would be great if the path could be relative to the `cwd` - i.e. running in the `dev` folder that would be:
```
[./network]
```

AlDemion:
As a workaround you can utilize something like this 
terragrunt run-all plan -lock=false --terragrunt-non-interactive | sed "s#$(pwd)/##g"

Issue #1853, opened 2021-10-10
iam_role is broken in v0.34.0


dudicoco:
Hi,

We're using the following configuration in order to make terragrunt work in a multi-account config:

```hcl
iam_role = "arn:aws:iam::${local.common_vars.account_id}:role/terragrunt"

remote_state {
  backend = "s3"

  config = {
    encrypt        = true
    bucket         = "${local.common_vars.account_alias}-terraform"
    region         = "us-east-1"
    key            = "${path_relative_to_include()}/terraform.tfstate"
    acl            = "bucket-owner-full-control"
    dynamodb_table = "terraform"

    dynamodb_table_tags = {
      Provisioner = "terraform"
    }
  }
}

locals {
  common_vars = merge(
    yamldecode(
      file(find_in_parent_folders("account.yaml", "${get_parent_terragrunt_dir()}/empty.yaml"))
    ),
    yamldecode(
      file(find_in_parent_folders("region.yaml", "${get_parent_terragrunt_dir()}/empty.yaml"))
    ),
    yamldecode(
      file(find_in_parent_folders("environment.yaml", "${get_parent_terragrunt_dir()}/empty.yaml"))
    ),
  )
}
```

After upgrading to v0.34.0, i'm getting the following error when running plan on a component in a different account:
```
│ Error: Failed to load state: AccessDenied: Access Denied
│ 	status code: 403, request id: xxxxxxxxxx, host id: xxxxxxxxxx
```

denis256:
Hello, recently was updated how `iam_role` to avoid cases when it is replaced from the environment variables and not evaluated from variables.

I prepared a small example in https://github.com/denis256/terragrunt-test-1853 which load `account_id` / `account_alias` from yaml files, I hope it will help:

```
$ aws sts get-caller-identity
...
"Arn": "arn:aws:sts::666:assumed-role/denis-account"
...

$ tree
.
├── account.yaml
├── app
│   ├── app.tf
│   └── terragrunt.hcl
├── empty.yaml
└── region.yaml
1 directory, 5 files
$ cd app
$ terragrunt_linux_amd64_0_34_1 plan  --terragrunt-log-level debug --terragrunt-debug
...
DEBU[0000] Assuming IAM role arn:aws:iam::test1:role/terragrunt with a session duration of 3600 seconds. 
...
```
So in this example, Terragrunt tries to assume IAM role `test1` defined in `account.yaml`.

[...]

PR #2507, opened 2023-03-31
Lazily evaluate the IAM Role

robmorgan:
## Description

Lazily evaluate the IAM Role. At the moment we are evaluating Terragrunt configurations for the IAM Role parameters regardless of whether the CLI args are set. When running Terragrunt commands against a large repo, the performance penalty quickly adds up.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -743,7 +743,7 @@ var iamRoleCache = NewIAMRoleOptionsCache()
 // setIAMRole - extract IAM role details from Terragrunt flags block
 func setIAMRole(configString string, terragruntOptions *options.TerragruntOptions, includeFromChild *IncludeConfig, filename string) error {
 	// Prefer the IAM Role CLI args if they were passed otherwise lazily evaluate the IamRoleOptions using the config.
-	if terragruntOptions.IAMRoleOptions != terragruntOptions.OriginalIAMRoleOptions {
+	if terragruntOptions.OriginalIAMRoleOptions.RoleARN != "" {
 		terragruntOptions.IAMRoleOptions = terragruntOptions.OriginalIAMRoleOptions
 	} else {
 		// as key is considered HCL code and include configuration
--- a/config/config.go
+++ b/config/config.go
@@ -742,8 +742,7 @@ var iamRoleCache = NewIAMRoleOptionsCache()
 
 // setIAMRole - extract IAM role details from Terragrunt flags block
 func setIAMRole(configString string, terragruntOptions *options.TerragruntOptions, includeFromChild *IncludeConfig, filename string) error {
-	// Prefer the IAM Role CLI args if they were passed otherwise lazily evaluate the IamRoleOptions if the CLI args
-	// weren't passed.
+	// Prefer the IAM Role CLI args if they were passed otherwise lazily evaluate the IamRoleOptions using the config.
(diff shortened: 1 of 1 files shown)


robmorgan:
Hey @denis256, I wonder if I can improve how we evaluate the IAM roles with this PR? At the moment we partially evaluate a Terragrunt configuration for the IAM role regardless of whether or not the CLI args are set. This happens for every file. I'm aware of https://github.com/gruntwork-io/terragrunt/issues/1853, but am I actually regressing anything with this PR?

The one catch with this approach is that any missing IAM CLI arguments won't get merged with what's in a Terragrunt configuration.

config/config.go
```diff
@@ -743,23 +742,28 @@ var iamRoleCache = NewIAMRoleOptionsCache()
...
-	terragruntOptions.IAMRoleOptions = options.MergeIAMRoleOptions(
-		config,
-		terragruntOptions.OriginalIAMRoleOptions,
-	)
+		// We merge the OriginalIAMRoleOptions into the one from the config, because the CLI passed IAMRoleOptions has
+		// precedence.
+		terragruntOptions.IAMRoleOptions = options.MergeIAMRoleOptions(
+			config,
+			terragruntOptions.OriginalIAMRoleOptions,
+		)
```
robmorgan, line 765: I might need to remove this, as OriginalIAMRoleOptions will likely be empty if we fall into this conditional block.

denis256: AFAIK, OriginalIAMRoleOptions can be populated from CLI ARGs

denis256:
So far I thought about 2 approaches that require more refinement/tests:
* parallel evaluation of HCL files 
* lazy evaluation of IAM role (and other fields too) - single parsing of HCL files and fetching of "promises" for fields which are evaluated(once and cached) on reading value

config/config.go
robmorgan: So I guess the other fields will be overridden unless `RoleARN` is set from the CLI args.

robmorgan:
@denis256 

> So far I thought about 2 approaches that require more refinement/tests:
>parallel evaluation of HCL files
> lazy evaluation of IAM role (and other fields too) - single parsing of HCL files and fetching of "promises" for fields which are evaluated(once and cached) on reading value

Ok nice! Do you think they'll take a while to implement? I guess I'm primarily trying to see if we can find a quick win that is mostly backwards compatible. At the moment on some basic benchmarks, we're spending about ~15% of the CPU budget in the `config.setIAMRole` function. I was hoping to dramatically reduce that if the user specifies one (`--terragrunt-iam-role`) or all of the IAM CLI args.

Screenshot 2023-04-03 at 7 35 58 pm

denis256:
It is not easy a simple change - both approaches will require a significant change in the handling of HCL files

robmorgan:
@denis256 okay just so I gather this correctly. We cannot merge this PR "as-is", even if the tests pass? Do I need to take a different approach?

robmorgan:
@denis256 hey Denis, just a friendly bump on the message above.

denis256:
I think with implemented changes, will be good enough to validate through existing integration tests, which seems to pass now

denis256 (approved):

robmorgan:
@denis256 okay, thats good to hear! Hopefully I haven't regressed anything.

PR #2204, opened 2022-07-19
Improve partial parsing HCL strings by adding cache

maunzCache:
## Description

Fixes #2203.

This PR will add a caching behavior to `PartialParseConfigString` in `config.go`. The original function has been wrapped by a new function called `TerragruntConfigFromPartialConfigString` which handles cache access if the flag `--terragrunt-use-partial-parse-config-cache` was provided.

This fix was co-authored by @nilreml.

We are very open to suggestions or objections of this PR. If you find anything that we overlooked, please speak out.

### Technical subtleties

The implementation uses the same approach as the `setIAMRole` cache in `config.go` (#2010) but drops the additional hashing of the key used in the cache map. We decided to drop this as our understanding is that this interferes with the native behavior of go maps. Go implements the map type as hashmap which already handles uniqueness and collisions by providing a cache bucket implementation. See [the docs and implementation](https://github.com/golang/go/blob/master/src/runtime/map.go) for reference. 

### Testing
As we have no concrete unit test for this implementation besides verifying that our cache works in general, we decided to add two fixtures.

The fixtures are very similar: Both implement dependencies and require heavy parsing due to massive stub data. The first variant implements includes by creating locals which call `read_terragrunt_config`. The second implements the multi-include approach which was implemented in 0.32.
Both fixtures are used in [benchmark tests](https://pkg.go.dev/testing#hdr-Benchmarks) to verify the impact of the implementation.

One benchmark has been implemented for the execution of `runGraphDependencies` which is the fastest functionality that accesses `PartialParseConfigString` and could additionally be used to verify if dependency order is correct even after enabling the cache. The second benchmark was implemented for `ReadTerragruntConfig` as this will call all subsequent functionality which is necessary for parsing. Each benchmark consists of one of the fixtures plus the caching enabled or disabled.

The benchmarks verify that CPU and memory usage don't change drastically when using the cache. However, for huge projects which surely exceed the fixture, use cases should be considered to implement a cache invalidation to prevent out of memory errors.

We realize the limited benefits since this project doesn't make use of benchmarks yet and may have to decide on how to handle them. We suggest implementing least a benchmark call in the CI/CD pipeline which is executing the unit and integration tests.
For the future all benchmarks should be collected and somehow be compared to easily find regressions if features are added or changed. [This blog post](https://blog.logrocket.com/benchmarking-golang-improve-function-performance/) might help with an implementation.

It has to be noted that benchmark tests are not representative but validate our use case.

### Implementation notes

[...]

Issue #2203, opened 2022-07-19
Improve partial parsing HCL strings by adding cache


maunzCache:
### Motivation
As outlined in #1971 my customer still has issues with the parsing time of the later Terragrunt versions since 0.32. Still there is a requirement for my team to get a new version of Terraform and Terragrunt working ASAP. So @nilreml and I took some time to understand the Terragrunt code better and find a potential solution to our issue.
To get this straight: We are working on a big project containing ~150 AWS accounts, 6 environments and ~75 Terraform modules. The Terragrunt code is sometimes very complex containing multiple dependency layers. We already took the time to optimize and review all of the `*.hcl` files to reduce `includes/read_terragrunt_config` calls, calls to other custom functions and simplifying logic of loops. Still we only get minor improvements in our parsing times (time until all dependencies are resolved and Terraform is triggered the first time).

After the merge of #2010 and the 0.36.6 release another colleague of mine did some re-measuring of execution times in our environment:

| tg version | init 1 | init 2 | init 3 | plan 1 | avg | factor to 0.29.2 |
|------------|-------|-------|-------|--------|-----|------------------|
| 0.29.2 parsing | 0:01:11 | 0:00:50 | 0:01:14 | 0:01:02 | 64s | 1.0 |
| 0.29.2 complete | 0:04:11 | 0:04:32 | 0:04:13 | 0:04:14 | 257s | 1.0 |
| 0.33.2 parsing | 0:01:15 | 0:01:44 | -  | -  | 127s | 2.0 |
| 0.33.2 complete | 0:05:48 | 0:07:34 | -  | -  | 401s | 1.6 |
| 0.34.3 parsing | 0:05:54 | 0:05:21 | -  | -  | 337s | 5.3 |
| 0.34.3 complete | 0:24:59 | 0:21:02 | -  | -  | 1381s | 5.4 |
| 0.35.20 parsing | 0:04:50 | 0:04:34 | -  | -  | 282s | 4.4 |
| 0.35.20 complete | 0:23:39 | 0:23:14 | -  | -  | 1406s | 5.4 |
| 0.36.3 parsing | 0:04:23 | 0:04:44 | -  | -  | 274s | 4.3 |
| 0.36.3 complete | 0:24:26 | 0:21:57 | -  | -  | 1391s | 5.4 |
| 0.36.6 parsing | 0:01:16 | 0:02:09 | 0:01:52 | 0:02:11 | 112s | 1.8 |
[...]

PR #2518, opened 2023-04-05
#2517 fix parsing cache key construction

### LINKED ISSUE #2517: fix parsing cache key construction
**Preface**

With terragrunt v0.38.9 our HCL parsing cache contribution from issue #2203 (PR #2204) was added to terragrunt which improved the parsing speed of HCL files by introducing an in-memory cache:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.38.9

For detailed information, please refer to #2203 and #2204

**Motivation for this issue**

During our daily work, we encountered that the new parsing cache key has a bug which occurs only in certain, rare conditions.
Specifically, it only occurs if a Terragrunt module .hcl file depends on itself in a different folder, which was previously not the case and therefor couldn't been tested in our environments.

**Solution**

Adding a missing parameter (`filename`) in the cache key construction of `TerragruntConfigFromPartialConfigString`:

Stripped snippet to demonstrate the change:

``` go
func TerragruntConfigFromPartialConfigString(...) (...) {
...
- var cacheKey = fmt.Sprintf("%#v-%#v-%#v", configString, includeFromChild, decodeList)
+ var cacheKey = fmt.Sprintf("%#v-%#v-%#v-%#v", filename, configString, includeFromChild, decodeList)
...
}
```

Implementation will be provided in the following Pull Request: #2518

untcha:
## Description

**Preface**

With v0.38.9 our HCL parsing cache contribution from issue #2203 (PR #2204) was added to terragrunt which improved the parsing speed of HCL files by introducing an in-memory cache:
https://github.com/gruntwork-io/terragrunt/releases/tag/v0.38.9

For detailed information, please refer to #2203 and #2204

**Motivation for this issue**

During our daily work, we encountered that the new parsing cache key has a bug which occurs only in certain, rare conditions.
Specifically, it only occurs if a Terragrunt module .hcl file depends on itself in a different folder, which was previously not the case and therefor couldn't been tested in our environments.

**Solution**

Adding a missing parameter (`filename`) in the cache key construction of `TerragruntConfigFromPartialConfigString`

### Documentation

n/a

## TODOs

Please ensure all of these TODOs are completed before asking for a review.

- [x] Ensure the branch is named correctly with the issue number.
- [x] Keep the changes backward compatible where possible.
- [x] Run the pre-commit checks successfully.
- [x] Run the relevant tests successfully.

## Related Issues

Fixes #2517 

Related to #2203 and #2204


maunzCache:
I don't want to vote against this MR in general, but adding the filename will have an major effect on the caching implementation. The logic was to cache duplicate blocks of code and not cache individual files. Addin the filename will break the initial idea of that because you will only cache files. While this will still boost reoccurring dependency parsing, it does not have the benefit of different files containing the same code logic.

In addition, i want to point out that there was no consideration in changing the test suite for the implementation to confirm that the changes have an impact or not.

denis256:
Hi,
I agree with the previous note - will be helpful to have test that will show improvements added by this PR

maunzCache:
> Hi, I agree with the previous note - will be helpful to have test that will show improvements added by this PR

This is also a nice case for the benchmarks that we included in the initial MR ;)

untcha:
Hi Kevin (@maunzCache), hi Denis (@denis256),

in general I agree with Kevin. Adding the `filename` is against the initial idea.

But in our case the following issue occurs. I need to explain a bit about the environment.
Kevin should know it (since he worked for us in the past), but I guess not the latest changes.

One remark: performance is not an issue. The performance decreased only slightly (tested by Merlin)

### Explaining our environment and dependencies

In the following picture you can see a snippet which shows a simple representation of our terragrunt environment:

terragrunt_upstream_merge

The important part is the multi-region config in this example account: `aws-account-001`
We have `eu-central-1` and `eu-west-3` both using a so called `baseline-common` module.
Both `terragrunt.hcl` files are *Symlinks* to the one and only `baseline-common.hcl` in the `default-hcls` folder.

In there we have a dependency block to a kms key.

The `config_path` is built by checking if we are in `eu-central-1` if yes then the multiregion key
is generated and later used in eu-central-1 by using this path `"../../global/baseline-common-global"`

If not, then we are in `eu-west-3` and then this path is used `"../../eu-central-1/baseline-common”` and in eu-west-3 the replica kms key.

``` hcl
dependency "baseline_common_kms_key" {
  config_path = local.region == "eu-central-1" ? "../../global/baseline-common-global" : "../../eu-central-1/baseline-common"
  mock_outputs = {
    default_cmk_arn    = "arn:aws:kms:eu-central-1:..."
    cloudtrail_cmk_arn = "arn:aws:kms:eu-central-1:..."
  }
}
```

### Reproducing the issue without and with the `filename` parameter in the cacheKey

#### Test 1:
- terragrunt version v0.45.2 (NO FORK!)
- TERRAGRUNT_USE_PARTIAL_PARSE_CONFIG_CACHE=**true**
- terragrunt run-all init in /terragrunt-environment/dev3/spoke/aws-spoke-dev3-a/eu-west-3/baseline-common

#### Test 1 result: init NOT successful

``` shell
time=2023-04-12T16:24:29+02:00 Executing: terragrunt run-all init --terragrunt-ignore-external-dependencies  --terragrunt-parallelism 20

time=2023-04-12T16:24:31+02:00 level=error msg=Found a dependency cycle between modules: /Users/untcha/.../terragrunt-environment/dev3/spoke/aws-spoke-dev3-a/eu-central-1/baseline-common -> /Users/untcha/.../terragrunt-environment/dev3/spoke/aws-spoke-dev3-a/eu-central-1/baseline-common
time=2023-04-12T16:24:31+02:00 level=error msg=Unable to determine underlying exit code, so Terragrunt will exit with error code 1

…/.../terragrunt-environment/dev3/spoke/aws-spoke-dev3-a/eu-west-3/baseline-common [test] terragrunt run-all init —terragrunt-ignore-external-dependencies  --terragrunt-parallelism 20
```

#### Test 2:
- terragrunt version v0.45.2 (NO FORK!)
- TERRAGRUNT_USE_PARTIAL_PARSE_CONFIG_CACHE=**false**
- terragrunt run-all init in /terragrunt-environment/dev3/spoke/aws-spoke-dev3-a/eu-west-3/baseline-common

#### Test 2 result: init successful

#### Test 3:
- terragrunt version v0.45.2.1 (forked version with `filename` param)
- TERRAGRUNT_USE_PARTIAL_PARSE_CONFIG_CACHE=**true**
- terragrunt run-all init in /terragrunt-environment/dev3/spoke/aws-spoke-dev3-a/eu-west-3/baseline-common

#### Test 3 result: init successful

denis256:
Hello,
will be very helpful to provide an example repository where this issue occurs

I tried to reproduce a similar setup in https://github.com/denis256/terragrunt-tests/tree/master/issue-2518 but still can't get `Found a dependency cycle between modules` errors during Terragrunt execution

untcha:
Hi @denis256,

thanks a lot for your effort. We will try to provide a repo with an environment to reproduce.
This will take us a bit of time.

Thanks a lot!
Alex

untcha:
Hi @denis256,

we spent some time on stripping down one of our environments to make it as small as possible with a good use case to show the error. Since this was still too much. One of our developers (@jkarkoszka) tried the other way round and created a very small example repository to show exactly our error.

Here you can find this sandbox repo: https://github.com/jkarkoszka/terragrunt-environment-for-pr-2518

The repository also includes two terragrunt binaries. The original v0.44.5 and the forked one from us, which includes the change from this PR. So it should be very easy without configuration to clone the repo and follow the instructions to reproduce the error.

Our current approach to get the region is to retrieve it from the directory path. We thought that this is maybe not the way gruntwork would propose it to do. @sdahlen-reply tried the approach to retrieve the region from a `region.hcl` file. The result is the same dependency cycle with that approach.
You can find this in the branch `region-from-file`

Thanks a lot!

Thanks to Sebastian (@sdahlen-reply) and Jakub (@jkarkoszka)

denis256:
The original issue is that the cached value of `TerragruntConfig` already has an evaluated value of `config_path` which leads to a cycle error between dependencies

https://github.com/jkarkoszka/terragrunt-environment-for-pr-2518/blob/main/default-hcls/baseline-common.hcl#L18

denis256 (approved):
Raised separated issue https://github.com/gruntwork-io/terragrunt/issues/2643 to implement tracking of HCL cache performance

PR #2547, opened 2023-04-25
Add explanation for terraform errors

### LINKED ISSUE #2512: Terragrunt fails to report useful error information.
In a number of situations terragrunt will output one to many errors that apply to specific dependencies of a modules without providing any context of what module was being operated on at the time of the issue.

In the following example a terragrunt module with 3 dependencies is in an odd state (the .terragrunt-cache directories need to be remove for one reason or another). Running `terragrunt init` produces the following output.

```
terragrunt init

Error: Initialization required. Please see the error message above.



Error: Initialization required. Please see the error message above.



Error: Initialization required. Please see the error message above.


ERRO[0016] exit status 1    
```

Ideally, terragrunt would produce usable contextual information about the issue encountered to help end users debug. In this particular situation, I would expect the name of each module to be printed, as well as the path to the module. Additionally `Please see the error message above.` suggests that some additional debugging information is being suppressed.

### LINKED ISSUE #2513: Terragrunt often and randomly needs the .terragrunt-cache directory manually removed.
Our developers often receive the following output when attempting to run a terragrunt plan , terragrunt apply, or terragrunt init.

```terragrunt init

Error: Initialization required. Please see the error message above.



Error: Initialization required. Please see the error message above.



Error: Initialization required. Please see the error message above.


ERRO[0016] exit status 1    
```
The only apparent solution is deleting all occurrences of `.terragrunt-cache` in the module structure. Ideally terragrunt would be able to detect that a `.terragrunt-cache` was in a bad state and be able to report more helpfully, or be able to avoid the initial situation from occurring.

### LINKED ISSUE #2514: Terragrunt error messages when not auth'd are confusing
When not auth'd, terragrunt gives the message `Error: Initialization required. Please see the error message above.` - there is no message above, which is not intuitive to know what this means, and that auth is required.

Secondly, when auth'd in the wrong environment (when there is a dependency, it is fine - `Failed to load state: AccessDenied: Access Denied`, when there is not, it is a bit more baffling - `Remote state S3 bucket dev-terraform-state does not exist or you don't have permissions to access it. Would you like Terragrunt to create it? (y/n)`). Again, it would be great to know that this is an auth error, rather than having a dev bucket made in stage, for example.

denis256:
## Description

Updated handling of terraform output to print explanation messages in case of errors.

Fixes #2514.
Fixes #2512.
Fixes #2513.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added terraform errors explanation message.

### DIFF
--- a/main.go
+++ b/main.go
@@ -41,6 +41,9 @@ func checkForErrorsAndExit(err error) {
 			exitCode = 1
 			util.GlobalFallbackLogEntry.Errorf("Unable to determine underlying exit code, so Terragrunt will exit with error code 1")
 		}
+		if explain := shell.ExplainError(err); len(explain) > 0 {
+			util.GlobalFallbackLogEntry.Errorf("Error handling suggestions: %s", explain)
+		}
 		os.Exit(exitCode)
 	}
 
--- a/main.go
+++ b/main.go
@@ -42,7 +42,7 @@ func checkForErrorsAndExit(err error) {
 			util.GlobalFallbackLogEntry.Errorf("Unable to determine underlying exit code, so Terragrunt will exit with error code 1")
 		}
 		if explain := shell.ExplainError(err); len(explain) > 0 {
-			util.GlobalFallbackLogEntry.Errorf("Error handling suggestions: %s", explain)
+			util.GlobalFallbackLogEntry.Errorf("Suggestions on how to fix failure: %s", explain)
 		}
 		os.Exit(exitCode)
 	}
--- a/main.go
+++ b/main.go
@@ -42,7 +42,7 @@ func checkForErrorsAndExit(err error) {
 			util.GlobalFallbackLogEntry.Errorf("Unable to determine underlying exit code, so Terragrunt will exit with error code 1")
 		}
 		if explain := shell.ExplainError(err); len(explain) > 0 {
-			util.GlobalFallbackLogEntry.Errorf("Suggested fixes: %s", explain)
+			util.GlobalFallbackLogEntry.Errorf("Suggested fixes: \n%s", explain)
 		}
 		os.Exit(exitCode)
 	}
--- a/main.go
+++ b/main.go
@@ -42,7 +42,7 @@ func checkForErrorsAndExit(err error) {
 			util.GlobalFallbackLogEntry.Errorf("Unable to determine underlying exit code, so Terragrunt will exit with error code 1")
 		}
 		if explain := shell.ExplainError(err); len(explain) > 0 {
-			util.GlobalFallbackLogEntry.Errorf("Suggestions on how to fix failure: %s", explain)
+			util.GlobalFallbackLogEntry.Errorf("Suggested fixes: %s", explain)
 		}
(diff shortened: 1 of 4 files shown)


shell/error_explainer.go
```diff
@@ -0,0 +1,45 @@
...
+	"fmt"
+	"regexp"
+
+	"github.com/gruntwork-io/terragrunt/errors"
+	"github.com/hashicorp/go-multierror"
+)
+
+// terraformErrorsMatcher List of errors that we know how to explain to the user. The key is a regex that matches the error message, and the value is the explanation.
+var terraformErrorsMatcher = map[string]string{
+	"(?s).*Error refreshing state: AccessDenied: Access Denied(?s).*": "You don't have access to the S3 bucket where the state is stored. Check your AWS credentials and permissions.",
```
zackproser, line 13: NIT / just curious: is there any way we could output the actual bucket name that they were not able to access? That might make it even easier to understand what's happening?

zackproser (dismissed):
LGTM

main.go
```diff
@@ -41,6 +41,9 @@ func checkForErrorsAndExit(err error) {
 			exitCode = 1
 			util.GlobalFallbackLogEntry.Errorf("Unable to determine underlying exit code, so Terragrunt will exit with error code 1")
 		}
+		if explain := shell.ExplainError(err); len(explain) > 0 {
+			util.GlobalFallbackLogEntry.Errorf("Suggestions on how to fix failure: %s", explain)
```
ellisonc, line 45:
```suggestion
			util.GlobalFallbackLogEntry.Errorf("Suggested fixes: %s", explain)
```

shell/error_explainer.go
```diff
@@ -0,0 +1,45 @@
...
+		}
+		processError, ok := originalError.(ProcessExecutionError)
+		if !ok {
+			continue
+		}
+		errorOutput := processError.Stderr
+		for regex, explanation := range terraformErrorsMatcher {
+			if match, _ := regexp.MatchString(regex, errorOutput); match {
+				// append explanation to result
+				result = fmt.Sprintf("%s\n%s", result, explanation)
```
ellisonc, line 40: Is there any scenario where this could cause duplication of the same explanation? For example if multiple modules had source changes?  Or does terragrunt fail fast in that case? If this is a risk I'd collect these in a set and only print the distinct messages at the end

ellisonc (commented):
A small NIT and 1 question for clarification

zackproser (approved):
Curious about the two open questions but otherwise LGTM - approving to unblock you.

shell/error_explainer.go
denis256: Added deduplication of explanations

shell/error_explainer.go
denis256: Having bucket name may help in troubleshooting, but since name is not included in output it makes it more complicated to be printed.

Additionally, it will require implementing some kind of templating language on explanations to allow inserting custom values.

zackproser (approved):
LGTM

Issue #61, opened 2016-12-01
release lock on SIGTERM too.


mildred:
On SIGINT (Interrupt from keyboard), terragrunt release the lock, but not on SIGTERM (Termination signal). This should probably be fixed. See `locks/lock.go`.

```
diff --git a/locks/lock.go b/locks/lock.go
index 58ace31..13af32b 100644
--- a/locks/lock.go
+++ b/locks/lock.go
@@ -5,6 +5,7 @@ import (
        "github.com/gruntwork-io/terragrunt/errors"
        "os"
        "os/signal"
+       "syscall"
 )
 
 // Every type of lock must implement this interface
@@ -48,6 +49,7 @@ func WithLock(lock Lock, action func() error) (finalErr error) {
        // the blocking call to action() to return normally.
        signalChannel := make(chan os.Signal, 1)
        signal.Notify(signalChannel, os.Interrupt)
+       signal.Notify(signalChannel, syscall.SIGTERM)
        go func() { util.Logger.Printf("Caught signal '%s'. Terraform should be shutting down gracefully now.", <- signalChannel) }()
 
        return action()
```

mildred:
Additionally, terragrunt should forward SIGTERM signals to terraform. There is no guaantee that terraform will have received it. See also hashicorp/terraform#10459

mildred:
In fact, on some cases (default on systemd services) SIGTERM is sent to all processes in the cgroup. In other cases (sysvservice, docker) the SIGTERM signal is only sent to the parent process.

brikis98:
Good point. Looks like you have some code written for this already. Would you like to submit a PR?

mildred:
Yes, I will submit a PR once I get something with terraform. There are two aspects:

- handle SIGTERM as well as SIGINT
- make sure terraform has received SIGTERM. When we type Ctrl-C in a terminal, the VT will make sure that all VT clients are sent a SIGINT signal. With SIGTERM, we are not sure that terraform received SIGTERM and we must probably forward the signal.

Are you ok with forwarding the SIGTERM signal ?

When forwarding SIGTERM we must make sure terraform can handle it. Currently it does not respond to SIGTERM and doesn't cleanly finish like it does with Ctrl-C. I'm currently submitting a PR on terraform for that.

brikis98:
Probably makes sense to wait for your PR to be merged into Terraform itself.

mildred:
yes

brikis98:
[...]

PR #2559, opened 2023-05-06
fix: disable sending a second interrupt signal to `terraform`

### LINKED ISSUE #2120: Send ctrl+c one time but terragrunt send twice to terraform
Sorry that might be a stupid question.  When I abort the apply when resource is being created via ctrl+c one time(want it to be gracefully shutdown), terragrunt send twice to terraform so there is no gracefully shutdown and result in state lock not being released and the hash value in the table would be mismatch even I run force-unlock.  May I know is it the design or only my laptop encounter this issue.

```
^C
Interrupt received.
Please wait for Terraform to exit or data loss may occur.
Gracefully shutting down...


Two interrupts received. Exiting immediately. Note that data loss may have
occurred.

╷
│ Error: operation canceled
```

terragrunt version v0.37.1
Terraform v1.2.1
on darwin_amd64
+ provider registry.terraform.io/hashicorp/aws v4.15.1
+ provider registry.terraform.io/hashicorp/cloudinit v2.2.0
+ provider registry.terraform.io/hashicorp/kubernetes v2.11.0
+ provider registry.terraform.io/hashicorp/local v2.2.3
+ provider registry.terraform.io/hashicorp/tls v3.4.0

levkoburburas:
## Description

When `CTRL+c` keys are pressed, `terragrunt` forwards this interrupt signal to the child process `terraform`, while the shell already passes interrupt signal to the main process (in our case `terragrunt`) and its children as well (in our case `terraform`). As a result, `terraform` receives the interrupt signal twice and cancels graceful shutdown.

```
Interrupt received.
Please wait for Terraform to exit or data loss may occur.
Gracefully shutting down...


Two interrupts received. Exiting immediately. Note that data loss may have
occurred.
```

Fixes #2120.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

fix: disable sending a second interrupt signal to `terraform`

### DIFF
--- a/shell/run_shell_cmd.go
+++ b/shell/run_shell_cmd.go
@@ -21,8 +21,8 @@ import (
 
 // The signal can be sent to the main process (only `terragrunt`) as well as the process group (`terragrunt` and `terraform`), for example:
 // kill -INT <pid>  # sends SIGINT only to the main process
-// kill -INT -<pid> # sends SIGINT to the process group using
-// Since we cannot know how the signal was sent, we should give `terraform` time to gracefully exit if it receives the signal directly from the shell, to avoid sending the second interrupt signal to `terraform`.
+// kill -INT -<pid> # sends SIGINT to the process group
+// Since we cannot know how the signal is sent, we should give `terraform` time to gracefully exit if it receives the signal directly from the shell, to avoid sending the second interrupt signal to `terraform`.
 const signalForwardingDelay = time.Second * 30
 
 // Commands that implement a REPL need a pseudo TTY when run as a subprocess in order for the readline properties to be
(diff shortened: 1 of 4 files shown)


Etiene:
Thanks for the PR! How have you tested this? Will it not make [this issue](https://github.com/gruntwork-io/terragrunt/issues/61) resurface?

levkoburburas:
> How have you tested this? Will it not make [this issue](https://github.com/gruntwork-io/terragrunt/issues/61) resurface?

Yes, it could. This [comment](https://github.com/gruntwork-io/terragrunt/issues/61#issuecomment-264428096) made it clear to me that the signal can be sent in a variety of ways. I tested in a shell that sends a signal to a process group, in which case we shouldn't forward the signal, but it can also only be sent to the main process and than we need to forward the signal. Thus, we must consider all cases.

Fixed and added unit test.

Thank you for the help @Etiene!

zackproser (approved):
LGTM

denis256 (approved):

levkoburburas:
Thank you for the review @zackproser, @denis256!

PR #2586, opened 2023-05-31
feat: support git tag in `--terragrunt-source-map` option

### LINKED ISSUE #2460: `--terragrunt-source-map` does not work with `?ref=`
In my workflow one of the main use cases for `--terragrunt-source-map` would be to point all my module-sources to a feature branch I am working on like this:

```
terragrunt plan --terragrunt-source-map git::ssh://git@github.com/myORG/terraform-modules.git=git::ssh://git@github.com/myORG/terraform-modules.git?ref=FEATURE
```

Unfortunately due to the way the source-map is parsed having a `=` in the argument breaks it:

```
ERRO[0000] Invalid key-value pair. Expected format KEY=VALUE, got git::ssh://git@github.com/myORG/terraform-modules.git=git::ssh://git@github.com/myORG/terraform-modules.git?ref=FEATURE.
```

I tried escaping the second `=`, but this didn't help.

levkohimins:
## Description

Added support git tag in URL for `--terragrunt-source-map` option. For example:
```
--terragrunt-source-map github.com/old-org/modules.git=git::github.com/new-org/modules.git?ref=feature
```

Fixes #2460.

## TODOs

- [] Update the docs.

## Release Notes (draft)

Added support git tag in URL for `--terragrunt-source-map` option.

### DIFF
--- a/util/collections.go
+++ b/util/collections.go
@@ -172,8 +172,27 @@ func StringListInsert(list []string, element string, index int) []string {
 // KeyValuePairListToMap converts a list of key value pair encoded as `key=value` strings into a map.
 func KeyValuePairStringListToMap(asList []string) (map[string]string, error) {
 	asMap := map[string]string{}
+
 	for _, arg := range asList {
-		parts := strings.Split(arg, "=")
+		var parts []string
+
+		// splits the string to the slice `parts`, using `=` sign as delimiter, but taking into account that the `=` sign can also be used as a git tag, e.g. `git@github.com/test.git?ref=feature`
+		for i, l := 0, 0; i <= len(arg); i++ {
+			if i == len(arg) {
+				parts = append(parts, arg[l:])
+				break
+			}
+
+			if arg[i] == '=' {
+				if i >= 4 && arg[i-4:i] == "?ref" {
+					continue
+				}
+
+				parts = append(parts, arg[l:i])
+				l = i + 1
+			}
+		}
+
 		if len(parts) != 2 {
 			return nil, errors.WithStackTrace(InvalidKeyValue(arg))
 		}
--- a/cli/download_source.go
+++ b/cli/download_source.go
@@ -175,9 +175,10 @@ func readVersionFile(terraformSource *tfsource.TerraformSource) (string, error) 
 }
 
 // updateGetters returns the customized go-getter interfaces that Terragrunt relies on. Specifically:
-// - Local file path getter is updated to copy the files instead of creating symlinks, which is what go-getter defaults
-//   to.
-// - Include the customized getter for fetching sources from the Terraform Registry.
+//   - Local file path getter is updated to copy the files instead of creating symlinks, which is what go-getter defaults
+//     to.
+//   - Include the customized getter for fetching sources from the Terraform Registry.
+//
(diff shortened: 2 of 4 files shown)


util/collections.go
```diff
@@ -172,8 +172,27 @@ func StringListInsert(list []string, element string, index int) []string {
...
+
+		// splits the string to the slice `parts`, using `=` sign as delimiter, but taking into account that the `=` sign can also be used as a git tag, e.g. `git@github.com/test.git?ref=feature`
+		for i, l := 0, 0; i <= len(arg); i++ {
+			if i == len(arg) {
+				parts = append(parts, arg[l:])
+				break
+			}
+
+			if arg[i] == '=' {
+				if i >= 4 && arg[i-4:i] == "?ref" {
```
denis256, line 187: Not sure if in `util/collections.go`, which can be used in multiple places, should be implemented handling of a specific case for "?ref" strings, it may have side effects in other places

denis256: Can we try to parse map in handling of `optTerragruntSourceMap`?
https://github.com/gruntwork-io/terragrunt/blob/master/cli/args.go#LL101C12-L101C12

levkohimins: > Can we try to parse map in handling of `optTerragruntSourceMap`? https://github.com/gruntwork-io/terragrunt/blob/master/cli/args.go#LL101C12-L101C12

In this case, URLs passed via env var `TERRAGRUNT_SOURCE_MAP` will not be parsed `?ref=` as part of the URL and will break the URL at that point. In my opinion, data passed through the flag or the environment variable should be processed the same way. Or have I missed something?
https://github.com/gruntwork-io/terragrunt/blob/1494e777b1f2ce18797ce1c9e46859bdef7f33ed/cli/args.go#L97

> Not sure if in `util/collections.go`, which can be used in multiple places, should be implemented handling of a specific case for "?ref" strings, it may have side effects in other places

I checked, the func is used only in two places and is relevant in both places, the reason described above.
```
cli/args.go:499:14:     return util.KeyValuePairStringListToMap(asList)
cli/args.go:511:14:     return util.KeyValuePairStringListToMap(mappingsAsList)
```

denis256: For me, looks like the single responsibility of `KeyValuePairStringListToMap()` is broken - instead of splitting string by `=` it is also checking for `?ref` values.

I was wondering if we can prepare input data to be handled correctly by `KeyValuePairStringListToMap()`

levkohimins: I don't think so, this function keeps going to do the same task, splitting string to _source_ and _destination_ URL, for example, the func accepts the follow string 
`git@github.com/old.git?ref=some-tag=git@github.com/new.git?ref=another-tag`, and everything it does is to split string at the correct point 
`git@github.com/old.git?ref=some-tag` `git@github.com/new.git?ref=another-tag` instead of 
`git@github.com/old.git?ref` `some-tag` `git@github.com/new.git?ref` `another-tag`. 
So we can't do any preparation, because any preparation requires proper string splitting first.

denis256 (approved):

levkohimins:
@denis256, thanks for the review and improvement idea.

PR #2652, opened 2023-08-01
use impersonate package for GCP service impersonation

### LINKED ISSUE #1997: GCS bucket creation fails when service account impersonation is performed from another service account
Hi,

I'm trying to use Terragrunt on GCP with a GCS based remote backend for the Terraform state. We use a setup with service account impersonation to do our deploys. We have one service account with an access key that authenticates and then impersonates a second service account that has access to the different required Google APIs. 

The Terragrunt remote config looks like this:
```
remote_state {
  backend = "gcs"

  config = {
    project                     = local.project_id
    bucket                      = "${local.project_name}-tf-state"
    prefix                      = path_relative_to_include()
    location                    = local.gcp_region
    impersonate_service_account = local.service_account
    skip_bucket_creation        = true
    skip_bucket_versioning      = true
  }

  generate = {
    path = "backend.tf"
  }
}

```

When running Terraform we run into the following error, Terragrunt fails to auto initialize the GCS bucket:
```
Create GCS bucket YYYYYYYY returned an error: Post "https://storage.googleapis.com/storage/v1/b?alt=json&prettyPrint=false&project=ZZZZZZZZ": impersonate: status code 403: {
  "error": {
    "code": 403,
    "message": "Request had insufficient authentication scopes.",
    "status": "PERMISSION_DENIED",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.ErrorInfo",
        "reason": "ACCESS_TOKEN_SCOPE_INSUFFICIENT",
        "domain": "googleapis.com",
        "metadata": {
          "method": "google.iam.credentials.v1.IAMCredentials.GenerateAccessToken",
          "service": "iamcredentials.googleapis.com"
        }
      }
    ]
  }
}
```

Terragrunt and Terraform versions:
```
➜ terraform -v
Terraform v1.1.4
on linux_amd64
➜ terragrunt -v
terragrunt version v0.29.10
```
I also tested terragrunt version v0.36.0, with same result.

If we reconfigure the remote to skip bucket creation and versioning and instead create the bucket manually everything works and Terraform can access GCP using SA impersonation without any problems.

I did some digging around and it seems like a similar problem was discovered some time ago in the Terraform project (https://github.com/hashicorp/terraform/issues/28139) but it has now been resolved by replacing one of the libraries (cloud.google.com/go/storage) used for communicating with GCP APIs. It looks like Terragrunt is still using the cloud.google.com/go/storage library (https://github.com/gruntwork-io/terragrunt/blob/master/go.mod#L7), so that makes me think that's the root cause of the issue I'm encountering.

richardjennings:
## Description

Fixes #1997.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Updated GCS Client with Impersonate package for Service Impersonation.

### DIFF
--- a/remote/remote_state_gcs.go
+++ b/remote/remote_state_gcs.go
@@ -18,6 +18,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/jwt"
+	"google.golang.org/api/impersonate"
 	"google.golang.org/api/option"
 )
 
@@ -408,7 +409,7 @@ func DoesGCSBucketExist(gcsClient *storage.Client, config *RemoteStateConfigGCS)
 	bucket := gcsClient.Bucket(config.Bucket)
 
 	// TODO - the code below attempts to determine whether the storage bucket exists by making a making a number of API
-	// calls, then attemping to list the contents of the bucket. It was adapted from Google's own integration tests and
+	// calls, then attempting to list the contents of the bucket. It was adapted from Google's own integration tests and
 	// should be improved once the appropriate API call is added. For more info see:
 	// https://github.com/GoogleCloudPlatform/google-cloud-go/blob/de879f7be552d57556875b8aaa383bce9396cc8c/storage/integration_test.go#L1231
 	if _, err := bucket.Attrs(ctx); err != nil {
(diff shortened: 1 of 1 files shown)


yhakbar:
@richardjennings

Sorry we didn't get to this PR before it went stale. This has been addressed, as referenced in the linked issue.

PR #2664, opened 2023-08-14
Prevent parsing of excluded modules

### LINKED ISSUE #2167: An `excluded-dir` causes a failure on `run-all apply`
I ran a `run-all apply` from `$BASE_PATH/env_A` (referred to in the error log below). The `Required environment variable` wasn't set indeed, but **should/may an `excluded-dir` be able to cause this, or any, failure**? This `some_service` is a leaf in the dependencies graph.

> BASE_PATH=...
> EXCLUDED_DIR=...
> terragrunt run-all apply --terragrunt-exclude-dir $EXCLUDED_DIR                                                          
ERRO[0000] Error processing module at '$BASE_PATH/env_A/$EXCLUDED_DIR/terragrunt.hcl'. How this module was found: Terragrunt config file found in a subdirectory of $BASE_PATH/env_A/. Underlying error: $BASE_PATH/env_COMMON/some_service.hcl:25,23-31: Error in function call; Call to function "get_env" failed: EnvVarNotFound: Required environment variable TF_VAR_foo - not found. 

v0.37.1

levkohimins:
## Description

Prevent parsing of excluded modules. For example, these modules may have invalid syntax or incomplete configuration, causing terragrunt to fail when parsed.

## Related Issues

Fixes #2167

### DIFF
--- a/util/file.go
+++ b/util/file.go
@@ -67,7 +67,7 @@ func EnsureDirectory(path string) error {
 	return nil
 }
 
-// Return the canonical version of the given path, relative to the given base path. That is, if the given path is a
+// CanonicalPath returns the canonical version of the given path, relative to the given base path. That is, if the given path is a
 // relative path, assume it is relative to the given base path. A canonical path is an absolute path with all relative
 // components (e.g. "../") fully resolved, which makes it safe to compare paths as strings.
 func CanonicalPath(path string, basePath string) (string, error) {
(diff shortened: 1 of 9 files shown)


util/file.go
```diff
@@ -76,12 +76,51 @@ func CanonicalPath(path string, basePath string) (string, error) {
...
+
+	var paths []string
+
+	for _, globPath := range globPaths {
+		// Ensure globPath are absolute
+		if !filepath.IsAbs(globPath) {
+			globPath = filepath.Join(basePath, globPath)
+		}
+
+		matches, err := zglob.Glob(globPath)
```
MoonMoon1919, line 107: What is the desired behavior here if `err` is not nil?

MoonMoon1919 (commented):
One minor comment, otherwise LGTM

util/file.go
levkohimins: Just skip this path as it was before
https://github.com/gruntwork-io/terragrunt/blob/40e0d4b5db0fcfae9d9d36e3430995b7d142d800/configstack/module.go#L182-L187

Perhaps there is a reason for this behaviour, I don't know, but I would check the correctness of this path and return an error in cases of non-existence.

levkohimins: Do you think we should change this behaviour?

MoonMoon1919 (approved):

util/file.go
MoonMoon1919: Makes sense, no worries! Let's leave it for now

levkohimins:
Thanks for the review @MoonMoon1919!

levkohimins:
Resolved in [v0.50.1](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.50.1) release.

PR #2697, opened 2023-09-01
feat: disable terraform command validation

### LINKED ISSUE #2680: Terragrunt (0.50.6) does not handle Service Account impersonation in GCP, defaulting to the caller Service account
**Describe the bug**
Terragrunt doesn't target the right GCP Service account when using Service Account impersonation.

```
Remote state GCS bucket ProjectPrivBucket does not exist or you don't have permissions to access it. Would you like Terragrunt to create it? (y/n) time=2023-08-22T11:06:51Z level=error msg=Error: Error in function call
[...]
time=2023-08-22T11:06:51Z level=error msg=Call to function "read_terragrunt_config" failed: googleapi: Error 403: SA-Caller@ProjectCaller.iam.gserviceaccount.com does not have storage.buckets.get access to the Google Cloud Storage bucket. Permission 'storage.buckets.get' denied on resource (or it may not exist)., forbidden.
```

`SA-Caller` Service account should not be used, as there is an impersonated Service account in the environment/workspace (correctly used by `gcloud` commands).

**To Reproduce**
1. Create `SA-Caller` in `ProjectCaller`.
2. Create `SA-Priv` in `ProjectPriv`. Give it the Editor role.
3. Give `SA-Caller` the `roles/iam.serviceAccountTokenCreator` role on `SA-Priv`.
4. Auth into `ProjectCaller`, using the `SA-Caller` SA and the Github action `google-github-actions/auth@v1` setting `export_environment_variables` to `false`.
5. Run the following commands (in the Github action workflow):
```
# credentials_file_path is an output from `google-github-actions/auth@v1`
gcloud auth login --cred-file=${{ steps.auth.outputs.credentials_file_path }} --update-adc
gcloud config set auth/impersonate_service_account "SA-Priv@ProjectPriv.iam.gserviceaccount.com"
export GOOGLE_IMPERSONATE_SERVICE_ACCOUNT="SA-Priv@ProjectPriv.iam.gserviceaccount.com"
gcloud config set project ProjectPriv
          
gcloud alpha storage ls
echo "n" | terragrunt run-all init
```

**Expected behavior**
`gcloud alpha storage ls` and `terragrunt run-all init` should not fail.

The former should display a list of all the buckets in `ProjectPriv` (as it does) without requiring any extra gcloud flags, such as `--impersonate-service-account` or `--project`.

The latter should target `ProjectPriv` using the `SA-Priv` Service account (impersonated) and run `init` against all folders/stacks/components.

**Nice to have**
`gcloud config list` output:
```
[auth]
Your active configuration is: [default]
impersonate_service_account = SA-Priv@ProjectPriv.iam.gserviceaccount.com
[core]
account = SA-Caller@ProjectCaller.iam.gserviceaccount.com
disable_usage_reporting = True
project = ProjectPriv
```

**Versions**
- Terragrunt version: `0.50.6`
- Terraform version: `1.3.5`
- Environment details: Github actions runner, `ubuntu-latest`

**Additional context**
Looking at a recent PR (#2679) it seems like `impersonate_service_account` might be required as part of the `remote_state` block for the `gcs` backend. 

Is that the case? If so, Terragrunt is able to retrieve the right Google credentials from the workspace, how come it can't get the details about the impersonated SA, and they have to be explicitly passed to the remote state?

Is this a Terraform limitation? I can see [TF gcs backend docs](https://developer.hashicorp.com/terraform/language/settings/backends/gcs#impersonate_service_account) mention `impersonate_service_account` as well as 2 extra environment variables: `GOOGLE_BACKEND_IMPERSONATE_SERVICE_ACCOUNT` and `GOOGLE_IMPERSONATE_SERVICE_ACCOUNT`. But Terragrunt still fails even when setting the `GOOGLE_IMPERSONATE_SERVICE_ACCOUNT` env var, which is meant to be equivalent to setting `impersonate_service_account`?

levkohimins:
## Description

The PR adds a new flag `--terragrunt-disable-command-validation` to disable terraform command validation.

## Related issues

Fixes #2680 .

### DIFF
--- a/cli/commands/flags.go
+++ b/cli/commands/flags.go
@@ -39,6 +39,7 @@ const (
 	FlagNameTerragruntIncludeModulePrefix            = "terragrunt-include-module-prefix"
 	FlagNameTerragruntFailOnStateBucketCreation      = "terragrunt-fail-on-state-bucket-creation"
 	FlagNameTerragruntDisableBucketUpdate            = "terragrunt-disable-bucket-update"
+	FlagNameTerragruntDisableCommandValidation       = "terragrunt-disable-command-validation"
 
 	FlagNameHelp    = "help"
 	FlagNameVersion = "version"
@@ -230,6 +231,12 @@ func NewGlobalFlags(opts *options.TerragruntOptions) cli.Flags {
 			Destination: &opts.DisableBucketUpdate,
 			EnvVar:      "TERRAGRUNT_DISABLE_BUCKET_UPDATE",
 			Usage:       "When this flag is set Terragrunt will not update the remote state bucket.",
+		},
+		&cli.BoolFlag{
+			Name:        FlagNameTerragruntDisableCommandValidation,
+			Destination: &opts.DisableCommandValidation,
+			EnvVar:      "TERRAGRUNT_DISABLE_COMMAND_VALIDATION",
+			Usage:       "When this flag is set, Terragrunt will not validate the terraform command.",
 		},
 	}
 
--- a/cli/commands/terraform/command.go
+++ b/cli/commands/terraform/command.go
@@ -30,7 +30,7 @@ func action(opts *options.TerragruntOptions) func(ctx *cli.Context) error {
 			opts.CheckDependentModules = true
 		}
 
-		if !collections.ListContainsElement(nativeTerraformCommands, opts.TerraformCommand) {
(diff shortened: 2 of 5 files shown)


denis256 (approved):

levkohimins:
@denis256 thanks for the review!

PR #2704, opened 2023-09-07
Fixed searching for configs with non-default names #2704.

### LINKED ISSUE #2031: Terragrunt "run-all plan" do not respects TERRAGRUNT_CONFIG env variable
Hi, I've set `TERRAGRUNT_CONFIG=.terragrunt.hcl`. 
So my root and leaf folders have this .terragrunt.hcl file.
Everything works well when I'm trying to run terragrunt from child folder, but I get "Could not find any subfolders with Terragrunt configuration files" error when trying to run from `terragrunt run-all plan` from root. Renaming .terragrunt.hcl -> terragrunt.hcl fixes the issue, but I would prefer to use "dot" prefix if possible.

It seems like [this line](https://github.com/gruntwork-io/terragrunt/blob/f6b5661906b71dfe2f261a029ae3809f233d06a7/config/config.go#L582) is using default config name and not the parsed argument.

levkohimins:
## Description

* Fixed searching for config with explicitly specified `--terragrunt-config` and `run-all` command.
* Added the ability to explicitly specify the name of dependency configuration files, which may differ from the default name `terragrunt.conf`.

Fixes #2031 .

### DIFF
--- a/config/dependency.go
+++ b/config/dependency.go
@@ -177,7 +177,7 @@ func decodeAndRetrieveOutputs(
 
 // Convert the list of parsed Dependency blocks into a list of module dependencies. Each output block should
 // become a dependency of the current config, since that module has to be applied before we can read the output.
-func dependencyBlocksToModuleDependencies(decodedDependencyBlocks []Dependency) *ModuleDependencies {
+func dependencyBlocksToModuleDependencies(workingDir string, decodedDependencyBlocks []Dependency) *ModuleDependencies {
 	if len(decodedDependencyBlocks) == 0 {
 		return nil
 	}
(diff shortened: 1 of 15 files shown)


config/dependency.go
```diff
@@ -177,20 +177,27 @@ func decodeAndRetrieveOutputs(
...
 	paths := []string{}
 	for _, decodedDependencyBlock := range decodedDependencyBlocks {
 		configPath := decodedDependencyBlock.ConfigPath
-		if util.IsFile(configPath) && filepath.Base(configPath) == DefaultTerragruntConfigPath {
-			// dependencies system expects the directory containing the terragrunt.hcl file
-			configPath = filepath.Dir(configPath)
-		}
+
+		// absolutePath := configPath
+		// if !filepath.IsAbs(absolutePath) {
```
denis256, line 190: This part is required?

levkohimins: Oh, forgot to delete this code. Thanks. Fixed.

denis256 (approved):

levkohimins:
@denis256, thanks for the review!

levkohimins:
Resolved in [v0.50.15](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.50.15) release.

PR #2712, opened 2023-09-13
Add enabled property for dependency

### LINKED ISSUE #1483: Support Optional Dependencies
I'm currently working on a project where I have a number of different modules which make up our web application stack. We have many of these stacks defined, all of which pull from a set of top-level .hcl terragrunt files (one for each module). Some of these modules are optional for a given stack, in my case for example whether or not to enable AzureAD single sign-on is an optional, and is defined as a separate module. However, I have a module which is dependant on the outputs of the AzureAD module when present. Therefore, having an option within the dependency block to simply enable or disable a given dependency would be really helpful. Right now even if I set `skip_outputs` to `true` terragrunt still expects the module directory (as pointed to by the `config_path` attribute) to exist and contain a `terragrunt.hcl` file.

denis256:
## Description

Included changes:
* added property `enabled` to `dependency`
* added logic to skip dependency if `enabled = false`
* added tests to track that flag is working

Fixes #1483.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added `enabled` property on `dependency` block, once set to `false` - dependency will be skipped.


MoonMoon1919 (approved):

jlepere-everlaw:
Should we inject `mock_outputs` into the dependency variable's outputs if `enabled = false`? Otherwise, we'd have to add a ternary for each dependency output usage based on the `enabled` properties value.

denis256:
Hello,
will be helpful to raise a separate enhancement issue to see how this can be implemented

PR #2719, opened 2023-09-18
Removes trailing slash appended to the end of `get_path_to_repo_root()`

### LINKED ISSUE #2158: `get_path_to_repo_root()` includes trailing slash unlike other functions
The path returned by `get_path_to_repo_root()` includes a trailing `/` where it looks like all other path functions do not.  When building a `source` based on this, it causes `//` problems.

For instance, using the normal expression like this:
```
source = "${get_path_to_repo_root()}/modules//my-thing"
```
it actually resolves to the path `../..//modules//my-thing`.  This causes the entire repo to get copied as the source.  So, now I have to write my expression as:
```
source = "${get_path_to_repo_root()}modules//my-thing"
```
to get the expected path of `../../modules//my-thing`.

Correcting this will likely cause problems for folks.  Either they lose the `//` they currently expect or their existing expression (like my second example above) is no longer a valid path.

levkohimins:
## Description

This PR removes the trailing slash from the path returned by the `get_path_to_repo_root()` built-in function.

Fixes #2158.

## Release Notes
* Removes trailing slash appended to the end of `get_path_to_repo_root()`

### Migration Guide

* Migration:
  * `source = "${get_path_to_repo_root()}/modules` -> `source = "${get_path_to_repo_root()}//modules`
  * `source = "${get_path_to_repo_root()}modules` -> `source = "${get_path_to_repo_root()}/modules`


denis256 (approved):

levkohimins:
@denis256 Thanks for the review!

PR #2721, opened 2023-09-18
S3 bucket errors handling

### LINKED ISSUE #2267: S3 backend permissions: Missing debug output, silently ignored errors

denis256:
## Description

Included changes:
* added returning of errors in case of S3 bucket configuration errors
* added logging of executed function

Fixes #2267.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

### DIFF
--- a/remote/remote_state_s3.go
+++ b/remote/remote_state_s3.go
@@ -372,40 +372,41 @@ func validateS3Config(extendedConfig *ExtendedRemoteStateConfigS3, terragruntOpt
 // If the bucket specified in the given config doesn't already exist, prompt the user to create it, and if the user
 // confirms, create the bucket and enable versioning for it.
 func createS3BucketIfNecessary(s3Client *s3.S3, config *ExtendedRemoteStateConfigS3, terragruntOptions *options.TerragruntOptions) error {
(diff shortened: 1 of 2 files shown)


levkohimins (approved):
LGTM!

PR #2740, opened 2023-10-02
Update render-json to include list of dependent modules

### LINKED ISSUE #2198: Add "reverse dependency" information to render-json
Currently, when you run the `render-json` command for module `foo`, it includes `dependency` information, showing you other modules that `foo` depends on: e.g., if `foo` contains a `dependency` block pointing to `bar`, then `bar` is in the `dependency` output of `render-json`. It would be useful to add "reverse dependency" information, showing what other modules depend on `foo`: e.g., if `baz` contains a `dependency` block pointing to `foo`, then `baz` shows up in the `reverse-dependency` output of `render-json` for `foo`. This way, any tool introspecting Terragrunt configs using the `render-json` command would be able to see the dependency graph in both directions.

I believe we already know how to fetch this information as per https://github.com/gruntwork-io/terragrunt/pull/1823, so I would guess it's just a matter of exposing it in `render-json`?

denis256:
## Description

Update `render-json` to include new field `dependent_modules` which includes list of paths to dependent modules.

Example output:

```
# test/fixture-destroy-warning

.
├── app-v1
│   ├── main.tf
│   └── terragrunt.hcl
├── app-v2
│   ├── main.tf
│   └── terragrunt.hcl
├── terragrunt.hcl
└── vpc
    ├── main.tf
    └── terragrunt.hcl

# cd vpc && terragrunt render-json
...
  "dependent_modules": [
    "/home/denis/projects/gruntwork/terragrunt/test/fixture-destroy-warning/app-v1",
    "/home/denis/projects/gruntwork/terragrunt/test/fixture-destroy-warning/app-v2"
  ],
...

# cd vpc && terragrunt render-json --with-metadata
...
  "dependent_modules": {
    "metadata": {
      "found_in_file": "/home/denis/projects/gruntwork/terragrunt/test/fixture-destroy-warning/vpc/terragrunt.hcl"
    },
    "value": [
      "/home/denis/projects/gruntwork/terragrunt/test/fixture-destroy-warning/app-v2",
      "/home/denis/projects/gruntwork/terragrunt/test/fixture-destroy-warning/app-v1"
    ]
  },
...


```

Fixes #2198.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added `dependent_modules` field to `render-json` output with list of paths to dependent modules.


configstack/module.go
```diff
@@ -536,9 +537,16 @@ func FindWhereWorkingDirIsIncluded(terragruntOptions *options.TerragruntOptions,
 	var matchedModulesMap = make(map[string]*TerraformModule)
 
 	gitTopLevelDir, err := shell.GitTopLevelDir(terragruntOptions, terragruntOptions.WorkingDir)
+	useIncludes := false
```
maunzCache, line 540: If i read the code correctly, there is never a case where false is required as value. Instead of creating the if condition we could maybe rearrange the code to be more straightforward.

Line 541 only runs when line 539 doesn't produce an error. The error is not handled until bit later so you might want to do that instead and simply set `useIncludes := true`. Call `buildDirList` safely after and only override with `false` when no errors are present.

Actually, after reading the line, just put 549-557 into a function and call it. `useIncludes` might not be required at all.

configstack/module.go
```diff
@@ -553,24 +561,22 @@ func FindWhereWorkingDirIsIncluded(terragruntOptions *options.TerragruntOptions,
...
+		cfgOptions.Logger.Logger.AddHook(hook)
 
 		stack, err := FindStackInSubfolders(cfgOptions, terragruntConfig)
 		if err != nil {
-			// loggign error as debug since in some cases stack building may fail because parent files can be designed
+			// log error as debug since in some cases stack building may fail because parent files can be designed
 			// to work with relative paths from downstream modules
 			terragruntOptions.Logger.Debugf("Failed to build module stack %v", err)
-			return nil
+			continue
```
maunzCache, line 579: Due to this change, stack might be nil in 582

maunzCache (commented):
I'd suggest to do a minor refactoring of the mentioned lines. Functionality might be fine without it.

configstack/module.go
denis256: on line 579 execution will go to next element in for loop, will not be invoked `stack.Modules`

MoonMoon1919 (approved):

configstack/module.go
maunzCache: You are correct. The for loop was hidden when i reviewed this section.

PR #2738, opened 2023-10-02
Add opentofu notice

MoonMoon1919:
## Description

Adds notice that Terragrunt supports OpenTofu as of TG version v0.52.0.

image

## TODOs

Read the [Gruntwork contribution guidelines](https://gruntwork.notion.site/Gruntwork-Coding-Methodology-02fdcd6e4b004e818553684760bf691e).

- [ ] Update the docs.
- [ ] Run the relevant tests successfully, including pre-commit checks.
- [ ] Ensure any 3rd party code adheres with our [license policy](https://www.notion.so/gruntwork/Gruntwork-licenses-and-open-source-usage-policy-f7dece1f780341c7b69c1763f22b1378) or delete this line if its not applicable.
- [ ] Include release notes. If this PR is backward incompatible, include a migration guide.

## Release Notes (draft)

Added / Removed / Updated [X].

PR #2745, opened 2023-10-04
Add OpenTofu support

### LINKED ISSUE #2690: Add OpenTofu support
[As recently announced](https://opentofu.org/blog/the-opentofu-fork-is-now-available), the OpenTofu fork should drop soon. We should make sure that Terragrunt supports it out of the box. 

Here's an idea for an initial implementation: anywhere Terragrunt calls out to the `terraform` binary, we should update it to check for the `terraform` binary first; if that's not found, then try the `tofu` binary; and if that's not found, show an error (as we do today). Whatever we pick, we should mention in the debug log.

In theory, you can already use OpenTofu by using the [`terraform_binary`  config](https://terragrunt.gruntwork.io/docs/reference/config-blocks-and-attributes/#terraform_binary) or `--terragrunt-tfpath` CLI flag or `TERRAGRUNT_TFPATH` env var, but those all require extra steps, whereas I'd like OpenTofu to "just work" with Terragrunt. 

**Additional context**
OpenTofu should work as a "drop in" replacement for Terraform, so in theory, no other changes should be necessary. However, we should run our tests to make sure once the `tofu` binary is available for testing. One place where OpenTofu will differ is `stdout` and `stderr` output will say OpenTofu instead of Terraform; if Terragrunt is parsing that log output anywhere, we'll need to update the code accordingly.

This is just the initial implementation in the sense that it should be backward compatible. In the future, we can reverse the lookup order, and perhaps even drop support for `terraform`, but those would be significant breaking changes (especially the latter), so we'd want to see how things go with OpenTofu and what the community wants first.

**Note on registry usage**
HashiCorp recently did an update (one that was totally unannounced, BTW) to its [Registry terms](https://registry.terraform.io/terms) to say:

> You may download providers, modules, policy libraries and/or other Services or Content from this website solely for use with, or in support of, HashiCorp Terraform.

Terragrunt uses go-getter to download modules specified in the `source` URL, and it's possible that URL points to the Terraform Registry. For now, Terragrunt only works with Terraform, so that is still inline with the terms. If we add support for OpenTofu, then in theory, anyone using Terragrunt with OpenTofu might be violating the registry terms.

Possible solutions:

1. Cheapest/fastest: If we detect a registry `source` URL and `tofu` as the binary, give the user a warning.
2. Better solution: If we detect a registry `source` URL and `tofu` as the binary, automatically use OpenTofu's registry (coming soon) instead of HashiCorp's.

denis256:
## Description

Included changes:
* added support for wrapping `tofu` command: if `terraform` command is not available, will be used `tofu`
* updated version parsing to handle OpenTofu output
* separated integration tests to run separated the same test suits against `terraform` and `tofu`(terraform binary is removed before starting tofu tests)
* added field in `TerragruntOptions` to track used Terraform implementation
* updated logs to reflect used implementation
* usage of OpenTofu in Windows tests

image

Fixes #2690.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added support for OpenTofu in Terragrunt, by default, will be wrapped `terraform` command with a fallback to `tofu`.

### Migration Guide

Install OpenTofu cli tool https://github.com/opentofu/opentofu

### DIFF
--- a/docs/_docs/01_getting-started/supported-versions.md
+++ b/docs/_docs/01_getting-started/supported-versions.md
@@ -35,7 +35,9 @@ The officially supported versions are:
 | 0.12.x            | [0.19.0](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.19.0) - [0.24.4](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.24.4) |
 | 0.11.x            | [0.14.0](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.14.0) - [0.18.7](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.18.7) |
 
-With the HashiCorp Terraform BSL license change, will be supported Terraform versions `<= 1.5.7`, for newer features, please use OpenTofu.
+Due to the HashiCorp BSL license change, we are restricting Terraform to the latest open source version (`1.5.7`).
+Going forward we will be adding support for OpenTofu. To learn more, see the official [OpenTofu website](https://opentofu.org/)
+and [project status](https://github.com/opentofu/opentofu/blob/main/WEEKLY_UPDATES.md).
 
 However, note that these are the versions that are officially tested in the CI process. In practice, the version
 compatibility is more relaxed than documented above. For example, we've found that Terraform 0.13 works with any version
--- a/docs/_docs/01_getting-started/supported-versions.md
+++ b/docs/_docs/01_getting-started/supported-versions.md
@@ -35,10 +35,6 @@ The officially supported versions are:
 | 0.12.x            | [0.19.0](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.19.0) - [0.24.4](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.24.4) |
 | 0.11.x            | [0.14.0](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.14.0) - [0.18.7](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.18.7) |
 
-Due to the HashiCorp BSL license change, we are restricting Terraform to the latest open source version (`1.5.7`).
-Going forward we will be adding support for OpenTofu. To learn more, see the official [OpenTofu website](https://opentofu.org/)
-and [project status](https://github.com/opentofu/opentofu/blob/main/WEEKLY_UPDATES.md).
-
 However, note that these are the versions that are officially tested in the CI process. In practice, the version
 compatibility is more relaxed than documented above. For example, we've found that Terraform 0.13 works with any version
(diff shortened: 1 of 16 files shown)


docs/_docs/01_getting-started/supported-versions.md
```diff
@@ -27,6 +35,9 @@ The officially supported versions are:
 | 0.12.x            | [0.19.0](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.19.0) - [0.24.4](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.24.4) |
 | 0.11.x            | [0.14.0](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.14.0) - [0.18.7](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.18.7) |
 
+Due to the HashiCorp BSL license change, we are restricting Terraform to the latest open source version (`1.5.7`).
+Going forward we will be adding support for OpenTofu. To learn more, see the official [OpenTofu website](https://opentofu.org/)
+and [project status](https://github.com/opentofu/opentofu/blob/main/WEEKLY_UPDATES.md).
```
brikis98, line 40: Terragrunt should keep working with newer Terraform versions, even BSL ones. It's an open source wrapper, and shouldn't be an issue with BSL.

options/options.go
```diff
@@ -265,7 +281,7 @@ func MergeIAMRoleOptions(target IAMRoleOptions, source IAMRoleOptions) IAMRoleOp
 // Create a new TerragruntOptions object with reasonable defaults for real usage
 func NewTerragruntOptions() *TerragruntOptions {
 	return &TerragruntOptions{
-		TerraformPath:                  TerraformDefaultPath,
+		TerraformPath:                  TofuPath,
```
brikis98, line 284: We should probably stick with `terraform` as the default and `tofu` as the fallback. Otherwise, it's a backward incompatible change.

brikis98 (commented):
Thanks for putting this together! A few high level thoughts:

1. The Terragrunt docs site should mention we support both OpenTofu and Terraform. Probably somewhere right on the home page and a dedicated page somewhere in the docs.
2. Are we now testing with _only_ OpenTofu? If so, I think we want to test with both Terraform and OpenTofu, as we will have users from both.

docs/_docs/01_getting-started/supported-versions.md
denis256: Will update implementation to work with BSL Terraform

brikis98 (approved):
Code changes LGTM, thanks for getting this turned around so quickly after the alpha dropped!

We should still update the Terragrunt website home page to make it clear Terragrunt works with both Terraform and OpenTofu... But that can be a follow up PR.

denis256:
I think it is handled in https://github.com/gruntwork-io/terragrunt/pull/2738

PR #2752, opened 2023-10-11
`sops_decrypt_file()`: resolve path argument relative to `terragrunt.hcl`

sybereal:
## Description

Update path resolution logic in `sops_decrypt_file()` to resolve relative path arguments relative to the file containing the call to `sops_decrypt_file()`, rather than relative to the working directory of the Terragrunt process.

This change should be invisible for the common case of running Terragrunt directly inside the directory containing `terragrunt.hcl`. However, it enables intuitive use of `terragrunt run-all` with `sops_decrypt_file()`, ensuring configurations behave the same, no matter if they're evaluated directly or under `run-all`.

An additional test case for `sops_decrypt_file()` was added that validates its behavior in combination with `run-all`.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Updated `sops_decrypt_file()` path resolution to be relative to `terragrunt.hcl` instead of Terragrunt's working directory.

### Migration Guide

If you were relying on the behavior of `terragrunt run-all` on a set of modules using `sops_decrypt_file()` using a file in the current working directory, such as a shared secret, you can wrap the argument to `sops_decrypt_file()` in a call to `find_in_parent_folders()`, as shown in the documentation for `sops_decrypt_file()`.

### DIFF
--- a/config/config_helpers.go
+++ b/config/config_helpers.go
@@ -746,7 +746,7 @@ func sopsDecryptFile(ctx *ParsingContext, params []string) (string, error) {
 	if err != nil {
 		return "", errors.WithStackTrace(err)
 	}
-	canonicalSourceFile, err := util.CanonicalPath(sourceFile, ctx.TerragruntOptions.WorkingDir)
+	canonicalSourceFile, err := util.CanonicalPath(sourceFile, filepath.Dir(ctx.TerragruntOptions.TerragruntConfigPath))
 	if err != nil {
 		return "", errors.WithStackTrace(err)
 	}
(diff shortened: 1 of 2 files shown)


sybereal:
My bad. I just copied the existing test case that was there at the time I made the change without thinking. Should be fine now, I hope.

yhakbar:
Hey @sybereal

I've run the tests for this functionality, and they still aren't passing. Are you able to run the tests locally?

This PR has been open for a long time, and while this functionality seems very convenient, there are workarounds via functions like [get_terragrunt_dir](https://terragrunt.gruntwork.io/docs/reference/built-in-functions/#get_terragrunt_dir).

We can leave the PR open a little bit longer, but if you could ensure that tests are passing before re-requesting review from me, that would be great.

sybereal:
Could you tell me which tests are failing that aren't failing on master? I can't see the CircleCI logs even after creating an account, and locally I can't run the full suite because I do not have an AWS account set up. I specifically ran only the sops tests with `go test ./... -run '.*Sops.*'`, and those were fine, but ofc that way I can't tell if my change broke anything else.

yhakbar (approved):

yhakbar:
Running the tests after rebase worked!

PR #2763, opened 2023-10-20
Fix output-module-groups failing if terragrunt.hcl file not found in working dir

### LINKED ISSUE #2685: `terragrunt output-module-groups` fails if not run from a Terragrunt module
**Describe the bug**
The description of `output-module-groups` says:

> This will recursively search the current working directory for any folders that contain Terragrunt modules and build the dependency graph based on dependency and dependencies blocks.

However, if I run `terragrunt output-module-groups` from the root of my repo, the command fails. I do not have a terragrunt.hcl at root, as all Terragrunt modules are organized within my repo. Here's an example of my layout:

```
- root
  - aws
    - aws.hcl (shared config for all AWS Terragrunt modules)
    - commercial
      - account1
        - terragrunt.hcl
        - test.tfvars
      - account2
        - terragrunt.hcl
        - test.tfvars
        - submodule
          - terragrunt.hcl
          - test.tfvars
  - azure
    - azure.hcl (shared config for all Azure Terragrunt modules)
    - ...
```

Here's the error I'm getting:
```
$ terragrunt output-module-groups
ERRO[0000] Error reading file at path /terragrunt_test/terragrunt.hcl: open /terragrunt_test/terragrunt.hcl: no such file or directory
```

**To Reproduce**
Attempt to run `terragrunt output-module-groups` from a a directory that contains Terragrunt modules in subfolders, but is not a Terragrunt module itself.

**Expected behavior**
I would expect this command to recursively search the current directory for all Terragrunt modules and build the dependency graph, even if there is no terragrunt.hcl in the current dir.

**Workaround**
As a workaround, you can create an empty terragrunt.hcl in the root directory to get it to recursively search for all Terragrunt modules.

**Versions**
- Terragrunt version: v0.50.6
- Terraform version: v1.5.4
- Environment details (Ubuntu 20.04, Windows 10, etc.): Rocky Linux 8

evsl:
## Description

Fixes #2685

After a quick look, it seems like the changes introduced from 73d2c83 added (seemingly unnecessary?) code in the Run func for `output-module-groups` which treated it as if it were a terraform command, which would of course need to find the root configuration in order to build the temp dir in which to run terraform. 

I updated the relevant test to move the `terragrunt.hcl` file out of the root of the working directory in the context of the test, which indeed caused it to fail as described in the issue. Removing the terraform invocation from the action matches the documented behavior and all relevant tests pass.

This is my first contribution to this repo and I've only dug through enough code to identify the root cause at a high level, so please let me know if I've misunderstood the intent of the code I removed.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Updated `output-module-groups` to no longer require a `terragrunt.hcl` file in the working directory.


denis256:
Hello,
It would be helpful to add an integration test that will track whether this issue will not occur again.

Examples in: https://github.com/gruntwork-io/terragrunt/tree/master/test

evsl:
> Hello, It would be helpful to add an integration test that will track whether this issue will not occur again.
> 
> Examples in: https://github.com/gruntwork-io/terragrunt/tree/master/test

There was already an integration test around this functionality, but it seemed like it was only passing because the test fixture itself already had a `terragrunt.hcl` file in the root (which I simply removed -- I'm not familiar with the reasoning behind setting up that fixture as such to test this func). Given that, is there anything else necessary?

Additionally, there are changes over in #2769 that builds additional functionality on top of this function that looks interesting, and @bt-macole has suggested cherry picking these changes into that, so perhaps we can continue improving upon this function in that PR.

evsl:
Closed to track in #2769

PR #2772, opened 2023-10-30
fix: output the default prompt answer for better UX

meysam81:


yhakbar:
@meysam81 

I'm not sure why this is better UX. Closing as stale.

In the future, please create an issue to document the problem fully, then include a descriptive PR so that it's clear why a change is beneficial.

Issue #2814, opened 2023-11-29
Bad error message for 404s with source URLs


brikis98:
**Describe the bug**
If you have a typo in your `source` URL, Terragrunt gives a very unhelpful error message.

**To Reproduce**
Create a `terragrunt.hcl` with a `source` URL that doesn't work:

```hcl
terraform {
  source = "https://github.com/totallyfakedoesnotexist/notreal.git//foo?ref=v1.2.3"
}
```

Run `terragrunt plan` and you'll see:

```
$ terragrunt plan
ERRO[0000] 1 error occurred:
	* bad response code: 404
```

**Expected behavior**
There should be a little more context to help the user understand where the 404 is coming from!

E.g.,:

```
$ terragrunt plan
ERRO[0000] 1 error occurred downloading source URL "https://github.com/totallyfakedoesnotexist/notreal.git//foo?ref=v1.2.3":
	* bad response code: 404
```

**Versions**
- Terragrunt version: 0.48.1

**Additional context**
If you set the log level to debug, the error is clearer:

```
$ terragrunt plan --terragrunt-log-level debug
(... truncated ...)
DEBU[0000] Downloading Terraform configurations from https://github.com/totallyfakedoesnotexist/notreal.git?ref=v1.2.3 into /tmp/example/.terragrunt-cache/lS-k1MxVrT9JmxvyztQGGooKYDc/dVXZU_WvzSwc9Ox3-4qP7a1G7bo
ERRO[0001] 1 error occurred:
	* bad response code: 404
```

But ideally, you shouldn't need a higher log level to understand such a basic error.

PR #2820, opened 2023-12-01
feat: update the log level from debug to info

gqrlt1207:
To resolve the issue #2814 
https://github.com/gruntwork-io/terragrunt/issues/2814

Update the log level from Debug to Info


gqrlt1207:
@denis256 could you please take a look when you get a moment? it's a tiny change. thanks.

denis256:
Hi,
will be helpful to add an integration test which will track that error message contains required information

gqrlt1207:
> Hi, will be helpful to add an integration test which will track that error message contains required information

thanks, will do

gqrlt1207:
> Hi, will be helpful to add an integration test which will track that error message contains required information

added integration test, please review, thanks.

cli/commands/terraform/download_source.go
```diff
@@ -102,7 +103,9 @@ func downloadTerraformSourceIfNecessary(terraformSource *terraform.Source, terra
 	})
 
 	if downloadErr != nil {
-		return downloadErr
+		download_info := fmt.Errorf("Downloading Terraform source from %s", terraformSource.CanonicalSourceURL.String())
```
denis256, line 106: * why can't be created specific error struct like `WorkingDirNotFound`?
* use camel case instead of snake case 
* why can't be used `go-multierror` to combine multiple errors?

denis256: Linter suggestion "Error string should not be capitalized or end with punctuation mark "

gqrlt1207: thanks, will update it accordingly.

gqrlt1207: if we use go-multierror, we got something as below:
image
Actually, there is only 1 error.
If we keep the current code, we get something as below:
image

gqrlt1207: For creating a struct, since we only need the 'Url' string, in my opinion, we do not need a struct in this case. just my opinion. please review and advise. thanks,

gqrlt1207: @denis256 updated as you suggested. added a struct 'DownloadingTerraformSourceNotFound', removed the snake case, and updated the test accordingly. please review and advise. thanks.

denis256: I think error name `DownloadingTerraformSourceNotFound` is too specific ( "NotFound" part) - downloadErr can be returned because of 404/500 etc...

denis256:
Will be helpful also to have fixture/integration test like https://github.com/gruntwork-io/terragrunt/tree/master/test  which will track end to end printed error.

cli/commands/terraform/download_source.go
gqrlt1207: updated error name 'DownloadingTerraformSourceNotFound'  to 'DownloadingTerraformSourceErr'.

gqrlt1207:
> Will be helpful to have a fixture and integration test like https://github.com/gruntwork-io/terragrunt/tree/master/test which will track end-to-end printed errors.

added integration test as suggested, please review. thanks

test/integration_download_test.go
```diff
@@ -144,6 +145,17 @@ func TestRemoteDownload(t *testing.T) {
 	runTerragrunt(t, fmt.Sprintf("terragrunt apply -auto-approve --terragrunt-non-interactive --terragrunt-working-dir %s", testFixtureRemoteDownloadPath))
 }
 
+func TestInvalidRemoteDownload(t *testing.T) {
```
denis256, line 148:
```
time=2023-12-15T16:57:27Z level=debug msg=Enabling enforced TLS access for S3 bucket terragrunt-test-bucket-0n6bas prefix=[/tmp/terragrunt-parent-child-test516286081/stage/my-app] 
=== NAME  TestInvalidRemoteDownload
    integration_test.go:4097: Failed to run Terragrunt command 'terragrunt apply -auto-approve --terragrunt-non-interactive --terragrunt-working-dir fixture-download/remote-invalid' due to error: downloading source url https://github.com/totallyfakedoesnotexist/notreal.git?ref=v1.2.3
        1 error occurred:
        	* bad response code: 404
        
        
        
        Stdout: (see log output above)
        
        Stderr: (see log output above)
--- FAIL: TestInvalidRemoteDownload (2.28s)

```

test/integration_download_test.go
```diff
@@ -144,6 +145,17 @@ func TestRemoteDownload(t *testing.T) {
...
+	t.Parallel()
+
+	cleanupTerraformFolder(t, testFixtureInvalidRemoteDownloadPath)
+
+	runTerragrunt(t, fmt.Sprintf("terragrunt apply -auto-approve --terragrunt-non-interactive --terragrunt-working-dir %s", testFixtureInvalidRemoteDownloadPath))
```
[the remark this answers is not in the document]
gqrlt1207: okay, will add it.

test/integration_download_test.go
gqrlt1207: @denis256 thanks for taking the time to review. I updated the test section as suggested. please review. many thanks

test/integration_download_test.go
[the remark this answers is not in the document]
gqrlt1207: updated by adding assert to check that error, please review. thanks.

denis256 (approved):

PR #2821, opened 2023-12-01
[WIP] Terragrunt scaffold 

denis256:
## Description

Fixes #000.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)


cli/commands/scaffold/action.go
```diff
@@ -0,0 +1,351 @@
...
+	sourceUrlTypeVar    = "SourceUrlType"
+	sourceGitSshUserVar = "SourceGitSshUser"
+	refVar              = "Ref"
+	// refParam - ?ref param from url
+	refParam = "ref"
+
+	moduleUrlPattern = `git::([^:]+)://([^/]+)(/.*)`
+	moduleUrlParts   = 4
+
+	defaultBoilerplateConfig = `
```
brikis98, line 44: - I think the  more "idiomatic" way to use boilerplate is to have these as files on disk in a folder (i.e., `boilerplate.yml`, `terragrunt.hcl`, etc), and to run `boilerplate` against that folder.  
- This "default template" is just our starting point for the `scaffold` command. We'll want other built-in templates for Terragrunt, such as scaffolding out a new module (e.g., `main.tf`, `variables.tf`, etc), scaffolding out a wrapper module, scaffolding out an entire `infra-live` repo, and so on. So all of these templates should live in folders in a GitHub repo: e.g., a `boilerplate` folder in the Terragrunt repo, with this default template under `boilerplate/default-terragrunt-module`. 
- The `--template-url` arg of `boilerplate` already supports URLs, so you could point it to the TG repo URL (which is open source) and it should "just work."

That said, it's OK for this first version to leave the default template "embedded" in the binary like this. It also makes local testing easier (while iterating on the template), which is something we'd have to solve for with the approach above.

So, for now, could you file a GH issue for this, and we can come back to it a bit later?

cli/commands/scaffold/action.go
```diff
@@ -0,0 +1,351 @@
...
+	var dirsToClean []string
+	defer func() {
+		for _, dir := range dirsToClean {
+			if err := os.RemoveAll(dir); err != nil {
+				opts.Logger.Warnf("Failed to clean up dir %s: %v", dir, err)
+			}
+		}
+	}()
+
+	// scaffold only in empty directories
```
brikis98, line 93: Why only in empty dirs? I think scaffolding into an existing dir is OK?

I suppose we could have a warning / confirmation prompt (unless `--non-interactive` is set) telling the user the existing contents may be overwritten, but even that is a bit of a nice to have...

cli/commands/scaffold/action.go
```diff
@@ -0,0 +1,351 @@
...
+			return errors.WithStackTrace(err)
+		}
+		dirsToClean = append(dirsToClean, templateDir)
+
+		// downloading template
+		opts.Logger.Infof("Using template from %s", templateUrl)
+		if err := getter.GetAny(templateDir, templateUrl); err != nil {
+			return errors.WithStackTrace(err)
+		}
+	}
```
brikis98, line 171: Can you not pass the template URL directly to `--template-url` of boilerplate? It should know how to download templates...

cli/commands/scaffold/action.go
```diff
@@ -0,0 +1,351 @@
...
+  # Type: {{ .Type }}
+  # {{ .Name }} = {{ .DefaultValue }}
+  {{ end }}
+}
+`
+)
+
+var moduleUrlRegex = regexp.MustCompile(moduleUrlPattern)
+
+func Run(opts *options.TerragruntOptions) error {
```
brikis98, line 79: NIT: this is a really long function that does many different things:

- Prepare temp folder
- Parse source URL and download
- Parse template URL and download
- Parse variables in module
- Render template, passing it the variables

Perhaps break each of these large behaviors into separate, named functions, so this top-level one is a bit easier to scan at a glance?

cli/commands/scaffold/action.go
```diff
@@ -0,0 +1,351 @@
...
+	if err != nil {
+		return errors.WithStackTrace(err)
+	}
+
+	// separate variables that require value and with default value
+	var requiredVariables []*config.ParsedVariable
+	var optionalVariables []*config.ParsedVariable
+
+	for _, value := range inputs {
+		if value.DefaultValue == "" {
```
brikis98, line 189: Does this handle string variables where the default value is an empty string (as opposed to no default value at all)?

cli/commands/scaffold/action.go
```diff
@@ -0,0 +1,351 @@
...
+		if err := os.WriteFile(util.JoinPath(boilerplateDir, "boilerplate.yml"), []byte(defaultBoilerplateConfig), 0644); err != nil {
+			return errors.WithStackTrace(err)
+		}
+	}
+
+	// add additional variables
+	vars["requiredVariables"] = requiredVariables
+	vars["optionalVariables"] = optionalVariables
+
+	vars["sourceUrl"] = moduleUrl
```
brikis98, line 226: We need to treat this and document it as an API. 

That is, whatever variables Terragrunt makes available to the boilerplate template need to be well-defined and captured somewhere as an explicit public API, and one for which we'll want to maintain backward compatibility (and document breaking changes & migration guides if necessary). Consider defining these as a struct or somewhere centralized, with appropriate comments, so it's clear this is an API that we need to treat carefully.

cli/commands/scaffold/action.go
```diff
@@ -0,0 +1,351 @@
...
+
+	opts.Logger.Infof("Running boilerplate generation to %s", opts.WorkingDir)
+	boilerplateOpts := &boilerplate_options.BoilerplateOptions{
+		TemplateFolder:  boilerplateDir,
+		OutputFolder:    opts.WorkingDir,
+		OnMissingKey:    boilerplate_options.DefaultMissingKeyAction,
+		OnMissingConfig: boilerplate_options.DefaultMissingConfigAction,
+		Vars:            vars,
+
+		NonInteractive: true,
```
brikis98, line 236: Why disable interactive mode? I think by default, if you didn't provide vars, Terragrunt should prompt you for any boilerplate inputs... But if you pass the `--terragrunt-non-interactive` flag, then it can set non-interactive for boilerplate too.

cli/commands/scaffold/action.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+
+	// add additional variables
+	vars["requiredVariables"] = requiredVariables
+	vars["optionalVariables"] = optionalVariables
+
+	vars["sourceUrl"] = moduleUrl
+
+	opts.Logger.Infof("Running boilerplate generation to %s", opts.WorkingDir)
+	boilerplateOpts := &boilerplate_options.BoilerplateOptions{
```
brikis98, line 229: IMPORTANT: please disable scripts and the `shell` helper by default. Those are a security risk and we should only turn them on if requested, via explicit flags. But those flags can be added in a separate PR later; here, we should ensure the default is safe.

cli/commands/scaffold/action.go
```diff
@@ -0,0 +1,351 @@
...
+	if err := hclfmt.Run(opts); err != nil {
+		return errors.WithStackTrace(err)
+	}
+
+	opts.Logger.Info("Scaffolding completed")
+
+	return nil
+}
+
+// rewriteModuleUrl rewrites module url to git ssh if required
```
brikis98, line 253: NIT: an example before & after in comments would help clarify what this method does.

cli/commands/scaffold/action.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+
+	parsedModuleUrl, err := terraform.ToSourceUrl(updatedModuleUrl, opts.WorkingDir)
+	if err != nil {
+		return nil, errors.WithStackTrace(err)
+	}
+	return parsedModuleUrl, nil
+}
+
+// rewriteTemplateUrl rewrites template url with reference to tag
```
brikis98, line 280: NIT: an example before & after in comments would help clarify what this method does.

cli/commands/scaffold/action.go
```diff
@@ -0,0 +1,351 @@
...
+	var templateParams = updatedTemplateUrl.Query()
+	ref := templateParams.Get(refParam)
+	if ref == "" {
+		rootSourceUrl, _, err := terraform.SplitSourceUrl(updatedTemplateUrl, opts.Logger)
+		if err != nil {
+			return nil, errors.WithStackTrace(err)
+		}
+		tag, err := shell.GitLastReleaseTag(opts, rootSourceUrl)
+		if err != nil || tag == "" {
+			opts.Logger.Warnf("Failed to find last release tag templae %s", rootSourceUrl)
```
brikis98, line 292:
```suggestion
			opts.Logger.Warnf("Failed to find last release tag for URL %s, so will not add a ref param to the URL", rootSourceUrl)
```

brikis98 (commented):
This is looking great! I left a bunch of comments.

A couple high thoughts:

1. We need to update the docs with this new command and the "API" for creating your own boilerplate templates for it.
2. Is there an integration test that runs `scaffold` end to end?

denis256:
* Will add documentation and examples of how to create boilerplate templates
* Integration end-to-end tests for boilerplate are located in `test/integration_scaffold_test.go` - for sure can be added more use cases

cli/commands/scaffold/action.go
denis256: I can try to pass template url but we will lose function for "detection of last release tag", not sure how important it is for template url

shell/run_shell_cmd.go
```diff
@@ -295,6 +308,73 @@ func GitTopLevelDir(terragruntOptions *options.TerragruntOptions, path string) (
...
+	stderr := bytes.Buffer{}
+	gitOpts, err := options.NewTerragruntOptionsWithConfigPath(opts.WorkingDir)
+	if err != nil {
+		return nil, err
+	}
+	gitOpts.Env = opts.Env
+	gitOpts.Writer = &stdout
+	gitOpts.ErrWriter = &stderr
+
+	output, err := RunShellCommandWithOutput(opts, opts.WorkingDir, true, false, "git", "ls-remote", "--tags", repoPath)
```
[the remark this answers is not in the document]
denis256: AFAIK, `ls-remote` is considered plumbing part of git

cli/commands/scaffold/action.go
denis256: Disabled

cli/commands/scaffold/action.go
denis256: Added warning message

cli/commands/scaffold/action.go
brikis98: (This can be in a follow-up PR)

config/variable.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+		file, diags := parser.ParseHCL(content, tfFile)
+		if diags.HasErrors() {
+			opts.Logger.Warnf("Failed to parse HCL in file %s: %v", tfFile, diags)
+			continue
+		}
+
+		ctx := &hcl.EvalContext{}
+
+		if body, ok := file.Body.(*hclsyntax.Body); ok {
```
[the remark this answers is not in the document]
brikis98: (If any of this is useful, can be changed in a follow-up PR, BTW)

cli/commands/scaffold/action.go
brikis98: Sorry, I forgot one: disable hooks by default as well! Those also execute arbitrary code.

cli/commands/scaffold/action.go
```diff
@@ -0,0 +1,392 @@
...
+	refParam = "ref"
+
+	moduleUrlPattern = `git::([^:]+)://([^/]+)(/.*)`
+	moduleUrlParts   = 4
+
+	defaultBoilerplateConfig = `
+variables:
+
+`
+	defaultTerragruntTemplate = `
```
brikis98, line 48: I forgot one thing to include in this default template:

```hcl
include "root" {
}
```

Could you add that just below the `terraform { }` block? And please wrap it in an if statement, where it only renders if the `EnableRootInclude` variable is true. This variable should default to `true`, but users can shut it off if they don't want it via `-var EnableRootInclude=false`.

brikis98 (commented):
Thanks for updating the docs! I left comments there.

docs/_docs/02_features/scaffold.md
```diff
@@ -0,0 +1,121 @@
...
+* The `inputs` section is generated for you automatically, and will list all required and optional variables from the module, with their types, descriptions, and defaults, so you can easily fill them in to configure the module as you like.
+
+#### Custom templates for scaffolding
+
+Terragrunt has a basic template built-in for rendering `terragrunt.hcl` files, but you can provide your own templates to customize what code is generated! Scaffolding is done via [boilerplate](https://github.com/gruntwork-io/boilerplate), and Terragrunt allows you to specify custom boilerplate templates via two mechanisms:
+1. You can specify a custom boilerplate template to use as the second argument of the `scaffold` command.
+1. You can define a custom boilerplate template in a `.boilerplate` subfolder of your module.
+   If you define input variables in your boilerplate template, Terragrunt will prompt users for the values. Those values can also be passed in via `--var` and `--var-file` arguments.
+   There are also a set of variables that Terragrunt will automatically expose to your boilerplate templates for rendering:
```
brikis98, line 73: This gets indented weirdly if there isn't a new line:

```suggestion

There are also a set of variables that Terragrunt will automatically expose to your boilerplate templates for rendering:
```

cli/commands/scaffold/action.go
brikis98: Bumping this one more time: do we have a test that checks for this API? I'm worried we'll break it in a backward incompatible way by accident, and not notice... And any custom templates users created will suddenly no longer work.

cli/commands/scaffold/action.go
brikis98: Bumping this one more time: hooks should be disabled by default.

brikis98 (approved):
LGTM!

Let's get this merged 👍 🎉

matiri132:
Hi @denis256 , there's any way to enable Hooks for Scaffold? I'm working with boilerplates that needs the hoooks for my catalog usage.

PR #2843, opened 2023-12-09
Parse frontmatter data in the `catalog` command 

### LINKED ISSUE #2832: Improve README parsing for the `catalog` command

1. Extract into a function with a name and comment that makes this clearer. 
2. Add unit tests for that function. Should be easy to write, as the function takes a string as input!
3. How are you extracting the title?:
    1. If the doc has "frontmatter," try to get the title from that ([example](https://github.com/gruntwork-io/terraform-aws-service-catalog/blob/master/modules/services/ecs-cluster/README.md?plain=1#L3)). That'll be the cleanest version.
    2. Failing that, find the "H1" header (`# xxx`) in the README, and use that ([example](https://github.com/gruntwork-io/terraform-aws-service-catalog/blob/master/modules/services/ecs-cluster/README.md?plain=1#L12)).
    3. Failing that, use the name of the folder in which the module lives.
5. How are you extracting the description?
    1. If the doc has "frontmatter," try to get the description from that ([example](https://github.com/gruntwork-io/terraform-aws-service-catalog/blob/master/modules/services/ecs-cluster/README.md?plain=1#L4)). That'll be the cleanest version.
    4. Failing that, we should look for the first sentence or two that is not in a header and is not in Markdown syntax like an image or link. E.g., Ideally [these lines in this README](https://github.com/gruntwork-io/terraform-aws-service-catalog/blob/master/modules/services/ecs-cluster/README.md?plain=1#L20-L21). For rendering the description in the catalog table, we should strip out any Markdown to make it more readable.
    5. Failing that, list "(no description found)").

levkohimins:
## Description

* Parsing frontmatter data
* Strip Markdown/AsciiDoc tags

Fixes #2832


cli/commands/catalog/module/doc.go
```diff
@@ -0,0 +1,307 @@
...
+
+		tagRegs:        make(map[docTagName]*regexp.Regexp),
+		frontmatterReg: regexp.MustCompile(`(?i)^[\s\n]*<!-- frontmatter[\s\n]*([\S\s]*?)[\s\n]*-->`),
+	}
+
+	switch fileExt {
+	case mdExt:
+		doc.tagRegs[tagH1Block] = regexp.MustCompile(`(?:^|\n)\#{1}\s([\S\s]+?)(?:[\r\n]+\#|[\r\n]*$)`)
+		doc.tagRegs[tagH2Block] = regexp.MustCompile(`(?:^|\n)\#{2}\s([\S\s]+?)(?:[\r\n]+\#|[\r\n]*$)`)
+		doc.tagStripRegs = DocRegs{
```
brikis98, line 71: Will these work if something wraps over multiple lines? E.g.,:

```md
[Here is a
link that wraps over multiple lines](https://www.gruntwork.io)
```

Also, what about nested expressions?

```md
> [Here is **bold text** and `code` inside a link](https://www.gruntwork.io) all of which is in a quote block
```

I believe that's all valid Markdown...

brikis98: It's worth noting there are Markdown parsers... Maybe one of those would be a better way to handle this?

cli/commands/catalog/module/doc.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+
+	return doc.parseTag(docContent)
+}
+
+func (doc *Doc) IsMarkDown() bool {
+	return doc.fileExt == mdExt
+}
+
+func (doc *Doc) parseFrontmatter(key docDataKey) string {
```
brikis98, line 229: NIT: some comments in this file explaining what we're doing (i.e., that some modules have Markdown files with frontmatter, which we use as the preferred title/description source) may help!

cli/commands/catalog/module/doc_test.go
```diff
@@ -0,0 +1,233 @@
...
+			doc := newDoc(testCase.content, "")
+
+			assert.Equal(t, testCase.expectedName, doc.Title(), "Frontmatter Name")
+			assert.Equal(t, testCase.expectedDesc, doc.Description(0), "Frontmatter Description")
+		})
+	}
+
+}
+
+var testH1EksK8sArgocd = `
```
brikis98, line 80: Nice tests 👍

brikis98 (approved):
Overall, this approach may work, but check my comments on some Markdown usage that can sometimes be problematic with regex. If the code handles that OK, then I'm fine with this.

cli/commands/catalog/module/doc.go
levkohimins: > Will these work if something wraps over multiple lines? E.g.,:
> 
> ```
> [Here is a
> link that wraps over multiple lines](https://www.gruntwork.io)
> ```

Checked, this is the result:

```
Here is a
link that wraps over multiple
```

> 
> Also, what about nested expressions?
> 
> ```
> > [Here is **bold text** and `code` inside a link](https://www.gruntwork.io) all of which is in a quote block
> ```

Checked, this is the result:
```
Here is bold text and code inside a link all of which is in a quote block
```

So it works pretty good.

levkohimins: > It's worth noting there are Markdown parsers... Maybe one of those would be a better way to handle this?

The most important is to convert `md/adoc` into a plain text. I tried to find ready-made solutions before implementation but found nothing for Golang.

cli/commands/catalog/module/doc.go
levkohimins: Oh, sorry, I should have commented these funcs.  I will do it in the next task in order not to again bother you with a request for a review.

levkohimins:
> Overall, this approach may work, but check my comments on some Markdown usage that can sometimes be problematic with regex. If the code handles that OK, then I'm fine with this.

Thank you for the review!  I will comment code in the next PR.

PR #2857, opened 2023-12-18
Speed up repo cloning for the `catalog` command

### LINKED ISSUE #2831: Speed up the `catalog` command

There are some things we can do to speed up the catalog command:

1. Check out just the minimum amount of code (i.e., only main branch), not all branches/tags/etc.
1. Check out to a predictable temp path for each URL, and if that path already exists, then just git pull to update it, rather than checking it out from scratch again.

levkohimins:
## Description

Fixed the temporary folder path and repository URL that is sent to `go-getter`.

Fixes #2831.

### DIFF
--- a/cli/commands/catalog/module/repo.go
+++ b/cli/commands/catalog/module/repo.go
@@ -13,7 +13,6 @@ import (
 	"github.com/gruntwork-io/go-commons/files"
 	"github.com/gruntwork-io/terragrunt/pkg/log"
 	"github.com/gruntwork-io/terragrunt/terraform"
-	"github.com/gruntwork-io/terragrunt/util"
 	"github.com/hashicorp/go-getter"
 	"gopkg.in/ini.v1"
 )
@@ -23,8 +22,6 @@ const (
 	gitlabHost    = "gitlab.com"
 	azuredevHost  = "dev.azure.com"
 	bitbucketHost = "bitbucket.org"
-
-	tempDirFormat = "catalog-%x"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


cli/commands/catalog/module/repo.go
```diff
@@ -181,18 +177,13 @@ func (repo *Repo) clone(ctx context.Context) error {
...
-	log.Infof("Cloning repository %q to temprory directory %q", repo.cloneUrl, tempDir)
-
-	// if no repo directory is specified, `go-getter` returns the error "git exited with 128: fatal: not a git repository (or any of the parent directories"
-	if !strings.Contains(sourceUrl.RequestURI(), "//") {
-		sourceUrl.Path += "//."
-	}
+	log.Infof("Cloning repository %q to temprory directory %q", repo.cloneUrl, repo.tempDir)
 
-	if err := getter.GetAny(tempDir, strings.Trim(sourceUrl.String(), "/"), getter.WithContext(ctx)); err != nil {
+	if err := getter.Get(repo.tempDir, strings.Trim(sourceUrl.String(), "/"), getter.WithContext(ctx)); err != nil {
```
brikis98, line 182: Two questions:

1. Is this method smart enough to update an existing repo if it's already checked out there?
2. Does this method do a full `git clone`? Or is there some way we can do a "lighter weight" one that _only_ checks out `main` and not all the other branches and full history?

levkohimins: > Two questions:
> 
> 1. Is this method smart enough to update an existing repo if it's already checked out there?

Yes, It is.

https://github.com/hashicorp/go-getter/blob/975961f5f06346ccc282cd0d9aa16e160d26f9e3/get_git.go#L122-L130

> 2. Does this method do a full `git clone`? Or is there some way we can do a "lighter weight" one that _only_ checks out `main` and not all the other branches and full history?

This method only checks out the default branch.

https://github.com/hashicorp/go-getter/blob/975961f5f06346ccc282cd0d9aa16e160d26f9e3/get_git.go#L192-L204

brikis98: Sweeeeet!

brikis98 (approved):
LGTM!

levkohimins:
> LGTM!

Thanks for the review!

PR #2862, opened 2023-12-19
Fix catalog configuration file search.

### LINKED ISSUE #2861: terragrunt catalog config problems
**Describe the bug**

There are several problems:

1. When running `terragrunt catalog` from the root of an `infra-live` repo, it doesn't seem to find the `terragrunt.hcl` in the root of the repo.
2. When running `terragrunt catalog` from subfolders of an `infra-live` repo that _don't_ contain `terragrunt.hcl` files, it finds the `terragrunt.hcl` in the root, but is unable to load it correctly due to `find_in_parent_folders` usage.
3. When running `terragrunt catalog` from subfolders of an `infra-live` repo that _do_ contain `terragrunt.hcl` files, it seems to skip them, still finds the `terragrunt.hcl` in the root, and again is unable to load it correctly due to `find_in_parent_folders` usage.

**To Reproduce**

Create an `infra-live` repo with the following folder structure:

```tree
.
├── common.hcl
├── folder-with-terragrunt-hcl
│   └── terragrunt.hcl
├── folder-without-terragrunt-hcl
└── terragrunt.hcl
```

Contents of `common.hcl`:

```hcl
locals {
  name_prefix = "example"
}
```

Contents of the root `terragrunt.hcl`:

```hcl
locals {
  common_vars = read_terragrunt_config(find_in_parent_folders("common.hcl"))
}

catalog {
  urls = [
    "https://github.com/gruntwork-io/terraform-aws-utilities",
  ]
}

inputs = {
  name_prefix = local.common_vars.locals.name_prefix
}
```

Contents of `folder-with-terragrunt-hcl/terragrunt.hcl`:

```hcl
include "root" {
}
```

Run `terragrunt catalog` in the root of the repo:

```
ERRO[0000] /tmp/terragrunt.hcl:8,10-33: Error in function call; Call to function "find_in_parent_folders" failed: ParentFileNotFound: Could not find a terragrunt.hcl in any of the parent folders of /tmp/terragrunt.hcl. Cause: Traversed all the way to the root.., and 1 other diagnostic(s)
```

Why wasn't it able to find the `terragrunt.hcl` in that folder?

Run `terragrunt catalog` in the `folder-with-terragrunt-hcl` folder:

```
ERRO[0000] Error: Error in function call

ERRO[0000]   on /tmp/infra-live-repro/terragrunt.hcl line 2, in locals:
ERRO[0000]    2:   common_vars = read_terragrunt_config(find_in_parent_folders("common.hcl"))
ERRO[0000]
ERRO[0000] Call to function "find_in_parent_folders" failed: ParentFileNotFound: Could not find a common.hcl in any of the parent folders of /tmp/infra-live-repro/terragrunt.hcl. Cause: Traversed all the way to the
root..
```

It looks like it again skipped the `terragrunt.hcl` in the current folder and loaded the one in the root folder, which won't quite work, as it is looking for a `common.hcl` in parent folders.

Run `terragrunt catalog` in the `folder-without-terragrunt-hcl` folder:

```
ERRO[0000] Error: Error in function call

ERRO[0000]   on /tmp/infra-live-repro/terragrunt.hcl line 2, in locals:
ERRO[0000]    2:   common_vars = read_terragrunt_config(find_in_parent_folders("common.hcl"))
ERRO[0000]
ERRO[0000] Call to function "find_in_parent_folders" failed: ParentFileNotFound: Could not find a common.hcl in any of the parent folders of /tmp/infra-live-repro/terragrunt.hcl. Cause: Traversed all the way to the
root..
```

This error is the same as the previous, and it's the only one that's somewhat expected. This is due to an oversight on my part. The way we write the root `terragrunt.hcl` files is that they only work when pulled in via an `include` from a child `terragrunt.hcl`. That child `terragrunt.hcl` is in a subfolder of the root, so `find_in_parent_folders` works as expected there.

**Expected behavior**

Running `terragrunt catalog` from any of those folders should work identically.

**Versions**

- Terragrunt version: `v0.54.8`

**Additional context**

Here's my guess as to what's happening and how to fix it:

1. For some reason, we are _not_ looking for a `terragrunt.hcl` in the current folder; we're only searching in parent folders. That explains issues (1) and (2).
2. In the case where there actually isn't a `terragrunt.hcl` in the current folder, and there's only one in a parent folder, there's a good chance that `terragrunt.hcl` is a "root" config that is only meant to be pulled in via an `include` block. The only solution I can think of is if TG detects there isn't a `terragrunt.hcl` in the current folder, it creates one in the current working directory with an `include` block with a `find_in_parent_folders()` in it (most likely, this actually happens in memory, via a string template, so we aren't litering the disk with these fake files), and uses that. That way, you get the `find_in_parent_folders()` automatically, and most of the logic in root `terragrunt.hcl` files should "just work."

levkohimins:
## Description

Fixed the issue with finding a configuration file when running an application in the same directory as the configuration file.

Fixes #2861.


brikis98 (commented):
It looks like this would fix finding a `terragrunt.hcl` in the current working dir, but will this fix the issue where that file includes something like `find_in_parent_dir()`?

levkohimins:
> It looks like this would fix finding a `terragrunt.hcl` in the current working dir, but will this fix the issue where that file includes something like `find_in_parent_dir()`?

Yes, it will. To be sure, I've added your example as a test case.

brikis98:
> > It looks like this would fix finding a `terragrunt.hcl` in the current working dir, but will this fix the issue where that file includes something like `find_in_parent_dir()`?
> 
> Yes, it will. To be sure, I've added your example as a test case.

Ah, OK... And how is it working? E.g.,, if the root `terragrunt.hcl` has a `find_in_parent_folders("common.hcl")`, and you run `terragrunt catalog` from the root folder... How do you not get an error about not finding `common.hcl` in parent folders?

levkohimins:
> > > It looks like this would fix finding a `terragrunt.hcl` in the current working dir, but will this fix the issue where that file includes something like `find_in_parent_dir()`?
> > 
> > 
> > Yes, it will. To be sure, I've added your example as a test case.
> 
> Ah, OK... And how is it working? E.g.,, if the root `terragrunt.hcl` has a `find_in_parent_folders("common.hcl")`, and you run `terragrunt catalog` from the root folder... How do you not get an error about not finding `common.hcl` in parent folders?

I got this error when running from the parent directory :)
I added a test case to check this error and complicated the config.

brikis98:
> > > > It looks like this would fix finding a `terragrunt.hcl` in the current working dir, but will this fix the issue where that file includes something like `find_in_parent_dir()`?
> > > 
> > > 
> > > Yes, it will. To be sure, I've added your example as a test case.
> > 
> > 
> > Ah, OK... And how is it working? E.g.,, if the root `terragrunt.hcl` has a `find_in_parent_folders("common.hcl")`, and you run `terragrunt catalog` from the root folder... How do you not get an error about not finding `common.hcl` in parent folders?
> 
> I got this error when running from the parent directory :) I added a test case to check this error and complicated the config. It's working as expected.

Not sure I follow... We _don't_ want this to be an error. `terragrunt catalog` should work, whether you run it from the root, a subfolder with a `terragrunt.hcl`, or a subfolder with no `terragrunt.hcl`.

levkohimins:
> > > > > It looks like this would fix finding a `terragrunt.hcl` in the current working dir, but will this fix the issue where that file includes something like `find_in_parent_dir()`?
> > > > 
> > > > 
> > > > Yes, it will. To be sure, I've added your example as a test case.
> > > 
> > > 
> > > Ah, OK... And how is it working? E.g.,, if the root `terragrunt.hcl` has a `find_in_parent_folders("common.hcl")`, and you run `terragrunt catalog` from the root folder... How do you not get an error about not finding `common.hcl` in parent folders?
> > 
> > 
> > I got this error when running from the parent directory :) I added a test case to check this error and complicated the config. It's working as expected.
> 
> Not sure I follow... We _don't_ want this to be an error. `terragrunt catalog` should work, whether you run it from the root, a subfolder with a `terragrunt.hcl`, or a subfolder with no `terragrunt.hcl`.

Oh, sorry, I didn't read your idea carefully in the issue. I really liked it and will implement it.

levkohimins:
> > > > > > It looks like this would fix finding a `terragrunt.hcl` in the current working dir, but will this fix the issue where that file includes something like `find_in_parent_dir()`?
> > > > > 
> > > > > 
> > > > > Yes, it will. To be sure, I've added your example as a test case.
> > > > 
> > > > 
> > > > Ah, OK... And how is it working? E.g.,, if the root `terragrunt.hcl` has a `find_in_parent_folders("common.hcl")`, and you run `terragrunt catalog` from the root folder... How do you not get an error about not finding `common.hcl` in parent folders?
> > > 
> > > 
> > > I got this error when running from the parent directory :) I added a test case to check this error and complicated the config. It's working as expected.
> > 
> > 
> > Not sure I follow... We _don't_ want this to be an error. `terragrunt catalog` should work, whether you run it from the root, a subfolder with a `terragrunt.hcl`, or a subfolder with no `terragrunt.hcl`.
> 
> Oh, sorry, I didn't read your idea carefully in the issue. I really liked it and will implement it.

@brikis98, I has implemented your idea (you are a genius).

config/catalog.go
```diff
@@ -27,3 +38,40 @@ func (config *CatalogConfig) normalize(cofnigPath string) {
 		}
 	}
 }
+
+func ReadCatalogConfig(opts *options.TerragruntOptions) (*TerragruntConfig, error) {
```
brikis98, line 42: NIT: do you mind adding a comment here explaining what this method does and _why_?

Namely:

- We want users to be able to browse to any folder in an `infra-live` repo, run `terragrunt catalog` (with no URL) arg, and we automatically find the "closest" terragrunt.hcl, and load the catalogs from its `catalog` config.
- So this method searches the current working directory and its parent folders for a `terragrunt.hcl`.
- We then need to parse this file. Our normal parsing won't always work, as some `terragrunt.hcl` files are meant to be used from an `include` and/or they might use `find_in_parent_folders`, so they only work from certain child folders.
- So if we find a `terragrunt.hcl` that already has an `include`, we parse that as-is; if we find a `terragrunt.hcl` without `include`, we generate a stub child `terragrunt.hcl` in memory with an `include` to pull in the one we found.

levkohimins: Of course, thanks for the example text. I will try not to miss important explanations of functions in the future.

brikis98 (approved):
LGTM!

PR #2871, opened 2023-12-27
Add graph command to terragrunt

### LINKED ISSUE #1862: run-all equivalent, but based on graph and not folder tree
Consider our reference architecture, where we have multiple modules in the `services` folder that deploy into the EKS cluster that are adjacent to each other. In a full deployment, the `services` folder also contains other modules that are not related to the EKS cluster.

Suppose you want to destroy the EKS cluster only. Ideally, you would be able to use `run-all destroy` to destroy EKS cluster and all the dependents in order, but in the above setup, that could accidentally include other modules that you don't want to destroy.

Instead, what you want is something like a graph prune logic, where you want to destroy the EKS cluster and all of its dependents.

This most likely needs to be a new command other than `run-all`, something like `destroy-graph`, such that if you run `terragrunt destroy-graph` in the `eks-cluster` folder, it will destroy all the modules that deploy in to the EKS cluster and the cluster itself, in the right order.

denis256:
## Description

Included changes:
* Implemented building of dependent modules from current module
* Added `graph` command which run terragrunt command on graph of dependencies
* Added cli argument to pass custom "root" from where will be scanned dependencies(git root by default)
* Destroy command is executed in reverse order

Example:

dependency-graph

```
# eks
terragrunt graph apply 

Group 1
- Module project/eks

Group 2
- Module project/services/eks-service-1
- Module project/services/eks-service-2

Group 3
- Module project/services/eks-service-2-v2
- Module project/services/eks-service-3
- Module project/services/eks-service-5

Group 4
- Module project/services/eks-service-3-v2
- Module project/services/eks-service-4

Group 5
- Module project/services/eks-service-3-v3

```

```
# eks
terragrunt graph destroy
Group 1
- Module project/services/eks-service-2-v2
- Module project/services/eks-service-3-v3
- Module project/services/eks-service-4
- Module project/services/eks-service-5

Group 2
- Module project/services/eks-service-3-v2

Group 3
- Module project/services/eks-service-3

Group 4
- Module project/services/eks-service-1
- Module project/services/eks-service-2

Group 5
- Module project/eks
```

```
# services/eks-service-3
terragrunt graph destroy
Group 1
- Module project/services/eks-service-3-v3
- Module project/services/eks-service-4

Group 2
- Module project/services/eks-service-3-v2

Group 3
- Module project/services/eks-service-3
```

Fixes #1862

Fixes: #2911

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added `graph` comment to Terragrunt which allows execution of Terraform commands on graph of dependencies.


cli/commands/graph/action.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	// filter dependencies to have keep only dependencies with working dir
+	filterDependencies(stack, opts.WorkingDir)
+
+	return runall.RunAllOnStack(opts, stack)
+}
+
+// filterDependencies updates the stack to only include modules that are dependent on the given path
+func filterDependencies(stack *configstack.Stack, workDir string) {
```
brikis98, line 41: Did we not have to create this logic already in the past? I'm struggling to remember where, but this seems so familiar. Maybe for showing a warning when you run `destroy` that there may be other modules that have a `dependency` block on the current module, and those should be destroyed first? Or maybe the `render-json` command? Not sure, but might be worth a quick check to ensure we're not duplicating work.

cli/commands/graph/action.go
```diff
@@ -0,0 +1,87 @@
...
+
+		if len(module.Dependencies) != 0 {
+			for _, dep := range module.Dependencies {
+				dependentModules[dep.Path] = util.RemoveDuplicatesFromList(append(dependentModules[dep.Path], module.Path))
+			}
+		}
+	}
+
+	// for each element from slice copy respective slice from map
+	// loop while are changes in copy
```
brikis98, line 58: Sorry, not sure I understand what this loop is doing. What "changes" or "updates" are you looking for here? What does that have to do with the `initialSize` param? Is there some other way you can explain it?

cli/commands/graph/comman.go
```diff
@@ -0,0 +1,73 @@
+package graph
```
brikis98, line 1: Typo in file name? `comman.go` should be `command.go`?

cli/commands/graph/comman.go
```diff
@@ -0,0 +1,73 @@
...
+			Destination: &opts.GraphRoot,
+			Usage:       "Root directory from where to build graph dependencies.",
+		})
+	return globalFlags
+}
+
+func NewCommand(opts *options.TerragruntOptions) *cli.Command {
+	return &cli.Command{
+		Name:                   CommandName,
+		Usage:                  "Execute command on dependency graph.",
```
brikis98, line 36: This probably needs more explanation. E.g., "Execute command on dependency graph. The dependency graph consists of all the modules that depend on the module in current folder via dependency or dependencies blocks, and all the modules that depend on those modules, and so on. This allows you to execute commands on the full graph of dependencies of the current module, in the right order."

docs/_docs/04_reference/cli-options.md
```diff
@@ -528,6 +529,97 @@ Launch the user interface for searching and managing your module catalog.
 
 More details in [catalog section](https://terragrunt.gruntwork.io/docs/features/catalog/).
 
+### graph
+
+Run provided terraform command against a graph of dependencies build git repository root or path provided as `--graph-root` argument.
+Command will be executed only on modules that are dependent from the current module.
+Command will run following the order of dependencies, except when the command is 'destroy' - in that case, it will execute in the opposite order of the dependencies.
```
brikis98, line 536:
```suggestion
Run the provided terraform command against the graph of dependencies for the module in the current working directory. The graph consists of all modules that depend on the module in the current working directory via a `depends_on` or `dependencies` block, plus all the modules that depend on those modules, and all the modules that depend on those modules, and so on, recursively up the tree, up to the Git repository root, or the path specified via the optional `--graph-root` argument.

The Command will be executed following the order of dependencies: so it'll run on the module in the current working directory first, then on modules that depend on it directly, then on the modules that depend on those modules, and so on. Note that if the command is `destroy`, it will execute in the opposite order of the dependencies.
```

docs/_docs/04_reference/cli-options.md
```diff
@@ -528,6 +529,97 @@ Launch the user interface for searching and managing your module catalog.
...
+
+Example:
+Having bellow dependencies:
+[![dependency-graph](/assets/img/collections/documentation/dependency-graph.png){: width="80%" }]({{site.baseurl}}/assets/img/collections/documentation/dependency-graph.png)
+
+Running `terragrunt graph apply` in `eks` module will lead to bellow execution order:
```
brikis98, line 542:
```suggestion
Running `terragrunt graph apply` in `eks` module will lead to the following execution order: 
```

docs/_docs/04_reference/cli-options.md
```diff
@@ -528,6 +529,97 @@ Launch the user interface for searching and managing your module catalog.
...
+- Module project/services/eks-service-4
+
+Group 5
+- Module project/services/eks-service-3-v3
+```
+Notes:
+* `lambda` modules aren't included in the graph, because they are not dependent on `eks` module.
+* execution is from bottom up based on dependencies
+
+Running `terragrunt graph destroy` in `eks` module will lead to bellow execution order:
```
brikis98, line 567:
```suggestion
Running `terragrunt graph destroy` in `eks` module will lead to the following execution order:
```

brikis98 (commented):
Very nice 👍

cli/commands/graph/action.go
denis256: Good point, we have a listing of one-level dependent modules in: https://github.com/gruntwork-io/terragrunt/blob/master/cli/commands/terraform/action.go#L208

But it does not include dependencies of dependencies in the output, I can try to merge it in one function

cli/commands/graph/action.go
denis256: Added more detailed description

cli/commands/graph/action.go
denis256: Merged changes in one place

configstack/module.go
```diff
@@ -650,3 +655,62 @@ func buildDirList(terragruntOptions *options.TerragruntOptions, topLevelDir stri
...
+
+		if len(module.Dependencies) != 0 {
+			for _, dep := range module.Dependencies {
+				dependentModules[dep.Path] = util.RemoveDuplicatesFromList(append(dependentModules[dep.Path], module.Path))
+			}
+		}
+	}
+
+	// Floyd–Warshall inspired approach to find dependent modules
+	// merge map slices by key until no more updates are possible
```
brikis98, line 676: Ah, much clearer, thank you!

brikis98: One NIT: we should probably have unit tests for this method in particular, testing it out against all sorts of graphs, including those with cycles. But that can be added in a follow up PR, no need to hold up this one.

brikis98 (approved):
LGTM, thank you!

configstack/module.go
denis256: Good point, added tests

brikis98 (approved):
LGTM!

PR #2888, opened 2024-01-14
Fix cloning repos for the catalog command.

levkohimins:
## Description

`terragrunt catalog` clones repos into a temporary directory `/var/folder/....` (on MacOS) with a unique, persistent directory names for reuse. MacOS has a garbage collector, which means files are deleted if they haven't been accessed for three days, see https://developer.apple.com/forums/thread/71382, but it only deletes files and keeps all directories. In this case, `github.com/hashicorp/go-getter` returns an error when try to update repos.
```
/usr/bin/git exited with 128: fatal: not a git repository (or any of the parent directories): .git
```

This PR implements a check that deletes the repository's temporary directory if the file `/var/folder/..../.git/HEAD` does not exist so that it can be cloned again by the `github.com/hashicorp/go-getter` package.

### DIFF
--- a/cli/commands/catalog/module/repo.go
+++ b/cli/commands/catalog/module/repo.go
@@ -25,25 +25,24 @@ const (
 )
 
 var (
-	gitHeadBranchName = regexp.MustCompile(`^.*?([^/]+)$`)
+	gitHeadBranchNameReg    = regexp.MustCompile(`^.*?([^/]+)$`)
+	repoNameFromCloneURLReg = regexp.MustCompile(`(?i)^.*?([-a-z_.]+)[^/]*?(?:\.git)?$`)
 
 	modulesPaths = []string{"modules"}
 )
 
 type Repo struct {
-	cloneUrl string
+	cloneURL string
 	path     string
-	tempDir  string
 
 	remoteURL  string
 	branchName string
 }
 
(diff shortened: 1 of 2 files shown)


cli/commands/catalog/module/repo.go
```diff
@@ -137,36 +136,49 @@ func (repo *Repo) moduleURL(moduleDir string) (string, error) {
...
+
+	repo.path = filepath.Join(repo.path, repoName)
+
+	if files.FileExists(repo.path) && !files.FileExists(repo.gitHeadfile()) {
+		log.Debugf("The repo dir exists but git file %q does not. Removing the repo dir for cloning from the remote source.", repo.gitHeadfile())
+
+		if err := os.RemoveAll(repo.path); err != nil {
+			return errors.WithStackTrace(err)
+		}
+	}
```
brikis98, line 179: NIT: please add a comment explaining what you're checking for here and why it's necessary (i.e., tmp folders being deleted by macOs). You could link back to this PR too for more context.

levkohimins: Fair point. Added the comment.

brikis98 (approved):
LGTM

levkohimins:
@brikis98 Thanks for the review!

PR #2893, opened 2024-01-16
Speed up git clones and catalog command

brikis98:
## Description

This PR started as a way to improve the speed of the `terragrunt catalog` command. However, I believe the improvements here may also speed up Terragrunt in general. Here's a summary of the changes:

1. Add `depth=1` parameter to `source` URLs. This will have `go-getter` do a shallow `git clone`. Whenever we clone a repo, it is for read-only usage on a single branch/ref anyway, so a shallow clone should work fine, and should be much faster, especially for larger/older repos.
2. Run the `git clone` steps of the `catalog` command concurrently using goroutines. So instead of cloning and parsing one repo at a time, we do them all in parallel (up to number of CPU cores), which speeds things up quite a bit as well.

I tested the `catalog` command with 4 URLs in my `terragrunt.hcl` config and the average time for first run when from ~20 seconds to ~5 seconds.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

### DIFF
--- a/cli/commands/catalog/action.go
+++ b/cli/commands/catalog/action.go
@@ -3,6 +3,9 @@ package catalog
 import (
 	"context"
 	"fmt"
+	"os"
+	"path/filepath"
+
 	"github.com/gruntwork-io/go-commons/errors"
 	"github.com/gruntwork-io/terragrunt/cli/commands/catalog/module"
 	"github.com/gruntwork-io/terragrunt/cli/commands/catalog/tui"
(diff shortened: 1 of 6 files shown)


brikis98:
Ah, the `scaffold` command tries to look up Git tags on the repo, so shallow clone doesn't work... Will have to think on this a bit more.

yhakbar:
Closing PR as stale. Something to revisit in the future!

PR #2938, opened 2024-02-13
Improved handling of disabled modules

### LINKED ISSUE #2935: `skip` not respected if `dependency` link is configured
Consider two modules where one depends on the other.

**To Reproduce**
The minimal scenario looks like this:

```shell
.
├── first
│   └── terragrunt.hcl
├── module
│   └── main.tf
└── second
    └── terragrunt.hcl
```
`module/main.tf` just to have a most basic module with input/output vars
```hcl
terraform {
  required_providers {
    random = {}
  }
}

variable "input" {
  type = string
  default = "abc"
}

resource "random_string" "random" {
  length           = 1 
  special          = false
}

locals {
  trimmed = length(var.input) > 3 ? substr(var.input, 0, 3) : var.input
}

output "random_output" {
  value = random_string.random + local.trimmed
}
```

`first/terragrunt.hcl` that is being skipped:
```hcl
skip = true

terraform {
  source = "../module"
}

inputs = {
  input = "first"
}
```

and `second/terragrunt.hcl` that depends on `first`, but is also being skipped:
```hcl
skip = true

terraform {
  source = "../module"
}

dependency "first" {
  config_path = "../first"

  mock_outputs_allowed_terraform_commands = ["init", "destroy", "validate"]
  mock_outputs_merge_strategy_with_state  = "deep_map_only"
  mock_outputs = {
    random_output = ""
  }
}

inputs = {
  input = dependency.first.outputs.random_output
}
```

**Expected behavior**
Both modules would be skipped with no errors printed out. However, already with `terragrunt run-all init` we're getting errors from the `second` module:

```
ERRO[0000] Module .../second has finished with an error: Could not parse output from terragrunt config .../first/terragrunt.hcl. Underlying error: unexpected end of JSON input  prefix=[.../second] 
```

I would be expecting that:
1. if the module is skipped, there's no complain from it
2. alternatively, that the mock of the outputs of the first module saves the situation

**Nice to have**
- [X] Terminal output
- [ ] Screenshots

**Versions**
- Terragrunt version: 0.55.1
- Terraform version: 1.2.9
- Environment details (Ubuntu 20.04, Windows 10, etc.): Linux (OpenSUSE Leap, 15.3)
The issue also occurs on other linux systems. If more info is needed, I'll happily provide.

**Additional context**
Without the dependency declared, the `second` module is skipped as expected.

denis256:
## Description

Included changes:
* added check for `skip` value when processing dependencies to disable dependencies with `skip=true` flag
* disabled modules will use mock values
* added warning message in case if parsing fails to keep backward compatibility and for cases when disabled modules is "broken" and can't be parsed as HCL

Fixes #2935.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Improved dependency management by ensuring 'skip' flag properly excludes evaluation of skipped dependencies.

### DIFF
--- a/config/dependency.go
+++ b/config/dependency.go
@@ -180,13 +180,17 @@ func decodeAndRetrieveOutputs(ctx *ParsingContext, file *hclparse.File) (*cty.Va
 	updatedDependencies := terragruntDependency{}
 	for _, dep := range decodedDependency.Dependencies {
 		depPath := getCleanedTargetConfigPath(dep.ConfigPath, ctx.TerragruntOptions.TerragruntConfigPath)
-		depConfig, err := PartialParseConfigFile(ctx.WithDecodeList(TerragruntFlags), depPath, nil)
-		if err != nil {
-			return nil, errors.WithStackTrace(err)
-		}
-		if depConfig.Skip {
-			ctx.TerragruntOptions.Logger.Debugf("Skipping outputs reading for disabled dependency %s", dep.Name)
-			dep.Enabled = new(bool)
+		if dep.isEnabled() && util.FileExists(depPath) {
+			depConfig, err := PartialParseConfigFile(ctx.WithDecodeList(TerragruntFlags), depPath, nil)
+			if err == nil {
+				if depConfig.Skip {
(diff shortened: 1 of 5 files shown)


brikis98 (approved):
LGTM!

Issue #2952, opened 2024-02-20
Terragrunt catalog freezes consistently after scaffolding 


iangrunt:
**Describe the bug**
When running `terragrunt catalog` and using the `Scaffold` subcommand, it appears the `terragrunt.hcl` is generated in the current directory, but my terminal becomes unresponsive to the usual UNIX keyboard shortcuts to kill, bg, etc. the current process. 

The workaround is to manually find the `terragrunt catalog ..` process and SIGKILL it. 

**To Reproduce**
Steps to reproduce the behavior, code snippets, and examples that can be used to reproduce the issue.

```bash
(base) ~/tmp/new_dir_20240220110839_26288/infrastructure-live/networking-01/us-east-2/inspection-vpc 0 $ terragrunt catalog git@github.com:gruntwork-io/terraform-aws-cis-service-catalog.git

```

Screenshots: I ran a `catalog` against the VPC repo, and it froze on:

<img width="1840" alt="Screenshot 2024-02-20 at 5 57 47 PM" src="https://github.com/gruntwork-io/terragrunt/assets/101607944/b953b254-501c-4bb4-9b9c-a8820f07df25">
<img width="1840" alt="Screenshot 2024-02-20 at 5 57 43 PM" src="https://github.com/gruntwork-io/terragrunt/assets/101607944/0c5412c2-858a-4feb-857d-b17ab295fd44">

```
(base) ~ 0 $ terragrunt --version
terragrunt version v0.55.3
```

**Expected behavior**

**Nice to have**
- [x] Terminal output
- [x] Screenshots

**Versions**
- Terragrunt version:
- Terraform version:
- Environment details (Ubuntu 20.04, Windows 10, etc.):

**Additional context**

brikis98:
@denis256 Could you take a look?

denis256:
@iangrunt hi, I have a couple of questions:
* when UI freeze? after selecting a module(which module?) or it is not printing any output? 
* which OS/Terminal was used?

in my test on Ubuntu 22.04 catalog and scaffold generate HCL files

![image](https://github.com/gruntwork-io/terragrunt/assets/10694338/81b861ca-c1c3-4aa7-b19a-49e5778c9121)

iangrunt:
Hey @denis256, I'm using Mac OS and the default terminal - this issue has chased me around since the first release, and I have tried to resolve it keeping my Terragrunt up to date. 

[...]

PR #2993, opened 2024-03-12
Catalog TUI Improvements

robmorgan:
## Description

This PR refactors our catalog TUI to try to prevent some of the terminal errors from occurring.

Fixes https://github.com/gruntwork-io/terragrunt/issues/2952.

## TODOs

- [x] Update the docs.

## Release Notes (draft)


cli/commands/catalog/tui/components/buttonbar/buttonbar.go
```diff
@@ -0,0 +1,103 @@
...
+
+	return s.String()
+}
+
+func (b *ButtonBar) activeBtnCmd() tea.Msg {
+	return ActiveBtnMsg(b.activeButton)
+}
+
+// SelectBtnCmd is a bubbletea command that selects the button at the given index.
+func SelectBtnCmd(button int) tea.Cmd {
```
denis256, line 99: function `SelectBtnCmd` seems to not be used

cli/commands/catalog/tui/view.go
```diff
@@ -0,0 +1,53 @@
...
+	infoLineStyle     = lipgloss.NewStyle().Foreground(lipgloss.Color("#1D252"))
+	infoHelp          = lipgloss.NewStyle().Padding(2, 0, 0, 2)
+)
+
+// View is the main view, which just calls the appropriate sub-view and returns a string representation of the TUI
+// based on the application's state.
+func (m model) View() string {
+	var s string
+
+	switch m.state {
```
denis256, line 22: Missing default case

cli/commands/catalog/tui/update.go
```diff
@@ -0,0 +1,184 @@
...
+
+	case rendererErrMsg:
+		m.viewport.SetContent(fmt.Sprintf("there was an error rendering markdown: %s", msg.err.Error()))
+		// ensure we show the viewport
+		m.state = pagerState
+	}
+
+	// Hand off the message and model to the appropriate update function for the
+	// appropriate view based on the current state.
+	switch m.state {
```
denis256, line 159: Missing default case

cli/commands/catalog/tui/components/buttonbar/buttonbar.go
robmorgan: thanks, I'll remove it.

cli/commands/catalog/tui/view.go
robmorgan: thanks, I'll add one.

cli/commands/catalog/tui/update.go
robmorgan: This is a bit trickier, and I'd rather avoid it to be honest. As the `Update` eventloop [runs so many times per second](https://github.com/charmbracelet/bubbletea/blob/f611ed97a3a63728987874354424b34d7d8efef6/tea.go#L307), I think it would make sense to let the control flow through to the `return m, nil` statement below. I've noticed a lot of their own examples and production apps mainly omit the `default` case. I suspect this is for performance reasons. Note: I'm not bullish on this and I haven't done any benchmarks. If having it there makes you feel more comfortable/satisfies the linter I'll put something in.

robmorgan:
@denis256 thanks I'll fix this one:
> INFO [runner] linters took 40.220868596s with stages: goanalysis_metalinter: 40.215411498s 
cli/commands/catalog/tui/model.go:70:80: newModel - result 1 (error) is always nil (unparam)
func newModel(modules module.Modules, opts *options.TerragruntOptions) (model, error) {

robmorgan:
@denis256 regarding the other remarks. I won't fix the `2 / magic number` errors. It seems we'd need to use a constant which would be a bit overkill in certain places or stop using the `go-mnd` linter. What do you think? https://forum.golangbridge.org/t/how-to-solve-the-magic-number-linter-warning/20421

denis256:
> @denis256 regarding the other remarks. I won't fix the `2 / magic number` errors. It seems we'd need to use a constant which would be a bit overkill in certain places or stop using the `go-mnd` linter. What do you think? https://forum.golangbridge.org/t/how-to-solve-the-magic-number-linter-warning/20421

I think it is fine to leave a magic value and set a linter comment to ignore this checked in this specific file through `//nolint:gomnd`

robmorgan:
@denis256 
> I think it is fine to leave a magic value and set a linter comment to ignore this checked in this specific file through //nolint:gomnd

Okay I've done that.

denis256 (approved):

PR #3001, opened 2024-03-14
Caching Terraform providers

### LINKED ISSUE #2920: Add guidance/features for reducing disk space and bandwidth usage
# The problem

Many of our customers struggle with TG eating up a ton of disk space and bandwidth: hundreds of gigabytes in some cases! I think this comes from a few sources (note: https://github.com/gruntwork-io/terragrunt/issues/2919 may help provide the data we need understand this better):

1. **Providers**. By default, Terraform downloads providers from scratch every time you run `init`. This isn't a problem for a single module, but if you do `run-all` in a repo that has, say, 50 `terragrunt.hcl` files, each one runs `init` on a TF module, each module downloads an average of, say, 10 providers, then that's 50 * 10 = 500 provider downloads—_even if it's the exact same 10 providers across all 50 modules_! 
2. **Repos**. When you set a `source` URL in TG, it downloads the whole repo into a `.terragrunt-cache` folder. If you have 50 `terragrunt.hcl` files with `source` URLs, and do `run-all`, it will download the repos 50 times—_even if all 50 repos are the same!_
3. **Git clone**. I think TG is doing a "full" Git clone for the code in the `source` URL. We should consider doing a shallow clone, as that would be much faster/smaller. 
4. **Modules**. When you run `init`, Terraform downloads repos to a `.modules` folder. If you have 50 `terragrunt.hcl` files, each of which has a `source` URL pointing to TF code that contains, say, 10 modules, then when you do `run-all`, you'll end up downloading 50 * 10 = 500 repos—_even if all the modules are exactly the same!_
5. **Ephemeral storage**. Many users run TG in places with ephemeral storage—e.g., in a K8S cluster—where the disk is totally empty/fresh on each run. So all the stuff you downloaded in previous runs isn't even available, and you have to download everything from scratch each time.

# Goals

We should have the following goals:

1. Each unique provider should only ever be downloaded once on a given computer.
2. Each unique repo should only ever be cloned once on a given computer.
3. Each unique module should only ever be cloned once on a given computer.
4. `git clone` should be as efficient as possible: e.g., use shallow clones.

# The solution

The solution will need to be some mix of:

1. **Provide guidance**. One big part of this should be adding new docs to the TG docs site that provide clear, step-by-step instructions to optimize disk space and bandwidth usage. This could include:
    1. How to use TF features such as [provider plugin caching](https://developer.hashicorp.com/terraform/cli/config/config-file#provider-plugin-cache) and the [`provider_installation` block](https://developer.hashicorp.com/terraform/cli/config/config-file#explicit-installation-method-configuration) to achieve the goals in the previous section.
    2. How to configure K8S and other ephemeral systems to re-use previously downloaded data.
2. **Add new features**. We should consider what new features could be added to TG to help. Some initial ideas:
    1. A new command to automatically configure the current system for "optimal" provider caching as per the guidance in the previous point.
    2. Switch to using shallow clones everywhere. See https://github.com/gruntwork-io/terragrunt/pull/2893 for some experimentation with adding a `depth=1` param to tell go-getter to do a shallow clone.
    3. Use some sort of repo/module cache to avoid re-downloading modules over and over again.

# Notes

- We can implement these improvements incrementally. We don't have to fix everything in a single giant PR.
- The providers seem to take up the most disk space and bandwidth, so we should prioritize that first: in fact, we should do the research and write up our guidance as the very first step, and then consider if new features are merited in follow-up steps.
- We might not be able to achieve all the goals in the goals section, as some may be dependent on internal TF implementation details. But we should strive to get as close as we can.

levkohimins:
## Description

Using the implemented built-in private registry, Terragrunt caches Terraform providers without using `plugin_cache_dir`, thereby ensuring that multiple Terraform processes can safely run concurrently.

Fixes #2920.


terraform/registry/server.go
```diff
@@ -0,0 +1,110 @@
...
+func (server *Server) ProviderURL() *url.URL {
+	return &url.URL{
+		Scheme: "http",
+		Host:   server.config.Addr(),
+		Path:   server.providerController.Path(),
+	}
+}
+
+// Run starts the webserver and workers.
+func (server *Server) Run(ctx context.Context) error {
```
brikis98, line 79: What happens if you run multiple TG commands concurrently on the same system? Will they clash because they both try to fire up a server on the same port?

terraform/registry/server.go
```diff
@@ -0,0 +1,110 @@
...
+	log.Infof("Start Private Registry")
+
+	addr := server.config.Addr()
+	srv := &http.Server{Addr: addr, Handler: server}
+
+	errGroup, ctx := errgroup.WithContext(ctx)
+	errGroup.Go(func() error {
+		return server.providerService.RunCacheWorker(ctx)
+	})
+	errGroup.Go(func() error {
```
brikis98, line 89: We need "unit tests" for this server. I say "unit test," because it's perhaps closer to an integration test, but I'd really like to have tests that just test this server as a single unit: run server, hit various endpoints, check they work as expected, check auth is enforced, check multiple concurrent servers work, etc.

test/integration_test.go
```diff
@@ -197,6 +199,82 @@ const (
 	fixtureDownload = "fixture-download"
 )
 
+func TestTerragruntProviderCache(t *testing.T) {
```
brikis98, line 202: Thanks for adding this test! 

Given the rather large impact this has across all TG functionality, could we find a way to run _all_ TG integration tests with (a) caching enabled and (b) caching disabled? I want to be sure all existing functionality still works with caching.

test/integration_test.go
```diff
@@ -197,6 +199,82 @@ const (
...
+			"hashicorp/kubernetes/2.27.0",
+		},
+	}
+
+	registryName := "registry.terraform.io"
+	if strings.Contains(wrappedBinary(), "tofu") {
+		registryName = "registry.opentofu.org"
+	}
+
+	for subDir, providers := range providers {
```
brikis98, line 228: Does this test have some way of validating that each provider was downloaded and stored exactly once?

cli/commands/run-all/command.go
```diff
@@ -17,32 +18,89 @@ import (
...
+			Name:        FlagNameTerragruntProviderCache,
+			Destination: &opts.ProviderCache,
+			EnvVar:      "TERRAGRUNT_PROVIDER_CACHE",
+			Usage:       "Enables provider cache.",
+		},
+		&cli.GenericFlag[string]{
+			Name:        FlagNameTerragruntProviderCacheDir,
+			Destination: &opts.ProviderCacheDir,
+			EnvVar:      "TERRAGRUNT_PROVIDER_CACHE_DIR",
+			Usage:       "The path to the cache directory. Default is .terragrunt-cache/provider-cache in the working directory.",
```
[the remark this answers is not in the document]
levkohimins: Using the same directory by multiple TG processes could cause a conflict, the same as running multiple TF processes with the same PluginCacheDir path. I modified the code, now _Terragrunt Provider Cache_ uses a flock to prevent this. 

Now it uses the native Go function https://pkg.go.dev/os#UserCacheDir + "terragrunt/providers" to determine the default cache directory.

cli/commands/run-all/command.go
```diff
@@ -17,32 +18,89 @@ import (
...
+			Name:        FlagNameTerragruntProviderCompleteLock,
+			Destination: &opts.ProviderCompleteLock,
+			EnvVar:      "TERRAGRUNT_PROVIDER_COMPLETE_LOCK",
+			Usage:       "Disables terraform 'plugin_cache_may_break_dependency_lock_file' feature.",
+		},
+		&cli.GenericFlag[string]{
+			Name:        FlagNameTerragruntRegistryToken,
+			Destination: &opts.RegistryToken,
+			EnvVar:      "TERRAGRUNT_REGISTRY_TOKEN",
+			Usage:       "The Token for connecting to the built-in Private Registry server. By default generated automatically.",
```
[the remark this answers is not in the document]
levkohimins: Agree. Renamed to "Terragrunt Provider Cache".

cli/commands/run-all/command.go
```diff
@@ -17,32 +18,89 @@ import (
...
+		},
+		&cli.GenericFlag[string]{
+			Name:        FlagNameTerragruntRegistryHostname,
+			Destination: &opts.RegistryHostname,
+			EnvVar:      "TERRAGRUNT_REGISTRY_HOSTNAME",
+			Usage:       "The hostname of the built-in Private Registry server. Default is 'localhsot'.",
```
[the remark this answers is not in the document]
levkohimins: Thanks

cli/commands/run-all/command.go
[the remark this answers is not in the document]
levkohimins: This can be any text except an empty string.

cli/commands/run-all/provider_cache.go
```diff
@@ -0,0 +1,207 @@
...
+	})
+
+	ctx = terraformcmd.ContextWithRetry(ctx, &terraformcmd.Retry{
+		HookFunc: func(ctx context.Context, opts *options.TerragruntOptions, callback terraformcmd.RetryCallback) (*shell.CmdOutput, error) {
+			// Use Hook only for the `terraform init` command, which can be run explicitly by the user or Terragrunt's `auto-init` feature.
+			if util.FirstArg(opts.TerraformCliArgs) == terraform.CommandNameInit {
+				log.Debugf("Caching providers for %q", opts.WorkingDir)
+
+				// Before each init, we warm up the global cache to ensure that all necessary providers are cached.
+				// It's low cost operation, because ecause it does not cache the same provider twice, but only new previously non-existent providers.
```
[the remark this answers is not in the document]
levkohimins: Thanks

cli/commands/run-all/provider_cache.go
```diff
@@ -0,0 +1,207 @@
...
+				if err := RunTerraformProvidersLockCommand(opts, "-platform="+controllers.PlatformNameCacheProvider); err != nil {
+					return nil, err
+				}
+				provider.WaitForCacheReady()
+
+				// Create copmplete terraform lock files. By default this feature is disabled, since it's not superfast.
```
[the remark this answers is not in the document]
levkohimins: Thanks

cli/commands/run-all/provider_cache.go
```diff
@@ -0,0 +1,207 @@
...
+				}
+			}
+
+			return callback(ctx, opts)
+		},
+	})
+	if err := Run(ctx, opts); err != nil {
+		return err
+	}
+	cancel()
```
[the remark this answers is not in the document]
levkohimins: You are right, improved.

cli/commands/run-all/provider_cache.go
```diff
@@ -0,0 +1,207 @@
...
+
+	return errGroup.Wait()
+}
+
+// RunTerraformProvidersLockCommand runs `terraform providers lock` for two purposes:
+// 1. First, warm up the global cache
+// 2. To create complete terraform lock files, if `opts.ProviderCompleteLock` is true
```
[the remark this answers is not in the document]
levkohimins: Simplified the code. I've added your suggestion in another place. Thanks.

cli/commands/run-all/provider_cache.go
```diff
@@ -0,0 +1,207 @@
...
+//
+//	provider_installation {
+//		filesystem_mirror {
+//			path    = "/path/to/the/provider/cache"
+//			include = ["example.com/*/*"]
+//		}
+//		direct {
+//			exclude = ["example.com/*/*"]
+//		}
+//	}
```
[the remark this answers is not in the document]
levkohimins: Added

cli/commands/run-all/provider_cache.go
```diff
@@ -0,0 +1,207 @@
...
+	for _, registryName := range opts.RegistryNames {
+		envName := fmt.Sprintf(terraform.EnvNameTFTokenFmt, strings.ReplaceAll(registryName, ".", "_"))
+		opts.Env[envName] = opts.RegistryToken
+	}
+
+	if !opts.ProviderCompleteLock {
+		opts.Env[terraform.EnvNameTFPluginCacheMayBreakDependencyLockFile] = "1"
+	}
+
+	opts.Env[terraform.EnvNameTFCLIConfigFile] = cliConfigFile
```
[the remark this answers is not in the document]
levkohimins: It's correct. Added.

terraform/config.go
```diff
@@ -2,17 +2,143 @@ package terraform
...
 
-// IsPluginCacheUsed returns true if the terraform plugin cache dir is specified, https://developer.hashicorp.com/terraform/cli/config/config-file#provider-plugin-cache
-func IsPluginCacheUsed() bool {
-	if strings.TrimSpace(os.Getenv("TF_PLUGIN_CACHE_DIR")) != "" {
-		return true
+var (
+	configParamNamePluginCacheDirReg = regexp.MustCompile(`(?mi)^.*plugin_cache_dir.*$`)
+)
+
+// ProviderInstallationMethod represents an installation method block inside a provider_installation block.
```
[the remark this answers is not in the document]
levkohimins: Added

terraform/config.go
```diff
@@ -2,17 +2,143 @@ package terraform
...
+		FilesystemMirror: filesystemMethod,
+		Direct:           directMethod,
+	}
+
+	cfg.ProviderInstallation = append(cfg.ProviderInstallation, providerInstallation)
+}
+
+// Save marshalls and saves CLI config with the given config path.
+// In order to not lose user/default settings, if the user/default CLI config file exists, read this config and place at the top our the config file.
+func (cfg *Config) Save(configPath string) error {
```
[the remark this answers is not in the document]
levkohimins: No, I don't. Unfortunately, there was not enough time for this.

terraform/registry/server.go
levkohimins: Unfortunately, there was not enough time for this.

terraform/registry/server.go
levkohimins: It was like that. Now the free port of the port is resolved automatically.

terraform/registry/handlers/auth.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/pkg/errors"
+)
+
+const AuthorizationApiKeyHeaderName = "x-api-key"
+
+type Authorization struct {
+	Token string
+}
+
+func (auth *Authorization) Auth(bearerToken string, ctx echo.Context) (bool, error) {
```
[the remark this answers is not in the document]
levkohimins: Updated.

test/integration_test.go
levkohimins: Added tasks with provider cache.

test/integration_test.go
levkohimins: Yes, it does all these checks.

terraform/registry/config.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+type Config struct {
+	Hostname        string
+	Port            int
+	Token           string
+	ShutdownTimeout time.Duration
+}
+
+func NewConfig() *Config {
```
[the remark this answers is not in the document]
levkohimins: Fixed.

terraform/registry/controllers/provider.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/gruntwork-io/terragrunt/terraform/registry/handlers"
+	"github.com/gruntwork-io/terragrunt/terraform/registry/models"
+	"github.com/gruntwork-io/terragrunt/terraform/registry/router"
+	"github.com/gruntwork-io/terragrunt/terraform/registry/services"
+	"github.com/labstack/echo/v4"
+)
+
+const (
+	// name using for the discovery
+	porviderName = "providers.v1"
```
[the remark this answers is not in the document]
levkohimins: Thanks.

util/trap_writer.go
```diff
@@ -0,0 +1,42 @@
...
+
+var ansiReg = regexp.MustCompile(ansi)
+
+type TrapWriter struct {
+	writer      io.Writer
+	reg         *regexp.Regexp
+	trappedMsgs []string
+}
+
+func NewTrapWriter(writer io.Writer, reg *regexp.Regexp) *TrapWriter {
```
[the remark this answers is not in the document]
levkohimins: Added.

util/lockfile.go
```diff
@@ -0,0 +1,43 @@
...
+func (lockfile *Lockfile) TryLock() error {
+	if locked, err := lockfile.Flock.TryLock(); err != nil {
+		return errors.WithStackTrace(err)
+	} else if !locked {
+		return errors.Errorf("unable to lock file %s", lockfile.Path())
+	}
+	return nil
+}
+
+func AcquireLockfile(filename string) (*Lockfile, error) {
```
denis256, line 39: `AcquireLockfile` seems to be unused

```
grep -irn "AcquireLockfile" .
./util/lockfile.go:39:func AcquireLockfile(filename string) (*Lockfile, error) {
./terraform/cache/services/provider.go:144:func (cache *ProviderCache) acquireLockFile(ctx context.Context) (*util.Lockfile, error) {
./terraform/cache/services/provider.go:268:                             lockfile, err := cache.acquireLockFile(ctx)

```

terraform/cache/services/provider.go
```diff
@@ -0,0 +1,298 @@
...
+	// The path to store archive providers that are retrieved from the source registry and cached to reduce traffic.
+	baseArchiveDir string
+
+	providerCaches        ProviderCaches
+	providerCacheWarmUpCh chan *ProviderCache
+
+	cacheMu      sync.RWMutex
+	cacheReadyMu sync.RWMutex
+
+	// If needCacheArchive is true, ensures that not only the unarchived binary is cached, but also its archive. We need acrhives in order to reduce the bandwidth, because `terraform lock provider` always loads providers from a remote registry to create a lock file rather than using a cached one. This is only used when opts.ProviderCompleteLock is true.
```
denis256, line 176: `archives` I suspect

terraform/cache/services/provider.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+}
+
+// WaitForCacheReady blocks the call until all providers are cached.
+func (service *ProviderService) WaitForCacheReady() {
+	service.cacheReadyMu.Lock()
+	defer service.cacheReadyMu.Unlock()
+}
+
+// CacheProvider starts caching the given provider using non-blocking approch.
```
denis256, line 196: `approach`

terraform/cache/services/provider.go
```diff
@@ -0,0 +1,298 @@
...
+	"github.com/gruntwork-io/terragrunt/terraform/cache/models"
+	"github.com/gruntwork-io/terragrunt/util"
+	"github.com/hashicorp/go-getter/v2"
+	"github.com/hashicorp/go-multierror"
+	"github.com/hashicorp/terraform/command/cliconfig"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
+)
+
+var (
```
denis256, line 24: Looks like variables are used as constants, can it be declared as `const`?

util/lockfile.go
levkohimins: Removed.

util/fetch.go
```diff
@@ -0,0 +1,49 @@
...
+	"os"
+	"time"
+
+	"github.com/gruntwork-io/go-commons/errors"
+	"github.com/hashicorp/go-getter/v2"
+)
+
+// FetchFile downloads the file from the given `downloadURL` into the specified `saveToFile` file.
+func FetchFile(ctx context.Context, downloadURL, saveToFile string) error {
+	req, err := http.NewRequestWithContext(ctx, "GET", downloadURL, nil)
```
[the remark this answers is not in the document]
levkohimins: Good advice. Replaced.

test/integration_test.go
```diff
@@ -197,6 +199,86 @@ const (
...
+		},
+		"second": []string{
+			"hashicorp/aws/5.40.0",
+			"hashicorp/azurerm/3.95.0",
+			"hashicorp/kubernetes/2.27.0",
+		},
+	}
+
+	registryName := "registry.terraform.io"
+	if strings.Contains(wrappedBinary(), "tofu") {
```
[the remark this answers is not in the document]
levkohimins: Ah, did notice it. Replaced.

test/integration_serial_test.go
```diff
@@ -420,12 +420,7 @@ func TestTerragruntOutputJson(t *testing.T) {
 	err := runTerragruntCommand(t, fmt.Sprintf("terragrunt apply --terragrunt-json-log --terragrunt-non-interactive --terragrunt-working-dir %s", testPath), &stdout, &stderr)
 	assert.Error(t, err)
 
-	var output map[string]interface{}
-
-	err = json.Unmarshal(stderr.Bytes(), &output)
-	assert.NoError(t, err)
-
-	assert.Contains(t, output["msg"], "Downloading Terraform configurations from git::https://github.com/gruntwork-io/terragrunt.git?ref=v0.9.9")
+	assert.Contains(t, stderr.String(), "Downloading Terraform configurations from git::https://github.com/gruntwork-io/terragrunt.git?ref=v0.9.9")
```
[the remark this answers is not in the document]
levkohimins: The output is not JSON, it displays multiple JSONs,
```
{...}
{...}
{...}
```
This cannot be parsed. That worked before, because there was only one record of log, and it looked like json, but after adding the logs, it broke. This is a set of json logs of `logrus`, and not the output of terraform, I think this is not critical.

terraform/cache/services/provider.go
levkohimins: Renamed.

terraform/cache/services/provider.go
levkohimins: Fixed.

terraform/cache/services/provider.go
levkohimins: Ah, indeed, thanks. Fixed.

docs/_docs/02_features/provider-cache.md
```diff
@@ -0,0 +1,62 @@
...
+excerpt: Learn how to use provider cache with run-all commands.
+tags: ["cache"]
+order: 280
+nav_title: Documentation
+nav_title_link: /docs/
+---
+
+## Provider Cache
+
+Terraform has provider caching feature [Provider Plugin Cache](https://developer.hashicorp.com/terraform/cli/config/config-file#provider-plugin-cache), that does the job well until you run multiple Terraform processes simultaneously. Then the Terraform processes begin conflict by overwriting each other’s cache, which causes the error `Error: Failed to install provider`. By default, to speed up processing, Terragrunt with the [`run-all`](https://terragrunt.gruntwork.io/docs/reference/cli-options/#run-all) command processes modules simultaneously, which in turn prevents the use of [Provider Plugin Cache](https://developer.hashicorp.com/terraform/cli/config/config-file#provider-plugin-cache) to avoid conflicts or force Terragrunt to process modules sequentially by specifying the `--terragrunt-parallelism 1` flag, which significantly slows down the work.
```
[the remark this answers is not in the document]
levkohimins: Initially I did it only for `run-all` commands, but now it works for all commands.

test/integration_serial_test.go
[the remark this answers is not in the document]
denis256: Yes, it is by intention in this way to be parsed by other applications like Datadog

terraform/cache/controllers/discovery.go
```diff
@@ -0,0 +1,42 @@
...
+func (controller *DiscoveryController) terraformAction(ctx echo.Context) error {
+	endpoints := make(map[string]any)
+
+	for _, endpointer := range controller.Endpointers {
+		for name, urlPath := range endpointer.Endpoints() {
+			endpoints[name] = urlPath
+		}
+	}
+
+	return ctx.JSON(http.StatusOK, endpoints)
```
denis256, line 41: Does it make sense to wrap the returned error?

terraform/cache/controllers/downloader.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	if cache := controller.ProviderService.GetProviderCache(provider); cache != nil {
+		if filename := cache.ArchiveFilename(); filename != "" {
+			log.Debugf("Provider %q uses cached file %q", cache.Provider, filename)
+			return ctx.File(filename)
+		}
+	}
+
+	return controller.ReverseProxy.NewRequest(ctx, provider.DownloadURL)
```
denis256, line 64: Does it make sense to wrap the returned error?

cli/provider_cache.go
```diff
@@ -0,0 +1,231 @@
...
+	if cfgDir := filepath.Dir(cliConfigFile); !util.FileExists(cfgDir) {
+		if err := os.MkdirAll(cfgDir, os.ModePerm); err != nil {
+			return errors.WithStackTrace(err)
+		}
+	}
+
+	if err := cfg.Save(cliConfigFile); err != nil {
+		return err
+	}
+	return nil
```
denis256, line 230: err check seems to be redundant 
```suggestion
	return cfg.Save(cliConfigFile)
```

levkohimins: You are right, simplified the code.

terraform/cache/controllers/downloader.go
levkohimins: This is not the error that is returned to the main function, this error is returned to the HTTP client that made the request for this URL. So we will never see this error in the terminal. Therefore, there is no point in wrapping.

terraform/cache/controllers/discovery.go
levkohimins: This is not the error that is returned to the main function, this error is returned to the HTTP client that made the request for this URL. So we will never see this error in the terminal. Therefore, there is no point in wrapping.

cli/provider_cache.go
```diff
@@ -0,0 +1,231 @@
...
+		if err := runTerraformCommand(ctx, opts, terraform.CommandNameProviders, terraform.CommandNameLock); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func runTerraformCommand(ctx context.Context, opts *options.TerragruntOptions, args ...string) error {
+	// We use custom writter in order to trap the log from `terraform providers lock -platform=provider-cache` command, which terraform considers an error, but to us a success.
```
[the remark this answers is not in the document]
levkohimins: Thanks

cli/provider_cache.go
```diff
@@ -0,0 +1,231 @@
...
+		args = append(args, terraform.FlagNameNoColor)
+	}
+
+	cloneOpts := opts.Clone(opts.TerragruntConfigPath)
+	cloneOpts.ErrWriter = errWritter
+	cloneOpts.WorkingDir = opts.WorkingDir
+	cloneOpts.TerraformCliArgs = args
+
+	// If the Terraform error matches `HTTPStatusCacheProviderReg` we ignore it and hide the log from users, otherwise we process the error as is.
+	if err := shell.RunTerraformCommand(ctx, cloneOpts, cloneOpts.TerraformCliArgs...); err != nil && len(errWritter.Msgs()) == 0 {
```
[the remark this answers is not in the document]
levkohimins: Thanks

cli/commands/run-all/command.go
```diff
@@ -17,32 +17,37 @@ import (
 
 const (
 	CommandName = "run-all"
+
+	FlagNameTerragruntProviderCache        = "terragrunt-provider-cache"
```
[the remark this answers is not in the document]
levkohimins: Ah, forgot to remove  after moving it to another file. Thanks.

cli/commands/flags.go
```diff
@@ -8,250 +8,309 @@ import (
...
+			Name:        TerragruntProviderCacheDirFlagName,
+			Destination: &opts.ProviderCacheDir,
+			EnvVar:      TerragruntProviderCacheDirEnvVarName,
+			Usage:       "The path to the cache directory. By default, 'terragrunt/providers' folder in the user cache directory.",
+		},
+		&cli.BoolFlag{
+			Name:        TerragruntProviderCompleteLockFlagName,
+			Destination: &opts.ProviderCacheCompleteLock,
+			EnvVar:      TerragruntProviderCacheCompleteLockEnvVarName,
+			Usage:       "Disables terraform 'plugin_cache_may_break_dependency_lock_file' feature.",
```
brikis98, line 288:
```suggestion
			Usage:       "Don't use 'plugin_cache_may_break_dependency_lock_file' with Terragrunt provider caching. Provider downloads for modules without lock files will be much slower.",
```

docs/_docs/02_features/provider-cache.md
```diff
@@ -0,0 +1,132 @@
...
+
+or the environment variable `TERRAGRUNT_PROVIDER_CACHE_REGISTRY_NAMES`:
+
+``` shell
+TERRAGRUNT_PROVIDER_CACHE=1 \
+TERRAGRUNT_PROVIDER_CACHE_REGISTRY_NAMES=example1.com,example2.com \
+terragrunt apply
+```
+
+Since Terragrunt Cache is essentially a Private Registry server that accepts requests from Terraform, downloads and saves providers to the cache directory, there are a few more flags that are unlikely to be needed, but are useful to know about:
```
brikis98, line 78: I'd move all these docs under a new "Configure the Terragrunt Cache Provider" sub-section (`###` heading) that goes _below_ the "How Terragrunt Provider Caching works" section, as until you read that, it's not clear why there's any server at all!

docs/_docs/02_features/provider-cache.md
```diff
@@ -0,0 +1,132 @@
...
+* Call `terraform providers lock -platform=cache_provider`. When the cache server receives this request, it returns HTTP status _429 Locked_ and starts caching the provider (for any other requests, cache server acts as a proxy):
+  * Create the [lock file](https://en.wikipedia.org/wiki/File_locking) to prevent multiple cache servers from overwriting the same provider cache.
+  * Download the provider from the remote registry, unpack and store into the cache directory or [create a symlink](#Why Cache Server creates symlinks to providers from the user plugins directory) if the required provider exists in the user plugins directory.
+  * Remove the [lock file](https://en.wikipedia.org/wiki/File_locking).
+* Receive the response _429 Locked_ from the cache server, wait until all providers are cached.
+* If the [`terragrunt-provider-cache-complete-lock`](https://terragrunt.gruntwork.io/docs/reference/cli-options/#terragrunt-provider-cache-complete-lock) flag is set, call `terraform providers lock` to generate a complete `.terraform.lock.hcl` file.
+* Run `terragrunt init`, which in turn finds providers in the cache directory and creates symlinks to them.
```
brikis98, line 120:
```suggestion
* Start a server on localhost. This is the _Terragrunt Provider Cache server_.
* Configure Terraform instances to use the cache server as a remote registry:
  * Create local CLI config file `.terragrunt-cache/.terraformrc` that concatenates the user configuration from the Terraform [CLI config file](https://developer.hashicorp.com/terraform/cli/config/config-file) with additional sections:
     * [provider-installation](https://developer.hashicorp.com/terraform/cli/config/config-file#provider-installation) forces Terraform to look for for the required providers in the cache directory and create symbolic links to them, if not found, then request them from the remote registry.
     * [host](https://github.com/hashicorp/terraform/issues/28309) forces Terraform to [forward](#how-forwarding-terraform-requests-through-the-cache-server-works) all provider requests through the cache server.
  * Set environment variables:
     * [TF_PLUGIN_CACHE_MAY_BREAK_DEPENDENCY_LOCK_FILE](https://developer.hashicorp.com/terraform/cli/config/config-file#allowing-the-provider-plugin-cache-to-break-the-dependency-lock-file) allows to generate `.terraform.lock.hcl` files based only on provider hashes from the cache directory.
     * [TF_CLI_CONFIG_FILE](https://developer.hashicorp.com/terraform/cli/config/environment-variables#tf_plugin_cache_dir) sets to use just created local CLI config `.terragrunt-cache/.terraformrc`
     * [TF_TOKEN_*](https://developer.hashicorp.com/terraform/cli/config/config-file#environment-variable-credentials) sets per-remote-registry tokens for authentication to cache server.
* Any time Terragrunt is going to run `init`:
    * Call `terraform providers lock -platform=cache_provider`. This gets Terraform to request all the providers it needs from the Terragrunt Provider Cache server. 
    * The Terragrunt Provider Cache server will download the provider from the remote registry, unpack and store it into the cache directory or [create a symlink](#Why Cache Server creates symlinks to providers from the user plugins directory) if the required provider exists in the user plugins directory. Note that the Terragrunt Provider Cache server will ensure that each unique provider is only ever downloaded and stored on disk once, handling concurrency (from multiple Terraform and Terragrunt instances) correctly. 
    * The Terragrunt Provider Cache server returns the HTTP status _429 Locked_ to Terraform. This is because we do _not_ want Terraform to actually download any providers as a result of calling `terraform providers lock` (as that command does _not_ use any caching); we only use that command to get the list of required providers.
    * At this point, all providers are downloaded and cached, so finally, we run `terragrunt init`, which will find all the providers it needs in the cache, and it'll create symlinks to them nearly instantly, with no additional downloading.
    * Note that if a Terraform module doesn't have a lock file, Terraform does _not_ use the cache, so it would end up downloading all the providers from scratch. To work around this, by default, we use the `plugin_cache_may_break_dependency_lock_file` feature, which, for modules without lock files, will allow Terraform to automatically generate a partial lock file if it finds the providers it needs in the cache, with no additional downloading. This ensures that no additional downloads are necessary, but at the cost of partial lock files. If you wish to disable this feature, set the [`terragrunt-provider-cache-complete-lock`](https://terragrunt.gruntwork.io/docs/reference/cli-options/#terragrunt-provider-cache-complete-lock) flag, and for modules without a lock file, Terragrunt will call `terraform providers lock`, which will give you a complete lock file, at the cost of more downloading and bandwidth.
```

docs/_docs/04_reference/cli-options.md
```diff
@@ -1150,3 +1157,66 @@ When this flag is set, Terragrunt will output its logs in JSON format.
...
 When this flag is set, Terragrunt will wrap Terraform `stdout` and `stderr` in JSON log messages. Works only with `--terragrunt-json-log` flag.
+
+### terragrunt-provider-cache
+
+**CLI Arg**: `--terragrunt-provider-cache`
+**Environment Variable**: `TERRAGRUNT_PROVIDER_CACHE`
+**Commands**:
+- [run-all](#run-all)
+
+Enables provider caching. This forces Terraform to make provider requests through the built-in Provider Cache server.
```
brikis98, line 1168:
```suggestion
Enables Terragrunt's provider caching. This forces Terraform to make provider requests through the built-in Terragrunt Provider Cache server.
```

docs/_docs/04_reference/cli-options.md
```diff
@@ -1150,3 +1157,66 @@ When this flag is set, Terragrunt will output its logs in JSON format.
 **Environment Variable**: `TERRAGRUNT_TF_JSON_LOG` (set to `true`)
 
 When this flag is set, Terragrunt will wrap Terraform `stdout` and `stderr` in JSON log messages. Works only with `--terragrunt-json-log` flag.
+
+### terragrunt-provider-cache
```
brikis98, line 1161: For each of these new provider caching flags, add a link to the dedicated provider caching docs page!

docs/_docs/04_reference/cli-options.md
```diff
@@ -1150,3 +1157,66 @@ When this flag is set, Terragrunt will output its logs in JSON format.
...
+Enables provider caching. This forces Terraform to make provider requests through the built-in Provider Cache server.
+
+### terragrunt-provider-cache-dir
+
+**CLI Arg**: `--terragrunt-provider-cache-dir`
+**Environment Variable**: `TERRAGRUNT_PROVIDER_CACHE_DIR`
+**Commands**:
+- [run-all](#run-all)
+
+The path to the cache directory. By default, `terragrunt/providers` folder in the user cache directory: `$HOME/.cache` on Unix systems, `$HOME/Library/Caches` on Darwin, `%LocalAppData%` on Windows. The file structure of the cache directory is identical to the Terraform [plugin_cache_dir](https://developer.hashicorp.com/terraform/cli/config/config-file#provider-plugin-cache) directory.
```
brikis98, line 1177:
```suggestion
The path to the Terragrunt provider cache directory. By default, `terragrunt/providers` folder in the user cache directory: `$HOME/.cache` on Unix systems, `$HOME/Library/Caches` on Darwin, `%LocalAppData%` on Windows. The file structure of the cache directory is identical to the Terraform [plugin_cache_dir](https://developer.hashicorp.com/terraform/cli/config/config-file#provider-plugin-cache) directory.
```

docs/_docs/04_reference/cli-options.md
```diff
@@ -1150,3 +1157,66 @@ When this flag is set, Terragrunt will output its logs in JSON format.
...
+
+### terragrunt-provider-cache-complete-lock
+
+**CLI Arg**: `--terragrunt-provider-cache-complete-lock`
+**Environment Variable**: `TERRAGRUNT_PROVIDER_CACHE_COMPLETE_LOCK`
+**Commands**:
+- [run-all](#run-all)
+
+Disables Terraform [_plugin_cache_may_break_dependency_lock_file_](https://developer.hashicorp.com/terraform/cli/config/config-file#provider-installation) feature. When this feature is enabled, Terraform generates `.terraform.lock.hcl` files based on the files in the cache directory. If the flag is set Terragrunt disable this feature and before each `init` command runs `terraform providers lock` to generate complete `.terraform.lock.hcl`, which in turn downloads providers from the original registry, which is not fast.
```
brikis98, line 1186:
```suggestion
By default, Terraform does _not_ use the cache for modules without a lock file. This results in lots of extra provider downloading. To work around this, for modules without a lock file, Terragurnt provider caching enables the [_plugin_cache_may_break_dependency_lock_file_](https://developer.hashicorp.com/terraform/cli/config/config-file#provider-installation) feature, which allows Terraform to generate a partial lock file if it finds the providers it needs in the cache. This avoids lots of unnecessary provider downloads, but results in partial lock files. If you wish to disable this feature, set this flag flag, and Terragrunt will run `terraform providers lock` before `init` for modules without lock files, which will generate a complete lock file, but at the cost of more provider downloads.
```

docs/_docs/04_reference/cli-options.md
```diff
@@ -1150,3 +1157,66 @@ When this flag is set, Terragrunt will output its logs in JSON format.
...
+
+### terragrunt-provider-cache-hostname
+
+**CLI Arg**: `--terragrunt-provider-cache-hostname`
+**Environment Variable**: `TERRAGRUNT_PROVIDER_CACHE_HOSTNAME`
+**Commands**:
+- [run-all](#run-all)
+
+The hostname of the built-in Terragrunt Cache server. By default, 'localhost'.
```
brikis98, line 1195:
```suggestion
The hostname of the built-in Terragrunt Provider Cache server. By default, 'localhost'.
```

docs/_docs/04_reference/cli-options.md
```diff
@@ -1150,3 +1157,66 @@ When this flag is set, Terragrunt will output its logs in JSON format.
...
+
+### terragrunt-provider-cache-port
+
+**CLI Arg**: `--terragrunt-provider-cache-port`
+**Environment Variable**: `TERRAGRUNT_PROVIDER_CACHE_PORT`
+**Commands**:
+- [run-all](#run-all)
+
+The port of the built-in Terragrunt Cache server. By default, assigned automatically.
```
brikis98, line 1204:
```suggestion
The port of the built-in Terragrunt Provider Cache server. By default, assigned automatically.
```

docs/_docs/04_reference/cli-options.md
```diff
@@ -1150,3 +1157,66 @@ When this flag is set, Terragrunt will output its logs in JSON format.
...
+
+### terragrunt-provider-cache-token
+
+**CLI Arg**: `--terragrunt-provider-cache-token`
+**Environment Variable**: `TERRAGRUNT_PROVIDER_CACHE_TOKEN`
+**Commands**:
+- [run-all](#run-all)
+
+The Token for authentication on the built-in Terragrunt Cache server. By default, assigned automatically.
```
brikis98, line 1213:
```suggestion
The Token for authentication on the built-in Terragrunt Provider Cache server. By default, assigned automatically.
```

docs/_docs/04_reference/cli-options.md
```diff
@@ -1150,3 +1157,66 @@ When this flag is set, Terragrunt will output its logs in JSON format.
...
+
+### terragrunt-provider-cache-registry-names
+
+**CLI Arg**: `--terragrunt-provider-cache-registry-names`
+**Environment Variable**: `TERRAGRUNT_PROVIDER_CACHE_REGISTRY_NAMES`
+**Commands**:
+- [run-all](#run-all)
+
+The list of remote registries to cached by Terragrunt Cache server. By default, 'registry.terraform.io', 'registry.opentofu.org'.
```
brikis98, line 1222:
```suggestion
The list of remote registries to cached by Terragrunt Provider Cache server. By default, 'registry.terraform.io', 'registry.opentofu.org'.
```

options/options.go
```diff
@@ -269,6 +273,30 @@ type TerragruntOptions struct {
 
 	// Disable listing of dependent modules in render json output
 	JsonDisableDependentModules bool
+
+	// Enables provider cache using the built-in private registry.
```
brikis98, line 277: OK, I'm going to stop fixing these on a one off basis, and just give a more general comment:

Let's standardize on "Terragrunt Provider Cache" everywhere. Not just "provider cache" or "terragrunt cache," but the full "Terragrunt Provider Cache" to make it clear this is a TG specific feature.

terraform/cache/config.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+const (
+	defaultHostname        = "localhost"
+	defaultShutdownTimeout = time.Second * 30
+)
+
+type Option func(Config) Config
+
+func WithHostname(hostname string) Option {
```
brikis98, line 16: Do we have a test that checks this server can't be accessed without proper auth?

cli/commands/flags.go
```diff
@@ -8,250 +8,309 @@ import (
...
 			Destination: &opts.DisableCommandValidation,
 			EnvVar:      "TERRAGRUNT_DISABLE_COMMAND_VALIDATION",
 			Usage:       "When this flag is set, Terragrunt will not validate the terraform command.",
 		},
+		// provider cache flags
+		&cli.BoolFlag{
+			Name:        TerragruntProviderCacheFlagName,
+			Destination: &opts.ProviderCache,
+			EnvVar:      TerragruntProviderCacheEnvVarName,
+			Usage:       "Enables provider caching.",
```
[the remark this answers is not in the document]
levkohimins: Agree. Thanks.

cli/commands/flags.go
```diff
@@ -8,250 +8,309 @@ import (
...
+		},
+		&cli.GenericFlag[string]{
+			Name:        TerragruntProviderCacheDirFlagName,
+			Destination: &opts.ProviderCacheDir,
+			EnvVar:      TerragruntProviderCacheDirEnvVarName,
+			Usage:       "The path to the cache directory. By default, 'terragrunt/providers' folder in the user cache directory.",
```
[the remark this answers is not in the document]
levkohimins: Thanks.

cli/commands/flags.go
```diff
@@ -8,250 +8,309 @@ import (
...
+			Name:        TerragruntProviderCompleteLockFlagName,
+			Destination: &opts.ProviderCacheCompleteLock,
+			EnvVar:      TerragruntProviderCacheCompleteLockEnvVarName,
+			Usage:       "Disables terraform 'plugin_cache_may_break_dependency_lock_file' feature.",
+		},
+		&cli.GenericFlag[string]{
+			Name:        TerragruntProviderCacheTokenFlagName,
+			Destination: &opts.ProviderCacheToken,
+			EnvVar:      TerragruntProviderCacheTokenEnvVarName,
+			Usage:       "The Token for authentication on the built-in Terragrunt Cache server. By default, assigned automatically.",
```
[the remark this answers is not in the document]
levkohimins: Thanks.

cli/commands/flags.go
```diff
@@ -8,250 +8,309 @@ import (
...
+		},
+		&cli.GenericFlag[string]{
+			Name:        TerragruntProviderCacheHostnameFlagName,
+			Destination: &opts.ProviderCacheHostname,
+			EnvVar:      TerragruntProviderCacheHostnameEnvVarName,
+			Usage:       "The hostname of the built-in Terragrunt Cache server. By default, 'localhost'.",
```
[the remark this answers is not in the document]
levkohimins: Thanks.

cli/commands/flags.go
```diff
@@ -8,250 +8,309 @@ import (
...
+		},
+		&cli.GenericFlag[int]{
+			Name:        TerragruntProviderCachePortFlagName,
+			Destination: &opts.ProviderCachePort,
+			EnvVar:      TerragruntProviderCachePortEnvVarName,
+			Usage:       "The port of the built-in Terragrunt Cache server. By default, assigned automatically.",
```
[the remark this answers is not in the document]
levkohimins: Thanks.

cli/commands/flags.go
```diff
@@ -8,250 +8,309 @@ import (
...
 
-	FlagNameHelp    = "help"
-	FlagNameVersion = "version"
+	// provider cache flags/envs
+	TerragruntProviderCacheFlagName                = "terragrunt-provider-cache"
+	TerragruntProviderCacheEnvVarName              = "TERRAGRUNT_PROVIDER_CACHE"
+	TerragruntProviderCacheDirFlagName             = "terragrunt-provider-cache-dir"
+	TerragruntProviderCacheDirEnvVarName           = "TERRAGRUNT_PROVIDER_CACHE_DIR"
+	TerragruntProviderCompleteLockFlagName         = "terragrunt-provider-cache-complete-lock"
+	TerragruntProviderCacheCompleteLockEnvVarName  = "TERRAGRUNT_PROVIDER_CACHE_COMPLETE_LOCK"
```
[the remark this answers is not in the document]
levkohimins: I agree with you, by the way I didn't like "complete lock" too :)  I just couldn't come up with the better name.  Renamed everywhere.

docs/_docs/02_features/provider-cache.md
```diff
@@ -0,0 +1,132 @@
...
+
+#### Why Terraform's built-in provider caching doesn't work
+
+Terraform has provider caching feature [Provider Plugin Cache](https://developer.hashicorp.com/terraform/cli/config/config-file#provider-plugin-cache), that does the job well until you run multiple Terraform processes simultaneously. Then the Terraform processes begin conflict by overwriting each other’s cache, which causes the error like `Error: Failed to install provider`. By default, to speed up processing, Terragrunt runs multiple Terraform modules simultaneously, which in turn prevents the use of [Provider Plugin Cache](https://developer.hashicorp.com/terraform/cli/config/config-file#provider-plugin-cache) to avoid conflicts or force Terragrunt to run Terraform modules sequentially by specifying the `--terragrunt-parallelism 1` flag, which significantly slows down the work.
```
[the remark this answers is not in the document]
levkohimins: Thanks.

docs/_docs/02_features/provider-cache.md
levkohimins: I did as you said. Thank you.

docs/_docs/02_features/provider-cache.md
levkohimins: Thanks.

docs/_docs/04_reference/cli-options.md
levkohimins: Thanks.

docs/_docs/04_reference/cli-options.md
levkohimins: Added. Thanks.

docs/_docs/04_reference/cli-options.md
levkohimins: Thanks.

docs/_docs/04_reference/cli-options.md
levkohimins: Thanks.

docs/_docs/04_reference/cli-options.md
levkohimins: Thanks.

docs/_docs/04_reference/cli-options.md
levkohimins: Thanks.

options/options.go
levkohimins: Thanks for all these suggestions! I have replaced the rest, I hope I didn’t miss anything.

levkohimins:
> This is looking very close! Most of my comments are just related to docs.

Thanks for the review!

> One important question: since we return 429 for the `providers lock` command, does TF still have a chance to check the hashes of the providers it uses? Or does our cache somehow result in that hash check being bypassed?

Terragrunt Provider Cache returns 429, for the command with `terraform providers lock -platform=cache_provider`  TF simply exits with an error and does nothing, but TG catches this error and hides it from the user, any other errors other than 429 are shown to the user as is. 

_If TG or an user in CLI, specify `./terragrunt providers lock` without the fake platform `-platform=cache_provider` it works as it was intended - updates the lock file._

Probably the answer to your question will be, 
then TG run `terraform init` command which should create hashes for all providers,
TF tries to make them from the cache directory, if the cache was not found downloads providers directly from remote registries, this can happen if the user uses registries other than the default and didn't specify them using the `--terragrunt-provider-cache-registry-names` flag.
If TF can't create a hash for the provider, it will simply return an error.  Therefore, TG did not check hashes in the lock file.

And for the `auto-init` function, I added a check for the presence of a lock file in the working directory.
In other words, caching (and all subsequent operations such as creating a lock file) starts only when the user explicitly runs the `init` command or using the `auto-init` function.

brikis98:
> > One important question: since we return 429 for the `providers lock` command, does TF still have a chance to check the hashes of the providers it uses? Or does our cache somehow result in that hash check being bypassed?
> 
> Terragrunt Provider Cache returns 429, for the command with `terraform providers lock -platform=cache_provider` TF simply exits with an error and does nothing, but TG catches this error and hides it from the user, any other errors other than 429 are shown to the user as is.
> 
> _If TG or an user in CLI, specify `./terragrunt providers lock` without the fake platform `-platform=cache_provider` it works as it was intended - updates the lock file._
> 
> Probably the answer to your question will be, then TG run `terraform init` command which should create hashes for all providers, TF tries to make them from the cache directory, if the cache was not found downloads providers directly from remote registries, this can happen if the user uses registries other than the default and didn't specify them using the `--terragrunt-provider-cache-registry-names` flag. If TF can't create a hash for the provider, it simply returns an error. Therefore, TG does not check hashes in the lock file.
> 
> And for the `auto-init` feature, I added a check for the presence of the lock file in the working directory. In other words, caching (and all subsequent operations such as creating a lock file) starts only when the user explicitly runs the `init` command or using the `auto-init` feature.

Hm, I'm not sure this answers my question. Normally, when you run `terraform init`, it does the following:

1. Go out and download providers into the cache (if not there already) and record their hash into the lock file.
2. Check that the hash of the binary matches the hash in the lock file.

I want to make sure we're not losing the check in (2) when provider caching is enabled. 

Thinking through the use cases:

1. If there is a lock file already, we run `terraform providers lock`, download the providers into the cache, and then run `terraform init`, which will check the providers in the cache against the ones in the lock file. So I believe this should work as expected.
2. If there is no lock file, then there are two possibilities:
    1. By default, we run `terraform providers lock` as above, then `terraform init`, but on this `init`, we tell Terraform to create the lock file _from the data in the cache_. But did anyone check that the binaries we downloaded into the cache actually match the hashes in their hash files? If Terragrunt provider cache doesn't check hashes, and Terraform creates the partial lock file from the cache, is there any hash checking?
    2. If the user passes in a flag, then we have the module create the full lock file, and so it seems like this will do a full hash check, and work as expected.

docs/_docs/02_features/provider-cache.md
```diff
@@ -0,0 +1,134 @@
...
+     * [TF_CLI_CONFIG_FILE](https://developer.hashicorp.com/terraform/cli/config/environment-variables#tf_plugin_cache_dir) sets to use just created local CLI config `.terragrunt-cache/.terraformrc`
+     * [TF_TOKEN_*](https://developer.hashicorp.com/terraform/cli/config/config-file#environment-variable-credentials) sets per-remote-registry tokens for authentication to Terragrunt Provider Cache server.
+* Any time Terragrunt is going to run `init`:
+    * Call `terraform providers lock -platform=cache_provider`. This gets Terraform to request all the providers it needs from the Terragrunt Provider Cache server.
+    * The Terragrunt Provider Cache server will download the provider from the remote registry, unpack and store it into the cache directory or [create a symlink](#Why Cache Server creates symlinks to providers from the user plugins directory) if the required provider exists in the user plugins directory. Note that the Terragrunt Provider Cache server will ensure that each unique provider is only ever downloaded and stored on disk once, handling concurrency (from multiple Terraform and Terragrunt instances) correctly.
```
brikis98, line 91: Broken link:

```suggestion
    * The Terragrunt Provider Cache server will download the provider from the remote registry, unpack and store it into the cache directory or [create a symlink](#reusing-providers-from-the-user-plugins-directory) if the required provider exists in the user plugins directory. Note that the Terragrunt Provider Cache server will ensure that each unique provider is only ever downloaded and stored on disk once, handling concurrency (from multiple Terraform and Terragrunt instances) correctly.
```

brikis98 (commented):
One small broken link and one question left!

levkohimins:
> Hm, I'm not sure this answers my question. Normally, when you run `terraform init`, it does the following:
> 
> 1. Go out and download providers into the cache (if not there already) and record their hash into the lock file.
> 2. Check that the hash of the binary matches the hash in the lock file.
> 
> I want to make sure we're not losing the check in (2) when provider caching is enabled.
> 
> Thinking through the use cases:
> 
> 1. If there is a lock file already, we run `terraform providers lock`, download the providers into the cache, and then run `terraform init`, which will check the providers in the cache against the ones in the lock file. So I believe this should work as expected.
> 2. If there is no lock file, then there are two possibilities:
>    
>    1. By default, we run `terraform providers lock` as above, then `terraform init`, but on this `init`, we tell Terraform to create the lock file _from the data in the cache_. But did anyone check that the binaries we downloaded into the cache actually match the hashes in their hash files? If Terragrunt provider cache doesn't check hashes, and Terraform creates the partial lock file from the cache, is there any hash checking?
>    2. If the user passes in a flag, then we have the module create the full lock file, and so it seems like this will do a full hash check, and work as expected.

without `--terragrunt-provider-cache-disable-partial-lock-file`, TG does

`terraform providers lock -platform=cache_provider` 
`terraform init` + env `TF_PLUGIN_CACHE_MAY_BREAK_DEPENDENCY_LOCK_FILE`

In this case, TF trusts the files from the cache. TG also doesn't check the hashes. So there are no checks. I don't know what kind of check we can do here.


with `--terragrunt-provider-cache-disable-partial-lock-file`, TG does

`terraform providers lock -platform=cache_provider` _caches  archive files of the providers_
`terraform providers lock`  requests all three files from the Cache server, the large one Cache returns from the cache and rest two handles via  proxy. Ultimately a complete lock file will be created.
`terraform init` If the hashes do not match the unpacked provider from the cache dir, TF returns an error.

docs/_docs/02_features/provider-cache.md
levkohimins: Thanks.

brikis98:
> without `--terragrunt-provider-cache-disable-partial-lock-file`, TG does
> 
> * `terraform providers lock -platform=cache_provider`
> * `terraform init` + env `TF_PLUGIN_CACHE_MAY_BREAK_DEPENDENCY_LOCK_FILE`
> 
> In this case, TF trusting files from the cache, creates the lock file. TG also doesn't check the hashes. So there are no checks. I don't know what kind of check we can do here.

Yea, this is the part that worries me: the lack of hash/signature checking. Is there any way to ensure that happens?

levkohimins:
Additional: 


> > without `--terragrunt-provider-cache-disable-partial-lock-file`, TG does
> > 
> > * `terraform providers lock -platform=cache_provider`
> > * `terraform init` + env `TF_PLUGIN_CACHE_MAY_BREAK_DEPENDENCY_LOCK_FILE`
> > 
> > In this case, TF trusting files from the cache, creates the lock file. TG also doesn't check the hashes. So there are no checks. I don't know what kind of check we can do here.
> 
> Yea, this is the part that worries me: the lack of hash/signature checking. Is there any way to ensure that happens?

Good question. If we don't have a full lock file, we need to check the binaries' providers hashes ourselves. Do this every time TG starts or only at the moment when we cache the provider. In theory, we could carry out a similar check of what TG should do when creating lock files. But to say for sure, I need to investigate about it.
But this is definitely not a quick task, so it’s probably best to implement this separately.

levkohimins:
> Additional:
> 
> > > without `--terragrunt-provider-cache-disable-partial-lock-file`, TG does
> > > 
> > > * `terraform providers lock -platform=cache_provider`
> > > * `terraform init` + env `TF_PLUGIN_CACHE_MAY_BREAK_DEPENDENCY_LOCK_FILE`
> > > 
> > > In this case, TF trusting files from the cache, creates the lock file. TG also doesn't check the hashes. So there are no checks. I don't know what kind of check we can do here.
> > 
> > 
> > Yea, this is the part that worries me: the lack of hash/signature checking. Is there any way to ensure that happens?
> 
> Good question. If we don't have a full lock file, we need to check the binaries' providers hashes ourselves. Do this every time TG starts or only at the moment when we cache the provider. In theory, we could carry out a similar check of what TF does when creating lock files. But to say for sure, I need to investigate this. But this is definitely not a quick task, so it’s probably best to implement this separately.



> Additional:
> 
> > > without `--terragrunt-provider-cache-disable-partial-lock-file`, TG does
> > > 
> > > * `terraform providers lock -platform=cache_provider`
> > > * `terraform init` + env `TF_PLUGIN_CACHE_MAY_BREAK_DEPENDENCY_LOCK_FILE`
> > > 
> > > In this case, TF trusting files from the cache, creates the lock file. TG also doesn't check the hashes. So there are no checks. I don't know what kind of check we can do here.
> > 
> > 
> > Yea, this is the part that worries me: the lack of hash/signature checking. Is there any way to ensure that happens?
> 
> Good question. If we don't have a full lock file, we need to check the binaries' providers ourselves. Do this every time TG starts or only at the moment when we cache the provider. In theory, we could carry out a similar check of what TF does when creating lock files, to say for sure, I need to investigate this. But this is definitely not a quick task, so it’s probably best to implement this separately.

If this is critical, we could make this flag `--terragrunt-provider-cache-disable-partial-lock-file` enabled by default, and add another `--terragrunt-provider-cache-enable-partial-lock-file`

brikis98:
> > > > without `--terragrunt-provider-cache-disable-partial-lock-file`, TG does
> > > > 
> > > > * `terraform providers lock -platform=cache_provider`
> > > > * `terraform init` + env `TF_PLUGIN_CACHE_MAY_BREAK_DEPENDENCY_LOCK_FILE`
> > > > 
> > > > In this case, TF trusting files from the cache, creates the lock file. TG also doesn't check the hashes. So there are no checks. I don't know what kind of check we can do here.
> > > 
> > > 
> > > Yea, this is the part that worries me: the lack of hash/signature checking. Is there any way to ensure that happens?
> > 
> > 
> > Good question. If we don't have a full lock file, we need to check the binaries' providers hashes ourselves. Do this every time TG starts or only at the moment when we cache the provider. In theory, we could carry out a similar check of what TF does when creating lock files. But to say for sure, I need to investigate this. But this is definitely not a quick task, so it’s probably best to implement this separately.
> 
> If this is critical, we could make this flag `--terragrunt-provider-cache-disable-partial-lock-file` enabled by default, and add another `--terragrunt-provider-cache-enable-partial-lock-file`

I'd prefer to keep partial lock files enabled by default, as the whole point of provider caching is to go much faster.

That said, hash & signature checking is important. 

With the partial lock file stuff, does TF not check the signature or has _at all_?

levkohimins:
> > > > > without `--terragrunt-provider-cache-disable-partial-lock-file`, TG does
> 
> > > > > 
> 
> > > > > * `terraform providers lock -platform=cache_provider`
> 
> > > > > * `terraform init` + env `TF_PLUGIN_CACHE_MAY_BREAK_DEPENDENCY_LOCK_FILE`
> 
> > > > > 
> 
> > > > > In this case, TF trusting files from the cache, creates the lock file. TG also doesn't check the hashes. So there are no checks. I don't know what kind of check we can do here.
> 
> > > > 
> 
> > > > 
> 
> > > > Yea, this is the part that worries me: the lack of hash/signature checking. Is there any way to ensure that happens?
> 
> > > 
> 
> > > 
> 
> > > Good question. If we don't have a full lock file, we need to check the binaries' providers hashes ourselves. Do this every time TG starts or only at the moment when we cache the provider. In theory, we could carry out a similar check of what TF does when creating lock files. But to say for sure, I need to investigate this. But this is definitely not a quick task, so it’s probably best to implement this separately.
> 
> > 
> 
> > If this is critical, we could make this flag `--terragrunt-provider-cache-disable-partial-lock-file` enabled by default, and add another `--terragrunt-provider-cache-enable-partial-lock-file`
> 
> 
> 
> I'd prefer to keep partial lock files enabled by default, as the whole point of provider caching is to go much faster.
> 
> 
> 
> That said, hash & signature checking is important. 
> 
> 
> 
> With the partial lock file stuff, does TF not check the signature or has _at all_?

For new entries in the lock file, the signature is not checked at all. To check, it needs to download the rest of the two files, but it doesn’t do this with partial lock, _but_ if the hash already exists in the lock file and the provider binary is modified, it shows an error - mismatch hashes.

levkohimins:
> * `terraform lock providers` requests providers even they already exist in the lock file, which is not effective if we assume that the lock file already contains 9 providers (hashes) and we need to add a hash for new one.

I have an idea to solve the issue above, is to use `terraform init` instead of `terraform lock providers  -platform=cache_provider`, on the same principle, in two requests.
* run `terraform init` without `provider_installation`, which will force terraform to send requests to get only those providers that are not in the lock file. The cache server:
  * Returns the 429 response. 
  * Downloads the provider if necessary (what has already been implemented), but each time it downloads the signature and hashes to check the archive.
* run `TF_PLUGIN_CACHE_MAY_BREAK_DEPENDENCY_LOCK_FILE=1 terraform init` with `provider_installation` to create provider symlinks from the cache directory.

_Replacing `terraform init` with `terraform lockprovider -platform=cache_provider` itself should be a simple task since almost nothing needs to be changed._

levkohimins:
In any case, I would suggest merging this pull request first and then making the changes separately. Since after each new update in the `master` branch I have to resolve conflicts.

brikis98 (approved):
Roger. PR approved. Let's file a bug and figure out the signature/hash check thing next!

levkohimins:
Thanks for the review! Created issue #3047.

Merging and creating a new release.

levkohimins:
Resolved in [v0.56.4](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.56.4) release. Make sure to read [Provider Caching](https://terragrunt.gruntwork.io/docs/features/provider-cache/).

jmreicha:
Hey @levkohimins great work! Just want to make sure I fully understand - this solution addresses the concurrency race condition present in Terraform where providers fail to download if multiple downloads happen at the same time, is that correct?

levkohimins:
> Hey @levkohimins great work! Just want to make sure I fully understand - this solution addresses the concurrency race condition present in Terraform where providers fail to download if multiple downloads happen at the same time, is that correct?

Hey @jmreicha, you are absolutely right.

jmreicha:
Amazing. This is huge, really good stuff guys.

PR #3004, opened 2024-03-15
Add configuration options for AWS S3 server access logging

### LINKED ISSUE #3003: Expose additional configuration options for AWS S3 server access logging

In our environment, the AWS S3 bucket where S3 access logs should be stored is pre-created by the "central" team, and we don't have permissions to make any changes to it. Therefore, we don't need to:
- Create it: The current code creates the logging bucket only if it doesn't exist, which is already appropriate.
- Change configuration of the logging bucket, such as ACL, public access, encryption, or TLS enforcement.

Currently, we cannot configure [server access logging](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerLogs.html) because we have to use the pre-created bucket. Additionally, it would be beneficial to have the option to configure `PartitionDateSource` in the logging configuration.
Example:

```
{
  "LoggingEnabled": {
    "TargetBucket": "<bucket_name> ",
    "TargetPrefix": "",
    "TargetObjectKeyFormat": {
      "PartitionedPrefix": {
        "PartitionDateSource": "EventTime"
      }
    }
  }
}
```

There is also a **small bug** because the current code silently fails on the permission error when attempting to configure the logging bucket.

The only alternative is to configure access logging outside of Terragrunt.

findmyname666:
## Description

Fixes #3003.

In our environment the S3 bucket where should be stored S3 access logs is pre-created by "central" team and we don't have permissions to do any changes. In such case we don't have to:
- Create it - the current code creates the logging bucket only if it doesn't exist. No change is needed.
- Configure logging bucket e.g. ACL, public access, encryption, tls enforcement - adding option to skip the configuration for access log bucket.
- PartitionDateSource - Adding one more option to allow users to configure `PartitionDateSource`.

Additionally 1) adding error logging into the function `configureAccessLogBucket`. Without it I wasn't able to detect why the policy isn't configured. The error is returned but ignored. 2) Adding code to detect if the logging policy / configuration was change. The previous code just check if some policy is enabled.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

### Migration Guide

N/A.


findmyname666:
I did typo in a branch name so I created a new PR #3006.

PR #3005, opened 2024-03-17
Scaffold command, fix unable to handle https URL

### LINKED ISSUE #2987: Scaffold command doesn't handle "https" in a URL correctly, and also doesn't handle errors correctly
**Describe the bug**

1. The `scaffold` command can handle `github.com` URLs, but not `https://github.com` URLs. 
2. When the `scaffold` command fails to handle a URL, it logs a warning, but then proceeds with code generation anyway, so you end up with a largely empty `terragrunt.hcl`.

**To Reproduce**

```hcl
terragrunt scaffold https://github.com/gruntwork-io/terragrunt-infrastructure-modules-example//modules/s3-bucket
```

In the logs, you'll briefly see:

```
WARN[0000] Failed to parse url https://github.com/gruntwork-io/terragrunt-infrastructure-modules-example/tree/master/modules/s3-bucket
WARN[0000] Failed to parse module url https://github.com/gruntwork-io/terragrunt-infrastructure-modules-example/tree/master/modules/s3-bucket
fatal: remote error:
 gruntwork-io/terragrunt-infrastructure-modules-example/tree/master/modules/s3-bucket is not a valid repository name
Visit https://support.github.com/ for help
WARN[0001] Failed to find last release tag for https://github.com/gruntwork-io/terragrunt-infrastructure-modules-example/tree/master/modules/s3-bucket
```

Then the screen is cleared and you get:

```
INFO[0001] Scaffolding a new Terragrunt module https://github.com/gruntwork-io/terragrunt-infrastructure-modules-example/tree/master/modules/s3-bucket to /tmp/test-scaffolding
INFO[0001] Running boilerplate generation to /tmp/test-scaffolding
```

And you then end up with a `terragrunt.hcl` where none of the required or optional input variables have been filled in.

**Expected behavior**

1. If `scaffold` can't process a URL, it should show an error and exit.
2. That said, since users are likely to copy/paste URLs from their browser, we should probably make URLs with the `https://` prefix work. FWIW, I believe the `catalog` command _does_ handle the "https://" prefix OK, so perhaps we can reuse the code from there?

**Versions**
- Terragrunt version: v0.55.13

levkohimins:
## Description

Fixes the issue where the `scaffold` command is unable to parse a URL with the http(s) scheme.

Fixes #2987.

### DIFF
--- a/terraform/source.go
+++ b/terraform/source.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"net/url"
 	"os"
-	"path"
 	"path/filepath"
 	"regexp"
 	"strings"
@@ -19,8 +18,8 @@ import (
 )
 
 var (
-	forcedRegexp            = regexp.MustCompile(`^([A-Za-z0-9]+)::(.+)$`)
-	secondLevelDomainRegexp = regexp.MustCompile(`(?i)\.?([^.]+\.[^.]+)$`)
+	forcedRegexp     = regexp.MustCompile(`^([A-Za-z0-9]+)::(.+)$`)
+	httpSchemeRegexp = regexp.MustCompile(`(?i)https?://`)
 )
 
 const matchCount = 2
@@ -192,6 +191,9 @@ func NewSource(source string, downloadDir string, workingDir string, logger *log
 // Convert the given source into a URL struct. This method should be able to handle all source URLs that the terraform
 // init command can handle, parsing local file paths, Git paths, and HTTP URLs correctly.
 func ToSourceUrl(source string, workingDir string) (*url.URL, error) {
+	// we need to remove the http(s) scheme to allow `getter.Detect` to add the source type
+	source = httpSchemeRegexp.ReplaceAllString(source, "")
+
 	// The go-getter library is what Terraform's init command uses to download source URLs. Use that library to
 	// parse the URL.
 	rawSourceUrlWithGetter, err := getter.Detect(source, workingDir, getter.Detectors)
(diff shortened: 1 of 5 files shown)


autero1 (dismissed):
LGTM!

terraform/source_test.go
```diff
@@ -54,12 +54,41 @@ func TestSplitSourceUrl(t *testing.T) {
...
 		})
 	}
 }
 
+func TestToSourceUrl(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		sourceURL         string
+		workingDir        string
```
denis256, line 68: looks like `workingDir` is empty in each test case, do we need it in the struct?

levkohimins: Thanks. Added a few test cases with the workingDir values.

denis256 (dismissed):

terraform/source.go
```diff
@@ -195,7 +195,16 @@ func ToSourceUrl(source string, workingDir string) (*url.URL, error) {
...
+	sourceUrl, err := parseSourceUrl(rawSourceUrlWithGetter)
+	if err != nil {
+		return nil, err
+	}
+
+	// explicitly add prefix git:: for the http scheme to avoid further errors
+	if strings.HasPrefix(sourceUrl.Scheme, "http") {
+		sourceUrl.Scheme = "git::" + sourceUrl.Scheme
+	}
+	return sourceUrl, nil
```
brikis98, line 207: How do we know that an http URL is always Git? Terraform supports a variety of [source URLs](https://developer.hashicorp.com/terraform/language/modules/sources), including some http URLs that are _not_ Git.

levkohimins: Here we add the default _git_ protocol if the user has not specified anything. Otherwise, we will then get an error when we ask go-getter to clone a repository.
I just thought, that we should add `git` it not only to `https://`, but also to `ssh://`

levkohimins: without default protocol `getter.Get` returns 
```
 *errors.errorString error downloading 'https://github.com/gruntwork-io/terraform-aws-lambda': bad response code: 404
```

brikis98: Hm, I'm not sure that's right. I think we need to be more picky and do checks based on the supported [module source URLs](https://developer.hashicorp.com/terraform/language/modules/sources):

- If it starts with `http` and the domain is `github.com`, `gitlab.com`, or `bitbucket.org`, or the URL contains `.git`, then I think we can safely prepend `git::`.
- If it starts with `http` and the domain includes `amazonaws.com`, we prepend `s3::`.
- If it starts with `http` and the domain includes `googleapis.com`, we prepend `gcs::`.

levkohimins: Ah, understood. It's redone, thank you.

terraform/source.go
```diff
@@ -195,7 +199,57 @@ func ToSourceUrl(source string, workingDir string) (*url.URL, error) {
 		return nil, errors.WithStackTrace(err)
 	}
 
-	return parseSourceUrl(rawSourceUrlWithGetter)
+	sourceUrl, err := parseSourceUrl(rawSourceUrlWithGetter)
```
brikis98, line 202: Sorry I didn't notice this before, but I'm noticing you're passing in `rawSourceUrlWithGetter` to this method. The `WithGetter` part suggests to me this is a "getter" as parsed by the `go-getter` library. is it possible that the getter is _already_ set to the value we need, and we just need to prepend it at the end? In other words, we don't need to write our own custom logic to figure out the getter?

terraform/source.go
```diff
@@ -195,7 +199,57 @@ func ToSourceUrl(source string, workingDir string) (*url.URL, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return PrependSourceType(sourceUrl), nil
+}
+
+// PrependSourceType prepends one of the "s3, gcs3, hg, git" source type to the http(s)/ssh scheme, if the user has omittied it.
```
brikis98, line 210: NIT: link to module source URL docs or maybe go-getter docs as context on what these are.

terraform/source.go
levkohimins: Understand your thought, because I was thinking the same, since this is the same Golang package, it would be logical that the result from `getter.Detect` could be immediately used for `getter.Get`, but unfortunately this is not the case. I tested, `getter.Detect` does prepend a source type if we only pass URL without scheme:

 `github.com/gruntwork-io/terraform-aws-lambda` -> `git::https://github.com/gruntwork-io/terraform-aws-lambda.git`

and doesn't if there's the scheme:

`https://github.com/gruntwork-io/terraform-aws-lambda` -> `https://github.com/gruntwork-io/terraform-aws-lambda.git`

I also tested to pass `https://github.com/gruntwork-io/terraform-aws-lambda` to `getter.Get`, and got the error

```
'https://github.com/gruntwork-io/terraform-aws-lambda': bad response code: 404
```

terraform/source.go
levkohimins: Updated the comment. Thanks.

terraform/source_test.go
```diff
@@ -54,12 +54,50 @@ func TestSplitSourceUrl(t *testing.T) {
...
 			require.NoError(t, err)
 
-			assert.Equal(t, testCase.expectedRootRepo, actualRootRepo.String())
+			assert.Equal(t, testCase.expectedSo, actualRootRepo.String())
 			assert.Equal(t, testCase.expectedModulePath, actualModulePath)
 		})
 	}
 }
 
+func TestPrependSourceType(t *testing.T) {
```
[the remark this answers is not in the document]
levkohimins: Thanks.

terraform/source.go
levkohimins: The only way is if we remove the schema before passing it to `Detect`.

levkohimins: There is a new version 2.2.1 of `go-getter`, but it is very different from the 1.7.1 version we are using now.
We'll have to rewrite the code to use it. I don’t yet know how it behaves with the http scheme. Should I check this?

brikis98: Thanks for looking into it. Prob not worth a big upgrade now. Let's go with what you have here.

brikis98 (approved):
LGTM

levkohimins:
> LGTM

Thanks for the review!

Issue #2980, opened 2024-03-04
Extremely poor performance in v0.55.5+


syphernl:
**Describe the bug**
It seems that the performance since v0.55.6 is very bad. A simple `terragrunt apply` without caches (that is with `.terragrunt-cache` and `.terraform.lock.hcl` recursively removed) is taking ages.  
What took about a minute on v0.55.5 didn't even finish after 10 minutes on v0.55.6.

I then downgraded back to v0.55.5 and it worked again.
It seems like it is getting caught in some kind of cyclic redundancy as it keeps looping over the same modules over and over again ("Evaluated X locals (remaining XX)")

**Versions**
- Terragrunt version: v0.55.6
- Terraform version: 1.7.4
- Environment details Ubuntu 23.10

**Additional context**
- Cause could be the changes done in #2957 to fix #2955
- Other users are reporting performance issues as well in #2806

syphernl:
Renovate planned an upgrade of Terragrunt today after I downgraded it yesterday. 
One of the `plan` jobs that is impacted by this change took 25 minutes (!) to execute where it would normally take 1~2 minutes at most.

crisbal:
I have seen similar behaviors with "noticeable" slowdowns across various different releases. For the moment I am back using version `0.50.14` because after that version I begin to notice slowdowns.

seanmacisaac:
My experience is that the slow down is far worse in the examples directory of our infrastructure-modules repo, where there is a long pause before it does any output.  I’m on 55.1 now due to an earlier problem with dependencies that started in 55.4.  I’ll try going back to 50.14.

denis256:
Hi,
will be helpful to get feedback on changes from https://github.com/gruntwork-io/terragrunt/pull/2983 to see if performance is improved for real users

RafaelWO:
Hi @denis256, I noticed the same performance issue as @syphernl and was also able to pin it down to the release of [v0.55.6](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.55.6).

I did some "benchmarking" across different versions of Terragrunt by running `terragrunt run-all init` (no `plan` or `apply`)  in a project with 35 modules and the performance issues do not seem to be solved until the latest version v0.56.5.

I created a plot of the timespans between every module that shows a massive increase of runtime for the final modules (looks like it's increasing exponentially). 

[...]

PR #3118, opened 2024-05-02
feat: add "recurse-dependencies" option

michaeldelago:
## Description

This change reverts to the pre 0.55.6 behavior of not recursing dependencies for all operations. In case a user desires this behavior, they can enable it with `--recurse-dependencies` or `TERRAGRUNT_RECURSE_DEPENDENCIES`.

The changes in this PR should mitigate the performance issues noticed in #2980

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added `--recurse-dependencies` flag. This flag recursively parses dependencies instead of parsing only the original terragrunt module's dependencies.

### DIFF
--- a/cli/commands/flags.go
+++ b/cli/commands/flags.go
@@ -42,6 +42,7 @@ const (
 	TerragruntFailOnStateBucketCreationFlagName      = "terragrunt-fail-on-state-bucket-creation"
 	TerragruntDisableBucketUpdateFlagName            = "terragrunt-disable-bucket-update"
 	TerragruntDisableCommandValidationFlagName       = "terragrunt-disable-command-validation"
+	TerragruntRecurseDependenciesFlagName            = "terragrunt-recurse-dependencies"
 
 	TerragruntOutDirFlagEnvVarName = "TERRAGRUNT_OUT_DIR"
 	TerragruntOutDirFlagName       = "terragrunt-out-dir"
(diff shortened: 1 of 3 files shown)


michaeldelago:
Updated to latest version of `master`

yhakbar:
Thanks for the PR @michaeldelago . I'm discussing internally with the team to see if this is the right approach for this.

My concern is that the behavior you're introducing here is a breaking change, so we need to make sure it makes sense to have the default be that recursive parsing of configurations isn't done.

If we determine that recursive parsing is something that all users should expect to be done by default, I'll ask that you flip the default behavior of the flag so that users have to opt-in to recursive parsing. If it's the case that we think this is an edge case that users aren't likely to need and that it's acceptable for this to be opt-in, we'll accept your PR, assuming everything else looks good.

yhakbar:
@michaeldelago 

Discussed this with the team, and there are problems with introducing this breaking change.

Say you have the following dependency chain:
A <-- B <-- C

Where A depends on B, which depends on C.

In this example, if B re-exports an input from dependency C as an output to A, A will either get an empty value or a mock value from the dependency on C.

This is non-obvious behavior that shouldn't be the default experience for users.

For this reason, please update the default to recursively parse dependencies.

This new functionality that you're introducing also does not come with any new tests. To make sure that this behavior is well documented, and the edge cases are tested, please introduce tests that test this behavior both with the flag set, and without it.

Thanks!

PR #3170, opened 2024-05-29
[WIP] Update 'trragrunt' to 'terragrunt' in TestTerragruntSkipConfirmExternalDependencies

jalessio:
## Description

It appears that the `TestTerragruntSkipConfirmExternalDependencies` test function in `test/integration_test.go` has a typo when calling the `terragrunt` command. I'm surprised to see the test suite passes with the command as `trragrunt` instead of `terragrunt`.

This PR updates the command from `trragrunt` to `terragrunt`.

## TODOs

- [ ] Update the docs.
- [ ] Does this exposes a gap in test coverage?

## Release Notes (draft)

### DIFF
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -6468,7 +6468,7 @@ func TestTerragruntSkipConfirmExternalDependencies(t *testing.T) {
 	oldStdout := os.Stderr
 	os.Stderr = w
 
-	err = runTerragruntCommand(t, fmt.Sprintf("trragrunt destroy --terragrunt-working-dir %s", testPath), &stdout, &stderr)
+	err = runTerragruntCommand(t, fmt.Sprintf("terragrunt destroy --terragrunt-working-dir %s", testPath), &stdout, &stderr)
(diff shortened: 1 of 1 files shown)


yhakbar (approved):

yhakbar:
Bizarre!

jalessio:
@yhakbar I was expecting the test suite to run against my branch so that I could see if this change caused any tests to fail. If it does _NOT_ I _suspect_ the `TestTerragruntSkipConfirmExternalDependencies` test is not actually doing anything.

denis256:
The executable name is ignored when tests are executed and app instances are created, so the test works in the same way as before

https://github.com/gruntwork-io/terragrunt/blob/master/test/integration_test.go#L4182

jalessio:
> The executable name is ignored when tests are executed and app instances are created, so the test works in the same way as before
> 
> https://github.com/gruntwork-io/terragrunt/blob/master/test/integration_test.go#L4182

Ah, thanks for the explanation @denis256! 🧠

PR #3180, opened 2024-06-03
Try out GitHub Actions

josh-padnick:


josh-padnick:
We concluded that we were able to use GitHub Actions, however we assume we only receive 2,000 minutes per month across all Gruntwork public repos, so use the time wisely!

PR #3193, opened 2024-06-11
feat: Adding initialization check caching

yhakbar:
## Description

Adds initialization check caching so that remote state checks only have to be done once for a given bucket if the bucket can be initialized successfully.

Mitigates the need to have any network traffic to inspect the initialization status of a bucket after it's been initialized once.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Updated `remote` package to include a cache for initialization checks.

### DIFF
--- a/remote/remote_state.go
+++ b/remote/remote_state.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/gruntwork-io/go-commons/errors"
 	"github.com/gruntwork-io/terragrunt/codegen"
+	"github.com/gruntwork-io/terragrunt/internal/cache"
 	"github.com/gruntwork-io/terragrunt/options"
 )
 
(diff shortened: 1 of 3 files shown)


remote/remote_state.go
```diff
@@ -30,6 +31,10 @@ type stateAccess struct {
 
 var stateAccessLock = newStateAccess()
 
+// initializedRemoteStateCache is a cache to store the result of a remote state initialization check.
+// This is used to avoid checking to see if remote state needs to be initialized multiple times.
+var initializedRemoteStateCache = cache.NewCache[bool]()
```
levkohimins, line 36: I strongly recommend against using global variables. Our integration tests run in parallel, which can cause conflicts.

levkohimins (changes requested):
Overall everything looks great, I just have some concerns about global variables.

remote/remote_state.go
yhakbar: I thought this might be OK, given that there's a mutex to guard access to shared state, and that we're already doing this with `stateAccessLock`. Could you recommend an alternate approach for how I should organize this?

yhakbar: Also, FWIW, it's how we handle all of our caching right now. Maybe it's something to refactor later?

levkohimins: I would put this variable into the `RemoteState` structure. We already have a similar issue with another global variable.

https://github.com/gruntwork-io/terragrunt/blob/17fd480320b856096ada087b784c12a86ddd5703/config/dependency.go#L156-L158

This forces us to exclude such tests from parallel running

https://github.com/gruntwork-io/terragrunt/blob/17fd480320b856096ada087b784c12a86ddd5703/test/integration_test.go#L5023-L5025

I mentioned about hit issue in Slack:
https://gruntwork-io.slack.com/archives/C06KU8JJU5V/p1713977826964509

levkohimins (approved):
LGTM!

remote/remote_state.go
yhakbar: Created https://github.com/gruntwork-io/terragrunt/issues/3198 to address this later.

PR #3263, opened 2024-07-11
Added passing of extra ENVs to the hook executed command

### LINKED ISSUE #3211: Identify after_hook trigger
### Summary

A way to identity the after_hook trigger if hook can be triggered by multiple commands

### Motivation

I am running an after_hook which executes on both commands plan and apply.

The hook starts a Go script which creates database users.

I wish to know within the script which of those has initiated the trigger such that I can print the outcome of what will happen when an apply will be performed (simply a dry run of the script).

### Proposal

A simplified version of the hook:

```
terraform {
  after_hook "create_users" {
    commands = ["plan", "apply"]
    execute = [
      "go", "run", "main.go",
    ]
  }
}
```

For my use case, this is a Go CLI with quite many arguments, thus I do not want to split it into two distinct after_hooks. I also have to update the CLI arguments on both hooks for it to work in that case.

My proposal is for example to add some type of environment variable such that I can do a os.Getenv(...) in the Go script.

### Technical Details

A non-breaking feature for hooks.

### Press Release

A new environment variable for hooks to distinguish between triggers.

### Drawbacks

No drawbacks.

### Alternatives

*No response*

### Migration Strategy

*No response*

### Unresolved Questions

Any alternative or existing solution?

### References

*No response*

### Proof of Concept Pull Request

*No response*

### Support Level

- [ ] I have Terragrunt Enterprise Support
- [ ] I am a paying Gruntwork customer

### Customer Name

*No response*

levkohimins:
## Description

Fixes #3211.

## TODOs

- [x] Update the docs.

## Release Notes (draft)


cli/commands/terraform/hook.go
```diff
@@ -169,3 +171,12 @@ func executeTFLint(ctx context.Context, terragruntOptions *options.TerragruntOpt
 	}
 	return nil
 }
+
+func cloneTerragruntOptionsForHook(opts *options.TerragruntOptions, hookName string) *options.TerragruntOptions {
+	newOpts := *opts
+	newOpts.Env = util.CloneStringMap(opts.Env)
+	newOpts.Env["TERRAFORM_COMMAND"] = opts.TerraformCommand
```
yhakbar, line 178: Let's make these environment variables more generic.

e.g.
`TG_CTX_COMMAND`
`TG_CTX_HOOK_NAME`

The motivation behind requesting that they're named like this is that we want them to definitely avoid collisions with user defined environment variables (users should know that they should never set `TG_CTX_` prefixed variables if they don't want the values clobbered by Terragrunt), and it avoids using terminology specific to Terraform, when that might be confusing, given that the user might be wrapping OpenTofu instead.

yhakbar (changes requested):

cli/commands/terraform/hook.go
levkohimins: Agree, renamed. Thanks for the suggestion!

yhakbar (approved):

yhakbar:
@levkohimins , it's not worth blocking this pull request, but let's also make sure we follow up with more documentation on how the environment variables are populated, and an example in the docs (you can basically just c+p the test fixture, and point out what it will emit in the docs).

yhakbar (approved):

levkohimins:
Resolved in [v0.61.1](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.61.1) release.

PR #3268, opened 2024-07-16
Hiding diagnostic error when first parsing `remote_state`

### LINKED ISSUE #3036: "No variable named dependency" when chaining dependency from parent (since v0.54.16)
**Describe the bug**

v0.54.16 has introduced a regression related to chained dependencies leading up to the parent module which in turn has a dependency.  v0.54.15 is fine, and the issue persists with v0.55.20.

Consider the following layout:

```
.
├── tfstate/
│   ├── outputs.tf
│   └── terragrunt.hcl
├── vpc/
│   ├── outputs.tf
│   ├── terragrunt.hcl
├── eks/
│   ├── main.tf
│   ├── terragrunt.hcl
└── terragrunt.hcl
```

With dependencies:

image

The `tfstate` dependency is declared in the root-level terragrunt.hcl for purposes of generating `backend.tf`.  The vpc module works fine, however the eks module generates this error:

```
ERRO[0000] Error: Unknown variable

ERRO[0000]   on /home/jtackaberry/projects/terragrunt-dep-bug/terragrunt.hcl line 12, in remote_state:
ERRO[0000]   12:     path = "${dependency.tfstate.outputs.prefix}-${path_relative_to_include()}.tfstate"
ERRO[0000]
ERRO[0000] There is no variable named "dependency".
```

An error like the above is generated for each `dependency.tfstate.outputs` reference in the `remote_state` block.

We use this pattern quite a lot in our projects, where we have a standalone `tfstate` module in the repo (with no dependencies or includes) that bootstraps an S3 bucket and DynamoDB table using a custom module that provides the details as outputs.  Then `terragrunt.hcl` in the root sources this terragrunt module as a dependency to configure the backend.

**To Reproduce**

I've created a sample project at https://github.com/jtackaberry/terragrunt-dep-bug 

First, ensure you have a terragrunt version greater than or equal to v0.54.16.  Then:

```bash
# Clone the example
git clone https://github.com/jtackaberry/terragrunt-dep-bug.git

# Apply mock tfstate module -- this is fine
cd terragrunt-dep-bug/tfstate
terragrunt apply

# Apply mock vpc module -- this is fine
cd ../vpc
terragrunt apply

# Apply mock eks module -- error occurs here
cd ../eks
terragrunt apply
```

Though not reproduced above, I also believe #2974 is related, and possibly may have the same root cause.  The versions line up, certainly.

**Expected behavior**

The error message itself actually _may_ be innocuous.  The `eks` module in the example project does seem to work (insofar as `backend.tf` is correctly populated).

However the expected behavior is that of pre-0.54.16 where no error is printed.

**Versions**
- Terragrunt version: v0.54.15 (works), v0.54.16 (broken), v0.55.20 (broken)
- Terraform version: v1.5.7
- Environment details: Ubuntu 20.04, Ubuntu 22.04

levkohimins:
## Description

After refactoring the code in version v0.54.16, and moving the diags logger writer to inside of the `hclparse` package, it began to display a diags parse error during a first parsing of the `remote_state` block, when it should not. In fact, the fix hides the diag errors.

https://github.com/gruntwork-io/terragrunt/blob/95f356ec87e295d1a300c3e2acda904e865828f7/config/dependency.go#L646

Fixes #3036.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

### DIFF
--- a/cli/commands/hclfmt/action.go
+++ b/cli/commands/hclfmt/action.go
@@ -131,7 +131,9 @@ func formatTgHCL(opts *options.TerragruntOptions, tgHclFile string) error {
 func checkErrors(logger *logrus.Entry, disableColor bool, contents []byte, tgHclFile string) error {
 	parser := hclparse.NewParser()
 	_, diags := parser.ParseHCL(contents, tgHclFile)
-	diagWriter := util.GetDiagnosticsWriter(logger, parser, disableColor)
+
+	writer := &util.LogWriter{Logger: logger, Level: logrus.ErrorLevel}
+	diagWriter := util.GetDiagnosticsWriter(writer, parser, disableColor)
 	err := diagWriter.WriteDiagnostics(diags)
 	if err != nil {
 		return errors.WithStackTrace(err)
--- a/config/config.go
+++ b/config/config.go
@@ -73,8 +73,10 @@ var (
 	}
 
 	DefaultParserOptions = func(opts *options.TerragruntOptions) []hclparse.Option {
+		writer := &util.LogWriter{Logger: opts.Logger, Level: logrus.ErrorLevel}
+
 		return []hclparse.Option{
-			hclparse.WithLogger(opts.Logger, opts.DisableLogColors),
+			hclparse.WithDiagnosticsWriter(writer, opts.DisableLogColors),
 			hclparse.WithFileUpdate(updateBareIncludeBlock),
 		}
 	}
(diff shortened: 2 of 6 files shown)


yhakbar (approved):
LGTM

levkohimins:
@yhakbar thanks for the review!

levkohimins:
Resolved in [v0.62.2](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.62.2) release.

PR #3286, opened 2024-07-23
3274 - fix: Include nothing when value for `terragrunt-include-dir` doesn't exist on the filesystem

### LINKED ISSUE #3274: CLI argument`--terragrunt-include-dir` includes all folders if the pattern does not match since version `0.50.1`
## Describe the bug

A run of `terragrunt run-all apply --terragrunt-include-dir "does-not-exist-*"` leads to apply of all folders in the current directory. Especially dangerous when running in the CI and you upgraded from 0.50.0 to any later version!

The weird behaviour of this was introduced with this PR: https://github.com/gruntwork-io/terragrunt/pull/2664

IMO this is a breaking change but was added as a patch increment.

## Steps To Reproduce

- Create a folder with multiple subfolders that contain `terragrunt.hcl` files.
- Head to the folder and execute the above mentioned command, make sure the include pattern you specify does not match.
- You should see that it selects all folders for the execution plan.

I confirmed in my local tests that it is not related to glob matching, so any non existing folder in the include should work to reproduce the problem.

## Expected behavior

- After executing the above command and no folder is matching the specified include pattern, I expect an empty execution and nothing to happen.
- The argument `--terragrunt-include-dir` should work like a whitelist (which it did before version 0.50.1)

## Nice to haves

A screenshot of a CLI session to reproduce:
Screenshot 2024-07-18 at 13 00 21

## Versions
- Terragrunt version: 0.50.0

yhakbar:
## Description

Fixes #3274.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Updated `--terragrunt-include-dir` to properly exclude by default when set, even if the value is not found.

### DIFF
--- a/options/options.go
+++ b/options/options.go
@@ -203,6 +203,10 @@ type TerragruntOptions struct {
 	// Unix-style glob of directories to include when running *-all commands
 	IncludeDirs []string
 
+	// If set to true, exclude all directories by default when running *-all commands
+	// Is set automatically if IncludeDirs is set
+	ExcludeByDefault bool
+
 	// If set to true, do not include dependencies when processing IncludeDirs (unless they are in the included dirs)
 	StrictInclude bool
 
--- a/options/options.go
+++ b/options/options.go
@@ -533,6 +533,7 @@ func (opts *TerragruntOptions) Clone(terragruntConfigPath string) *TerragruntOpt
 		ExcludesFile:                   opts.ExcludesFile,
 		ExcludeDirs:                    opts.ExcludeDirs,
 		IncludeDirs:                    opts.IncludeDirs,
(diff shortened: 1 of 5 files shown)


options/options.go
```diff
@@ -203,6 +203,10 @@ type TerragruntOptions struct {
 	// Unix-style glob of directories to include when running *-all commands
 	IncludeDirs []string
 
+	// If set to true, exclude all directories by default when running *-all commands
+	// Is set automatically if IncludeDirs is set
+	ExcludeByDefault bool
```
denis256, line 208: Usually, when a field is added to `TerragruntOptions`, it should be also updated `Clone()` function to copy the field

denis256 (approved):

PR #3299, opened 2024-07-25
Adding registry url discovery for Terragrunt Provider Cache

### LINKED ISSUE #3283: Provider caching does not follow "well-known" discovery specification
## Describe the bug

The terraform [service discovery process](https://developer.hashicorp.com/terraform/internals/remote-service-discovery#discovery-process) for modules and providers states that terraform native services should be discovered using the `/.well-known/terraform.json` document present at a specific host.

This will then give you the base paths for a specific service, for example, a provider registry.

The specific example provided on the website is:

```
{
  "modules.v1": "https://modules.example.com/v1/"
}
```

Which is the path that the hashicorp registry and opentofu use for their modules and providers.

However, one can choose to host these services under a different base path, such as `/foo/bar/v1/...`

The current iteration of the provider cache makes the [assumption this is not possible](https://github.com/gruntwork-io/terragrunt/blob/e5e020932745a734e1929f4854df5b5977da6d26/terraform/cache/handlers/provider_direct.go#L59).

```go
		Path:   path.Join("/v1/providers", provider.Namespace, provider.Name, "versions"),
```

## Steps To Reproduce

Run a tf registry hosting a provider registry in path that does not start with `/v1`. Attempt to enable provider cache. Terragrunt run will fail, as cache server will look for modules at `/v1` directly and get a `404` status.

Possibly adding a test case [here](https://github.com/gruntwork-io/terragrunt/blob/e5e020932745a734e1929f4854df5b5977da6d26/cli/provider_cache_test.go#L38) will be enough to surface the issue.

## Expected behavior

The cache server should respect the base path and be able to download the provider.

## Versions

- Terragrunt version: [v0.62.0](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.62.0)
- OpenTofu/Terraform version: 1.9.2
- Environment details (Ubuntu 20.04, Windows 10, etc.): `alpine/terragrunt` container, [this tag](https://hub.docker.com/layers/alpine/terragrunt/1.9.2/images/sha256-7abd33686bd184e247349d51bcc41c555aadbbcc2f3542acc038b00c6628e8a9?context=explore); plus reproduced with latest master ([e5e0209](https://github.com/gruntwork-io/terragrunt/commit/e5e020932745a734e1929f4854df5b5977da6d26))

## Additional context

I had this surface in a client's internal env, apologies due to the lack of screenshots/etc. In this case the issue was worked around by changing the registry to serve directly from `hostname/v1` (that was possible to do); but this should still be looked into because the registry was working fine with terraform but not the cache.

My suggestion is to grab the base path from the `.well-known/terraform.json` and storing it in the `Provider` model, which then can be referenced to build the URL.

I'm happy to submit a PR, but I'm still kind of unfamiliar with the codebase so it may take a while to get merged.

levkohimins:
## Description

Implemented [Discovery Process](https://developer.hashicorp.com/terraform/internals/remote-service-discovery#discovery-process) for Terragrunt Provider Cache. The PR does not need an integration test because we are replacing the static URLs with a dynamically discovered ones

Fixes #3283.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)


cli/provider_cache.go
```diff
@@ -236,10 +244,21 @@ func (cache *ProviderCache) createLocalCLIConfig(opts *options.TerragruntOptions
 	for _, registryName := range opts.ProviderCacheRegistryNames {
 		providerInstallationIncludes = append(providerInstallationIncludes, fmt.Sprintf("%s/*/*", registryName))
 
+		urls, err := DiscoveryURL(ctx, registryName)
+		if err != nil {
+			if !liberrors.As(err, &NotFoundWellKnownURL{}) {
+				return err
+			}
+			urls = DefaultRegistryURLs
+			opts.Logger.Debugf("Unable to discovery %q registry URLs, reason: %q, use default URLs: %s", registryName, err, urls)
```
yhakbar, line 253:
```suggestion
			opts.Logger.Debugf("Unable to discover %q registry URLs, reason: %q, use default URLs: %s", registryName, err, urls)
```

yhakbar (approved):

cli/provider_cache.go
levkohimins: Thanks Yousif!

denis256 (approved):

yhakbar (approved):

levkohimins:
Resolved in [v0.64.1](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.64.1) release.

Issue #3103, opened 2024-04-26
Terragrunt IAC Engine Plugin System


yhakbar:
### Summary

Introduce the ability to integrate with plugins to drive custom behavior in the underlying IAC tool orchestrated by Terragrunt (like OpenTofu or Terraform).

### Motivation

Users have been lacking two significant capabilities that are addressed by this RFC:

1. The ability to customize the usage of `tofu` and `terraform` when called by Terragrunt.

    Users have been relying on ensuring that a particular versions of tools have been set prior to executing `terragrunt` or utilizing a shim to alter the execution of the underlying IAC tool.

2. The ability to alter the context of the IAC execution separate from the Terragrunt execution.

    So far, there has been no way to isolate the IAM access that the underlying IAC tool has from the access that Terragrunt has. The IAC tool has also had to run in the same compute environment and on the same filesystem as Terragrunt.

    Terragrunt super users would like to be able to isolate the compute resources allocated to Terragrunt from the compute allocated to the underlying IAC tool so that they can fan out IAC updates across multiple instances/containers/pods.

### Proposal

Allow users to optionally specify an IAC `engine`, which will control how the underlying IAC operations like plans, applies, etc will be carried out instead of directly calling the `tofu` or `terraform` binaries.

Users will be able to use a configuration block that looks like the following to configure their engine in the relevant `terragrunt.hcl`:

```hcl
engine {
   source  = "github.com/acme/terragrunt-plugin-custom-opentofu"
   version = "v0.0.1" # Optionally specify version
   type    = "rpc" # Optionally specify the type of plugin

  # Optional JSON serializable metadata that is engine-specific
   metadata = {
     tofu_version = "v1.7.2"
   }
}
```

The `source` field would be either the path to a local binary (signified by starting the value with `.` or `/`) or a URL pointing to a GitHub repository with a releases page containing an asset that can be used (the appropriate architecture and platform would be guessed based on detected values of the host machine, and can be explicitly set via environment variables).

The optional `version` field would indicate the git tag associated with the release to download, when the `source` is not a local binary. Throws an error if set for a local path, and is the latest release by default for remote sources.

[...]

PR #3300, opened 2024-07-26
Engine verification

denis256:
## Description

Implemented check flow:
* Download SHA256SUMS 
* Download SHA256SUMS.SIG 
* Validate SHA256SUMS with GPG signature from SHA256SUMS.SIG and public key
* Validate Checksum of engine file from SHA256SUMS with real file

Checks are performed for the downloaded zip file with the engine and before loading engine file itself

Additional included changes:
* added public GPG signature used for official OpenTofu engine
* added skipping of engine verification if env variable `TG_ENGINE_SKIP_CHECK=1` is set
* added end-to-end tests to track engine checking
 
https://github.com/gruntwork-io/terragrunt/issues/3103

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added support for engine verification before running


engine/engine.go
```diff
@@ -152,28 +154,48 @@ func DownloadEngine(ctx context.Context, opts *options.TerragruntOptions) error
...
+		opts.Logger.Infof("Verifying checksum for %s", downloadFile)
+		if err := verifyFile(downloadFile, checksumFile, checksumSigFile); err != nil {
+			return errors.WithStackTrace(err)
+		}
+	} else {
+		opts.Logger.Warnf("Skipping verification for %s", downloadFile)
 	}
-	opts.Logger.Infof("Engine downloaded to %s", downloadFile)
+
 	if err := extractArchive(opts, downloadFile, localEngineFile); err != nil {
```
levkohimins, line 199: I know that this was added before, but consider to use:

```go
// Borrow the "unpack a zip cache into a target directory" logic from go-getter
var unzip = getter.ZipDecompressor{}

func(.... 
	if err := unzip.Decompress(dst, src, isDir, umask ); err != nil {
	}
```

engine/engine.go
```diff
@@ -578,3 +623,12 @@ func convertMetaToProtobuf(meta map[string]interface{}) (map[string]*anypb.Any,
...
 	return protoMeta, nil
 }
+
+// skipChecksumCheck returns true if the engine checksum check is skipped.
+func skipEngineCheck() bool {
+	switch strings.ToLower(os.Getenv(EngineSkipCheckEnv)) {
+	case "1", "yes", "true", "on":
+		return true
+	}
+	return false
```
levkohimins, line 633:
```suggestion
        ok, _ := strconv.ParseBool(os.Getenv(EngineSkipCheckEnv)) //nolint:errcheck
        return ok
```

engine/verification.go
```diff
@@ -0,0 +1,97 @@
...
+FuCDZbQHXPn3jzSEDQPSA62+ZyRGxXfpqgVPpT8IPzAdCRAhMuUZb62h+WX2ey91
+rnRFIXOlPTbrOMPLaMnGBjDnsWuCQmxBCXeevBh3u8q4Wa1xCZiqN8T7PSUusalu
+xita/w5ZA+Tzwxe8VqrgutCJdj5m5OxXX4v10xbbyPfhhnaahMduGL60MV/noxy6
+9TFlXIhgDj8dxV8wt8Tv/GZSSUALaBuPs9U7Q+fGiPFpC48Q75y0uS0QWXm0taxs
+Rm6AMowi8dJEq1BIKvdEO2lDJ1uSw5Xcamj6Nu0JrM8tc1uCFNYOAEHw2bDHIcHK
++uFqASiTa9QRj1SpSRkbPJB3yuPgUr1DgEoolSlMOnUiI46K3I9APD54nuShVbVq
+iE6bHk4c9kBU
+=TmYc
+-----END PGP PUBLIC KEY BLOCK-----`
+)
```
levkohimins, line 58: I would create another file named `public_keys.go` to keep this constant there

example: https://github.com/opentofu/opentofu/blob/main/internal/getproviders/public_keys.go

engine/engine.go
```diff
@@ -152,28 +154,48 @@ func DownloadEngine(ctx context.Context, opts *options.TerragruntOptions) error
 		return nil
 	}
 	downloadFile := filepath.Join(path, enginePackageName(e))
-	var downloadURL string
+
+	downloads := make(map[string]string)
+	checksumFile := ""
+	checksumSigFile := ""
 	if strings.HasPrefix(e.Source, "http") {
```
levkohimins, line 161:
```suggestion
	if strings.HasPrefix(e.Source, "://") {
```

Since `go-getter` supports not only `http`, I think this expands possible sources.  
_Not sure of the correct suggestion, requires rechecking_

levkohimins (changes requested):

engine/engine.go
denis256: Hm, I think `archiver` package has more applications here:
* it can detect if the downloaded file is an archive or not - people may use engine files as is
* I think it can also extract more archive formats in case if people will use different archiving tools

levkohimins: Roger

levkohimins (dismissed):
LGTM!

levkohimins (approved):
LGTM!

PR #3309, opened 2024-07-31
Sorting output and fixing minor bugs in `hclvalidate` command

levkohimins:
## Description

1. Renamed the `--terragrunt-hclvalidate-invalid` flag to `--terragrunt-hclvalidate-show-config-path`
2. Sorted output by file name.
3. Fixed `invalid character ',' looking for beginning of value` error when mock output is not specified.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

### DIFF
--- a/cli/commands/hclvalidate/command.go
+++ b/cli/commands/hclvalidate/command.go
@@ -12,7 +12,7 @@ const (
 	CommandName = "hclvalidate"
 
 	ShowConfigPathFlagName   = "terragrunt-hclvalidate-show-config-path"
-	ShowConfigPathEnvVarName = "TERRAGRUNT_HCLVALIDATE_INVALID"
+	ShowConfigPathEnvVarName = "TERRAGRUNT_HCLVALIDATE_SHOW_CONFIG_PATH"
 
 	JSONOutputFlagName   = "terragrunt-hclvalidate-json"
 	JSONOutputEnvVarName = "TERRAGRUNT_HCLVALIDATE_JSON"
--- a/cli/commands/hclvalidate/command.go
+++ b/cli/commands/hclvalidate/command.go
@@ -11,8 +11,8 @@ import (
 const (
 	CommandName = "hclvalidate"
 
-	InvalidFlagName   = "terragrunt-hclvalidate-invalid"
-	InvalidEnvVarName = "TERRAGRUNT_HCLVALIDATE_INVALID"
+	ShowConfigPathFlagName   = "terragrunt-hclvalidate-show-config-path"
+	ShowConfigPathEnvVarName = "TERRAGRUNT_HCLVALIDATE_INVALID"
 
 	JSONOutputFlagName   = "terragrunt-hclvalidate-json"
 	JSONOutputEnvVarName = "TERRAGRUNT_HCLVALIDATE_JSON"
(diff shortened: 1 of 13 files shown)


cli/commands/hclvalidate/command.go
```diff
@@ -11,8 +11,8 @@ import (
 const (
 	CommandName = "hclvalidate"
 
-	InvalidFlagName   = "terragrunt-hclvalidate-invalid"
-	InvalidEnvVarName = "TERRAGRUNT_HCLVALIDATE_INVALID"
+	ShowConfigPathFlagName   = "terragrunt-hclvalidate-show-config-path"
```
yhakbar, line 14: Maybe `terragrunt-hclvalidate-show-invalid-configs`?

cli/commands/hclvalidate/command.go
```diff
@@ -11,8 +11,8 @@ import (
...
+	ShowConfigPathEnvVarName = "TERRAGRUNT_HCLVALIDATE_INVALID"
```
yhakbar, line 15: If you're already making a breaking change, I think you should rename this to `TERRAGRUNT_HCLVALIDATE_SHOW_CONFIG_PATH` or `TERRAGRUNT_HCLVALIDATE_SHOW_INVALID_CONFIGS`

cli/commands/hclvalidate/command.go
levkohimins: I think the word "invalid" is a little misleading since `hclvalidate` outputs only invalid _error explanations/config paths_. so users might think that if `--terragrunt-hclvalidate-show-invalid-configs` is not specified, this command will also output the _valid_ config path.

cli/commands/hclvalidate/command.go
levkohimins: Indeed, I missed it. Thank you.

cli/commands/hclvalidate/command.go
levkohimins: Thus `show-config-path` tells us that instead of explaining errors of invalid configs, just show their paths.

yhakbar (approved):

levkohimins:
@yhakbar thanks for the review!

cli/commands/hclvalidate/command.go
yhakbar: This might be something we want to revisit in the later. The point of this flag is to make it so that we can conveniently work on invalid configurations, so it's a little weird that we're not referencing that in the flag.

I think what we might want to explore is having something like [ripgrep](https://github.com/BurntSushi/ripgrep) where users can turn on and off stdout for filenames, line numbers or diagnostics, etc.

Maybe the way users should approach this is to turn off diagnostic messages instead of turning on filenames, or something.

levkohimins: I also does not like this `--terragrunt-hclvalidate-show-config-path` flag.  The idea of ​​filtering diagnostic information is interesting.

Let's assume we have a detailed error diagnostics:

```hcl
│ Error: Missing required argument
│
│   on /workload/terragrunt.hcl line 25, in dependency "iam":
│   25: dependency iam {
│
│ The argument "config_path" is required, but no definition was found.
╵
╷
│ Error: Unsupported argument
│
│   on /workload/terragrunt.hcl line 26, in dependency "iam":
│   26:   config_pathx = "../iam"
│
│ An argument named "config_pathx" is not expected here. Did you mean "config_path"?
╵
╷
│ Error: Unsupported attribute
│
│   on /workload/terragrunt.hcl line 44:
│   44:   applications_credentials = dependency.iam.outputs.applications_credentials
│
│ This object does not have an attribute named "iam".
╵
╷
│ Error: Unsupported attribute
│
│   on /workload/terragrunt.hcl line 50:
│   50:       kubeconfig_values = dependency[paas.dependency].outputs.kubeconfig_values
│     ├────────────────
│     │ paas.dependency is "tst"
│
│ This object does not have an attribute named "outputs".
```

Perhaps we could add `--terragrunt-hclvalidate-show` flags to list comma separated what information should be shown, eg `--terragrunt-hclvalidate-show config-path,severity,summary,detail,range,snippet`

levkohimins:
Resolved in [v0.64.5](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.64.5) release.

PR #3395, opened 2024-09-09
Added handling of TERRAGRUNT_TEMP_QUOTE_NULL

denis256:
## Description

Included changes:
* added handling of TERRAGRUNT_TEMP_QUOTE_NULL env var
* added test for checking passing null values
* added basic integration test to track this change

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added `TERRAGRUNT_TEMP_QUOTE_NULL`

This undcoumented feature flag allows users to temporarily revert the breaking change in release [v0.50.0](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.50.0) to avoid sending the string "null" as the value of an input to OpenTofu/Terraform instead of not setting an input at all.

The flag is being introduced into the codebase for a temporary duration to support Enterprise customers that have asked for this as a stop gap while they update their codebase to no longer need it.

Terragrunt will not support this flag in the future. Do not rely on it.

We will be coordinating closely with Enterprise customers to determine when they no longer need it supported.

### DIFF
--- a/cli/commands/terraform/action.go
+++ b/cli/commands/terraform/action.go
@@ -787,7 +787,7 @@ func filterTerraformEnvVarsFromExtraArgs(terragruntOptions *options.TerragruntOp
 // natively.
 func ToTerraformEnvVars(opts *options.TerragruntOptions, vars map[string]interface{}) (map[string]string, error) {
 	if useLegacyNullValues() {
-		opts.Logger.Warnf("%s is a temporary workaround to bypass the breaking change in #2663.\nThis flag will be removed in the future.\nDo not rely on it.", useLegacyNullValuesEnvVar)
+		opts.Logger.Warnf("⚠️ %s is a temporary workaround to bypass the breaking change in #2663.\nThis flag will be removed in the future.\nDo not rely on it.", useLegacyNullValuesEnvVar)
 	}
 
 	out := map[string]string{}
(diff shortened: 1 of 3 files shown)


cli/commands/terraform/action.go
```diff
@@ -781,13 +785,22 @@ func filterTerraformEnvVarsFromExtraArgs(terragruntOptions *options.TerragruntOp
 // Convert the given variables to a map of environment variables that will expose those variables to Terraform. The
 // keys will be of the format TF_VAR_xxx and the values will be converted to JSON, which Terraform knows how to read
 // natively.
-func ToTerraformEnvVars(vars map[string]interface{}) (map[string]string, error) {
+func ToTerraformEnvVars(opts *options.TerragruntOptions, vars map[string]interface{}) (map[string]string, error) {
+	if useLegacyNullValues() {
+		opts.Logger.Warnf("%s is a temporary workaround to bypass the breaking change in #2663.", useLegacyNullValuesEnvVar)
```
yhakbar, line 790: NIT: It might be better to group this into one logging message for users with JSON logging.

cli/commands/terraform/action.go
```diff
@@ -832,3 +845,7 @@ func setTerragruntNullValues(terragruntOptions *options.TerragruntOptions, terra
 
 	return varFile, nil
 }
+
+func useLegacyNullValues() bool {
+	return os.Getenv(useLegacyNullValuesEnvVar) != ""
```
yhakbar, line 850: Let's check that it's set to `"1"` instead of anything other than `""`.

That way people don't have to unset the env var if they want it dynamically set.

yhakbar (approved):

yhakbar (approved):

PR #3443, opened 2024-09-26
Engine log level propagation

denis256:
## Description

Included changes:
* added propagation of log level to the engine through env variables
* added env var documentation
* added tests to check log passing
* updated handling of env vars through flags
* simplified tests

## TODOs

- [x] Update the docs.

## Release Notes (draft)

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -522,8 +522,8 @@ func createEngine(terragruntOptions *options.TerragruntOptions) (*proto.EngineCl
 	engineLogLevel := os.Getenv(EngineLogLevelEnv)
 	if engineLogLevel == "" {
 		engineLogLevel = terragruntOptions.LogLevel.String()
-		// turn off log formatting if disabled for terragrunt
-		if terragruntOptions.DisableLogFormatting {
+		// turn off log formatting if disabled for Terragrunt
+		if terragruntOptions.DisableLog {
 			engineLogLevel = hclog.Off.String()
 		}
 	}
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -127,7 +127,6 @@ func Run(
 
 // WithEngineValues add to context default values for engine.
 func WithEngineValues(ctx context.Context) context.Context {
-
 	ctx = context.WithValue(ctx, TerraformCommandContextKey, &sync.Map{})
 	ctx = context.WithValue(ctx, LocksContextKey, util.NewKeyLocks())
 	ctx = context.WithValue(ctx, LatestVersionsContextKey, cache.NewCache[string]("engineVersions"))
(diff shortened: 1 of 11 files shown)


yhakbar (dismissed):

engine/engine.go
```diff
@@ -518,10 +519,17 @@ func createEngine(terragruntOptions *options.TerragruntOptions) (*proto.EngineCl
 
 	terragruntOptions.Logger.Debugf("Creating engine %s", localEnginePath)
 
+	engineLogLevel := os.Getenv(EngineLogLevelEnv)
```
yhakbar, line 522: Consider whether you need special behavior to handle this:
https://github.com/gruntwork-io/terragrunt/blob/58e2c20ed2c50b4bdc3c278a01bc447fab650d78/cli/commands/flags.go#L352-L361

We might want to suppress logging in the engine too when it's disabled in Terragrunt.

yhakbar (changes requested):

engine/engine.go
denis256: Added setting of "off" log level

engine/engine.go
```diff
@@ -518,10 +519,21 @@ func createEngine(terragruntOptions *options.TerragruntOptions) (*proto.EngineCl
...
+	if engineLogLevel == "" {
+		engineLogLevel = terragruntOptions.LogLevel.String()
+		// turn off log formatting if disabled for terragrunt
+		if terragruntOptions.DisableLogFormatting {
+			engineLogLevel = hclog.Off.String()
```
yhakbar, line 527: This is different. Disabling log formatting only makes the logs output in key-value pairs, it doesn't stop them from logging altogether.

engine/engine.go
```diff
@@ -518,10 +519,21 @@ func createEngine(terragruntOptions *options.TerragruntOptions) (*proto.EngineCl
 
 	terragruntOptions.Logger.Debugf("Creating engine %s", localEnginePath)
 
+	engineLogLevel := os.Getenv(EngineLogLevelEnv)
```
levkohimins, line 522: It will not work with, `false`, `0`, and will not notify about a wrong value. The same with `EngineCachePathEnv`, `TG_ENGINE_SKIP_CHECK`.

In this particular case I would recommend adding `opts.DisableLog`

```go
		&cli.BoolFlag{
			Name:   TerragruntLogDisableFlagName,
			EnvVar: TerragruntLogDisableEnvName,
			Destination: &opts.DisableLog,
```

Let's agree that the only entry point for **_all_** environment variables is CLI flags, this will help avoid issues in the future, when we forget about the existence of this "orphan", and it will live its own life. For example, when globally renaming a prefix or redesign the CLI interface. If for some reason you don't need to show the evn variable flag in the CLI help, you can simply hide it from users with `Hidden: true`.

levkohimins (changes requested):

levkohimins (approved):
Great job!

yhakbar (approved):

PR #3473, opened 2024-10-12
[WIP] feature: Move force_path_style to use_path_style

### LINKED ISSUE #3472: force_path_style in S3 backend is now deprecated
## Describe the bug

force_path_style is now deprecated and should be replaced with use_path_style

## Steps To Reproduce

Steps to reproduce the behavior, code snippets and examples which can be used to reproduce the issue.

Be sure that the maintainers can actually reproduce the issue. Bug reports that are too vague or hard to reproduce are hard to troubleshoot and fix.

```hcl
force_path_style = true
```

Will produce the error

```
tofu: │ Warning: Deprecated Parameter
tofu: │
tofu: │   on provider.tf line 6, in terraform:
tofu: │    6:   backend "s3" {}
tofu: │
tofu: │ Parameter "force_path_style" is deprecated. Use "use_path_style" instead.

This should be changed to

```hcl
use_path_style = true
```

## Versions

- Terragrunt version: latest
- OpenTofu/Terraform version: latest

## Additional context

https://developer.hashicorp.com/terraform/language/backend/s3#force_path_style

I have a branch with the code changes to change the existing configuration option to the new one. It is a straight replacement. If the community thinks it is better I can add both but I think a straight replacement with a breaking change warning would be better as it will stop any new deployments of the deprecated configuration option.

lazzurs:
## Description

As per the documentation linked below the force_path_style option is now deprecated and should be changed to use_path_style.

https://developer.hashicorp.com/terraform/language/backend/s3#force_path_style

It may be better to add both to allow people to migrate from one to the other but given this is only in one place in the docs I expect there isn't a lot of use and highlighting as a breaking change will make future migration easier and stop being using the now deprecated option.

Fixes #3472 .

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Updated force_path_style

### Migration Guide

Change

```hcl
force_path_style = true
```

to 

```hcl
use_path_style = true
```

### DIFF
--- a/docs/_docs/02_features/keep-your-remote-state-configuration-dry.md
+++ b/docs/_docs/02_features/keep-your-remote-state-configuration-dry.md
@@ -245,7 +245,7 @@ remote_state {
 
     shared_credentials_file     = "/path/to/credentials/file"
     skip_metadata_api_check     = true
-    force_path_style            = true
+    use_path_style            = true
   }
 }
 ```
--- a/remote/remote_state_s3.go
+++ b/remote/remote_state_s3.go
@@ -115,7 +115,7 @@ type RemoteStateConfigS3 struct {
 	LockTable        string                        `mapstructure:"lock_table"`   // Deprecated in Terraform version 0.13 or newer.
 	DynamoDBTable    string                        `mapstructure:"dynamodb_table"`
 	CredsFilename    string                        `mapstructure:"shared_credentials_file"`
-	S3ForcePathStyle bool                          `mapstructure:"force_path_style"`
+	S3UsePathStyle   bool                          `mapstructure:"use_path_style"`
 	AssumeRole       RemoteStateConfigS3AssumeRole `mapstructure:"assume_role"`
 }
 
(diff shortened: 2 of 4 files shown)


yhakbar:
Hey @lazzurs ,

Thanks for cutting this PR!

We do not want to have Terragrunt require users to use the latest version of OpenTofu/Terraform when they use it. As a consequence, we'll want to support the legacy configuration, in addition to the modern version. 

One approach we can take to make this an opt-in breaking change is to include it in [Strict Mode](https://terragrunt.gruntwork.io/docs/reference/strict-mode/).

Please adjust this PR to support both approaches for the sake of backwards compatibility.

Thanks again!

lazzurs:
Thanks for the comment.

As the PR is currently structured it would support both the legacy and current configuration without any issues other than if someone decides to use the new option with an older version of Terraform.

Happy to support strict mode, that looks good. I will get this done in the next few days. Thanks for the feedback. It's important to ensure the users have the best experience and I appreciate your focus on that.

lazzurs:
I will get that addressed shortly.

yhakbar:
Hey @lazzurs ,

Do you think you can revisit this PR? It has gone stale.

PR #3550, opened 2024-11-11
feat: Adding units-reading

### LINKED ISSUE #3418: Add `--terragrunt-queue-include-units-reading` flag to Terragrunt
### Summary

Add a new flag that works similarly to the [terragrunt-modules-that-include](https://terragrunt.gruntwork.io/docs/reference/cli-options/#terragrunt-modules-that-include) flag, but instead adds Units (terragrunt.hcl files) to the queue of units that will be added to the Directed Acyclic Graph (DAG), if they read a file using a Terragrunt function.

NOTE: This RFC will use the relatively new language of "Unit" to refer to the directory where a `terragrunt.hcl` file leaf node is. See [Stacks](https://github.com/gruntwork-io/terragrunt/issues/3313) for more info. This language will become standard eventually.

### Motivation

Users have asked for similar functionality to that which is available with [terragrunt-modules-that-include](https://terragrunt.gruntwork.io/docs/reference/cli-options/#terragrunt-modules-that-include), but specifically to handle things like updated JSON files and updated SOPS encrypted files to determine which Units need to be added to the DAG.

Right now, the best solution that users have available to them is to manually stitch together the list of Units that read external files by manually parsing `terragrunt.hcl` files. This can be error prone, however, as users have to work around issues like relative paths, and interpolated HCL functions changing which file is read.

### Proposal

Introduce a hook that is used in every HCL function Terragrunt supports which reads a file to update a central mapping of files being read to the Units that read them.

Also introduce the flag `--terragrunt-queue-include-units-reading` that can be used in `run-all` commands to include Units that match up with the central mapping derived from the hook in HCL functions reading files.

### Technical Details

Functions that definitely need the hook:
- `read_terragrunt_config`
- `sops_decrypt_file`
- `read_tfvars_file`

Over time, more functions can support this behavior.

Whenever any of these HCL functions are called, they will update the shared mapping of files to Units. As part of the `run-all` command, all Units in a given working directory are read and parsed, those parses should result in the mapping being updated, which will change how the queue is updated accordingly.

### Press Release

Terragrunt can now watch for changes in external files!

Using the new `--terragrunt-queue-include-units-reading` flag, users can now drive updates to any Unit that reads from a file using an HCL function. This allows for CI/CD workflows to intelligently handle updates to files in formats other than HCL, like YML or JSON, including encrypted files using SOPS.

### Drawbacks

This might result in a performance penalty for functions that read files. 

We have to be careful introducing functionality like this, as functions often be repeatedly called during parsing in a way that degrades performance. Smart memoization, etc can reduce the impact of this.

### Alternatives

1. Manually search through `terragrunt.hcl` files for references to files that read files in HCL functions.
2. Always wrap file reads with includes, so that [terragrunt-modules-that-include](https://terragrunt.gruntwork.io/docs/reference/cli-options/#terragrunt-modules-that-include) can be used instead.

### Migration Strategy

_No response_

### Unresolved Questions

_No response_

### References

_No response_

### Proof of Concept Pull Request

_No response_

### Support Level

- [ ] I have Terragrunt Enterprise Support
- [ ] I am a paying Gruntwork customer

### Customer Name

_No response_

yhakbar:
## Description

Fixes #3418.

Adds hooks to HCL functions that read files so that changes relevant to non-Terragrunt configuration files can be used to drive changes to the Terragrunt queue.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added `--terragrunt-queue-include-units-reading` flag. When set, the flag will add any Unit that reads the file to the Terragrunt queue.


options/options.go
```diff
@@ -245,6 +245,10 @@ type TerragruntOptions struct {
 	// in this list.
 	ModulesThatInclude []string
 
+	// When used with `run-all`, restrict the units in the stack to only those that read at least one of the files
+	// in this list.
+	UnitsReading []string
```
denis256, line 250: I think clone() function should be updated to copy conents

options/options.go
```diff
@@ -717,6 +725,37 @@ func (opts *TerragruntOptions) DataDir() string {
...
+			return
+		}
+	}
+
+	opts.Logger.Debugf("Tracking that file %s was read by %s.", file, unit)
+	opts.readFiles[file] = append(opts.readFiles[file], unit)
+}
+
+// DidReadFile checks if a given file was read by a given unit.
+func (opts *TerragruntOptions) DidReadFile(file string, unit string) bool {
```
denis256, line 745: file, unit string

options/options.go
denis256: Can be used a map to store files, and avoid iterations to find

yhakbar: Can you elaborate on how you'd want that implemented?

yhakbar:
@denis256 

This should be working now, but still happy to get input on implementation details, of course.

Still need to document. Will look to get to that soon.

denis256 (approved):

Issue #3134, opened 2024-05-13
Feature Flags, Errors and Exclude


yhakbar:
### Summary

Provide first class support for feature flags as part of Terragrunt HCL configuration.

Allow for dynamic configuration of behavior in select `terragrunt.hcl` files based on the presence or absence of feature flags that are set via environment variables and CLI flags.

In addition, update how Terragrunt errors and exckudes are handled to ensure that unstable configuration can be handled gracefully.

### Motivation

Terragrunt is frequently used in monorepo contexts, and it lends itself to this in how it segments IAC state into separate directories. One definition of monorepos is a single codebase with multiple independent, but related, projects. By this definition, Terragrunt is very much an IAC monorepo tool. Multiple units of IAC are defined independently, and as a whole, they represent a repository of IAC.

Feature flags are a common way to manage the complexity of a monorepo. They allow for the gradual rollout of new features, the ability to turn off features that are not ready for production, and the ability to manage the complexity of a large codebase.

This is especially important in the context of Terragrunt, where infrastructure is most safely updated when updated in small, incremental changes. In addition, the ability to control how failure is handled in IAC is extremely important. Preventing full resolution of an apply across multiple Terragrunt units because a known flaky unit is failing is not always something that can be remediated by the use of retries, and it can be expensive to do so. Occasionally, it is better to ignore the failure of a known flaky unit and continue with the rest of applies, assuming that the failure is not critical to the overall success of the apply.

An example of such a failure would be a dependency chain where one service is deployed by Terragrunt, and has a `url` output where the service can be accessed, and another service which uses a `dependency` block to pass that `url` into the environment variables of a second service.

In this example, if the first service fails to deploy, the second service will also fail to deploy. However, if the first service is known to be flaky, and the second service is not dependent on the first service *being deployed successfully*, it is better to ignore the failure of the first service and continue with the deployment of the second service, leveraging the `url` output from a previous successful apply.

Reasons that a unit might be marked in this way include:

- Known releases of underlying module updates that are not yet stable.
- Known problematic behavior of the underlying resource that is not yet resolved.
- Known dependence on network connectivity that is not yet reliable.

### Proposal

Provide a combination of:

1. A method to signal Terragrunt that a flag is set.
2. A method to define alternate behavior in Terragrunt when a flag is set.

### Proposed Syntax

```bash
TG_FLAG_feature_name="value" terragrunt command
# or
terragrunt --feature "feature_name=value" command
```

```hcl
[...]

PR #3584, opened 2024-11-20
WIP: Errors block

denis256:
## Description

RFC: https://github.com/gruntwork-io/terragrunt/issues/3134

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

### DIFF
--- a/config/errors.go
+++ b/config/errors.go
@@ -5,6 +5,15 @@ import (
 	"strings"
 )
 
+type ErrorConfig struct {
+}
+
+type IgnoreConfig struct {
+}
+
+type RetryConfig struct {
+}
+
 // Custom error types
 
 type InvalidArgError string


options/options.go
```diff
@@ -795,5 +803,240 @@ type EngineOptions struct {
 	Meta    map[string]interface{}
 }
 
+// ErrorsConfig extracted errors handling configuration.
+type ErrorsConfig struct {
+	Retry  map[string]*RetryConfig
+	Ignore map[string]*IgnoreConfig
+}
+
+// RetryConfig represents the configuration for retrying specific errors
```
levkohimins, line 812:
```suggestion
// RetryConfig represents the configuration for retrying specific errors.
```
strict-list complains about a comment without a dot at the end.

options/options.go
```diff
@@ -795,5 +803,240 @@ type EngineOptions struct {
...
+			opts.Logger.Warnf(
+				"Encountered retryable error: %s\nAttempt %d of %d. Waiting %d second(s) before retrying...",
+				action.RetryMessage,
+				currentAttempt,
+				action.RetryAttempts,
+				action.RetrySleepSecs,
+			)
+
+			// Sleep before retry
+			time.Sleep(time.Duration(action.RetrySleepSecs) * time.Second)
```
levkohimins, line 922:
```suggestion
			select {
			case <-time.After(time.Duration(action.RetrySleepSecs) * time.Second):
				// try again
			case <-ctx.Done():
				return errors.New(ctx.Err())
			}

```

Let's imagine that if an user sets RetrySleepSecs to say 60 seconds and RetryAttempts to 4 times and then wants to terminate the application by pressing ctrl-c, nothing will happen for 240 seconds. To prevent this, we need to listen to the context in case it was cancelled.

levkohimins (changes requested):

levkohimins (approved):

docs/_docs/04_reference/config-blocks-and-attributes.md
```diff
@@ -1291,6 +1292,99 @@ This setup is useful for scenarios where output evaluation is still needed, even
...
+
+Parameters:
+
+- `ignorable_errors`: A list of regex patterns to define errors to ignore.
+  - `"Error: safe-to-ignore.*"`: Ignores errors containing `Error: safe-to-ignore`.
+  - `"!Error: critical.*"`: Ensures errors containing `Error: critical` are not ignored.
+- `message` (Optional): A warning message displayed when an error is ignored.
+  - Example: `"Ignoring safe-to-ignore errors"`.
+- `signals` (Optional): Key-value pairs used to emit signals to external systems.
+  - Example: `safe_to_revert = true` indicates it is safe to revert the operation if it fails.
```
yhakbar, line 1354: I think we should add a note here indicating what this does.

e.g.

```
Populating values into the `signals` attribute results in a JSON file named `error-signals.json` being emitted on failure. This file can be useful to inspect in CI/CD systems to determine what the recommended course of action is to address the failure.

In this example, an error has occurred, the author of the unit has signaled that it is safe to revert adjustments to this unit. A CI/CD system could have a standard process of finding all units with files named `error-signals.json`, then checkout the previous commit and apply those units in that state, reverting their updates.
```

options/options.go
```diff
@@ -795,5 +803,245 @@ type EngineOptions struct {
...
+	if err == nil {
+		return nil, nil
+	}
+
+	errStr := err.Error()
+	action := &ErrorAction{}
+
+	// First check ignore rules
+	for _, ignoreBlock := range c.Ignore {
+		isIgnorable, err := matchesAnyPattern(errStr, ignoreBlock.IgnorableErrors)
```
yhakbar, line 979: We should compile the regexes  in configurations before we call `ProcessError`, by parsing the configurations, then immediately doing the post-processing of the relevant regexes.

That would allow this method to be safer, as the user shouldn't learn that their regex is invalid in the middle of an error during a `run-all`. They should learn it while the DAG is being constructed because parsing failed.

options/options.go
```diff
@@ -795,5 +803,245 @@ type EngineOptions struct {
...
+				action.IgnoreSignals[k] = v
+			}
+
+			return action, nil
+		}
+	}
+
+	// Then check retry rules
+	for _, retryBlock := range c.Retry {
+		isRetryable, err := matchesAnyPattern(errStr, retryBlock.RetryableErrors)
```
yhakbar, line 1000: Same issue here as described above.

options/options.go
```diff
@@ -795,5 +803,245 @@ type EngineOptions struct {
...
+func matchesAnyPattern(input string, patterns []string) (bool, error) {
+	for _, pattern := range patterns {
+		// Handle negative patterns (patterns starting with !)
+		isNegative := false
+		if len(pattern) > 0 && pattern[0] == '!' {
+			isNegative = true
+			pattern = pattern[1:]
+		}
+
+		matched, err := regexp.MatchString(pattern, input)
```
yhakbar, line 1033: In addition to the comment above, note that it's also more efficient to have this compiled earlier so that we don't recompile all the patterns for every error.

docs/_docs/04_reference/config-blocks-and-attributes.md
denis256: Updated docs

yhakbar (approved):

PR #3588, opened 2024-11-21
feat: Deprecating `terragrunt.hcl` as root

yhakbar:
## Description

Making it so that usage of a root `terragrunt.hcl` throws a warning for users, and gives them guidance to move to a differently named root configuration file.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Updated `strict` package.
Updated `config` package.
Updated `scaffold` command.
Updated `catalog` command.

### Migration Guide

Current idiomatic usage of Terragrunt involves using a root configuration file named `terragrunt.hcl`.

This causes a lot of confusion for users, and potential errors. With this change, the new recommendation of naming it anything else will be enforced via the `strict` package.

It will likely take a very long time before this functionality can be removed, so we should start updating our documentation to reflect this change and throw warnings that users are using less reliable configurations.


test/integration_test.go
```diff
@@ -1085,48 +1087,6 @@ func TestTerragruntExcludeExternalDependencies(t *testing.T) {
 	assert.NotContains(t, applyAllStdoutString, "Hello World, "+excludedModule)
 }
 
-func TestApplySkipTrue(t *testing.T) {
```
yhakbar, line 1088: - [ ] TODO: Make sure to reintroduce this.

denis256 (dismissed):

test/integration_gcp_test.go
```diff
@@ -121,10 +121,10 @@ func TestGcpParallelStateInit(t *testing.T) {
 		require.NoError(t, err)
 	}
 
-	tmpTerragruntConfigFile := util.JoinPath(tmpEnvPath, "terragrunt.hcl")
+	tmpTerragruntConfigFile := util.JoinPath(tmpEnvPath, "root.hcl")
 	project := os.Getenv("GOOGLE_CLOUD_PROJECT")
 	gcsBucketName := "terragrunt-test-bucket-" + strings.ToLower(helpers.UniqueID())
-	tmpTerragruntGCSConfigPath := createTmpTerragruntGCSConfig(t, testFixtureGcsParallelStateInit, project, terraformRemoteStateGcpRegion, gcsBucketName, config.DefaultTerragruntConfigPath)
+	tmpTerragruntGCSConfigPath := createTmpTerragruntGCSConfig(t, testFixtureGcsParallelStateInit, project, terraformRemoteStateGcpRegion, gcsBucketName, "root.hcl")
```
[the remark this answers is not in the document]
yhakbar: Ya, but it's a pain to update them all, and I don't think it's that important. I'll look to use that constant more consistently in the future.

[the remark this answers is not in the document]
denis256: Ok

levkohimins (changes requested):
You don't really need the `Name` field  in `strict.Control `

```diff
--- a/internal/strict/strict.go
+++ b/internal/strict/strict.go
@@ -21,8 +21,6 @@ import (
 // Control represents a control that can be enabled or disabled in strict mode.
 // When the control is enabled, Terragrunt will behave in a way that is not backwards compatible.
 type Control struct {
-	// Name is the name of the strict control.
-	Name string
 	// Error is the error that will be returned when the control is enabled.
 	Error error
 	// Warning is a warning that will be logged when the control is not enabled.
@@ -57,7 +55,7 @@ const (
 )
 
 // GetStrictControl returns the strict control with the given name.
-func GetStrictControl(name string) (Control, bool) {
+func GetStrictControl(name string) (*Control, bool) {
 	control, ok := StrictControls[name]
 
 	return control, ok
@@ -65,8 +63,8 @@ func GetStrictControl(name string) (Control, bool) {
 
 // Evaluate returns a warning if the control is not enabled, an indication of whether the control has already been triggered,
 // and an error if the control is enabled.
-func (control Control) Evaluate(opts *options.TerragruntOptions) (string, bool, error) {
-	_, triggered := TriggeredControls.LoadAndStore(control.Name, true)
+func (control *Control) Evaluate(opts *options.TerragruntOptions) (string, bool, error) {
+	_, triggered := TriggeredControls.LoadAndStore(control, true)
 
 	if opts.StrictMode {
 		return "", triggered, control.Error
@@ -88,73 +86,61 @@ func (control Control) Evaluate(opts *options.TerragruntOptions) (string, bool,
 	return control.Warning, triggered, nil
 }
 
-type Controls map[string]Control
+type Controls map[string]*Control
 
 //nolint:lll,gochecknoglobals,stylecheck,revive
 var StrictControls = Controls{
 	SpinUp: {
-		Name:    SpinUp,
 		Error:   errors.Errorf("The `%s` command is no longer supported. Use `terragrunt run-all apply` instead.", SpinUp),
 		Warning: fmt.Sprintf("The `%s` command is deprecated and will be removed in a future version. Use `terragrunt run-all apply` instead.", SpinUp),
 	},
 	TearDown: {
-		Name:    TearDown,
 		Error:   errors.Errorf("The `%s` command is no longer supported. Use `terragrunt run-all destroy` instead.", TearDown),
 		Warning: fmt.Sprintf("The `%s` command is deprecated and will be removed in a future version. Use `terragrunt run-all destroy` instead.", TearDown),
 	},
 	PlanAll: {
-		Name:    PlanAll,
 		Error:   errors.Errorf("The `%s` command is no longer supported. Use `terragrunt run-all plan` instead.", PlanAll),
 		Warning: fmt.Sprintf("The `%s` command is deprecated and will be removed in a future version. Use `terragrunt run-all plan` instead.", PlanAll),
 	},
 	ApplyAll: {
-		Name:    ApplyAll,
 		Error:   errors.Errorf("The `%s` command is no longer supported. Use `terragrunt run-all apply` instead.", ApplyAll),
 		Warning: fmt.Sprintf("The `%s` command is deprecated and will be removed in a future version. Use `terragrunt run-all apply` instead.", ApplyAll),
 	},
 	DestroyAll: {
-		Name:    DestroyAll,
 		Error:   errors.Errorf("The `%s` command is no longer supported. Use `terragrunt run-all destroy` instead.", DestroyAll),
 		Warning: fmt.Sprintf("The `%s` command is deprecated and will be removed in a future version. Use `terragrunt run-all destroy` instead.", DestroyAll),
 	},
 	OutputAll: {
-		Name:    OutputAll,
 		Error:   errors.Errorf("The `%s` command is no longer supported. Use `terragrunt run-all output` instead.", OutputAll),
 		Warning: fmt.Sprintf("The `%s` command is deprecated and will be removed in a future version. Use `terragrunt run-all output` instead.", OutputAll),
 	},
 	ValidateAll: {
-		Name:    ValidateAll,
 		Error:   errors.Errorf("The `%s` command is no longer supported. Use `terragrunt run-all validate` instead.", ValidateAll),
 		Warning: fmt.Sprintf("The `%s` command is deprecated and will be removed in a future version. Use `terragrunt run-all validate` instead.", ValidateAll),
 	},
 	SkipDependenciesInputs: {
-		Name:    SkipDependenciesInputs,
 		Error:   errors.Errorf("The `%s` option is deprecated. Reading inputs from dependencies has been deprecated and will be removed in a future version of Terragrunt. To continue using inputs from dependencies, forward them as outputs.", SkipDependenciesInputs),
 		Warning: fmt.Sprintf("The `%s` option is deprecated and will be removed in a future version of Terragrunt. Reading inputs from dependencies has been deprecated. To continue using inputs from dependencies, forward them as outputs.", SkipDependenciesInputs),
 	},
 	DisableLogFormatting: {
-		Name:    DisableLogFormatting,
 		Error:   errors.Errorf("The `--%s` flag is no longer supported. Use `--terragrunt-log-format=key-value` instead.", DisableLogFormatting),
 		Warning: fmt.Sprintf("The `--%s` flag is deprecated and will be removed in a future version. Use `--terragrunt-log-format=key-value` instead.", DisableLogFormatting),
 	},
 	JSONLog: {
-		Name:    JSONLog,
 		Error:   errors.Errorf("The `--%s` flag is no longer supported. Use `--terragrunt-log-format=json` instead.", JSONLog),
 		Warning: fmt.Sprintf("The `--%s` flag is deprecated and will be removed in a future version. Use `--terragrunt-log-format=json` instead.", JSONLog),
 	},
 	TfLogJSON: {
-		Name:    TfLogJSON,
 		Error:   errors.Errorf("The `--%s` flag is no longer supported. Use `--terragrunt-log-format=json` instead.", TfLogJSON),
 		Warning: fmt.Sprintf("The `--%s` flag is deprecated and will be removed in a future version. Use `--terragrunt-log-format=json` instead.", TfLogJSON),
 	},
 	RootTerragruntHCL: {
-		Name:    RootTerragruntHCL,
 		Error:   errors.Errorf("Using `terragrunt.hcl` as the root of Terragrunt configurations is an anti-pattern, and no longer supported. Use a differently named file like `root.hcl` instead. For more information, see https://terragrunt.gruntwork.io/docs/migrate/migrating-from-root-terragrunt-hcl"),
 		Warning: "Using `terragrunt.hcl` as the root of Terragrunt configurations is an anti-pattern, and no longer recommended. In a future version of Terragrunt, this will result in an error. You are advised to use a differently named file like `root.hcl` instead. For more information, see https://terragrunt.gruntwork.io/docs/migrate/migrating-from-root-terragrunt-hcl",
 	},
 }
 
-var TriggeredControls = xsync.NewMapOf[string, bool]()
+var TriggeredControls = xsync.NewMapOf[*Control, bool]()
 
 // Names returns the names of all strict controls.
 func (controls Controls) Names() []string {
```

internal/strict/strict.go
```diff
@@ -58,78 +63,99 @@ func GetStrictControl(name string) (Control, bool) {
...
 		Warning: fmt.Sprintf("The `--%s` flag is deprecated and will be removed in a future version. Use `--terragrunt-log-format=json` instead.", TfLogJSON),
 	},
+	RootTerragruntHCL: {
+		Name:    RootTerragruntHCL,
+		Error:   errors.Errorf("Using `terragrunt.hcl` as the root of Terragrunt configurations is an anti-pattern, and no longer supported. Use a differently named file like `root.hcl` instead. For more information, see https://terragrunt.gruntwork.io/docs/migrate/migrating-from-root-terragrunt-hcl"),
+		Warning: "Using `terragrunt.hcl` as the root of Terragrunt configurations is an anti-pattern, and no longer recommended. In a future version of Terragrunt, this will result in an error. You are advised to use a differently named file like `root.hcl` instead. For more information, see https://terragrunt.gruntwork.io/docs/migrate/migrating-from-root-terragrunt-hcl",
+	},
 }
 
+var TriggeredControls = xsync.NewMapOf[string, bool]()
```
levkohimins, line 157: Global variable can cause problems in tests, especially with `-race`.

yhakbar: Ya, but does it really matter? 

As long as we test the stderr check integration test out of parallel, we should be fine, right?
https://github.com/gruntwork-io/terragrunt/pull/3588/files#diff-4e1bb7ad3f548fa9a54e7e0f4b57c046d000c335ba4802df083a8e93416d5abfR85

In unit tests, I just ignore the `triggered` status:
https://github.com/gruntwork-io/terragrunt/pull/3588/files#diff-5e4aa2fd41e755849a1962333f14613e6189d0aaf00d10eac0a756cebe1692ccR68

yhakbar:
> You don't really need the `Name` field in `strict.Control`.

I considered this, but I wonder if it's actually better. I don't like Golang pointers as it's easy to create nil pointer exceptions, and I wouldn't want to put the whole struct in the triggered map.

Is it that bad to duplicate the name of the control in the value found by the key?

internal/strict/strict.go
levkohimins: Until we create another test to check the strict-mode (in config or CLI), this should work stably. It's up to you, we just need remember this.

levkohimins:
> > You don't really need the `Name` field in `strict.Control`.
> 
> I considered this, but I wonder if it's actually better. I don't like Golang pointers as it's easy to create nil pointer exceptions, and I wouldn't want to put the whole struct in the triggered map.
> 
> Is it that bad to duplicate the name of the control in the value found by the key?

Not sure if it's a good idea to duplicate the same string data in the map key and field structure, I see two options:
1. What I already suggested (I don't see anything wrong with using a pointer).
2. Remake `StrictControls` from map to slice.

docs/_docs/04_reference/built-in-functions.md
```diff
@@ -91,21 +91,15 @@ If there is a specific function you would like to see supported directly in Terr
 
 ## find_in_parent_folders
```
yhakbar, line 92: Add in a note on the historical legacy behavior of no params.

internal/strict/strict.go
```diff
@@ -147,23 +167,37 @@ var (
...
+	var warning string
+	if len(completedControls) > 0 {
+		warning = fmt.Sprintf("The following strict control(s) are already completed: %s. Please remove any completed strict controls, as setting them no longer does anything. For a list of all ongoing strict controls, and the outcomes of previous strict controls, see https://terragrunt.gruntwork.io/docs/reference/strict-mode",
+			strings.Join(completedControls, ", "))
+	}
+
+	var err error
 	if len(invalidControls) > 0 {
-		return fmt.Errorf("%w\nInvalid value(s):\n- %s\nAllowed value(s):\n- %s",
+		err = fmt.Errorf("%w\nInvalid control(s):\n- %s\nAllowed control(s):\n- %s",
```
levkohimins, line 195:
```suggestion
		err = errors.Errorf("%w\nInvalid control(s):\n- %s\nAllowed control(s):\n- %s",
```

We should use our `internal/errors` package everywhere, otherwise we will lose the stack trace.

cli/commands/catalog/tui/tui.go
```diff
@@ -8,10 +8,27 @@ import (
 	tea "github.com/charmbracelet/bubbletea"
 
 	"github.com/gruntwork-io/terragrunt/cli/commands/catalog/module"
+	"github.com/gruntwork-io/terragrunt/config"
+	"github.com/gruntwork-io/terragrunt/internal/strict"
 	"github.com/gruntwork-io/terragrunt/options"
 )
 
 func Run(ctx context.Context, modules module.Modules, opts *options.TerragruntOptions) error {
+	if opts.ScaffoldRootFileName == "" {
```
levkohimins, line 17: You also check this option in `cli/commands/catalog/command.go` `Action` flag func. Maybe it would be better to combine them and keep them in this func?

cli/commands/scaffold/action.go
```diff
@@ -96,6 +101,21 @@ inputs = {
 var moduleURLRegex = regexp.MustCompile(moduleURLPattern)
 
 func Run(ctx context.Context, opts *options.TerragruntOptions, moduleURL, templateURL string) error {
+	if opts.ScaffoldRootFileName == "" {
```
levkohimins, line 104: Duplicated code. If someone else works on this code, he most likely won't know that changes need to be synchronized in several places. I would recommend keeping this code in one place and just referencing it, or even better moving it into a scaffold.RootFileName flag and defining it in `commands/flags.go`

```go
func NewRootFileNameFlag(opts *options.TerragruntOptions) cli.Flag {
	return &cli.GenericFlag[string]{
			Name:        scaffold.RootFileName,
			Destination: &opts.ScaffoldRootFileName,
			Usage:       "Name of the root Terragrunt configuration file, if used.",
			Action: func(ctx *cli.Context, value string) error {
				if value == config.DefaultTerragruntConfigPath {
					if control, ok := strict.GetStrictControl(strict.RootTerragruntHCL); ok {
						warn, triggered, err := control.Evaluate(opts)
						if err != nil {
							return err
						}

						if !triggered {
							opts.Logger.Warnf(warn)
						}
					}
				}

				return nil
			},
		},
}
```

them in both places scaffold/command.go and catalog/command.go just referencing to that func

```go
func NewFlags(opts *options.TerragruntOptions) cli.Flags {
	return cli.Flags{
		commands.NewRootFileNameFlag(opts),
                ....
	}
}
```

cli/commands/scaffold/action.go
```diff
@@ -166,6 +186,19 @@ func Run(ctx context.Context, opts *options.TerragruntOptions, moduleURL, templa
...
+		vars["EnableRootInclude"] = !opts.ScaffoldNoIncludeRoot
+	} else {
+		opts.Logger.Warnf("The EnableRootInclude variable is already set in the var flag(s). The --%s flag will be ignored.", NoIncludeRoot)
+	}
+
+	if _, found := vars["RootFileName"]; !found {
+		vars["RootFileName"] = opts.ScaffoldRootFileName
+	} else {
+		opts.Logger.Warnf("The RootFileName variable is already set in the var flag(s). The --%s flag will be ignored.", RootFileName)
+	}
```
levkohimins, line 200:
```go
const rootFileNameVarName = "RootFileName"
```

```suggestion
	if _, found := vars[rootFileNameVarName]; !found {
		vars[rootFileNameVarName] = opts.ScaffoldRootFileName
	} else {
		opts.Logger.Warnf("The %s variable is already set in the var flag(s). The --%s flag will be ignored.", rootFileNameVarName, RootFileName)
	}
```

Consider using a constant. This way we will be sure that the variable name in the message log will always be up-to-date.

levkohimins (changes requested):
Left some recommendations for the code.

options/options.go
```diff
@@ -306,6 +306,12 @@ type TerragruntOptions struct {
 	// Files with variables to be used in modules scaffolding.
 	ScaffoldVarFiles []string
 
+	// Do not include root unit in scaffolding.
+	ScaffoldNoIncludeRoot bool
+
+	// Name of the root Terragrunt configuration file, if used.
+	ScaffoldRootFileName string
```
denis256, line 313: I think `Clone()` should be updated

yhakbar: Why would these need to be cloned? Scaffold isn't used in `run-all`.

yhakbar: I added it just to keep this moving.

denis256 (approved):

levkohimins (approved):
Great job!

PR #3602, opened 2024-11-25
fix: Wrap TF stdout and stderr in JSON

levkohimins:
## Description

The issue is that new logger does not wrap stdout and stderr in json, as it worked with the `--terragrunt-tf-logs-to-json` flag.
This PR will be deprecated this flag, and using it will not change anything. Since the json format is intended for parsing, one more field does not make the logs worse. Therefore the `executedCommandArgs` field will now always be present in the json log, and there is no longer any point in using the `--terragrunt-tf-logs-to-json` flag. 

## TODOs

- [ ] Update the docs.

## Release Notes (draft)


cli/provider_cache.go
```diff
@@ -350,14 +351,18 @@ func runTerraformCommand(ctx context.Context, opts *options.TerragruntOptions, a
 	cloneOpts.WorkingDir = opts.WorkingDir
 	cloneOpts.TerraformCliArgs = args
 	cloneOpts.Env = envs
-	cloneOpts.ForwardTFStdout = true
 
+	output, err := shell.RunTerraformCommandWithOutput(ctx, cloneOpts, cloneOpts.TerraformCliArgs...)
 	// If the Terraform error matches `httpStatusCacheProviderReg` we ignore it and hide the log from users, otherwise we process the error as is.
-	if output, err := shell.RunTerraformCommandWithOutput(ctx, cloneOpts, cloneOpts.TerraformCliArgs...); err != nil && len(errWriter.Msgs()) == 0 {
-		return output, err
+	if err != nil && httpStatusCacheProviderReg.Match(output.Stderr.Bytes()) {
```
yhakbar, line 357: Nice 👍🏽

docs/_docs/02_features/custom-log-format.md
```diff
@@ -56,7 +56,7 @@ Placeholders have preset names:
 
 * `%prefix` - Path to the working directory were Terragrunt is running.
 
-* `%tfpath` - Path to the OpenTofu/Terraform executable (as defined by [terragrunt-tfpath](https://terragrunt.gruntwork.io/docs/reference/cli-options/#terragrunt-tfpath)).
+* `%tf-path` - Path to the OpenTofu/Terraform executable (as defined by [terragrunt-tfpath](https://terragrunt.gruntwork.io/docs/reference/cli-options/#terragrunt-tfpath)).
```
yhakbar, line 59: Make sure to call out this change in the release notes.

internal/strict/strict.go
```diff
@@ -83,45 +85,49 @@ type Controls map[string]Control
 //nolint:lll,gochecknoglobals,stylecheck
 var StrictControls = Controls{
 	SpinUp: {
-		Error:   errors.Errorf("The `%s` command is no longer supported. Use `terragrunt run-all apply` instead.", SpinUp),
```
yhakbar, line 86: Because this error is user facing, and it's more likely to involve multiple sentences don't you think it's better to have them be full sentences? If so, we should just add a `nolint` for whatever made you adjust this.

yhakbar (changes requested):

internal/strict/strict.go
yhakbar: Alternatively, maybe you know a special way to convert the error messages to proper casing before they are reported to users?

docs/_docs/02_features/custom-log-format.md
levkohimins: Sure, I will call this out. Thanks.

internal/strict/strict.go
levkohimins: Sure, I did what you suggested.

levkohimins: > Alternatively, maybe you know a special way to convert the error messages to proper casing before they are reported to users?

Ah no, I don't know of any ready-made function that could do this, only manual message conversion, but I don't think it's worth it.

yhakbar (approved):

PR #3607, opened 2024-11-28
add new built-in function get_aws_account_alias(), fix #2483

### LINKED ISSUE #2483: Need new get_aws_account_aliases() built-in function
Hello

I'm using terragrunt in a company where the account alias is used to create the name of the tfstate bucket.

Currently, we are running an external `aws` command to list the accounts alias from `terragrunt.hcl`, but this is brittle.

To be able to get rid of this brittle code, I need a way to get the list of aliases with a builtin function. 

Could you add a new function similar to `get_aws_account_id` to get the list of account aliases ?

All the best

wakeful:
## Description
add new built-in function `get_aws_account_alias()`. Fixes #2483.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added new built-in function `get_aws_account_alias()`. Fixes #2483.

### Migration Guide
n/a

### DIFF
--- a/awshelper/config.go
+++ b/awshelper/config.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws/request"
+	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/gruntwork-io/go-commons/version"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -355,6 +356,31 @@ func GetAWSPartition(config *AwsSessionConfig, terragruntOptions *options.Terrag
 	}
 
 	return arn.Partition, nil
+}
+
+// GetAWSAccountAlias gets the AWS account Alias of the current session configuration,
+// if there is no alias an empty string is return.
+func GetAWSAccountAlias(config *AwsSessionConfig, terragruntOptions *options.TerragruntOptions) (string, error) {
+	sess, err := CreateAwsSession(config, terragruntOptions)
+	if err != nil {
+		return "", errors.New(err)
+	}
+
+	aliases, err := iam.New(sess).ListAccountAliases(nil)
+	if err != nil {
+		return "", errors.New(err)
+	}
+
+	if len(aliases.AccountAliases) != 1 { // AWS supports only one alias per account https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListAccountAliases.html
+		return "", nil
+	}
+
+	alias := aliases.AccountAliases[0]
+	if alias == nil {
+		return "", errors.Errorf("expected AWS account alias, got nil")
+	}
+
+	return *alias, nil
 }
 
 // GetAWSAccountID gets the AWS account ID of the current session configuration.
--- a/docs/_docs/04_reference/built-in-functions.md
+++ b/docs/_docs/04_reference/built-in-functions.md
@@ -29,6 +29,7 @@ Terragrunt allows you to use built-in functions anywhere in `terragrunt.hcl`, ju
 - [get\_terraform\_commands\_that\_need\_input](#get_terraform_commands_that_need_input)
 - [get\_terraform\_commands\_that\_need\_locking](#get_terraform_commands_that_need_locking)
 - [get\_terraform\_commands\_that\_need\_parallelism](#get_terraform_commands_that_need_parallelism)
+- [get\_aws\_account\_alias](#get_aws_account_alias)
 - [get\_aws\_account\_id](#get_aws_account_id)
 - [get\_aws\_caller\_identity\_arn](#get_aws_caller_identity_arn)
 - [get\_terraform\_command](#get_terraform_command)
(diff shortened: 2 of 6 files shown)


docs/_docs/04_reference/built-in-functions.md
```diff
@@ -552,6 +553,17 @@ terraform {
   }
 }
 ```
+## get_aws_account_alias
```
levkohimins, line 556:
```suggestion

## get_aws_account_alias
```

markdownlint error:

> docs/_docs/04_reference/built-in-functions.md:556 MD022/blanks-around-headings Headings should be surrounded by blank lines [Expected: 1; Actual: 0; Above] [Context: "## get_aws_account_alias"]

levkohimins (changes requested):
Thank you for the contribution! Could you add an integration test that checks this feature to avoid regression in the future?

wakeful:
> Thank you for the contribution! Could you add an integration test that checks this feature to avoid regression in the future?

hey @levkohimins

After reviewing the code, most functions in the `awshelper` package are missing tests. Can you confirm how I should implement the `integration test`?

- Should I `mock` the AWS client, or do you prefer making actual calls to one of your AWS accounts with hard-coded values in the assertions?
- Should I guard the test with a `//go:build aws` directive to prevent it from running outside of an AWS environment?

awshelper/config.go
```diff
@@ -357,6 +358,25 @@ func GetAWSPartition(config *AwsSessionConfig, terragruntOptions *options.Terrag
...
+		return "", errors.New(err)
+	}
+
+	aliases, err := iam.New(sess).ListAccountAliases(nil)
+	if err != nil {
+		return "", errors.New(err)
+	}
+
+	if len(aliases.AccountAliases) > 0 {
+		return *aliases.AccountAliases[0], nil
```
denis256, line 374: can `res.AccountAliases[0]` have value `nil`?

levkohimins:
Hey @wakeful,

> * Should I `mock` the AWS client, or do you prefer making actual calls to one of your AWS accounts with hard-coded values in the assertions?

Unit tests we keep in the same package (folder) as the code, but integration tests in the `test/` folder. You can use [this](https://github.com/gruntwork-io/terragrunt/blob/d8313f9099d92500a008b1677c663253e82fcbd9/test/integration_aws_test.go#L472) example that tests `get_aws_account_id` built-in function.

> * Should I guard the test with a `//go:build aws` directive to prevent it from running outside of an AWS environment?

`test/integration_aws_test.go` already contains this directive, so just please add your test to that file.

awshelper/config.go
wakeful: the underlying type is `[]*string`, so a `nil` value is possible, but according to AWS documentation

>AWS supports only one alias per account https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListAccountAliases.html

a `nil` value would suggest a bug in the AWS API, but I will add an extra check.

wakeful:
hey @denis256, @levkohimins 

I need advice on the implementation/behaviour.

After some additional research, it seems that AWS returns an empty slice when the AWS account does not have an alias set up.

I found two scenarios:
- If the AWS account is created manually, AWS does not set an alias by default; the admin needs to add it manually in the IAM Dashboard.
- If the AWS account is created via AWS Organisations, the account name is used as the alias. The admin can then create a new alias in the IAM Dashboard. If the admin deletes the alias, AWS will return an empty response (the AWS Organisations account name will no longer be used).

I would like to ask for your guidance on what we should do in case AWS returns an empty response - when the AWS account has no alias (or it was recently deleted).

Originally, I wanted to fall back to returning the AWS Account ID, but since there is a chance this function will be used as part of the `state path`, perhaps we should return an empty string instead to allow the Terragrunt user to `make a decision` e.g., use a ternary operator to call `get_aws_account_id` in a more obvious way. Or should we crash the app with an error?

levkohimins:
> hey @denis256, @levkohimins
> 
> I need advice on the implementation/behaviour.
> 
> After some additional research, it seems that AWS returns an empty slice when the AWS account does not have an alias set up.
> 
> I found two scenarios:
> 
> * If the AWS account is created manually, AWS does not set an alias by default; the admin needs to add it manually in the IAM Dashboard.
> * If the AWS account is created via AWS Organisations, the account name is used as the alias. The admin can then create a new alias in the IAM Dashboard. If the admin deletes the alias, AWS will return an empty response (the AWS Organisations account name will no longer be used).
> 
> I would like to ask for your guidance on what we should do in case AWS returns an empty response - when the AWS account has no alias (or it was recently deleted).
> 
> Originally, I wanted to fall back to returning the AWS Account ID, but since there is a chance this function will be used as part of the `state path`, perhaps we should return an empty string instead to allow the Terragrunt user to `make a decision` e.g., use a ternary operator to call `get_aws_account_id` in a more obvious way. Or should we crash the app with an error?

Hey @wakeful, I think it should return just an empty string if the AWS account does not have an alias. And further actions should be taken by users, applying the conditions.

denis256:
If AWS returns an empty response, we should return an empty string. 
This approach will allow Terragrunt users to handle missing aliases explicitly, such as using a ternary operator to fall back to the AWS Account ID.

wakeful:
Thanks for your suggestions, @denis256 & @levkohimins. The code has been updated, and the new function will now return an empty string if there is no AWS Account Alias.

levkohimins (approved):
Great job @wakeful. Thanks!

levkohimins:
Resolved in [v0.69.7](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.69.7) release.

PR #3627, opened 2024-12-05
Stacks: generate

denis256:
## Description

Add support for `stacks generate` cli command

Included changes:
* added `stacks generate` cli command
* added tests for local and remote stacks
* added cli documentation

In this release, *recursive stacks* are not yet supported.

RFC: #3313

## TODOs

- [x] Update the docs.

## Release Notes (draft)


cli/commands/stack/command.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/gruntwork-io/terragrunt/pkg/cli"
+)
+
+const (
+	CommandName = "stack"
+)
+
+func NewFlags(opts *options.TerragruntOptions) cli.Flags {
+	return cli.Flags{
+		commands.NewNoIncludeRootFlag(opts),
```
yhakbar, line 15: I know this is WIP, but I'm curious about this. Do you think we need this for `stack` commands?

cli/commands/stack/action.go
```diff
@@ -0,0 +1,228 @@
...
+		opts.Logger.Infof("Processing unit: %s (%s) to %s", unit.Name, src, dest)
+
+		client := &getter.Client{
+			Src:             src,
+			Dst:             dest,
+			Mode:            getter.ClientModeAny,
+			Dir:             true,
+			DisableSymlinks: true,
+			Options: []getter.ClientOption{
+				getter.WithInsecure(),
```
yhakbar, line 87: Why are we doing this? Do you mind adding a comment documenting that?

cli/commands/stack/action.go
```diff
@@ -0,0 +1,228 @@
...
+
+		if err := client.Get(); err != nil {
+			return errors.New(err)
+		}
+	}
+
+	return nil
+}
+
+func isURL(str string) bool {
```
yhakbar, line 112: Does this mean that `github.com/...` URLs won't be supported?

cli/commands/stack/action.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+// ClientMode determines if we're getting a directory or single file
+func (p *StacksFileProvider) ClientMode(u *url.URL) (getter.ClientMode, error) {
+	fi, err := os.Stat(u.Path)
+	if err != nil {
+		return getter.ClientModeInvalid, errors.New(err)
+	}
+
+	if fi.IsDir() {
```
yhakbar, line 154: Why are we ever fetching a file? We can just always set it to get a directory, right? Do we recursively fetch the individual files after loading the directory?

config/stack.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/gruntwork-io/terragrunt/config/hclparse"
+	"github.com/gruntwork-io/terragrunt/internal/errors"
+	"github.com/gruntwork-io/terragrunt/options"
+)
+
+// StackConfigFile represents the structure of terragrunt.stack.hcl stack file
+type StackConfigFile struct {
+	Locals *terragruntLocal `cty:"locals"  hcl:"locals,block"`
+	Units  []*Unit          `cty:"unit" hcl:"unit,block"`
```
yhakbar, line 16: We should point out in the PR body that we're not supporting recursive stacks yet. It should also make its way to the release notes.

config/stack.go
```diff
@@ -0,0 +1,97 @@
...
+
+	if err != nil {
+		return errors.New(err)
+	}
+
+	if len(localsBlock) == 0 {
+		return nil
+	}
+
+	attrs, err := localsBlock[0].JustAttributes()
```
yhakbar, line 57: Is this always how locals work? The first block is the only one that matters? If so, we should either throw an error if users use multiple locals blocks, or properly combine them all.

config/stack.go
```diff
@@ -0,0 +1,97 @@
...
+		return errors.New(err)
+	}
+
+	evaluatedLocals := map[string]cty.Value{}
+	evaluated := true
+
+	for iterations := 0; len(attrs) > 0 && evaluated; iterations++ {
+		if iterations > MaxIter {
+			// Reached maximum supported iterations, which is most likely an infinite loop bug so cut the iteration
+			// short an return an error.
```
yhakbar, line 69:
```suggestion
			// short and return an error.
```

config/unit.go
```diff
@@ -0,0 +1,54 @@
+package config
+
+import (
+	"fmt"
+
+	"github.com/zclconf/go-cty/cty"
+	"github.com/zclconf/go-cty/cty/gocty"
+)
+
+// Unit represents a list of units.
```
yhakbar, line 10: No it doesn't? It represents a single unit, right?

config/unit.go
```diff
@@ -0,0 +1,54 @@
...
+type Unit struct {
+	Name   string `cty:"name"    hcl:",label"`
```
yhakbar, line 12: NIT: Let's keep the `cty` and `hcl` tags on the same side, please.

config/unit.go
```diff
@@ -0,0 +1,54 @@
...
+	Source string `hcl:"source,attr" cty:"source"`
+	Path   string `hcl:"path,attr" cty:"path"`
+}
+
+// ToCtyValue converts StackConfigFile to cty.Value
+func (s *StackConfigFile) ToCtyValue() (cty.Value, error) {
```
yhakbar, line 18: Do we need to have the file available as a cty value right now? We're going to integrate reference to the generating `terragrunt.stack.hcl` file in the future once we support a `values` attribute, but is this doing anything right now?

docs/_docs/04_reference/cli-options.md
```diff
@@ -741,6 +742,48 @@ Notes:
 
 - destroy will be executed only on subset of services dependent from `eks-service-3`
 
+### stack
+
+The `terragrunt stack` commands provide a new interface for managing collections of Terragrunt units defined in `terragrunt.stack.hcl` files.
```
yhakbar, line 747:
```suggestion
The `terragrunt stack` commands provide an interface for managing collections of Terragrunt units defined in `terragrunt.stack.hcl` files.
```

Let's try to keep these docs evergreen.

docs/_docs/04_reference/cli-options.md
```diff
@@ -741,6 +742,48 @@ Notes:
...
+These commands simplify the process of handling multiple infrastructure units by grouping them into a "stack," reducing code duplication and streamlining operations across environments.
```
yhakbar, line 748:
```suggestion
These commands simplify the process of handling multiple infrastructure units by grouping them into a "stack", reducing code duplication and streamlining operations across environments.
```

docs/_docs/04_reference/cli-options.md
```diff
@@ -741,6 +742,48 @@ Notes:
...
+
+The `terragrunt stack generate` command is used to generate a stack of Terragrunt `hcl` files based on the configuration provided in the `terragrunt.stack.hcl` file.
```
yhakbar, line 750:
```suggestion
The `terragrunt stack generate` command is used to generate a stack of `terragrunt.hcl` files based on the configuration provided in the `terragrunt.stack.hcl` file.
```

cli/commands/stack/action.go
yhakbar: It might be better to see if the source starts with `.` or `/` to indicate that it's a relative or absolute file path respectively.

options/options.go
```diff
@@ -82,6 +82,8 @@ type TerragruntOptions struct {
 	// Location of the Terragrunt config file
 	TerragruntConfigPath string
 
+	TerragrungStackConfigPath string
```
yhakbar, line 85: Typo. `TerragrungStackConfigPath` --> `TerragruntStackConfigPath`

yhakbar (changes requested):
I have mostly small adjustments requested here.

The most important changes I'm requesting are documentation changes for the docs site and the release notes.

cli/commands/stack/action.go
denis256: In case of `github.com/` you don't know if it is local directory or remote... added workaround to check for `//`

yhakbar: I was saying that we can require that local sources always start with `./` for a relative path or `/` for an absolute path.

I'm not sure a user would ever want to do this, but could they define their stack definition at the root of a repository?

e.g.

`github.com/acme/my-favorite-stack`

cli/commands/stack/action.go
denis256: Removed, it was used for local tests/verification of handling symlinks

cli/commands/stack/action.go
denis256: Are copied files recursively even if we have multiple subdirectories

config/stack.go
denis256: Updated description

config/stack.go
denis256: I added a check for multiple blocks to get an error in the case of multiple locals.

config/unit.go
denis256: Merged unit.go with stack.go since we only used unit structure from it.

yhakbar:
@denis256 , if you don't mind, could you put the stacks command behind an experiment flag for now? Users should have to opt-in to use stacks until we're ready for general usage so that we can feel more comfortable changing the API.

cli/commands/stack/action.go
```diff
@@ -0,0 +1,242 @@
...
+
+// Run runs the stack command.
+func Run(ctx context.Context, opts *options.TerragruntOptions, subCommand string) error {
+	stacksEnabled := opts.Experiments[experiment.Stacks]
+	if !stacksEnabled.Enabled {
+		return errors.New("stacks experiment is not enabled use --experiment stacks to enable it")
+	}
+
+	if subCommand == "" {
+		return errors.New("No command specified")
```
yhakbar, line 36: Sorry, I didn't think of this until now, but could you return a help message instead of only throwing an error? I think that's a better pattern to follow for commands with required arguments.

config/stack.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		return errors.New(err)
+	}
+
+	if len(localsBlock) == 0 {
+		return nil
+	}
+
+	if len(localsBlock) > 1 {
+		return errors.New(fmt.Sprintf("only one locals block is allowed %s stack file, "+
```
yhakbar, line 67: Let's change this to:

"Only one locals block is allowed per stack file, but found %d in %s"

test/fixtures/stacks/remote/terragrunt.stack.hcl
```diff
@@ -0,0 +1,14 @@
+locals {
+	version = "v0.6.0"
+}
+
+unit "app1" {
+	source = "github.com/gruntwork-io/terraform-google-sql.git//modules/cloud-sql?ref=${local.version}"
```
yhakbar, line 6: Why are we using this for our tests? Also, this isn't a unit but a module, so this is a little confusing. 

If you need to be able to reference a remote fixture in the Terragrunt repo that isn't tagged yet, you can use the commit. If the issue is squashed commits from merges, we can update the commit after the squash or merge in the fixture as a separate PR.

yhakbar (approved):
🚀

PR #3649, opened 2024-12-11
Don't require HOME if TERRAGRUNT_PROVIDER_CACHE_DIR is set

g7r:
## Description

I run Terragrunt in a homeless environment. I explicitly specify `TERRAGRUNT_PROVIDER_CACHE_DIR` but Terragrunt unconditionally checks whether `HOME` or `XDG_CACHE_HOME` exists despite that the result isn't used.

### Updated
Having no `HOME` doesn't trigger error anymore if `TERRAGRUNT_PROVIDER_CACHE_DIR` is set.

### DIFF
--- a/cli/provider_cache.go
+++ b/cli/provider_cache.go
@@ -62,17 +62,18 @@ type ProviderCache struct {
 }
 
 func InitProviderCacheServer(opts *options.TerragruntOptions) (*ProviderCache, error) {
-	cacheDir, err := util.GetCacheDir()
-	if err != nil {
-		return nil, err
-	}
-
 	// ProviderCacheDir has the same file structure as terraform plugin_cache_dir.
 	// https://developer.hashicorp.com/terraform/cli/config/config-file#provider-plugin-cache
 	if opts.ProviderCacheDir == "" {
+		cacheDir, err := util.GetCacheDir()
+		if err != nil {
+			return nil, err
+		}
+
 		opts.ProviderCacheDir = filepath.Join(cacheDir, "providers")
 	}
 
+	var err error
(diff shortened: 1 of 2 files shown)


cli/provider_cache.go
```diff
@@ -62,17 +62,18 @@ type ProviderCache struct {
...
 	if opts.ProviderCacheDir == "" {
+		cacheDir, err := util.GetCacheDir()
+		if err != nil {
+			return nil, err
+		}
+
 		opts.ProviderCacheDir = filepath.Join(cacheDir, "providers")
 	}
 
+	var err error
```
yhakbar, line 76: Why did you explicitly set this var err? Can't it be assigned as part of the if statement?

```go
if opts.ProviderCacheDir, err := filepath.Abs(opts.ProviderCacheDir); err != nil {
```

g7r: Because of `opts.ProviderCacheDir` I can't use `:=` here. See https://go.dev/play/p/jLjNDukKcyr

yhakbar: Oh gotcha.

yhakbar:
Sorry, but I need one more thing before we merge this, @g7r . Do you mind adding a test for a run of the provider cache server in a homeless environment? I imagine it can be done by doing something like this:

```go
t.Setenv("HOME", "")
os.Unsetenv("HOME")
```

In the test. I'd like a tests that acts as verification the provider cache works in homeless environments.

Feel free to push back if you don't think that makes sense.

g7r:
> Sorry, but I need one more thing before we merge this, @g7r . Do you mind adding a test for a run of the provider cache server in a homeless environment?

No need to apologize for requesting tests! 😄 
I've added them. Let me know if there's anything else you'd like!

yhakbar (approved):

PR #3653, opened 2024-12-12
Add redirect for unused contact page.

panzacoder:
Changes:

## Description
This seems to be an orphaned contact page that needs to be removed and redirected for the time being to ensure we don't lose leads.

### Changes
- Remove content of /contact
- Add a redirect to https://gruntwork.io/contact

## Release Notes (draft)

Removed `/contact`, replaced with redirect to `gruntwork.io/contact`

### DIFF
--- a/docs/_pages/contact/index.html
+++ b/docs/_pages/contact/index.html
@@ -1,34 +1,5 @@
 ---
-layout: contact
-title: contact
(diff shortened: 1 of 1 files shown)


yhakbar (approved):
Thanks!

PR #3672, opened 2024-12-16
fix: Reintroduce symlink work behind experiment flag

### LINKED ISSUE #3654: Revert Recent Symlink Work and Reintroduce Behind Experiment Flag
## Describe the enhancement

The recent effort to add support for symlink support in Terragrunt:
- https://github.com/gruntwork-io/terragrunt/pull/3562

Has resulted in some unexpected regression for users:
- https://github.com/gruntwork-io/terragrunt/issues/3622
- https://github.com/gruntwork-io/terragrunt/issues/3616

To avoid issues like this for users while we resolve edge cases and improve testing, revert the recent feature, and reintroduce it behind an experiment flag. This will give users that want to continue using the current unstable behavior a chance to adopt it, while work continues to handle edge cases.

In addition to doing this, add a standard process for managing experimental features like this.

Do something very similar to what is available for Strict Controls, but instead of allowing opt-in breakage of future breaking functionality, add a system for opt-in of future stable functionality that is still unstable.

## Additional context

We need to develop a systematic approach to handling experiments anyways, and this is a good test case for the new system.

This will end up being a medium effort endeavor, with documentation requirements, introducing new flags (experiment and experiment mode), and a breaking change to make symlinks no longer work by default.

- [x] Changes required.
- [x] Implications of the feature.
- [x] Alternatives considered.
- [x] Level of effort.

## PoC (Proof of Concept)

- [ ] [PoC]()

## RFC Not Needed

yhakbar:
## Description

Closes #3654.

We have other experiments like the following that need to be pulled into this system if we don't have any problems with the design:

- [Provider Cache Server](https://terragrunt.gruntwork.io/docs/features/provider-cache-server/).
- [IaC Engines](https://terragrunt.gruntwork.io/docs/features/engine/).

To ensure backwards compatibility, when they get pulled in, the existing flags that they use will be added to a strict control that warns users they are using a legacy experiment control mechanism, and should switch to using this new system.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added new `experiment` system. This system allows you to enable experimental features of Terragrunt before they're stable.
Removed Symlink support by default. Users have to opt-in to use it via the new `experiment` system.

### Migration Guide

Users using the current ability to find symlinked units in their stacks will now need to explicitly enable the `symlinks` experiment to continue using this feature.

e.g.

```bash
terragrunt run-all plan --experiment symlinks
```

For information on how to use the new `experiment` system, see the [Experiments documentation](https://terragrunt.gruntwork.io/docs/reference/experiments/) (which will go live once this PR is merged).

### DIFF
--- a/cli/commands/catalog/action.go
+++ b/cli/commands/catalog/action.go
@@ -10,6 +10,7 @@ import (
 	"github.com/gruntwork-io/terragrunt/cli/commands/catalog/tui"
 	"github.com/gruntwork-io/terragrunt/config"
 	"github.com/gruntwork-io/terragrunt/internal/errors"
+	"github.com/gruntwork-io/terragrunt/internal/experiment"
 	"github.com/gruntwork-io/terragrunt/options"
 	"github.com/gruntwork-io/terragrunt/util"
 )
@@ -39,7 +40,10 @@ func Run(ctx context.Context, opts *options.TerragruntOptions, repoURL string) e
 	for _, repoURL := range repoURLs {
 		tempDir := filepath.Join(os.TempDir(), fmt.Sprintf(tempDirFormat, util.EncodeBase64Sha1(repoURL)))
 
-		repo, err := module.NewRepo(ctx, opts.Logger, repoURL, tempDir)
+		experiment := opts.Experiments[experiment.Symlinks]
+		walkWithSymlinks := experiment.Evaluate(opts.ExperimentMode)
+
+		repo, err := module.NewRepo(ctx, opts.Logger, repoURL, tempDir, walkWithSymlinks)
 		if err != nil {
 			return err
 		}
--- a/cli/commands/catalog/action.go
+++ b/cli/commands/catalog/action.go
@@ -37,11 +37,11 @@ func Run(ctx context.Context, opts *options.TerragruntOptions, repoURL string) e
 
 	var modules module.Modules
 
+	experiment := opts.Experiments[experiment.Symlinks]
+	walkWithSymlinks := experiment.Evaluate(opts.ExperimentMode)
+
 	for _, repoURL := range repoURLs {
 		tempDir := filepath.Join(os.TempDir(), fmt.Sprintf(tempDirFormat, util.EncodeBase64Sha1(repoURL)))
-
-		experiment := opts.Experiments[experiment.Symlinks]
-		walkWithSymlinks := experiment.Evaluate(opts.ExperimentMode)
 
 		repo, err := module.NewRepo(ctx, opts.Logger, repoURL, tempDir, walkWithSymlinks)
 		if err != nil {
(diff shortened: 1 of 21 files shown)


cli/commands/catalog/action.go
```diff
@@ -39,7 +40,10 @@ func Run(ctx context.Context, opts *options.TerragruntOptions, repoURL string) e
 	for _, repoURL := range repoURLs {
 		tempDir := filepath.Join(os.TempDir(), fmt.Sprintf(tempDirFormat, util.EncodeBase64Sha1(repoURL)))
 
-		repo, err := module.NewRepo(ctx, opts.Logger, repoURL, tempDir)
+		experiment := opts.Experiments[experiment.Symlinks]
+		walkWithSymlinks := experiment.Evaluate(opts.ExperimentMode)
```
levkohimins, line 44: Since this code is the same for all `repoURLs`, it might be better to move it out of the `for` block.

internal/experiment/experiment.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+
+		experiment.Enabled = true
+		(*e)[name] = experiment
+	}
+
+	if len(invalidExperiments) > 0 {
+		return InvalidExperimentsError{
+			ExperimentNames: invalidExperiments,
+		}
```
levkohimins, line 106:
```suggestion
		return errors.New(InvalidExperimentsError{
			ExperimentNames: invalidExperiments,
		})
```

to have a stack trace.

options/options.go
```diff
@@ -364,6 +365,12 @@ type TerragruntOptions struct {
 	// StrictControls is a slice of strict controls enabled.
 	StrictControls []string
 
+	// ExperimentMode is a flag to enable experiment mode for terragrunt.
+	ExperimentMode bool
```
denis256, line 369: I think should be updated `Clone()` function too, otherwise can be issues in `run-all` cases

yhakbar: We should add a test for this at some point. I always forget to do this.

terraform/source.go
```diff
@@ -146,7 +179,7 @@ func (src Source) WriteVersionFile() error {
 //  1. Always download source URLs pointing to local file paths.
 //  2. Only download source URLs pointing to remote paths if /T/W/H doesn't already exist or, if it does exist, if the
 //     version number in /T/W/H/.terragrunt-source-version doesn't match the current version.
-func NewSource(source string, downloadDir string, workingDir string, logger log.Logger) (*Source, error) {
+func NewSource(source string, downloadDir string, workingDir string, logger log.Logger, walkWithSymlinks bool) (*Source, error) {
```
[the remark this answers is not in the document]
yhakbar: If you don't mind, I'd like to punt on that. I think we should eventually just switch to options from a variadic argument.

denis256 (approved):

levkohimins (approved):

jasonwbarnett:
Thanks so much for addressing this so quickly and thoughtfully! I really appreciate the effort to put this behind a feature flag—it’s a great solution that balances flexibility and safety, even though it’s more work on your end. Your responsiveness and attention to detail are what make contributing to this project such a great experience. 🚀

PR #3682, opened 2024-12-18
fix: Fixing `root-file-name` default

yhakbar:
## Description

I missed that the `Action` for a flag wouldn't fire when not set, so I accidentally made it so that `scaffold` and `catalog` had no defaults.

This fixes that, and makes it so that they are set to the expected defaults when no flag is set.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Updated `scaffold`.
Updated `catalog`.

### DIFF
--- a/cli/commands/flags.go
+++ b/cli/commands/flags.go
@@ -707,40 +707,36 @@ func NewVersionFlag(opts *options.TerragruntOptions) cli.Flag {
 // Scaffold/Catalog shared flags
 
 const (
+	LegacyParentConfigName      = "terragrunt.hcl"
 	RecommendedParentConfigName = "root.hcl"
 
 	RootFileNameFlagName  = "root-file-name"
 	NoIncludeRootFlagName = "no-include-root"
 )
 
+func GetDefaultRootFileName(opts *options.TerragruntOptions) string {
+	if control, ok := strict.GetStrictControl(strict.RootTerragruntHCL); ok {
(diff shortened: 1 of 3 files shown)


cli/commands/flags.go
```diff
@@ -707,40 +707,36 @@ func NewVersionFlag(opts *options.TerragruntOptions) cli.Flag {
 // Scaffold/Catalog shared flags
 
 const (
+	LegacyParentConfigName      = "terragrunt.hcl"
```
levkohimins, line 710: There is same constant in `config.DefaultTerragruntConfigPath`

cli/commands/flags.go
```diff
@@ -707,40 +707,36 @@ func NewVersionFlag(opts *options.TerragruntOptions) cli.Flag {
...
 	RecommendedParentConfigName = "root.hcl"
```
levkohimins, line 711: I think we need to keep the config names in one place.

levkohimins (changes requested):

cli/commands/flags.go
yhakbar: Importing `config.DefaultTerragruntConfigPath` would cause an import loop cycle.

They also have different semantic meaning, even though they have the same value. I'm not referring to the default unit configuration path, "terragrunt.hcl", I'm referring to the legacy root terragrunt configuration name, "terragrunt.hcl".

levkohimins (approved):

PR #3692, opened 2024-12-20
Enhanced clarity of error messages for missing mock outputs

### LINKED ISSUE #3567: Missing Dependency Outputs Has Insufficiently Helpful Error Message
## Describe the bug

When a unit depends on outputs from another unit with no dependencies, it throws an error that looks like the following:

```
08:57:09.936 ERROR  error occurred:

* ./foo/terragrunt.hcl is a dependency of ./bar/terragrunt.hcl but detected no outputs. Either the target module has not been applied yet, or the module has no outputs. If this is expected, set the skip_outputs flag to true on the dependency block.

08:57:09.936 ERROR  Unable to determine underlying exit code, so Terragrunt will exit with error code 1
```

While the error does accurately describe the problem, it doesn't offer the best prescriptive guidance on how to fix the issue.

## Steps To Reproduce

Create the following fixture:

```bash
mkdir foo
mkdir bar
touch foo/main.tf
touch bar/main.tf
touch foo/terragrunt.hcl
touch bar/terragrunt.hcl
```

Then populate `bar/terragrunt.hcl` with the following:

```hcl
dependency "foo" {
	config_path = "../foo"
}
```

## Expected behavior

The error that Terragrunt could not determine the underlying exit code should be removed, as it's clear what the underlying error is, and there should be a hint that a user could have used mocks to avoid the issue if they encountered the error during a plan of a dependency that wasn't applied yet.

Ideally, there should actually be an HCL sample emitted that a user could copy + paste to bypass the issue.

## Nice to haves

- [x] Terminal output
- [ ] Screenshots

## Versions

- Terragrunt version: `v0.68.13`
- OpenTofu version: `v1.8.5`
- Environment details (Ubuntu 20.04, Windows 10, etc.): MacOS

## Additional context

I encountered this as an expected error while writing up a guide for new users to adopt Terragrunt. These are the most important errors to get right, and they should be as good as possible.

wakeful:
## Description

Enhanced clarity of error messages for missing mock outputs when using the dependencies block - Fixes #3567 

before:
before

after:
after

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Updated - Enhanced clarity of error messages for missing mock outputs when using the dependencies block, Fix #3567 

### Migration Guide

n/a

### DIFF
--- a/config/errors.go
+++ b/config/errors.go
@@ -226,15 +226,45 @@ func (err TerragruntOutputListEncodingError) Error() string {
 }
 
 type TerragruntOutputTargetNoOutputs struct {
+	targetName    string
+	targetPath    string
 	targetConfig  string
 	currentConfig string
 }
 
+func (err TerragruntOutputTargetNoOutputs) ExitCode() int {
+	return 1
+}
+
+func (err TerragruntOutputTargetNoOutputs) Unwrap() error {
+	return nil
+}
+
 func (err TerragruntOutputTargetNoOutputs) Error() string {
+	msg := `
+If this dependency is accessed before the outputs are ready (which can happen during the planning phase of an unapplied stack), consider using mock_outputs.
+e.g.
+
+dependency "` + err.targetName + `" {
+    config_path = "` + err.targetPath + `"
+
+    mock_outputs = {
+        ` + err.targetName + `_output = "mock-` + err.targetName + `-output"
+    }
+}
+
+For further details, refer to this resource:
+https://terragrunt.gruntwork.io/docs/features/stacks/#unapplied-dependency-and-mock-outputs
+
+If you do not require outputs from your dependency, consider using the dependencies block instead:
+https://terragrunt.gruntwork.io/docs/reference/config-blocks-and-attributes/#dependencies
+`
+
 	return fmt.Sprintf(
-		"%s is a dependency of %s but detected no outputs. Either the target module has not been applied yet, or the module has no outputs. If this is expected, set the skip_outputs flag to true on the dependency block.",
+		"%s is a dependency of %s but detected no outputs. Either the target module has not been applied yet, or the module has no outputs. If this is expected, set the skip_outputs flag to true on the dependency block.\n%s",
 		err.targetConfig,
 		err.currentConfig,
+		msg,
 	)
 }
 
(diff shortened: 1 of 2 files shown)


config/errors.go
```diff
@@ -226,15 +226,44 @@ func (err TerragruntOutputListEncodingError) Error() string {
...
+}
+
+func (err TerragruntOutputTargetNoOutputs) Unwrap() error {
+	return nil
+}
+
 func (err TerragruntOutputTargetNoOutputs) Error() string {
+	msg := `
+If this dependency is accessed before the outputs are ready (which can happen during the planning phase of an unapplied stack), consider using mock_outputs.
+e.q.
```
yhakbar, line 246:
```suggestion
e.g.
```

config/errors.go
```diff
@@ -226,15 +226,44 @@ func (err TerragruntOutputListEncodingError) Error() string {
...
 
 type TerragruntOutputTargetNoOutputs struct {
+	targetName    string
+	targetPath    string
 	targetConfig  string
 	currentConfig string
 }
 
+func (err TerragruntOutputTargetNoOutputs) ExitCode() int {
+	return 2
```
yhakbar, line 236: Why 2?

yhakbar (changes requested):
There's a typo in `e.g.` being spelled `e.q`.

This is fantastic! Kicking off CI for the change.

config/errors.go
wakeful: It's always hard to choose the correct exit code, but over the years I have typically used `2` for `Incorrect usage, such as invalid options or missing arguments`.

Happy to change it, just not sure if it should be `1`.

yhakbar: Ah, let's leave it at 1 for now. 

A concern I have is that a user might be using `-detailed-exitcode` and mistakenly think that they have a successful plan with a non-empty diff:
https://opentofu.org/docs/cli/commands/plan/#other-options

I think it's safer to leave exit codes as 1 unless we specifically want to support adding tooling that detects unique exit codes for whatever reason.

yhakbar:
@wakeful 

Getting the following linting errors in CI:

```
config/errors.go:236:9: Magic number: 2, in <return> detected (mnd)
        return 2
               ^
config/errors.go:262:2: return statements should not be cuddled if block has more than two lines (wsl)
        return fmt.Sprintf(
        ^
```

wakeful:
> Getting the following linting errors in CI:
> 
> ```
> config/errors.go:236:9: Magic number: 2, in <return> detected (mnd)
>         return 2
>                ^
> config/errors.go:262:2: return statements should not be cuddled if block has more than two lines (wsl)
>         return fmt.Sprintf(
>         ^
> ```

sorted!

yhakbar (approved):

PR #3796, opened 2025-01-20
Stacks: output

denis256:
## Description

Add stacks output command:
* output formatting
* tests
* documentation

Example:

stack-output

RFC: https://github.com/gruntwork-io/terragrunt/issues/3313

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

- **New Features**
  - Introduced a new “stack output” command that enables users to retrieve outputs from multiple stack units in various formats (default, JSON, raw).
  - Added an option to configure the desired output format for better control of output presentation.
  - Enhanced stack configuration capabilities with new units and output declarations.

- **Documentation**
  - Updated CLI documentation to explain the usage of the new output command and available format options.
  - Revised stabilization criteria for the "stacks" feature to reflect the completion of support for stack output commands.

- **Tests**
  - Expanded test coverage to ensure reliable retrieval and formatting of stack outputs, including various output formats and indexed values.


yhakbar:
@denis256 

This is fantastic!

I think this will work for most people, but I think there's one thing that is gonna end up being confusing for people:

```bash
$ terragrunt stack output --format=raw app1.output
app1.output = ./test.txt
```

It would be better if the result of that was `./test.txt`.

The use-case for this format is to make it easy to integrate with shell scripts, just like the `-raw` flag for OpenTofu.

It's OK if we can't support all cty types with this flag, but the most important types to support are the ones that can be interpreted directly as simple bash strings.

If, for example, a nested unit deep in a stack had a `url` output, a user should be able to use it like this:

```bash
curl -s "$(terragrunt stack output --format=raw services.service["service-1"].url)"
```

yhakbar (dismissed):

yhakbar (approved):

Issue #3769, opened 2025-01-15
Give run_cmd() AWS credentials from --terragrunt-iam-role


rorynolan:
## Describe the enhancement

Have `run_cmd()` in `terraform.hcl` run with credentials from `--terragrunt-iam-role "$TERRAGRUNT_ROLE_ARN"`.

## Additional context

I'm trying to use a `terragrunt.hcl` that looks like this:

```
remote_state {
  backend = "s3"
  config = {
    bucket         = "my-${get_env("AWS_ACCT_ALIAS_SHORT")}-tf-backend"
    dynamodb_table = "my-${get_env("AWS_ACCT_ALIAS_SHORT")}-tf-backend"
    key            = "state/tf.state"
    encrypt        = true
    kms_key_id     = run_cmd("python3", "get_tfstate_kms_key_arn.py")
    region         = "us-west-1"
  }
}
```

So for the different buckets `my-dev-tf-backend`, `my-prod-tf-backend` I have different KMS keys which can be detected by my `get_tfstate_kms_key_arn.py` script but 

The problem is that my python program runs and complains of not having credentials to access AWS, even though I am running terragrunt with 

```
          --terragrunt-iam-web-identity-token "$GITLAB_OIDC_TOKEN" \
          --terragrunt-iam-role "$TERRAGRUNT_ROLE_ARN"
```

I have a workaround, which involves expanding my python script to use the token and role ARN to get the credentials for itself, but part of the nice thing about terragrunt is it takes care of that sort of boilerplate (ideally).

I guess `run_cmd()` in this context isn't aware of the credentials.

## PoC (Proof of Concept)

I do not have a proof of concept.

## RFC

An RFC may be needed, I don't know.

PR #3798, opened 2025-01-21
Further document `--terragrunt-auth-provider-cmd` and redo the contents.

rorynolan:
## Description

Document the advice in https://github.com/gruntwork-io/terragrunt/issues/3769 and fix the contents (add missing sections, reorder to match the text below).

## Release Notes (draft)

Further document `--terragrunt-auth-provider-cmd`

### DIFF
--- a/docs/_docs/04_reference/02-cli-options.md
+++ b/docs/_docs/04_reference/02-cli-options.md
@@ -1090,7 +1090,7 @@ The output must be valid JSON of the following schema:
 
 This allows Terragrunt to acquire different credentials at runtime without changing any `terragrunt.hcl` configuration. You can use this flag to set arbitrary credentials for continuous integration, authentication with providers other than AWS and more.
 
-As long as the standard output of the command passed to `terragrunt-auth-provider-cmd` results in JSON matching the schema above, corresponding environment variables will be set before Terragrunt begins IAC execution for a `terragrunt.hcl` file.
+As long as the standard output of the command passed to `terragrunt-auth-provider-cmd` results in JSON matching the schema above, corresponding environment variables will be set (and/or roles assumed) before Terragrunt begins parsing an `terragrunt.hcl` file or running an OpenTofu/Terraform command.
 
 The simplest approach to leverage this flag is to write a script that fetches desired credentials, and emits them to STDOUT in the JSON format listed above:
 
--- a/docs/_docs/04_reference/02-cli-options.md
+++ b/docs/_docs/04_reference/02-cli-options.md
@@ -1121,7 +1121,7 @@ Other credential configurations will be supported in the future, but until then,
 Assume the specified IAM role ARN for running OpenTofu/Terraform or AWS commands. This is a convenient way to use Terragrunt
 and OpenTofu/Terraform with multiple AWS accounts. 
 
-When using this option, the AWS authentication happens _after_ the `terragrunt.hcl` files have been processed. Therefore, `run_cmd()` calls in `terragrunt.hcl` files which need AWS to be authenticated for them to run will fail. If you need this, use [`terragrunt-auth-provider-cmd`](#terragrunt-auth-provider-cmd) instead.
+When using this option, AWS authentication takes place right before an OpenTofu/Terraform run. This takes place after `terragrunt.hcl` files are fully parsed, so HCL functions like [`get_aws_account_id`](/docs/reference/built-in-functions/#get_aws_account_id) and [`run_cmd`](/docs/reference/built-in-functions/#run_cmd) will not run after assuming the role. If you need roles to be assumed prior to parsing Terragrunt configurations, use [`terragrunt-auth-provider-cmd`](#terragrunt-auth-provider-cmd) instead.
 
 ### terragrunt-iam-assume-role-duration
 
--- a/docs/_docs/04_reference/02-cli-options.md
+++ b/docs/_docs/04_reference/02-cli-options.md
@@ -1100,7 +1100,7 @@ The simplest approach to leverage this flag is to write a script that fetches de
 echo -n '{"envs": {"KEY": "a secret"}}'
 ```
 
-You can use any technology you'd like, however, as long as Terragrunt can execute it. The expected pattern for using this flag is to populate the values dynamically using a secret store, etc.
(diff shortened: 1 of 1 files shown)


docs/_docs/04_reference/02-cli-options.md
```diff
@@ -1068,16 +1057,71 @@ When passed in, delete the contents of the temporary folder before downloading O
...
+  },
+  "envs": {
+    "ANY_KEY": ""
+  }
+}
+```
+
+This allows Terragrunt to acquire different credentials at runtime without changing any `terragrunt.hcl` configuration. You can use this flag to set arbitrary credentials for continuous integration, authentication with providers other than AWS and more.
+
+As long as the standard output of the command passed to `terragrunt-auth-provider-cmd` results in JSON matching the schema above, corresponding environment variables will be set before Terragrunt begins IAC execution for a `terragrunt.hcl` file.
```
yhakbar, line 1093:
```suggestion
As long as the standard output of the command passed to `terragrunt-auth-provider-cmd` results in JSON matching the schema above, corresponding environment variables will be set (and/or roles assumed) before Terragrunt begins IAC execution for a `terragrunt.hcl` file.
```

docs/_docs/04_reference/02-cli-options.md
```diff
@@ -1068,16 +1057,71 @@ When passed in, delete the contents of the temporary folder before downloading O
...
+
+The simplest approach to leverage this flag is to write a script that fetches desired credentials, and emits them to STDOUT in the JSON format listed above:
+
+```bash
+#!/usr/bin/env bash
+
+echo -n '{"envs": {"KEY": "a secret"}}'
+```
+
+You can use any technology you'd like, however, as long as Terragrunt can execute it. The expected pattern for using this flag is to populate the values dynamically using a secret store, etc.
```
yhakbar, line 1103:
```suggestion
You can use any technology for the authentication provider you'd like, however, as long as Terragrunt can execute it. The expected pattern for using this flag is to author a script/program that will dynamically fetch secret values from a secret store, etc. then emit them to STDOUT for consumption by Terragrunt.
```

docs/_docs/04_reference/02-cli-options.md
```diff
@@ -35,67 +35,68 @@ This page documents the CLI commands and options available with Terragrunt:
   - [graph](#graph)
   - [stack](#stack)
 - [CLI options](#cli-options)
-  - [terragrunt-check](#terragrunt-check)
```
yhakbar, line 38: Why was this updated?

docs/_docs/04_reference/02-cli-options.md
```diff
@@ -1068,16 +1057,71 @@ When passed in, delete the contents of the temporary folder before downloading O
...
 **CLI Arg**: `--terragrunt-iam-role`<br/>
 **Environment Variable**: `TERRAGRUNT_IAM_ROLE`<br/>
 **Requires an argument**: `--terragrunt-iam-role "arn:aws:iam::ACCOUNT_ID:role/ROLE_NAME"`<br/>
 
-Assume the specified IAM role ARN before running OpenTofu/Terraform or AWS commands. This is a convenient way to use Terragrunt
-and OpenTofu/Terraform with multiple AWS accounts.
+Assume the specified IAM role ARN for running OpenTofu/Terraform or AWS commands. This is a convenient way to use Terragrunt
+and OpenTofu/Terraform with multiple AWS accounts. 
+
+When using this option, the AWS authentication happens _after_ the `terragrunt.hcl` files have been processed. Therefore, `run_cmd()` calls in `terragrunt.hcl` files which need AWS to be authenticated for them to run will fail. If you need this, use [`terragrunt-auth-provider-cmd`](#terragrunt-auth-provider-cmd) instead.
```
yhakbar, line 1124:
```suggestion
When using this option, AWS authentication takes place right before an OpenTofu/Terraform run. This takes place after `terragrunt.hcl` files are fully parsed, so HCL functions like [`get_aws_account_id`](/docs/reference/built-in-functions/#get_aws_account_id) and [`run_cmd`](/docs/reference/built-in-functions/#run_cmd) will not run after assuming the role. If you need roles to be assumed prior to parsing Terragrunt configurations, use [`terragrunt-auth-provider-cmd`](#terragrunt-auth-provider-cmd) instead.

happens _after_ the `terragrunt.hcl` files have been processed. Therefore, `run_cmd()` calls in `terragrunt.hcl` files which need AWS to be authenticated for them to run will fail. If you need this, use [`terragrunt-auth-provider-cmd`](#terragrunt-auth-provider-cmd) instead.
```

docs/_docs/04_reference/02-cli-options.md
```diff
@@ -1103,6 +1147,8 @@ Used as the session name for the STS session which assumes the role defined in `
 
 Used as the web identity token for assuming a role temporarily using the AWS Security Token Service (STS) with the [AssumeRoleWithWebIdentity](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html) API.
 
+When using this option, the AWS authentication happens _after_ the `terragrunt.hcl` files have been processed. Therefore, `run_cmd()` calls in `terragrunt.hcl` files which need AWS to be authenticated for them to run will fail. If you need this, use [`terragrunt-auth-provider-cmd`](#terragrunt-auth-provider-cmd) instead.
```
yhakbar, line 1150:
```suggestion
When using this option, AWS authentication takes place right before an OpenTofu/Terraform run. This takes place after `terragrunt.hcl` files are fully parsed, so HCL functions like [`get_aws_account_id`](/docs/reference/built-in-functions/#get_aws_account_id) and [`run_cmd`](/docs/reference/built-in-functions/#run_cmd) will not run after assuming the role. If you need roles to be assumed prior to parsing Terragrunt configurations, use [`terragrunt-auth-provider-cmd`](#terragrunt-auth-provider-cmd) instead.
```

docs/_docs/04_reference/02-cli-options.md
rorynolan: I forgot that I had done this. I tried to match the table of contents at the top with the order of stuff below. You can see that I've moved it rather than deleted it.

PR #3879, opened 2025-02-12
fix: Fixing excessive warning for dependency input usage

### LINKED ISSUE #3873: TG warns about skip-dependencies-inputs option even if not used
## Describe the bug

I don't use any `skip-dependencies-inputs` options or `dependency.abc.inputs.xyz` so I don' understand from where it comes and what to do to fix it.
```
10:32:51.747 WARN   [../dep] The `skip-dependencies-inputs` option is deprecated and will be removed in a future version of Terragrunt. Reading inputs from dependencies has been deprecated. To continue using inputs from dependencies, forward them as outputs.
```

## Steps To Reproduce

```
.
├── dep
│   ├── main.tf
│   ├── terraform.tfstate
│   └── terragrunt.hcl
└── main
    └── terragrunt.hcl
```

`dep/main.tf` and `dep/terragrunt.hcl` are empty.

`main/terragrunt.hcl`

```hcl
dependency "dep" {
  config_path = find_in_parent_folders("dep")
}
```

```
terragrunt apply --working-dir dep
terragrunt apply --working-dir main
```

## Expected behavior

No warning about `skip-dependencies-inputs` as options is not used as well as inputs are not used.

## Nice to haves

- [x] Terminal output
- [X] Screenshots

Image

## Versions

- Terragrunt version: 0.73.0
- OpenTofu/Terraform version: Terraform v1.10.5 on darwin_arm64
- Environment details (Ubuntu 20.04, Windows 10, etc.): Macos M1 15.3.1

## Additional context

Here is only minimal code. I have multiple dependencies and modules however the issue is reproducible even on that small example.

yhakbar:
## Description

Fixes #3873

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Updated strict mode to allow for error suppression.

## Summary by CodeRabbit

- **New Features**
  - Updated environment variable usage now adopts the new names (TG_TF_PATH with a default value change and TG_PROVIDER_CACHE added) to streamline integration workflows.
  - Enhanced validation and clearer error/warning messages when using deprecated dependency inputs and legacy command flags.

- **Documentation**
  - Revised dependency block docs with a new attribute for defining mock outputs merge strategy and deprecation notice for the previous attribute.

- **Tests**
  - Updated integration tests to reflect simplified command-line flags and improved error messages for deprecated functionalities.
  - Added assertions to check for deprecation warnings related to reading inputs from dependencies.


levkohimins:
There is no need to change the interface. There are two options:

1. Easy, just remove warn message:
``` go
	skipDependenciesInputsControl := &Control{
		// TODO: `ErrorFmt` and `WarnFmt` of this control are not displayed anywhere and needs to be reworked.
		Name:        SkipDependenciesInputs,
		Description: "Disable reading of dependency inputs to enhance dependency resolution performance by preventing recursively parsing Terragrunt inputs from dependencies.",
		Error:       errors.Errorf("Reading inputs from dependencies is no longer supported. To acquire values from dependencies, use outputs."),
		Category:    stageCategory,
	}
```
If `ctrl.Warning` is empty it does nothing: https://github.com/gruntwork-io/terragrunt/blob/a02986e130e46861da59d892a10f98a306d800ed/internal/strict/controls/control.go#L118

2. Flexible, creating a new struct, like `DeprecatedEnvVar`, with implementing your own `Evaluate` method
```go
type ControlWithoutWarning struct {
	*Control
}

func (ctrl *ControlWithoutWarning) Evaluate(ctx context.Context) error {

	// Do what you want.

	return nil
}
```

levkohimins (commented):
In general, we can go with this, but I would recommend simplifying the code (I explaned in the comments).

levkohimins:
The new structure approach is justified when you need to write different logic for `Evaluate(ctx context.Context)`, but in a situation where logic is the same, if only it complicates the future maintenance of the same logic in two places.

levkohimins:
Sorry if I confused you with the options initially.

config/config_partial.go
```diff
@@ -457,11 +457,17 @@ func PartialParseConfig(ctx *ParsingContext, file *hclparse.File, includeFromChi
...
 
+			skipDependenciesInputs := allControls.Find(controls.SkipDependenciesInputs)
+			if skipDependenciesInputs == nil {
+				return nil, errors.New("failed to find control " + controls.SkipDependenciesInputs)
+			}
+
+			skipDependenciesInputs.SuppressWarning()
+
+			logger := log.ContextWithLogger(ctx, ctx.TerragruntOptions.Logger)
 			if err := skipDependenciesInputs.Evaluate(logger); err != nil {
```
levkohimins, line 469:
```suggestion
			logger := log.ContextWithLogger(ctx, ctx.TerragruntOptions.Logger)
			if err := skipDependenciesInputs.SuppressWarning().Evaluate(logger); err != nil {
```

internal/strict/controls/control.go
```diff
@@ -97,6 +100,11 @@ func (ctrl *Control) AddSubcontrols(newCtrls ...strict.Control) {
 	ctrl.Subcontrols = append(ctrl.Subcontrols, newCtrls...)
 }
 
+// SuppressWarning suppresses the warning message from being displayed.
+func (ctrl *Control) SuppressWarning() {
+	ctrl.Suppress = true
+}
+
```
levkohimins, line 107: It's better not to change the current control, but to return a copy of the control with a new value. Imagine that you need to use the same control, in one place with the warning suppressed, and in another - not.

```suggestion
// SuppressWarning suppresses the warning message from being displayed.
func (ctrl *Control) SuppressWarning() *Control {
        newCtrl := *ctrl
	newCtrl.Suppress = true
	
	return &newCtrl
}

```

don't forget to extend the `Control` interface in `strict/control.go`

```go
type Control interface {
   SuppressWarning() Control
}
```

levkohimins (dismissed):
It's already good. It's up to you, apply the latest suggestion or not.

internal/strict/controls/control.go
yhakbar: We just mutate enable, so I'll avoid doing this to keep things consistent.

config/config_partial.go
yhakbar: See [this](https://github.com/gruntwork-io/terragrunt/pull/3879#discussion_r1954901866) about keeping things consistent with enable.

config/config.go
```diff
@@ -868,6 +870,20 @@ func ParseConfigString(ctx *ParsingContext, configPath string, configString stri
...
+	if detectInputsCtyUsage(file) {
+		allControls := ctx.TerragruntOptions.StrictControls
+
+		skipDependenciesInputs := allControls.Find(controls.SkipDependenciesInputs)
+		if skipDependenciesInputs == nil {
+			return nil, errors.New("failed to find control " + controls.SkipDependenciesInputs)
+		}
+
+		logger := log.ContextWithLogger(ctx, ctx.TerragruntOptions.Logger)
+		if err := skipDependenciesInputs.Evaluate(logger); err != nil {
```
levkohimins, line 882: it's a bit confusing, creates the wrong impression that `Evaluate` takes the Logger type. 

```suggestion
		evalCtx := log.ContextWithLogger(ctx, ctx.TerragruntOptions.Logger)
		if err := skipDependenciesInputs.Evaluate(evalCtx); err != nil {
```

Maybe it's even worth to move this code `if detectInputsCtyUsage(file) {....}` into a separate function. Up to you.

config/config.go
```diff
@@ -940,6 +956,40 @@ func ParseConfig(ctx *ParsingContext, file *hclparse.File, includeFromChild *Inc
...
+			if len(traversal) < dependencyInputsIdentifierMinParts {
+				continue
+			}
+
+			root, ok := traversal[0].(hcl.TraverseRoot)
+			if !ok || root.Name != "dependency" {
+				continue
+			}
+
+			attrTraversal, ok := traversal[2].(hcl.TraverseAttr)
```
levkohimins, line 981: Do you need to check if the index exists?

config/config.go
```diff
@@ -940,6 +956,40 @@ func ParseConfig(ctx *ParsingContext, file *hclparse.File, includeFromChild *Inc
...
+			if !ok || attrTraversal.Name != "inputs" {
```
levkohimins, line 982:
```suggestion
			if !ok || attrTraversal.Name != MetadataInputs {
```

config/config.go
```diff
@@ -940,6 +956,40 @@ func ParseConfig(ctx *ParsingContext, file *hclparse.File, includeFromChild *Inc
...
+	for _, attr := range body.Attributes {
+		for _, traversal := range attr.Expr.Variables() {
+			const dependencyInputsIdentifierMinParts = 3
+
+			if len(traversal) < dependencyInputsIdentifierMinParts {
+				continue
+			}
+
+			root, ok := traversal[0].(hcl.TraverseRoot)
+			if !ok || root.Name != "dependency" {
```
levkohimins, line 977:
```suggestion
			if !ok || root.Name != MetadataDependency {
```

config/config_partial.go
```diff
@@ -457,11 +457,17 @@ func PartialParseConfig(ctx *ParsingContext, file *hclparse.File, includeFromChi
...
 			allControls := ctx.TerragruntOptions.StrictControls
-			skipDependenciesInputs := allControls.FilterByNames(controls.SkipDependenciesInputs)
-			logger := log.ContextWithLogger(ctx, ctx.TerragruntOptions.Logger)
 
+			skipDependenciesInputs := allControls.Find(controls.SkipDependenciesInputs)
+			if skipDependenciesInputs == nil {
+				return nil, errors.New("failed to find control " + controls.SkipDependenciesInputs)
+			}
+
+			skipDependenciesInputs.SuppressWarning()
```
levkohimins, line 466: `skipDependenciesInputs` is a pointer. If you change it, it will also affect `opts.Strict Controls`.
It would be better not to modify `opts.StrictControls` at all. In another place where you use it, it will be unclear why the value is different from what is specified by default. This may especially surprise other developers who are not aware.

yhakbar: Ya, I would prefer that we not mutate the control, but we already do it for `Enable()`, and I think it's better to keep things consistent. We shouldn't _sometimes_ mutate in methods of pointer receivers.

We should either always do that, and have that be the expectation, or never do it.

config/config.go
yhakbar: We already know that it's a slice larger than 3 above:
https://github.com/gruntwork-io/terragrunt/pull/3879/files/2d5e986605324b8e16a38a4294c1972ac047780f#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R972

config/config_partial.go
levkohimins: In some ways you are right, but these are slightly different things :) Enable is done to configure the controls before starting work, and this is definitely something the user does explicitly. But in this case, it's like a magic. If you think it's better this way, so be it.

levkohimins (approved):

PR #3917, opened 2025-02-21
fix #2072 make sure the `generated` code is formatted before writing it to a file.

### LINKED ISSUE #2072: Make hclfmt format generated code
Sometimes we want to generate terraform code and we write configurations like:

```hcl

generate "data" {
  path      = "data_gen.tf"
  if_exists = "overwrite"
  contents  = <<EOF
data "yandex_dns_zone" "infra" {
  name = "${local.common.infra_dns_name}"
  folder_id = "${local.common.common_folder_id}"
}
}
```

And if we run `terragrunt hclfmt` and `terraform fmt -recursive` this config will not be formatted. 

However, if we run `terragrunt fmt -check=true` it will produce errors like:

```
data_gen.tf
ERRO[0003] 1 error occurred:
        * exit status 3
```

I propose to add autoformatting of generated objects by setting some special flag (`terragrunt hclfmt --format-generated`)

wakeful:
## Description

Fixes #2072 make sure the `generated` code is formatted before writing it to a file.

## TODOs

- [x] Update the docs.

## Release Notes (draft)
Updated the `generated` block to format the `HCL` syntax before writing it to a file, fix for #2072.

### Migration Guide

n/a

## Summary by CodeRabbit

- **New Features**
  - Introduced a new `hcl_fmt` attribute in the `generate` block for automatic formatting of `.hcl`, `.tf`, and `.tofu` files before writing.
  
- **Tests**
  - Added automated tests to validate the formatting of generated files, ensuring compliance with the new capabilities.

### DIFF
--- a/codegen/generate.go
+++ b/codegen/generate.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclsimple"
-
 	"github.com/hashicorp/hcl/v2/hclwrite"
 	ctyjson "github.com/zclconf/go-cty/cty/json"
 
@@ -87,6 +86,7 @@ type GenerateConfig struct {
 	Contents         string `cty:"contents"`
 	DisableSignature bool   `cty:"disable_signature"`
 	Disable          bool   `cty:"disable"`
+	HclFmt           *bool  `cty:"hcl_fmt"`
 }
 
 // WriteToFile will generate a new file at the given target path with the given contents. If a file already exists at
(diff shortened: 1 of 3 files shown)


codegen/generate.go
```diff
@@ -138,7 +138,7 @@ func WriteToFile(terragruntOptions *options.TerragruntOptions, basePath string,
 	contentsToWrite := fmt.Sprintf("%s%s", prefix, config.Contents)
 
 	const ownerWriteGlobalReadPerms = 0644
-	if err := os.WriteFile(targetPath, []byte(contentsToWrite), ownerWriteGlobalReadPerms); err != nil {
+	if err := os.WriteFile(targetPath, hclwrite.Format([]byte(contentsToWrite)), ownerWriteGlobalReadPerms); err != nil {
```
yhakbar, line 141: @wakeful , this needs to be opt-in. The `generate` block can write arbitrary content, not just HCL content.

We could have a `fmt` attribute that determines whether to format the file, and users can set that to format the file.

We can also look to leverage the filename extension to have a smarter default for formatting (e.g. if the file has extension `hcl`, `tf` or `tofu`, default to format the file if not defined.

yhakbar (changes requested):

wakeful:
hey @yhakbar,

I added an `hcl_fmt` attribute to the `generate` block and limited the formatting to specific file extensions.
Am I on the right track now?

yhakbar:
> hey @yhakbar,
> 
> I added an `hcl_fmt` attribute to the `generate` block and limited the formatting to specific file extensions. Am I on the right track now?

@wakeful 

We're very close now. What I thought would be done is to define the value of `HclFmt` as a pointer of bool, and immediately look at the value of `Path` after parsing the config, and default the value to `true` if the extension is in the list of `.hcl`, `.tf` or `.tofu`, and default the value to `false` otherwise. If a user explicitly sets the value to something, we should blindly accept what they've requested (in-case some other extension we haven't thought of uses HCL and we should format it anyways, or they would very intentionally like to avoid formatting a file, even though it's an HCL file).

wakeful:
@yhakbar I modified the code based on your suggestions.

codegen/generate_test.go
```diff
@@ -144,6 +145,81 @@ func TestRemoteStateConfigToTerraformCode(t *testing.T) {
...
+			fmt:      &bFalse,
+			path:     fmt.Sprintf("%s/%s", testDir, "ignore_hcl_fmt.hcl"),
+			contents: "variable \"msg\"{\ntype=string\n  default=\"hello\"\n}\n",
+			expected: "variable \"msg\"{\ntype=string\n  default=\"hello\"\n}\n",
+			ifExists: codegen.ExistsError,
+		},
+	}
+
+	for _, tt := range tc {
+		tt := tt
```
yhakbar, line 191: @wakeful FYI, you don't have to do this in modern golang.

codegen/generate_test.go
```diff
@@ -144,6 +145,81 @@ func TestRemoteStateConfigToTerraformCode(t *testing.T) {
...
+			}
+
+			opts, err := options.NewTerragruntOptionsForTest("mock-path-for-test.hcl")
+			require.NoError(t, err)
+			assert.NotNil(t, opts)
+
+			err = codegen.WriteToFile(opts, "", config)
+			require.NoError(t, err)
+
+			assert.True(t, util.FileExists(tt.path))
```
yhakbar, line 213: FYI, you can use this: https://pkg.go.dev/github.com/stretchr/testify/assert#Assertions.FileExists

yhakbar (approved):

PR #3926, opened 2025-02-24
[WIP] feat: Support version constraints for Terraform Registry modules

### LINKED ISSUE #1930: Support Version Constraint Syntax for modules loaded from a Terraform registry
Support Version Constraint Syntax for modules loaded from a Terraform registry:
- https://www.terraform.io/docs/language/expressions/version-constraints.html#version-constraint-syntax

At the moment it is not possible to use the version constrains to load a module from the terraform registry
The current configuration looks like:

```
terraform {
  source = "tfr://myregistry.io/mymodule/azurerm?version=0.0.1"
```
However I would like to be able to use any build version, in Terraform you would configure it like:

```
module "mymodule" {
  source = "myregistry.io/mymodule/azurerm"
  version = "~> 0.0.1"
}
```

Version:
```
Terraform v1.0.10
Terragrunt version v0.35.12

```

Other resources:
- https://www.terraform.io/docs/internals/module-registry-protocol.html

### LINKED ISSUE #2572: Feature request: support latest version when using `tfr` protocol in `terraform.source`

```hcl
terraform {
}
```

```hcl
module "vpc" {
}
```

There are a few options here:

**Additional context**
N/A

juan-vg:
## Description

Fixes #1930
Fixes #2572 

## TODOs

- [x] Update the docs.
- [X] Ensure any 3rd party code adheres with our [license policy](https://www.notion.so/gruntwork/Gruntwork-licenses-and-open-source-usage-policy-f7dece1f780341c7b69c1763f22b1378) or delete this line if its not applicable.
- [X] Include release notes. If this PR is backward incompatible, include a migration guide.

## Release Notes (draft)

Added support for version constraints for Terraform Registry modules

## Summary by CodeRabbit

- **New Features**
  - Enhanced module version management with clearer error messages for various failure scenarios.
  - Improved module version retrieval process, ensuring accurate version selection and increased logging transparency during downloads.
  - New configuration file for Terragrunt to retrieve modules from the public Terraform registry.
  - Added a new test for Terraform registry version resolution to ensure correct functionality.

- **Documentation**
  - Updated documentation for the `VERSION` argument in the `terraform` block to clarify acceptable formats.

- **Tests**
  - Expanded test coverage for the `GetTargetVersion` function to handle various input scenarios.

### DIFF
--- a/docs/_docs/04_reference/04-config-blocks-and-attributes.md
+++ b/docs/_docs/04_reference/04-config-blocks-and-attributes.md
@@ -97,7 +97,7 @@ The `terraform` block supports the following arguments:
     registry]({{site.baseurl}}/docs/getting-started/quick-start#a-note-about-using-modules-from-the-registry) for more
     information about using modules from the Terraform Registry with Terragrunt.
   - VERSION can be a specific version number (e.g., `1.2.3`) or a version constraint (e.g., `>= 1.2.3, <=3.2.1`) following syntax
-    defined in the [Terraform documentation](https://developer.hashicorp.com/terraform/language/expressions/version-constraints).
+    defined in the [OpenTofu documentation](https://opentofu.org/docs/language/expressions/version-constraints/).
 
 - `include_in_copy` (attribute): A list of glob patterns (e.g., `["*.txt"]`) that should always be copied into the
   OpenTofu/Terraform working directory. When you use the `source` param in your Terragrunt config and run `terragrunt <command>`,
--- a/docs/_docs/04_reference/04-config-blocks-and-attributes.md
+++ b/docs/_docs/04_reference/04-config-blocks-and-attributes.md
@@ -96,6 +96,8 @@ The `terraform` block supports the following arguments:
   - Refer to [A note about using modules from the
     registry]({{site.baseurl}}/docs/getting-started/quick-start#a-note-about-using-modules-from-the-registry) for more
     information about using modules from the Terraform Registry with Terragrunt.
+  - VERSION can be a specific version number (e.g., `1.2.3`) or a version constraint (e.g., `>= 1.2.3, <=3.2.1`) following syntax
+    defined in the [Terraform documentation](https://developer.hashicorp.com/terraform/language/expressions/version-constraints).
 
 - `include_in_copy` (attribute): A list of glob patterns (e.g., `["*.txt"]`) that should always be copied into the
   OpenTofu/Terraform working directory. When you use the `source` param in your Terragrunt config and run `terragrunt <command>`,
--- a/test/integration_registry_test.go
+++ b/test/integration_registry_test.go
@@ -17,6 +17,7 @@ const (
 	registryFixtureRootShorthandModulePath       = "root-shorthand"
 	registryFixtureSubdirModulePath              = "subdir"
 	registryFixtureSubdirWithReferenceModulePath = "subdir-with-reference"
+	registryFixtureVersion                       = "version"
 )
 
 func TestTerraformRegistryFetchingRootModule(t *testing.T) {
@@ -55,5 +56,15 @@ func testTerraformRegistryFetching(t *testing.T, modPath, expectedOutputKey stri
 	require.NoError(t, json.Unmarshal(stdout.Bytes(), &outputs))
 	_, hasOutput := outputs[expectedOutputKey]
 	assert.True(t, hasOutput)
+}
 
+// test that the version of the module is correctly resolved and downloaded when running a terragrunt init
+func TestTerraformRegistryVersionResolution(t *testing.T) {
+	t.Parallel()
+
+	versionFixture := util.JoinPath(registryFixturePath, registryFixtureVersion)
+	helpers.CleanupTerragruntFolder(t, versionFixture)
+
+	_, _, err := helpers.RunTerragruntCommandWithOutput(t, "terragrunt init --working-dir "+versionFixture)
+	require.NoError(t, err)
 }
--- a/test/integration_registry_test.go
+++ b/test/integration_registry_test.go
@@ -58,7 +58,6 @@ func testTerraformRegistryFetching(t *testing.T, modPath, expectedOutputKey stri
 	assert.True(t, hasOutput)
 }
 
-// test that the version of the module is correctly resolved and downloaded when running a terragrunt init
 func TestTerraformRegistryVersionResolution(t *testing.T) {
 	t.Parallel()
 
--- a/tf/getter.go
+++ b/tf/getter.go
@@ -198,7 +198,7 @@ func (tfrGetter *RegistryGetter) Get(dstPath string, srcURL *url.URL) error {
 }
 
 // getHighestVersion returns the highest version from the list of versions, or an empty string if the list is empty.
-func getHighestVersion(logger log.Logger, availableVersions []*version.Version) string {
+func getHighestVersion(availableVersions []*version.Version) string {
 	if len(availableVersions) == 0 {
 		return ""
 	}
@@ -216,12 +216,14 @@ func GetTargetVersion(ctx context.Context, logger log.Logger, registryDomain str
 	if err != nil {
 		return "", errors.New(err)
 	}
-	body, _, err := httpGETAndGetResponse(ctx, logger, *moduleVersionsURL)
+	bodyData, _, err := httpGETAndGetResponse(ctx, logger, *moduleVersionsURL)
 	if err != nil {
 		return "", errors.New(ModuleVersionsErr{moduleName: modulePath})
 	}
 	var responseJSON Modules
-	json.Unmarshal(body, &responseJSON)
+	if err := json.Unmarshal(bodyData, &responseJSON); err != nil {
+		return "", errors.New(ModuleVersionsErr{moduleName: modulePath})
+	}
 	availableVersions := responseJSON.Modules[0].ModuleVersions
 
 	// Filter the available versions based on the version constraint
@@ -241,9 +243,9 @@ func GetTargetVersion(ctx context.Context, logger log.Logger, registryDomain str
 	}
 
 	// Get the highest version from the filtered versions
-	targetVersion := getHighestVersion(logger, filteredVersions)
+	targetVersion := getHighestVersion(filteredVersions)
 	if targetVersion == "" {
-		return "", errors.New(ModuleVersionConstraintErr{versionConstraint: targetVersion})
+		return "", errors.New(ModuleVersionConstraintErr{versionConstraint: versionQuery})
 	}
 	return targetVersion, nil
 }
--- a/tf/getter.go
+++ b/tf/getter.go
@@ -224,6 +224,9 @@ func GetTargetVersion(ctx context.Context, logger log.Logger, registryDomain str
 	if err := json.Unmarshal(bodyData, &responseJSON); err != nil {
 		return "", errors.New(ModuleVersionsErr{moduleName: modulePath})
 	}
+	if len(responseJSON.Modules) == 0 {
+		return "", errors.New(ModuleVersionsErr{moduleName: modulePath})
+	}
 	availableVersions := responseJSON.Modules[0].ModuleVersions
 
 	// Filter the available versions based on the version constraint
--- a/tf/getter.go
+++ b/tf/getter.go
@@ -159,7 +159,12 @@ func (tfrGetter *RegistryGetter) Get(dstPath string, srcURL *url.URL) error {
 		return err
 	}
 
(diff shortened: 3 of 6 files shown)


docs/_docs/04_reference/04-config-blocks-and-attributes.md
```diff
@@ -96,6 +96,8 @@ The `terraform` block supports the following arguments:
   - Refer to [A note about using modules from the
     registry]({{site.baseurl}}/docs/getting-started/quick-start#a-note-about-using-modules-from-the-registry) for more
     information about using modules from the Terraform Registry with Terragrunt.
+  - VERSION can be a specific version number (e.g., `1.2.3`) or a version constraint (e.g., `>= 1.2.3, <=3.2.1`) following syntax
+    defined in the [Terraform documentation](https://developer.hashicorp.com/terraform/language/expressions/version-constraints).
```
yhakbar, line 100:
```suggestion
    defined in the [OpenTofu documentation](https://opentofu.org/docs/language/expressions/version-constraints/).
```

We always attempt to default to OpenTofu documentation. Please make sure there isn't anything missed in the rest of the pull request that prefers Terraform documentation over OpenTofu.

yhakbar (changes requested):
@juan-vg 

Our linter found some issues in CI:

```
tf/getter.go:251:3: if statements should only be cuddled with assignments (wsl)
                if versionConstraint.Check(availableVersionParsed) {
                ^
tf/getter.go:231:2: declarations should never be cuddled (wsl)
        var responseJSON Modules
        ^
tf/getter.go:235:2: if statements should only be cuddled with assignments (wsl)
        if len(responseJSON.Modules) == 0 || len(responseJSON.Modules[0].ModuleVersions) == 0 {
        ^
tf/getter.go:243:2: only one cuddle assignment allowed before if statement (wsl)
        if err != nil {
        ^
tf/getter.go:206:2: return statements should not be cuddled if block has more than two lines (wsl)
        return availableVersions[len(availableVersions)-1].String()
        ^
tf/getter.go:227:2: assignments should only be cuddled with other assignments (wsl)
        bodyData, _, err := httpGETAndGetResponse(ctx, logger, *moduleVersionsURL)
        ^
tf/getter.go:232:2: only one cuddle assignment allowed before if statement (wsl)
        if err := json.Unmarshal(bodyData, &responseJSON); err != nil {
        ^
tf/getter.go:171:2: expressions should not be cuddled with blocks (wsl)
        tfrGetter.TerragruntOptions.Logger.Infof("Downloading module from %s", moduleURL.String())
        ^
tf/getter.go:246:2: only one cuddle assignment allowed before range statement (wsl)
        for _, availableVersion := range availableVersions {
        ^
tf/getter.go:261:2: return statements should not be cuddled if block has more than two lines (wsl)
        return targetVersion, nil
        ^
tf/getter.go:238:2: assignments should only be cuddled with other assignments (wsl)
        availableVersions := responseJSON.Modules[0].ModuleVersions
        ^
tf/getter.go:205:2: expressions should not be cuddled with blocks (wsl)
        sort.Sort(version.Collection(availableVersions))
        ^
tf/getter.go:224:2: only one cuddle assignment allowed before if statement (wsl)
        if err != nil {
        ^
tf/getter.go:228:2: only one cuddle assignment allowed before if statement (wsl)
        if err != nil {
        ^
INFO File cache stats: 403 entries of total size 2.3MiB 
INFO Memory: 2474 samples, avg is 2695.3MB, max is 3747.3MB 
INFO Execution took 4m11.814409778s               
make: *** [Makefile:45: run-lint] Error 1

Exited with code exit status 2

```

Please request re-review when you've addressed these issues.

yhakbar:
Thanks for addressing those, @juan-vg .

Upon re-review, I realized that the implementation you've added uses a query string parameter for the source URL:

```hcl
locals {
  base_source = "tfr:///terraform-aws-modules/iam/aws"
  version     = "~>5.0, <5.51.0, !=5.50.0"
}
terraform {
  source = "${local.base_source}?version=${local.version}"
}
```

Instead of a new `version` attribute:

```hcl
terraform {
  source  = "tfr:///terraform-aws-modules/iam/aws"
  version = "~>5.0, <5.51.0, !=5.50.0"
}
```

I believe a new `version` attribute would be more in-line with the UI/UX of OpenTofu/Terraform, so users may find it more convenient to follow that pattern as well. What do you think? Would you be willing to try out an implementation that uses a `version` attribute?

tf/getter.go
```diff
@@ -135,18 +152,30 @@ func (tfrGetter *RegistryGetter) Get(dstPath string, srcURL *url.URL) error {
...
+	if err != nil {
+		return err
+	}
+
+	moduleURL, err := BuildRequestURL(registryDomain, moduleRegistryBasePath, modulePath, targetVersion)
 	if err != nil {
 		return err
 	}
 
+	tfrGetter.TerragruntOptions.Logger.Infof("Downloading module from %s", moduleURL.String())
```
denis256, line 177: I'm not entirely sure we need this log message at the Info level. It may produce multiple unused lines, so would it make more sense to set it to Debug instead?

tf/getter.go
```diff
@@ -268,6 +297,39 @@ func GetModuleRegistryURLBasePath(ctx context.Context, logger log.Logger, domain
 	return respJSON.ModulesPath, nil
 }
 
+// GetTargetVersion retrieves the target version of the module based on the version constraint provided. This function
+// will return the highest version that satisfies the version constraint. If no version satisfies the constraint, an
+// error will be returned.
+func GetTargetVersion(ctx context.Context, logger log.Logger, url url.URL, versionQuery string) (string, error) {
```
denis256, line 303: It appears that this function is only used within the tf package. Would it make sense to mark it as private?

tf/getter.go
```diff
@@ -268,6 +297,39 @@ func GetModuleRegistryURLBasePath(ctx context.Context, logger log.Logger, domain
...
+		return "", errors.New(ModuleVersionsFetchErr{sourceURL: url.String()})
+	}
+
+	var responseJSON Modules
+	if err := json.Unmarshal(body, &responseJSON); err != nil {
+		return "", errors.New(ModuleVersionsFetchErr{sourceURL: url.String()})
+	}
+
+	if len(responseJSON.Modules) == 0 || len(responseJSON.Modules[0].ModuleVersions) == 0 {
+		return "", errors.New(ModuleVersionsFetchErr{sourceURL: url.String()})
```
denis256, line 315: It appears that we're reusing the same ModuleVersionsFetchErr error for multiple scenarios. This could be confusing to users, as the same error is thrown when the GET request fails, when the JSON response cannot be parsed, or when no module versions are found.

tf/getter.go
```diff
@@ -380,3 +462,38 @@ func BuildRequestURL(registryDomain string, moduleRegistryBasePath string, modul
...
+
+	sort.Sort(version.Collection(availableVersions))
+
+	return availableVersions[len(availableVersions)-1].String()
+}
+
+// getCompatibleVersions returns the list of versions that satisfy the version constraint.
+func getCompatibleVersions(availableVersions []ModuleVersion, versionQuery string) ([]*version.Version, error) {
+	// Filter the available versions based on the version constraint
+	filteredVersions := []*version.Version{}
```
denis256, line 480: var filteredVersions []*version.Version
or even pre-allocate
filteredVersions := make([]*version.Version, 0, len(availableVersions))

juan-vg:
@yhakbar 
>I believe a new version attribute would be more in-line with the UI/UX of OpenTofu/Terraform, so users may find it more convenient to follow that pattern as well. What do you think? Would you be willing to try out an implementation that uses a version attribute?

While I agree the new version attribute completely makes sense (and I'm willing to try that implementation), I'm wondering if it belongs to the scope of this PR since the version attribute is already there via query string. I mean, adding that attribute implies a (probably full) refactor of this code and at the end the target is to add the new attribute and not to support version constraints (even if they're related). From my POV the version attribute should be added as an incremental, later change after supporting version constraints. However, that's my take. WDYT?

tf/getter.go
juan-vg: I tried to make it private, but then I get an error on the getter_test :/

tf/getter.go
juan-vg: I have doubts also. However, since there's another Info log related to this and it includes the query string, I considered useful to have now this second log line clarifying which version is finally being downloaded without having to go some levels up (way far from getter.go) to change that behaviour.

As an example:
image

That said, I believe this is an intermediate step. The next one should be to implement the version attribute and properly handle the version constraints at an upper level (adding the ability to invalidate tg cache when there's a new version even if the code hasn't changed, among other changes). Then, both log lines should be addressed and merged into an more accurate one. WDYT?

test/integration_registry_test.go
```diff
@@ -55,5 +56,14 @@ func testTerraformRegistryFetching(t *testing.T, modPath, expectedOutputKey stri
 	require.NoError(t, json.Unmarshal(stdout.Bytes(), &outputs))
 	_, hasOutput := outputs[expectedOutputKey]
 	assert.True(t, hasOutput)
+}
+
+func TestTerraformRegistryVersionResolution(t *testing.T) {
```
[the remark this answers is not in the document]
juan-vg: Sure, I'm completely open to it. Do you have any idea or clarification about the path I should follow? In fact I tried to make the test more accurate by reading the stdout and trying to read the info log-line where the version picked was shown (to check the picked version), but I discovered logs are at a different level and this was not possible. Any help here is very welcome!

tf/getter.go
juan-vg: Since filteredVersions is a subset of availableVersions and the length can't be known in advance, I'll go for the first option. Thank you!

tf/getter.go
juan-vg: While I agree with you, many of those errors are a consequence of the same problem (network issue or lacking proper authentication). As an example, both JSON response parsing error and no ModuleVersions in the response are a consequence of authentication lack OR module not existing, and you can't know in advance. As an example, [this is a non existing org/module](https://app.spacelift.io/registry/modules/v1/test-org/test-module/aws/versions) but there's still a partial answer, which is exactly the same as when the org/module exists but the auth is missing or wrong. On the other hand, [TFE seems to better implement this for their private registry](https://app.terraform.io/api/registry/v1/modules/test-org/test-module/aws/versions): at least they return an error. However, since the answers differ from one implementation to another, I consider it's useful to condense these errors into a more abstract one (FetchErr).

What I could do for sure is to create separate errors for the GET request and these other tricky cases, and try to refine as much as possible. Let me know if you have any ideas here, please.

yhakbar:
> While I agree the new version attribute completely makes sense (and I'm willing to try that implementation), I'm wondering if it belongs to the scope of this PR since the version attribute is already there via query string. ... However, that's my take. WDYT?

@juan-vg 

Normally, I'd be totally onboard with an incremental improvement rather than hold up a PR, but in the case of TG configuration, we want to be more conservative. This introduces configuration for a URL that is not URL encoded, and when we support the attribute, users should switch off using the query string parameter, which would be a breaking change.

If you really don't have time to rework with PR to use an attribute, I'd be open to reviewing it after the changes have been gated behind an experiment flag. That way, we can document the feature as incomplete, and require user opt in to use behavior that might break in the future.

juan-vg:
@yhakbar 

>Normally, I'd be totally onboard with an incremental improvement rather than hold up a PR, but in the case of TG configuration, we want to be more conservative. This introduces configuration for a URL that is not URL encoded, and when we support the attribute, users should switch off using the query string parameter, which would be a breaking change.

I understand

>If you really don't have time to rework with PR to use an attribute, I'd be open to reviewing it after the changes have been gated behind an experiment flag. That way, we can document the feature as incomplete, and require user opt in to use behavior that might break in the future.

Nah, let's do it! I have some time and much willing, and any guiding or help will be appreciated :)

tf/getter.go
juan-vg: FYI: Since the scope of the PR has been changed to also support a new version attribute, and this will imply modifying way more than the tf getter, I believe this logs issue could be addressed also at the root, so this log line in getter won't make sense anymore

tf/getter.go
juan-vg: FYI: Since the scope of the PR has been changed to also support a new version attribute, I believe I'll use this function from other places. Let's wait until that refactor is done to evaluate again.

stephenbawks:
Bumping this to see if there were any more updates.  We would really like to see this adopted.

OpsTalkJordan:
What is this PR still missing? Would love to see this implemented.

yhakbar:
This is supported through the [`version-attribute`](https://docs.terragrunt.com/reference/experiments/active/#version-attribute) experiment.

Thanks for opening this PR @juan-vg . Sorry we didn't help get it resolved sooner.

Issue #3445, opened 2024-09-27
CLI Redesign


yhakbar:
### Summary

This RFC was initially vetted in a Gruntwork internal Notion page, and increased in size and complexity until it became detailed enough that it was no longer practical to publish as a GitHub issue.

To read the RFC, navigate to the publicly published version of the Notion page here:
https://gruntwork.notion.site/PUBLIC-Terragrunt-CLI-Redesign-10ed8759a6828042ac6bc03aef4d9dfb

That page will be updated according to feedback, with comments placed on this issue to track those updates.

All feedback, conversation and updates will take place on this issue.

### Motivation

See the Notion page.

### Proposal

See the Notion page.

### Technical Details

See the Notion page.

### Press Release

See the Notion page.

### Support Level

- [ ] I have Terragrunt Enterprise Support
- [ ] I am a paying Gruntwork customer

slaughtr:
It's great to see this RFC, this would surely clean up some mess and make Terragrunt more clear to new users. That said, it's not super clear in the text how the new `run` and `exec` commands would work. I'd love to see more examples comparing common use cases in new vs old style.

For my two cents on specific changes:

* The `--` syntax is familiar to me, and probably many people. It's the same syntax that many other tools use (aws-vault and chamber being two common ones). The only issues I could see with it are how it might interact with those other tools, but as I recall from combining aws-vault and chamber in the past that might not be a problem after all?
* Removing `terragrunt` from CLI flags makes sense, though per the RFC flag collisions would be a concern. I'd also wonder whether removing `terragrunt` from the flags would really help with anything more than a reduction in characters - right now it's very obvious to users when a flag relates only to Terragrunt, and this change would remove that clarity. If I were in your shoes and erring on the side of caution, I'd probably not make this change/make this change much later in the rollout after people have had some time to get familiar with the rest of the changes.
* The approach to backend provisioning absolutely will reduce confusion for many people. Teams I support run into confusion with the current approach very often. 10/10

yhakbar:
Thanks for the feedback @slaughtr !

### `run` vs `exec`

I'll think through that feedback and add more detail to distinguish `run` vs `exec` in the RFC. It's feedback that I also got internally, and I want to address it well. 

[...]

PR #3981, opened 2025-03-06
feat: Adding `find` command

yhakbar:
## Description

Adds `find` command. Relates to #3445.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added `find` command.

### DIFF
--- a/cli/commands/find/action.go
+++ b/cli/commands/find/action.go
@@ -1,0 +1,88 @@
+package find
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/gruntwork-io/terragrunt/internal/discovery"
+	"github.com/gruntwork-io/terragrunt/internal/errors"
+	"github.com/mgutz/ansi"
+)
+
+func Run(ctx context.Context, opts *Options) error {
+	configs, err := discovery.DiscoverConfigs(opts.TerragruntOptions)
+	if err != nil {
+		return err
+	}
+
+	switch opts.Format {
+	case "text":
+		return outputText(opts, configs)
+	case "json":
+		return outputJSON(configs)
+	default:
+		return errors.New("invalid format: " + opts.Format)
+	}
+}
+
+type JSONOutput struct {
+	Units  []string `json:"units,omitempty"`
+	Stacks []string `json:"stacks,omitempty"`
+}
+
+func outputJSON(configs discovery.DiscoveredConfigs) error {
+	output := JSONOutput{
+		Units:  configs.Filter(discovery.ConfigTypeUnit).Paths(),
+		Stacks: configs.Filter(discovery.ConfigTypeStack).Paths(),
+	}
+
+	jsonBytes, err := json.MarshalIndent(output, "", "  ")
+	if err != nil {
+		return err
+	}
+
+	fmt.Println(string(jsonBytes))
+	return nil
+}
+
+type Colorizer struct {
+	unitColorizer  func(string) string
+	stackColorizer func(string) string
+}
+
+func NewColorizer() *Colorizer {
+	return &Colorizer{
+		unitColorizer:  ansi.ColorFunc(ansi.LightGreen),
+		stackColorizer: ansi.ColorFunc(ansi.LightBlue),
+	}
+}
+
+func (c *Colorizer) colorize(config *discovery.DiscoveredConfig) string {
+	switch config.ConfigType() {
+	case discovery.ConfigTypeUnit:
+		return c.unitColorizer(config.Path())
+	case discovery.ConfigTypeStack:
+		return c.stackColorizer(config.Path())
+	default:
+		return config.Path()
+	}
+}
+
+func outputText(opts *Options, configs discovery.DiscoveredConfigs) error {
+	if opts.TerragruntOptions.Logger.Formatter().DisabledColors() {
+		for _, config := range configs {
+			fmt.Println(config.Path(), ansi.Reset)
+		}
+
+		return nil
+	}
+
+	colorizer := NewColorizer()
+
+	for _, config := range configs {
+		fmt.Println(colorizer.colorize(config), ansi.Reset)
+	}
+
+	return nil
+}
--- a/cli/commands/find/command.go
+++ b/cli/commands/find/command.go
@@ -1,0 +1,64 @@
+// Package find provides the ability to find Terragrunt configurations in your codebase
+// via the `terragrunt find` command.
+package find
+
+import (
+	"github.com/gruntwork-io/terragrunt/cli/flags"
+	"github.com/gruntwork-io/terragrunt/internal/cli"
+	"github.com/gruntwork-io/terragrunt/internal/errors"
+	"github.com/gruntwork-io/terragrunt/internal/experiment"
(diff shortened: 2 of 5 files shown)


internal/discovery/discovery.go
```diff
@@ -0,0 +1,163 @@
+// Package discovery provides functionality for discovering Terragrunt configurations.
+package discovery
+
+import (
+	"errors"
```
levkohimins, line 5:
```suggestion
	"github.com/gruntwork-io/terragrunt/internal/errors"
```

internal/discovery/discovery.go
```diff
@@ -0,0 +1,163 @@
...
+func (c *DiscoveredConfig) String() string {
+	return string(c.Type) + ": " + c.Path
+}
+
+func (d *Discovery) Discover() (DiscoveredConfigs, error) {
+	var units DiscoveredConfigs
+
+	walkFn := func(path string, e fs.DirEntry, err error) error {
+		if err != nil {
+			return errors.New(err.Error())
```
levkohimins, line 74:
```suggestion
			return errors.New(err)
```

internal/discovery/discovery.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+
+		if e.IsDir() {
+			return nil
+		}
+
+		path, err = filepath.Rel(d.WorkingDir, path)
+		if err != nil {
+			return errors.New(err.Error())
```
levkohimins, line 83:
```suggestion
			return errors.New(err)
```

internal/discovery/discovery.go
```diff
@@ -0,0 +1,163 @@
...
+				Type: ConfigTypeStack,
+				Path: filepath.Dir(path),
+			})
+		}
+
+		return nil
+	}
+
+	if err := filepath.WalkDir(d.WorkingDir, walkFn); err != nil {
+		return nil, errors.New(err.Error())
```
levkohimins, line 107:
```suggestion
		return nil, errors.New(err)
```

levkohimins (changes requested):
Great job 👍

cli/commands/find/command.go
```diff
@@ -0,0 +1,69 @@
...
+			EnvVars:     tgPrefix.EnvVars(FormatFlagName),
+			Destination: &opts.Format,
+			Usage:       "Output format for the find results. Valid values: text, json",
+			DefaultText: "text",
+		}),
+		flags.NewFlag(&cli.GenericFlag[string]{
+			Name:        SortFlagName,
+			EnvVars:     tgPrefix.EnvVars(SortFlagName),
+			Destination: &opts.Sort,
+			Usage:       "Sort order for the find results. Valid values: alpha", // TODO: add dag
```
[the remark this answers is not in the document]
yhakbar: It'll help me not to forget to update the Usage once I add support for that, ya.

cli/commands/find/action.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/gruntwork-io/terragrunt/internal/errors"
+	"github.com/mgutz/ansi"
+)
+
+func Run(ctx context.Context, opts *Options) error {
+	d := discovery.NewDiscovery(opts.WorkingDir)
+
+	configs, err := d.Discover()
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
yhakbar: I am really not a fan of the fact that we do this everywhere. I know we do it largely so that we can ensure we get stack traces, but it seems to me to be non-standard Go.

internal/discovery/discovery.go
```diff
@@ -0,0 +1,163 @@
...
+			units = append(units, &DiscoveredConfig{
+				Type: ConfigTypeStack,
+				Path: filepath.Dir(path),
+			})
+		}
+
+		return nil
+	}
+
+	if err := filepath.WalkDir(d.WorkingDir, walkFn); err != nil {
```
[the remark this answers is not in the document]
yhakbar: I'll add that afterwards. Adding to the experiment docs that it's not supported yet.

cli/commands/find/command.go
```diff
@@ -0,0 +1,84 @@
...
+			Name:        HiddenFlagName,
+			EnvVars:     tgPrefix.EnvVars(HiddenFlagName),
+			Destination: &opts.Hidden,
+			Usage:       "Include hidden directories in the find results.",
+		}),
+	}
+}
+
+func NewCommand(opts *options.TerragruntOptions) *cli.Command {
+	cmdOpts := NewOptions(opts)
```
levkohimins, line 57:
```suggestion
        prefix := flags.Prefix{CommandName}
	cmdOpts := NewOptions(opts)
```

cli/commands/find/command.go
```diff
@@ -0,0 +1,84 @@
...
+
+	return &cli.Command{
+		Name:                 CommandName,
+		Aliases:              []string{CommandAlias},
+		Usage:                "Find relevant Terragrunt configurations.",
+		ErrorOnUndefinedFlag: true,
+		Flags:                NewFlags(cmdOpts, nil),
```
levkohimins, line 64:
```suggestion
		Flags:                NewFlags(cmdOpts, prefix),
```

levkohimins (commented):
The suggested changes prepend the command name to the env var names, for example  `TG_FIND_JSON` instead of `TG_JSON`

cli/commands/find/command.go
yhakbar: Why?

It's confusing to have environment variables sometimes pretended with the command name, and other times not (e.g. `TG_AUTH_PROVIDER_CMD` is not `TG_RUN_AUTH_PROVIDER_CMD`).

The majority of flags don't collide anymore.

levkohimins: What I think, `run` flags shouldn't contain command name prefix. So `TG_AUTH_PROVIDER_CMD` should leave as it is `TG_AUTH_PROVIDER_CMD`. But other command flags should have the command prefix. Why?!

Environment variables can be set permanent in the shell. In these case some flags can collide, for example:

`hcl validate -json` TG_JSON
`find -json` TG_JSON

That's why I suggest that they be called that way

```
TG_FIND_JSON
TG_HCL_VALIDATE_JSON
```

levkohimins: `TG_HIDDEN`

Also such short names of environment variables do not say much about what they refer to.

levkohimins: The case from my task:

`terragurnt backend delete -bucket`

`TG_BUCKET` or `TG_BACKEND_DELETE_BUCKET`

yhakbar: If we are consistent about following the rule that you can always set a flag by using `TG_<NAME_OF_FLAG>`, then it's easy to guess what every flag means. It means the same thing as the name of the flag. If users want, they can unset env vars or rely on CLI flags instead of env vars if they want to avoid collisions.

Whenever we can, we should design the interface so that nobody has to read the docs, or `--help` to know how to use it. That includes having to learn that there's a different naming convention for `run` (and `exec`) flags, and the rest.

levkohimins (approved):

PR #4056, opened 2025-03-20
[WIP] Stacks RFC: values with dependencies

cgetzen:
## Description

Easily pass values between unit definitions from the terragrunt.stack.hcl file while maintaining backwards compatibility with units.

Context:
- [Original Request](https://github.com/gruntwork-io/terragrunt/issues/3313#issuecomment-2705391439): Now that stacks can create multiple instances of a single unit, should the coupling of implementation details with units be revisited? It feels at odds with the stack concept of dynamically setting up units.
- [Response](https://github.com/gruntwork-io/terragrunt/issues/3313#issuecomment-2721552458): Can you think of a way to define dependencies for units inside the terragrunt.stack.hcl file that doesn't break this?
- [Solution Description](https://github.com/gruntwork-io/terragrunt/issues/3313#issuecomment-2723936374): Continue to use the values file interface.

This PR implements the following format:

```
# terragrunt.stack.hcl

unit "dep" {
  source = "modules/imadep"
  path   = "imadep"
}

unit "test" {
  source = "modules/test"
  path   = "test"
  values = {
    a = "a"
    b = unit.dep.output_b
  }
}


# .terragrunt-stack/test/terragrunt.values.hcl (auto-generated)

values {
    a = "a"
    b = dependency.dep.outputs.output_b
}

dependency "dep" {
    config_path = "../imadep"
}
```

Before merging:
- Confirm that this behavior is in line with the Stacks RFC vision
- Handle all error cases
- Ensure performance is not degraded

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

### DIFF
--- a/config/hclparse/file.go
+++ b/config/hclparse/file.go
@@ -75,11 +75,22 @@ func (file *File) Decode(out any, evalContext *hcl.EvalContext) (err error) {
 
 // Blocks takes a parsed HCL file and extracts a reference to the `name` block, if there are defined.
 func (file *File) Blocks(name string, isMultipleAllowed bool) ([]*Block, error) {
+	var labelNames []string
+	switch name {
+	case "dependency":
+		labelNames = []string{"name"}
+	case "values":
+		// values block doesn't have labels
+		labelNames = []string{}
+	default:
+		labelNames = []string{"name"}
+	}
+
 	catalogSchema := &hcl.BodySchema{
 		Blocks: []hcl.BlockHeaderSchema{
 			{
 				Type:       name,
-				LabelNames: []string{"name"},
+				LabelNames: labelNames,
 			},
 		},
 	}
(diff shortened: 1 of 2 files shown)


yhakbar:
Thanks for submitting this PoC, @cgetzen . It's a neat idea! 

Please [create an RFC following our GitHub issues based process](https://github.com/gruntwork-io/terragrunt/issues/new?template=02-rfc.yml) so we can get more details on the design thought process, think through alternatives, etc. Make sure you link this PoC to it. It really helps to have something like this to understand how the implementation would work.

Regarding the design as-is: I don't think it would be a good idea to have dependencies in `terragrunt.values.hcl` files. They are intended to be flat, simple HCL attributes, making them easy to parse and work with. When you submit the RFC, I would encourage you to update the design to perhaps leverage a separate `terragrunt.dependencies.hcl` file or something (to be clear, I'm not saying that we'd accept a proposal that did that, I'm just saying it would be a preferred alternative to overloading the `terragrunt.values.hcl` file).

I'd also like to get a better understanding of the edge cases in this proposal: What if two dependencies are named the same thing across these different files? What's the behavior? Can the `terragrunt.hcl` file make reference to them? Are there any alternative designs that you think would work just as well? How are mock outputs handled?

PR #4089, opened 2025-03-27
fix: Fixing `backend delete` safety features

yhakbar:
## Description

Adds requisite check that the backend state bucket is versioned before performing any state deletion as a safeguard for users.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added `--force` flag to bypass safety feature disallowing usage of `backend delete` on buckets that are not versioned.

## Summary by CodeRabbit

- **New Features**
  - Added a new CLI option, `--force`, for backend deletion. When used, it allows the deletion to proceed even if bucket versioning isn’t enabled.
  - Introduced a configuration option to conditionally skip bucket versioning during remote state bootstrapping.
  
- **Tests**
  - Expanded integration tests for AWS and GCP backends to cover scenarios with versioning disabled and to verify the behavior of the new force deletion capability.

### DIFF
--- a/internal/cli/command.go
+++ b/internal/cli/command.go
@@ -115,15 +115,7 @@ func (cmd *Command) VisibleSubcommands() Commands {
 		return nil
 	}
 
-	visibleSubcommands := make(Commands, 0, len(cmd.Subcommands))
-
-	for _, subcommand := range cmd.Subcommands {
-		if !subcommand.Hidden {
-			visibleSubcommands = append(visibleSubcommands, subcommand)
-		}
-	}
-
-	return visibleSubcommands
+	return cmd.Subcommands.VisibleCommands()
 }
 
 // Run parses the given args for the presence of flags as well as subcommands.
--- a/internal/cli/command.go
+++ b/internal/cli/command.go
@@ -115,7 +115,15 @@ func (cmd *Command) VisibleSubcommands() Commands {
 		return nil
 	}
 
-	return cmd.Subcommands.VisibleCommands()
+	visibleSubcommands := make(Commands, 0, len(cmd.Subcommands))
+
+	for _, subcommand := range cmd.Subcommands {
+		if !subcommand.Hidden {
+			visibleSubcommands = append(visibleSubcommands, subcommand)
+		}
+	}
+
+	return visibleSubcommands
 }
 
 // Run parses the given args for the presence of flags as well as subcommands.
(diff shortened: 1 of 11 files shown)


internal/cli/command.go
```diff
@@ -105,7 +105,15 @@ func (cmd *Command) Subcommand(name string) *Command {
...
-	return cmd.Flags
+	visibleFlags := make(Flags, 0, len(cmd.Flags))
+
+	for _, flag := range cmd.Flags {
+		if !flag.GetHidden() {
+			visibleFlags = append(visibleFlags, flag)
+		}
+	}
+
+	return visibleFlags
```
levkohimins, line 116:
```suggestion
	return cmd.Flags.VisibleFlags()
```

internal/remotestate/backend/s3/backend.go
```diff
@@ -138,6 +139,17 @@ func (backend *Backend) Delete(ctx context.Context, backendConfig backend.Config
 		return err
 	}
 
+	if !opts.ForceBackendDelete {
```
levkohimins, line 142: NIT: I'd rather pass it as an argument to the function.

yhakbar: We shouldn't have conflicting sources of truth. If we're going to take it as an argument, then we shouldn't have `opts` as an argument, as that will have another copy of the same value.

levkohimins: It would be great if `remotestate` didn't know anything about options at all, as stated in the book `Clean Architecture`, the called method shouldn't know anything about the properties etc. of the called method.

levkohimins: I tried to implement this during the refactoring, but because we pass some authorization parameters via options, I didn't spend much time on it.

yhakbar: I totally agree, but as it stands, it does. While it has access to the `opts` variable, it shouldn't receive potentially conflicting data from a different variable. 

If we want to make it agnostic, we need to refactor out all usage of `opts`, including in `NewClient`, etc. so that there isn't risk of conflicting information. 

e.g. What if I set the argument `forceBackendDelete` to true, but `opts.ForceBackendDelete` is false? It's an inconsistency that can lead to bugs.

yhakbar: It's better to avoid that kind of refactor without making it all we do in the PR. We use `opts` everywhere, so we need to be careful with how we remove usage.

levkohimins (dismissed):

levkohimins (approved):
Great job 👍

Issue #3535, opened 2024-11-07
Road to 1.0 Deprecated Feature Removal Schedule


yhakbar:
On the road to the first major release of Terragrunt 1.0, some breaking changes will take place that remove functionality in Terragrunt.

While features will primarily be introduced as backwards compatible, deprecated features will be gradually removed leading up to 1.0, and remain supported until the next major version, 2.0. More on this can be found [here, in the CLI Redesign RFC](https://www.notion.so/gruntwork/PUBLIC-Terragrunt-CLI-Redesign-10ed8759a6828042ac6bc03aef4d9dfb?pvs=4#10ed8759a68281768cb1dab32dfa4bae).

Note that users can also leverage [Strict Mode](https://terragrunt.gruntwork.io/docs/reference/strict-mode/) to _opt-in_ to these breaking changes before they occur to confirm that they are ready to adopt the releases where the changes take place. All deprecated functionality will be introduced to a strict mode control before it is fully removed to give users the chance to try out the removal preemptively.

The items listed below will be listed in order of their eventual removal. Initially, they will be listed with To Be Decided (TBD) deadlines for their removal, but as we get closer to 1.0, they will be scheduled, and removed slowly.

This is a living document. If you have concerns or would like to request that the order or deadline for feature removal change, please share your feedback. More may be added to this schedule as we get closer to 1.0.

## Removal Schedule

1. Removal of longstanding deprecated commands (e.g. `plan-all`) - TBD
2. Removal of legacy Terragrunt flags (all flags that start with `terragrunt-`) - TBD
3. Removal of backend provisioning as default behavior - TBD
4. Removal of command forwarding by default - TBD
5. Removal of deprecated commands (e.g. `terragrunt-info`) - TBD
6. Removal of behavior where inputs are read from dependencies - TBD

## Removed Features

The following is a breakdown of the items listed above, and why they are being removed.

### Removal of longstanding deprecated commands

For quite a long time, commands like `plan-all` have been deprecated in favor of the `run-all` command, which can be used in combination with other commands like `plan`, `apply`, etc instead of a dedicated command. Most users have already adopted `run-all`, so this should be relatively safe to do today.

### Removal of legacy Terragrunt flags

[...]

PR #4120, opened 2025-04-02
feat: Add performance guide

yhakbar:
## Description

Adds some basic guidance on performance improvements for Terragrunt users/developers.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added  performance documentation for Terragrunt users/developers.

## Summary by CodeRabbit

- **Documentation**
  - Added a new performance optimization guide with strategies for measuring and enhancing Terragrunt’s performance.
  - Updated troubleshooting content to highlight trace collection using OpenTelemetry.
  - Improved catalog pages with enhanced header structure and spacing for better readability.
  - Expanded contribution guidelines with detailed sections on testing and linting practices.
  - Clarified CLI options documentation, including validation processes and experimental features.
  - Introduced additional notes on the Provider Cache Server as a performance optimization feature.
  - Enhanced strict mode documentation with new sections for better organization and clarity.
  - Updated guidelines for the `dependency-fetch-output-from-state` flag, emphasizing version management and performance optimization tips.

### DIFF
--- a/docs-starlight/src/content/docs/04-reference/03-strict-controls.mdx
+++ b/docs-starlight/src/content/docs/04-reference/03-strict-controls.mdx
@@ -6,6 +6,8 @@ sidebar:
   order: 3
 ---
 
+import { Aside } from '@astrojs/starlight/components';
+
 Terragrunt supports operating in a mode referred to as "Strict Mode".
 
 Strict Mode is a set of controls that can be enabled to ensure that your Terragrunt usage is future-proof
@@ -140,6 +142,10 @@ Throw an error when using the `validate-all` command.
 ### skip-dependencies-inputs
 
 Disable reading of dependency inputs to enhance dependency resolution performance by preventing recursively parsing Terragrunt inputs from dependencies.
+
+<Aside type="tip">
+Skipping dependency inputs is a performance optimization. For more details on performance optimizations, their tradeoffs, and other performance tips, read the dedicated [Performance documentation](/docs/troubleshooting/performance).
+</Aside>
 
 **Reason**: Enabling the `skip-dependencies-inputs` option prevents the recursive parsing of Terragrunt inputs, leading to optimized performance during dependency resolution.
 
--- a/docs-starlight/src/content/docs/05-troubleshooting/03-performance.mdx
+++ b/docs-starlight/src/content/docs/05-troubleshooting/03-performance.mdx
@@ -16,10 +16,10 @@ However, given the nature of the problems that Terragrunt solves, there are some
 
 One of the most expensive things that OpenTofu/Terraform does, from a bandwidth and disk utilization perspective, is download and install providers. These are large binary files that are downloaded from the internet, and not cached across units by default.
 
-You can significantly reduce the amount of time taken by Terragrunt runs by enabling the [provider cache server](/docs/features/provider-cache-server), like this:
+You can significantly reduce the amount of time taken by Terragrunt runs by enabling the provider cache server, like this:
 
 ```shell
-terragrunt run-all plan --provider-cache
+terragrunt run --all plan --provider-cache
 ```
 
 #### Provider Cache - Gotchas
@@ -45,12 +45,14 @@ You can significantly improve the performance of dependency blocks by using the 
 For example:
 
 ```shell
-terragrunt run-all plan --dependency-fetch-output-from-state
+terragrunt run --all plan --dependency-fetch-output-from-state
 ```
 
 #### Fetching Output From State - Gotchas
 
-**NOTE**: This is an experimental feature, use with caution. There is no guarantee that OpenTofu/Terraform will maintain the existing schema of their state files, so there is also no guarantee that the flag will work as expected in future versions of OpenTofu/Terraform. Also, note that the only backend that supports this feature is the AWS S3 backend.
+The first thing you need to be aware of when considering usage of the `--dependency-fetch-output-from-state` flag is that it only works for S3 backends. If you are using a different backend, you should not use this flag.
+
+Next, you should be aware that there is no guarantee that OpenTofu/Terraform will maintain the existing schema of their state files, so there is also no guarantee that the flag will work as expected in future versions of OpenTofu/Terraform.
 
 We are coordinating with the OpenTofu team to improve the performance of the `output` command, and we hope that this flag will be unnecessary for most users in the future.
 
@@ -62,7 +64,7 @@ A rarely used capability of Terragrunt dependency blocks is the ability to read 
 
 The reason for this is that Terragrunt needs to recursively read inputs from dependencies, and their dependencies, all the way down to the ancestral Terragrunt unit they all have in common to be able to expose the inputs to the dependent unit.
 
-You can mitigate this performance penalty by using the `skip-dependency-inputs` strict control. For more information about strict controls, refer to the [Strict Mode documentation](/docs/reference/strict-mode).
+You can mitigate this performance penalty by using the `skip-dependency-inputs` strict control.
 
 Enabling this strict control will enable a breaking change in Terragrunt behavior, making it so that dependency blocks no longer read inputs from dependencies.
 
--- a/docs-starlight/src/content/docs/05-troubleshooting/03-performance.mdx
+++ b/docs-starlight/src/content/docs/05-troubleshooting/03-performance.mdx
@@ -60,13 +60,13 @@ See [#1549](https://github.com/opentofu/opentofu/issues/1549) for more details.
 
 ### Skip Dependency Inputs
 
-A rarely used capability of Terragrunt dependency blocks is the ability to read inputs from dependencies. There's very little in the way of valid use-cases for this, and supporting this behavior is a source of significant performance penalty for Terragrunt users.
+Terragrunt dependency blocks allow reading inputs directly from other dependencies. However, the mechanism required to support this capability introduces performance overhead during Terragrunt operations. Due to this performance impact, using this feature is heavily discouraged.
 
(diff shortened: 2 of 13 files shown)


docs-starlight/src/content/docs/04-reference/03-strict-controls.mdx
```diff
@@ -141,6 +143,10 @@ Throw an error when using the `validate-all` command.
 
 Disable reading of dependency inputs to enhance dependency resolution performance by preventing recursively parsing Terragrunt inputs from dependencies.
 
+<Aside type="tip">
+Skipping dependency inputs is a performance optimization. For more details on performance optimizations, their tradeoffs, and other performance tips, read the dedicated [Performance documentation](/docs/troubleshooting/performance).
```
ZachGoldberg, line 147: as a user reading this, the question that comes to mind is "So why not have it always disabled?"

docs-starlight/src/content/docs/05-troubleshooting/03-performance.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+For example:
+
+```shell
+terragrunt run --all plan --dependency-fetch-output-from-state
+```
+
+#### Fetching Output From State - Gotchas
+
+The first thing you need to be aware of when considering usage of the `--dependency-fetch-output-from-state` flag is that it only works for S3 backends. If you are using a different backend, you should not use this flag.
```
ZachGoldberg, line 53: s/should not/can not/?  What happens if they do, i assume it blows up?

docs-starlight/src/content/docs/05-troubleshooting/03-performance.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+Next, you should be aware that there is no guarantee that OpenTofu/Terraform will maintain the existing schema of their state files, so there is also no guarantee that the flag will work as expected in future versions of OpenTofu/Terraform.
+
+We are coordinating with the OpenTofu team to improve the performance of the `output` command, and we hope that this flag will be unnecessary for most users in the future.
+
+See [#1549](https://github.com/opentofu/opentofu/issues/1549) for more details.
+
+### Skip Dependency Inputs
+
+A rarely used capability of Terragrunt dependency blocks is the ability to read inputs from dependencies. There's very little in the way of valid use-cases for this, and supporting this behavior is a source of significant performance penalty for Terragrunt users.
```
ZachGoldberg, line 63: There's a handful of handwaivy qualifiers in this paragraph.  "Rarely used".  "Very little in the way of valid use-cases".  "Significant performance penalty."

Can we tighten it up a bit and restate plainly/factfully?

docs-starlight/src/content/docs/05-troubleshooting/03-performance.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+The reason for this is that Terragrunt needs to recursively read inputs from dependencies, and their dependencies, all the way down to the ancestral Terragrunt unit they all have in common to be able to expose the inputs to the dependent unit.
+
+You can mitigate this performance penalty by using the `skip-dependency-inputs` strict control.
+
+Enabling this strict control will enable a breaking change in Terragrunt behavior, making it so that dependency blocks no longer read inputs from dependencies.
+
+In the future, we will be removing this capability from Terragrunt, so you won't need to use this strict control.
```
ZachGoldberg, line 71: Do we know when the removal will be? 1.0?  Can we say that explicitly?

docs-starlight/src/content/docs/05-troubleshooting/03-performance.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+### Skip Dependency Inputs - Gotchas
+
+In general, you should use this strict control. The only reason not to use it is if you are using dependency blocks to read inputs from dependencies, which you should stop doing as soon as possible.
```
ZachGoldberg, line 75: Should we include this in our day 1 tutorials?  Or maybe have a point in time where we'll set the flag by default and turn the feature into an opt-in instead of opt-out?

docs-starlight/src/content/docs/05-troubleshooting/03-performance.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+You can use configurations like the `--warmup` flag to do some warmup runs before the actual benchmarking. This is useful to get a more accurate measurement of the performance of Terragrunt with cache populated, etc.
+
+Here's an example of how to use Hyperfine to benchmark the performance of Terragrunt with two different configurations:
+
+```shell
+hyperfine -w 3 -r 5 'terragrunt run-all plan' 'terragrunt run-all plan --dependency-fetch-output-from-state'
+```
+
+### Developer
```
ZachGoldberg, line 105: This is for _terragrunt_ developers, right?  Maybe worth adding that qualifier, I was confused for a minute why these docs were teaching me about golang benchmarking

docs-starlight/src/data/flags/dependency-fetch-output-from-state.mdx
```diff
@@ -16,7 +16,11 @@ The main benefit this flag provides is performance. Reading directly from state
 The limitation of this approach is that it is only supported by the S3 backend, and OpenTofu/Terraform may change the schema of the state file in the future, breaking this functionality.
 
 <Aside type="caution">
-This is an experimental feature. When enabled, Terragrunt will fetch dependency outputs directly from the state file instead of using `tofu output`.
+This is an experimental feature, use with caution.
```
ZachGoldberg, line 19: What does "use with caution" really mean?  What should people be aware of?  IMO we should be pretty explicit about when to use/not use based on tradeoffs.  Yes its ultimately end user responsibility to maintain their stuff, but they come to use to get best practice and to gain understanding.  We're the experts.

docs/_docs/04_reference/02-cli-options.md
```diff
@@ -2208,8 +2208,10 @@ block is not evaluated until _after_ the queue has been populated with units to
 
 When using many dependencies, this option can speed up the dependency processing by fetching dependency output directly
 from the state file instead of using `tofu/terraform output` to fetch them.
-NOTE: This is an experimental feature, use with caution.
-Currently only AWS S3 backend is supported.
+
+**NOTE**: This is an experimental feature, use with caution. There is no guarantee that OpenTofu/Terraform will maintain the existing schema of their state files, so there is also no guarantee that the flag will work as expected in future versions of OpenTofu/Terraform. Also, note that the only backend that supports this feature is the AWS S3 backend.
```
ZachGoldberg, line 2212: Maybe the suggestion is, rather than "use with caution" it's "Use with pinned versions of OpenTofu/Terraform, and be on the lookout when upgrading that if the tools change the underlying scheme, Terragrunt will also require updates and there may be version incompatibilities for a period of time" or similar.

ZachGoldberg: All that said -- has the s3 state file format changed in years?  Is this really something our users need to be concerned about, or is it more of a "true, but very unlikely to be a problem" kinda thing?

docs-starlight/src/content/docs/04-reference/03-strict-controls.mdx
yhakbar: In the performance documentation docs, I explicitly say basically "you should definitely always have this disabled"

docs-starlight/src/content/docs/05-troubleshooting/03-performance.mdx
yhakbar: Na. It just ignores the flag.

docs-starlight/src/content/docs/05-troubleshooting/03-performance.mdx
yhakbar: It's on the road to 1.0, and there's no deadline at the moment.

docs-starlight/src/content/docs/05-troubleshooting/03-performance.mdx
yhakbar: We're trying hard to avoid breaking changes except when scheduled.

Once we're done with the CLI Redesign, we'll start removing deprecated behavior following this schedule:
https://github.com/gruntwork-io/terragrunt/issues/3535

docs-starlight/src/data/flags/dependency-fetch-output-from-state.mdx
yhakbar: It's the next line...

There is no guarantee that OpenTofu/Terraform will maintain the existing schema of their state files, so there is also no guarantee that the flag will work as expected in future versions of OpenTofu/Terraform. Also, note that the only backend that supports this feature is the AWS S3 backend.

docs/_docs/04_reference/02-cli-options.md
yhakbar: AFAIK, it's never changed, and we're coordinating with OpenTofu maintainers to avoid any changes there if possible, but we have no influence on Terraform development.

yhakbar: Added some notes.

docs-starlight/src/content/docs/05-troubleshooting/03-performance.mdx
yhakbar: Updated.

docs-starlight/src/content/docs/05-troubleshooting/03-performance.mdx
ZachGoldberg: > In general, you should use this strict control.

My concern is that we have this kind of universal "everyone should do this" guidance buried in some performance documentation that most terragrunt users will never read, so most users won't actually be getting the value here.

ZachGoldberg (approved):

PR #4135, opened 2025-04-07
feat: Implementation `backend migrate` command

levkohimins:
## Description

Implementation of the `backend migrate` command for AWS S3 and GCS backends.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

- **New Features**
  - The application now displays help by default when no command is provided.
  - Introduced a new "migrate" command that enables remote state migration across various cloud backends.
  - Enhanced configuration options with dynamic key prefix support and new flags for bucket deletion and forced migration.

- **Bug Fixes**
  - Improved error handling for non-existent buckets during version control checks and migrations.

- **Tests**
  - Expanded the test suite to validate migration and remote state operations across AWS, GCS, and S3 environments.
  - Added new tests specifically for backend migration scenarios.


cli/commands/backend/migrate/cli.go
```diff
@@ -0,0 +1,55 @@
...
+	cmd := &cli.Command{
+		Name:                 CommandName,
+		Usage:                "Migrate OpenTofu/Terraform state from one location to another.",
+		UsageText:            "terragrunt backend migrate [options] <src-path> <dst-path>",
+		Flags:                NewFlags(opts, nil),
+		ErrorOnUndefinedFlag: true,
+		Action: func(ctx *cli.Context) error {
+			srcPath := ctx.Args().First()
+			if srcPath == "" {
+				return errors.New("not specified src-path")
```
yhakbar, line 42: Let's replace this with a reference to the usage text.

cli/commands/backend/migrate/cli.go
```diff
@@ -0,0 +1,55 @@
...
+			}
+
+			dstPath := ctx.Args().Second()
+			if dstPath == "" {
+				return errors.New("not specified dst-path")
```
yhakbar, line 47: Let's replace this with a reference to the usage text.

cli/commands/backend/migrate/migrate.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"context"
+
+	"github.com/gruntwork-io/terragrunt/config"
+	"github.com/gruntwork-io/terragrunt/internal/errors"
+	"github.com/gruntwork-io/terragrunt/internal/remotestate/backend"
+	"github.com/gruntwork-io/terragrunt/options"
+)
+
+func Run(ctx context.Context, srcPath, dstPath string, opts *options.TerragruntOptions) error {
```
yhakbar, line 13: NIT: We can replace this with `src` and `dst`.

cli/commands/backend/migrate/migrate.go
```diff
@@ -0,0 +1,37 @@
...
+	remoteState, err := config.ParseRemoteState(ctx, opts)
```
yhakbar, line 14: This is parsing remote state configurations for the current directory? If so, why? Shouldn't this always be parsing state in a particular directory?

cli/commands/backend/migrate/migrate.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	if remoteState == nil {
+		opts.Logger.Debug("Did not find remote `remote_state` block in the config")
+
+		return nil
+	}
+
+	if !opts.ForceBackendMigrate {
+		enabled, err := remoteState.IsVersionControlEnabled(ctx, opts)
```
yhakbar, line 26: We only need to check this for the source directory, right?

yhakbar (dismissed):
My comments can be addressed later. Looks good!

internal/remotestate/backend/gcs/client.go
```diff
@@ -387,3 +391,74 @@ func (client *Client) DeleteGCSObjects(ctx context.Context, bucketName, prefix s
...
+	}
+
+	client.logger.Debugf("Remote state GCS bucket %s object %s does not exist or you don't have permissions to access it.", bucketName, key)
+
+	return false, nil
+}
+
+// MoveGCSObject copies the GCS object at the specified srcKey to dstKey and then removes srcKey.
+func (client *Client) MoveGCSObject(ctx context.Context, bucketName, srcKey, dstKey string) error {
+	if err := client.CopyGCSBucketObject(ctx, bucketName, srcKey, dstKey); err != nil {
```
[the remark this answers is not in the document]
levkohimins: This code moves a remote state file inside a bucket. As far as I know `state/push` uploads/downloads a state file, to cloud/local. Thus, we need to change the configuration several times, first to download the remote state from the old path and then to upload it to the new path. The idea is interesting, but it will require a lot of changes.

cli/commands/backend/migrate/migrate.go
levkohimins: Versioning can only be enabled for the entire bucket. We take the bucket settings from the remote state block, where there can only be one.

cli/commands/backend/migrate/migrate.go
levkohimins: > This is parsing remote state configurations for the current directory?
No, it works the same as for other commands, mainly takes into account `working-dir`, and so on.

cli/commands/backend/migrate/migrate.go
levkohimins: Replaced

cli/commands/backend/migrate/cli.go
levkohimins: Replaced

cli/commands/backend/migrate/cli.go
levkohimins: Replaced

cli/commands/backend/migrate/migrate.go
```diff
@@ -0,0 +1,69 @@
...
+		if err != nil && !errors.As(err, new(backend.BucketDoesNotExistError)) {
+			return err
+		}
+
+		if !enabled {
+			return errors.Errorf("src bucket is not versioned, refusing to migrate backend state. If you are sure you want to migrate the backend state anyways, use the --%s flag", ForceBackendMigrateFlagName)
+		}
+	}
+
+	return srcRemoteState.Migrate(ctx, dstRemoteState.BackendConfig, opts)
```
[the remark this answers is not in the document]
levkohimins: Right, it will work only with the same type of backend and in the same for example AWS region, but it supports different bucket names. Because we currently only support copying within AWS S3 or GCS using their API, for example

```go
	input := &s3.CopyObjectInput{
		Bucket:     aws.String(dstBucketName),
		Key:        aws.String(dstKey),
		CopySource: aws.String(path.Join(srcBucketName, srcKey)),
	}
```

To copy data between different AWS regions or even between different clouds, we should copy the remote state to local and then upload to the cloud, perhaps using the `state pull` and `state push` commands you suggested.

So, yes, to fully support migration, we will need another improvement.

PR #4288, opened 2025-05-15
feat: performance improvement on -version execution

denis256:
## Description

* added caching of -version execution
* added integration test to track that -version is invoked only once

Test results on running 10 times a stack of 20 units on different versions vs current branch

hyperfine-v2

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

- **New Features**
	- Improved performance by adding caching for Terraform version detection, reducing redundant version checks.
- **Tests**
	- Added tests to verify that the Terraform version command is invoked only once and correctly across different directories.
- **Bug Fixes**
	- Corrected the working directory used in a dependency validation test.


cli/commands/run/version_check.go
```diff
@@ -212,6 +237,31 @@ func parseTerraformImplementationType(versionCommandOutput string) (options.Terr
 	}
 }
 
+// Helper to compute a cache key from the checksums of .terraform-version and .tool-versions
+func computeVersionFilesCacheKey(workingDir string) string {
+	var hashes []string
+
+	files := []string{".terraform-version", ".tool-versions"}
```
yhakbar, line 244: This could be many more files including `mise.toml` or `.mise.toml`.

test/integration_test.go
```diff
@@ -4136,3 +4136,20 @@ func TestTfPath(t *testing.T) {
 
 	assert.Regexp(t, "(?i)(terraform|opentofu)", stdout+stderr)
 }
+
+func TestVersionIsInvokedOnlyOnce(t *testing.T) {
```
yhakbar, line 4140: We should include a version where at least two fire from `.tool-version` files, etc.

cli/commands/run/version_check.go
```diff
@@ -212,6 +237,31 @@ func parseTerraformImplementationType(versionCommandOutput string) (options.Terr
...
+		}
+	}
+
+	cacheKey := "no-version-files"
+
+	if len(hashes) != 0 {
+		cacheKey = strings.Join(hashes, "|")
+	}
+
+	return util.EncodeBase64Sha1(cacheKey)
```
yhakbar, line 262: NIT: Should we be SHA-ing this? It's going to get hashed by the standard library when used for the key of the map, right? I don't think it's super important, just trying to confirm my understanding right.

cli/commands/run/version_check.go
```diff
@@ -27,6 +31,8 @@ var TerraformVersionRegex = regexp.MustCompile(`^(\S+)\s(v?\d+\.\d+\.\d+)`)
 
 const versionParts = 3
 
+var versionFiles = []string{".terraform-version", ".tool-versions", "mise.toml", ".mise.toml"}
```
yhakbar, line 34: Do you think we should make a ticket to backlog making this configurable? Maybe a user can just feed Terragrunt the files that might be used to control TF versions outside these defaults?

yhakbar (dismissed):
We can do the bonus test for checking that version files trigger multiple `-version` calls later, IMO. It doesn't have to block you.

cli/commands/run/version_check.go
denis256: `util.FileSHA256(path)` - returns 8kb byte array, encoded in hex it is quite large key.

`sha1.Sum([]byte(str))` - return 20 bytes

cli/commands/run/version_check.go
denis256: Will raise another ticket for this

yhakbar (approved):

cli/commands/run/version_check.go
yhakbar: I was just saying that I don't know that we need to hash this, as it's used as the key for the hashmap, which is going to get hashed by the standard library, right?

Issue #3629, opened 2024-12-05
Runner Pool


yhakbar:
### Summary

Change the underlying concurrency model of Terragrunt so that a pool of runners are leveraged instead of run groups.

Add Units to the pool of runners when they are ready.

When all Units have completed their runs, end the run.

### Motivation

Terragrunt currently runs Units in a concurrency model where Units are grouped based on their dependencies, and groups are run in parallel when they do not have any pending group they depend upon.

e.g.

```bash
$ terragrunt run-all plan
14:09:52.456 INFO   The stack at . will be processed in the following order for command plan:
Group 1
- Module ./unit-a
- Module ./unit-b

Group 2
- Module ./unit-depends-on-unit-a
- Module ./unit-depends-on-unit-b
```

This is a simple concurrency model, and is easy to display in logs.

Individual Units failing during runs can cause entire groups, and dependent groups to fail, ultimately meaning that individual failing Units can cause widespread failure for a Stack.

In addition, there is wasted time in a run, as groups execute when they have no dependent groups they are waiting on. A group dependent on another group will only start running when the slowest Unit in the dependency completes.

```bash
$ terragrunt run-all plan
14:09:52.456 INFO   The stack at . will be processed in the following order for command plan:
Group 1
- Module ./slow-unit
- Module ./fast-unit

Group 2
- Module ./unit-depends-on-slow-unit
- Module ./unit-depends-on-fast-unit
```

### Proposal

When Terragrunt starts a run, create a runner pool and a Unit queue, then add unblocked Units to the pool from the queue and run them until the queue is empty.

This will make it so that runs can complete more efficiently and so that individual failing Units do not cause the entire run to fail.

Some users may prefer to support the current behavior where Terragrunt will fail early if an individual Unit fails, and to support that use-case, a `--terragrunt-fail-early` flag will be introduced.

### Algorithm

The algorithm for working with the runner pool will be as follows:

1. Discover Units and add them to a queue with metadata regarding their dependencies. 
    
    Store that metadata as a slice of Units the discovered Unit is `blocked by`. 
    
    In addition store a slice of `dependencies` for each Unit.
    
    Set the status of the Units to `ready` if they have an empty list of `blocked by` and set their status to `blocked` if they do not.
    
    Sort the queue by:
    
    1. Units with `ready` status first.
    2. Units with more dependencies before Units with less dependencies.
2. Create a runner pool equal to the minimum of 
      1. The total number of Units
[...]

PR #4434, opened 2025-06-16
Runner pool implementation

denis256:
## Description

* Extracted separated packages to allow implementation of runners
* `internal/runner/common` - base blocks to implement runners
* `internal/runner/configstack` - original configstack runner implementation
* `internal/runner/runnerpool` - runner pool implementation
* `internal/runner/runner.go` - logic to create runners based on experiments
* Extracted data structures to represent Stack and Unit
* Added interface for running stacks `StackRunner` - which is implemented in configstack and runnerpool
* Low-level execution on unit extracted to `UnitRunner` struct
* In configstack, to track dependencies extracted separated struct `DependencyController` to track execution of dependencies
* Added support for `fail-fast` flag, so far only in runner pool implementation
* Tests update

RFC: https://github.com/gruntwork-io/terragrunt/issues/3629

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **New Features**
  * Introduced a new `--fail-fast` CLI flag and environment variable to immediately halt execution if any unit fails (currently effective with the experimental runner-pool engine).
  * Enhanced parallel execution and dependency management for Terragrunt units, improving reliability and error reporting.
  * Improved output and error messages to consistently use "Unit" instead of "Module".

* **Bug Fixes**
  * Corrected terminology and error messages to reflect updated concepts and improve clarity.

* **Documentation**
  * Added and updated documentation for the new `--fail-fast` flag and runner-pool experiment progress.
  * Updated CLI reference and usage examples for new features and terminology.

* **Refactor**
  * Replaced the previous module-centric stack system with a new, unit-centric runner architecture for improved performance and maintainability.
  * Removed legacy code and tests related to the old stack implementation.
  * Migrated and expanded tests to cover new dependency execution logic and error handling.

* **Tests**
  * Added comprehensive tests for new dependency controller, unit runner, queue, and DAG runner logic.
  * Updated integration and unit tests to validate new features and terminology.

### DIFF
--- a/internal/queue/queue.go
+++ b/internal/queue/queue.go
@@ -378,17 +378,6 @@ func (q *Queue) RemainingDeps(e *Entry) int {
 	return count
 }
 
-// AllTerminal returns true if all entries are in a terminal state (Succeeded or Failed).
-func (q *Queue) AllTerminal() bool {
-	for _, e := range q.Entries {
-		if e.Status != StatusSucceeded && e.Status != StatusFailed && e.Status != StatusEarlyExit {
-			return false
-		}
-	}
-
-	return true
-}
-
 // isTerminal returns true if the status is terminal.
 func isTerminal(status Status) bool {
 	return status == StatusSucceeded || status == StatusFailed || status == StatusEarlyExit
--- a/internal/queue/queue_test.go
+++ b/internal/queue/queue_test.go
@@ -239,7 +239,7 @@ func TestQueue_LinearDependencyExecution(t *testing.T) {
 	require.NoError(t, err)
 
 	// Initially, not all are terminal
-	assert.False(t, q.AllTerminal(), "AllTerminal should be false at start")
+	assert.False(t, q.Finished(), "Finished should be false at start")
 
 	// Check that all entries are ready initially and in order A, B, C
 	readyEntries := q.GetReadyWithDependencies()
@@ -251,7 +251,7 @@ func TestQueue_LinearDependencyExecution(t *testing.T) {
 	entryA := readyEntries[0]
 	entryA.Status = queue.StatusSucceeded
 
-	assert.False(t, q.AllTerminal(), "AllTerminal should be false after A is done")
+	assert.False(t, q.Finished(), "Finished should be false after A is done")
 
 	readyEntries = q.GetReadyWithDependencies()
 	assert.Len(t, readyEntries, 1, "After A is done, only B should be ready")
@@ -261,7 +261,7 @@ func TestQueue_LinearDependencyExecution(t *testing.T) {
 	entryB := readyEntries[0]
 	entryB.Status = queue.StatusSucceeded
 
-	assert.False(t, q.AllTerminal(), "AllTerminal should be false after B is done")
+	assert.False(t, q.Finished(), "Finished should be false after B is done")
 
 	readyEntries = q.GetReadyWithDependencies()
 	assert.Len(t, readyEntries, 1, "After B is done, only C should be ready")
@@ -272,7 +272,7 @@ func TestQueue_LinearDependencyExecution(t *testing.T) {
 	entryC.Status = queue.StatusSucceeded
 
 	// Now all should be terminal
-	assert.True(t, q.AllTerminal(), "AllTerminal should be true after all succeeded")
+	assert.True(t, q.Finished(), "Finished should be true after all succeeded")
 
 	readyEntries = q.GetReadyWithDependencies()
 	assert.Empty(t, readyEntries, "After C is done, no entries should be ready")
@@ -350,7 +350,7 @@ func TestQueue_FailFast(t *testing.T) {
 	require.NoError(t, err)
 	q.FailFast = true
 
-	assert.False(t, q.AllTerminal(), "AllTerminal should be false at start")
+	assert.False(t, q.Finished(), "Finished should be false at start")
 
 	// Simulate A failing
 	var entryA *queue.Entry
@@ -380,7 +380,7 @@ func TestQueue_FailFast(t *testing.T) {
 	}
 
 	// Now all should be terminal
-	assert.True(t, q.AllTerminal(), "AllTerminal should be true after fail-fast triggers")
+	assert.True(t, q.Finished(), "Finished should be true after fail-fast triggers")
 
 	// No entries should be ready after fail-fast
 	readyEntries := q.GetReadyWithDependencies()
@@ -525,7 +525,7 @@ func TestQueue_AdvancedDependency_BFails_NoFailFast(t *testing.T) {
 	require.NoError(t, err)
 	q.FailFast = false
 
-	assert.False(t, q.AllTerminal(), "AllTerminal should be false at start")
+	assert.False(t, q.Finished(), "Finished should be false at start")
 
 	// 1. Initially, only A should be ready
 	readyEntries := q.GetReadyWithDependencies()
@@ -536,7 +536,7 @@ func TestQueue_AdvancedDependency_BFails_NoFailFast(t *testing.T) {
 	entryA := readyEntries[0]
 	entryA.Status = queue.StatusSucceeded
 
-	assert.False(t, q.AllTerminal(), "AllTerminal should be false after A is done")
+	assert.False(t, q.Finished(), "Finished should be false after A is done")
 
 	// 2. After A, B and C should be ready
 	readyEntries = q.GetReadyWithDependencies()
@@ -556,7 +556,7 @@ func TestQueue_AdvancedDependency_BFails_NoFailFast(t *testing.T) {
 	entryB.Status = queue.StatusRunning
 	q.FailEntry(entryB)
 
-	assert.False(t, q.AllTerminal(), "AllTerminal should be false after B fails if C is not done")
+	assert.False(t, q.Finished(), "Finished should be false after B fails if C is not done")
 
 	// D and E should be marked as failed due to dependency on B
 	assert.Equal(t, queue.StatusFailed, q.EntryByPath("B").Status)
@@ -572,7 +572,7 @@ func TestQueue_AdvancedDependency_BFails_NoFailFast(t *testing.T) {
 	entryC.Status = queue.StatusSucceeded
 
 	// After C is done, now all should be terminal
-	assert.True(t, q.AllTerminal(), "AllTerminal should be true after all entries are terminal")
+	assert.True(t, q.Finished(), "Finished should be true after all entries are terminal")
 
 	// After C is done, nothing should be ready
 	readyEntries = q.GetReadyWithDependencies()
@@ -591,7 +591,7 @@ func TestQueue_FailFast_SequentialOrder(t *testing.T) {
 	require.NoError(t, err)
 	q.FailFast = true
 
-	assert.False(t, q.AllTerminal(), "AllTerminal should be false at start")
+	assert.False(t, q.Finished(), "Finished should be false at start")
 
 	// Only A should be ready
 	readyEntries := q.GetReadyWithDependencies()
@@ -613,8 +613,8 @@ func TestQueue_FailFast_SequentialOrder(t *testing.T) {
 		}
 	}
 
-	// AllTerminal should be true
-	assert.True(t, q.AllTerminal(), "AllTerminal should be true after fail-fast triggers")
+	// Finished should be true
+	assert.True(t, q.Finished(), "Finished should be true after fail-fast triggers")
 
 	// No entries should be ready
 	readyEntries = q.GetReadyWithDependencies()


internal/runner/configstack/unit.go
```diff
@@ -629,7 +596,7 @@ func (modules TerraformModules) flagExcludedDirs(l log.Logger, opts *options.Ter
 
 var existingModules = cache.NewCache[*TerraformModulesMap](existingModulesCacheName)
 
-type TerraformModulesMap map[string]*TerraformModule
+type TerraformModulesMap map[string]*Unit
```
dnp1: Why not rename to UnitsMap since you already renamed the TerraformModule?

cli/commands/common/runall/runall.go
```diff
@@ -80,30 +82,30 @@ func Run(ctx context.Context, l log.Logger, opts *options.TerragruntOptions) err
...
+	if err := runner.LogUnitDeployOrder(l, opts.TerraformCommand); err != nil {
 		return err
 	}
 
 	var prompt string
 
 	switch opts.TerraformCommand {
 	case tf.CommandNameApply:
-		prompt = "Are you sure you want to run 'terragrunt apply' in each folder of the stack described above?"
+		prompt = "Are you sure you want to run 'terragrunt apply' in each folder of the runner described above?"
```
yhakbar:
```suggestion
		prompt = "Are you sure you want to run 'terragrunt apply' in each unit of the run queue displayed above?"
```

cli/commands/common/runall/runall.go
```diff
@@ -80,30 +82,30 @@ func Run(ctx context.Context, l log.Logger, opts *options.TerragruntOptions) err
...
 	case tf.CommandNameDestroy:
-		prompt = "WARNING: Are you sure you want to run `terragrunt destroy` in each folder of the stack described above? There is no undo!"
+		prompt = "WARNING: Are you sure you want to run `terragrunt destroy` in each folder of the runner described above? There is no undo!"
```
yhakbar:
```suggestion
		prompt = "WARNING: Are you sure you want to run `terragrunt destroy` in each unit of the run queue displayed above? There is no undo!"
```

cli/commands/common/runall/runall.go
```diff
@@ -80,30 +82,30 @@ func Run(ctx context.Context, l log.Logger, opts *options.TerragruntOptions) err
...
 	case tf.CommandNameState:
-		prompt = "Are you sure you want to manipulate the state with `terragrunt state` in each folder of the stack described above? Note that absolute paths are shared, while relative paths will be relative to each working directory."
+		prompt = "Are you sure you want to manipulate the state with `terragrunt state` in each folder of the runner described above? Note that absolute paths are shared, while relative paths will be relative to each working directory."
```
yhakbar:
```suggestion
		prompt = "Are you sure you want to manipulate the state with `terragrunt state` in each unit of the run queue displayed above? Note that absolute paths are shared, while relative paths will be relative to each working directory."
```

cli/commands/run/errors.go
```diff
@@ -47,7 +47,7 @@ type ModuleIsProtected struct {
 }
 
 func (err ModuleIsProtected) Error() string {
-	return fmt.Sprintf("Module is protected by the prevent_destroy flag in %s. Set it to false or delete it to allow destroying of the module.", err.Opts.TerragruntConfigPath)
+	return fmt.Sprintf("Unit is protected by the prevent_destroy flag in %s. Set it to false or delete it to allow destroying of the module.", err.Opts.TerragruntConfigPath)
```
yhakbar:
```suggestion
	return fmt.Sprintf("Unit is protected by the prevent_destroy flag in %s. Set it to false or remove it to allow destruction of the unit.", err.Opts.TerragruntConfigPath)
```

internal/runner/configstack/dependency_controller.go
```diff
@@ -0,0 +1,372 @@
...
+	Dependencies   map[string]*DependencyController
+	NotifyWhenDone []*DependencyController
+}
+
+// NewDependencyController Create a new dependency controller for the given unit.
+func NewDependencyController(unit *runbase.Unit) *DependencyController {
+	return &DependencyController{
+		Runner:         runbase.NewUnitRunner(unit),
+		DependencyDone: make(chan *DependencyController, channelSize),
+		Dependencies:   map[string]*DependencyController{},
```
yhakbar, line 42: I wonder if we should either make this private and only expose access via methods or initialize it here to avoid accidental nil pointer exceptions when it's not initialized.

internal/runner/configstack/dependency_controller.go
```diff
@@ -0,0 +1,372 @@
...
+		NotifyWhenDone: []*DependencyController{},
+	}
+}
+
+// Run a unit once all of its dependencies have finished executing.
```
yhakbar: NIT: We should stick to the idiomatic `runUnitWhenReady runs a unit once all of its dependencies have finished running` type godoc comments.

internal/runner/configstack/dependency_controller.go
```diff
@@ -0,0 +1,372 @@
...
+				ctrl.Runner.Unit.Logger.Errorf("Dependency %s of unit %s just finished with an error. Unit %s will have to return an error too. However, because of --queue-ignore-errors, unit %s will run anyway.", doneDependency.Runner.Unit.Path, ctrl.Runner.Unit.Path, ctrl.Runner.Unit.Path, ctrl.Runner.Unit.Path)
+			} else {
+				ctrl.Runner.Unit.Logger.Errorf("Dependency %s of unit %s just finished with an error. Unit %s will have to return an error too.", doneDependency.Runner.Unit.Path, ctrl.Runner.Unit.Path, ctrl.Runner.Unit.Path)
+
+				if opts.Experiments.Evaluate(experiment.Report) {
+					run, err := r.GetRun(ctrl.Runner.Unit.Path)
+					if err != nil {
+						if errors.Is(err, report.ErrRunNotFound) {
+							run, err = report.NewRun(ctrl.Runner.Unit.Path)
+							if err != nil {
```
yhakbar: NIT: This is a lot of nesting. I wonder if we can extract out some logic so that it's easier to read?

internal/runner/configstack/dependency_controller.go
```diff
@@ -0,0 +1,372 @@
...
+func (ctrl *DependencyController) waitForDependencies(opts *options.TerragruntOptions, r *report.Report) error {
+	ctrl.Runner.Unit.Logger.Debugf("Unit %s must wait for %d dependencies to finish", ctrl.Runner.Unit.Path, len(ctrl.Dependencies))
+
+	for len(ctrl.Dependencies) > 0 {
+		doneDependency := <-ctrl.DependencyDone
+		delete(ctrl.Dependencies, doneDependency.Runner.Unit.Path)
+
+		if doneDependency.Runner.Err != nil {
+			if ctrl.Runner.Unit.TerragruntOptions.IgnoreDependencyErrors {
+				ctrl.Runner.Unit.Logger.Errorf("Dependency %s of unit %s just finished with an error. Unit %s will have to return an error too. However, because of --queue-ignore-errors, unit %s will run anyway.", doneDependency.Runner.Unit.Path, ctrl.Runner.Unit.Path, ctrl.Runner.Unit.Path, ctrl.Runner.Unit.Path)
```
yhakbar: This isn't true.

```suggestion
				ctrl.Runner.Unit.Logger.Errorf("Dependency %s of unit %s just finished with an error. Normally, unit %s would exit early, however, because --queue-ignore-errors has been set, unit %s will run anyway.", doneDependency.Runner.Unit.Path, ctrl.Runner.Unit.Path, ctrl.Runner.Unit.Path, ctrl.Runner.Unit.Path)
```

yhakbar (commented):
I didn't have time to review it all, but I wanted to share some early feedback if it helps.

cli/commands/common/runall/runall.go
```diff
@@ -80,30 +82,30 @@ func Run(ctx context.Context, l log.Logger, opts *options.TerragruntOptions) err
...
 	switch opts.TerraformCommand {
 	case tf.CommandNameApply:
-		prompt = "Are you sure you want to run 'terragrunt apply' in each folder of the stack described above?"
+		prompt = "Are you sure you want to run 'terragrunt apply' in each unit of the run queue displayed described above?"
 	case tf.CommandNameDestroy:
-		prompt = "WARNING: Are you sure you want to run `terragrunt destroy` in each folder of the stack described above? There is no undo!"
+		prompt = "WARNING: Are you sure you want to run `terragrunt destroy` in each unit of the run queue displayed described above? There is no undo!"
 	case tf.CommandNameState:
-		prompt = "Are you sure you want to manipulate the state with `terragrunt state` in each folder of the stack described above? Note that absolute paths are shared, while relative paths will be relative to each working directory."
+		prompt = "Are you sure you want to manipulate the state with `terragrunt state` in each unit of the run queue displayed described above? Note that absolute paths are shared, while relative paths will be relative to each working directory."
```
yhakbar:
```suggestion
		prompt = "Are you sure you want to manipulate the state with `terragrunt state` in each unit of the run queue displayed above? Note that absolute paths are shared, while relative paths will be relative to each working directory."
```

docs/_docs/04_reference/02-cli-options.md
```diff
@@ -2450,6 +2451,22 @@ export TERRAGRUNT_FEATURE=int_feature_flag=123,bool_feature_flag=true,string_fea
...
+This is useful for enforcing strict failure handling in CI/CD pipelines or when you want to stop processing on the first error.
+
+Example:
+
+```bash
+terragrunt run --all apply --fail-fast
+```
+
+By default, Terragrunt will continue running other units even if one fails (unless other error-handling flags are set). Use `--fail-fast` to change this behavior and stop on the first failure.
+
```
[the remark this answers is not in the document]
yhakbar: Good point robot. Take note of this, @denis256

internal/runner/configstack/dependency_controller.go
```diff
@@ -0,0 +1,388 @@
...
+
+	for _, toNotify := range ctrl.NotifyWhenDone {
+		toNotify.DependencyDone <- ctrl
+	}
+}
+
+// RunningUnits is a map of unit path to DependencyController, representing the units that are currently running or
+type RunningUnits map[string]*DependencyController
+
+// toTerraformUnitGroups converts the RunningUnits into a slice of runbase.Units, where each slice represents a group of
```
yhakbar: a group of what?

internal/runner/runbase/unit_runner.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// Run a unit right now by executing the runTerragrunt command of its TerragruntOptions field.
+func (runner *UnitRunner) Run(ctx context.Context, rootOptions *options.TerragruntOptions, r *report.Report) error {
+	runner.Status = Running
+
+	if runner.Unit.AssumeAlreadyApplied {
+		runner.Unit.Logger.Debugf("Assuming unit %s has already been applied and skipping it", runner.Unit.Path)
+		return nil
+	} else {
```
yhakbar: We don't need this else.

internal/runner/runnerpool/pool.go
```diff
@@ -0,0 +1,77 @@
...
+			if p.q.Empty() {
+				l.Debugf("RunnerPool: queue is Empty, breaking loop")
+				break
+			}
+
+			l.Tracef("RunnerPool: no ready tasks, yielding (queue not Empty)")
+			runtime.Gosched()
+
+			continue
+		}
```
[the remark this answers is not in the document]
yhakbar: What do you think about this, @denis256 ? Worth addressing in a future PR?

internal/runner/runnerpool/pool.go
```diff
@@ -0,0 +1,77 @@
...
+// RunnerPool orchestrates concurrent execution over a DAG.
+type RunnerPool struct {
+	q           *DagQueue
+	runner      TaskRunner
+	concurrency int
+	failFast    bool
+}
+
+// New creates a new RunnerPool with the given units, runner function.
+func New(units []*runbase.Unit, r TaskRunner, maxConc int, failFast bool) *RunnerPool {
```
yhakbar: NIT: Consider making these options so that we can avoid super long parameter lists, and so that the default is easier to build.

internal/runner/runnerpool/pool.go
```diff
@@ -0,0 +1,93 @@
...
+				defer func() {
+					<-sem
+					wg.Done()
+				}()
+
+				ent.Result = p.runner(ctx, ent.Task)
+				p.q.MarkDone(ent, p.failFast)
+				signalReady() // notify that new tasks may be ready
+			}(e)
+		}
```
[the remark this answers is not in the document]
yhakbar: Good point. Consider this, @denis256 .

docs/_docs/04_reference/02-cli-options.md
```diff
@@ -2450,6 +2451,24 @@ export TERRAGRUNT_FEATURE=int_feature_flag=123,bool_feature_flag=true,string_fea
...
+When enabled, Terragrunt will fail the entire run as soon as any unit fails. This means that if any unit encounters an error during execution, all remaining units will be skipped and the run will exit with a failure status immediately.
+This is useful for enforcing strict failure handling in CI/CD pipelines or when you want to stop processing on the first error.
+
+Example:
+
+```bash
+terragrunt run --all apply --fail-fast
+```
+
+NOTE: NOTE: Currently, `--fail-fast` is honored only when the experimental _runner-pool_ execution engine is enabled (`--experiment runner-pool` or the matching env var). When using the legacy _configstack_ runner, the flag is ignored.
```
yhakbar: Double note?

yhakbar: Nobody is going to know what `configstack` runner means.

internal/runner/runnerpool/pool.go
yhakbar: Also, prefer `NewRunnerPool`, unless you're trying to make this generic to satisfy a shared interface.

internal/runner/runnerpool/pool_test.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	pool := runnerpool.New(units, runner, 2, false)
+	results := pool.Run(t.Context(), logger.CreateLogger())
+
+	for _, res := range results {
+		assert.Equal(t, 0, res.ExitCode)
+	}
+	// A must run before B and C
+	assert.Less(t, indexOf(executionOrder, "A"), indexOf(executionOrder, "B"))
```
yhakbar: I'm wondering here if we shouldn't be using reports instead of a new results construct.

internal/runner/runnerpool/queue.go
```diff
@@ -0,0 +1,212 @@
...
+
+var (
+	ErrSkippedFailFast       = errors.New("skipped due to fail-fast")
+	ErrSkippedAncestorFailed = errors.New("skipped due to ancestor failure")
+)
+
+// DagQueue holds Task nodes and tracks their dependency State in a
+// concurrency‑safe way. Each child keeps a `RemainingDeps` counter so we never
+// rescan the whole map when a parent finishes.
+type DagQueue struct {
```
yhakbar: Why aren't we using the queue package here? It's a generic problem, right? We need to be able to reliably display the queue as rendered during discovery, etc.

internal/runner/runnerpool/queue.go
```diff
@@ -0,0 +1,212 @@
...
+	for _, u := range units {
+		e := &Entry{Task: taskFromUnit(u), State: StatusPending}
+		q.Index[e.Task.ID()] = e
+		q.Ordered = append(q.Ordered, e)
+	}
+
+	// 2. wire dependencies and Dependents, set initial status / counters
+	for _, e := range q.Ordered {
+		parents := e.Task.Parents()
+		e.RemainingDeps = len(parents)
```
yhakbar: Why don't we just recalculate this instead of risking it becoming de-synchronized?

internal/runner/runnerpool/queue.go
```diff
@@ -0,0 +1,212 @@
...
+			if p, ok := q.Index[pid]; ok {
+				p.Dependents = append(p.Dependents, e)
+			}
+		}
+	}
+
+	return q
+}
+
+// GetReady returns up to max ready entries, respecting the original order.
```
yhakbar: There's no limit on this.

internal/runner/runnerpool/queue.go
```diff
@@ -0,0 +1,212 @@
...
+func (q *DagQueue) GetReady() []*Entry {
+	q.mu.Lock()
```
yhakbar: Do we need to lock the whole queue for this, or can we just lock the entry being set to ready?

internal/runner/runnerpool/queue.go
```diff
@@ -0,0 +1,212 @@
...
+			out = append(out, e)
+		}
+	}
+
+	return out
+}
+
+// updateEntryState updates the state of the entry based on its result.
+func (q *DagQueue) updateEntryState(e *Entry) {
+	if e.Result.ExitCode == 0 && e.Result.Err == nil {
```
yhakbar: Is that right? For a `run --all -- plan -detailed-exitcode`, we should expect non-zero exit codes, right?

internal/runner/runnerpool/queue.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+}
+
+// applyFailFast sets the state of all pending, blocked, or ready entries to StatusFailFast.
+func (q *DagQueue) applyFailFast() {
+	for _, n := range q.Ordered {
+		switch n.State {
+		case StatusPending, StatusBlocked, StatusReady:
+			n.State = StatusFailFast
+		case StatusRunning, StatusSucceeded, StatusFailed, StatusAncestorFailed, StatusFailFast:
```
yhakbar: Let's use the status of early exit to stay consistent with the report, please.

internal/runner/runnerpool/queue.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+}
+
+// updateDependents updates the states of the dependents of the entry based on its success or failure.
+func (q *DagQueue) updateDependents(e *Entry) {
+	success := e.State == StatusSucceeded
+
+	for _, child := range e.Dependents {
+		switch {
+		case success:
```
yhakbar: My brain hurts. Why is this a switch? Is there some magic Golang thing that makes 104 not result in a boolean here?

internal/runner/runnerpool/queue.go
```diff
@@ -0,0 +1,212 @@
...
+			child.RemainingDeps--
```
yhakbar: Again, I think not maintaining a counter would be better to avoid having to remember things like this.

internal/runner/runnerpool/queue.go
```diff
@@ -0,0 +1,212 @@
...
+	q.mu.Lock()
+	defer q.mu.Unlock()
+
+	if e.State != StatusRunning {
+		return // double call safeguard
+	}
+
+	q.updateEntryState(e)
+
+	if e.State == StatusFailed && failFast {
```
yhakbar: q has `failFast` on it, why are we passing it in here?

internal/runner/runnerpool/queue.go
```diff
@@ -0,0 +1,212 @@
...
+		q.applyFailFast()
+	}
+
+	q.updateDependents(e)
+}
+
+// Empty reports when no runnable or running tasks remain.
+func (q *DagQueue) Empty() bool {
+	q.mu.Lock()
```
yhakbar: Do we need a write lock for this? Do we need a lock at all? This shouldn't be accessed concurrently, should it? Won't the single manager be reading this?

internal/runner/runnerpool/runner.go
```diff
@@ -0,0 +1,318 @@
...
+
+// Runner implements the Stack interface for runner pool execution.
+type Runner struct {
+	Stack            *runbase.Stack
+	planErrorBuffers []bytes.Buffer // Store plan error buffers for summarizePlanAllErrors
+}
+
+// NewRunnerPoolStack creates a new stack from discovered units.
+func NewRunnerPoolStack(l log.Logger, terragruntOptions *options.TerragruntOptions, discovered discovery.DiscoveredConfigs, opts ...runbase.Option) (runbase.StackRunner, error) {
+	unitsMap := make(runbase.UnitsMap, len(discovered))
```
yhakbar: Is it a good idea to be double-tracking like this? Why not just have the orderedUnits slice?

internal/runner/runnerpool/runner.go
```diff
@@ -0,0 +1,318 @@
...
+	if opts.OutputFolder != "" {
+		for _, u := range r.Stack.Units {
+			planFile := u.OutputFile(l, opts)
+			if err := os.MkdirAll(filepath.Dir(planFile), os.ModePerm); err != nil {
+				return err
+			}
+		}
+	}
+
+	if util.ListContainsElement(config.TerraformCommandsNeedInput, terraformCmd) {
```
yhakbar, line 116: Can we not use `slices.Contains` here?

internal/runner/runnerpool/runner.go
```diff
@@ -0,0 +1,318 @@
...
+		}
+
+		return res
+	}
+
+	pool := New(r.Stack.Units, taskRun, opts.Parallelism, opts.FailFast)
+
+	results := pool.Run(ctx, l)
+
+	var allErrs []error
```
yhakbar: NIT: I think it's more idiomatic to name this `errs`

internal/runner/runnerpool/runner.go
```diff
@@ -0,0 +1,318 @@
...
+	return string(j), nil
+}
+
+func (r *Runner) ListStackDependentUnits() map[string][]string {
+	dependentUnits := make(map[string][]string)
+
+	for _, unit := range r.Stack.Units {
+		if len(unit.Dependencies) != 0 {
+			for _, dep := range unit.Dependencies {
+				dependentUnits[dep.Path] = util.RemoveDuplicatesFromList(append(dependentUnits[dep.Path], unit.Path))
```
yhakbar, line 230: We might prefer to use `slices.Sort().Compact()`

internal/runner/runnerpool/runner.go
```diff
@@ -0,0 +1,318 @@
...
+		if noUpdates {
+			break
+		}
+	}
+
+	return dependentUnits
+}
+
+// Sync the TerraformCliArgs for each unit in the stack to match the provided terragruntOptions struct.
+func (r *Runner) syncTerraformCliArgs(l log.Logger, opts *options.TerragruntOptions) {
```
yhakbar, line 260: Is there any way we can avoid increasing reliance on TerragruntOptions here? Totally fine if the answer is no, or no, we should tackle that later.

internal/runner/runnerpool/runner.go
```diff
@@ -0,0 +1,318 @@
...
+func (r *Runner) GetStack() *runbase.Stack {
+	return r.Stack
+}
+
+// SetTerragruntConfig sets the report for the stack.
+func (r *Runner) SetTerragruntConfig(config *config.TerragruntConfig) {
+	r.Stack.ChildTerragruntConfig = config
+}
+
+// SetParseOptions sets the report for the stack.
```
yhakbar:
```suggestion
// SetParseOptions sets the ParseOptions for the stack.
```

test/integration_aws_test.go
```diff
@@ -1393,7 +1393,7 @@ func TestAwsRunAllCommandPrompt(t *testing.T) {
 	err := helpers.RunTerragruntCommand(t, "terragrunt run --all apply --working-dir "+environmentPath, &stdout, &stderr)
 	helpers.LogBufferContentsLineByLine(t, stdout, "stdout")
 	helpers.LogBufferContentsLineByLine(t, stderr, "stderr")
-	assert.Contains(t, stderr.String(), "Are you sure you want to run 'terragrunt apply' in each folder of the stack described above? (y/n)")
+	assert.Contains(t, stderr.String(), "Are you sure you want to run 'terragrunt apply' in each unit of the run queue displayed described above? (y/n)")
```
yhakbar:
```suggestion
	assert.Contains(t, stderr.String(), "Are you sure you want to run 'terragrunt apply' in each unit of the run queue displayed above? (y/n)")
```

yhakbar (changes requested):
I don't want to block this for too long, so if we could have a talk about it, and address the typos in the docs, and log messages, I'm happy to have us merge this, then iterate.

docs-starlight/src/data/flags/fail-fast.mdx
```diff
@@ -0,0 +1,17 @@
...
+---
+
+import { Aside } from '@astrojs/starlight/components';
+
+When enabled, Terragrunt will fail the entire run as soon as any unit fails. If any unit encounters an error during execution, all remaining units will be skipped and the run will exit with a failure status immediately.
+
+This is useful for enforcing strict failure handling in CI/CD pipelines or when you want to stop processing on the first error.
+
+<Aside type="warning">
+Currently, `--fail-fast` is honored only when the experimental _runner-pool_ is enabled.
```
yhakbar:
```suggestion
Currently, `--fail-fast` is honored only when the `runner-pool` experiment is enabled.
```

NIT

yhakbar:
```suggestion
Currently, `--fail-fast` is honored only when the `runner-pool` experiment is enabled.
```

docs/_docs/04_reference/02-cli-options.md
```diff
@@ -2461,6 +2462,22 @@ export TERRAGRUNT_FEATURE=int_feature_flag=123,bool_feature_flag=true,string_fea
...
+When enabled, Terragrunt will fail the entire run as soon as any unit fails. This means that if any unit encounters an error during execution, all remaining units will be skipped and the run will exit with a failure status immediately.
+This is useful for enforcing strict failure handling in CI/CD pipelines or when you want to stop processing on the first error.
+
+Example:
+
+```bash
+terragrunt run --all apply --fail-fast
+```
+
+NOTE: NOTE: Currently, `--fail-fast` is honored only when the experimental _runner-pool_ is enabled.
```
yhakbar:
```suggestion
NOTE: Currently, `--fail-fast` is honored only when the `runner-pool` experiment is enabled.
```

internal/queue/queue.go
```diff
@@ -30,18 +30,34 @@ import (
 	"github.com/gruntwork-io/terragrunt/internal/discovery"
 )
 
+// Entry represents a node in the execution queue/DAG. Each Entry corresponds to a single Terragrunt configuration
+// and tracks its execution status and relationships to other entries in the queue.
 type Entry struct {
+	// Config is the Terragrunt configuration associated with this entry. It contains all metadata about the module,
```
yhakbar:
```suggestion
	// Config is the Terragrunt configuration associated with this entry. It contains all metadata about the unit/stack,
```

internal/queue/queue.go
```diff
@@ -30,18 +30,34 @@ import (
...
+	// including its path, dependencies, and discovery context (such as the command being run).
 	Config *discovery.DiscoveredConfig
+
+	// Dependents is a list of entries that directly depend on this entry. If this entry fails (and fail-fast is not enabled),
```
yhakbar: Is this more accurate?

```
    // Dependents is a list of entries that directly depend on this entry. 
    // 
    // By default:
    // - "Up" (e.g. `apply`) commands won't run until dependencies run successfully.
    // - "Down" (e.g. `destroy`) commands won't run until dependents run successfully.
    //
    // Whenever a unit completes its run, all of its dependents (for "up" commands) or dependencies (for "down" commands) will receive a signal to start their run (or early exit their runs on failure of a dependent/dependency).
```

`fail-fast` isn't really related to this, as it has to do with initiating an early exit for _all_ units, even those that aren't dependencies, right?

NOTE: We should be sure that we're respecting `--queue-ignore-dag-order`.

internal/queue/queue.go
```diff
@@ -30,18 +30,34 @@ import (
...
+	// all its dependents will also be marked as failed. This field is populated during queue construction and is used to
+	// propagate status changes and determine execution order for destroy operations.
+	Dependents []*Entry
```
yhakbar: I just realized that we're storing dependents here. Why? We can dynamically recompute this based on `DiscoveredConfig.Dependencies`, right? I think it's a bad sign that we might have two sources of truth here, and we need to be careful with that. It's likely that one of these two are going to get out of sync at some point.

I think, ideally, determining dependents should just be computed as necessary by looking at dependencies. If we have measurable slowdown because of this, we can memoize the results or optimize the computation, before going down this road and computing it ahead of time and storing two copies of the same data.

internal/queue/queue.go
```diff
@@ -30,18 +30,34 @@ import (
...
 type Status byte
 
 const (
 	StatusPending Status = iota
 	StatusBlocked
 	StatusUnsorted
 	StatusReady
+	StatusRunning
+	StatusSucceeded
+	StatusFailed
```
yhakbar, line 55: I think it makes sense to have a different status for "failed" vs "early exit". Both are terminal statuses for an overall failure, but they mean something different semantically.

internal/queue/queue.go
```diff
@@ -127,7 +143,9 @@ func (e *Entry) IsUp() bool {
 }
 
 type Queue struct {
-	Entries Entries
+	Index    map[string]*Entry
```
yhakbar: Why do we have both a map and a slice for this? I assume it's for performance, but do we actually know that it's going to end up being faster to have both? If the reason you want to do this is performance, I want to challenge you to test it and prove it if you have time.

internal/queue/queue.go
```diff
@@ -173,22 +191,37 @@ func NewQueue(discovered discovery.DiscoveredConfigs) (*Queue, error) {
...
+		entry := &Entry{
+			Config:     cfg,
+			Status:     StatusPending,
+			Dependents: []*Entry{},
+		}
+		entries = append(entries, entry)
+		index[cfg.Path] = entry
+	}
+
+	// Populate Dependents for each entry
```
yhakbar: See note above about concerns that we might not need this.

internal/queue/queue.go
```diff
@@ -247,3 +280,104 @@ func NewQueue(discovered discovery.DiscoveredConfigs) (*Queue, error) {
 
 	return q, errors.New("cycle detected during queue construction")
 }
+
+// GetReadyWithDependencies returns all entries that are ready to run and have all dependencies completed(or no dependencies).
```
yhakbar:
```suggestion
// GetReadyWithDependencies returns all entries that are ready to run and have all dependencies completed (or no dependencies).
```

internal/queue/queue.go
```diff
@@ -247,3 +280,104 @@ func NewQueue(discovered discovery.DiscoveredConfigs) (*Queue, error) {
...
+				out = append(out, e)
+			}
+		}
+	}
+
+	return out
+}
+
+// SetStatus sets the status of the given entry, handles fail-fast logic if enabled, and updates the statuses of dependents accordingly.
+// This method should be used for all status transitions. If fail-fast is enabled and the entry fails, all not-yet-started tasks are marked as failed.
```
yhakbar:
```suggestion
// This method should be used for all status transitions. If fail-fast is enabled and the entry fails, all not-yet-started entries are marked as failed.
```

internal/queue/queue.go
```diff
@@ -247,3 +280,104 @@ func NewQueue(discovered discovery.DiscoveredConfigs) (*Queue, error) {
...
+// It also unblocks children or marks them as failed/ancestor-failed as appropriate.
```
yhakbar: NIT: Consider whether it's better to have that be a different method. Something like "PropagateFailure". Maybe an option on a `StatusSetter` or something. Nothing about the method signature here indicates that these side-effects are going to occur. You have to read the comment to know that they'll happen.

internal/queue/queue.go
```diff
@@ -247,3 +280,104 @@ func NewQueue(discovered discovery.DiscoveredConfigs) (*Queue, error) {
...
+func (q *Queue) SetStatus(e *Entry, status Status) {
+	e.Status = status
+
+	// If this entry failed, recursively fail all dependents using the Dependents field
```
yhakbar: No it doesn't, right?

internal/queue/queue.go
```diff
@@ -247,3 +280,104 @@ func NewQueue(discovered discovery.DiscoveredConfigs) (*Queue, error) {
...
+	if status == StatusFailed {
+		// If fail-fast is enabled, mark all not-yet-started tasks (Pending, Blocked, Ready) as Failed to prevent further execution.
+		if q.FailFast {
+			for _, n := range q.Entries {
+				if isTerminalOrRunning(n.Status) {
+					continue
+				}
+
+				n.Status = StatusFailed
```
yhakbar: Should this be a call to `q.SetStatus(n, StatusFailed)`? That would make it recursive, I believe. See my comment above about whether this should change to "Propagate Failure"

Also, see my comment above about this potentially being better off as `StatusEarlyExit` to stay consistent with reporting.

internal/queue/queue.go
```diff
@@ -247,3 +280,104 @@ func NewQueue(discovered discovery.DiscoveredConfigs) (*Queue, error) {
...
+
+			return
+		}
+
+		for _, depEntry := range e.Dependents {
+			if isTerminalOrRunning(depEntry.Status) {
+				continue
+			}
+
+			depEntry.Status = StatusFailed
```
yhakbar: Should this be a call to `q.SetStatus(n, StatusFailed)`? That would make it recursive, I believe. See my comment above about whether this should change to "Propagate Failure".

Also, see my comment above about this potentially being better off as `StatusEarlyExit` to stay consistent with reporting.

internal/queue/queue.go
```diff
@@ -247,3 +280,104 @@ func NewQueue(discovered discovery.DiscoveredConfigs) (*Queue, error) {
...
+		}
+	}
+}
+
+// Empty reports when no runnable or running tasks remain.
+func (q *Queue) Empty() bool {
```
yhakbar: NIT: Consider `Finished` for this, as it doesn't actually check if the queue is empty by looking at length.

internal/queue/queue.go
yhakbar: NIT

internal/queue/queue.go
yhakbar: NIT

docs-starlight/src/data/flags/fail-fast.mdx
yhakbar: NIT

docs/_docs/04_reference/02-cli-options.md
yhakbar: NIT

internal/queue/queue.go
yhakbar: NIT

internal/queue/queue.go
yhakbar: NIT

internal/queue/queue.go
yhakbar: NIT

internal/queue/queue.go
yhakbar: NIT

internal/queue/queue.go
yhakbar: NIT

internal/queue/queue.go
yhakbar: NIT

yhakbar (changes requested):

docs/_docs/04_reference/02-cli-options.md
```diff
@@ -2461,6 +2462,22 @@ export TERRAGRUNT_FEATURE=int_feature_flag=123,bool_feature_flag=true,string_fea
...
+When enabled, Terragrunt will fail the entire run as soon as any unit fails. This means that if any unit encounters an error during execution, all remaining units will be skipped and the run will exit with a failure status immediately.
+This is useful for enforcing strict failure handling in CI/CD pipelines or when you want to stop processing on the first error.
+
+Example:
+
+```bash
+terragrunt run --all apply --fail-fast
+```
+
+NOTE: NOTE: Currently, `--fail-fast` is honored only when the `runner-pool` experiment is enabled.
```
yhakbar, line 2479:
```suggestion
NOTE: Currently, `--fail-fast` is honored only when the `runner-pool` experiment is enabled.
```

internal/queue/queue.go
```diff
@@ -247,3 +275,115 @@ func NewQueue(discovered discovery.DiscoveredConfigs) (*Queue, error) {
...
+			continue
+		}
+
+		for _, dep := range entry.Config.Dependencies {
+			if dep.Path == e.Config.Path {
+				if isTerminalOrRunning(entry.Status) {
+					continue
+				}
+
+				entry.Status = StatusFailed
```
yhakbar, line 344: Aren't these still early exit?

internal/queue/queue.go
```diff
@@ -247,3 +275,115 @@ func NewQueue(discovered discovery.DiscoveredConfigs) (*Queue, error) {
...
+			}
+		}
+	}
+
+	return out
+}
+
+// FailEntry marks the given entry as failed, handles fail-fast logic if enabled, and updates the statuses of dependents accordingly.
+// This method should be used only for failure transitions. For other status changes, set the Status field directly.
+func (q *Queue) FailEntry(e *Entry) {
```
yhakbar, line 316: Why isn't this trying to be recursive?

internal/queue/queue.go
```diff
@@ -247,3 +275,115 @@ func NewQueue(discovered discovery.DiscoveredConfigs) (*Queue, error) {
...
+			count++
+		}
+	}
+
+	return count
+}
+
+// isTerminal returns true if the status is terminal.
+func isTerminal(status Status) bool {
+	return status == StatusSucceeded || status == StatusFailed || status == StatusEarlyExit
```
yhakbar, line 383: NIT: Prefer switch here?

internal/queue/queue.go
```diff
@@ -247,3 +275,115 @@ func NewQueue(discovered discovery.DiscoveredConfigs) (*Queue, error) {
...
+}
+
+// isTerminalOrRunning returns true if the status is terminal.
+func isTerminalOrRunning(status Status) bool {
+	return status == StatusSucceeded || status == StatusFailed || status == StatusRunning || status == StatusEarlyExit
```
yhakbar, line 388: NIT: Prefer switch here?

internal/runner/common/errors.go
```diff
@@ -0,0 +1,83 @@
...
+	Err        error
+}
+
+func (err ProcessingUnitDependencyError) Error() string {
+	return errors.Errorf("Cannot process unit %s because one of its dependencies, %s, finished with an error: %s", err.Unit, err.Dependency, err.Err).Error()
+}
+
+func (err ProcessingUnitDependencyError) ExitStatus() (int, error) {
+	if exitCode, err := util.GetExitCode(err.Err); err == nil {
+		return exitCode, nil
```
yhakbar, line 66: Why don't we get the original error returned here, but do get it in line 69? Shouldn't we either get `err` in both or stop shadowing `err` in line 65, and return that alternate error on line 69?

internal/runner/common/unit.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+
+	for _, dependencyPath := range unit.Config.Dependencies.Paths {
+		dependencyUnitPath, err := util.CanonicalPath(dependencyPath, unit.Path)
+		if err != nil {
+			// TODO: Remove lint suppression
+			return dependencies, nil //nolint:nilerr
+		}
+
+		if files.FileExists(dependencyUnitPath) && !files.IsDir(dependencyUnitPath) {
```
yhakbar, line 133: NIT: Isn't this going to end up stat-ing the file twice? Do we need to do that?

internal/runner/common/unit.go
```diff
@@ -0,0 +1,299 @@
...
+	out := UnitsMap{}
+
+	maps.Copy(out, externalDependencies)
+
+	maps.Copy(out, unitsMap)
+
+	return out
+}
+
+// FindByPath returns the unit that matches the given path, or nil if no such unit exists in the map.
```
yhakbar, line 166: Don't you normally have to check the ok?

internal/runner/common/unit.go
```diff
@@ -0,0 +1,299 @@
...
+
+		dependencies = append(dependencies, dependencyUnit)
+	}
+
+	return dependencies, nil
+}
+
+// MergeMaps the given external dependencies into the given map of units if those dependencies aren't already in the
+// units map
+func (unitsMap UnitsMap) MergeMaps(externalDependencies UnitsMap) UnitsMap {
```
yhakbar, line 156: I don't understand why we have this. Discovery will already return external dependencies. Why are we creating a new map here?

internal/runner/common/unit.go
```diff
@@ -0,0 +1,299 @@
...
+}
+
+// FindByPath returns the unit that matches the given path, or nil if no such unit exists in the map.
+func (unitsMap UnitsMap) FindByPath(path string) *Unit {
+	return unitsMap[path]
+}
+
+// CrossLinkDependencies Go through each unit in the given map and cross-link its dependencies to the other units in that same map. If
+// a dependency is referenced that is not in the given map, return an error.
+func (unitsMap UnitsMap) CrossLinkDependencies(canonicalTerragruntConfigPaths []string) (Units, error) {
```
yhakbar, line 173: Do we need to do this? It's already done during Discovery.

internal/runner/common/unit.go
```diff
@@ -0,0 +1,299 @@
...
+		unit.Dependencies = dependencies
+		units = append(units, unit)
+	}
+
+	return units, nil
+}
+
+// WriteDot is used to emit a GraphViz compatible definition
+// for a directed graph. It can be used to dump a .dot file.
+// This is a similar implementation to terraform's digraph https://github.com/hashicorp/terraform/blob/master/digraph/graphviz.go
```
yhakbar, line 195: We should update this link to 1.5.7. We looked at the file before then, but we don't want to accidentally show contributors BSL licensed code.

internal/runner/common/unit.go
```diff
@@ -0,0 +1,299 @@
...
+			if err != nil {
+				return errors.New(err)
+			}
+		}
+	}
+
+	return nil
+}
+
+// CheckForCycles checks for dependency cycles in the given list of units and return an error if one is found.
```
yhakbar, line 242: Do we always need to do this? We check for cycles during discovery. I believe right now we just remove cycles, but we could explicitly error instead.

internal/runner/common/unit_runner.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/gruntwork-io/terragrunt/internal/experiment"
+	"github.com/gruntwork-io/terragrunt/internal/report"
+	"github.com/gruntwork-io/terragrunt/options"
+	"github.com/gruntwork-io/terragrunt/tf"
+	"github.com/gruntwork-io/terragrunt/util"
+)
+
+// UnitStatus represents the status of a unit that we are
+// trying to apply or destroy as part of the run --all apply or run --all destroy command
+type UnitStatus int
```
yhakbar, line 18: Why do we have this as a different status than `Entry.Status`?

internal/runner/common/unit_runner.go
```diff
@@ -0,0 +1,112 @@
...
+
+const (
+	Waiting UnitStatus = iota
```
yhakbar, line 21: We use `StatusPending` in queue entries. Can we use one term for both to avoid complicating our language?

internal/runner/common/unit_runner.go
```diff
@@ -0,0 +1,112 @@
...
+		if err := r.AddRun(run); err != nil {
+			return err
+		}
+	}
+
+	return opts.RunTerragrunt(ctx, runner.Unit.Logger, opts, r)
+}
+
+// Run a unit right now by executing the runTerragrunt command of its TerragruntOptions field.
+func (runner *UnitRunner) Run(ctx context.Context, rootOptions *options.TerragruntOptions, r *report.Report) error {
```
yhakbar, line 68: Why is this `rootOptions` instead of `opts`?

internal/runner/common/unit_runner.go
```diff
@@ -0,0 +1,112 @@
...
+		if err != nil {
+			return err
+		}
+
+		if err := r.AddRun(run); err != nil {
+			return err
+		}
+	}
+
+	return opts.RunTerragrunt(ctx, runner.Unit.Logger, opts, r)
```
yhakbar, line 64: Do we still need this in the new model? Every queue entry knows what it's going to run, right? Do we need to maintain that tight dependency on this anonymous function here?

internal/runner/runner.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/gruntwork-io/terragrunt/pkg/log"
+
+	"github.com/gruntwork-io/terragrunt/options"
+)
+
+// FindStackInSubfolders finds all the Terraform modules in the subfolders of the working directory of the given TerragruntOptions and
+// assemble them into a Stack object that can be applied or destroyed in a single command
+func FindStackInSubfolders(ctx context.Context, l log.Logger, terragruntOptions *options.TerragruntOptions, opts ...common.Option) (common.StackRunner, error) {
+	if terragruntOptions.Experiments.Evaluate(experiment.RunnerPool) {
+		l.Infof("Using RunnerPoolStackBuilder to build stack for %s", terragruntOptions.WorkingDir)
```
yhakbar, line 27: Confusing log message here. Nothing is called `RunnerPoolStackBuilder` here.

internal/runner/runnerpool/dag_runner.go
yhakbar, line 1: NIT: I think the recommendation from the go team is to just call files like this `dagrunner.go`, etc.

internal/runner/runnerpool/dag_runner.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// NewDAGRunner creates a new DAGRunner with the given options and a pre-built queue.
+func NewDAGRunner(q *queue.Queue, units []*common.Unit, opts ...DAGRunnerOption) *DAGRunner {
+	dr := &DAGRunner{
+		q:       q,
+		readyCh: make(chan struct{}, 1), // buffered to avoid blocking
+	}
+	// Build unitsMap from units slice
+	unitsMap := make(map[string]*common.Unit)
```
yhakbar, line 56: Do we really benefit from these conversions to maps from slices?

internal/runner/runnerpool/dag_runner.go
```diff
@@ -0,0 +1,179 @@
...
+		case <-dr.readyCh:
+		case <-ctx.Done():
+			wg.Wait()
+			return nil
+		}
+	}
+
+	wg.Wait()
+
+	// Preserve original order
```
yhakbar, line 169: Oh this is interesting... are we doing this just to make testing easier?

internal/runner/runnerpool/dag_runner.go
```diff
@@ -0,0 +1,179 @@
...
+// (order preserved).  The function:
+//
+//   - never blocks forever – if progress is impossible it bails out;
+//   - is data-race free (results map is mutex-protected);
+//   - needs no special-casing for fail-fast.
+func (dr *DAGRunner) Run(ctx context.Context, l log.Logger) []RunResult {
+	var (
+		wg      sync.WaitGroup
+		sem     = make(chan struct{}, dr.concurrency)
+		results = xsync.NewMapOf[string, RunResult]()
```
yhakbar, line 93: Why are we tracking results, btw? Could we not read the report?

internal/runner/runnerpool/runner.go
```diff
@@ -0,0 +1,336 @@
...
+
+// NewRunnerPoolStack creates a new stack from discovered units.
+func NewRunnerPoolStack(l log.Logger, terragruntOptions *options.TerragruntOptions, discovered discovery.DiscoveredConfigs, opts ...common.Option) (common.StackRunner, error) {
+	q, queueErr := queue.NewQueue(discovered)
+	if queueErr != nil {
+		return nil, queueErr
+	}
+
+	unitsMap := make(common.UnitsMap, len(discovered))
+	orderedUnits := make(common.Units, 0, len(discovered))
```
yhakbar, line 45: We keep duplicating data between maps and slices. Is that really a good idea?

internal/runner/runnerpool/runner.go
```diff
@@ -0,0 +1,336 @@
...
+	for _, cfg := range discovered {
+		configPath := config.GetDefaultConfigPath(cfg.Path)
+
+		if cfg.Parsed == nil {
+			// Skip configurations that could not be parsed
+			l.Warnf("Skipping unit at %s due to parse error", cfg.Path)
+			continue
+		}
+
+		modLogger, modOpts, err := terragruntOptions.CloneWithConfigPath(l, configPath)
```
yhakbar, line 65:
```suggestion
		unitLogger, unitOpts, err := terragruntOptions.CloneWithConfigPath(l, configPath)
```

internal/runner/runnerpool/runner.go
```diff
@@ -0,0 +1,336 @@
...
+			Logger:            modLogger,
+			Path:              cfg.Path,
+			Config:            *cfg.Parsed,
+		}
+
+		orderedUnits = append(orderedUnits, mod)
+		unitsMap[cfg.Path] = mod
+	}
+
+	// cross-link dependencies units based on the discovered configurations
```
yhakbar, line 83: Why do we do this so many times? Shouldn't this already be done during discovery?

internal/runner/runnerpool/runner.go
```diff
@@ -0,0 +1,336 @@
...
+	if planDefer {
+		defer r.summarizePlanAllErrors(l, r.planErrorBuffers)
+	}
+
+	taskRun := func(ctx context.Context, u *common.Unit) (int, error) {
+		unitRunner := common.NewUnitRunner(u)
+
+		err := unitRunner.Run(ctx, u.TerragruntOptions, r.Stack.Report)
+		if err != nil {
+			return 1, err
```
yhakbar, line 143: Why do we do this? That's not really the exit code, right?

internal/runner/runnerpool/runner.go
```diff
@@ -0,0 +1,336 @@
...
+		}
+
+		return 0, nil
```
yhakbar, line 146: Why do we do this? That's not really the exit code, right?

PR #4439, opened 2025-06-18
perf: Improving CAS performance a bit

yhakbar:
## Description

Optimizing performance a bit in the `cas` package. The main win here is to add the `-r` flag to the `ls-tree` `git` operation that we perform so that Git does the recursion instead of doing it in Go.

image

image

I bet we can shave off more time.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Updated `cas` package to address some performance issues.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved accuracy of benchmark timing and added error handling in benchmark tests.
  - Enhanced test reliability by adding explicit error checks when initializing the Git runner in all relevant tests.

- **New Features**
  - Added a method to recursively list all blobs in a Git tree in a single operation.
  - The application now automatically detects the Git executable path for all Git operations.

- **Refactor**
  - Simplified and optimized tree and blob storage logic for improved performance and clarity.
  - Refactored tree linking to use sequential processing for greater reliability and maintainability.
  - Streamlined URL transformation logic for Git SSH URLs.

- **Chores**
  - Updated dependency versions and added new indirect dependencies.
  - Added a .gitignore rule to exclude test files in the internal/cas directory.


internal/cas/git.go
```diff
@@ -18,20 +18,30 @@ const (
 
 // GitRunner handles git command execution
 type GitRunner struct {
+	GitPath string
```
denis256, line 21: Clone(), depth argument seems to be unused, it is only checked for > 0

internal/cas/tree.go
```diff
@@ -101,96 +97,41 @@ func ParseTree(output, path string) (*Tree, error) {
...
+		entryPath := filepath.Join(targetDir, entry.Path)
+		if err := os.MkdirAll(filepath.Dir(entryPath), DefaultDirPerms); err != nil {
+			return wrapError("mkdir_all", entryPath, err)
+		}
 
-			select {
-			case <-ctx.Done():
-				return
-			default:
+		switch entry.Type {
```
denis256, line 115: No handling of unknown types

internal/cas/git.go
```diff
@@ -206,6 +216,32 @@ func (g *GitRunner) LsTree(ctx context.Context, reference, path string) (*Tree,
 	return ParseTree(stdout.String(), path)
 }
 
+// LsTreeRecursive runs git ls-tree -r and returns all blobs recursively
+// This eliminates the need for multiple separate ls-tree calls on subtrees
+func (g *GitRunner) LsTreeRecursive(ctx context.Context, reference, path string) (*Tree, error) {
```
denis256, line 221: Do we need to validate inputs for reference, path?

denis256 (approved):

internal/cas/tree.go
yhakbar: What do you think we should do? I could put a default that logs the type, but I don't think we really care about any other type.

internal/cas/git.go
yhakbar: What kind of validations would you want?

internal/cas/tree.go
denis256: Since the type is loaded from files, we need some basic validation for empty or  invalid value, most probably we need to log unknown types

Most probably we also need to have version system of CAS files.

internal/cas/git.go
denis256: At least empty/nil reference,path

PR #4502, opened 2025-07-02
fix: Adjusting `awshelper`

yhakbar:
## Description

Moves `awshelper` into an `internal` package, and upgrading all usage of the AWS SDK to v2.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Updated AWS SDK to v2 and moved `awshelper` into an `internal` package.


awshelper/config.go
```diff
@@ -1,428 +0,0 @@
...
-	SessionName             string
-	S3ForcePathStyle        bool
-	DisableComputeChecksums bool
-}
-
-// addUserAgent - Add terragrunt version to the user agent for AWS API calls.
-var addUserAgent = request.NamedHandler{
-	Name: "terragrunt.UserAgentHandler",
-	Fn: request.MakeAddToUserAgentHandler(
-		"terragrunt", version.GetVersion()),
```
denis256, line 46: How will the user agent be passed to AWS?

From what I read, it can be set with AppID

https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws

denis256 (dismissed):

internal/awshelper/config.go
```diff
@@ -0,0 +1,469 @@
...
+		if opts != nil && opts.IAMRoleOptions.RoleARN != "" {
+			if opts.IAMRoleOptions.WebIdentityToken != "" {
+				l.Debugf("Assuming role %s using WebIdentity token", opts.IAMRoleOptions.RoleARN)
+				cfg.Credentials = getWebIdentityCredentialsFromIAMRoleOptions(cfg, opts.IAMRoleOptions)
+			} else {
+				l.Debugf("Assuming role %s", opts.IAMRoleOptions.RoleARN)
+				cfg.Credentials = getSTSCredentialsFromIAMRoleOptions(cfg, opts.IAMRoleOptions, "")
+			}
+		} else if opts != nil {
+			if creds := getCredentialsFromEnvs(opts); creds != nil {
```
wakeful: Do we really need this?
The code around line 243:
```go
		cfg, err = config.LoadDefaultConfig(ctx)
```
should already cover the case of loading credentials from environment variables

internal/awshelper/config.go
```diff
@@ -0,0 +1,469 @@
...
+// CreateAwsConfigFromConfig returns an AWS config object for the given config region (required), profile name (optional), and IAM role to assume
+// (optional), ensuring that the credentials are available.
+func CreateAwsConfigFromConfig(ctx context.Context, awsCfg *AwsSessionConfig, opts *options.TerragruntOptions) (aws.Config, error) {
+	var configOptions []func(*config.LoadOptions) error
+
+	// Set region
+	if awsCfg.Region != "" {
+		configOptions = append(configOptions, config.WithRegion(awsCfg.Region))
+	} else {
+		configOptions = append(configOptions, config.WithRegion("us-east-1"))
```
wakeful, line 52: The default region is already being set in the `CreateAwsConfig` function
```go
		// Ensure a region is set - fallback to us-east-1 if none is configured
		if cfg.Region == "" {
			l.Debugf("No region configured, using default region us-east-1")
			cfg.Region = "us-east-1"
		}
```
comment for this function suggests that instead of performing a fallback, we should return an error

yhakbar: I added this because my changes introduced a regression in our test suite without it. I do think we need it, or another approach that's better to handle the fallback when no region is defined to avoid regression.

internal/awshelper/config.go
yhakbar: It's a good question. I'll see if I can get rid of it.

internal/awshelper/config.go
yhakbar: Can confirm it's necessary:
https://github.com/gruntwork-io/terragrunt/actions/runs/16523674724/job/46731375349#step:10:493

awshelper/config.go
yhakbar: Added.

adamcharnock:
Hi @yhakbar, I suspect these changes may have introduced the regression I have detailed in issue #4732

PR #4533, opened 2025-07-11
perf: speed up `EncodeSourceVersion` by multiple optimizations

healthy-pod:
## Description

1. Use `WalkDir` to only call `os.Stat` when needed (i.e. when access to file modification time is needed). Previously, it was called on all entries. This cut `EncodeSourceVersion` time in one of our private workspaces from 43s to 34s.

2. Skip walking cache directories completely by returning `filepath.SkipDir` when hitting one. Previously, we just returned nil which meant we still had to walk everything in the file tree.
This cut `EncodeSourceVersion` time in one of our private workspaces from 34s to ~nil.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

Updated `EncodeSourceVersion` to make less `stat` calls and skip walking cache dirs.


denis256:
Hi,
Consider running `make run-lint` to run lint checks locally:
```
level=info msg="File cache stats: 1 entries of total size 14.2KiB"
level=info msg="Memory: 857 samples, avg is 111.6MB, max is 1052.8MB"
level=info msg="Execution took 1m25.991943112s"
tf/source.go:114:5: assignments should only be cuddled with other assignments (wsl)
				fileModified := fileInfo.ModTime().UnixMicro()
				^
1 issues:
* wsl: 1
make: *** [Makefile:[42](https://github.com/gruntwork-io/terragrunt/actions/runs/16296496947/job/46019409624?pr=4533#step:8:43): run-lint] Error 1
```

Looks like TestInitSkipCache is failing
```
2025-07-15T14:49:42.0137091Z --- FAIL: TestInitSkipCache (23.88s)
```
[FAILTestInitSkipCache.txt](https://github.com/user-attachments/files/21236343/FAILTestInitSkipCache.txt)

healthy-pod:
@denis256 @yhakbar @ThisGuyCodes Can you take a look when you get a chance? Thanks!

yhakbar:
Hey @healthy-pod,

This looks good, and if CI passes, I'll merge it.

This PR does introduce an inconsistency, though, along with highlighting a bug in the logic used for matching `util.TerragruntCacheDir`.

You only updated the `WalkDir` branch. You didn't also update the `WalkDirWithSymlinks` branch. The two can likely be consolidated into one call to a function stored in a variable, determined by `src.WalkDirWithSymlinks` so that we can't accidentally only update one branch instead of the other.

In addition, the existing logic for `strings.Contains` is fragile. We shouldn't match on the path containing any reference to the cache dir, or we'll also match on directories named `not_.terragrunt-cache_dir`, etc. A more accurate match would be to check `d.Name()` to ensure that we're just matching against the basename of the path.

There's also likely another optimization that could be made here to skip over `.git` directories, etc. There's [logic in `discovery`](https://github.com/gruntwork-io/terragrunt/blob/main/internal/discovery/helpers.go#L126) that does that. We can probably refactor the logic used in these two paths to have consistent logic for determining when we can early exit from a walk.

Are you interested in taking any/all of this on?

healthy-pod:
@yhakbar Thanks for merging! Do you think the `no-hidden` flag should go under global? If not, which commands end up calling `EncodeSourceVersion` and will depend on its value?

yhakbar:
Hey @healthy-pod , no the `--no-hidden` flag is only used by discovery commands (`find` and `list`), I believe.

I think I'd need to do some tracing to find out which commands end up calling `EncodeSourceVersion`. Why do you ask?

healthy-pod:
@yhakbar I plan to move [this](https://github.com/gruntwork-io/terragrunt/blob/main/internal/discovery/phase_filesystem.go#L100-L113) to a shared place like `internal/util` then call it from `internal/discovery` like now and also inside `EncodeSourceVersion` under `internal/tf`. I am asking about the commands that end up calling `EncodeSourceVersion` to make sure I configure `no-hidden` flag for them. Another option is configuring it globally under `internal/cli/flags/global`.

yhakbar:
Ah, I see. `--no-hidden` is really a concern for discovery, so I don't think it's necessary to refactor out the logic for all of the private `FilesystemPhase.skipDirIfIgnorable` method into a separate location. It should be viable to only migrate the logic for the private helper function `skipDirIfIgnorable` into a separate location, so that callers can reliably skip directories that never have meaningful content for any walk in Terragrunt.

I don't think it makes sense for `EncodeSourceVersion` to conditionally ignore _all_ hidden directories, and it would probably be difficult for an end user to opt in to something like that, as it's a fairly low-level operation.

healthy-pod:
@yhakbar Just opened https://github.com/gruntwork-io/terragrunt/pull/5559

PR #4540, opened 2025-07-16
allow overriding of boilerplate hooks and shell flags in catalog config

### LINKED ISSUE #3659: allow overriding of boilerplate hooks and shell flags in catalog config
## Describe the enhancement

It would seem that when running `terragrunt scaffold` (through `terragrunt catalog` ) boilerplate hooks and shell scripts are disabled. That is, it seems the following flags are added to the boilerplate call:
```
--disable-hooks
--disable-shell
```

This doesn't seem to be documented anywhere, which would be the first suggestion for enhancement. The second suggestion would be to let the user override these in the catalog config.

```
catalog {
  disable_hooks: false,  # defaults to true
  disable_shell: false, # defaults to true
  urls = [
    "https://github.com/gruntwork-io/terragrunt-infrastructure-modules-example",
    "https://github.com/gruntwork-io/terraform-aws-utilities",
    "https://github.com/gruntwork-io/terraform-kubernetes-namespace"
  ]
}
```

## Additional context

I understand the reasoning for disabling both hooks and shells, as they can introduce security concerns when run against third party terraform modules. If the end user has control over the catalog config however, and if they also have control over all the configured terraform module repos in the catalog config, it would seem to me that the user should be able to make the call on whether or not to allow boilerplate to execute scripts. 

Scripts can significantly extend the possibilities of what can be achieved with boilerplate.  As long as safe defaults are set, the trade off of security versus functionality should lie with the end user.

pseudomorph:
## Description

Allow overriding the default behavior of disabled shell and hooks for boilerplate in scaffold/catalog.

Fixes #3659.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added --enable-shell and --enable-hooks CLI flags and catalog configuration support for scaffold templates.


denis256:
Hi,
Looks like code is failing on lint

```
config/locals_test.go:21:18: string `terragrunt.hcl` has 5 occurrences, make it a constant (goconst)
	mockFilename := "terragrunt.hcl"
	                ^
cli/commands/scaffold/scaffold_test.go:136:13: fieldalignment: struct with 104 pointer bytes could be 96 (govet)
	tests := []struct {
	           ^
config/catalog.go:42:20: fieldalignment: struct with 56 pointer bytes could be [40](https://github.com/gruntwork-io/terragrunt/actions/runs/16354418547/job/46209080937?pr=4540#step:8:41) (govet)
type CatalogConfig struct {
                   ^
config/catalog_test.go:127:21: fieldalignment: struct with 40 pointer bytes could be 32 (govet)
	scaffoldTests := []struct {
	                   ^
level=info msg="File cache stats: 6 entries of total size 64.4KiB"
level=info msg="Memory: 851 samples, avg is 110.4MB, max is 1043.2MB"
level=info msg="Execution took 1m25.69647162s"
options/options.go:100:24: fieldalignment: struct of size 1128 could be 1120 (govet)
type TerragruntOptions struct {
                       ^
config/catalog.go:58:2: if statements should only be cuddled with assignments (wsl)
	if cfg.EnableHooks != nil {
	^
config/catalog.go:61:2: return statements should not be cuddled if block has more than two lines (wsl)
	return fmt.Sprintf("Catalog{URLs = %v, DefaultTemplate = %v, EnableShell = %v, EnableHooks = %v}", cfg.URLs, cfg.DefaultTemplate, enableShell, enableHooks)
	^
internal/services/catalog/catalog.go:112:4: if statements should only be cuddled with assignments (wsl)
			if !s.opts.ScaffoldEnableHooks && catalogCfg.EnableHooks != nil && *catalogCfg.EnableHooks {
			^
8 issues:
* goconst: 1
* govet: 4
* wsl: 3
make: *** [Makefile:[42](https://github.com/gruntwork-io/terragrunt/actions/runs/16354418547/job/46209080937?pr=4540#step:8:43): run-lint] Error 1
```

You can run it locally through `make run-lint`

denis256:
Hi,
Looks like fails TestCatalogParseConfigFile tests

```
2025-07-18T14:18:42.1935666Z --- FAIL: TestCatalogParseConfigFile (0.01s)
2025-07-18T14:18:42.1935956Z     --- FAIL: TestCatalogParseConfigFile/scaffold_both_enabled (0.00s)
2025-07-18T14:18:42.1976684Z     --- FAIL: TestCatalogParseConfigFile/scaffold_enable_hooks_false (0.00s)
2025-07-18T14:18:42.1977784Z     --- FAIL: TestCatalogParseConfigFile/scaffold_enable_shell_true (0.00s)
```

yhakbar:
@pseudomorph 

This functionality has been discussed internally before at Gruntwork, and we previously reached the conclusion that the right way to tackle this functionality in Terragrunt is to introduce a `--no-shell-hook` to disable shell hooks in Boilerplate.

The reason we'd like to bias towards enabling the hooks instead of disabling them is for two major reasons:
1. It aligns with how Boilerplate behaves by default, reducing the complexity that would be present if the two had different default behaviors.
2. It avoids giving users a false sense of security. Most of the time, what you are scaffolding is Terragrunt configurations, and Terragrunt configurations can execute arbitrary commands anyways (via `run_cmd` for example). Users shouldn't scaffold configurations they don't trust to begin with, even while disabling scaffold hooks.

Let me know if that makes sense, and if you agree. Sorry I didn't bring this up earlier when you asked about this.

pseudomorph:
@yhakbar - No problem, shall I invert the logic/options here?

yhakbar:
> @yhakbar - No problem, shall I invert the logic/options here?

Yes please

pseudomorph:
@yhakbar -- qq. do we want a singular flag for both (--no-shell-hooks), or break out for each (--no-shell, --no-hooks) or the exact same flags as boilerplate (--disable-hooks, --disable-shell)?

yhakbar:
> @yhakbar -- qq. do we want a singular flag for both (--no-shell-hooks), or break out for each (--no-shell, --no-hooks) or the exact same flags as boilerplate (--disable-hooks, --disable-shell)?

Great question. I think we should have two flags for parity, but break parity with the names. Terragrunt flags should always start with `non` or `no` when they disable default behavior per the [CLI rules](https://terragrunt-v1.gruntwork.io/docs/reference/cli/rules/).

Do you agree?

pseudomorph:
Makes sense! I'll update the PR.

config/catalog.go
```diff
@@ -38,13 +38,29 @@ var (
 	catalogBlockReg = regexp.MustCompile(fmt.Sprintf(hclBlockRegExprFmt, MetadataCatalog))
 )
 
+// CatalogConfig represents the configuration for the Terragrunt catalog.
 type CatalogConfig struct {
-	DefaultTemplate string   `hcl:"default_template,optional" cty:"default_template"`
-	URLs            []string `hcl:"urls,attr" cty:"urls"`
+	// DisableShell disables shell functions in scaffold templates (CLI flags take precedence)
```
yhakbar: Let's avoid having these diverge, @pseudomorph .

Can we name these `NoShell` (and `no_shell`) and `NoHooks` (and `no_hooks`), just like the CLI flags? That will make it easier for folks to guess what they're called, and how they're overridden in the CLI.

cli/commands/scaffold/scaffold_test.go
```diff
@@ -95,3 +95,151 @@ func TestDefaultTemplateVariables(t *testing.T) {
 	require.True(t, found)
 	require.Equal(t, "git::https://github.com/gruntwork-io/terragrunt.git//test/fixtures/inputs?ref=v0.53.8", *cfg.Terraform.Source)
 }
+
+// TestScaffoldFeatures validates shell and hooks behavior using a table-driven approach.
+// This consolidates testing of both --enable-shell and --enable-hooks flags with their positive and negative cases.
```
yhakbar:
```suggestion
// This consolidates testing of both --no-shell and --no-hooks flags with their positive and negative cases.
```

docs-starlight/src/content/docs/02-features/06-catalog.md
```diff
@@ -41,6 +45,56 @@ This will recursively search for OpenTofu/Terraform modules in the root of the r
 1. See the docs for a selected module: `ENTER`.
 1. Use [`terragrunt scaffold`](/docs/features/scaffold/) to render a `terragrunt.hcl` for using the module: `S`.
 
+## Scaffold Configuration
+
+The catalog configuration supports scaffold-specific settings that control the behavior when scaffolding modules:
+
+- **`disable_shell`** - Controls shell functions in boilerplate templates. When `true`, disables shell commands using `{{shell "command" "args"}}` syntax (enabled by default).
```
yhakbar: Make sure to update this too.

docs-starlight/src/content/docs/02-features/06-catalog.md
```diff
@@ -41,6 +45,56 @@ This will recursively search for OpenTofu/Terraform modules in the root of the r
...
+- **`disable_hooks`** - Controls hooks in boilerplate templates. When `true`, disables `before` and `after` hooks in their `boilerplate.yml` configuration (enabled by default).
+
+**Security Note**: Both shell functions and hooks can execute arbitrary commands on your system. Consider disabling these features when using untrusted templates or for enhanced security.
```
yhakbar:
```suggestion
**Security Note**: Both shell functions and hooks can execute arbitrary commands on your system. Consider disabling these features for enhanced security. Do not scaffold templates you do not trust.
```

docs-starlight/src/content/docs/02-features/07-scaffold.md
```diff
@@ -63,6 +63,78 @@ Important notes:
...
+To disable hooks for security reasons:
+
+```bash
+# Via CLI flag
+terragrunt scaffold github.com/org/repo//modules/mysql --no-hooks
+
+# Via catalog configuration (see below)
+```
+
+**Security Warning**: Both shell functions and hooks can execute arbitrary commands on your system. Consider disabling these features when using untrusted templates or for enhanced security.
```
yhakbar:
```suggestion
**Security Note**: Both shell functions and hooks can execute arbitrary commands on your system. Consider disabling these features for enhanced security. Do not scaffold templates you do not trust.
```

docs-starlight/src/content/docs/02-features/07-scaffold.md
```diff
@@ -63,6 +63,78 @@ Important notes:
...
+
+### Configuration Precedence
+
+When using scaffold with catalog integration, configuration follows this precedence order (highest to lowest):
```
yhakbar:
```suggestion
When using scaffold via the catalog integration, configuration follows this precedence order (highest to lowest):
```

docs-starlight/src/content/docs/02-features/06-catalog.md
```diff
@@ -41,6 +45,56 @@ This will recursively search for OpenTofu/Terraform modules in the root of the r
...
+
+### Configuration Precedence
+
+Scaffold configuration follows this precedence order (highest to lowest):
```
yhakbar:
```suggestion
When using scaffold via the catalog integration, configuration follows this precedence order (highest to lowest):
```

docs-starlight/src/data/flags/catalog-no-hooks.mdx
```diff
@@ -0,0 +1,23 @@
...
+---
+
+Hooks in scaffold templates are **enabled by default**. The `--no-hooks` flag allows you to disable them for security reasons when scaffolding modules through the catalog.
+
+## Configuration Precedence
+
+The setting can be configured in multiple ways with the following precedence order:
+
+1. CLI flag (`--no-hooks`)
+2. Catalog configuration (`disable_hooks = true`)
```
yhakbar:
```suggestion
2. Catalog configuration (`no_hooks = true`)
```

docs-starlight/src/data/flags/catalog-no-shell.mdx
```diff
@@ -0,0 +1,23 @@
...
+---
+
+Shell functions in scaffold templates are **enabled by default**. The `--no-shell` flag allows you to disable them for security reasons when scaffolding modules through the catalog.
+
+## Configuration Precedence
+
+The setting can be configured in multiple ways with the following precedence order:
+
+1. CLI flag (`--no-shell`)
+2. Catalog configuration (`disable_shell = true`)
```
yhakbar:
```suggestion
2. Catalog configuration (`no_shell = true`)
```

docs-starlight/src/data/flags/scaffold-no-hooks.mdx
```diff
@@ -0,0 +1,23 @@
...
+---
+
+Hooks in scaffold templates are **enabled by default**. The `--no-hooks` flag allows you to disable them for security reasons when needed.
+
+## Configuration Precedence
+
+The setting can be configured in multiple ways with the following precedence order:
+
+1. CLI flag (`--no-hooks`)
+2. Catalog configuration (`disable_hooks = true`)
```
yhakbar:
```suggestion
2. Catalog configuration (`no_hooks = true`)
```

docs-starlight/src/data/flags/scaffold-no-shell.mdx
```diff
@@ -0,0 +1,23 @@
...
+---
+
+Shell functions in scaffold templates are **enabled by default**. The `--no-shell` flag allows you to disable them for security reasons when needed.
+
+## Configuration Precedence
+
+The setting can be configured in multiple ways with the following precedence order:
+
+1. CLI flag (`--no-shell`)
+2. Catalog configuration (`disable_shell = true`)
```
yhakbar:
```suggestion
2. Catalog configuration (`no_shell = true`)
```

yhakbar:
I bet the test failures are flakes, but the PR now has conflicts. Do you mind addressing them, @pseudomorph ?

yhakbar:
This has sparked active discussion! We're taking a step back, and changing the way that Boilerplate works in order to try to get the best of both worlds here. 

We're going to introduce a prompt in Boilerplate whenever a shell function or hook is detected in a template, requiring opt-in from users before running. We'll then cut a breaking change release in Boilerplate, and pull in the updated release to Terragrunt, where that same prompt will be presented to users before a hook/shell is run. That work is going to block merging this in.

The `--non-interactive` flag will ignore the prompt in both Boilerplate and Terragrunt.

pseudomorph:
Roger! Sounds like this PR will be unnecessary then. Please lmk if anything is needed from me.

yhakbar:
> Roger! Sounds like this PR will be unnecessary then. Please lmk if anything is needed from me.

We can leave it open. Ideally, we can bump the boilerplate dependency, rebase, then merge it in.

pseudomorph:
Oh, perhaps I misunderstood -- would there would be 2 paths to enable/disable shell/hooks?

1. `--non-interactive` flag to allow hooks/shell
2. `--no-(hooks|shell)` to disable hooks/shell

If so, would `--no-(hooks|shell)` take precedent over `--non-interactive`?

yhakbar:
> Oh, perhaps I misunderstood -- would there would be 2 paths to enable/disable shell/hooks?
> 
>     1. `--non-interactive` flag to allow hooks/shell
> 
>     2. `--no-(hooks|shell)` to disable hooks/shell
> 
> 
> If so, would `--no-(hooks|shell)` take precedent over `--non-interactive`?

Yes. If a user supplied `--no-hooks/shell`, and `--non-interactive`, and they tried to scaffold a template that had one of those, they would get the same behavior that Boilerplate currently has when passing `--disable-hooks/shell`.

The difference would be that if a user doesn't supply `--non-interactive`, and they don't use `--no-hooks/shell`, they'll get a prompt to decide to (dis)allow it.

Issue #4504, opened 2025-07-03
Support `for_each` and `count` meta-parameters for dependencies, stacks, and units


pseudomorph:
### Summary

This RFC proposes adding support for for_each and count meta-arguments to dependency blocks, stack blocks, and unit blocks in Terragrunt configurations. This feature would allow users to dynamically create multiple instances of these blocks based on iterable data (sets, map of strings), similar to how Terraform handles resource iteration.

### Motivation

Currently, Terragrunt users must manually duplicate dependency, stack, and unit blocks when they need multiple similar configurations. This leads to:

1. **Code duplication** - Repeated blocks with minor variations
2. **Maintenance overhead** - Changes must be applied to multiple blocks
3. **Error-prone configuration** - Inconsistencies between duplicated blocks
4. **Scalability issues** - Managing dozens or hundreds of similar blocks becomes unwieldy

The problem becomes particularly acute in enterprise environments where:

* Multiple services require similar infrastructure patterns (databases, caches, etc.)
* Cross-service dependencies require connections to multiple similar resources
* Infrastructure must scale across multiple environments, regions, or teams
* Configuration drift occurs when manually maintaining dozens of similar blocks

### Current Pain Points: Examples

The real-world applicability or valididity of the following scenarios is not meant to be scrutinized. The following examples are proviced only to highlight a possible scenario where `count` and `for_each` could provide benefit.

#### Dependencies

Consider a scenario where you're deploying an ECS service that runs database migrations across multiple Aurora clusters. Each service has its own Aurora cluster with specific connection details and credentials that the migration service needs to access:

```hcl
# Manual duplication for each service's Aurora cluster
dependency "aurora-web" {
  config_path = "../services/web/aurora"
  
  mock_outputs = {
    cluster_endpoint = "web-aurora-cluster.cluster-xyz.us-east-1.rds.amazonaws.com"
    cluster_reader_endpoint = "web-aurora-cluster.cluster-ro-xyz.us-east-1.rds.amazonaws.com"
    port = 5432
    database_name = "web_production"
    master_username = "web_admin"
    security_group_id = "sg-web-aurora-12345"
  }
}

dependency "aurora-api" {
  config_path = "../services/api/aurora"
  
  mock_outputs = {
    cluster_endpoint = "api-aurora-cluster.cluster-xyz.us-east-1.rds.amazonaws.com"
[...]

PR #4595, opened 2025-07-31
Allow unit values file to be written as json.

pseudomorph:
## Description

This allows for unit values files to be written as either `terragrunt.values.hcl` or `terragrunt.values.json`. Default is current behavior (writing `terragrunt.values.hcl`).

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added support for generating values files in JSON format using the `--json-values` flag with `terragrunt stack generate`.

### Migration Guide

N/A

## Summary by CodeRabbit

* **New Features**
  * Added support for generating stack values files in JSON format using a new `--json-values` flag with stack commands. Both HCL and JSON formats are now supported and functionally equivalent.
  * JSON values files are generated with sorted keys for deterministic output and include a comment field for context.
  * When both HCL and JSON values files exist, the JSON file takes precedence during stack operations.

* **Documentation**
  * Updated documentation to describe the new `--json-values` flag, explain values file formats, and provide usage examples for both HCL and JSON output options.

* **Tests**
  * Added comprehensive unit and integration tests to verify JSON values file generation, reading, precedence, and backward compatibility with HCL files.


yhakbar:
@pseudomorph could you explain why you need this feature? The Terragrunt CLI has a _lot_ of flags, and adding another one should be justified with functionality that is valuable and hard to achieve otherwise.

pseudomorph:
@yhakbar - Sure!

We're currently working on a framework which leverages scaffold and stacks to replace our current homegrown set up to manage a large monorepo (3000+ deployments) which is deployed with Atlantis.

At first I was looking at a soft adoption of stacks around our current setup, but the proliferation of repetitive files in Git feel less than ideal. Foregoing the soft adoption would be nice from a DRY perspective, but here is still some necessity of visibility (for the end-user) and change tracking in Git (for Atlantis). The idea I have is render out metadata files for each of the generated stack units, which information about the rendered units (file hashes), along with unit-specific values. That way users can understand what values a specific deployment is getting (and why), and Atlantis can target the appropriate point changes.

We're planning to use stack operations for wholesale standup and teardown of infra, but do point operations on the day-to-day management... Lots of individual teams managing their specific slice of infra.

Having the values file rendered out as json as part of stack generation allows for more efficient processing (no secondary operations or tooling). It is possible to use `terragrunt render` along with a purpose built hcl file, but this is clunky and add extra time to CI operations. We can use external tooling like hcl2json, but that's still extra compute cycles, and it would be preferable if we could keep everything within one tool.

pseudomorph:
Actually, taking a step back, would it make more sense to place this functionality into a stack command?

Something like: `terragrunt stack info`, where information about stacks/units could be rendered from config. Output could contain source paths, values, etc..

This information could be used within workflows to help with the orphan units/state matter raised in #4504 .

Apologies, I should have started this off as an issue.

yhakbar:
@pseudomorph it's all good! Sometimes seeing a prototype helps shape better understanding of the problem and potential solutions. I would recommend creating an RFC, and then soliciting feedback from the community on the design. We want to think through all reasonable alternatives, and approaches (Discord is a good place to do that. Please tag me if you ask for feedback there).

PR #4605, opened 2025-08-01
feat: Discovey false positive parsing errors

denis256:
## Description

Fixes #000.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

### DIFF
--- a/internal/discovery/discovery.go
+++ b/internal/discovery/discovery.go
@@ -252,12 +252,17 @@ func (c *DiscoveredConfig) Parse(ctx context.Context, l log.Logger, opts *option
 	cfg, err := config.ParseConfigFile(parsingCtx, l, parseOpts.TerragruntConfigPath, nil)
 	if err != nil {
 		if !suppressParseErrors || cfg == nil {
-			l.Debugf("Unrecoverable parse error for %s: %s", parseOpts.TerragruntConfigPath, err)
-
+			// Only log unrecoverable errors when not suppressing
+			if !suppressParseErrors {
+				l.Debugf("Unrecoverable parse error for %s: %s", parseOpts.TerragruntConfigPath, err)
+			}
 			return errors.New(err)
 		}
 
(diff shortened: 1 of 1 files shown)


internal/discovery/discovery.go
```diff
@@ -248,6 +250,12 @@ func (c *DiscoveredConfig) Parse(ctx context.Context, l log.Logger, opts *option
 		config.ExcludeBlock,
 	)
 
+	if suppressParseErrors {
+		// If suppressing parse errors, we want to avoid writing diagnostics to the log.
+		parseOptions := append(parsingCtx.ParserOptions, hclparse.WithDiagnosticsWriter(io.Discard, true))
+		parsingCtx = parsingCtx.WithParseOption(parseOptions)
```
yhakbar, line 271: What if the error is actually important? Shouldn't we try to filter diagnostics?

My concern is that something like an HCL function fails, preventing parse of something important in the dependency/dependent and the user doesn't know why it isn't working.

yhakbar (approved):

internal/discovery/discovery.go
```diff
@@ -248,6 +251,26 @@ func (c *DiscoveredConfig) Parse(ctx context.Context, l log.Logger, opts *option
...
 	)
 
+	if suppressParseErrors {
+		// If suppressing parse errors, we want to filter diagnostics that contain references to outputs,
+		// while leaving other diagnostics as is.
+		parseOptions := append(parsingCtx.ParserOptions, hclparse.WithDiagnosticsHandler(func(file *hcl.File, hclDiags hcl.Diagnostics) (hcl.Diagnostics, error) {
+			filteredDiags := hcl.Diagnostics{}
+
+			for _, hclDiag := range hclDiags {
+				containsOutputRef := strings.Contains(strings.ToLower(hclDiag.Summary), "output") ||
```
yhakbar, line 261: I think we could get more specific here, but I think it's fine.

PR #4579, opened 2025-07-29
fix(remote_state): allow Terraform backend init when disable_init=tru…

oyiz-michael:
# fix(remote_state): allow Terraform backend init when disable_init=true (#1422)

## Description

Fixes #1422.

When `remote_state.disable_init = true`, Terragrunt was incorrectly passing `-backend=false` to all Terraform `init` calls, preventing Terraform from initializing its backend. This broke CI/CD workflows where teams need to run `terragrunt validate` or `terragrunt plan` without creating remote state resources.

**Root Cause:** The `GetTFInitArgs()` function was conflating two separate concerns:
1. Terragrunt's remote state resource provisioning (S3 buckets, DynamoDB tables, etc.)
2. Terraform's backend initialization

**The Fix:**
- Removes the `-backend=false` flag for Terraform backend init calls
- Keeps Terragrunt's own remote state provisioning disabled as intended
- Allows Terraform to properly initialize its backend configuration
- Enables CI validation workflows without requiring backend credentials for resource creation

**Files Changed:**
- Updates `GetTFInitArgs()` in `internal/remotestate/remote_state.go`
- Revises the corresponding test in `internal/remotestate/remote_state_test.go`
- Adds comprehensive integration test in `test/integration_test.go`
- Updates documentation in `docs/02-blocks.mdx` and `docs/04-state-backend.mdx`

**Impact:** This change enables teams to use `disable_init=true` for CI validation while maintaining the intended behavior of not provisioning remote state infrastructure.

## TODOs

Read the [Gruntwork contribution guidelines](https://gruntwork.notion.site/Gruntwork-Coding-Methodology-02fdcd6e4b004e818553684760bf691e).

- [x] I authored this code entirely myself
- [x] I am submitting code based on open source software (e.g. MIT, MPL-2.0, Apache)
- [x] I am adding or upgrading a dependency or adapted code and confirm it has a compatible open source license
- [x] Update the docs.
- [x] Run the relevant tests successfully, including pre-commit checks.
- [x] Include release notes. If this PR is backward incompatible, include a migration guide.

## Release Notes (draft)

Fixed `remote_state.disable_init = true` to allow Terraform backend initialization, enabling CI/CD validation workflows with `terragrunt validate` and `terragrunt plan` without requiring permissions to create remote state resources.

### Migration Guide

This change is backward-compatible. No migration required. Existing configurations with `disable_init=true` will continue to work as before, but `terragrunt validate` and similar commands will now work properly in CI environments.

PR #4611, opened 2025-08-04
Fix: allow Terraform backend init when disable_init is true

### LINKED ISSUE #1422: Terraform fails when using remote_state.disable_init
I'm using disable_init to prevent terragrunt from accessing remote state bucket on its own (reason explained below). However, this also disables initializing the backend by terraform: it passes `-backend=false` to terraform init. As a result, the backend is not initialized and all terragrunt commands result in errors:

```
$ terragrunt plan
...
[terragrunt] [/tf] 2020/11/11 21:14:49 Running command: terraform init -backend=false
...
...
[terragrunt] 2020/11/11 21:14:56 Running command: terraform plan
Backend reinitialization required. Please run "terraform init".
Reason: Initial configuration of the requested backend "gcs"
...
Error: Initialization required. Please see the error message above
```

I think a better behavior would be have remote_state.disable_init prevent terragrunt itself from creating the state file in the remote bucket (or accessing the remote bucket bucket in any way), but not prevent terraform itself from initializing the state backend.

The reason this is needed in my case is that I'm using `impersonate_service_account` option in terraform 0.14-rc1, and terragrunt doesn't understand it yet, so it fails whenever it tries to access the remote bucket. I hoped I could solve it by using `emote_state.disable_init`, but I run into this problem.

sjlewis25:
When remote_state.disable_init = true, Terragrunt currently passes `-backend=false` to Terraform, which prevents Terraform from initializing the backend and causes an error like:

  Backend reinitialization required. Please run "terraform init".

This PR removes that flag so that Terraform is allowed to initialize the backend normally. Terragrunt still skips creating S3 buckets or DynamoDB tables, as intended by disable_init.

Fixes #1422.

### DIFF
--- a/internal/remotestate/remote_state.go
+++ b/internal/remotestate/remote_state.go
@@ -118,7 +118,9 @@ func (remote *RemoteState) NeedsBootstrap(ctx context.Context, l log.Logger, opt
 // GetTFInitArgs converts the RemoteState config into the format used by the `tofu init` command.
 func (remote *RemoteState) GetTFInitArgs() []string {
 	if remote.DisableInit {
-		return []string{"-backend=false"}
+		// Don't pass -backend=false anymore; we still want Terraform to init the backend,
+		// but we don't want Terragrunt to manage bucket creation.
+		return []string{}
 	}
 
 	if remote.Generate != nil {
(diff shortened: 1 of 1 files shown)


yhakbar:
Thanks for opening this. There's already a PR open for this:
https://github.com/gruntwork-io/terragrunt/pull/4579

If you find issue with that implementation, please comment there. We'll look to re-review it soon.

PR #4622, opened 2025-08-06
Fixing double border bug at mobile

karlcarstensen:
fixes double border issue on mobile


yhakbar:
@karlcarstensen is there a way we can preserve the command k thing being on the right side of the search bar?
image

How it looks on terragrunt-v1.gruntwork.io:
image

PR #4625, opened 2025-08-06
Fix typos

karlcarstensen:
Remove repeated word
Fix incorrect class


yhakbar:
I really appreciate you catching that, @karlcarstensen .

PR #4659, opened 2025-08-12
Adding light dark toggle

karlcarstensen:
Adding theme switcher

Screenshot 2025-08-12 at 3 34 45 PM


yhakbar:
@karlcarstensen 

This is great! 

Could we invert the icons so that the moon triggers dark mode and the sun triggers light mode?

Also, could we hide the toggle on the homepage as we don't have dark and light variants there (unless you secretly went and built that already)?

PR #4671, opened 2025-08-14
docs: Updating migration guide for `--terragrunt-` flag removal

yhakbar:
## Description

Explicitly maps every legacy flag that's being removed to the modern equivalent.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

### DIFF
--- a/docs-starlight/src/content/docs/06-migrate/03-cli-redesign.md
+++ b/docs-starlight/src/content/docs/06-migrate/03-cli-redesign.md
@@ -61,6 +61,65 @@ terragrunt plan --inputs-debug
 
 You can find the new flag names in the [CLI reference](/docs/reference/cli/) (including the deprecated flags they replace).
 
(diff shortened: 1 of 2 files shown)


wakeful (approved):

regressEdo:
what about `--terragrunt-quiet`, I don't see it as a part of the list, nevertheless it begins with `--terragrunt-`

yhakbar:
> what about `--terragrunt-quiet`, I don't see it as a part of the list, nevertheless it begins with `--terragrunt-`

Great question! It is not a CLI flag, and will not be impacted by breaking changes on the road to 1.0. We may consider deprecating the `run_cmd` HCL function in favor of one that has different semantics (we've been thinking of replacing it with an `exec` HCL function that is quiet by default, and allows users to opt into verbosity with a `--verbose` flag), but that's something we'd be considering on the timeline of a 2.0 release.

PR #4683, opened 2025-08-18
Handle trailing doublestar ** in --queue-include-dir and --queue-exclude-dir

### LINKED ISSUE #4690: Include and exclude glob implementation does not handle ** intuitively
## Describe the bug

The glob implementation used by include and exclude flags does not exclude a directory tree when passing `**` as the last segment of the glob.

## Expected behaviour

Given a tree of units e.g.
```
units/terragrunt.hcl
units/a/terragrunt.hcl
units/b/child/terragrunt.hcl
```
Calling terragrunt with `--queue-exclude-dirs units/**` should exclude the entire tree.

Currently only `units` and `units/a/` is excluded, but `units/b/child` is not.

This behaviour is possible by using `units/**/*`, but in my opinion this is incorrect as `units/**/*` should not exclude `units/terragrunt.hcl` as the glob indicates there must be at least one additional level of depth due to the second `/`.

## Versions

- Terragrunt version: v0.85.0

## Additional context

While my suggested syntax of `units/**` seems intuitive to me, looking into doublestar and globbing I found that different shells (bash, zsh) support different options and [handle globs differently by default](https://unix.stackexchange.com/a/62665)

I'm happy to be pointed at a canonical definition of how globbing in Terragrunt should work and align on that, but if none exists I'd suggest attempting to align on [gitignore syntax](https://git-scm.com/docs/gitignore#_pattern_format) as that's what I'm most familiar with.

Changing the 'ignore entire directory tree' behaviour from `/**/*` to `/**` is a breaking change.

Resonance1584:
## Description

Fixes #4690

Switched the glob library used by `--queue-include-dir` and `--queue-exclude-dir` from https://github.com/mattn/go-zglob to https://github.com/gobwas/glob to handle trailing `**` matches https://github.com/mattn/go-zglob/issues/38

zglob (the previous library) performed a walk of the directory tree and returned a set of paths that matched the globs. Switching allows us to instead test the discovered unit paths in UnitResolver against the patterns, reducing the total number of filesystem walks. This also allows us to remove the hack to handle terragrunt stack paths as the matching now occurs after stacks have been generated.

zglob contained a concurrent walk implementation, this concurrent walk is preserved in GetGlobPaths (which is currently only used in tests) by introducing the [fastwalk library](https://github.com/charlievieth/fastwalk). I'd like to also introduce fastwalk into UnitResolver and run benchmarks, but the current implementation differs with respect to the symlink experiment, and this PR should focus only on changes to globbing.

Implementation is feature-flagged by a strict-control `double-star` to enable a slow roll out of the breaking change. Globs ending in `**/*` will print a warning when the strict-control is not enabled.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Updated --queue-include-dir and --queue-exclude-dir syntax to support a recursive directory tree with ** operator when strict-control double-star is enabled.

### Migration Guide

Including or excluding a directory tree with `**/*` will not be supported in a future version of Terragrunt. This syntax will indicate at least one level of depth must exist. Instead use `**`. E.g. `--queue-exclude-dir "units/**"`. To opt-in to this behavior enable the strict-control `double-star`.

## Summary by CodeRabbit

- **New Features**
  - Optional "double-star" strict control to alter ** / **/* globbing behavior for include/exclude patterns, with improved matching and diagnostic logs.

- **Bug Fixes**
  - Better validation, deduplication, and error reporting for include/exclude configurations, including merges of file-sourced excludes.

- **Documentation**
  - Added double-star reference; updated queue-include-dir and queue-exclude-dir guidance and examples.

- **Tests**
  - Expanded tests to exercise both default and double-star behaviors.

- **Chores**
  - Added a glob matching dependency and deprecation/warning guidance for legacy glob patterns.

### DIFF
--- a/util/file_test.go
+++ b/util/file_test.go
@@ -143,7 +143,10 @@ func getGlobPaths(ctx context.Context, l log.Logger, basePath string, compiledGl
 
 	var paths []string
 
-	filepath.WalkDir(basePath, func(path string, d fs.DirEntry, err error) error {
+	err = filepath.WalkDir(basePath, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
 		path = filepath.ToSlash(path)
 		for globPath, compiledGlob := range compiledGlobs {
 			ll := l.WithField("glob_path", globPath)


util/file_test.go
```diff
@@ -97,18 +97,25 @@ func TestGlobCanonicalPath(t *testing.T) {
 		paths    []string
 		expected []string
 	}{
+		{[]string{"*"}, []string{expectedHelper("module-a"), expectedHelper("module-b")}},
+		{[]string{"**"}, []string{expectedHelper("module-a"), expectedHelper("module-a/terragrunt.hcl"), expectedHelper("module-b"), expectedHelper("module-b/root.hcl"), expectedHelper("module-b/module-b-child"), expectedHelper("module-b/module-b-child/main.tf"), expectedHelper("module-b/module-b-child/terragrunt.hcl")}},
 		{[]string{"module-a", "module-b/module-b-child/.."}, []string{expectedHelper("module-a"), expectedHelper("module-b")}},
 		{[]string{"*-a", "*-b"}, []string{expectedHelper("module-a"), expectedHelper("module-b")}},
 		{[]string{"module-*"}, []string{expectedHelper("module-a"), expectedHelper("module-b")}},
 		{[]string{"module-*/*.hcl"}, []string{expectedHelper("module-a/terragrunt.hcl"), expectedHelper("module-b/root.hcl")}},
-		{[]string{"module-*/**/*.hcl"}, []string{expectedHelper("module-a/terragrunt.hcl"), expectedHelper("module-b/root.hcl"), expectedHelper("module-b/module-b-child/terragrunt.hcl")}},
```
Resonance1584, line 104: Please note this breaking change

go.mod
```diff
@@ -83,8 +83,10 @@ require (
 
 require (
 	github.com/aws/aws-sdk-go-v2/service/s3 v1.83.0
+	github.com/charlievieth/fastwalk v1.0.12
 	github.com/charmbracelet/x/exp/teatest v0.0.0-20250611152503-f53cdd7e01ef
 	github.com/charmbracelet/x/term v0.2.1
+	github.com/gobwas/glob v0.2.3
 	github.com/invopop/jsonschema v0.13.0
```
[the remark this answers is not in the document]
Resonance1584: zglob is still imported in file.go

docs-starlight/src/data/flags/queue-exclude-dir.mdx
```diff
@@ -21,7 +21,9 @@ Dependencies of excluded units will still be run unless they are also explicitly
 </Aside>
 
 <Aside type="note">
-The glob curly braces expansion behaves differently when using environment variables versus command line parameters.
+Glob patterns must use unix forward slashes `/` to separate directories.
```
yhakbar:
```suggestion
Glob patterns must use Unix forward slashes `/` to separate directories, even on Windows machines.
```

internal/runner/common/unit_resolver.go
```diff
@@ -22,14 +23,44 @@ import (
...
+	excludeGlobs      map[string]glob.Glob
+	doubleStarEnabled bool
 }
 
 // NewUnitResolver creates a new UnitResolver with the given stack.
-func NewUnitResolver(stack *Stack) *UnitResolver {
-	return &UnitResolver{
-		Stack: stack,
+func NewUnitResolver(ctx context.Context, stack *Stack) (*UnitResolver, error) {
+	var includeGlobs map[string]glob.Glob
```
yhakbar: NIT: If you were doing this because of the cuddle lint, the more straightforward way is to use:

```go
var (
  includeGlobs
  excludeGlobs
)
```

internal/runner/common/unit_resolver.go
```diff
@@ -740,12 +806,32 @@ func (r *UnitResolver) flagUnitsThatRead(opts *options.TerragruntOptions, units
...
+				l.Debugf("Unit %s is excluded by glob %s", unit.Path, globPath)
+				return true
+			}
+		}
+
+		return false
+	}
+	if !r.doubleStarEnabled {
+		excludeFn = func(l log.Logger, unit *Unit) bool {
+			if unit.FindUnitInPath(opts.ExcludeDirs) {
```
yhakbar: Why aren't we just returning `unit.FindUnitInPath(opts.ExcludeDirs)` here?

internal/runner/configstack/runner_test.go
```diff
@@ -683,7 +683,8 @@ func TestResolveTerraformModulesTwoModulesWithDependenciesExcludedDirsWithDepend
 	t.Parallel()
 
 	opts, _ := options.NewTerragruntOptionsForTest("running_module_test")
-	opts.ExcludeDirs = globCanonical(t, "../../../test/fixtures/modules/module-a*")
+	opts.ExcludeDirs = []string{"../../../test/fixtures/modules/module-a*"}
+	opts.StrictControls.FilterByNames("double-star").Enable()
```
yhakbar: We should check if strict mode is enabled here, then skip this test if it is. The test name should also highlight that we are using the strict control here.

test/integration_download_test.go
```diff
@@ -308,7 +312,11 @@ func TestExcludeDirs(t *testing.T) {
 		}
 
 		// Apply modules according to test cases
-		err := helpers.RunTerragruntCommand(t, fmt.Sprintf("terragrunt run --all apply --non-interactive --log-level trace --working-dir %s %s", tc.workingDir, tc.excludeArgs), &applyAllStdout, &applyAllStderr)
+		strictControl := ""
+		if tc.enableDoubleStar {
+			strictControl = "--strict-control double-star"
```
yhakbar, line 326: Make sure that you check whether or not strict mode is enabled and disable it before you add this here.

util/file.go
```diff
@@ -779,20 +863,16 @@ func GetExcludeDirsFromFile(baseDir, filename string) ([]string, error) {
 		return nil, err
 	}
 
-	var dirs []string
+	var dirs []string //nolint: prealloc
```
yhakbar: I think we can optimistically prealloc this to `len(lines)`

PR #4696, opened 2025-08-18
Lower padding margins

karlcarstensen:
Reduce margins and gap

## Summary by CodeRabbit

- New Features
  - Added a Testimonials section to the homepage to showcase user feedback.

- Style
  - Reduced vertical spacing between homepage sections on medium+ screens for a tighter layout.
  - Removed margins from the main hero heading for cleaner alignment.

### DIFF
--- a/docs-starlight/src/components/dv-Hero.astro
+++ b/docs-starlight/src/components/dv-Hero.astro
@@ -33,7 +33,7 @@ import TerraformIcon from '@assets/terraform-logo.svg';
 
     <div class="px-5 md:px-0 relative z-20 flex flex-col items-center gap-10 w-full text-center">
         <div class="flex flex-col justify-start items-center gap-5 w-full pt-5 pd:pt-0">
-            <h1 class="text-[42px] md:text-[64px] text-white leading-12 md:leading-18">The Open Source<br>IaC Orchestrator<br>Platform Teams Trust</h1>
+            <h1 class="text-[42px] md:text-[64px] text-white leading-12 md:leading-18 m-0">The Open Source<br>IaC Orchestrator<br>Platform Teams Trust</h1>
             <p class="px-12 lg:px-40 md:px-0 text-lg text-[var(--color-gray-1)]">
                 Organize and speed up your Infrastructure as Code, with the #1 IaC Orchestrator
             </p>
--- a/docs-starlight/src/pages/index.astro
+++ b/docs-starlight/src/pages/index.astro
@@ -37,7 +37,7 @@ import '@styles/global.css';
         <!-- FIXME: Between 1024px - 1046px the dotted lines do not match the edge of the sections -->
         <div class="hidden md:block pointer-events-none absolute left-1/2 top-0 -translate-x-1/2 md:w-[633px] lg:w-[852px] h-full z-0 border-x border-dashed border-gray-3"></div>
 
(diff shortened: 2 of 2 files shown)


docs-starlight/src/components/dv-Hero.astro
```diff
@@ -33,7 +33,7 @@ import TerraformIcon from '@assets/terraform-logo.svg';
 
     <div class="px-5 md:px-0 relative z-20 flex flex-col items-center gap-10 w-full text-center">
         <div class="flex flex-col justify-start items-center gap-5 w-full pt-5 pd:pt-0">
-            <h1 class="text-[42px] md:text-[64px] text-white leading-12 md:leading-18">The Open Source<br>IaC Orchestrator<br>Platform Teams Trust</h1>
+            <h1 class="text-[42px] md:text-[64px] text-white leading-12 md:leading-18 m-0">The Open Source<br>IaC Orchestrator<br>Platform Teams Trust</h1>
```
josh-padnick, line 36: @karlcarstensen Should we need to explicitly specify `m-0` here? I'm wondering if we're better off adopting the [Tailwind pre-flight styles](https://tailwindcss.com/docs/preflight) so that all margins are removed automatically. That will also fix every other situation where we use an `h1` or any other header.

josh-padnick: I'd prefer we use `rem` values to honor users who adjust their default font size in the browser. With the `px` values, the users who zoom will see a bigger font, but not the users who adjust their browser default font size. In fact, since the `leading` values use `rem` text-` values use `px`, it will make the pages look pretty crowded in general.

Could you also replace `text-[42px] md:text-[64px] text-white` with ```text-4xl md:text-6xl```?

karlcarstensen: This is great feedback and I agree, however I didn't set up this project. This is the only H1 in the codebase currently, so it was much more predictable specifying `m-0` to this single instance than adding pre-flight styles with an unknown impact. I am trying to be respectful of Gruntworks' budget and time. If there is a desire to set up a more robust stylesheet and/or design system I would be happy to do that!

I have just updated the classnames with `text-4xl md:text-6xl` that will be in a separate PR. If there is a desire to switch to rems I can do that also! Just let me know or make a ticket and assign me and it will get done.

karlcarstensen: Here's the PR https://github.com/gruntwork-io/terragrunt/pull/4710

yhakbar: Strongly agree that we should ticket and prioritize work for @karlcarstensen . Tin should make the call on priority of those items.

josh-padnick: Thanks for the quick turnaround! I think we want to be paying down or at the very least not adding tech debt, and today's tech debt is already slowing us down through a seemingly endless list of minor fixes. 

Maybe you could put together a list of improvement ideas, timebox yourself to 2 - 3 hours, do a final check with Tin, and see what you can knock out. Some things that are top of mind for me from looking at the code base:

1. Tailwind preflights, rems (mentioned above)
2. Standardizing the components we use in the site
3. Cleaning up the global CSS so that most of the CSS is in the components as Tailwind classes

I'm sure there's other stuff if I took a second look, but that's what's top of mind for now.

PR #4784, opened 2025-09-04
A collection of website polishing

josh-padnick:
This makes a collection of small polish improvements to the Terragrunt website. I have _not_ QA'd this on mobile yet, though I wouldn't expect any of these changes to negatively affect mobile since I'd used `rem` for most changes (which will scale with default font size). 

The changes include:

1. Tighten up the sidebar spacing
2. Standardize the CSS ordering so that dev and prod render the same (previously, there was a large margin to the left of the body text in dev, but not prod)
3. Improve main body formatting

## Summary by CodeRabbit

- Style
  - Refined sidebar layout: adjusted spacing, margins, and link font sizing for clearer navigation.
  - Updated main content styling: improved paragraph line-height, margins, and code font size; standardized container margins.
  - Harmonized component styling via reorganized CSS layers, ensuring consistent appearance across buttons, lists, contact form, and search.
  - Minor padding tweaks for tighter sidebar list items.

### DIFF
--- a/docs-starlight/src/styles/global.css
+++ b/docs-starlight/src/styles/global.css
@@ -321,6 +321,10 @@ p code {
   margin-bottom: 1.5rem;
 }
 
+.main-pane starlight-file-tree {
+  margin-bottom: 1.8rem;
+}
+
 .main-pane .code {
   font-size: calc(14 / 16 * 1rem);
 }
--- a/docs-starlight/src/styles/global.css
+++ b/docs-starlight/src/styles/global.css
@@ -316,6 +316,11 @@ p code {
   margin-inline: 0;
 }
 
+.main-pane p {
+  line-height: 1.6;
+  margin-bottom: 1.5rem;
+}
+
 .main-pane .code {
   font-size: calc(14 / 16 * 1rem);
 }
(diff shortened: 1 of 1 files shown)


docs-starlight/src/styles/global.css
```diff
@@ -282,18 +282,35 @@ p code {
...
-  padding-bottom: 0.5em;
 }
 
 #starlight__sidebar ul li ul li {
   padding-left: 1em;
 }
 
-#starlight__sidebar details summary {
+#starlight__sidebar ul li a {
+  font-size: calc(16 / 16 * 1rem);
```
karlcarstensen, line 292: I've not seen this before - why is it done this way vs. just setting it to 1rem?

josh-padnick: I prefer to think in px, but it’s important to use rem whenever we want the thing to scale as the user increases their default font size. So by using `calc(14 / 16 * 1rem)` I can say “make this font-size `14px`, but express it in `rem`.”

But admittedly `16 / 16` is a bit strange.

yhakbar (approved):

josh-padnick:
FYI, I do just want to acknowledge that I didn't do sufficient QA testing for mobile here. I'll tackle that shortly in a separate PR.

PR #4810, opened 2025-09-09
Adjustments to form

karlcarstensen:


yhakbar (changes requested):
Seems busted. Probably a typo.

yhakbar (approved):

yhakbar (approved):

karlcarstensen:
We set this functionality up as a trojan horse for tracking on Hubspot. But turns out we don't need this functionality on TS forms because it already goes to a different area in hubspot. Closing.

PR #4818, opened 2025-09-10
fix: always fetch dep output from remote state when selected

pseudomorph:
## Description

There are cases where parallelization might affect an init-ed dir during read, such as when using Terragrunt with Atlantis. In these cases, when a user passes the option to fetch dependency output from (remote) state, it should not use output from init-ed dirs.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

Updated: --fetch-dependency-output-from-state now will always pull directly from state, even when dependency dirs are init-ed.

### DIFF
--- a/config/dependency.go
+++ b/config/dependency.go
@@ -809,7 +809,8 @@ func getTerragruntOutputJSON(ctx *ParsingContext, l log.Logger, targetConfig str
 
 	ctx.TerragruntOptions.Engine = engineOpts
 
-	if isInit {
+	// Do not fetch from init-ed folder if user has explicitly asked to fetch from state.
+	if isInit && !ctx.TerragruntOptions.FetchDependencyOutputFromState {
 		return getTerragruntOutputJSONFromInitFolder(ctx, l, workingDir, remoteStateTGConfig.GetIAMRoleOptions())
 	}
 


BradBarnich:
This can also happen when there is a terragrunt unit that is using loose tf files instead of a `terraform { source = ... }` block. In this case the init happens in the unit directory (not in .terragrunt-cache, see screenshot). 

Screenshot 2025-09-11 143019


When this unit is used as dependency, the code path in question is triggered, and then even with the 'fetch output from state` option set, it instead calls terraform output

```
brad@DESKTOP-608UVN8 ~/c/S/s/a/p/d/c/u/p/ssh_bastion (master)> env | grep TG_DEPENDENCY_FETCH_OUTPUT_FROM_STATE
TG_DEPENDENCY_FETCH_OUTPUT_FROM_STATE=true
```

```
brad@DESKTOP-608UVN8 ~/c/S/s/a/p/d/c/u/p/ssh_bastion (master)> terragrunt plan -log-level debug
...
14:35:30.488 DEBUG  [../vpc] Detected module ../vpc/terragrunt.hcl is already init-ed. Retrieving outputs directly from working directory.
14:35:30.488 DEBUG  [../vpc] Running command: terraform output -json
```

I think the proposed fix is fine, I had dabbed with reordering the operations to reach for remote state before running output in an init'd dependency https://github.com/gruntwork-io/terragrunt/commit/d69eed79c10610fab8a60b120b827b1b818f0afb

I'm not familiar enough with terragrunt to know which is a more appropriate fix. Because the init check doesn't hit on the more common path `terraform { source = ... }`, it feels like it is more or less dead code

yhakbar:
@pseudomorph thanks for opening this PR. It seems entirely reasonable. Do you mind adding some testing to confirm that this resolves the issue we're looking to address here?

PR #4860, opened 2025-09-18
Adds CTA for Terragrunt Scale on sidebar

karlcarstensen:
## Summary by CodeRabbit

* **New Features**
  * Added a custom PageSidebar (registered in site config) that preserves existing sidebar content and adds a decorative divider plus a promo block for Terragrunt Scale (heading, description, CTA with icon; promo hidden on smaller viewports).
* **Style**
  * Minor whitespace/formatting cleanup in global stylesheet with no behavioral changes.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,2 +1,2 @@
-* @denis256 @wakeful @yhakbar
+* @denis256 @yhakbar
 
--- a/config/config_helpers.go
+++ b/config/config_helpers.go
@@ -1238,7 +1238,7 @@ func extractSopsErrors(err error) *errors.MultiError {
 	// using reflection extract GroupResults from getDataKeyError
 	// may not be compatible with future versions
 	errValue := reflect.ValueOf(err)
-	if errValue.Kind() == reflect.Ptr {
+	if errValue.Kind() == reflect.Pointer {
 		errValue = errValue.Elem()
 	}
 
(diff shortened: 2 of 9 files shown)


docs-starlight/src/styles/global.css
```diff
@@ -169,6 +169,13 @@
         padding: 10px 10px;
     }
   }
+
+  @media (min-width: 1152px) {
+    .right-sidebar {
+      padding-top: 0 !important;
+      position: relative !important;
+    }
+  }
```
josh-padnick: Can we use standard tailwind breakpoints instead of a custom media query? Can we put these styles in the component instead of this `global.css` file which is kind of a grab bag of everything right now?

docs-starlight/src/components/PageSidebar.astro
```diff
@@ -0,0 +1,23 @@
+---
+
+import DefaultPageSidebar from '@astrojs/starlight/components/PageSidebar.astro';
+import { Icon } from '@astrojs/starlight/components';
+
+---
+
+<DefaultPageSidebar>
+  <slot />
+</DefaultPageSidebar>
```
josh-padnick: This is a nice way of implementing this!

docs-starlight/src/styles/global.css
karlcarstensen: No and no, unfortunately. I tried both. What this is doing is overriding a baked-in starlight style for the right sidebar which is part of their structure and has some very strange markup

karlcarstensen: The design was updated and I was able to remove this. Resolving.

PR #4893, opened 2025-09-27
Fix inconsistencies in CLI documentation

### LINKED ISSUE #4892: Update docs regarding CLI Redesign
## "Migrate" topic quick fix for deprecated run-all

**'Before'** and **'After'** definitions of the run-all command are repeated, which looks like it's not intended with the context of the documentation explanation.

Assuming it's intending to describe the difference between the deprecated 'run-all' and the current 'run --all' command, the official docs on the website are showing the same command both for **'Before'** and **'After'** definitions.

Image

- [x] Changes required.
- [ ] Level of effort.

## RFC Not Needed

Doesn't seem like this needs much explanation, however if it's intended and needed, would like to understand why this is.

pedromtsilva:
Updated command examples for clarity and accuracy regarding the use of the 'run' command in Terragrunt CLI, which was presenting itself with a repeated command definition.

Also changed upper reference regarding the `graph` flag to match the definition on the bottom of the page.

## Description

Fixes #4892.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **Documentation**
  * Updated migration guide to reflect revised CLI usage and examples.
  * Corrected command examples (e.g., use terragrunt run --graph plan).
  * Fixed spacing/flag formatting in code snippets (e.g., removed extra spaces).
  * Replaced prose references of “run --all” with “run-all” and updated Before/After snippets for clarity.


denis256 (approved):

docs-starlight/src/content/docs/07-migrate/03-cli-redesign.md
```diff
@@ -208,25 +208,25 @@ terragrunt run -- workspace ls
...
 In addition to allowing for explicit invocation of OpenTofu/Terraform instead of using shortcuts, the `run` command also takes on the responsibilities of the now deprecated `run-all` and `graph` commands using flags.
 
-For example, if you are currently using the `terragrunt run --all` command, you can switch to the `run` command with the `--all` flag instead.
+For example, if you are currently using the `terragrunt run-all` command, you can switch to the `run` command with the `--all` flag instead.
 
 Before:
 
 ```bash
-terragrunt run --all plan
+terragrunt run-all plan
```
denis256, line 229: `run-all plan` command is deprecated and will be removed in a future version of Terragrunt. Use `terragrunt plan --all` instead

docs-starlight/src/content/docs/07-migrate/03-cli-redesign.md
```diff
@@ -208,25 +208,25 @@ terragrunt run -- workspace ls
 Similarly, commands like `graph` won't be supported as a shortcut, as `graph` is a now deprecated command in the Terragrunt CLI. Supporting it as a shortcut would be misleading, so you can use the `run` command to run it explicitly:
 
 ```bash
-terragrunt run graph
+terragrunt run --graph plan
```
yhakbar, line 211: Wait, this is wrong. The point of the command above is that the OpenTofu/Terraform command can't be run without using the `run` command.

yhakbar (changes requested):

docs-starlight/src/content/docs/07-migrate/03-cli-redesign.md
yhakbar: This is right. This is the `before`, not the `after`.

docs-starlight/src/content/docs/07-migrate/03-cli-redesign.md
pedromtsilva: I don't have much knowledge on this flag. CodeRabbit, when i changed it to `terragrunt run --graph` only, suggested the changes above.

I followed [this run --graph](https://terragrunt.gruntwork.io/docs/reference/cli/commands/run/#graph) command usage description, in the documentation. Also, [here](https://terragrunt.gruntwork.io/docs/migrate/cli-redesign/#use-the-new-run-command), in the new CLI redesign section, it says to use `terragrunt run graph`. Please indicate the right usage to correctly apply the changes.

yhakbar: I know it's confusing, but there are two different `graph` commands, which drove some of the adjustments in the CLI Redesign project.

- There's the [OpenTofu/Terraform graph command](https://opentofu.org/docs/cli/commands/graph/)
- Terragrunt also has used to have a command named `graph`, which was removed, and the behavior of `run` was updated to include the [`--graph` flag](https://terragrunt.gruntwork.io/docs/reference/cli/commands/run/#graph), which replaced it.

For some OpenTofu/Terraform commands, if there's no ambiguity, we directly forward the command from Terragrunt to OpenTofu/Terraform.

e.g.

```bash
terragrunt plan
```

Under the hood, calls either `tofu plan` or `terraform plan`, depending on the user's configurations.

This is just a shortcut, though, and it's the same as running the following:

```bash
terragrunt run plan
```

For other commands like `graph`, we don't want users who were using the old Terragrunt `graph` command to get confused when they see OpenTofu running the `graph` command, so we instead require that they explicitly use the `run` form if they want Terragrunt to run the OpenTofu command.

```bash
terragrunt run graph
```

You can see in the line above where you're proposing changes that we're explaining this:

> Similarly, commands like `graph` won't be supported as a shortcut, as `graph` is a now deprecated command in the Terragrunt CLI. Supporting it as a shortcut would be misleading, so you can use the `run` command to run it explicitly:

If you think we can be clearer about the explanation, please consider updating that line to make it clearer.

yhakbar:
Hey @pedromtsilva ,

Do you have a sec to rebase this PR and address the feedback?

pedromtsilva:
Hey @yhakbar 

Apologies for the long time taken regarding this issue. I've had other personal issues, will look into this.

docs-starlight/src/content/docs/07-migrate/03-cli-redesign.md
pedromtsilva: Okay, so the currently existing OpenTofu/Terraform command "graph" now explicitly requires the use of "terragrunt run graph" to be executed instead. 

And the old 'terragrunt graph' command is now replaced with the explicit command "terragrunt run --graph".

Will be waiting for confirmation on this, and try to make the changes accordingly. I believe I got a bit confused because I was unsure of what it meant for a command to have a shortcut.

PR #4953, opened 2025-10-14
fix: Fixing `--queue-ignore-errors` resulting in ignored DAG order

### LINKED ISSUE #4947: v0.84.4 (and 0.90.0) ignores dependency order on destroy
## Describe the bug
Dag dependency order seems to be broken in 0.90.0 and 0.89.4 when destroying and using --queue-ignore-errors

## Steps To Reproduce
Have a few folders with terragrunt files having dependencies from each other and run destroy on it
cluster/terrragrunt.hcl:
```
dependency "label" {
  config_path = "../_label"
}
```
istio: (note that one, as it will be used in the screenshots below)
```
dependency "label" {
  config_path = "../_label"
}

dependency "cluster" {
  config_path = "../cluster"
}
```
addons/base/terragrunt.hcl:
```
dependency "cluster" {
  config_path = "../../cluster"
}
```
addons/logging/terragrunt.hcl:
```
dependency "cluster" {
  config_path = "../../cluster"
}
dependency "base" {
  config_path = "../base"
}
```
[etc....]

When running a destroy with --queue-ignore-errors, it destroyed the cluster, then _label .. and then everything else just failed

## Expected behavior
--queue-ignore-errors should just prevent failure on errors, not disable the whole dependency tree

## Nice to haves

- [ ] Terminal output
- [ X ] Screenshots

## Versions

- Terragrunt version: 0.87.0 vs 0.90.0
- OpenTofu/Terraform version: v1.10.6
- Environment details (Ubuntu 20.04, Windows 10, etc.): ubuntu 24.04

## Additional context

Start of destroy (all good):

Image

Then it destroys the cluster: (note the timestamps)

Image

And then starts planning the destroy for other modules, while still destroying the cluster:

Image

(note: all those dependencies failed to destroy, since the rights for the user was already gone as part of the EKS cluster module :) )

yhakbar:
## Description

Fixes #4947.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Updated `--queue-ignore-errors` so that it wouldn't also ignore DAG order.

## Summary by CodeRabbit

- New Features
  - Destroy now supports ignoring dependency errors while preserving correct dependency order. Progress continues as modules reach terminal states (succeeded/failed), improving resilience in large runs.
- Bug Fixes
  - Prevents stalls during destroy by consistently evaluating readiness of dependencies/dependents, even when some modules fail but are terminal.
- Tests
  - Added unit and integration tests to validate destroy order and continued progress when ignoring errors.

### DIFF
--- a/internal/queue/queue.go
+++ b/internal/queue/queue.go
@@ -315,79 +315,78 @@ func (q *Queue) GetReadyWithDependencies() []*Entry {
 		}
 
 		if e.IsUp() {
-			// Up logic: all dependencies must be succeeded (or terminal if ignoring errors)
-			allDepsReady := true
-
-			for _, dep := range e.Config.Dependencies {
-				depEntry := q.entryByPathUnsafe(dep.Path)
-				if depEntry == nil {
-					allDepsReady = false
-					break
-				}
-
-				// When ignoring dependency errors, allow scheduling if dependencies are in a terminal state
-				// (succeeded OR failed), not just succeeded
-				if q.IgnoreDependencyErrors {
-					if !isTerminal(depEntry.Status) {
-						allDepsReady = false
-						break
-					}
-
-					continue
-				}
-
-				if depEntry.Status != StatusSucceeded {
-					allDepsReady = false
-					break
-				}
-			}
-
-			if allDepsReady {
+			if q.areDependenciesReadyUnsafe(e) {
 				out = append(out, e)
 			}
 
 			continue
 		}
 
-		// Down logic: all dependents must be succeeded (or terminal if ignoring errors)
-		allDependentsReady := true
-
-		for _, other := range q.Entries {
-			if other == e || other.Config.Dependencies == nil {
-				continue
-			}
-
-			for _, dep := range other.Config.Dependencies {
-				if dep.Path == e.Config.Path {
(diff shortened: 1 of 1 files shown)


internal/queue/queue.go
```diff
@@ -315,63 +315,78 @@ func (q *Queue) GetReadyWithDependencies() []*Entry {
...
-			for _, dep := range other.Config.Dependencies {
-				if dep.Path == e.Config.Path {
-					if other.Status != StatusSucceeded {
-						allDependentsReady = false
-						break
+// areDependentsReadyUnsafe checks if all dependents of an entry are ready for "down" commands.
+// For down commands, all dependents must be in a succeeded state (or terminal if ignoring errors).
+// Should only be called when the caller already holds a read lock.
+func (q *Queue) areDependentsReadyUnsafe(e *Entry) bool {
+	for _, other := range q.Entries {
```
denis256, line 365: can we add a fast path to exit if len(e.Config.Dependencies) == 0?

yhakbar: Hmmmm... Maybe? I'm not sure about that. 

We're checking here if other units depend on this unit, and that they're ready, right? Would this entry having no dependencies mean that we can short circuit?

PR #4958, opened 2025-10-14
feat: Adding `filter` package

yhakbar:
## Description

Adds the initial implementation of the `filter` package, so we can start to add basic integration into the codebase under the `filter` experiment.

## TODOs

Read the [Gruntwork contribution guidelines](https://gruntwork.notion.site/Gruntwork-Coding-Methodology-02fdcd6e4b004e818553684760bf691e).

- [ ] I authored this code entirely myself
- [x] I am submitting code based on open source software (e.g. MIT, MPL-2.0, Apache)]
- [x] I am adding or upgrading a dependency or adapted code and confirm it has a compatible open source license
- [x] Update the docs.
- [x] Run the relevant tests successfully, including pre-commit checks.
- [x] Include release notes. If this PR is backward incompatible, include a migration guide.

## Release Notes (draft)

Added `filter` package.

PR #4963, opened 2025-10-15
chore: Refactor out `component` package

yhakbar:
## Description

Refactoring out a new package named `component` that supports a high-level struct for units and stacks.

Makes it easier to avoid cyclic dependencies when discovery requires stuff from other packages, as they won't have to depend on `discovery` directly.

Started doing this in #4958, but I figured it would be cleaner to break this part out into its own PR.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added `component` package to support high-level structs for units and stacks.

## Summary by CodeRabbit

- New Features
  - Clearer errors for dependency cycles during discovery.
  - Unified component model for units and stacks across find and list commands.
  - More consistent colorization of types in CLI output.

- Bug Fixes
  - Improved accuracy in filtering and sorting units and stacks.

- Refactor
  - Internal rearchitecture to a component-based model for discovery and dependencies; no user action required.

- Tests
  - Added comprehensive tests for component sorting, filtering, and cycle detection; updated command tests to match new model.


internal/component/component.go
```diff
@@ -0,0 +1,153 @@
...
+
+// Kind is the type of Terragrunt component.
+type Kind string
+
+// Component represents a discovered Terragrunt configuration.
+type Component struct {
+	Parsed           *config.TerragruntConfig
+	DiscoveryContext *DiscoveryContext
+
+	Kind Kind
```
denis256, line 32: We may need a "version" field to support multiple implementations 🤔

yhakbar: 🤔 ...

I don't understand.

denis256: We can add a top-level version field in each HCL file to support future syntax or behavior changes without breaking existing setups.

yhakbar: Ah, I see. Maybe something to tackle when we get around to redoing our HCL parsing strategy.

I'm concerned it might result in performance penalties if we have to do a preemptive partial parse to determine the version, then do one of two different parses depending on the version.

PR #5021, opened 2025-10-24
chore: Disable experimental tests ignore in CI

yhakbar:
## Description

Requires that experimental integration tests pass.

As a consequence of requiring that these tests, some bug fixes were made. 

These include:

- Fixing CAS usage logic to skip local sources, even when CAS is allowed.
- Updated experiment enable logic to ensure that they're getting enabled correctly.
- Updated infix `RequiresHCLParsing` logic to return the parent expression instead of any half of the expression (e.g. return `./foo | external=true` instead of `external=true` if we're preventing attribute based expressions).
- Fixed logic for early return in evaluation of empty filters on a list of components to return the full slice of components instead of an empty slice when evaluating on files.
- Updated provider version pins to be on a floating `hashicorp/null` version instead of `3.2.3` (helps with M series macbooks running the tests not having that version available).
-  Pinned the version of Terragrunt referenced for remote modules to be a working commit that has the adjustment above made (will need to have the commit moved, then changed to a tag once this is merged).
- Updated `CopyEnvironment` to always evaluate symlinks, which helps folks on M series macbooks as they run tests (t.Tempdir() points to a symlinked directory, which breaks directory comparison throughout tests).
- Fixed `hcl validate` and `hcl filter` `--filter` integration tests.
- Cleaned up dummy counters in `runCmdFlagsFixture` on every run, and added a `.gitignore` to prevent them from being accidentally committed going forward.
- Updated `TestVersionIsInvokedInDifferentDirectory` to no longer expect an extra version call, as the OpenTofu auto provider cache dir feature is now GA.
- Made access to the `ProviderCache` fields synchronized using mutices to ensure they're accessed safely across goroutines.
- Upgraded OpenTofu to the latest version in `mise.toml`.
- Updated all usage of `WalkWithSymlinks` to `WalkDirWithSymlinks`. This was necessary to avoid issues with the `symlinks` experiment causing the experiment mode integration test to fail.

This also incidentally required adding support for `terragrunt.stack.hcl` files in `hcl validate`.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **Chores**
  * CI now fails fast on setup or test errors for more reliable pipelines.
  * Many example/test Terraform provider constraints updated to allow newer patch releases (~> 3.2.4).

* **New Features / Improvements**
  * HCL validation/reporting improved with clearer, file-scoped diagnostics for stacks.
  * Downloads can use content-addressable storage with safe fallbacks; relative path resolution improved.
  * File discovery, symlink handling and path normalization made more reliable across platforms.
  * Provider cache made concurrency-safe to improve stability.

* **Tests**
  * Integration tests updated for consistent, symlink-resolved paths, streamlined command invocation, and simplified expectations.

### DIFF
--- a/test/fixtures/download/hello-world/main.tf
+++ b/test/fixtures/download/hello-world/main.tf
@@ -29,6 +29,6 @@ output "test" {
 }
 
 module "remote" {
-  source = "github.com/gruntwork-io/terragrunt.git//test/fixtures/download/hello-world-no-remote?ref=4de37b32367af50a3d5612e42a0965d8e477cbec"
+  source = "github.com/gruntwork-io/terragrunt.git//test/fixtures/download/hello-world-no-remote?ref=2ca67fe2dbd3001c4cffa66df9567ca497182cb1"
   name   = var.name
 }
--- a/test/fixtures/download/remote-ref/terragrunt.hcl
+++ b/test/fixtures/download/remote-ref/terragrunt.hcl
@@ -3,5 +3,5 @@ inputs = {
 }
 
 terraform {
-  source = "git::git@github.com:gruntwork-io/terragrunt.git//test/fixtures/download/hello-world-no-remote?ref=4de37b32367af50a3d5612e42a0965d8e477cbec"
+  source = "git::git@github.com:gruntwork-io/terragrunt.git//test/fixtures/download/hello-world-no-remote?ref=2ca67fe2dbd3001c4cffa66df9567ca497182cb1"
 }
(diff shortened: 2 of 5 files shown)


test/fixtures/download/hello-world/main.tf
```diff
@@ -29,6 +29,6 @@ output "test" {
 }
 
 module "remote" {
-  source = "github.com/gruntwork-io/terragrunt.git//test/fixtures/download/hello-world-no-remote?ref=v0.77.22"
+  source = "github.com/gruntwork-io/terragrunt.git//test/fixtures/download/hello-world-no-remote?ref=2ca67fe2dbd3001c4cffa66df9567ca497182cb1"
```
denis256, line 32: Hm, why do we need a hash reference? it will not work through git tag?

denis256: Found in description

PR #5041, opened 2025-10-30
allow input validation with variable in source or version

### LINKED ISSUE #4986: Terragrunt cannot validate inputs if Terraform module uses variables in "source" or "version" fields
## Describe the bug

Terragrunt cannot validate inputs if Terraform module uses variables in "source" or "version" fields. Being able to use variables (or locals) in `source` or `version` fields within a Terraform module is an OpenTofu only feature available since OpenTofu 1.8 (I believe).

When Terragrunt is ran with a command like:

```
terragrunt hcl validate --all --strict --inputs
```

an error similar to this is given:

```
19:51:02.367 ERROR  Run failed: 2 errors occurred:
* Variables not allowed: Variables may not be used here. (and 17 other messages)
* Variables not allowed: Variables may not be used here. (and 3 other messages)
```

## Steps To Reproduce

- Create a Terraform module, that relies on another Teraform submodule. Use a local variable when referring to the submodule version. Example:

```hcl
################################################################################

terraform {
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "6.16.0"
    }
  }
}

provider "aws" {
  region = var.region

  allowed_account_ids = [
    var.account_id
  ]
}

################################################################################

locals {
  module_identity_aws_release = "1.1.2"
}
################################################################################

module "iam_role" {
  source        = "gitlab.com/bitservices/identity/aws//role"
  version       = local.module_identity_aws_release
  name          = "test-iam-role"
  assume_policy = <<POLICY
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "AWS": [
          "arn:aws:iam::${var.account_id}:root"
        ]
      },
      "Action": "sts:AssumeRole",
      "Condition": {}
    }
  ]
}
POLICY
}

module "iam_role_policy_ec2" {
  source  = "gitlab.com/bitservices/identity/aws//role-policy"
  version = local.module_identity_aws_release
  name    = format("%s-ec2", module.iam_role.name)
  role    = module.iam_role.name
  policy  = <<POLICY
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "ec2:*"
            ],
            "Resource": "*"
        },
    ]
}
POLICY
}

################################################################################
```

- Reference this Terraform module with any basic Terragrunt configuration
- The errors shown in the problem statement occur.

## Expected behavior

For validation to pass (as long as inputs match required/optional variables). Even if they don't match an error message stating this rather than the errors above which imply Terragrunt is unable to process the module at all.

## Must haves

## Nice to haves

- [x] Terminal output
- [ ] Screenshots

## Versions

- Terragrunt version: 0.90.0
- OpenTofu/Terraform version: 1.10.6 (not relevant as not invoked)
- Environment details (Ubuntu 20.04, Windows 10, etc.): Archlinux

## Additional context

The problem seems to be in this file: https://github.com/gruntwork-io/terragrunt/blob/main/tf/tf.go
Due to import: `"github.com/hashicorp/terraform-config-inspect/tfconfig"`.

Terraform does not support variables in module references, OpenTofu does.

Terraform has this: https://github.com/hashicorp/terraform-config-inspect

I cannot find an equivalent for OpenTofu. Until one exists this might be a bit hard for me to work on myself as I guess the solution (package) will need to come from OpenTofu. Someone more involved in the OpenTofu project may know if a package that provides similar functionality is available perhaps under a different name.

thisguycodes:
## Description

Terragrunt does not support using variables in the `source` or `version` fields of `module` blocks, but OpenTofu does.

This becomes a problem because when validating inputs via `terragrunt hcl validate --inputs` we use github.com/hashicorp/terraform-config-inspect to read the variables defined by the target module; since this is a Go module designed for Terraform it returns error diagnostics when variables are used as defined above.

Thus a valid OpenTofu module will fail a run of `terragrunt hcl validate -inputs`.

Fixes #4986

Despite these error diagnostics, as long as the syntax is valid we don't actually care (we just want to know what variables are defined, and if they have defaults defined).

This PR totally removes this use of github.com/hashicorp/terraform-config-inspect , instead parsing the Terraform / Tofu code *just* for variable definitions, ignoring everything else (our goal is validating the Terragrunt code anyway, validating the Tofu / Terraform code should be left to the respective tools).

Also reworded the docs on `--inputs`, as I found it a bit confusing.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Allow variables in module source / version

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Terraform module variable parsing and detection accuracy, including broader file-format support.

* **Tests**
  * Expanded test coverage with fixtures for edge cases: variables in module fields, circular references, invalid locals, and required/optional inputs. Added integration tests validating successful hcl validation.

* **Documentation**
  * Clarified HCL validation docs to state that modules require certain variables.

* **Chores**
  * Updated formatting config to exclude test fixtures from automatic formatting.

### DIFF
--- a/tf/tf.go
+++ b/tf/tf.go
@@ -4,7 +4,6 @@ import (
 	"os"
 	"path/filepath"
 	"slices"
-	"strings"
 
 	"github.com/gruntwork-io/terragrunt/internal/errors"
 	"github.com/hashicorp/hcl/v2"
@@ -153,16 +152,13 @@ func ModuleVariables(modulePath string) ([]string, []string, error) {
 
 		parseFunc := parser.ParseHCLFile
 
-		suffix := ""
-		for s := range strings.SplitSeq(file.Name(), ".") {
-			suffix = s
-		}
+		suffix := filepath.Ext(file.Name())
 
-		if suffix == "json" {
+		if suffix == ".json" {
 			parseFunc = parser.ParseJSONFile
 		}
 
-		if !(slices.Contains([]string{"tf", "tofu", "json"}, suffix)) {
+		if !(slices.Contains([]string{".tf", ".tofu", ".json"}, suffix)) {
 			continue
 		}
 


docs-starlight/src/data/flags/hcl-validate-inputs.mdx
```diff
@@ -6,7 +6,7 @@ env:
   - TG_INPUTS
 ---
 
-When enabled, Terragrunt will validate that all variables are set by the module a unit provisions.
+When enabled, Terragrunt will validate that all variables a module (provisioned by a unit) requires are set.
```
thisguycodes, line 9: Open to input on this, or even removing it. But I definitely don't think the current wording is correct (since we're checking that the unit sets variables consumed by the module, the module doesn't set them).

test/fixtures/hclvalidate/valid/var-in-source/main.tf
thisguycodes, line 1: I was able to make an even more minimal example than the one in https://github.com/gruntwork-io/terragrunt/issues/4986

test/fixtures/hclvalidate/valid/var-in-source/terragrunt.hcl
thisguycodes, line 1: I know empty files seem silly, but in the spirit of "minimal reproduction" I don't think you can get more minimal than empty!

test/integration_test.go
```diff
@@ -614,6 +614,47 @@ func TestTerragruntExcludesFile(t *testing.T) {
...
+
+	t.Run("using --all", func(t *testing.T) {
+		t.Parallel()
+		helpers.CleanupTerraformFolder(t, testFixtureHclvalidate)
+		tmpEnvPath := helpers.CopyEnvironment(t, testFixtureHclvalidate)
+		rootPath := util.JoinPath(tmpEnvPath, testFixtureHclvalidate)
+
+		_, _, err := helpers.RunTerragruntCommandWithOutput(t, "terragrunt hcl validate --all --strict --inputs --working-dir "+path.Join(rootPath, "valid"))
+		require.NoError(t, err)
+	})
```
thisguycodes, line 628: I added this test, but actually it was already passing! Is `--all` not supposed to return non-zero?

tf/tf.go
```diff
@@ -127,10 +129,35 @@ var (
...
+func diagnosticDoesNotAffectModuleVariables(d tfconfig.Diagnostic) bool {
+	ignorableErrors := []tfconfig.Diagnostic{
+		// These two occur when a module block uses a variable in the `version` or `source` fields.
+		// Terraform doesn't support this, but OpenTofu does. Either way our ability to extract variables is unaffected.
+		//
+		// What we really need is an OpenTofu version of terraform-config-inspect. This may work for now but as / if
+		// syntax continues to diverge we may run into other issues.
+		{Summary: "Variables not allowed", Detail: "Variables may not be used here."},
+		{Summary: "Unsuitable value type", Detail: "Unsuitable value: value must be known"},
+	}
```
thisguycodes: I wanted to keep this scoped to this function, plus putting it outside the function would require a more descriptive name.

I'm 90% sure the compiler notices it's never changed and pre-allocates it, but am happy to be proven wrong (and will move it if so).

tf/tf.go
```diff
@@ -127,10 +129,35 @@ var (
...
+	if d.Severity != tfconfig.DiagError {
+		return true
+	}
+
+	i := slices.IndexFunc(ignorableErrors, func(ie tfconfig.Diagnostic) bool {
```
denis256: I was thinking that this can be simplified to

```
	return slices.ContainsFunc(ignorableErrors, func(ie tfconfig.Diagnostic) bool {
		return ie.Summary == d.Summary && ie.Detail == d.Detail
	})

```

thisguycodes: That makes way more sense. I already know about contains so I don't know what I was thinking when I wrote this

docs-starlight/src/data/flags/hcl-validate-inputs.mdx
yhakbar: Would better wording be: "When enabled, Terragrunt will validate that all inputs set by Terragrunt units are valid OpenTofu/Terraform variables consumed by their respective modules"?

test/fixtures/hclvalidate/valid/var-in-source/terragrunt.hcl
yhakbar: Nothing wrong with it! If you want to be explicit about this, you can leave a one-line comment in the file that it's intentionally empty.

test/integration_test.go
yhakbar: It's entirely possible there's a bug in the error handling. I think it should be returning an error if anything fails during an `--all` run.

test/fixtures/hclvalidate/valid/var-in-source/main.tf
yhakbar, line 1: Love the names on these fixtures. Reading the name of the fixture should tell you what the fixture is for.

tf/tf.go
yhakbar: I don't think it's a big deal either way.

yhakbar (dismissed):
I think this does the job, so I'll approve it. It's not how I thought you'd tackle this.

I was assuming you'd do something like this untested code:

```go
type Variable struct {
	Name    string `hcl:"name,label"`
	Default string `hcl:"default,optional"`

	Remainder hcl.Body `hcl:",remain"`
}

type ModuleVars struct {
	Variables []Variable `hcl:"variable,block"`
}

// ModuleVariables will return all the variables defined in the downloaded terraform modules, taking into
// account all the generated sources. This function will return the required and optional variables separately.
func ModuleVariables(modulePath string) ([]string, []string, error) {
	parser := hclparse.NewParser()

	files, err := os.ReadDir(modulePath)
	if err != nil {
		return nil, nil, err
	}

	hclFiles := []*hcl.File{}
	errs := []error{}

	for _, file := range files {
		if file.IsDir() {
			continue
		}

		// This would only solve the issue for HCL files. We'd need to also check JSON files.
		if !strings.HasSuffix(file.Name(), ".tf") && !strings.HasSuffix(file.Name(), ".tofu") {
			continue
		}

		hclFile, err := parser.ParseFromFile(file.Name())
		if err != nil {
			errs = append(errs, err)
		}

		hclFiles = append(hclFiles, hclFile.File)
	}

	if len(errs) > 0 {
		return nil, nil, errors.New(errs)
	}

	body := hcl.MergeFiles(hclFiles)

	moduleVars := ModuleVars{}

	if err := gohcl.DecodeBody(body, nil, &moduleVars); err != nil {
		return nil, nil, err
	}

	required := []string{}
	optional := []string{}

	for _, variable := range moduleVars.Variables {
		if variable.Default != "" {
			optional = append(optional, variable.Name)
		}
	}

	return required, optional, nil
}
```

To fully drop the dependency on `github.com/hashicorp/terraform-config-inspect/tfconfig`.

It's a bigger change than what you're doing here, so I'll accept that you might prefer your current solution. Please confirm that you've evaluated this, and don't want to make this or the changes that @denis256 recommended before merging if you're happy with the solution.

docs-starlight/src/data/flags/hcl-validate-inputs.mdx
thisguycodes: I feel that only is accurate when using `--strict`, without `--strict` you can set inputs that are not consumed / don't exist as variables

thisguycodes:
@yhakbar I had considered it, and really would prefer it, but I had two thoughts:

1. I'd have to define the schema we expect, this duplication of schema (the "original" being inside terraform / opentofu) is a pattern that makes me even more itchy about drift.
    * The code as you wrote would produce error diagnostics on *anything that isn't a variable block* (it infers the schema from the struct tags, then "requires" that all the hcl match the schema) 
2. We'd be potentially allowing an even wider set of "invalid" Tofu / Terraform code than intended. Strictly speaking this code path is only meant to validate the Terragrunt files; but at present it also does some implicit validation of Tofu / Terraform code structure.

However [text in the terraform-config-inspect readme](https://github.com/hashicorp/terraform-config-inspect/blob/225c06ed5fd99341e37f199aed9c91e829c14be1/README.md?plain=1#L23-L26) made a really good point about the compatibility promise (alleviating the first point), and the presence of your suggestion mostly alleviates my concern for the second point (since it implies you're not concerned about this sort of implicit validation), so I went ahead and did it.

I also added more test cases, some I just thought of when thinking about this "implicit validation" thing; ~and one specifically that currently fails (using `tfconfig`), and will pass when we do our own parsing~ never mind I was wrong. I apparently cannot come up with a case that will behave the way I thought this would!

test/fixtures/hclvalidate/valid/duplicate-attributes-in-required-providers/main.tf
thisguycodes, line 1: this test case fails using the `tfconfig` package, but passes doing our own parsing

test/fixtures/hclvalidate/valid/duplicate-attributes-in-required-providers/main.tf
```diff
@@ -0,0 +1,6 @@
+terraform {
+  required_providers {
+    attr  = no
+    attr = yes
+  }
+}
```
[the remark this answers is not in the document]
thisguycodes: this terraform code is intentionally invalid

test/fixtures/hclvalidate/valid/duplicate-attributes-in-required-providers/main.tf
thisguycodes: I stand corrected! I swear I saw this pass locally, but apparently I was imagining things.

yhakbar (approved):
Looks good! Great job!

rlees85:
Awesome, thank you very much. I was really not expecting this problem to be resolved quickly or by somebody else.

PR #5049, opened 2025-11-03
feat: Adding support for graph expressions in `--filter`

yhakbar:
## Description

Adds support for graph expressions in `--filter` to allow users to selectively match components based on their relationships in the dependency graph.

This is a pretty massive PR, so I might end up breaking this down significantly to make it more digestible.

Changes:

- Changes the default logic of discovery so that it opts out of hidden configurations, instead of opting in (`NoHidden` instead of `Hidden`). I changed this because the majority of discovery logic actually requires hidden files to be supported, with explicit exceptions like `.git`, `.terragrunt-cache`, etc.
- Reorganized `Unit` and `Stack` structs in `component` into their own files so that the top-level interface is easier to reason about.
- Added a new `ThreadSafeComponents` struct that bundles a `components` field with a RW mutex for safe concurrent access. This is necessary for the concurrent component discovery we do for dependencies and/or dependents.
- Added a new `mu` onto the `Unit` and `Stack` structs so that they also have their dependencies/dependents accessed with synchronization controls to avoid duplicate graph edges.
- Broke `dependencydiscovery.go` into its own file so that we can reason about the logic there more easily.
- `DependencyDiscovery` now allows for a slice of `startingComponents` as the start of dependency discovery. This is used when the form `--filter 'foo...' is used, as we only want to do discovery for components that match the filter nested in the graph expression.
- `DependencyDiscovery` now also discovers dependencies concurrently, including discovered dependencies (which are dependencies of dependencies).
- The logic for determining if something is external is different now. Instead of checking to see if the component was in the initial set of components we were working with before, we instead check the filepath of the component to see if it's outside the filepath of the working directory where discovery started. This is important because we now allow for only some of the initially discovered components to have their dependencies discovered. As such, we might discover new components during dependency discovery that are not external.
- Added a new `DependentDiscovery`. This works differently than dependencies in that it traverses up the filesystem from a set of initial starting components up the filesystem until all components that depend on the starting components are discovered. 
- `DependentDiscovery` will concurrently find candidates that are possible dependents for each component being discovered, parse their dependencies to see if they are dependent, and add it to the discovered component if so. It also kicks off discovery both for that discovered dependent and in a parent directory, accordingly.
- The `Discovery` struct now tracks graph expressions pulled from `Filters`, and uses that for determining whether any dependency/dependent discovery needs to take place.
- Added a new `GraphExpression` struct in the `filter` package that stores parsed graph expressions (filter expressions that can traverse up and/or down the tree).
- Added new `RequiresDependencyDiscovery` and `RequiresDependentDiscovery` methods to `Filter` so that we can surface expressions that require discovery through the dependency graph to operate correctly.

## Limitations

There's an annoying hack that I had to include because of how we construct the runner pool:
https://github.com/gruntwork-io/terragrunt/pull/5049/files#diff-916e6863e1245c673b4e5965c98dc27bafbd72650fdb38ce65ea73ee6304e027R1136
https://github.com/gruntwork-io/terragrunt/pull/5049/files#diff-916e6863e1245c673b4e5965c98dc27bafbd72650fdb38ce65ea73ee6304e027R1281

Right now, the runner pool requires that all the components are discovered as top-level results, then it's supposed to filter them out in runs. Rather than fight that this PR, I'm going to open it with the knowledge that this is busted for filter flags, and look to address it by untangling that in the `runner` package.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Added support for graph expressions in `--filter`.

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added graph-based filtering syntax with new operators (`...`, `^`) for discovering components by their dependencies and dependents.
  * Enhanced component discovery system to better handle and traverse dependency relationships.

* **Improvements**
  * Improved visibility and tracking of external dependencies during discovery.
  * Added logging support to filter evaluation for better debugging and monitoring.
  * Refined handling of hidden components in the discovery process.

### DIFF
--- a/internal/discovery/dependencydiscovery.go
+++ b/internal/discovery/dependencydiscovery.go
@@ -151,14 +151,15 @@ func (dd *DependencyDiscovery) DiscoverDependencies(
 		return errors.New("expected Unit component but got different type")
 	}
 
-	if unit.Config() == nil {
+	cfg := unit.Config()
+	if cfg == nil {
 		err := Parse(dComponent, ctx, l, opts, dd.suppressParseErrors, dd.parserOptions)
 		if err != nil {
 			return errors.New(err)
 		}
-	}
 
-	cfg := unit.Config()
+		cfg = unit.Config()
+	}
 
 	depPaths, err := extractDependencyPaths(cfg, dComponent)
 	if err != nil {
--- a/internal/discovery/constructor.go
+++ b/internal/discovery/constructor.go
@@ -35,7 +35,8 @@ type HCLCommandOptions struct {
 // NewForDiscoveryCommand creates a Discovery configured for discovery commands (find/list).
 func NewForDiscoveryCommand(opts DiscoveryCommandOptions) (*Discovery, error) {
 	d := NewDiscovery(opts.WorkingDir).
-		WithSuppressParseErrors()
+		WithSuppressParseErrors().
+		WithBreakCycles()
 
 	if opts.NoHidden {
 		d = d.WithNoHidden()
--- a/internal/discovery/discovery.go
+++ b/internal/discovery/discovery.go
@@ -3,6 +3,7 @@ package discovery
 
 import (
 	"context"
+	stderrs "errors"
 	"io"
 	"io/fs"
 	"os"
@@ -160,6 +161,9 @@ type Discovery struct {
 
 	// filterFlagEnabled determines whether the filter flag experiment is active
 	filterFlagEnabled bool
+
+	// breakCycles determines whether to break cycles in the dependency graph if any exist.
+	breakCycles bool
 }
 
 // DiscoveryOption is a function that modifies a Discovery.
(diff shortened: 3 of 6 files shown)


internal/component/stack.go
```diff
@@ -0,0 +1,176 @@
...
+func NewStack(path string) *Stack {
+	return &Stack{
+		path:         path,
+		dependencies: make(Components, 0),
+		dependents:   make(Components, 0),
+	}
+}
+
+// NewStackWithConfig creates a new Stack component with the given path and config.
+func NewStackWithConfig(path string, cfg *config.StackConfig) *Stack {
```
denis256, line 36: Looks like `NewStackWithConfig()` is not used

internal/discovery/dependencydiscovery.go
```diff
@@ -0,0 +1,320 @@
...
+	dComponent component.Component,
+	depthRemaining int,
+) error {
+	if depthRemaining <= 0 {
+		return errors.New("max dependency depth reached while discovering dependencies")
+	}
+
+	// Stack configs don't have dependencies (at least for now),
+	// so we can return early.
+	if _, ok := dComponent.(*component.Stack); ok {
```
denis256, line 145: can be checked through Kind() ?

yhakbar: Ya, but I think this works out better because we'll know the actual type of the component, and if we wanted to get the config of a stack in discovery later, we can do it through this same mechanism.

internal/component/stack.go
yhakbar: I shouldn't have introduced it here, but we will need it eventually.

Issue #3474, opened 2024-10-15
Another hash mismatch prevents building with GOPROXY=direct


ChandlerSwift:
**Describe the bug**
I generally run `go` with `GOPROXY=direct`, which bypasses the Google-run module cache and downloads modules sources directly from their repos. Terragrunt fails to be built with this option set, due to a hash mismatch on a dependency.

I had reported a similar bug in https://github.com/gruntwork-io/terragrunt/issues/2900, but that one resolved itself with https://github.com/gruntwork-io/terragrunt/commit/d7423b88d7e75972ff459acf97bc507ae7572928#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R42. 

**To Reproduce**
I ran these commands in a current `golang` docker container, to make sure nothing in my environment was causing (additional) issues:

```sh
docker pull golang:latest
docker run -it golang:latest
```

```text
$ git clone --depth=1 https://github.com/gruntwork-io/terragrunt.git
[...]
$ cd terragrunt
$ git rev-parse HEAD
af89a982c98d50d8a0015016c70f311462d5c57f
$ GOPROXY=direct go build main.go
go: downloading github.com/go-errors/errors v1.4.2
[...]
go: downloading github.com/hashicorp/terraform-registry-address v0.2.0
go: downloading github.com/googleapis/enterprise-certificate-proxy v0.3.3
go: downloading github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da
go: downloading github.com/aymerick/douceur v0.2.0
go: downloading github.com/felixge/httpsnoop v1.0.4
verifying github.com/googleapis/enterprise-certificate-proxy@v0.3.3: checksum mismatch
	downloaded: h1:G6q7VHBoU74wQHXFsZSLMPl0rFw0ZDrlZ3rt6/aTBII=
	go.sum:     h1:QRje2j5GZimBzlbhGA2V2QlGNgL8G6e+wGo/+/2bWI0=

SECURITY ERROR
This download does NOT match an earlier download recorded in go.sum.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.
```

**Expected behavior**
This build should succeed.

**Additional context**
[...]

PR #5066, opened 2025-11-05
chore: Github action release improvements

denis256:
## Description

* moved MacOS signing from CircleCI to GitHub Actions
* moved release process to Github action
* added ZIPs and checksum files releasing
* added Windows executables signing and Icon setting
* moved no_proxy build to run on main branch, to save GH workers

Example artifacts publication:
https://github.com/gruntwork-io/terragrunt/releases/tag/beta-2025111005

https://github.com/gruntwork-io/terragrunt/releases/tag/alpha-2025111102
 

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **New Features**
  * New automated release workflow: detect releases, build, sign per-OS, merge signed artifacts, package archives, and generate checksums.
  * Multiple release utilities to prepare artifacts, set permissions, verify and upload release assets.

* **Chores**
  * Migrated CI and signing from legacy provider to GitHub Actions; removed legacy CI config.
  * Added ignore patterns for Windows signing resources.

* **Documentation**
  * Added comprehensive release scripts README.

### DIFF
--- a/.github/scripts/release/README.md
+++ b/.github/scripts/release/README.md
@@ -508,12 +508,11 @@ These scripts are used by:
 
 The scripts support multiple version tag formats:
 
-| Format | Example | Windows FileVersion | Description |
-|--------|---------|---------------------|-------------|
-| Standard | `v0.93.4` | `0.93.4.0` | Semantic version with v prefix |
-| Beta | `beta-2025111001` | `2025.1110.01.0` | Pre-release with date timestamp |
-| Alpha | `alpha-2025110301` | `2025.1103.01.0` | Alpha with date timestamp |
-| Generic | `<prefix>-YYYYMMDDNN` | `YYYY.MMDD.NN.0` | Any prefix with date |
+| Format   | Example            | Windows FileVersion | Description                     |
+|----------|--------------------|---------------------|---------------------------------|
+| Standard | `v0.93.4`          | `0.93.4.0`          | Semantic version with v prefix  |
+| Beta     | `beta-2025111001`  | `2025.1110.01.0`    | Pre-release with date timestamp |
+| Alpha    | `alpha-2025110301` | `2025.1103.01.0`    | Alpha with date timestamp       |
 
 **Windows Version Constraints:**
 - Each component must be ≤ 65535


.circleci/config.yml
```diff
@@ -1,153 +1,8 @@
...
-    GRUNTWORK_INSTALLER_VERSION: v0.0.39
-    MODULE_CI_VERSION: v0.57.0
-    TERRAFORM_VERSION: "1.11.3"
-    TFLINT_VERSION: "0.47.0"
-
-defaults: &defaults
-  docker:
-    - image: 087285199408.dkr.ecr.us-east-1.amazonaws.com/circle-ci-test-image-base:go1.25.0-tf1.5-tg58.8-pck1.8-ci58.2
-
+# CircleCI configuration - DEPRECATED
```
yhakbar: Why don't we just get rid of this?

.github/workflows/build.yml
```diff
@@ -4,8 +4,32 @@ on:
...
+          echo "Git ref: $REF"
+
+          # Check if this is a release tag (v*, alpha*, beta*)
+          if [[ "$REF" =~ ^refs/tags/v.* ]] || \
+             [[ "$REF" =~ ^refs/tags/alpha.* ]] || \
+             [[ "$REF" =~ ^refs/tags/beta.* ]]; then
+            echo "is-release=true" >> "$GITHUB_OUTPUT"
+            echo "This is a RELEASE build - signing will be enabled"
+          else
+            echo "is-release=false" >> "$GITHUB_OUTPUT"
```
yhakbar: NIT: I think it's more consistent with our other names to use snake case for stuff referenced in YAML, even though there's no agreed upon standard for YAML casing.

.github/workflows/ci.yml
```diff
@@ -42,20 +42,22 @@ jobs:
     secrets: inherit
 
   build_no_proxy:
+    # Only run no_proxy builds on main branch to save CI time
+    if: github.ref == 'refs/heads/main' || github.ref == 'refs/heads/master'
```
yhakbar: NIT: Why check for master?

.github/workflows/release.yml
```diff
@@ -0,0 +1,341 @@
...
+      - name: Check if release exists
+        id: check_release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VERSION: ${{ steps.version.outputs.version }}
+        run: |
+          echo "Checking if release exists for tag: $VERSION"
+
+          # Try to get the release using gh CLI
+          if ! gh release view "$VERSION" --json id,uploadUrl,isDraft > /dev/null 2>&1; then
```
yhakbar: NIT: I recommend single quoting the parameter for `--json`.

.github/workflows/release.yml
```diff
@@ -0,0 +1,341 @@
...
+            echo "exists=false" >> "$GITHUB_OUTPUT"
+            echo "Release not found for tag $VERSION"
+            exit 1
+          fi
+
+          # Get release details
+          RELEASE_JSON=$(gh release view "$VERSION" --json id,uploadUrl,isDraft)
+          RELEASE_ID=$(echo "$RELEASE_JSON" | jq -r '.id')
```
yhakbar: NIT: Useless `echo`. Prefer `jq -r '.id' <<< "$RELEASE_JSON"`

.github/workflows/release.yml
```diff
@@ -0,0 +1,341 @@
...
+          name: all-signed-binaries
+          path: bin/
+
+      - name: Verify binaries downloaded
+        run: |
+          echo "Downloaded binaries:"
+          ls -lahrt bin/*
+
+          # Count binaries
+          binary_count=$(ls bin/ | wc -l)
```
yhakbar: NIT: Prefer find over ls for this.

.github/workflows/release.yml
```diff
@@ -0,0 +1,341 @@
...
+          fi
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+          echo "Release version: $VERSION"
+
+      - name: Check if release exists
+        id: check_release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VERSION: ${{ steps.version.outputs.version }}
+        run: |
```
yhakbar: NIT: Avoid doing in-line bash like this in the workflow files. We don't get syntax highlighting, or the ability to use shellcheck, etc. and it makes for less testable logic.

.github/workflows/release.yml
```diff
@@ -0,0 +1,341 @@
...
+          chmod +x terragrunt_darwin_arm64
+          chmod +x terragrunt_linux_386
+          chmod +x terragrunt_linux_amd64
+          chmod +x terragrunt_linux_arm64
+          chmod +x terragrunt_windows_386.exe
+          chmod +x terragrunt_windows_amd64.exe
+
+          echo "Execution permissions set on all binaries"
+
+      - name: Create ZIP archives
```
yhakbar: If we're gonna start adding ZIP archives, let's also add `.tar.gz` files. That way we can preserve file ownership.

EDIT: Apparently `.zip` preserves UNIX permissions too now. I still think it's nice to have both.

.github/assets/winres.json
```diff
@@ -0,0 +1,78 @@
...
+            {
+              "supportedOS": {
+                "Id": "{1f676c76-80e1-4239-95bb-83d0f6d0da78}",
+                "comment": "Windows 8.1 / Windows Server 2012 R2"
+              }
+            },
+            {
+              "supportedOS": {
+                "Id": "{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}",
+                "comment": "Windows 10, Windows 11 / Windows Server 2016, 2019, 2022"
```
thisguycodes: do we claim / offer support for these versions? I'm particularly hesitant about Windows Vista / 7 / 8 / 8.1 / 10 and Server 2008, which are all EOL. I understand if the IDs don't allow us to be so granular, but the top two entries are EOL only versions.

.github/workflows/release.yml
```diff
@@ -0,0 +1,341 @@
...
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VERSION: ${{ steps.version.outputs.version }}
+        run: |
+          echo "Uploading assets to existing release $VERSION..."
+
+          # Upload all files using gh CLI
+          cd bin
+          for file in *; do
+            echo "Uploading $file..."
+            if gh release upload "$VERSION" "$file" --clobber; then
```
yhakbar: Should `--clobber` be on by default? I feel like this should be a boolean for the workflow dispatch, and default to false in the release triggered workflow.

.github/assets/winres.json
```diff
@@ -0,0 +1,78 @@
...
+          "0409": {
+            "Comments": "Standardize IaC and manage growing infra complexity: define units, stacks, cut repetition with includes/hooks, execute modules in dependency order across environments",
+            "CompanyName": "Gruntwork, Inc.",
+            "FileDescription": "Terragrunt - Orchestrate OpenTofu and Terraform at Scale",
+            "FileVersion": "0.0.0",
+            "InternalName": "terragrunt",
+            "LegalCopyright": "Copyright (C) 2025 Gruntwork, Inc.",
+            "OriginalFilename": "terragrunt.exe",
+            "ProductName": "Terragrunt",
+            "ProductVersion": "0.0.0"
```
thisguycodes: NIT: this section makes me wonder if we should generate this json at build / sign time rather than have it checked in. Not necessary for the first version.

We already update the version numbers, but things like the copyright year, description, etc are also things we should want to be consistent. I think generating it (rather than reading it and updating certain fields) will encourage us to centralize even the mundane info as metadata.

.github/workflows/release.yml
```diff
@@ -0,0 +1,341 @@
...
+          echo "Verifying all assets are accessible..."
+
+          # Get list of assets in the release
+          ASSETS=$(gh release view "$VERSION" --json assets --jq '.assets[].name')
+          asset_count=$(echo "$ASSETS" | wc -l)
+
+          echo "Found $asset_count assets in release"
+
+          # Expected files (binaries + individual ZIPs + SHA256SUMS)
+          expected_files=(
```
yhakbar: Can we derive this from a single source of truth? I feel like we have to synchronize a lot of different filename lists right now.

.github/workflows/release.yml
```diff
@@ -0,0 +1,341 @@
...
+            echo "Warning: Expected 15 files, found $asset_count"
+          fi
+
+      - name: Upload summary
+        if: always()
+        env:
+          VERSION: ${{ steps.version.outputs.version }}
+          RELEASE_ID: ${{ steps.check_release.outputs.release_id }}
+          IS_DRAFT: ${{ steps.check_release.outputs.is_draft }}
+        run: |
```
yhakbar: This is very nice. Let's generate this using a template and `envsubst` or something, though.

.github/workflows/ci.yml
```diff
@@ -42,20 +42,22 @@ jobs:
     secrets: inherit
 
   build_no_proxy:
+    # Only run no_proxy builds on main branch to save CI time
+    if: github.ref == 'refs/heads/main'
```
thisguycodes, line 46: can someone explain why we do this at all?

yhakbar (changes requested):
This is fantastic, and I'm excited to have this shipped! 🎉 

The three things that should block this from merging are:
1. There's no `.tar.gz` packaged archive in addition to `.zip`. Given that we're gonna be making a big change to our release assets here, it would be nice to have both.
2. We have multiple different sources of truth for the assets that we're uploading and expecting. It would be nice to have one configuration (maybe in a JSON file that's parsed with `jq` or something) to track all that so we can't forget to make an update somewhere.
3. We're doing a lot of logic inline in the workflows. I think it'll be easier to maintain this and debug if we have the logic moved to bash scripts that we call from the workflows.

.github/workflows/release.yml
thisguycodes: NIT: remove the `--json` option entirely since we don't use it here, and only care about the exit code.
```suggestion
          if ! gh release view "$VERSION" > /dev/null 2>&1; then
```

.github/workflows/sign-windows.yml
```diff
@@ -0,0 +1,351 @@
...
+
+      # Restore P12 client certificate and set SM_CLIENT_CERT_FILE
+      - name: Restore P12 client certificate
+        shell: pwsh
+        run: |
+          $ErrorActionPreference = 'Stop'
+
+          Write-Host "Restoring P12 client certificate from base64..."
+
+          $bytes = [Convert]::FromBase64String("${{ secrets.WINDOWS_SIGNING_P12_BASE64 }}")
```
thisguycodes: Does powershell support reading this from an env var? I know we control this value, but quote escape issues make me itch.

Based on other code I think this might work, after adding the env var above
```suggestion
          $bytes = [Convert]::FromBase64String($env:SM_CLIENT_CERT_BASE64)
```

.github/scripts/release/sign-windows.ps1
```diff
@@ -0,0 +1,278 @@
...
+    } else {
+        Write-Host "Signature verified successfully"
+    }
+}
+
+function Main {
+    # Verify environment variables
+    Assert-EnvVar "SM_HOST"
+    Assert-EnvVar "SM_API_KEY"
+    Assert-EnvVar "SM_CLIENT_CERT_FILE"
```
thisguycodes, line 329: I cannot actually tell where this is used. Is this an env var used by smctl?

.github/workflows/sign-windows.yml
```diff
@@ -0,0 +1,81 @@
...
+
+      # Verify smctl is available
+      - name: Verify smctl installation
+        shell: pwsh
+        run: .github/scripts/release/verify-smctl.ps1
+
+      # Restore P12 client certificate and set SM_CLIENT_CERT_FILE
+      - name: Restore P12 client certificate
+        shell: pwsh
+        run: .github/scripts/release/restore-p12-certificate.ps1 -Base64Certificate "${{ secrets.WINDOWS_SIGNING_P12_BASE64 }}"
```
thisguycodes: I know we control this value, but shell quoting issues make me itch
```suggestion
        run: .github/scripts/release/restore-p12-certificate.ps1 -Base64Certificate "$SM_P12_BASE64"
```

...after adding the env above

.github/assets/winres.json
denis256: Other Windows versions don't support `supportedOS` config

.github/workflows/ci.yml
denis256: Some teams run Go with GOPROXY=direct to fetch modules straight from the original VCS (GitHub, GitLab, etc.) so they always see the latest changes and avoid any caching or lag from proxies.
Exist cases when releases are re-published with different checksums

https://github.com/gruntwork-io/terragrunt/issues/3474

.github/scripts/release/sign-windows.ps1
denis256: variables `SM_*` are used by smctl

.github/scripts/release/create-archives.sh
```diff
@@ -0,0 +1,45 @@
...
+
+function main {
+  local -r bin_dir="${1:-bin}"
+
+  if [[ ! -d "$bin_dir" ]]; then
+    echo "ERROR: Directory $bin_dir does not exist"
+    exit 1
+  fi
+
+  cd "$bin_dir"
```
yhakbar, line 16: NIT: Consider using pushd in scripts instead of cd unless you want the caller to have a different working directory. It avoids bugs that can arise from side effects when calling a script.

.github/scripts/release/get-version.sh
```diff
@@ -0,0 +1,58 @@
...
+
+  if [ "$EVENT_NAME" = "workflow_dispatch" ]; then
+    version="$INPUT_TAG"
+
+    # Validate workflow_dispatch input
+    if [[ -z "$version" ]]; then
+      echo "ERROR: INPUT_TAG is empty for workflow_dispatch event" >&2
+      exit 1
+    fi
+  else
```
yhakbar, line 24: NIT: This is me being neurotic, but I usually consider an `else` a smell. We can have a `resolve_version` function or something and return from that with an early return to clean up this section.

.github/assets/release-assets-config.json
```diff
@@ -0,0 +1,69 @@
...
+      "description": "TAR.GZ archive"
+    }
+  ],
+  "additional_files": [
+    {
+      "name": "SHA256SUMS",
+      "description": "Checksums for all files"
+    }
+  ],
+  "counts": {
```
yhakbar, line 62: NIT: I think these values should be computed so that we can't forget to update them later if we add a new platform. I don't think making that change should block merging this PR, though.

.github/scripts/release/upload-assets.sh
```diff
@@ -0,0 +1,59 @@
...
+      echo "Uploaded $file"
+    else
+      echo "Upload failed for $file (will retry in verification)"
+    fi
+  done
+
+  echo "Upload phase completed"
+}
+
+function assert_env_var_not_empty {
```
yhakbar, line 49: NIT: You copy + paste this around. You could move this to a lib as well.

yhakbar: FYI this is a short way to make the assertion:

```bash
: ${VAR:?ERROR: VAR is a required environment variable}"
```

.github/scripts/release/verify-assets-uploaded.sh
```diff
@@ -0,0 +1,118 @@
...
+  fi
+
+  echo "Verifying all assets are accessible..."
+
+  # Get list of assets in the release
+  local assets
+  assets=$(gh release view "$VERSION" --json 'assets' --jq '.assets[].name')
+
+  local asset_count
+  asset_count=$(echo "$assets" | wc -l)
```
yhakbar, line 39: NIT: Useless `echo`

.github/scripts/release/verify-assets-uploaded.sh
```diff
@@ -0,0 +1,118 @@
...
+  # Get expected files from centralized configuration
+  local expected_files
+  mapfile -t expected_files < <(get_all_expected_files)
+
+  # Check each expected file
+  for expected_file in "${expected_files[@]}"; do
+    echo "Checking $expected_file..."
+
+    # Check if file exists in release
+    if ! echo "$assets" | grep -q "^${expected_file}$"; then
```
yhakbar, line 52: NIT: Useless echo

.github/scripts/release/verify-assets-uploaded.sh
```diff
@@ -0,0 +1,118 @@
...
+      echo "$expected_file not found in release, uploading..."
+
+      # Upload the missing file
+      if [ -f "$bin_dir/$expected_file" ]; then
+        local i
+        for ((i=0; i<MAX_RETRIES; i++)); do
+          if gh release upload "$VERSION" "$bin_dir/$expected_file" $clobber_flag; then
+            echo "Uploaded $expected_file"
+            break
+          else
```
yhakbar, line 62: NIT: We don't need this else.

.github/scripts/release/verify-binaries-downloaded.sh
```diff
@@ -0,0 +1,54 @@
...
+# Source configuration library
+# shellcheck source=lib-release-config.sh
+source "$(dirname "$0")/lib-release-config.sh"
+
+function main {
+  local -r bin_dir="${1:-bin}"
+
+  # Get expected count from configuration, or use parameter if provided
+  local expected_count
+  if [[ -n "${2:-}" ]]; then
```
yhakbar, line 17: NIT: Avoid leveraging parameters here like this. We should give it a name at the entry to the function. Also, we can have expected count be a function to avoid this else.

.github/scripts/release/verify-files.sh
```diff
@@ -0,0 +1,59 @@
...
+
+  if [[ ! -d "$bin_dir" ]]; then
+    echo "ERROR: Directory $bin_dir does not exist"
+    exit 1
+  fi
+
+  echo "Verifying required files..."
+
+  # Check macOS binaries
+  for file in terragrunt_darwin_amd64 terragrunt_darwin_arm64; do
```
yhakbar, line 19: NIT: We can compute this off the JSON file using the `os` field.

PR #5100, opened 2025-11-13
feat: Usage of common for discovery and runnerpool

denis256:
## Description

* consolidated units filtering in discovery
* major runner pool simplification
* removed legacy code to filter units

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **New Features**
  * Graph-target discovery to limit results to a target and its dependents.
  * Per-unit execution context and isolated unit output writers for safer concurrent runs.

* **Bug Fixes**
  * Consistent symlink-resolved path handling (improves macOS correctness).
  * Clearer pre-checks and error messages for backend migration flows.

* **Improvements**
  * Enhanced include/exclude and queue filtering behavior for more accurate discovery.

* **Tests**
  * Added coverage for graph-target, runner pool, and per-unit helpers.

### DIFF
--- a/internal/discovery/discovery.go
+++ b/internal/discovery/discovery.go
@@ -289,9 +289,7 @@ func (d *Discovery) WithOptions(opts ...interface{}) *Discovery {
 
 	for _, opt := range opts {
 		if p, ok := opt.(interface{ GetParseOptions() []hclparse.Option }); ok {
-			if po := p.GetParseOptions(); len(po) > 0 {
-				parserOptions = append(parserOptions, po...)
-			}
+			parserOptions = append(parserOptions, p.GetParseOptions()...)
 		}
 
 		if g, ok := opt.(interface{ GraphTarget() string }); ok {


cli/commands/list/list.go
```diff
@@ -209,7 +233,19 @@ func discoveredToListed(components component.Components, opts *Options) (ListedC
 		listedCfg.Dependencies = make([]*ListedComponent, len(c.Dependencies()))
 
 		for i, dep := range c.Dependencies() {
-			relDepPath, err := filepath.Rel(opts.WorkingDir, dep.Path())
+			// Ensure dependency path is absolute for reliable relativity calculation
+			depPath := dep.Path()
```
yhakbar: Wait, why is the path ever relative here? I feel like we should fix the point of ingestion to make it so that components reliably store their paths as absolute paths.

cli/commands/list/list.go
```diff
@@ -187,7 +199,19 @@ func discoveredToListed(components component.Components, opts *Options) (ListedC
 			}
 		}
 
-		relPath, err := filepath.Rel(opts.WorkingDir, c.Path())
+		// Ensure component path is absolute for reliable relativity calculation
+		componentPath := c.Path()
+		if !filepath.IsAbs(componentPath) {
+			var err error
+
+			componentPath, err = filepath.Abs(componentPath)
```
yhakbar: I think this is wrong. We never want to care about the PWD of the process. We need the path to always be relative to the `--working-dir` flag.

cli/commands/hcl/validate/validate.go
```diff
@@ -108,7 +108,7 @@ func RunValidate(ctx context.Context, l log.Logger, opts *options.TerragruntOpti
 		parseOpts := opts.Clone()
 		parseOpts.WorkingDir = c.Path()
 
-		if _, ok := c.(*component.Stack); ok {
+		if c.Kind() == component.StackKind {
```
yhakbar: Prefer type assertions.

cli/commands/find/find.go
```diff
@@ -156,15 +170,17 @@ func discoveredToFound(components component.Components, opts *Options) (FoundCom
...
+				unit := c.(*component.Unit)
 				if cfg := unit.Config(); cfg != nil && cfg.Exclude != nil {
 					foundComponent.Exclude = cfg.Exclude.Clone()
 				}
 			}
 		}
 
 		if opts.Include {
-			if unit, ok := c.(*component.Unit); ok {
+			if c.Kind() == component.UnitKind {
```
yhakbar: Prefer type assertions.

cli/commands/find/find.go
```diff
@@ -156,15 +170,17 @@ func discoveredToFound(components component.Components, opts *Options) (FoundCom
 		}
 
 		if opts.Exclude {
-			if unit, ok := c.(*component.Unit); ok {
+			if c.Kind() == component.UnitKind {
```
yhakbar: Prefer type assertions.

internal/component/stack.go
```diff
@@ -1,26 +1,40 @@
...
-	discoveryContext *DiscoveryContext
-	dependencies     Components
-	dependents       Components
-	external         bool
-	mu               sync.RWMutex
+	logger                log.Logger
+	report                *report.Report
+	discoveryContext      *DiscoveryContext
+	childTerragruntConfig *config.TerragruntConfig
+	workingDir            string // Working directory for this stack (only field needed from TerragruntOptions)
```
yhakbar: We can store this in `discoveryContext`

internal/component/stack.go
```diff
@@ -1,26 +1,40 @@
...
-	reading          []string
-	discoveryContext *DiscoveryContext
-	dependencies     Components
-	dependents       Components
-	external         bool
-	mu               sync.RWMutex
+	logger                log.Logger
+	report                *report.Report
+	discoveryContext      *DiscoveryContext
+	childTerragruntConfig *config.TerragruntConfig
```
yhakbar: What does this mean?

internal/component/stack.go
```diff
@@ -1,26 +1,40 @@
...
 type Stack struct {
-	cfg              *config.StackConfig
-	path             string
-	reading          []string
-	discoveryContext *DiscoveryContext
-	dependencies     Components
-	dependents       Components
-	external         bool
-	mu               sync.RWMutex
+	logger                log.Logger
```
yhakbar: I'm not a fan of this being here.

cli/commands/list/list.go
```diff
@@ -166,6 +166,17 @@ func discoveredToListed(components component.Components, opts *Options) (ListedC
 	listedComponents := make(ListedComponents, 0, len(components))
 	errs := []error{}
 
+	// Ensure working directory is absolute for reliable path relativity calculations
+	absWorkingDir := opts.WorkingDir
+	if !filepath.IsAbs(absWorkingDir) {
+		var err error
+
+		absWorkingDir, err = filepath.Abs(absWorkingDir)
```
yhakbar: I think this is wrong. We never want to care about the PWD of the process. We need the path to always be relative to the --working-dir flag.

cli/commands/list/list.go
```diff
@@ -209,7 +233,19 @@ func discoveredToListed(components component.Components, opts *Options) (ListedC
 		listedCfg.Dependencies = make([]*ListedComponent, len(c.Dependencies()))
 
 		for i, dep := range c.Dependencies() {
-			relDepPath, err := filepath.Rel(opts.WorkingDir, dep.Path())
+			// Ensure dependency path is absolute for reliable relativity calculation
+			depPath := dep.Path()
+			if !filepath.IsAbs(depPath) {
+				var err error
+
+				depPath, err = filepath.Abs(depPath)
```
yhakbar: I think this is wrong. We never want to care about the PWD of the process. We need the path to always be relative to the --working-dir flag.

cli/commands/find/find.go
```diff
@@ -21,6 +21,19 @@ import (
 
 // Run runs the find command.
 func Run(ctx context.Context, l log.Logger, opts *Options) error {
+	// Ensure WorkingDir is absolute to avoid filepath.Rel errors when comparing with absolute unit paths
+	if !filepath.IsAbs(opts.WorkingDir) {
+		absWorkingDir, err := filepath.Abs(opts.WorkingDir)
```
yhakbar: I think this is wrong. We never want to care about the PWD of the process. We need the path to always be relative to the --working-dir flag.

cli/commands/common/graph/graph.go
thisguycodes, line 91: nit: this separation of graph and stack options is confusing to read, especially since the above code demonstrates that they can both affect the same things.

internal/component/component.go
```diff
@@ -115,6 +127,11 @@ func (c Components) RemoveByPath(path string) Components {
 func (c Components) Paths() []string {
 	paths := make([]string, 0, len(c))
 	for _, component := range c {
+		// Skip components explicitly marked as excluded if they support Excluded().
+		if excludable, ok := component.(interface{ Excluded() bool }); ok && excludable.Excluded() {
```
thisguycodes:
```suggestion
		if excludable, ok := component.(*Unit); ok && excludable.Excluded() {
```
Since a component can currently only be a `Unit` or a `Stack`, I think it's more clear to just test directly for if it's a Unit (which has this method).

Alternatively, if it makes sense, the `Excluded()` method can be added to `Stack` so we can add it to the `Component` interface.

What I'm trying to avoid is "magic" behavior that isn't clearly communicated by the types; just testing for this method inline here means I could delete this method from `Unit` and it'd still compile, plus the test failures we'd see (if any) would error pretty far from location so it could be hard to debug.

internal/component/stack.go
```diff
@@ -181,3 +195,31 @@ func (s *Stack) Dependents() Components {
 
 	return s.dependents
 }
+
+// String renders this stack as a human-readable string.
+func (s *Stack) String() string {
```
thisguycodes, line 206: nit: adding an example output to the docstring would be helpful to readers

internal/component/stack.go
```diff
@@ -181,3 +195,31 @@ func (s *Stack) Dependents() Components {
...
+	workingDir := s.path
+	if s.Execution != nil && s.Execution.TerragruntOptions != nil {
+		workingDir = s.Execution.TerragruntOptions.WorkingDir
+	}
+
+	return fmt.Sprintf("Stack at %s:\n%s", workingDir, strings.Join(units, "\n"))
+}
+
+// FindUnitByPath finds a unit in the stack by its path.
+func (s *Stack) FindUnitByPath(path string) *Unit {
```
ThisGuyCodes: do we need to do symlink resolution here? Or do we only want exact matches?

internal/component/stack.go
```diff
@@ -181,3 +195,31 @@ func (s *Stack) Dependents() Components {
...
+	for _, unit := range s.Units {
+		if unit.Path() == path {
```
thisguycodes, line 225: Do we need to do symlink resolution here? Or do we only want exact matches?

internal/component/unit.go
```diff
@@ -188,3 +216,125 @@ func (u *Unit) Dependents() Components {
 
 	return u.dependents
 }
+
+// String renders this unit as a human-readable string for debugging.
+func (u *Unit) String() string {
```
thisguycodes, line 225: same as with stacks: an example in the docstring would be helpful for readers and users

internal/component/unit.go
```diff
@@ -188,3 +216,125 @@ func (u *Unit) Dependents() Components {
...
+	// Snapshot values under read lock to avoid data races
+	u.rLock()
```
thisguycodes, line 227:
```suggestion
	u.rLock()
	defer u.rUnlock()
```

use defer to keep the cleanup code close to the setup code

internal/component/unit.go
```diff
@@ -188,3 +216,125 @@ func (u *Unit) Dependents() Components {
...
+	excluded := false
+	assumeApplied := false
+
+	if u.Execution != nil {
+		excluded = u.Execution.FlagExcluded
+		assumeApplied = u.Execution.AssumeAlreadyApplied
+	}
+
+	u.rUnlock()
+
```
thisguycodes:
```suggestion
```

internal/component/unit.go
```diff
@@ -188,3 +216,125 @@ func (u *Unit) Dependents() Components {
...
+		path, excluded, assumeApplied, strings.Join(deps, ", "),
+	)
+}
+
+// AbsolutePath returns the absolute path of the unit.
+// If path conversion fails, returns the original path and logs a warning if a logger is available.
+func (u *Unit) AbsolutePath() string {
+	if filepath.IsAbs(u.path) {
+		return u.path
+	}
```
thisguycodes: `filepath.Abs()` already contains code that looks like this:
```Go
	if IsAbs(path) {
		return Clean(path), nil
	}
```

if we do this check first then converted paths and non-converted paths will be different in whether they are `Clean()`'d. If this is desired then fine (e.g. "if already absolute then don't touch it"), but if there isn't a perceived downside I'd rather the behavior be as similar as possible between the two.


```suggestion
```

if we simply delete these three lines then we'll get the behavior I'm asking for.

internal/component/unit_output.go
```diff
@@ -0,0 +1,64 @@
...
+
+// flusher is any writer that supports Flush() error.
+type flusher interface {
+	Flush() error
+}
+
+// parentWriterProvider is any writer that can provide its underlying parent writer.
+// This is used to create writer-based locks that serialize flushes to the same parent.
+type parentWriterProvider interface {
+	ParentWriter() io.Writer
```
thisguycodes, line 17: nit: if we don't actually want to give the ability to write to the parent writer, we can use a `unique.Handle[io.Writer]`. And I would also want to rename this to indicate the intended use:
```suggestion
	UniqueWriter() unique.Handle[io.Writer]
```
(the name is just a suggestion)

which gives us the comparability without giving access to the writer itself.

If we *do* want to give access to the parent writer, I'd argue we use the [io.ReaderFrom](https://pkg.go.dev/io#ReaderFrom) pattern to accomplish that instead. Among other things this works natively with `io.Copy`.

internal/component/unit_output.go
```diff
@@ -0,0 +1,64 @@
...
+		key = fmt.Sprintf("%p", pwp.ParentWriter())
+	}
+
+	mu := unitOutputLock(key)
+
+	mu.Lock()
+	defer mu.Unlock()
+
+	return writer.Flush()
+}
```
thisguycodes, line 64: I kind of feel like this should be implemented inside of `Unit.Flush()`? A few things immediately come to mind:

- do writes happen *only* when `.Flush()` is called? Or also (sometimes) when writes are made? If the later, then this synchronization is needed elsewhere as well.
- the *caller* of `Unit.Flush()` is responsible for knowing if the underlying writer is safe to write to

The immediate ideal setup that occurs to me would be that the underlying writer on the unit handles this synchronization (which at that level could be as simple as a mutex on some wrapper writer). That feels like it *might* be hard to track down all the places the writers are created? But however far down the stack we can push this seems better.

internal/discovery/dependentdiscovery.go
```diff
@@ -339,7 +339,8 @@ func (dd *DependentDiscovery) DiscoverDependents(
 
 				candidate.AddDependency(c)
 
-				if dep == target.Path() {
+				// Use resolved paths for comparison to handle symlinks (e.g., macOS /var -> /private/var)
+				if resolvePath(dep) == resolvePath(target.Path()) {
```
thisguycodes: nit: I wonder if it'd be reasonable at some point to do this path sanitization consistently closer to parse time rather than sprinkled throughout when we compare things 🤔 

this would probably require a universal "display vs execution" concept for all paths everywhere

internal/discovery/discovery.go
```diff
@@ -300,15 +301,21 @@ func (d *Discovery) SetParseOptions(options []hclparse.Option) {
...
 				parserOptions = append(parserOptions, po...)
 			}
 		}
+
+		if g, ok := opt.(interface{ GraphTarget() string }); ok {
+			if target := g.GraphTarget(); target != "" {
+				d = d.WithGraphTarget(target)
+			}
+		}
 	}
```
thisguycodes, line 300: this use of `interface{}` makes me nervous; where previously changing signatures or methods would make this break (and thus easy to find + fix) now this will compile fine but do nothing at runtime, and "doesn't do a thing" is much harder to find and fix than "does a thing it shouldn't".

I can see that the previous code also depended on an interface assertion, but at least the input type was defined (I also don't super love the previous code).

The use of adding graph targets when calling `WithOptions` isn't intuitive to me, so I can't give any suggestion on how to remove the need for this generic interface. But any kind of firmer type safety on expected methods would make me happier. If I knew more I might suggest something like the graph target should be part of the parse options, but that also sounds like asking for a bigger refactor.

I'd like to suggest making this a generic method somehow, but the `...` syntax has quirks with that, and generic struct methods aren't supported anyway.

internal/discovery/discovery.go
```diff
@@ -300,15 +301,21 @@ func (d *Discovery) SetParseOptions(options []hclparse.Option) {
...
-		if p, ok := opt.(common.ParseOptionsProvider); ok {
+		if p, ok := opt.(interface{ GetParseOptions() []hclparse.Option }); ok {
 			if po := p.GetParseOptions(); len(po) > 0 {
 				parserOptions = append(parserOptions, po...)
 			}
 		}
+
+		if g, ok := opt.(interface{ GraphTarget() string }); ok {
+			if target := g.GraphTarget(); target != "" {
+				d = d.WithGraphTarget(target)
```
thisguycodes, line 297: does this mean if multiple opts with a `GraphTarget()` are passed in the last one wins and the earlier ones are thrown out? Is that the behavior we want?

internal/discovery/discovery_test.go
```diff
@@ -14,6 +14,14 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func normPath(p string) string {
```
thisguycodes: this looks identical in logic to `resolvePath`, I get since that's an internal function we can't use it here, but could we name it the same and make it a copy paste (the logic is structured slightly differently) to make the identical-ness more obvious?

internal/discovery/queue_filters.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	return false
+}
+
+// unitFrom returns the underlying *component.Unit if the given component is a unit.
+func unitFrom(c component.Component) (*component.Unit, bool) {
+	u, ok := c.(*component.Unit)
+	return u, ok
+}
```
thisguycodes, line 62: not to align with a relatively recent Torvalds rant, I think this helper method does more to obscure what's going on than it does to make the intentions clear

internal/discovery/queue_filters.go
```diff
@@ -0,0 +1,349 @@
...
+
+					break
+				}
+			}
+		}
+	}
+}
+
+// restorePreIncluded re-applies prior inclusions by un-excluding units that were previously included.
+func restorePreIncluded(components component.Components, preIncluded map[string]struct{}) {
```
thisguycodes, line 243: nit: the presence of this function makes me feel like our logic to determine if a unit is included could me more centralized....

Problems like this have three categories as far as I understand:

- manually included stuff (various means)
- manually excluded stuff (various means)
- items included via automated means (i.e. discovery)

rather than marking and un-marking things repeatedly, hoping we get the order correct, I could envision functions like:

`isManuallyIncluded()` which goes through the various ways a unit can be manually included
`isManuallyExcluded()`

and a main `shouldExclude()` which has something like:
```Go
if isManuallyIncluded() {
  return False
} else if isManuallyExcluded() {
  return True
}

return False
```

As "includes override excludes" is my assumption. Though that might not be true. But centralizing into a single `shouldExclude()` esq method would make the intended logic / interactions of these options more clear.

internal/runner/common/log_test.go
thisguycodes, line 1: just double checking that removing this was intentional

internal/runner/runnerpool/builder_helpers.go
```diff
@@ -0,0 +1,238 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return discovered, nil
+}
+
+// maybeRetryDiscovery retries discovery without exclude-by-default if none were discovered
+// and modules-that-include / units-reading flags are set.
+func maybeRetryDiscovery(
```
thisguycodes: is there a way this could be eliminated, and instead the caller of this could do the check, remove the option, and re-run `runDiscovery`?

Alternatively, could this be refactored to do the check, remove the option, and run `runDiscovery` internally rather than duplicating the logic?

internal/runner/common/writer.go
```diff
@@ -0,0 +1,54 @@
...
+	writer.mu.Lock()
+	defer writer.mu.Unlock()
+
+	return writer.flushUnsafe()
+}
+
+// flushUnsafe flushes buffer without acquiring a lock (useful when lock is held).
+func (writer *UnitWriter) flushUnsafe() error {
+	if writer.out != nil {
+		if _, err := writer.buffer.WriteTo(writer.out); err != nil {
```
thisguycodes, line 42: use `io.Copy` instead
```suggestion
		if _, err := io.Copy(writer.out, writer.buffer); err != nil {
```

This does e.g. `WriteTo` logic for us, and will benefit from future changes automatically.

internal/runner/common/writer.go
```diff
@@ -0,0 +1,54 @@
...
+	return &UnitWriter{
+		out: out,
+	}
+}
+
+func (writer *UnitWriter) Write(p []byte) (int, error) {
+	writer.mu.Lock()
+	defer writer.mu.Unlock()
+
+	return writer.buffer.Write(p)
```
thisguycodes, line 28: this *may* write to `writer.out`, so actually needs the same "same underlying writer" locking as way back in this PR.

I feel we should delete the `ParentWriter()` method entirely, and instead implement the lock-per-underlying-writer logic here, in both the `Write()` *and* `Flush()` methods.

The per-path locking of `FlushOutput` may still be useful (and if not, we should just delete that function entirely), and *also* holding that lock won't cause harm.

internal/component/stack.go
denis256: Path value is already resolved at this point

internal/runner/common/log_test.go
denis256: Yes

internal/discovery/queue_filters.go
denis256: Will handle in separated PR

internal/discovery/discovery.go
ThisGuyCodes: just double checking on this question

internal/component/component.go
```diff
@@ -183,20 +183,43 @@ func (c Components) CycleCheck() (Component, error) {
...
+
+// resolvedPathFor returns the cached resolved path for a component path if present,
+// otherwise resolves the path on the fly without mutating the cache.
+// Caller must hold at least a read lock.
+func (tsc *ThreadSafeComponents) resolvedPathFor(path string) string {
+	if resolved, ok := tsc.resolvedPaths[path]; ok {
+		return resolved
+	}
+
+	return resolvePath(path)
```
thisguycodes, line 217: caching these resolutions seems good on its face. But the structure and options all feels confusing:
- The cache is struct local (why not global?)
- the "resolve with cache" method doesn't populate on misses
- the package-level function doesn't use cache

There also appear to be places we don't use the cache that maybe we should? (e.g. line 4 of `addComponent`).

I see two possibilities.
I'm defining "common" as "more often % wise than the inverse of the performance difference between the two codepaths". So if checking / pulling from cache is 20x faster than doing the resolution, being 5% or more often is "common". I'd estimate pulling from a mutex'd map to be ~100x faster than doing the resolution.

- adding components that resolve to the same path is common, so cache provides a benefit

In this case I'd argue the package-level method should provide (and auto-populate) a global cache, and all struct-level logic for when to populate and use the cache gets ripped out.

Or:
- adding components that resolve to the same path is uncommon

In this case the potential benefit here is so rare the complexity isn't worth the benefit, and we should just remove the caching entirely.

internal/component/component_test.go
```diff
@@ -137,3 +138,106 @@ func TestComponentsCycleCheck(t *testing.T) {
...
+		wg.Add(1)
+
+		go func() {
+			defer wg.Done()
+
+			for range 100 {
+				s := unit.String()
+				assert.Contains(t, s, "/test/path")
+			}
+		}()
```
thisguycodes, line 163: nit: added in go 1.25, `waitGroup.Go`

```suggestion
		wg.Go(func() {
			for range 100 {
				s := unit.String()
				assert.Contains(t, s, "/test/path")
			}
		})
	}
```

internal/component/component_test.go
```diff
@@ -137,3 +138,106 @@ func TestComponentsCycleCheck(t *testing.T) {
...
+	for range goroutines {
+		wg.Add(1)
+
+		go func() {
+			defer wg.Done()
+
+			for range 100 {
+				_ = tsc.FindByPath("/test/path")
+				_ = tsc.Len()
+				_ = tsc.ToComponents()
```
thisguycodes, line 234: this might get compiled out?

if you declare variables right before the loop and re-assign to them every loop I think that stops it from doing that

internal/component/unit_output.go
thisguycodes, line 64: If we think this synchronization is important (which I'm not sure it is), then we actually need to handle this for all writes. Otherwise I don't understand what we're solving for.

Is `FlushOutput` somehow the only form of writing / flushing that may be called concurrently? Unless this is the case, I think we should change the `Writer` and `ErrWriter` on `TerragruntOptions` from an interface to a struct type of some `SyncWriter` that does this per-underlying-writer synchronization. Doing it here with a function we need to remember to use feels very strange and unsafe.

internal/component/component.go
```diff
@@ -166,20 +183,43 @@ func (c Components) CycleCheck() (Component, error) {
 
 // ThreadSafeComponents provides thread-safe access to a Components slice.
 // It uses an RWMutex to allow concurrent reads and serialized writes.
+// Resolved paths are cached to avoid repeated filepath.EvalSymlinks syscalls
+// and ensure consistent symlink-aware comparisons across all methods.
 type ThreadSafeComponents struct {
-	components Components
-	mu         sync.RWMutex
+	resolvedPaths map[string]string
```
thisguycodes, line 189: on its face this caching seems like a good thing. But if it's helpful I don't understand why it's struct-local instead of global, and it doesn't seem consistently used (e.g. line 4 of `addComponent` doesn't use the cache).

I vote we either rip out the caching entirely, or move it to be global (package level). In particular this "fetch from cache but don't populate on misses" option is confusing to me.

internal/component/unit_output.go
thisguycodes, line 64: if this flush synchroniation is something we actually need to solve, then we need to solve it for writes as well as flushes.

Unless somehow this `FlushOutput` function is the only function that may get called concurrently?

To solve this for all writes and flushes we need to do it at the unit. We could do this by changing the `Writer` and `ErrWriter`s from interfaces to concrete structs, say `SyncWriter`, which wraps the underlying writer and does this per-underlying-writer synchronization.

internal/discovery/discovery.go
```diff
@@ -283,13 +284,17 @@ func (d *Discovery) SetParseOptions(options []hclparse.Option) {
...
-		if p, ok := opt.(common.ParseOptionsProvider); ok {
-			if po := p.GetParseOptions(); len(po) > 0 {
-				parserOptions = append(parserOptions, po...)
+		if p, ok := opt.(interface{ GetParseOptions() []hclparse.Option }); ok {
+			parserOptions = append(parserOptions, p.GetParseOptions()...)
+		}
+
+		if g, ok := opt.(interface{ GraphTarget() string }); ok {
+			if target := g.GraphTarget(); target != "" {
+				d = d.WithGraphTarget(target)
```
thisguycodes, line 297: double checking on this question?

internal/discovery/discovery.go
```diff
@@ -283,13 +284,17 @@ func (d *Discovery) SetParseOptions(options []hclparse.Option) {
 // WithOptions ingests runner options and applies any discovery-relevant settings.
 // Currently, it extracts HCL parser options provided via common.ParseOptionsProvider
 // and forwards them to discovery's parser configuration.
-func (d *Discovery) WithOptions(opts ...common.Option) *Discovery {
+func (d *Discovery) WithOptions(opts ...interface{}) *Discovery {
```
thisguycodes, line 287: nit: still don't like this use of generic interface. If we can make this some kind of concrete type we can make this safer to use and expand.

internal/component/component.go
thisguycodes, line 189: if this cache is useful (which I'm not sure it is, how often to multiple components have the same underlying path?) then can we make it global instead?

The interaction of a struct-level cache along with both the struct and package level functions to resolve paths is confusing. Not to mention the decision to search the path but not populate it on misses?

I feel either the cache helps, and so we make it global and auto-populated (with a mutex or `sync.Map`), or it doesn't really help and we eliminate the complexity of it.

internal/component/unit_output.go
denis256: In case of JSON output, we don't want to mix outputs from different units, so we synchronize flushing

PR #5199, opened 2025-12-10
bug: runner pool external dependencies inclusion fix

### LINKED ISSUE #5195: Terragrunt v0.94.0 Regression: Incorrectly discovers Terragrunt modules outside target scope
## Describe the bug

Terragrunt v0.94.0 introduces a critical regression where the `run --all` command incorrectly discovers and executes against Terragrunt modules outside the intended scope (sibling and parent directories) instead of only within the current directory hierarchy. This causes unintended operations across wrong Terragrunt modules, including dangerous operations like resource deletions.

### Terragrunt Version

- **Broken version**: v0.94.0
- **Last working version**: v0.93.13

## Steps To Reproduce
1. Set up a modular Terragrunt configuration with multiple Terragrunt modules
2. Navigate to a specific Terragrunt module directory (e.g., `./deploy/aws/terraform/bastion`)
3. Run the command:

  ```bash
  terragrunt run --all destroy --non-interactive
  ```

## Expected Behavior

**v0.93.13 (correct behavior):**

- Command executes only in the current directory hierarchy
- Output shows: `Unit .` only
- Operations are scoped to Terragrunt modules within the current unit as intended
- Does NOT traverse to sibling or parent directories outside the target scope

## Actual Behavior

**v0.94.0 (incorrect behavior):**

- Command discovers and attempts to execute against Terragrunt modules in sibling and parent directories
- Output shows: `Unit ../<module> Unit ../<module> Unit ../<module>/<module>`
- Operations execute across unintended Terragrunt modules outside the target scope
- **Critical impact**: Causes deletion of AWS resources (e.g., KMS keys) in wrong Terragrunt modules that should not be affected

#### v0.93.13 (Working)

```text
Unit .
```

#### v0.94.0 (Broken)

```text
Unit ../<module>
Unit ../<module>
Unit ../<module>/<module>
```

**Directory structure:**

```text
deploy/aws/terraform/
├── bastion/
│   ├── terragrunt.hcl
│   └── ...
├── module1/
│   ├── terragrunt.hcl
│   └── ...
└── module2/
    ├── terragrunt.hcl
    └── ...
```

## Versions

- **Terraform Version**: v1.14.1
- **OS**: Linux (Ubuntu 24.04.3 LTS in dev container)
- **Shell**: bash
- **Execution context**: AWS infrastructure deployment

## Impact

This regression has severe consequences:

✗ **Data loss risk**: Unintended resource deletions (KMS keys, etc.)
✗ **Security risk**: Operations executed in wrong Terragrunt modules
✗ **Breaking change**: Previously working workflows now fail or cause damage
✗ **No warning**: Silent discovery of unintended Terragrunt modules without user consent

**Command executed from:**

```bash
cd ./deploy/aws/terraform/bastion
terragrunt run --all destroy --non-interactive
```

**Expected scope:** Only `bastion/` module
**Actual scope (v0.94.0):** Multiple Terragrunt modules discovered incorrectly, leading to **actual resource deletions** in wrong modules

### Workaround

Downgrade to v0.93.13:

```bash
wget https://github.com/gruntwork-io/terragrunt/releases/download/v0.93.13/terragrunt_linux_amd64
chmod +x terragrunt_linux_amd64
sudo mv terragrunt_linux_amd64 /usr/local/bin/terragrunt
```

## Additional context

- This issue was discovered during a routine `destroy` operation on a specific Terragrunt module
- The incorrect behavior was confirmed by comparing output between v0.93.13 and v0.94.0
- **Critical**: This caused actual resource deletions (AWS KMS keys, etc.) across unintended Terragrunt modules in production-like environments

### Request

Please investigate and fix this critical regression in the module discovery logic. This breaks existing workflows and poses significant risk to production infrastructure.

denis256:
## Description

* Implemented changes:
Logic v0.93.x    : discover all -> then filter out based on flags
Implemented logic in PR: only discover what you actually need
* Report of ignored dependencies
* Added basic integration test to track this issue

Fixes #5195.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Release Notes

* **Bug Fixes**
  * Fixed external dependency handling—external dependencies are now properly excluded by default from the queue. Use the `--queue-include-external` flag to include external dependencies when necessary.

* **Documentation**
  * Updated documentation for queue flags to clarify default behavior and reflect deprecation of `--queue-exclude-external`.

### DIFF
--- a/docs-starlight/src/content/docs/03-features/02-stacks.mdx
+++ b/docs-starlight/src/content/docs/03-features/02-stacks.mdx
@@ -23,9 +23,9 @@ Terragrunt supports two approaches to defining stacks:
 1. **Implicit Stacks**: Created by organizing units in a directory structure.
 2. **Explicit Stacks**: Defined using `terragrunt.stack.hcl` files.
 
-## Implicit Stacks: Directory-Based Organization
+## Implicit Stacks
 
-The simplest way to create a stack is to organize your units in a directory structure. When you have multiple units in a directory, Terragrunt automatically treats that directory as a stack.
+The simplest way to create a stack is to organize your units in a directory structure in your repository. When you have multiple units in a directory, Terragrunt automatically treats that directory as a stack for the purposes of commands like `terragrunt run --all apply`.
 
 ### Converting Terraform Modules to Units
 
@@ -100,7 +100,7 @@ terragrunt run --graph apply
 ### Advantages of Implicit Stacks
 
 - **Simple**: Just organize units in directory trees.
-- **Familiar**: Works like recommended best practices for OpenTofu/Terraform module organization.
+- **Familiar**: Organized following best practices for OpenTofu/Terraform repository structures.
 - **Flexible**: Easy to add/remove units by creating/deleting directories.
 - **Version Control Friendly**: Each unit is a separate directory with its own history.
 - **Backwards Compatible**: This has been the default way to work with Terragrunt for over eight years, and the majority of existing Terragrunt configurations use this approach.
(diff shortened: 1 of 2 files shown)


cli/flags/shared/shared.go
```diff
@@ -111,7 +111,7 @@ func NewQueueFlags(opts *options.TerragruntOptions, prefix flags.Prefix) cli.Fla
 				Name:        QueueExcludeExternalFlagName,
 				EnvVars:     tgPrefix.EnvVars(QueueExcludeExternalFlagName),
 				Destination: &opts.IgnoreExternalDependencies,
-				Usage:       "Ignore external dependencies for --all commands.",
+				Usage:       "DEPRECATED: External dependencies are now excluded by default. Use --queue-include-external to include them.",
```
yhakbar: Let's just add a strict control as the Action for the flag, and hide the flag.

test/fixtures/regressions/5195-scope-escape/bastion/.terraformrc
yhakbar, line 1: Probably want to get rid of this.

test/fixtures/regressions/5195-scope-escape/module2/.terraformrc
yhakbar, line 1: Probably want to get rid of this.

cli/flags/shared/shared.go
yhakbar: I did that here, which might help with the language for this:
https://github.com/gruntwork-io/terragrunt/pull/5189

HughesyMH:
Thanks for addressing this! I’ve checked the description and it seems to solve the issue I raised.

internal/runner/runnerpool/runner.go
```diff
@@ -909,6 +899,10 @@ func FilterDiscoveredUnits(discovered component.Components, units []*component.U
 // WithOptions updates the stack with the provided options.
 func (r *Runner) WithOptions(opts ...common.Option) *Runner {
 	for _, opt := range opts {
+		if opt == nil {
+			continue
+		}
+
```
thisguycodes, line 905: this feels so weird, we shouldn't be passing in nil options

internal/runner/runnerpool/runner_test.go
```diff
@@ -35,7 +35,7 @@ func TestDiscoveryResolverMatchesLegacyPaths(t *testing.T) {
 
 	l := thlogger.CreateLogger()
 
-	runner, err := runnerpool.NewRunnerPoolStack(context.Background(), l, opts, discovered)
+	runner, err := runnerpool.NewRunnerPoolStack(context.Background(), l, opts, discovered, nil)
```
thisguycodes, line 38: since the `nil` just executes `continue`, I don't understand why we need this here

ThisGuyCodes (approved):
looks fine, but I really don't understand (or like) the nil option thing. It appears to do nothing, but I trust it does something or we wouldn't have added it. Seems like a footgun waiting to go off.

PR #5241, opened 2025-12-16
chore: Backporting queue flags into filter flags

yhakbar:
## Description

Backports queue flags into filter flags. Makes the `--filter` flag generally available.

There are some breaking changes in here that require a chained PR out.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **New Features**
  * Added support for filter-based dependency and unit selection across commands.
  * Introduced configuration for handling external dependencies through filter expressions.

* **Bug Fixes**
  * External dependencies are now excluded from runs by default.

* **Deprecations**
  * Removed legacy include/exclude directory options and flags.
  * Removed `--external` flag behavior; use filters instead.
  * Hidden several internal flags from help output.

* **Documentation**
  * Updated run report schema to v2 with improved structure.


cli/flags/shared/shared.go
```diff
@@ -147,47 +154,91 @@ func NewQueueFlags(opts *options.TerragruntOptions, prefix flags.Prefix) cli.Fla
...
+				Usage:   "Unix-style glob of directories to include from the queue of Units to run.",
+				Action: func(_ *cli.Context, value []string) error {
+					if len(value) == 0 {
+						return nil
+					}
+
+					for _, v := range value {
+						// We explicitly wrap the value in curly braces to ensure that it is treated
+						// as a path expression, and not a name filter.
+						opts.FilterQueries = append(opts.FilterQueries, "{"+v+"}")
```
denis256, line 200: I suspect this may lead to issues if the included unit has dependencies - with `{"+v+"}`, dependencies will be ignored

cli/app_test.go
```diff
@@ -295,14 +558,50 @@ func TestFilterTerragruntArgs(t *testing.T) {
...
+				"/some/path/" + config.DefaultTerragruntConfigPath,
+			},
+			expected: []string{"plan", "-bar", "-baz"},
+		},
+		{
+			args:     []string{"run", "--all", "apply", "plan", "bar"},
+			expected: []string{tf.CommandNameApply, "plan", "bar"},
+		},
+		{
+			args:     []string{"run", "--all", "destroy", "--", "plan", "-foo", "--bar"},
```
denis256, line 598: Duplicates on cli/app_test.go:598 and cli/app_test.go:602

cli/flags/shared/shared.go
yhakbar: That's an intended breaking change here. Users should use `{./**}...` if they _want_ dependencies to be included.

cli/app_test.go
yhakbar: Removed

denis256:
Notes from review:
* --queue-exclude-external now effectively does nothing
* Dependency discovery is no longer invoked for standard runs: the constructor stopped enabling it and Discover only runs it when dependency graph filters exist
* `terragrunt find --external` no longer delivers external components. The flag is hidden and only appends a filter, but discovery never walks external deps anymore


I think changes should be released as a minor version increment

yhakbar:
That's not true, though. External dependencies will be pulled in if you use the replacement filter. That's why I aliased the flag to it.

PR #5264, opened 2025-12-19
chore: Remove `RunTarget` from `options`

yhakbar:
## Description

Removes the `RunTarget` lambda function in `options`.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Release Notes

* **Refactor**
  * Reorganized internal execution pipeline to improve code architecture and separation of concerns.
  * Made previously public internal types private to establish clearer API boundaries.
  * Consolidated preparation and initialization workflows for command execution.

* **Breaking Changes**
  * Removed `RunTerragrunt` field from public configuration options. Direct usage of this field will need to be updated.

### DIFF
--- a/internal/cli/commands/exec/exec.go
+++ b/internal/cli/commands/exec/exec.go
@@ -11,7 +11,6 @@ import (
 	"github.com/gruntwork-io/terragrunt/pkg/config"
 	"github.com/gruntwork-io/terragrunt/pkg/log"
 	"github.com/gruntwork-io/terragrunt/pkg/options"
-	"github.com/urfave/cli"
 )
 
 func Run(ctx context.Context, l log.Logger, opts *options.TerragruntOptions, cmdOpts *Options, args clihelper.Args) error {
@@ -57,7 +56,7 @@ func runTargetCommand(
 	cfg *config.TerragruntConfig,
 	r *report.Report,
 	cmdOpts *Options,
-	args cli.Args,
+	args clihelper.Args,
 ) error {
 	var (
 		command = args.CommandName()


cli/commands/hcl/validate/validate.go
```diff
@@ -295,9 +295,26 @@ func RunValidateInputs(ctx context.Context, l log.Logger, opts *options.Terragru
...
 
-		target := run.NewTarget(run.TargetPointGenerateConfig, runValidateInputs)
+		prepared, err := run.PrepareConfig(ctx, l, unitOpts)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+
+		// Download source
+		updatedOpts, err := run.PrepareSource(ctx, prepared.Logger, prepared.UpdatedOpts, prepared.TerragruntConfig, report.NewReport())
```
denis256: I’m not sure we should create a new report on each loop iteration (report.NewReport()). It probably makes more sense to use a single shared report instance and append each result to it.

internal/runner/run/prepare.go
```diff
@@ -0,0 +1,161 @@
...
+
+	terragruntConfig, err := config.ReadTerragruntConfig(ctx, l, opts, config.DefaultParserOptions(l, opts))
+	if err != nil {
+		return nil, err
+	}
+
+	return &PreparedConfig{
+		TerragruntConfig: terragruntConfig,
+		Logger:           l,
+		UpdatedOpts:      opts,
```
denis256: `UpdatedOpts` misleading name - `opts` is not updated, just copied from opts

internal/runner/runfn/runfn.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/gruntwork-io/terragrunt/internal/report"
+	"github.com/gruntwork-io/terragrunt/options"
+	"github.com/gruntwork-io/terragrunt/pkg/log"
+)
+
+// Func is the function signature for running terragrunt commands.
+type Func func(ctx context.Context, l log.Logger, opts *options.TerragruntOptions, r *report.Report) error
+
+// Run is the global run function that will be set by the run package during initialization.
+var Run Func
```
denis256, line 17: A global mutable variable is risky—it introduces hidden state and can lead to concurrency problems

internal/runner/run/prepare.go
yhakbar: ObtainAndUpdateEnvIfNecessary will update it, for example. It's not easy to see where opts is updated because of how much we pass it around.

internal/runner/runfn/runfn.go
yhakbar: I agree with you in principle, but it's only set in one place in the codebase, and it's only used to avoid a cyclic import:

```bash
$ rg runfn.Run
config/dependency.go
1182:   err := runfn.Run(ctx, l, pctx.TerragruntOptions, report.NewReport())

internal/runner/common/unit_runner.go
82:     runErr := runfn.Run(ctx, runner.Unit.Execution.Logger, opts, r)
155:            if err := runfn.Run(ctx, l, jsonOptions, adhocReport); err != nil {

cli/commands/commands.go
406:    runfn.Run = run.Run

```

I'm working on entirely removing this package, but it's more invasive:
https://github.com/gruntwork-io/terragrunt/pull/5359

PR #5361, opened 2026-01-14
chore: add internal-tflint strict control

thisguycodes:
## Description

Starts the deprecation process for the embeded version of `tflint`.

Tflint is tainted by the BUSL license in Terraform; this prevents us from being able to keep the embedded version updated; fortunately our hooks system makes running a local version of tflint pretty easy!

This PR creates a new strict control that forces the use of external tflint, maintaining the automatic `tflint init` and passing of variables.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Add strict control for deprecation of embedded tflint, our ability to embed having been hampered by the BUSL license change in Terraform.

### Migration Guide

No breaking changes right now. But if you use the embedded `tflint` right now you can ensure no breakages when we later remove it by adding the `--terragrunt-external-tflint` flag:

```hcl
terraform {
  before_hook "before_hook" {
    commands     = ["apply", "plan"]
    execute      = ["tflint", "--terragrunt-external-tflint"]
  }
}
```

Or by enabling the `legacy-internal-tflint` strict control.

When the embedded `tflint` is later removed, we will still strip this flag, perform automatic `tflint init`, and pass in variables when running the external tool.

## Summary by CodeRabbit

* **New Features**
  * Added a strict control to select internal vs external TFLint behavior and a command-line opt-in to force external TFLint.

* **Documentation**
  * Updated TFLint integration docs with deprecation notice for the internal mode, usage examples, and guidance for opting into external TFLint.

### DIFF
--- a/internal/tflint/tflint.go
+++ b/internal/tflint/tflint.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	"github.com/gruntwork-io/terratest/modules/collections"
@@ -30,8 +31,13 @@ const (
 
 // RunTflintWithOpts runs tflint with the given options and returns an error if there are any issues.
 func RunTflintWithOpts(ctx context.Context, l log.Logger, opts *options.TerragruntOptions, cfg *runcfg.RunConfig, hook *runcfg.Hook, externalTfLint bool) error {
+	hookExecute := slices.Clone(hook.Execute)
+	hookExecute = slices.DeleteFunc(hookExecute, func(arg string) bool {
+		return arg == TfExternalTFLint
+	})
+
 	// try to fetch configuration file from hook parameters
-	configFile, err := tflintConfigFilePath(l, opts, hook.Execute)
+	configFile, err := tflintConfigFilePath(l, opts, hookExecute)
 	if err != nil {
 		return err
 	}
@@ -55,7 +61,7 @@ func RunTflintWithOpts(ctx context.Context, l log.Logger, opts *options.Terragru
 		return errors.New(err)
 	}
 
-	tflintArgs := hook.Execute[1:]
+	tflintArgs := hookExecute[1:]
 
 	// tflint init
 	initArgs := []string{"tflint", "--init", "--config", configFile, "--chdir", opts.WorkingDir}


internal/runner/run/hook.go
```diff
@@ -178,27 +179,35 @@ func runHook(
...
-		)
-		if possibleError != nil {
-			l.Errorf("Error running hook %s with message: %s", curHook.Name, possibleError.Error())
-			return possibleError
+		if containsExternalTFLintFlag {
+			// delete the flag and continue to execute as a normal hook
+			actionParams = slices.DeleteFunc(actionParams, tflint.IsExternalTFLintFlag)
+		} else {
+			forceExternalTFLint := (opts.StrictControls.Find(controls.InternalTFLint).Evaluate(ctx) != nil)
+			if forceExternalTFLint {
```
yhakbar: You can invert this conditional to do an early return instead.

internal/runner/run/hook.go
```diff
@@ -178,27 +179,35 @@ func runHook(
...
-			actionToExecute, actionParams...,
-		)
-		if possibleError != nil {
-			l.Errorf("Error running hook %s with message: %s", curHook.Name, possibleError.Error())
-			return possibleError
+		if containsExternalTFLintFlag {
+			// delete the flag and continue to execute as a normal hook
+			actionParams = slices.DeleteFunc(actionParams, tflint.IsExternalTFLintFlag)
+		} else {
+			forceExternalTFLint := (opts.StrictControls.Find(controls.InternalTFLint).Evaluate(ctx) != nil)
```
yhakbar: You can fail to find here, which would panic. No reason we should expect that, but to be safe, we can just consider the strict control inactive if we fail to find.

yhakbar: Also I don't understand what the parentheses do here.

Ah, I guess it's that you don't want to do the comparison inline on 189.

internal/tflint/tflint.go
```diff
@@ -28,6 +28,12 @@ const (
 	tfExternalTFLint = "--terragrunt-external-tflint"
 )
 
+// IsExternalTFLintFlag satisfies XyzFunc signatures (i.e. for
+// slices.DeleteFunc) to match the external tflint flag.
+func IsExternalTFLintFlag(flag string) bool {
```
yhakbar: I don't think it's a big deal either way, but do we need it? Seems like it could be more straightforward to have a lambda function in DeleteFunc.

internal/runner/run/hook.go
thisguycodes: I got lul'd into security by Evaluate()'s handling of nil:
https://github.com/gruntwork-io/terragrunt/blob/543568cbbca2da49918a24825270028a3758eda5/internal/strict/controls/control.go#L119-L122

I think I'll actually update find to return a nil `*Control`. I think this makes things simpler to interact with.

thisguycodes: nope, can't do that without a dependency cycle, I'll add a nil check

internal/tflint/tflint.go
thisguycodes: yea I debated, currently the const isn't public, but I can rename it to be public instead

internal/runner/run/hook.go
thisguycodes: negation is harder to read, and do the if block here gives me a space to put a comment making the continuation of execution more explicit.

Otherwise future readers need to intuit the "absent" conditions that lead to a fallthrough in execution.

I'm not super happy with the control flow, so I thought a lot about how to make it as clear as possible. Unfortunately since I don't want to show the warning when the external flag is used, it has to be kinda complicated.

yhakbar: I'm not sure I agree that negation is harder to read than else's, but it doesn't need to hold this up. You could also put the comment above the early return.

internal/strict/controls/controls.go
```diff
@@ -260,6 +263,13 @@ func New() strict.Controls {
 			Error:       errors.New("The `--no-destroy-dependencies-check` flag is no longer supported. Use `--destroy-dependencies-check` to enable dependency checks during destroy operations."),
 			Warning:     "The `--no-destroy-dependencies-check` flag is deprecated and will be removed in a future version of Terragrunt. This flag is now ignored. Use `--destroy-dependencies-check` to enable dependency checks during destroy operations.",
 		},
+		&Control{
+			Name:        InternalTFLint,
+			Description: "Prevents the use of the deprecated embedded version of tflint, instead treating `tflint` as a normal hook.",
+			Category:    stageCategory,
+			Error:       errors.New("The embedded version of tflint is no longer supported."),
```
yhakbar: I think you still want this here:

```
Use the `--terragrunt-external-tflint` flag
```

So that users know what to do when they get the error with strict mode on.

thisguycodes: I put it in the warning, which shows up in the logs:
```
16:35:19.907 WARN   The embedded version of tflint is deprecated and will be removed in a future version of Terragrunt. Use the `--terragrunt-external-tflint` flag in your hook to opt in to running tflint externally.
```

The error here isn't used, but I'll add it anyway.

thisguycodes:
I've discovered that our internal running of `tflint` was actually automatically running `tflint --init`, even for external tflint binaries; this means my "totally bypass the tflit package" strategy doesn't work.

I'm refactoring to still do the init automatically.

docs-starlight/src/content/docs/03-features/10-hooks.md
```diff
@@ -268,11 +273,21 @@ config {
 
 By default, `tflint` is executed with the internal `tflint` built into Terragrunt, which will evaluate parameters passed in.
 
+<Aside type="warning">
+    The internal `tflint` has become horribly out of date due to upstream adoption of the BUSL license (preventing us
```
yhakbar: I think this can be toned down a bit.

"Previous versions of Terragrunt shipped with a version of `tflint` compiled into the binary to allow for convenient usage of `tflint` without installing it directly. Due to the adoption of a BUSL license of `tflint` in newer versions of `tflint`, the version included in Terragrunt was frozen.

docs-starlight/src/content/docs/03-features/10-hooks.mdx
yhakbar, line 1: This needs to be an mdx file for the import to work.

internal/runner/run/hook.go
```diff
@@ -175,30 +176,43 @@ func runHook(
...
-		if possibleError != nil {
-			l.Errorf("Error running hook %s with message: %s", curHook.Name, possibleError.Error())
-			return possibleError
+		if containsExternalTFLintFlag {
+			// delete the flag and run external tflint
+			curHook.Execute = slices.DeleteFunc(curHook.Execute, func(arg string) bool {
+				return arg == tflint.TfExternalTFLint
+			})
+
+			return executeTFLint(ctx, l, opts, cfg, curHook, workingDir, true)
```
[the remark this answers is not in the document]
yhakbar: This is valid.

docs-starlight/src/content/docs/03-features/10-hooks.mdx
thisguycodes: good catch!

docs-starlight/src/content/docs/03-features/10-hooks.md
thisguycodes: agreed. This "voice" stuff is a bit hard for me 😓

internal/tflint/tflint.go
```diff
@@ -85,13 +83,14 @@ func RunTflintWithOpts(ctx context.Context, l log.Logger, opts *options.Terragru
...
-	args := make([]string, 0, 6+len(variables)+len(tflintArgs))
+	args := make([]string, 0, 5+len(variables)+len(tfVariables)+len(tflintArgs))
 	args = append(args,
 		"tflint",
 		"--config", configFile,
 		"--chdir", opts.WorkingDir,
 	)
 	args = append(args, variables...)
+	args = append(args, tfVariables...)
 	args = append(args, tflintArgs...)
```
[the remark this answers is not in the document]
thisguycodes: this is valid

Issue #3927, opened 2025-02-24
Add curlable install to starlight docs


odgrim:
## Describe the enhancement

Our getting started docs could benefit greatly to have an even faster way to get started. Something like the hombrew "curl this to install":
<img width="823" alt="Image" src="https://github.com/user-attachments/assets/f0020a78-78ce-4a91-99f4-7b5333060e1d" />
and "For additional install methods, see our install guide."

PR #5364, opened 2026-01-15
feat: install script for Terragrunt

denis256:
## Description

* install script for Terragrunt
* added basic test to verify installation process
* added CICD to run installation tests on Linux and Macos
* by default are checked ony files checksums - added optional checking for gpg/cosign signatures
* removed unused `dv-Terminal.astro` component

Close: #3927

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **New Features**
  * Added a cross-platform installer script for Linux/macOS with multi-arch support and optional SHA256/GPG/Cosign verification.

* **Documentation**
  * Added a Quick Install section and script options guidance.

* **Tests**
  * Added a comprehensive install test suite covering unit and integration scenarios, plus a quick-run option.

* **CI**
  * Added CI workflow/job to run the install tests across Linux and macOS.

* **Refactor**
  * Removed the embedded terminal/install widget from the documentation UI.


docs-starlight/public/install
yhakbar, line 1: If you don't mind, let's make this file just `install` and at the root of public.

That makes the URL as short as possible so it'll fit nice on pages.

docs-starlight/public/scripts/install.sh
```diff
@@ -0,0 +1,396 @@
+#!/bin/bash
```
yhakbar: Prefer `#!/usr/bin/env bash`

docs-starlight/src/content/docs/01-getting-started/03-install.mdx
```diff
@@ -56,6 +56,67 @@ cosign verify-blob SHA256SUMS \
 
 <InstallTabs version={version} />
 
+## Quick Install Script
```
yhakbar: I would put this at the top.

docs-starlight/src/content/docs/01-getting-started/03-install.mdx
```diff
@@ -12,6 +12,67 @@ export const response = await fetch('https://api.github.com/repos/gruntwork-io/t
...
+  frame="terminal"
+/>
+
+Install to a custom directory:
+
+```bash
+curl -sL https://terragrunt.gruntwork.io/install.sh | bash -s -- -d ~/bin
+```
+
+Install with GPG signature verification:
```
yhakbar: Signature verification should be opt out, not opt in. Also, please prioritize GPG over Cosign if both are available (and if you think it's better to prefer cosign, let's update the asdf plugin too).

docs-starlight/src/content/docs/01-getting-started/03-install.mdx
```diff
@@ -12,6 +12,67 @@ export const response = await fetch('https://api.github.com/repos/gruntwork-io/t
...
 
+## Quick Install
+
+The quickest way to install Terragrunt on Linux or macOS:
+
+```bash
+curl -sL https://terragrunt.gruntwork.io/install.sh | bash
+```
+
+This will:
```
yhakbar: Please remove this, the script options and examples. We should just have an Aside that gives users the tip they can use the following to customize the install script:

```bash
curl -sL https://terragrunt.gruntwork.io/install.sh | bash -s -- --help
```

and a link to the script so they can read it. We want this section to be as short as possible, as we have a ton of details below.

docs-starlight/public/install
```diff
@@ -0,0 +1,605 @@
...
+            fi
+            ;;
+        auto)
+            # Try GPG first, then Cosign
+            if command -v gpg &>/dev/null; then
+                info "Using GPG for signature verification"
+                if verify_gpg "$version" "$checksums_path" "$tmpdir"; then
+                    return 0
+                fi
+                warn "GPG verification failed, trying Cosign..."
```
yhakbar: Let's not fallback to cosign when gpg verification fails. If users have gpg installed and gpg verification fails, we should throw.

docs-starlight/public/install
```diff
@@ -0,0 +1,605 @@
...
+                abort "GPG signature verification failed!"
+            fi
+            ;;
+        cosign)
+            if ! command -v cosign &>/dev/null; then
+                abort "Cosign verification requested but cosign is not installed"
+            fi
+            if verify_cosign "$version" "$checksums_path" "$tmpdir"; then
+                return 0
+            else
```
yhakbar: We don't need this else

docs-starlight/public/install
```diff
@@ -0,0 +1,605 @@
...
+    local method="$4"  # gpg, cosign, or auto
+
+    case "$method" in
+        gpg)
+            if ! command -v gpg &>/dev/null; then
+                abort "GPG verification requested but gpg is not installed"
+            fi
+            if verify_gpg "$version" "$checksums_path" "$tmpdir"; then
+                return 0
+            else
```
yhakbar: We don't need this else

docs-starlight/public/install
```diff
@@ -0,0 +1,605 @@
...
+            ;;
+        cosign)
+            if ! command -v cosign &>/dev/null; then
+                abort "Cosign verification requested but cosign is not installed.
+Install cosign or use --verify-sig for auto-detection."
+            fi
+            ;;
+        auto)
+            # Auto mode: user explicitly requested verification, abort if no tools
+            if ! command -v gpg &>/dev/null && ! command -v cosign &>/dev/null; then
```
yhakbar: We should require that users opt out of signature verification.

docs-starlight/public/install
```diff
@@ -0,0 +1,605 @@
...
+check_dependencies() {
+    if ! command -v curl &>/dev/null; then
+        abort "curl is required but not installed.
+Please install curl and try again."
+    fi
+
+    if [[ "$VERIFY_SHA" == true ]]; then
+        if ! command -v sha256sum &>/dev/null && ! command -v shasum &>/dev/null; then
+            abort "Neither sha256sum nor shasum found.
+Install one or use --no-verify to skip checksum verification."
```
yhakbar: When we suggest flags like this in the errors, we should use the full form with curl because users might not know how to pass the flags in.

e.g.

```bash
curl -sL https://terragrunt.gruntwork.io/install.sh | bash -s -- --no-verify
```

docs-starlight/public/install
```diff
@@ -0,0 +1,605 @@
...
+    fi
+
+    binary_name="terragrunt_${os}_${arch}"
+
+    info "Installing Terragrunt ${version} for ${os}/${arch}"
+
+    # Create temp directory
+    local tmpdir
+    tmpdir=$(mktemp -d 2>/dev/null || mktemp -d -t 'terragrunt-install')
+    trap 'rm -rf "${tmpdir:-}"' EXIT
```
yhakbar: Why are we ignoring the value being empty here? That seems unsafe. `trap` should only interpolate global variables.

docs-starlight/public/install
```diff
@@ -0,0 +1,605 @@
...
+# Environment:
+#   TERRAGRUNT_VERSION       Override version (same as -v)
+#   TERRAGRUNT_INSTALL_DIR   Override install directory (same as -d)
+
+set -euo pipefail
+
+# --- Constants ---
+readonly GITHUB_REPO="gruntwork-io/terragrunt"
+readonly GPG_KEY_URL="https://gruntwork.io/.well-known/pgp-key.txt"
+readonly DEFAULT_INSTALL_DIR="/usr/local/bin"
```
yhakbar: Feel free to defer this, but I think we should install to a user directory by default, and let the user update PATH, or update their rc file for them.

e.g.

```bash
$ curl -fsSL https://bun.com/install | bash
######################################################################## 100.0%
bun was installed successfully to ~/.bun/bin/bun

Added "~/.bun/bin" to $PATH in "~/.bashrc"

To get started, run:

  source /root/.bashrc
  bun --help
```

docs-starlight/public/install
```diff
@@ -0,0 +1,608 @@
...
+
+    # Create install directory if needed
+    [[ ! -d "$install_dir" ]] && {
+        info "Creating installation directory: $install_dir"
+        mkdir -p "$install_dir" 2>/dev/null || abort "Failed to create installation directory: $install_dir"
+    }
+
+    # Check write permissions
+    [[ ! -w "$install_dir" ]] && abort "Cannot write to $install_dir
+Try running with sudo or specify a different directory with -d"
```
yhakbar: It might not be clear to users how to do this. We might need to give them an example.

e.g.

```bash
curl ... | sudo bash
```

docs-starlight/public/install
```diff
@@ -0,0 +1,615 @@
...
+    if version=$(curl -fsL "https://api.github.com/repos/${GITHUB_REPO}/releases/latest" 2>/dev/null | grep -o '"tag_name": "[^"]*' | cut -d'"' -f4); then
+        if [[ -n "$version" ]]; then
+            echo "$version"
+            return 0
+        fi
+    fi
+
+    abort "Could not determine latest version.
+This may be due to GitHub API rate limits (60 requests/hour).
+Specify a version manually with -v, e.g.: -v v0.72.5"
```
yhakbar, line 217: Let's defer this, but this still doesn't use the curl* form to tell the user how to pass the flag.

docs-starlight/public/install
```diff
@@ -0,0 +1,615 @@
...
+}
+
+validate_version() {
+    local version="$1"
+    # Accept versions with or without 'v' prefix
+    if [[ ! "$version" =~ ^v?[0-9]+\.[0-9]+\.[0-9]+ ]]; then
```
yhakbar: This assumes we're not gonna support pre-releases. We might want to.

PR #5381, opened 2026-01-20
feat: Using updated `v0.1.0` version of `terragrunt-engine-go`

yhakbar:
## Description

Pulls in version `v0.1.0` of `github.com/gruntwork-io/terragrunt-engine-go`, which adjusts the schema for messages are sent between Terragrunt and engines.

This is a breaking change to the engine experiment, and will require that engine maintainers update their engines to use the new version of `terragrunt-engine-go`.

As part of this change, the `engine` package in Terragrunt now has a fix to the byte streaming from the engine to Terragrunt to ensure that runes are printed instead of priting byte values. This ensures that unicode values are printed properly as they're streamed from engines.

Engines can now also log messages in Terragrunt.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

Updated `engine` integration with `terragrunt-engine-go` to use `v0.1.0` of the package.

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable plugin shutdown with a graceful timeout before forced termination.

* **Improvements**
  * Improved logging and message routing for clearer diagnostics.
  * More accurate handling of command output (Unicode-safe) and better capture of stdout/stderr and exit codes.
  * The "engine" command no longer requires an init step.
  * Tests updated to use the structured run invocation and accommodate log-level output.

* **Chores**
  * Updated terragrunt-engine-go dependency to v0.1.0.

### DIFF
--- a/test/integration_engine_test.go
+++ b/test/integration_engine_test.go
@@ -47,7 +47,11 @@ var (
 func TestEngineLocalPlan(t *testing.T) {
 	rootPath := setupLocalEngine(t)
 
-	stdout, stderr, err := helpers.RunTerragruntCommandWithOutput(t, fmt.Sprintf("terragrunt plan --non-interactive --tf-forward-stdout --working-dir %s --log-level trace", rootPath))
+	stdout, stderr, err := helpers.RunTerragruntCommandWithOutput(
+		t,
+		"terragrunt run --non-interactive --log-level trace --tf-forward-stdout --working-dir "+
+			rootPath+" -- plan",
+	)
 	require.NoError(t, err)
 
 	assert.Contains(t, stderr, engineAssetName)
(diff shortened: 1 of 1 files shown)


denis256:
Looks like engine tests are failing

yhakbar:
> Looks like engine tests are failing

They're going to fail until engines get updated, and engines will fail their CI until Terragrunt gets updated:
https://github.com/gruntwork-io/terragrunt-engine-opentofu/pull/61

Once we're GA we can support backwards compatibility by supporting multiple different schemas for engines at once. I think we would support that by having something like [this](https://github.com/gruntwork-io/terragrunt-engine-go/blob/main/examples/client-server/client/main.go#L241), with multiple integer versions for different protocol versions, and an updated client config that uses `VersionedPlugins` instead of `ProtocolVersion` [here](https://github.com/gruntwork-io/terragrunt/blob/main/internal/engine/engine.go#L535).

PR #5383, opened 2026-01-20
docs: compatibility API

denis256:
## Description

* exposed Terragrunt compatibility table as JSON API
* added basic description of API
* added support for filtering of data based on the tool name

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **New Features**
  * Interactive compatibility tables for OpenTofu and Terraform showing corresponding Terragrunt version ranges.
  * Dynamic compatibility JSON API that returns content-driven compatibility data with an optional tool filter.

* **Documentation**
  * Replaced the old static supported-versions doc with a new page embedding the compatibility tables, notes on BSL/CI scope, and API usage examples.

* **Data**
  * Added a structured compatibility dataset that powers the tables and API.


docs-starlight/src/pages/api/compatibility.json.ts
```diff
@@ -1,111 +1,34 @@
...
+	const tool = url.searchParams.get('tool');
 
-	// Only return the terraform compatibility table
-	// if the user asks for it with the "tf" query parameter
-	if (searchParams.tf && searchParams.tf === "terraform") {
-		return new Response(
-			JSON.stringify(terraformCompatibility)
-		)
+	// Filter by tool if query param provided
+	let filtered = entries;
```
yhakbar: Let's use a function call with an early return to avoid this kind of mutability. As it stands, if the user passes in an empty string for `tool` or something other than `opentofu` or `terraform`, we'll stringify everything, not nothing.

docs-starlight/src/data/compatibility/compatibility.json
```diff
@@ -0,0 +1,210 @@
+[
+  {
+    "id": "opentofu-1.11.x",
+    "tool": "opentofu",
+    "version": "1.11.x",
+    "terragrunt_min": "0.95.0",
+    "terragrunt_max": null,
+    "order": 1
```
yhakbar: Can we invert this ordering logic so that it's easier to add new entries? Right now, we'll have to increment all the existing entries whenever we add a new version of OpenTofu we test against.

denis256: good point

docs-starlight/src/pages/api/v1/compatibility.ts
yhakbar, line 1: @denis256 could we also change this endpoint from `/api/compatibility.json` to `/api/compatibility`?

It's theoretically possible that we'll want to support alternate response payloads based on the Accept header. It's probably also a good idea to put this behind a `v1` so that we can support backwards compatibility in the future.

PR #5399, opened 2026-01-23
fix: Improving test speed

yhakbar:
## Description

Trying to make tests a bit faster.

We were dumping a huge amount of content from the trace logging in tests, and we're using log assertions in tests instead of using reports in some places. This removes any trace logging done in tests to reduce the amount we're dumping to stderr. Using reports instead of logs is also more reliable, so we'll hopefully have to do fewer retries for flaky tests.

I left more notes inline when I changed things unrelated to the logging adjustments when trying to make tests faster or more reliable.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a race condition in queue status handling that could cause concurrent operations to incorrectly overwrite early-exit status with success status.

* **Tests**
  * Enhanced test infrastructure with report-based validation for improved reliability.
  * Updated test fixtures to support broader Terraform provider version ranges for better compatibility.

### DIFF
--- a/test/integration_destroy_test.go
+++ b/test/integration_destroy_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 
 	"github.com/gruntwork-io/terragrunt/internal/git"
+	"github.com/gruntwork-io/terragrunt/internal/report"
 	"github.com/gruntwork-io/terragrunt/internal/runner/run"
 	"github.com/gruntwork-io/terragrunt/test/helpers"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


internal/cas/integration_test.go
```diff
@@ -113,21 +113,23 @@ func TestIntegration_TreeStorage(t *testing.T) {
...
 		}, "https://github.com/gruntwork-io/terragrunt.git"))
 
-		// Get the commit hash
+		// Get the commit hash for the tag
 		g, err := git.NewGitRunner()
 		require.NoError(t, err)
 
-		g = g.WithWorkDir(filepath.Join(tempDir, "repo"))
-		results, err := g.LsRemote(ctx, "https://github.com/gruntwork-io/terragrunt.git", "HEAD")
+		results, err := g.LsRemote(ctx, "https://github.com/gruntwork-io/terragrunt.git", testTag)
```
yhakbar, line 132: This was causing flakes because remote HEAD might have moved between time of check and time of use.

internal/queue/queue.go
```diff
@@ -414,10 +414,18 @@ func (q *Queue) areDependentsReadyUnsafe(e *Entry) bool {
...
 // SetEntryStatus safely sets the status of an entry with proper synchronization.
+//
+// If the entry is already in a terminal state (StatusSucceeded, StatusFailed, or StatusEarlyExit),
+// this operation is a no-op. This prevents race conditions where a concurrent success could
+// overwrite an early-exit status set by fail-fast mode.
 func (q *Queue) SetEntryStatus(e *Entry, status Status) {
 	q.mu.Lock()
 	defer q.mu.Unlock()
 
+	if isTerminal(e.Status) {
```
yhakbar, line 425: This was causing flakes because a unit with `StatusEarlyExit` could be overridden to start its run without this check.

test/fixtures/auto-provider-cache-dir/basic/unit/main.tf
yhakbar, line 1: These were causing annoying local test failures due to the provider of that version not being found for whatever reason. It either had to do with that version of `null` no longer being available in the provider registry, or a platform issue. In either case, a lock file is a more reliable way to test this, so I'm removing this restriction in `required_providers` and adding lockfiles in #5404

test/integration_download_test.go
```diff
@@ -386,21 +386,6 @@ func TestExcludeDirsWithFilter(t *testing.T) {
 			excludeArgs:           "--filter '!./integration-env/gce/module-gce-b' --filter '!./integration-env/gce/module-gce-c' --filter '!./**/module-aws*'",
 			excludedModuleOutputs: []string{"Module AWS A", "Module GCE B", "Module GCE C", "Module AWS D"},
 		},
-		{
```
yhakbar, line 389: These tests are duplicates.

test/integration_regressions_test.go
```diff
@@ -277,8 +277,11 @@ func TestExposedIncludeWithDeprecatedInputsSyntax(t *testing.T) {
 // configs with exposed includes containing generate blocks would fail during
 // discovery with "Could not find Terragrunt configuration settings" errors.
 //
+// Uses `list --dag --format=dot` instead of `run --all plan` since this is a parsing test
+// and doesn't need to run terraform operations.
+//
 // See: https://github.com/gruntwork-io/terragrunt/issues/4983
-func TestRunAllWithGenerateAndExpose(t *testing.T) {
+func TestParsingWithGenerateAndExpose(t *testing.T) {
```
yhakbar, line 284: This test gets way faster if we avoid using `run --all` and instead just parse with `list --dag`. Still tests the same underlying behavior.

internal/queue/queue_test.go
```diff
@@ -1010,3 +1011,40 @@ func TestQueue_DestroyWithIgnoreDependencyErrors_AllowsProgressAfterFailure(t *t
...
+		{queue.StatusFailed, queue.StatusSucceeded},
+		{queue.StatusFailed, queue.StatusEarlyExit},
+		{queue.StatusEarlyExit, queue.StatusSucceeded},
+		{queue.StatusEarlyExit, queue.StatusFailed},
+	}
+
+	for _, tc := range testCases {
+		t.Run(fmt.Sprintf("%v_to_%v", tc.initial, tc.attempted), func(t *testing.T) {
+			t.Parallel()
+
```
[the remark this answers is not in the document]
yhakbar: This is not how modern Golang behaves. Please don't suggest copying variables to ensure loop variable capture in this codebase again.

internal/queue/queue_test.go
```diff
@@ -1010,3 +1011,40 @@ func TestQueue_DestroyWithIgnoreDependencyErrors_AllowsProgressAfterFailure(t *t
...
+			configs := component.Components{component.NewUnit("Test")}
+			q, err := queue.NewQueue(configs)
+			require.NoError(t, err)
+
+			entry := q.EntryByPath("Test")
+
+			// Set initial terminal state directly
+			entry.Status = tc.initial
```
denis256, line 1042: Should be used `q.SetEntryStatus(...) ` to set initial status? or it will it break state machine logic?

internal/queue/queue.go
denis256: Should this function return a value indicating whether the status update was accepted or rejected?

internal/queue/queue_test.go
yhakbar: I think it's fine either way. I think this is a more direct test because if there is a bug in `SetEntryStatus` preventing the correct initial set, we won't know if the logic for precedence will work correctly.

yhakbar:
@denis256 

Sorry, I can't reply to the comment above for some reason.

> Should this function return a value indicating whether the status update was accepted or rejected?

What would we use that for? Some debug logging or something?

PR #5402, opened 2026-01-23
feat: .terragrunt-cache directory creation

### LINKED ISSUE #5397: Always create the `.terragrunt-cache` directory
## Describe the enhancement

Always create the .terragrunt-cache directory, even when no terraform.source is configured.

## Additional context

Currently, .terragrunt-cache is only created when a terraform.source block is present. 
To make the behavior consistent (and ensure a predictable working directory for downstream logic), Terragrunt should always create .terragrunt-cache and copy the working files into it, regardless of whether terraform.source is defined.

- [ ] Changes required.
- [ ] Level of effort.

## PoC (Proof of Concept)

- [ ] [PoC]()

## RFC Not Needed

- [ ] I have evaluated the complexity of this enhancement, and I believe it does not require an RFC.

denis256:
## Description

* Always copy/download source to .terragrunt-cache directory for consistent execution
* Fixed tests to reference .terragrunt-cache files
* Updated report path to use the original unit path in reports - save absolute paths and strip `.terragrunt-cache` part in report

Fixes #5397.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **New Features**
  * Unspecified module source is treated as the current directory, ensuring a .terragrunt-cache is consistently created and used.

* **Bug Fixes**
  * Hooks and generated files now run/land in the Terragrunt module directory.
  * Reporting, lock-file placement, and detailed-exit-code reporting derive paths from the original config location for more accurate outputs.
  * Init-from-module now executes for modules with no explicit source (treated as ".").

* **Tests**
  * New and updated fixtures and helpers to validate cache-based behavior and path semantics.

### DIFF
--- a/test/integration_runner_pool_test.go
+++ b/test/integration_runner_pool_test.go
@@ -146,12 +146,10 @@ func TestRunnerPoolFailFast(t *testing.T) {
 				reason string
 				cause  string
 			}{
-				"failing-unit":       {result: "failed", reason: "run error"},
-				"succeeding-unit":    {result: "succeeded"},
-				"depends-on-failing": {result: "early exit", reason: "ancestor error", cause: "failing-unit"},
-				// depends-on-succeeding can be either "early exit" or "succeeded" depending on timing.
-				// With parallel execution, it may complete before fail-fast propagates, or get stopped.
-				// Both outcomes are valid - we handle this case separately in the test below.
+				"failing-unit":          {result: "failed", reason: "run error"},
+				"succeeding-unit":       {result: "succeeded"},
+				"depends-on-failing":    {result: "early exit", reason: "ancestor error", cause: "failing-unit"},
+				"depends-on-succeeding": {result: "early exit", reason: "ancestor error"},
 			},
 		},
 	}
@@ -197,23 +195,6 @@ func TestRunnerPoolFailFast(t *testing.T) {
 				if expected.cause != "" {
 					require.NotNil(t, run.Cause, "expected cause for %q but got nil", unitName)
 					assert.Equal(t, expected.cause, *run.Cause, "unexpected cause for %q", unitName)
-				}
-			}
-
-			// Special handling for depends-on-succeeding in fail-fast=true case.
-			// Due to parallel execution, it can either complete before fail-fast propagates
-			// (result: "succeeded") or be stopped (result: "early exit"). Both are valid.
-			if tc.failFast {
-				run := runs.FindByName("depends-on-succeeding")
-				require.NotNil(t, run, "run %q not found in report", "depends-on-succeeding")
-
-				validResults := []string{"succeeded", "early exit"}
-				assert.Contains(t, validResults, run.Result,
-					"depends-on-succeeding should be either 'succeeded' or 'early exit', got %q", run.Result)
-
-				if run.Result == "early exit" {
-					require.NotNil(t, run.Reason, "expected reason for depends-on-succeeding but got nil")
-					assert.Equal(t, "ancestor error", *run.Reason)
 				}
 			}
 		})
--- a/test/fixtures/fail-fast-early-exit/depends-on-succeeding/terragrunt.hcl
+++ b/test/fixtures/fail-fast-early-exit/depends-on-succeeding/terragrunt.hcl
@@ -4,3 +4,13 @@ dependency "succeeding" {
     data = "mock"
   }
 }
+
+terraform {
+  before_hook "wait_for_signal" {
+    commands = ["apply"]
+    execute  = [
+      "bash", "-c",
+      "for i in $(seq 1 30); do [ -f '${get_terragrunt_dir()}/../.fail-signal' ] && exit 0; sleep 0.1; done; exit 0"
+    ]
+  }
+}


internal/report/summary.go
```diff
@@ -228,6 +228,8 @@ func (s *Summary) writeUnitLevelSummary(w io.Writer, colorizer *Colorizer) error
 			name = strings.TrimPrefix(name, s.workingDir+string(os.PathSeparator))
 		}
 
+		name = stripCachePath(name)
```
yhakbar: Why do we have to do this? The value stored in the report should be the path to the unit, not the OpenTofu/Terraform module it downloads/copies.

internal/runner/run/run.go
```diff
@@ -146,18 +146,22 @@ func Run(
...
+	}, func(ctx context.Context) error {
+		updatedTerragruntOptions, err = DownloadTerraformSource(ctx, l, sourceURL, opts, cfg, r)
+		return err
+	})
+	if err != nil {
+		return err
 	}
 
+	// Update opts.WorkingDir to match the cache directory so that
+	// exit codes and other logic that uses opts.WorkingDir are consistent.
```
yhakbar: Why do exit codes depend on the cache working directory? That's odd.

internal/runner/run/run.go
```diff
@@ -274,7 +278,8 @@ func runTerragruntWithConfig(
 			// case, we are using the user's working dir here, rather than just looking at the parent dir of the
 			// terragrunt.hcl. However, the default value for the user's working dir, set in options.go, IS just the
 			// parent dir of terragrunt.hcl, so these will likely always be the same.
-			lockFileError = runcfg.CopyLockFile(l, opts, opts.WorkingDir, originalOpts.WorkingDir)
+			// Use directory of OriginalTerragruntConfigPath since WorkingDir may have been changed to the cache directory.
```
yhakbar: Why is this a "may"?

test/fixtures/auth-provider-cmd/multiple-apps/root.hcl
```diff
@@ -1,6 +1,7 @@
 terraform {
   before_hook "before_hook" {
     commands = ["init"]
-    execute  = ["../test-creds.sh"]
+    execute  = ["${get_parent_terragrunt_dir()}/test-creds.sh"]
```
yhakbar: Would it be possible to avoid requiring this breaking change? As it is, do hooks run in the cache dir when using `terraform.source` and not when it's not defined?

denis256: Hooks run in .terragrunt-cache directory

```
09:26:34.684 INFO   Executing hook: sleep_test
/projects/gruntwork/terragrunt-tests/hooks/.terragrunt-cache/O0YS4i4EWlLIAOWmJ9tVMNj8O8M/66jrnDC7w2RZNHIx60Wfi-VS7CY
```

internal/runner/run/run.go
denis256: I was referencing findFailedDependency(), which check unit path


https://github.com/gruntwork-io/terragrunt/blob/df76a45bbba1b4ecf3b8a94b392869817b5a1108/internal/runner/runnerpool/errors.go#L48

internal/runner/common/unit_runner.go
```diff
@@ -97,10 +97,11 @@ func (runner *UnitRunner) runTerragrunt(
 	runErr := run.Run(ctx, runner.Unit.Execution.Logger, opts, r, cfg, credsGetter)
 
 	// Store the unit exit code in the global map using the unit path as key
-	// Get the exit code from the unit-scoped map (stored with empty string key in run_cmd.go)
+	// Get the exit code from the unit-scoped map using GetFinalDetailedExitCode, as the key in the map
+	// is the cache directory (which we don't have easy access to here), not opts.WorkingDir.
 	if globalExitCode != nil {
 		unitPath := filepath.Clean(runner.Unit.AbsolutePath())
-		code := unitExitCode.Get(opts.WorkingDir)
+		code := unitExitCode.GetFinalDetailedExitCode()
```
yhakbar: Why are we getting the aggregate exit code here? I think I might have caused some confusion with bad variable names when I wrote the surrounding code.

test/helpers/test_helpers.go
```diff
@@ -291,6 +291,54 @@ func determineToolName(command string) string {
 	return command
 }
 
+// FindCacheWorkingDir finds the working directory inside the .terragrunt-cache folder.
+// The cache layout is <root>/.terragrunt-cache/<hash>/<module>/..., so we return the first
+// two-level directory path we find.
+func FindCacheWorkingDir(t *testing.T, rootDir string) string {
```
yhakbar, line 297: Would it be better to use requires here instead of returning `""`?

If someone is using this function, they probably want the test to fail if they can't find `.terragrunt-cache`, or if there are extra directories other than the two paths used within `.terragrunt-cache` before getting to the directory.

Seems more reliable than returning the first directory found two levels deep.

internal/runner/common/unit_runner.go
denis256: `unitExitCode` have scope of this function and it has only one entry, since we don't want to calculate the key with `.terragrunt-cache` we simply fetch exit code with `GetFinalDetailedExitCode()`


I can change the logic how exit code is set, so we can fetch it with `unitPath` but it will add unrelated changes

test/helpers/test_helpers.go
```diff
@@ -291,6 +291,54 @@ func determineToolName(command string) string {
...
+	}
+
+	return ""
+}
+
+// FileExistsInCache checks if a file exists within the cache directory structure.
+func FileExistsInCache(t *testing.T, rootDir, filename string) bool {
+	t.Helper()
+
+	cacheWorkingDir := FindCacheWorkingDir(t, rootDir)
```
yhakbar, line 355: Again, I would think someone using this in a test would prefer that this can't fail.

test/helpers/test_helpers.go
```diff
@@ -291,6 +291,54 @@ func determineToolName(command string) string {
...
+	if cacheWorkingDir == "" {
+		return false
+	}
+
+	filePath := filepath.Join(cacheWorkingDir, filename)
+	_, err := os.Stat(filePath)
```
yhakbar, line 357: You can use `assert.FileExists` for an assertion with a better error message.

test/integration_aws_test.go
```diff
@@ -373,41 +373,16 @@ func TestAwsBootstrapBackendWithAccessLogging(t *testing.T) {
...
-	dualLockingPath := filepath.Join(rootPath, "dual-locking", "terragrunt.hcl")
-	helpers.CopyTerragruntConfigAndFillPlaceholders(
-		t,
-		dualLockingPath,
-		dualLockingPath,
-		s3BucketName,
-		dynamoDBName,
-		helpers.TerraformRemoteStateS3Region,
-	)
+	helpers.CopyTerragruntConfigAndFillPlaceholders(t, commonConfigPath, commonConfigPath, s3BucketName, dynamoDBName, helpers.TerraformRemoteStateS3Region)
```
yhakbar: Why did we go back to long lines :(

(Very unimportant)

test/integration_aws_test.go
```diff
@@ -667,14 +643,18 @@ func TestAwsBeforeAfterAndErrorMergeHook(t *testing.T) {
...
-	_, afterException := os.ReadFile(childPath + "/after.out")
-	_, afterParentException := os.ReadFile(childPath + "/after-parent.out")
-	_, errorHookParentException := os.ReadFile(childPath + "/error-hook-parent.out")
-	_, errorHookChildException := os.ReadFile(childPath + "/error-hook-child.out")
-	_, errorHookOverridenParentException := os.ReadFile(childPath + "/error-hook-merge-parent.out")
+	// Hook output files are now in .terragrunt-cache directory
+	cacheDir := helpers.FindCacheWorkingDir(t, childPath)
+	require.NotEmpty(t, cacheDir, "Cache directory should exist")
+
+	_, beforeException := os.ReadFile(cacheDir + "/before.out")
```
yhakbar: We should use filepath.Join here to avoid issues on Windows.

Also, we can use `assert.FileExists` instead for better error messages.

test/integration_runner_pool_test.go
```diff
@@ -199,6 +199,23 @@ func TestRunnerPoolFailFast(t *testing.T) {
 					assert.Equal(t, expected.cause, *run.Cause, "unexpected cause for %q", unitName)
 				}
 			}
+
+			// Special handling for depends-on-succeeding in fail-fast=true case.
+			// Due to parallel execution, it can either complete before fail-fast propagates
+			// (result: "succeeded") or be stopped (result: "early exit"). Both are valid.
+			if tc.failFast {
```
yhakbar: This is a bug in the test. We need to have depends on succeeding artificially block until failing unit fails so that we are actually testing that fail fast works. That's why the fixture has logic to wait for the sentinel file generated from the error hook of failing unit.

BeyondEvil:
According to [documentation](https://terragrunt.gruntwork.io/docs/features/provider-cache-server/#usage) the path on unix machines should be `$HOME/.terragrunt-cache/terragrunt/providers`

But on Ubuntu 24.04, I'm getting:

```sh
$ ls -laR "${HOME}/.terragrunt-cache/"

ls: cannot access '/home/runner/.terragrunt-cache/': No such file or directory
```

This is where I actually find the cache:

```sh
$ ls -laR "${HOME}/.cache/terragrunt/"

/home/runner/.cache/terragrunt/:
total 0
drwxr-xr-x. 3 runner runner 23 Jan 29 10:22 .
drwxr-xr-x. 4 runner runner 36 Jan 29 10:22 ..
drwxr-xr-x. 3 runner runner 35 Jan 29 10:22 providers
/home/runner/.cache/terragrunt/providers:
total 0
drwxr-xr-x. 3 runner runner 35 Jan 29 10:22 .
drwxr-xr-x. 3 runner runner 23 Jan 29 10:22 ..
drwxr-xr-x. 3 runner runner 23 Jan 29 10:22 registry.terraform.io
/home/runner/.cache/terragrunt/providers/registry.terraform.io:
total 0
drwxr-xr-x. 3 runner runner 23 Jan 29 10:22 .
drwxr-xr-x. 3 runner runner 35 Jan 29 10:22 ..
drwxr-xr-x. 3 runner runner 17 Jan 29 10:22 hashicorp
/home/runner/.cache/terragrunt/providers/registry.terraform.io/hashicorp:
total 0
drwxr-xr-x. 3 runner runner 17 Jan 29 10:22 .
drwxr-xr-x. 3 runner runner 23 Jan 29 10:22 ..
drwxr-xr-x. 5 runner runner 48 Jan 29 10:23 aws
/home/runner/.cache/terragrunt/providers/registry.terraform.io/hashicorp/aws:
total 0
drwxr-xr-x. 5 runner runner 48 Jan 29 10:23 .
drwxr-xr-x. 3 runner runner 17 Jan 29 10:22 ..
drwxr-xr-x. 3 runner runner 25 Jan 29 10:23 5.77.0
drwxr-xr-x. 3 runner runner 25 Jan 29 10:22 5.99.1
drwxr-xr-x. 3 runner runner 25 Jan 29 10:23 6.28.0

**snip**
```

- Is that expected and docs are just wrong?
- Will this PR change that behavior?

Versions:

```sh
$ terraform --version
Terraform v1.12.1
on linux_amd64
Your version of Terraform is out of date! The latest version
is 1.14.4. You can update by downloading from https://developer.hashicorp.com/terraform/install

$ terragrunt --version
terragrunt version v0.99.0
```

@yhakbar

PR #5413, opened 2026-01-26
feat: Improving filter parsing errors

yhakbar:
## Description

Improves the error messages returned when users use invalid filter queries.

This is what parse errors look like right now:
current-parse-errors

This is what they've been updated to look like:
updated-parse-errors

Generally, hints are only suggested when there's a reliable suggestion that can fix the user's filter query, and the error message tries to reference the exact location that's broken in the filter query when there's an error.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Release Notes

* **New Features**
  * Enhanced filter parsing error diagnostics with contextual hints, precise position markers, and formatted error messages for improved troubleshooting.
  * Added color-aware error output formatting for filter parsing errors when supported.

* **Tests**
  * Added comprehensive test suite for filter error diagnostics and hint generation.

### DIFF
--- a/internal/filter/hints.go
+++ b/internal/filter/hints.go
@@ -33,16 +33,6 @@ func GetHint(code ErrorCode, token, query string, position int) string {
 	return ""
 }
 
-// GetHints returns context-aware hints for a parse error (legacy, returns slice for compatibility).
-func GetHints(code ErrorCode, token, query string, position int) []string {
-	hint := GetHint(code, token, query, position)
-	if hint == "" {
-		return nil
-	}
-
-	return []string{hint}
-}
-
 // getUnexpectedTokenHint returns a single hint specific to unexpected token errors.
 func getUnexpectedTokenHint(token, query string, position int) string {
 	switch token {


internal/filter/diagnostic.go
```diff
@@ -0,0 +1,77 @@
...
+//	    [main...]
+//	            ^ Expected second Git reference after '...'
+//
+//	 hint: Git filters with '...' require a reference on each side. e.g. '[main...HEAD]'
+//
+// ```
+func FormatDiagnostic(err *ParseError, filterIndex int, useColor bool) string {
+	var sb strings.Builder
+
+	fmt.Fprintf(&sb, "Filter parsing error: %s\n", err.Title)
```
denis256, line 35: Error not handled

internal/filter/hints.go
```diff
@@ -0,0 +1,136 @@
...
+	// These are errors that don't have obvious hints that can be offered.
+	case ErrorCodeUnknown:
+		return ""
+	}
+
+	return ""
+}
+
+// GetHints returns context-aware hints for a parse error (legacy, returns slice for compatibility).
+func GetHints(code ErrorCode, token, query string, position int) []string {
```
denis256: Looks like GetHints() is not used

internal/filter/diagnostic.go
yhakbar: There's no point. `sb.Write` always returns nil:
https://pkg.go.dev/strings#Builder.Write

If we were a project that liked asserts, I could assert that this is true, but I don't really see the value here, and we probably want fuzzing or some more advance testing before we start adding in that kind of aggressive negative space logic.

internal/filter/hints.go
yhakbar: Removed.

PR #5414, opened 2026-01-26
IaC tool CLI arguments re-ordering

### LINKED ISSUE #5409: Regression in v0.98.0+: --all apply misorders flags and plan file
## Describe the bug

Starting with **Terragrunt v0.98.0**, applying a previously saved Terraform/OpenTofu plan using `terragrunt run --all apply` fails.

The failure is caused by Terragrunt generating the Terraform/OpenTofu command with the **plan file passed before flags**, which is not supported by Terraform/OpenTofu.

This issue only occurs when using the `--all` flag. Applying a saved plan without `--all` works as expected.

---

## Reproducing bugs

This bug can be reliably reproduced using a minimal Terragrunt setup.

The issue is specific to the `run --all apply` code path. The same command without `--all` does not exhibit the problem.

---

### Minimal reproducible example

Directory structure:

```
terragrunt-plan-apply-bug/
├── terragrunt.hcl
└── main.tf
```

### main.tf

```hcl
terraform {
  required_version = ">= 1.5.0"
}

resource "null_resource" "example" {}
```

### terragrunt.hcl

```hcl
terraform {
  source = "."
}
```

### Reproduction steps

1. Generate a saved plan:

```sh
terragrunt run --log-level=debug plan -- -out=tfplan
```

2. Apply the saved plan **with `--all`**:

```sh
terragrunt run --log-level=debug --all apply tfplan
```

3. Observe the generated Terraform/OpenTofu command in debug logs and the resulting failure.

### Working case (for comparison)

Running the same command **without `--all`** works:

```sh
terragrunt run --log-level=debug apply tfplan
```

This generates:

```text
terraform apply tfplan
```

---

## Expected behavior

When running:

```sh
terragrunt run --log-level=debug --all apply tfplan
```

Terragrunt should generate the following command:

```text
terraform apply -input=false -auto-approve tfplan
```

The plan file should be passed **after flags**, as required by Terraform/OpenTofu.

---

## Must haves

---

## Nice to haves

- [x] Terminal output
- [ ] Screenshots

---

## Versions

- Terragrunt version: 0.98.0+
- Terraform version: 1.14.3
- Environment details: Linux (Ubuntu)

---

## Additional context

Image

Actual command generated when using `--all`:

```text
terraform apply tfplan -input=false -auto-approve
```

Terraform/OpenTofu fails because the plan file is passed before flags.

This appears to be a regression introduced in Terragrunt v0.98.0 and affects workflows that rely on saved plans (e.g. CI/CD pipelines with plan/apply separation).

The issue seems specific to the `run --all` argument handling, as the same command works correctly without `--all`.

denis256:
## Description

* added parsing of cli arguments passed to IaC tool
* added support for arguments re-ordering
* added tests for plan passing arguments

Fixes #5409.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Release Notes

* **Bug Fixes**
  * Improved handling of custom plan file names (without `.tfplan` extension) in multi-resource deployments.

* **Refactor**
  * Restructured internal argument processing to provide more robust parsing and ordering of Terraform CLI commands and flags.

* **Tests**
  * Added comprehensive test coverage for custom plan file naming scenarios.

### DIFF
--- a/internal/clihelper/args.go
+++ b/internal/clihelper/args.go
@@ -212,17 +212,25 @@ var booleanFlags = []string{
 // IacArgs represents parsed IaC (terraform/tofu) CLI arguments
 // with separate command, flags, and arguments fields.
 // Provides a builder pattern for constructing CLI arguments.
+//
+// Structure: [Command] [SubCommand...] [Flags...] [Arguments...]
+// - Command: main terraform command (e.g., "apply", "providers")
+// - SubCommand: non-flag args before any flags (e.g., "lock" in "providers lock")
+// - Flags: all flags with their values
+// - Arguments: non-flag args after flags (e.g., plan files)
 type IacArgs struct {
-	Command   string   // e.g., "apply", "plan", "destroy"
-	Flags     []string // e.g., "-input=false", "-auto-approve"
-	Arguments []string // e.g., plan files, resource addresses
+	Command    string   // e.g., "apply", "plan", "providers"
+	SubCommand []string // e.g., "lock" in "providers lock -platform=..."
+	Flags      []string // e.g., "-input=false", "-auto-approve"
+	Arguments  []string // e.g., plan files, resource addresses
 }
 
 // NewIacArgs creates IacArgs from strings, parsing command/flags/arguments.
 func NewIacArgs(args ...string) *IacArgs {
 	result := &IacArgs{
-		Flags:     make([]string, 0),
-		Arguments: make([]string, 0),
+		SubCommand: make([]string, 0),
+		Flags:      make([]string, 0),
+		Arguments:  make([]string, 0),
 	}
 	result.parse(args)
 
@@ -309,12 +317,13 @@ func (a *IacArgs) MergeFlags(other *IacArgs) *IacArgs {
 
 // Slice returns args in correct order: [command] [flags...] [arguments...]
 func (a *IacArgs) Slice() []string {
-	result := make([]string, 0, 1+len(a.Flags)+len(a.Arguments))
+	result := make([]string, 0, 1+len(a.SubCommand)+len(a.Flags)+len(a.Arguments))
 
 	if a.Command != "" {
 		result = append(result, a.Command)
 	}
 
+	result = append(result, a.SubCommand...)
 	result = append(result, a.Flags...)
 	result = append(result, a.Arguments...)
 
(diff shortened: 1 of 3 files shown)


internal/clihelper/args.go
```diff
@@ -178,3 +178,355 @@ func (args Args) SubCommandName() string {
 func (args Args) Contains(target string) bool {
 	return slices.Contains(args, target)
 }
+
+// booleanFlags contains flags that don't take values (boolean flags).
+// Unknown flags are assumed to take space-separated values for safer parsing
+// of new Terraform/Tofu flags without requiring updates to this list.
+//
+// WARNING: If a new boolean flag is added to Terraform/Tofu and not listed here,
```
yhakbar, line 186: I'm not sure we should be doing this to flags that users can pass in. I think it's a good first step, as we already have a bug, and we aren't going to immediately have to deal with a new boolean flag from OpenTofu/Terraform, but I think we need to avoid having this kind of opinionated handling of inputs from users.

I think a better solution is being careful with how _we_ introduce stuff (always add flags right after the command by prepending the flags slice and always put arguments like plan file names at the end by appending to the argument slice), and treat everything supplied by the user as an argument (maybe in a special "user supplied arguments" slice). 

Everything they supply should go together, and be added after the stuff we manage, so there's never this concern about new boolean flags or anything.

internal/runner/runnerpool/runner.go
```diff
@@ -1047,9 +992,9 @@ func (r *Runner) SetReport(rpt *report.Report) {
 }
 
 // isDestroyCommand checks if the current command is a destroy operation
-func isDestroyCommand(cmd string, args []string) bool {
+func isDestroyCommand(cmd string, args *clihelper.IacArgs) bool {
 	return cmd == tf.CommandNameDestroy ||
-		slices.Contains(args, "-"+tf.CommandNameDestroy)
+		args.Contains("-"+tf.CommandNameDestroy)
```
yhakbar, line 997: Might as well do what the bot says here I think

pkg/config/dependency.go
```diff
@@ -779,7 +780,7 @@ func cloneTerragruntOptionsForDependencyOutput(ctx context.Context, pctx *Parsin
 	// just read outputs, so no need to check for dependent modules
 	targetOptions.CheckDependentUnits = false
 	targetOptions.TerraformCommand = "output"
-	targetOptions.TerraformCliArgs = []string{"output", "-json"}
+	targetOptions.TerraformCliArgs = clihelper.NewIacArgs("output", "-json")
```
yhakbar, line 783: I think when we're constructing the args like this, we should do the disambiguation ourselves at the call site because we have context, so we should set args to output and flags to `-json`.

pkg/options/options.go
```diff
@@ -504,62 +502,49 @@ func (opts *TerragruntOptions) CloneWithConfigPath(l log.Logger, configPath stri
...
 
-	opts.TerraformCliArgs = args
-}
-
-// AppendTerraformCliArgs appends the given argsToAppend after the current TerraformCliArgs.
-func (opts *TerragruntOptions) AppendTerraformCliArgs(argsToAppend ...string) {
-	opts.TerraformCliArgs = append(opts.TerraformCliArgs, argsToAppend...)
+	for _, arg := range parsed.Arguments {
+		opts.TerraformCliArgs.AppendArgument(arg)
+	}
```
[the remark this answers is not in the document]
yhakbar: I think this is potentially valid.

PR #5420, opened 2026-01-27
chore: Adding fuzz tests for `filter`

yhakbar:
## Description

Adds fuzz tests for the `filter` package with a workflow that runs on push to fuzz everything we fuzz in the repo with a 30s budget for each.

As we add more fuzz tests, we might want to drop down that budget so that we can save some time.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive fuzz tests to validate the filter component's robustness against diverse inputs, edge cases, and unexpected conditions.

* **Chores**
  * Enhanced the CI/CD pipeline with automated fuzz testing workflow to continuously validate functionality and catch potential issues early during development.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,10 @@ jobs:
     uses: ./.github/workflows/license-check.yml
     secrets: inherit
 
+  fuzz:
+    uses: ./.github/workflows/fuzz.yml
+    secrets: inherit
+
   base_tests:
     needs: [lint, precommit, codespell, go_mod_tidy_check, markdownlint, install_script_test, license_check]
     uses: ./.github/workflows/base-test.yml
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -1,0 +1,60 @@
+name: Fuzz
+
+on:
+  workflow_call:
+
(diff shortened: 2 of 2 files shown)


internal/filter/fuzz_test.go
```diff
@@ -0,0 +1,277 @@
...
+
+		// Tokenize until EOF - should never hang or panic
+		// Set a reasonable limit to prevent infinite loops in case of bugs
+		const maxTokens = 10000
+		for range maxTokens {
+			tok := lexer.NextToken()
+			if tok.Type == filter.EOF {
+				break
+			}
+		}
```
denis256, line 178: Should we handle the maxTokens limit as an error?

```
        for range maxTokens {
            tok := lexer.NextToken()
            if tok.Type == filter.EOF {
                return
            }
        }
        t.Fatal("Lexer max token limit")

```

yhakbar: I think we can, but isn't there a possibility that the input from the fuzzer will actually have 1000 tokens in it? I guess we'll need an upper bound error that we can throw to gracefully error on in the lexer for the bounds check to make sense here, right?

PR #5430, opened 2026-01-27
fix: Adding lock files to cache key for provider cache

yhakbar:
## Description

Ensures that we update cache when new `.terraform.lock.hcl` files show up or get updated.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Chores
* Optimized test workflow caching to improve CI/CD reliability and reduce unnecessary cache misses

### DIFF
--- a/.github/workflows/base-test.yml
+++ b/.github/workflows/base-test.yml
@@ -49,7 +49,9 @@ jobs:
         uses: actions/cache@v5
         with:
           path: ${{ runner.os == 'Linux' && '~/.cache/terragrunt' || '~/Library/Caches/terragrunt' }}
-          key: ${{ runner.os }}-terragrunt-provider-cache
+          key: ${{ runner.os }}-terragrunt-provider-cache-${{ hashFiles('**/.terraform.lock.hcl') }}
+          restore-keys: |
+            ${{ runner.os }}-terragrunt-provider-cache-
 
       - name: Run Tests
         id: run-tests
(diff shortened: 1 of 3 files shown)


.github/workflows/base-test.yml
```diff
@@ -49,7 +49,9 @@ jobs:
         uses: actions/cache@v5
         with:
           path: ${{ runner.os == 'Linux' && '~/.cache/terragrunt' || '~/Library/Caches/terragrunt' }}
-          key: ${{ runner.os }}-terragrunt-provider-cache
+          key: ${{ runner.os }}-terragrunt-provider-cache-${{ hashFiles('**/.terraform.lock.hcl') }}
```
thisguycodes, line 52: huh... we don't change them much, so it shouldn't *really* matter, but shouldn't the provider cache server handle this well anyway?

though I suppose this will also stop the cache download from getting bloated.

yhakbar: GitHub Actions won't update the cache entry if you have an exact match on the cache key. By doing this, we tell GitHub Actions that it should download the latest entry that matches `${{ runner.os }}-terragrunt-provider-cache-`, then upload the current state of cache as `${{ runner.os }}-terragrunt-provider-cache-${{ hashFiles('**/.terraform.lock.hcl') }}` if that value is any different than the one we downloaded (meaning someone changed a `.terraform.lock.hcl` file and we potentially downloaded new providers during our run).

Issue #3669, opened 2024-12-16
Terragrunt and TFLint above 0.50.0 breaking behavior


velkovb:
## Describe the enhancement

TFLint introduced a new way of handling values for undeclared variables in 0.50.0 - https://github.com/terraform-linters/tflint/pull/1941

This directly affects the Terragrunt hook integration as it does the following:

`tflint --config ./.terragrunt-cache/ellCj9USNaAqucT2-U5SmL6rf1A/Gu7P37gZ1L3NBihh7GJE9nHhW38/.tflint.hcl --chdir ./.terragrunt-cache/ellCj9USNaAqucT2-U5SmL6rf1A/Gu7P37gZ1L3NBihh7GJE9nHhW38 --var=test=foo` 

With Terragrunt release of `0.69.0` which bumped TFLint from 0.47.0 to 0.50.3 this started breaking our projects. We do generate a lot of metadata and automatic variables that are not used by every project. https://github.com/terraform-linters/tflint/discussions/2080#discussioncomment-11584584

The way that variables are passed to TFLint doesn't match the way they are passed to OpenTofu/ Terraform. For TFLint it seems to be with `--var` while for the latter it is `TF_VAR_*` which has no issues. If we pass OpenTofu/Terraform with `--var` it will error out as well.

Would be great if we can align both approaches and pass variables from input to TFLint as TF_VAR_* as well. If that cannot be done, passing a `.tfvars` file should work as well. I would guess the magic needs to happen somewhere here - https://github.com/gruntwork-io/terragrunt/blob/407f65fdb294559f3f70ac873aae5a41715d68dd/tflint/tflint.go#L154

Not sure if this is a bug or enhancement

## Additional context

Things you might want to address include:

- [ ] Changes required.
- [ ] Implications of the feature.
- [ ] Alternatives considered.
- [ ] Level of effort.

## PoC (Proof of Concept)

Link to a Proof of Concept if you have one:

- [ ] [PoC]()

Including a PoC can help others understand the feature better and implement it faster.

## RFC Not Needed

- [ ] I have evaluated the complexity of this enhancement, and I believe it does not require an RFC.

yhakbar:
Hey @velkovb ,

Thanks for creating this issue. 

Unfortunately, given that there's a very straightforward workaround (calling tflint with the binary installed in PATH however users like), there isn't much incentive to adjusting how the built-in `tflint` hook works.

[...]

PR #5437, opened 2026-01-28
docs: add instructions on bypassing tflint integration behavior

thisguycodes:
## Description

Reading through https://github.com/gruntwork-io/terragrunt/issues/3669 it became apparent that there can be valid reasons to bypass the tflint integration behavior (auto init and / or passing in variables).

This PR adds documentation on how one might do this bypassing.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance on using absolute paths to bypass integration behavior for tflint hooks, including a configuration example.

### DIFF
--- a/docs-starlight/src/content/docs/03-features/10-hooks.mdx
+++ b/docs-starlight/src/content/docs/03-features/10-hooks.mdx
@@ -300,6 +300,19 @@ terraform {
 }
 ```
 
+If you need to bypass the integration behavior (auto init or passing in variables), you can specify an absolute path to `tflint`, as this integration works by checking the first value in the `execute` array:
+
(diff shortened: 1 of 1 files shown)


thisguycodes:
> I think a better approach here is to add opt-out flags for the hook.
> 
> e.g. `--terragrunt-no-auto-tflint-init`, etc.
> 
> I initially assumed we'd just not do any special handling of the tflint invocation when we removed the dependency, so this is a little odd to me.

I also had this thought initially, but when adding the strict control I saw we did the specially handling for external tflint as well, and my immediate thought was "that's too useful to also break at the same time"

PR #5442, opened 2026-01-28
fix: Fixing `TestAwsBootstrapBackendWithAccessLoggingFlake`

yhakbar:
## Description

We have safeguards to protect against concurrent creation and error handling retries of the state logging buckets like we have for the main state bucket.

Results in less flakes for the `TestAwsBootstrapBackendWithAccessLogging` test.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced S3 bucket creation with robust retry logic for transient and concurrent scenarios.
  * Now waits for bucket existence and gracefully handles already-owned cases.
  * Improved detection and logging of retriable vs unretriable errors, and clearer bucket-name-aware messages.
  * Removed silent error suppression so failures are surfaced reliably.

### DIFF
--- a/internal/remotestate/backend/s3/client.go
+++ b/internal/remotestate/backend/s3/client.go
@@ -735,16 +735,16 @@ func isBucketAlreadyOwnedByYouError(err error) bool {
 func isBucketErrorRetriable(l log.Logger, err error) bool {
 	var apiErr smithy.APIError
 	if errors.As(err, &apiErr) {
-		retriable := apiErr.ErrorCode() == "InternalError" || apiErr.ErrorCode() == "OperationAborted" || apiErr.ErrorCode() == "InvalidParameter"
+		unrecoverable := apiErr.ErrorCode() == "InternalError" || apiErr.ErrorCode() == "OperationAborted" || apiErr.ErrorCode() == "InvalidParameter"
 
-		if !retriable {
+		if !unrecoverable {
 			l.Debugf(
 				"Encountered AWS API error '%s' during bucket creation. Assuming it's retriable and will retry.",
 				apiErr.ErrorCode(),
 			)
 		}
 
-		return retriable
+		return unrecoverable
 	}
 
 	l.Debugf(


internal/remotestate/backend/s3/client.go
```diff
@@ -715,13 +732,26 @@ func isBucketAlreadyOwnedByYouError(err error) bool {
...
+		retriable := apiErr.ErrorCode() == "InternalError" || apiErr.ErrorCode() == "OperationAborted" || apiErr.ErrorCode() == "InvalidParameter"
+
+		if !retriable {
+			l.Debugf(
+				"Encountered AWS API error '%s' during bucket creation. Assuming it's retriable and will retry.",
+				apiErr.ErrorCode(),
+			)
+		}
+
+		return retriable
```
thisguycodes: "if not retriable, retry" reads like it's a mistake? What's the intention here?

plus it actually returns `false` in these cases

ThisGuyCodes (dismissed):
mostly looks fine, but can you link to one of the flakes? I'm wondering if maybe we're seeing multiple test runs trying to use the same bucket name?

internal/remotestate/backend/s3/client.go
yhakbar: Good catch! Updated the name.

yhakbar:
> mostly looks fine, but can you link to one of the flakes? I'm wondering if maybe we're seeing multiple test runs trying to use the same bucket name?

Don't really want to be digging for it, but I think odds are somewhat low:

```go
	testID := strings.ToLower(helpers.UniqueID())

	s3BucketName := "terragrunt-test-bucket-" + testID
	s3AccessLogsBucketName := "terragrunt-test-bucket-" + testID + "-access-logs"
	dynamoDBName := "terragrunt-test-dynamodb-" + testID
```

We can start to also name the buckets after their test names to further reduce the likelihood of collisions, but we'll also have to be careful with S3 bucket name length restrictions.

PR #5447, opened 2026-01-28
feat: use github auth for engine release checks / downloads

thisguycodes:
## Description

Not having this auth might have caused a few flakes. But also anonymouns API limits are pretty tight, and IP based, so we should be good citizens and do this when possible.

Has the release check and downloads for engines use an available `GH_TOKEN` or `GITHUB_TOKEN`, same strategy as the `gh` cli.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

Use a github token when checking / downloading engine releases (if available)

## Summary by CodeRabbit

* **New Features**
  * Added GitHub token-based authentication via env vars (GH_TOKEN or GITHUB_TOKEN), with GH_TOKEN taking precedence; client now sends default auth and API-version headers on requests.

* **Tests**
  * Added tests for token-precedence and authenticated release-asset download flows.

* **Chores**
  * Minor test robustness and helper improvements.

### DIFF
--- a/internal/github/client.go
+++ b/internal/github/client.go
@@ -78,9 +78,12 @@ func WithGithubComDefaultAuth() GitHubAPIClientOption {
 func getGithubTokenFromEnv() string {
 	if tok := os.Getenv("GH_TOKEN"); tok != "" {
 		return tok
-	} else if tok := os.Getenv("GITHUB_TOKEN"); tok != "" {
+	}
+
+	if tok := os.Getenv("GITHUB_TOKEN"); tok != "" {
 		return tok
 	}
+
 	return ""
 }
 
--- a/internal/github/client_test.go
+++ b/internal/github/client_test.go
@@ -296,18 +296,22 @@ func TestDownloadReleaseAssetsGitHubRelease(t *testing.T) {
 			w.Header().Set("Content-Type", "application/zip")
 			fmt.Fprint(w, "fake-zip-content")
 			return
-		} else if strings.HasSuffix(path, "SHA256SUMS") {
+		}
+
+		if strings.HasSuffix(path, "SHA256SUMS") {
 			w.Header().Set("Content-Type", "text/plain")
 			fmt.Fprint(w, "fake-checksum-content")
 			return
-		} else if strings.HasSuffix(path, "SHA256SUMS.sig") {
+		}
+
+		if strings.HasSuffix(path, "SHA256SUMS.sig") {
 			w.Header().Set("Content-Type", "text/plain")
 			fmt.Fprint(w, "fake-signature-content")
 			return
-		} else {
-			w.WriteHeader(http.StatusNotFound)
-			return
 		}
+
+		w.WriteHeader(http.StatusNotFound)
+		return
 	}))
 	defer server.Close()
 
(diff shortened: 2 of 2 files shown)


internal/github/client.go
```diff
@@ -50,13 +54,42 @@ func WithBaseURL(baseURL string) GitHubAPIClientOption {
...
+		if tok := getGithubTokenFromEnv(); tok != "" {
+			c.defaultHeaders.Set("Authorization", "Bearer "+tok)
+		}
+	}
+}
+
+func getGithubTokenFromEnv() string {
+	if tok := os.Getenv("GH_TOKEN"); tok != "" {
+		return tok
+	} else if tok := os.Getenv("GITHUB_TOKEN"); tok != "" {
```
yhakbar: NIT: This else doesn't do anything.

internal/github/client_test.go
```diff
@@ -274,6 +308,78 @@ func TestDownloadReleaseAssetsGitHubRelease(t *testing.T) {
...
 }
 
+func TestDownloadReleaseAssetsGitHubReleaseUsesToken(t *testing.T) {
+	tempDir := helpers.TmpDirWOSymlinks(t)
+
+	// shared logic for handlers
+	doResp := func(w http.ResponseWriter, r *http.Request) {
+		// Serve different content based on the requested file
+		path := r.URL.Path
+		if strings.HasSuffix(path, "package.zip") {
```
yhakbar, line 348: NIT: Would really recommend early returns here, as this is really dense.

internal/github/client_test.go
```diff
@@ -274,6 +308,78 @@ func TestDownloadReleaseAssetsGitHubRelease(t *testing.T) {
 	verifyFileContent(t, result.PackageFile, "fake-zip-content")
 }
 
+func TestDownloadReleaseAssetsGitHubReleaseUsesToken(t *testing.T) {
```
yhakbar, line 340: The linter will tell you, but make sure these call `t.Parallel()`

internal/github/client_test.go
```diff
@@ -27,6 +27,40 @@ func TestNewClient(t *testing.T) {
...
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, "Bearer goodtoken", r.Header.Get("Authorization"))
+
+		w.Header().Set("Content-Type", "application/json")
+		response := `{
+			"tag_name": "v1.2.3",
+			"name": "Release v1.2.3",
+			"html_url": "https://github.com/owner/repo/releases/tag/v1.2.3"
+		}`
+		fmt.Fprint(w, response)
```
yhakbar: Make sure you require that this doesn't err.

internal/github/client_test.go
thisguycodes: I copied this from an existing test 😅 , but yea it can be cleaned up

internal/github/client.go
thisguycodes: I don't follow, there are two possible environment variables for github tokens. I link to the `gh` cli docs on this in a doc comment (on a different function, I can update): https://cli.github.com/manual/gh_help_environment

internal/github/client_test.go
thisguycodes: these tests use `t.Setenv()` so we can't run them parallel, and the linter recognizes this

internal/github/client.go
yhakbar: If you hit the branch before the else, you return, so you can make this easier (for me) to read by making this a standalone if statement instead of a chain of else and else ifs. I really value white space when reading the code in this code base as it makes it easier for me to think about one thing at a time.

It's totally valid if you just prefer this stylistically, though.

thisguycodes: finally understood what you're saying, and I agree

internal/github/client.go
```diff
@@ -249,6 +291,29 @@ func (c *GitHubReleasesDownloadClient) DownloadReleaseAssets(
...
+				// test servers don't use https, but we don't usually want to send auth tokens unencrypted
+				if testing.Testing() {
+					client.Getters["http"] = &getter.HttpGetter{
+						Netrc: true,
+						Header: http.Header{
+							"Authorization": {"Bearer " + tok},
+						},
+					}
+				}
+			}
```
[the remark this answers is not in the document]
thisguycodes: in this context we're only ever talking to github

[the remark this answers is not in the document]
thisguycodes: those are tests, because we don't want to talk to github in tests, they use a 127 address just as a consequence of implementation

PR #5454, opened 2026-01-29
feat: support gcs buckets as unit and stack source

### LINKED ISSUE #5453: Support all go-getter content providers as source in stacks
## Describe the enhancement

Currently terragrunt fails if a source other than git or local is defined in stacks

```hcl
stack "mystack" {
    source = "gcs::https://www.gogleapis.com/storage/v1/my-bucket/stack"
}
```

## Additional context

The effor is super small, a simple change in in the copyFiles method in pkg/config/stack.go is enough.

## PoC (Proof of Concept)

s3 and gcs should be usable in source

```hcl
stack "mystack" {
    source = "gcs::https://www.gogleapis.com/storage/v1/my-bucket/stack"
}
```

## RFC Not Needed

- [X] I have evaluated the complexity of this enhancement, and I believe it does not require an RFC.

mzeiher:
## Description

Fixes #5453 .

This PR enables other sources as stack and unit sources besides local, git, terraform registry. E.g. gcs or s3

## TODOs

- [X] I am submitting code based on open source software (e.g. MIT, MPL-2.0, Apache)]
- [X] I am adding or upgrading a dependency or adapted code and confirm it has a compatible open source license
- [ ] Update the docs.
- [X] Include release notes. If this PR is backward incompatible, include a migration guide.

## Release Notes (draft)

Added support for additional source providers in stacks and units.

## Summary by CodeRabbit

* **Chores**
  * Updated path handling dependency to latest stable version.

* **Refactor**
  * Simplified local path detection logic.
  * Improved context handling mechanism for file operations.

### DIFF
--- a/pkg/config/stack.go
+++ b/pkg/config/stack.go
@@ -14,7 +14,7 @@ import (
 	"github.com/gruntwork-io/terragrunt/internal/ctyhelper"
 	"github.com/gruntwork-io/terragrunt/internal/worker"
 
-	"github.com/hashicorp/go-getter/v2"
+	"github.com/hashicorp/go-getter"
 
 	"github.com/hashicorp/hcl/v2/hclsyntax"
 
@@ -325,7 +325,7 @@ func copyFiles(ctx context.Context, l log.Logger, identifier, sourceDir, src, de
 			return errors.Errorf("Failed to create directory %s for %s %w", dest, identifier, err)
 		}
 
-		if _, err := getter.GetAny(ctx, dest, src); err != nil {
+		if err := getter.GetAny(dest, src, getter.WithContext(ctx)); err != nil {
 			return errors.Errorf("Failed to fetch %s %s for %s %w", src, dest, identifier, err)
 		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/config/stack.go
```diff
@@ -344,27 +344,12 @@ func isLocal(l log.Logger, workingDir, src string) bool {
...
-		recognized, err := getter.Detect(req, g)
-		if err != nil {
-			l.Debugf("Error detecting getter for %s: %v", src, err)
-			continue
-		}
-
-		if recognized {
-			break
-		}
-	}
```
mzeiher, line 365: this did never actually work since the concatenated string was passed to the getter.Detect which was cwd + source
if cwd was /work and source was git::https://source.dev getter.Detect checked against "/workgit::https://source.dev

I now changed this to only check for local files and forward the original source to go-getter 1.8.5 which then detects e.g. the gcs:: scheme

mzeiher:
@yhakbar
thats interesting the v1.8.5 is more recent in respect to supporting different backends than the v2.2.3. In this case the 2.2.3 does not support these backends natively, I could also add the github.com/hashicorp/go-getter/gcs/v2 and add them to the detectors if this is the preffered way?

does the 2 vs 1 has something to do with the change in licenses?

yhakbar:
> I could also add the github.com/hashicorp/go-getter/gcs/v2 and add them to the detectors if this is the preffered way?

Ya, we'd also need some testing added in this PR to confirm that they work right.

> does the 2 vs 1 has something to do with the change in licenses?

As far as I know it's just that v2 had some flexibility introduced that required a breaking change.

Issue #5467, opened 2026-01-30
stack run plan with value is not known silently skips the terragrunt file


lorelei-rupp-imprivata:
## Describe the bug

We have a terragrunt unit that is being silently skipped when we run `terragrunt stack run plan` from the stack
We expect this to fail and say there is a syntax error/value not known because we removed something from an env.yaml it was expecting and now it cannot compile

This is what debug logs look like

```
09:21:46.915 DEBUG  Unrecoverable parse error for ./pod-identity/aaaa/terragrunt.hcl: value is not known
09:21:46.916 DEBUG  Found locals block: evaluating the expressions.
09:21:46.917 DEBUG  git show-toplevel result: ../../..
09:21:46.917 DEBUG  Suppressed parsing errors %!w(hcl.Diagnostics=[0xc000cf9a40])
09:21:46.917 DEBUG  Evaluated 2 locals (remaining 0): terraform_cache_dir, account
09:21:46.918 DEBUG  [Partial] Included config ../root_aws_new.hcl has strategy shallow merge: merging config in (shallow).
09:21:46.918 DEBUG  Suppressed parsing errors %!w(hcl.Diagnostics=[0xc0007e1080])
09:21:46.918 DEBUG  Suppressed parsing errors %!w(hcl.Diagnostics=[0xc0007e10e0])
09:21:48.087 WARN   Parsing errors were encountered while discovering relationships. They were suppressed, and can be found in the debug logs.
09:21:48.087 DEBUG  Errors: value is not known
value is not known
09:21:48.087 DEBUG  Runner pool discovery found 26 configs
09:21:48.090 DEBUG  [pod-identity/aaa] Unit ./pod-identity/aaa does not have an associated terraform configuration and will be skipped.
```

## Expectation
If I have a syntax error or im missing something somewhere in my values/env.yaml.. it should fail and let me know I have an issue. Right now it skips the module completely and I have no idea I have a problem and now my terragrunt module is "unlinked" from my pipelines

## Reproduce
Pull something out of the terragrunt.values.hcl file that your terragrunt expects

## Versions

- Terragrunt version: 0.99.1
- OpenTofu/Terraform version: 1.12.2

yhakbar:
Just wanted to confirm that there's definitely a bug here, and definitely an important bug to address. We'll be working on addressing it in the near term to get a fix out. It will just take a bit of internal restructuring to get the kind of fix we're looking for.

yhakbar:
[...]

PR #5477, opened 2026-01-31
chore: Refactor discovery for nuanced parse handling

yhakbar:
## Description

Redesigns the internal/discovery package to use a multi-phase pipeline architecture that lazily evaluates components based on filter requirements. This eliminates unnecessary parsing when there's a nuanced set of requirements for _which_ components need parsing.

For example, say a user uses the following filter query:

```bash
terragrunt run --filter './foo' --filter '{./bar}...' --filter '!./baz'
```

Currently, the logic in discovery sees that dependency expression on `bar` and decides that _everything_ discovered needs parsing by setting `RequiresParse` to `true`. That's a lot more coarse than things need to be. We don't, for example, need to parse `baz` unless it's in the dependency graph of `bar`. We also don't need to parse anything that isn't in the dependency graph of `bar` or `foo`.

The goal of this update is to classify the filter expressions provided and make decisions during discovery as to whether any particular component can be dropped earlier in discovery based on whether the filters users use make it impossible to end up in the final result set.

**Key changes:**
  - Replaced monolithic discovery call with 5-phase pipeline: Filesystem → Parse → Graph → Relationship → Filter
  - Added filter.Classifier that analyzes filter expressions upfront to determine minimum evaluation requirements
  - Implemented dual-channel output (discovered/candidates) enabling lazy evaluation between phases
  - Components only parsed when filters require it (e.g., reading=config/*, source=*)

Components are classified into three statuses during discovery:
  - StatusDiscovered: Matches a positive filter (path, attribute, or git expression). These are considered fully discovered and will end up at the final filter stage.
  - StatusCandidate: Needs further evaluation (graph target, requires parsing, or potential dependent). These are considered _candidates_ for discovery and are sent through the subsequent discovery phases to see if something can qualify it (e.g. it's confirmed to be a dependency of something in a graph expression).
  - StatusExcluded: Only matches negated filters, or positive filters exist but none match. These are disqualified early and are not sent to the final filter phase.

**Phase Flow**

  1. Filesystem + Worktree (concurrent): Walk directories, classify via CandidacyClassifier. For Git filters ([ref...ref]), discover in temporary worktrees
  2. Parse (if needed): Parse HCL for candidates with CandidacyReasonRequiresParse, re-classify
  3. Graph (if needed): Traverse dependencies (target|N) and/or dependents (...target)
  4. Relationship (optional): Build complete dependency graph for execution ordering (this phase is skipped for discovery commands without the use of `--dag`).
  5. Final: Merge, deduplicate, apply final filter evaluation, cycle detection

**Side-effects**

This PR started out because #5469 was too hacky, and we needed a fix for #5467.

We were silently ignoring failed parsing due to logic like [`ShouldSkipUnitWithoutTerraform`](https://github.com/gruntwork-io/terragrunt/pull/5477/changes#diff-79651e289f124fef748d811c508678ee163db9edd3f1fb902579cfe67568b389L119) and [`WithSuppressParseErrors`](https://github.com/gruntwork-io/terragrunt/pull/5477/changes#diff-e2175d9b0580a48170aefebba27455409d5f37c306bedc8095793588014d7d9fL104).

Removing this logic resulted in quite a lot of fallout, as we had tests that were silently passing due to the suppressions and skips. It's likely that there are users out there that had Terragrunt silently ignoring parsing errors, and that they'll need to start explicitly excluding things with negated filters or fix their configurations.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **Refactor**
  * Discovery rebuilt into a phased, builder-configurable pipeline with consistent path resolution and explicit phase outputs.

* **New Features**
  * Filesystem, parse, graph, relationship and worktree discovery phases; classifier-driven filters; graph-target pruning; read-file support; concurrent multi-worker discovery.

* **Bug Fixes**
  * Parse errors now surface by default; units without Terraform files are processed; improved graph and dependency handling.

* **Documentation & Tests**
  * Expanded package docs and a large suite of new unit and integration tests; some legacy tests skipped/removed.

---

> [!NOTE]
> **High Risk**
> Large refactor of core discovery and dependency graph construction; behavior around filtering, parsing, graph traversal, and exclusion semantics can change in subtle ways and impacts many commands.
> 
> **Overview**
> Refactors `internal/discovery` from a monolithic walk/parse/graph flow into a **multi-phase pipeline** (filesystem/worktree → parse → graph → optional relationship → final filtering) driven by a `filter.Classifier` so only components that need deeper evaluation are parsed or graph-traversed.
> 
> This removes the legacy `DependencyDiscovery`/`DependentDiscovery` implementations in favor of phase runners, updates discovery construction defaults (`NewDiscovery(dir)` with sensible defaults and `WithParseIncludes` rename), and centralizes symlink-aware comparisons via `util.ResolvePath` (including `ThreadSafeComponents` caching).
> 
> Behavioral changes include: earlier filter-based exclusion, optional pre-building of reverse dependency links for dependent graph filters, cycle removal via `WithBreakCycles`, and exclude blocks optionally marking dependencies as excluded when `exclude_dependencies=true` and `if=true`. A large new integration test suite and reorganized unit tests cover hidden dir handling, parsing/reading tracking, graph filters, external deps, cycles, and exclude semantics; package docs were added.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9e7349603abe683cd859ff238b438691781498b0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


yhakbar:
@coderabbit review

yhakbar:
@coderabbit review

yhakbar:
@coderabbit review

internal/discovery/constructor.go
```diff
@@ -138,21 +138,16 @@ func NewForStackGenerate(l log.Logger, opts StackGenerateOptions) (*Discovery, e
 	return d, nil
 }
 
-// NewDiscovery creates a new Discovery.
-func NewDiscovery(dir string, opts ...DiscoveryOption) *Discovery {
+// NewDiscovery creates a new Discovery with sensible defaults.
+func NewDiscovery(dir string) *Discovery {
```
denis256, line 142: Looks like a duplicate of `discovery.go:24`

internal/discovery/phase_worktree.go
```diff
@@ -0,0 +1,605 @@
...
+	// gitExpressions contains Git filter expressions that require worktree discovery.
+	gitExpressions filter.GitExpressions
+	// numWorkers is the number of concurrent workers.
+	numWorkers int
+}
+
+// NewWorktreePhase creates a new WorktreePhase.
+func NewWorktreePhase(gitExpressions filter.GitExpressions, numWorkers int) *WorktreePhase {
+	if numWorkers <= 0 {
+		numWorkers = runtime.NumCPU()
```
denis256, line 37: Can be shorter:
```
numWorkers = max(min(runtime.NumCPU(), maxDiscoveryWorkers), defaultDiscoveryWorkers)

```

yhakbar:
@coderabbit review

internal/discovery/phase_filesystem.go
```diff
@@ -0,0 +1,223 @@
...
+			default:
+			}
+
+			result := p.processFile(input, path, filenames)
+			if result == nil {
+				continue
+			}
+
+			select {
+			case <-ctx.Done():
```
denis256: It may hang forever if will not be closed by Rhase.Run() implementation, in case of an error, for example

internal/discovery/phase_worktree.go
```diff
@@ -0,0 +1,605 @@
...
+	ctx context.Context,
+	input *PhaseInput,
+	w *worktrees.Worktrees,
+) (component.Components, error) {
+	discoveredComponents := component.NewThreadSafeComponents(component.Components{})
+
+	stackDiff := w.Stacks()
+
+	g, ctx := errgroup.WithContext(ctx)
+	g.SetLimit(min(runtime.NumCPU(), len(stackDiff.Added)+len(stackDiff.Removed)+len(stackDiff.Changed)*2))
```
denis256: What will happen if the environment has 0.5 CPU?
I think on 0 limit it will panic, we may need to set some minimal value?  🤔

internal/discovery/phase_graph.go
```diff
@@ -0,0 +1,841 @@
...
+				Component: dependent,
+				Status:    StatusDiscovered,
+				Reason:    CandidacyReasonNone,
+				Phase:     PhaseGraph,
+			}:
+			case <-ctx.Done():
+				return ctx.Err()
+			}
+
+			err := p.discoverDependents(
```
denis256: Possible deadlock - on recursive call and same discovered channel

internal/discovery/phase_filesystem.go
```diff
@@ -0,0 +1,223 @@
...
+		return
+	}
+
+	filenames := discovery.configFilenames
+	if len(filenames) == 0 {
+		filenames = DefaultConfigFilenames
+	}
+
+	g, ctx := errgroup.WithContext(ctx)
+	g.SetLimit(p.numWorkers)
```
denis256: Limit is set on numWorkers, but g.Go() is invoked only 2 times

internal/discovery/phase_parse.go
```diff
@@ -0,0 +1,318 @@
...
+		Status:    candidate.Status,
+		Reason:    candidate.Reason,
+		Phase:     PhaseParse,
+	}, nil
+}
+
+// parseComponent parses a Terragrunt configuration.
+func parseComponent(
+	c component.Component,
+	ctx context.Context,
```
denis256, line 198: I think the convention is to pass context first

internal/discovery/phase_parse.go
```diff
@@ -0,0 +1,318 @@
...
+		config.ErrorsBlock,
+		config.RemoteStateBlock,
+	).WithSkipOutputsResolution()
+
+	if len(parserOptions) > 0 {
+		parsingCtx = parsingCtx.WithParseOption(parserOptions)
+	}
+
+	if suppressParseErrors {
+		opts := parsingCtx.ParserOptions
```
denis256: Possible shadow of variable `opts` is a parameter to the function

internal/discovery/phase_relationship.go
```diff
@@ -0,0 +1,292 @@
...
+// dependencyToDiscover resolves a dependency path and links it to the component.
+func (p *RelationshipPhase) dependencyToDiscover(
+	c component.Component,
+	path string,
+	allComponents *component.Components,
+	interTransientComponents *component.ThreadSafeComponents,
+	discovery *Discovery,
+) (component.Component, bool) {
+	for _, dep := range *allComponents {
+		if dep.Path() == path {
```
denis256, line 244: Can be a map lookup? 🤔

internal/discovery/phase_worktree.go
```diff
@@ -0,0 +1,605 @@
...
+func MatchComponentPairs(
+	fromComponents component.Components,
+	toComponents component.Components,
+) []ComponentPair {
+	componentPairs := make([]ComponentPair, 0, max(len(fromComponents), len(toComponents)))
+
+	for _, fromComponent := range fromComponents {
+		fromComponentSuffix := strings.TrimPrefix(
+			fromComponent.Path(),
+			fromComponent.DiscoveryContext().WorkingDir,
```
denis256, line 448: Can be a NPE if  fromComponent.DiscoveryContext() is not initialized?

internal/filter/classifier.go
```diff
@@ -0,0 +1,403 @@
...
+		}
+
+		return true
+	})
+}
+
+// matchesAnyNegated checks if the component matches any negated expression.
+func (c *Classifier) matchesAnyNegated(comp component.Component) bool {
+	return slices.ContainsFunc(c.negatedExprs, func(expr Expression) bool {
+		match, _ := MatchComponent(comp, expr)
```
denis256: Error is skipped

internal/discovery/phase_worktree.go
yhakbar: I believe `NumCPU` is guaranteed to return at least one, even in containerized environments with less than a ful core allocated. I can add a lower bound of 1 to be safe.

internal/discovery/phase_relationship.go
yhakbar: I'm not sure it would be faster in most cases. I'm generally of a mind that slices should be preferred unless there's a requirement for de-duplication or the cost of iteration is shown to be too high for the average user.

internal/discovery/phase_worktree.go
yhakbar: Added an explicit panic, because I don't think there's another sensible thing to do here.

internal/filter/classifier.go
yhakbar: Logging warning instead of ignoring now.

internal/discovery/phase_parse.go
yhakbar: shadow fixed.

internal/discovery/phase_parse.go
yhakbar: Fixed.

internal/discovery/phase_filesystem.go
yhakbar: Fixed.

internal/discovery/phase_filesystem.go
yhakbar: I just pulled out all these result channels. They weren't doing much to support the transition between phases anyways, as I wasn't able to avoid a channel drain between phases.

internal/discovery/phase_graph.go
yhakbar: Removed

internal/discovery/phase_worktree.go
```diff
@@ -0,0 +1,586 @@
...
+		// Writer interface, but we don't care about those errors.
+		_, _ = hash.Write([]byte(normalizedPath))
+		_, _ = hash.Write([]byte{0})
+
+		f, err := os.Open(path)
+		if err != nil {
+			return "", fmt.Errorf("could not open file %s: %w", path, err)
+		}
+
+		_, err = io.Copy(hash, f)
```
denis256, line 568: We read entire file... why not to use util.FileSHA256()? which read a small block to generate checksum

internal/discovery/phase_relationship.go
```diff
@@ -0,0 +1,265 @@
...
+	if len(paths) == 0 {
+		return nil
+	}
+
+	depsToDiscover := make(component.Components, 0, len(paths))
+
+	for _, path := range paths {
+		dep, created := p.dependencyToDiscover(c, path, state.allComponents, state.interTransientComponents, state.discovery)
+
+		terminalComponents = slices.DeleteFunc(terminalComponents, func(tc component.Component) bool {
```
denis256: possible race condition on terminalComponents - terminalComponents slice shared between goroutines

internal/filter/classifier_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+import (
+	"testing"
+
+	"github.com/gruntwork-io/terragrunt/internal/filter"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestClassifier_NegatedGraphExpression_HasGraphFilters(t *testing.T) {
```
denis256, line 13: no tests for `Classify()` function

internal/discovery/phase_relationship.go
yhakbar: Added mutex.

internal/discovery/phase_worktree.go
yhakbar: We chatted offline. It's probably fine.

internal/filter/classifier_test.go
yhakbar: Added tests.

internal/discovery/phase_worktree.go
```diff
@@ -0,0 +1,586 @@
...
+// MatchComponentPairs matches components between from and to stacks by their relative paths.
+func MatchComponentPairs(
+	fromComponents component.Components,
+	toComponents component.Components,
+) []ComponentPair {
+	componentPairs := make([]ComponentPair, 0, max(len(fromComponents), len(toComponents)))
+
+	for _, fromComponent := range fromComponents {
+		if fromComponent.DiscoveryContext() == nil {
+			panic(
```
denis256: Panic will break execution flow and will not be clear from where nil context apeared 🤔

internal/filter/classifier.go
```diff
@@ -0,0 +1,430 @@
...
+		return true
+	})
+}
+
+// matchesAnyNegated checks if the component matches any negated expression.
+func (c *Classifier) matchesAnyNegated(l log.Logger, comp component.Component) bool {
+	return slices.ContainsFunc(c.negatedExprs, func(expr Expression) bool {
+		match, err := MatchComponent(comp, expr)
+		if err != nil {
+			l.Warnf("Error matching component %s against negated expression: %v", comp.Path(), err)
```
denis256, line 285: Silently swallows `MatchComponent` errors

internal/filter/candidacy.go
```diff
@@ -0,0 +1,168 @@
...
+
+	case *InfixExpression:
+		leftInfo := CandidacyInfo{}
+		rightInfo := CandidacyInfo{}
+
+		analyzeExpressionCandidacy(node.Left, &leftInfo)
+		analyzeExpressionCandidacy(node.Right, &rightInfo)
+
+		// Merge: if either side requires more, the whole expression does
+		info.RequiresFilesystemOnly = leftInfo.RequiresFilesystemOnly && rightInfo.RequiresFilesystemOnly
```
denis256: Should be merged `IsNegated`, `ExcludeTarget`, `DependencyDepth` ?

internal/discovery/phase_parse.go
```diff
@@ -0,0 +1,292 @@
...
+		if discovery.suppressParseErrors {
+			var notFoundErr config.TerragruntConfigNotFoundError
+			if errors.As(err, &notFoundErr) {
+				l.Debugf("Skipping missing config during discovery: %s", parseOpts.TerragruntConfigPath)
+				return nil
+			}
+		}
+
+		if !discovery.suppressParseErrors || cfg == nil {
+			return errors.New(err)
```
denis256: Should be returned err?

internal/discovery/phase_worktree.go
yhakbar: We'll have the stack trace, and a repro if folks report the panic. There should be no valid way to hit this line anyways.

internal/filter/classifier.go
yhakbar: Let me defer this, please. I'll work to make it so that the compilation happens earlier and we won't have to worry about runtime regex compilation errors. I think now that we have the options updated it should be easier to make this kind of change.

internal/discovery/phase_parse.go
yhakbar: Updated it to just return err

internal/filter/candidacy.go
yhakbar: Just tossed this. It was only used in testing.

internal/discovery/phase_worktree.go
yhakbar: Updated it to just return an error.

denis256:
We can improve PR in subsequent updates

PR #5494, opened 2026-02-02
fix: Using `SyncWriter` to fix `TestTerragruntReportWithUnitTiming` flakes

yhakbar:
## Description

Resolves race conditions in `TestTerragruntReportWithUnitTiming`.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **Bug Fixes**
  * Eliminated data races during getter initialization and added .netrc support for HTTP/HTTPS retrievals.
  * Prevented concurrent-write races by introducing thread-safe writers for parallel operations.

* **Tests**
  * Added HTTP/HTTPS authentication tests (including .netrc) and expanded getter coverage.

* **Refactor**
  * Moved to explicit logger passing in file/manifest utilities, resulting in updated public constructor/method signatures.

### DIFF
--- a/internal/runner/run/download_source.go
+++ b/internal/runner/run/download_source.go
@@ -257,7 +257,7 @@ func readVersionFile(terraformSource *tf.Source) (string, error) {
 //
 // This creates a closure that returns a function so that we have access to the terragrunt configuration, which is
 // necessary for customizing the behavior of the file getter.
-func UpdateGetters(terragruntOptions *options.TerragruntOptions, cfg *runcfg.RunConfig) func(*getter.Client) error {
+func UpdateGetters(l log.Logger, terragruntOptions *options.TerragruntOptions, cfg *runcfg.RunConfig) func(*getter.Client) error {
 	return func(client *getter.Client) error {
 		// We iterate over the global getter.Getters map and clone each getter
 		// to avoid race conditions. The global map contains shared getter
@@ -275,6 +275,7 @@ func UpdateGetters(terragruntOptions *options.TerragruntOptions, cfg *runcfg.Run
 
 		// Override with Terragrunt-specific customizations
 		client.Getters["file"] = &FileCopyGetter{
+			Logger:          l,
 			IncludeInCopy:   cfg.Terraform.IncludeInCopy,
 			ExcludeFromCopy: cfg.Terraform.ExcludeFromCopy,
 		}
@@ -376,7 +377,7 @@ func downloadSource(
 
 	// Fallback to standard go-getter
 	return opts.RunWithErrorHandling(ctx, l, r, func() error {
-		return getter.GetAny(src.DownloadDir, src.CanonicalSourceURL.String(), UpdateGetters(opts, cfg), preserveSymlinksOption())
+		return getter.GetAny(src.DownloadDir, src.CanonicalSourceURL.String(), UpdateGetters(l, opts, cfg), preserveSymlinksOption())
 	})
 }
 
--- a/internal/runner/run/download_source_test.go
+++ b/internal/runner/run/download_source_test.go
@@ -612,7 +612,7 @@ func TestUpdateGettersExcludeFromCopy(t *testing.T) {
 			client := &getter.Client{}
 
 			// Call updateGetters
-			updateGettersFunc := run.UpdateGetters(terragruntOptions, tc.cfg)
+			updateGettersFunc := run.UpdateGetters(logger.CreateLogger(), terragruntOptions, tc.cfg)
 			err = updateGettersFunc(client)
 			require.NoError(t, err)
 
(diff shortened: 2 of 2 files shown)


internal/runner/run/download_source.go
```diff
@@ -239,23 +239,21 @@ func readVersionFile(terraformSource *tf.Source) (string, error) {
...
+		client.Getters = map[string]getter.Getter{
+			"file": &FileCopyGetter{
+				IncludeInCopy:   cfg.Terraform.IncludeInCopy,
+				ExcludeFromCopy: cfg.Terraform.ExcludeFromCopy,
+			},
+			"git":   &getter.GitGetter{},
+			"gcs":   &getter.GCSGetter{},
+			"hg":    &getter.HgGetter{},
+			"s3":    &getter.S3Getter{},
+			"http":  &getter.HttpGetter{},
```
denis256: Possible authentication regression since it is not set Netrc field

yhakbar: Added `Netrc` back and added a regression test for this.

internal/runner/run/download_source_test.go
```diff
@@ -917,3 +948,53 @@ func TestDownloadSourceWithCASMultipleSources(t *testing.T) {
 		})
 	}
 }
+
+// TestHTTPGetterNetrcAuthentication verifies that HTTP/HTTPS getters correctly authenticate
+// using ~/.netrc credentials when downloading OpenTofu/Terraform sources.
+func TestHTTPGetterNetrcAuthentication(t *testing.T) {
+	expectedUser := "testuser"
```
denis256, line 987: No parallel option because of `~/.netrc` ?

internal/runner/run/download_source.go
```diff
@@ -239,23 +239,21 @@ func readVersionFile(terraformSource *tf.Source) (string, error) {
...
+		// We create fresh getter instances for each client to avoid race conditions.
+		// The global getter.Getters map contains shared getter instances, and when
+		// SetClient is called on them from multiple goroutines, it causes data races.
+		// Creating new instances ensures each client has its own getter state.
+		client.Getters = map[string]getter.Getter{
+			"file": &FileCopyGetter{
+				IncludeInCopy:   cfg.Terraform.IncludeInCopy,
+				ExcludeFromCopy: cfg.Terraform.ExcludeFromCopy,
+			},
+			"git":   &getter.GitGetter{},
```
denis256: Not getting why we need this change, to hardcode all getter, why we can't  iterate and create new instances

internal/runner/run/download_source_test.go
yhakbar: Ya, I can add a comment documenting that.

internal/runner/run/download_source.go
yhakbar: We can, it's just going to involve using reflection, which I didn't think was really worth the runtime overhead, given how unlikely these defaults are to change.

Updated the implementation to iterate and create copies. Open to reverting if we'd prefer the hard-coded approach, or implementing a third strategy.

internal/runner/run/download_source.go
```diff
@@ -256,26 +257,30 @@ func readVersionFile(terraformSource *tf.Source) (string, error) {
...
-			}
+		// We iterate over the global getter.Getters map and clone each getter
+		// to avoid race conditions. The global map contains shared getter
+		// instances, and when SetClient is called on them from multiple
+		// goroutines, it causes data races. Cloning via dereference ensures
+		// each client has its own getter state, while automatically picking
+		// up any new getter types registered by go-getter.
+		client.Getters = make(map[string]getter.Getter, len(getter.Getters))
+		for name, g := range getter.Getters {
+			v := reflect.ValueOf(g).Elem()
```
denis256, line 270: May lead to NPE if reflect.ValueOf(g) is nil?

internal/util/file.go
```diff
@@ -763,8 +762,8 @@ func (manifest *fileManifest) Close() error {
 	return manifest.fileHandle.Close()
 }
 
-func NewFileManifest(logger log.Logger, manifestFolder string, manifestFile string) *fileManifest {
-	return &fileManifest{logger: logger, ManifestFolder: manifestFolder, ManifestFile: manifestFile}
+func NewFileManifest(manifestFolder string, manifestFile string) *fileManifest {
```
denis256, line 765: Public constructor returns unexported *fileManifest - something is not right, maybe make function pivate or returned type public 🤔

internal/runner/run/download_source.go
yhakbar: I don't think `ValueOf` returns a pointer:
https://pkg.go.dev/reflect#ValueOf
https://pkg.go.dev/reflect#Value

internal/util/file.go
yhakbar: I think it's perfectly valid to have a public constructor for a private type if it doesn't make usage inconvenient. Ensures that consumers use the constructor instead of manually constructing it.

PR #5507, opened 2026-02-04
fix: Adding `--report-file` support to single runs

yhakbar:
## Description

I believe it was decided in collaboration with the community that runs shouldn't emit summaries when running indpenendently without `--all`, as it wasn't clear how valuable the summary would be to the user.

That didn't make the `--report-file` flag less useful for single runs, and it is a little confusing to have a `--report-file` flag that's supported but silently ignored when `--all` isn't passed.

This adds support for generating reports for individual runs while avoiding introducing summaries for runs that aren't `--all`.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

### DIFF
--- a/internal/cli/commands/run/run.go
+++ b/internal/cli/commands/run/run.go
@@ -2,6 +2,7 @@ package run
 
 import (
 	"context"
+	"path/filepath"
 	"strings"
 
 	"github.com/gruntwork-io/terragrunt/internal/errors"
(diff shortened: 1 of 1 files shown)


internal/cli/commands/run/run.go
```diff
@@ -78,7 +101,46 @@ func Run(ctx context.Context, l log.Logger, opts *options.TerragruntOptions) err
 
 	runCfg := cfg.ToRunConfig(l)
 
-	return run.Run(ctx, l, tgOpts, r, runCfg, credsGetter)
+	absWorkingDir, err := filepath.Abs(opts.RootWorkingDir)
```
denis256, line 104: May be a problem if we invoke from a different directory

PR #5522, opened 2026-02-05
fix: SOPS decryption env variables locking

### LINKED ISSUE #5515: Terragrunt would fail to decrypt secrets correctly via sops if multiple resources are updated in the same PR
## Describe the bug

Issue with Terragrunt pipelines. it appear that Terragrunt cannot decrypt a secrets encrypted with sops to the correct value if multiple resources are updated in the same PR.
If you raise a PR with only a change done to a secret unit the pipeline will decrypt the correct secret value 
When working with secrets in terragrunt it appear that sops secret decry

## Steps To Reproduce

Have a repo with at least one secret that was encrypted via sops. See example:

```
.
├── repro-unit-secret
│   └── terragrunt.hcl   #add empty line
│   └── secret.enc.json

```
`secret_json = try(jsondecode(sops_decrypt_file("secret.enc.json")), {})`

Add an empty line to the file, and run a Terragrunt plan locally, this should show a plan with No changes detected.
Push this commit to remote and raise a PR, the Terragrunt pipeline will detect the change to the secret resource and will initiate the plan, the pipeline will display No changes as well for this object same as the local Terragrunt plan.

Now go back to the source code and instead of only adding an empty line to a single secret, add one empty line to the secret and also a different the object, see example:

```
.
├── repro-unit-secret
│   └── terragrunt.hcl   #add empty line
│   └── secret.enc.json
└── another-repro-unit
    └── terragrunt.hcl   #add empty line

```

Do a Terragrunt plan and you will see still No changes detected.
Commit and push to remote, the the Terragrunt plan pipeline will run again.
This time the pipeline will detect a change happening for the secret

Image

I tested letting the pipeline apply to see what value would set and it would wipe the secret information setting an empty json object for the secret {}

## Versions

- Terragrunt version: 0.99.1
- OpenTofu/Terraform version: Terraform 1.5.7
- Latest Gruntwork pipelines

denis256:
## Description

* Fixed a race condition when setting environment variables used to decode SOPS secrets.
* Added tests to catch and prevent regressions for this SOPS decoding race condition.

Fixes #5515.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a concurrency race in SOPS decryption that could cause data corruption or parallel failures by serializing decryption and stabilizing environment handling.

* **Improvements**
  * Credentials are now available during config/HCL evaluation for auth-dependent setups.
  * Normalized file-path error reporting, cache logging, and additional debug logs for decryption behavior.

* **Tests**
  * Added regression and integration tests covering concurrent SOPS decryptions and multi-unit parallel runs.

### DIFF
--- a/pkg/config/config_helpers.go
+++ b/pkg/config/config_helpers.go
@@ -1362,26 +1362,28 @@ func sopsDecryptFileImpl(ctx context.Context, pctx *ParsingContext, l log.Logger
 	locks.EnvLock.Lock()
 	defer locks.EnvLock.Unlock()
 
-	// Override process env with opts.Env for SOPS auth, restoring originals on exit.
-	// Only touches keys that actually differ (typically 3-5 auth-provider keys),
-	// skipping the hundreds of inherited process env vars that already match.
+	// Set env vars from opts.Env that are missing from process env.
+	// Auth-provider credentials (e.g., AWS_SESSION_TOKEN) may not exist
+	// in process env yet — SOPS needs them for KMS auth.
+	// Existing process env vars are preserved to avoid overriding real
+	// credentials with empty auth-provider values.
 	env := pctx.TerragruntOptions.Env
-	origEnv := make(map[string]string)
+
+	var setKeys []string
 
 	for k, v := range env {
-		current := os.Getenv(k)
-
-		if current == v {
+		if _, exists := os.LookupEnv(k); exists {
 			continue
 		}
 
-		origEnv[k] = current
 		os.Setenv(k, v) //nolint:errcheck
+
+		setKeys = append(setKeys, k)
 	}
 
 	defer func() {
-		for k, v := range origEnv {
-			os.Setenv(k, v) //nolint:errcheck
+		for _, k := range setKeys {
+			os.Unsetenv(k) //nolint:errcheck
 		}
 	}()
 
--- a/pkg/config/sops_test.go
+++ b/pkg/config/sops_test.go
@@ -4,6 +4,7 @@ package config //nolint:testpackage // needs access to sopsCache internals
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -182,8 +183,8 @@ func TestSOPSDecryptConcurrencyRace(t *testing.T) { //nolint:paralleltest // mut
 // loaded into opts.Env. This caused SOPS to return empty/wrong secrets.
 //
 // This test verifies the env propagation contract of sopsDecryptFileImpl:
-//   - Fresh credentials from opts.Env override stale process env during decrypt
-//   - Process env is restored to original values after decrypt completes
+//   - Existing process env vars are preserved (not overridden by opts.Env)
+//   - Missing env vars from opts.Env are set during decrypt and unset after
 //   - Without credentials, decrypt fails (reproduces the original bug)
 //   - Concurrent goroutines with different credentials are properly isolated
 func TestSOPSDecryptEnvPropagation(t *testing.T) { //nolint:paralleltest // mutates package-global sopsCache and process env vars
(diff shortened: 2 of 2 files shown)


pkg/config/config_helpers.go
```diff
@@ -1326,60 +1344,59 @@ func sopsDecryptFile(ctx context.Context, pctx *ParsingContext, l log.Logger, pa
...
-	if !filepath.IsAbs(path) {
-		path = filepath.Join(pctx.TerragruntOptions.WorkingDir, path)
-		path = filepath.Clean(path)
-	}
+	// Set env vars from opts.Env that are missing from process env.
+	// Auth-provider credentials (e.g., AWS_SESSION_TOKEN) may not exist
+	// in process env yet — SOPS needs them for KMS auth.
+	// Existing process env vars are preserved to avoid overriding real
+	// credentials with empty auth-provider values.
+	env := pctx.TerragruntOptions.Env
```
yhakbar, line 1370: NIT: There's a TOCTOU bug here. We should check the cache again _after_ we get the write lock because another cache miss might have populated the cache when racing for the lock. I don't think it's a big deal to decrypt the file twice, though.

pkg/config/config_helpers.go
```diff
@@ -1326,60 +1344,59 @@ func sopsDecryptFile(ctx context.Context, pctx *ParsingContext, l log.Logger, pa
...
+		setKeys = append(setKeys, k)
 	}
 
+	defer func() {
+		for _, k := range setKeys {
+			os.Unsetenv(k) //nolint:errcheck
+		}
+	}()
+
+	// Double-check: another goroutine may have populated cache while we waited for the lock.
```
yhakbar, line 1390: Oh, why are we doing this after mucking with the env vars?

pkg/config/sops_test.go
```diff
@@ -0,0 +1,374 @@
...
+// made by locked goroutines (set via os.Setenv, then deferred os.Unsetenv).
+// With KMS-based decryption, the network latency makes this race window large
+// enough to hit reliably.
+//
+// This test injects a delay into decryptFn to simulate KMS latency,
+// then detects when env vars disappear mid-operation.
+//
+// Without the fix (conditional lock): FAILS — env vars change mid-decrypt.
+// With the fix (always lock):         PASSES — operations are serialized.
+func TestSOPSDecryptConcurrencyRace(t *testing.T) { //nolint:paralleltest // mutates package-global sopsCache and process env vars via os.Setenv/os.Unsetenv
```
yhakbar, line 60: Should we move `sopsCache` to a `ctx` cache like we do with the other caches?

pkg/config/sops_test.go
```diff
@@ -0,0 +1,374 @@
...
+
+	// Mock decrypt function that simulates KMS latency and detects env var races.
+	mockDecryptFn := func(path string, format string) ([]byte, error) {
+		// Check if WE are the goroutine that set the env var.
+		// The production code does os.Setenv BEFORE calling decryptFn,
+		// so if the env var is set here, we're the "setter" goroutine.
+		if os.Getenv(testEnvKey) != "" {
+			// Setter goroutine: short delay so our deferred os.Unsetenv
+			// runs quickly, while unlocked goroutines are still mid-operation.
+			time.Sleep(5 * time.Millisecond)
```
yhakbar, line 83: This test feels like it's going to flake. Wouldn't it be better to design this test to fail if there's a race condition, then suffix the test to `WithRacing` so that it's automatically run as `-race` in CI?

pkg/config/sops_test.go
```diff
@@ -0,0 +1,374 @@
...
+
+				// Half goroutines set env var via opts.Env (like auth-provider).
+				// In buggy code only these acquire the lock.
+				// The other half run unlocked — that's the race.
+				if idx%2 == 0 {
+					opts.Env = map[string]string{testEnvKey: "valid-token"}
+				}
+
+				l := logger.CreateLogger()
+				ctx := context.Background()
```
yhakbar, line 153: NIT: Prefer `t.Context()`

yhakbar (approved):
I think this is likely better than what we have right now, but we can probably do better. I think the testing setup can be improved in a future iteration and we can probably avoid the environment variable manipulation when there's a cache hit after acquiring the write lock.

PR #4078, opened 2025-03-24
feat: stack validation improvements

denis256:
## Description

* Added test check for not existing path for stack and unit
* Removed unused constant for values file
* Added validation checks on generated units and stacks to contain terragrunt files
* Added hidden cli flag to disable validation check
* Improved error message in case of failed unit fetching
* Moved worker pool struct as internal 
* Documentation update

out

error

validation

Demo:

stack-validation-v2

RFC: https://github.com/gruntwork-io/terragrunt/issues/3313

## TODOs

Read the [Gruntwork contribution guidelines](https://gruntwork.notion.site/Gruntwork-Coding-Methodology-02fdcd6e4b004e818553684760bf691e).

- [x] Update the docs.
- [x] Run the relevant tests successfully, including pre-commit checks.
- [x] Ensure any 3rd party code adheres with our [license policy](https://www.notion.so/gruntwork/Gruntwork-licenses-and-open-source-usage-policy-f7dece1f780341c7b69c1763f22b1378) or delete this line if its not applicable.
- [x] Include release notes. If this PR is backward incompatible, include a migration guide.

## Release Notes (draft)

Added / Removed / Updated [X].

## Summary by CodeRabbit

- **New Features**
  - Enhanced validation during stack generation now verifies that each unit and stack contains the appropriate configuration files.
  - Advanced users can bypass this validation using a new hidden command-line flag (`--no-stack-validate`).
  - New units and stacks have been added to support various configurations.

- **Documentation**
  - Updated guides and CLI option references to include details about the new validation step and the additional flag.

- **Refactor**
  - Improved error messages provide clearer feedback when stack generation encounters issues.
  - Consolidated flag management for better structure in the command-line interface.

PR #5526, opened 2026-02-07
fix: duplicate error reporting in worker pool

### LINKED ISSUE #5525: Duplicate errors in commands that use the worker pool
## Description

The worker pool in `internal/worker/worker.go` appends every error to `allErrors` twice, causing all error output to be duplicated. Any command that uses the worker pool (e.g. `terragrunt stack generate`) reports 2N errors instead of N, with each error message appearing identically twice. I've sent a PR #5526 to fix this. The PR contains more implementation details.

## Steps To Reproduce

1. Create a `terragrunt.stack.hcl` with a unit pointing to a nonexistent path:

```hcl
unit "broken" {
  source = "/nonexistent/path"
  path   = "broken"
}
```

2. Run:

```bash
terragrunt stack generate
```

3. Observe the output:

```
ERROR  2 errors occurred:

* Failed to fetch unit broken
  Original error: stat /nonexistent/path: no such file or directory

* Failed to fetch unit broken
  Original error: stat /nonexistent/path: no such file or directory
```

The same error is listed twice. With multiple failing units, every error is doubled.

## Expected behavior

A single failing unit should produce a single error:

```
ERROR  error occurred:

* Failed to fetch unit broken
  Original error: stat /nonexistent/path: no such file or directory
```

## Must haves

## Nice to haves

- [x] Terminal output

## Versions

- Terragrunt version: v0.97.2 (bug introduced in PR #4078, commit `9f9929a1`)
- OpenTofu/Terraform version: N/A (bug is in terragrunt internals, not provider-dependent)
- Environment details: macOS 15.7.3, arm64, Go 1.25.5 (reproducible on all platforms)

## Additional context

The bug was introduced in PR #4078 ("stack validation improvements"). `Submit()` appends errors to `allErrors` via a direct `appendError()` call, and then also sends the error through `resultChan` to `collectResults()`, which appends it to `allErrors` a second time. In the normal case both paths fire, doubling every error.

anuragrao04:
Fixes #5525 

## Problem

Every error from the worker pool (e.g. during `terragrunt stack generate`) is reported twice. N errors become 2N in the output.

## Root cause

`Submit()` in `internal/worker/worker.go` appends each error to `allErrors` via two paths that both fire in the normal case:

1. Direct `wp.appendError(err)` call
2. Send to `resultChan` -> `collectResults()` -> `wp.allErrors.Append(err)`

The likely intent (added in PR #4078) was for `appendError()` to return an error, so the second `if err != nil` block would only fire as a fallback when the direct append failed:

```go
		// If there's an error, always record it directly first
		if err != nil {
			err = wp.appendError(err) 
			// expected to return an error and the second err != nil check was
			// supposed to check this error. But appendError() doesn't return an error
			// and the second err != nil ended up checking the original err = task() error
		}

		// Only try to send result to channel if there's an error and pool isn't stopping
		if err != nil {
			select {
			case <-wp.doneChan:
				// Pool is stopping, error already recorded directly via appendError
			case wp.resultChan <- err:
				// Successfully sent the error
			default: // Channel might be closed or full, but we already recorded the error via appendError, so we can safely continue without panic
			}
		}
	}()

```

But `appendError()` has no return value, so `err` remains the original task error and the second block always runs — both paths fire on every error.

## Fix

Removed the `resultChan` send from `Submit()`, keeping only the direct `appendError()` call. The direct call runs inside the task goroutine before `wg.Done()`, so it guarantees the error is recorded before `Wait()` returns.

## How To Reproduce

```hcl
# terragrunt.stack.hcl
unit "broken" {
  source = "/nonexistent/path"
  path   = "broken"
}
```

```
$ terragrunt stack generate
# Before fix: "2 errors occurred" with the same error listed twice
# After fix: "1 error occurred"
# If it doesn't happen the first time, run multiple times. It
# is a race condition
```

## Testing

- Added error count assertion to `TestSomeTasksReturnErrors` — previously it only checked `require.Error(t, errs)` without verifying the count, which is why the bug wasn't caught by CI
- All worker pool tests pass with `-race`
- Downstream stack config tests pass
- `go vet` clean
- Manual verification against reproduction case

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal error handling to centralize error accumulation and remove the previous mixed propagation paths, reducing complexity and improving reliability.

* **Tests**
  * Strengthened tests to assert the aggregated error type and exact wrapped error count, improving detection of multi-error scenarios and regression protection.

---

> [!NOTE]
> **Low Risk**
> Small, localized concurrency simplification in `internal/worker` that removes unused channels and makes error aggregation deterministic; minimal behavioral change beyond eliminating duplicate errors.
> 
> **Overview**
> Fixes worker pool error reporting so each task failure is aggregated **once** by removing the `resultChan`/`doneChan` plumbing and `collectResults()` goroutine; `Submit()` now only records errors via the thread-safe `appendError()` path.
> 
> Simplifies shutdown (`Start`/`Stop`/`GracefulStop`) by no longer recreating/closing the removed channels, and tightens tests to assert `Wait()` returns an `*errors.MultiError` with the expected wrapped error count (preventing regressions of the 2x error bug).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 49a98596e676ff1a0eb97856d51dcbbc536ddf88. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/internal/worker/worker.go
+++ b/internal/worker/worker.go
@@ -32,14 +32,10 @@ type Task func() error
 // Pool manages concurrent task execution with a configurable number of workers
 type Pool struct {
 	semaphore   chan struct{}
-	resultChan  chan error
-	doneChan    chan struct{}
 	allErrors   *errors.MultiError
-	errorsSlice []error
 	wg          sync.WaitGroup
 	maxWorkers  int
 	mu          sync.RWMutex
-	resultMu    sync.RWMutex
 	allErrorsMu sync.RWMutex
 	isStopping  atomic.Bool
 	isRunning   bool
@@ -52,13 +48,10 @@ func NewWorkerPool(maxWorkers int) *Pool {
 	}
 
 	return &Pool{
-		maxWorkers:  maxWorkers,
-		semaphore:   make(chan struct{}, maxWorkers),
-		resultChan:  make(chan error),
-		doneChan:    make(chan struct{}),
-		isRunning:   false,
-		errorsSlice: make([]error, 0),
-		allErrors:   &errors.MultiError{},
+		maxWorkers: maxWorkers,
+		semaphore:  make(chan struct{}, maxWorkers),
+		isRunning:  false,
+		allErrors:  &errors.MultiError{},
 	}
 }
 
@@ -74,13 +67,7 @@ func (wp *Pool) Start() {
 	wp.isRunning = true
 	wp.isStopping.Store(false)
 
-	// Recreate the channels if they've been closed
-	wp.resultChan = make(chan error)
-	wp.doneChan = make(chan struct{})
 	wp.semaphore = make(chan struct{}, wp.maxWorkers)
-
-	// Clear previous errors
-	wp.errorsSlice = make([]error, 0)
 
 	// Reset allErrors
 	wp.allErrorsMu.Lock()
@@ -88,35 +75,6 @@ func (wp *Pool) Start() {
 	wp.allErrorsMu.Unlock()
 
 	wp.mu.Unlock()
-
-	// Start the error collector
-	go wp.collectResults()
-}
-
-// collectResults collects the errors from the result channel
-func (wp *Pool) collectResults() {
-	for {
-		select {
-		case err, ok := <-wp.resultChan:
-			if !ok {
-				return
-			}
-
-			if err != nil {
-				// Add to allErrors safely
-				wp.allErrorsMu.Lock()
-				wp.allErrors = wp.allErrors.Append(err)
-				wp.allErrorsMu.Unlock()
-
-				// Also keep the slice for backward compatibility
-				wp.resultMu.Lock()
-				wp.errorsSlice = append(wp.errorsSlice, err)
-				wp.resultMu.Unlock()
-			}
-		case <-wp.doneChan:
-			return
-		}
-	}
 }
 
 // appendError safely appends an error to allErrors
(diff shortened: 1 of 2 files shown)


internal/worker/worker.go
```diff
@@ -157,21 +157,9 @@ func (wp *Pool) Submit(task Task) {
...
 		if err != nil {
 			wp.appendError(err)
 		}
-
-		// Only try to send result to channel if there's an error and pool isn't stopping
-		if err != nil {
-			select {
-			case <-wp.doneChan:
-				// Pool is stopping, error already recorded directly via appendError
-			case wp.resultChan <- err:
```
denis256, line 170: `resultChan` now is unused

denis256: Looks like if we remove resultChan path, the entire collectResults/channel architecture becomes unnecessary

denis256 (commented):
PR introduces a lot of dead code

denis256:
```
level=info msg="File cache stats: 2 entries of total size 9.4KiB"
internal/worker/worker_test.go:86:18: type assertion on error will fail on wrapped errors. Use errors.As to check for specific errors (errorlint)
	multiErr, ok := errs.(*errors.MultiError)
	                ^
internal/worker/worker.go:159:1: unnecessary whitespace (err) (wsl_v5)

^
2 issues:
* errorlint: 1
```

PR #5538, opened 2026-02-11
fix: Fixing partial parse config cache

### LINKED ISSUE #4123: Get dependency output from module outside of working directory
Hello,

I want to upgrade Terragrunt from `0.69.0` to `0.76.6` but I have an issue when running `terragrunt run-all plan` on my units. 

I have 3 different environments with the same infrastructure (dev, staging, prod). 
```
.
├── cloud_development
├── cloud_production
├── cloud_staging
├── _envcommon
└── terragrunt.hcl
```

One of the units in the staging environment uses the same DB cluster as dev and therefore has a dependency to the other environment:

```
# file: cloud_staging/eu-west-1/staging/bs/infra/data_lake_private_link/terragrunt.hcl
...
dependency "bs_infra_data_lake" {
  config_path = "${get_repo_root()}/cloud_development/eu-west-1/dev/bs/infra/data_lake"
}
...
```

With Terragrunt `0.69.0`, the plan and apply work just fine (the `data_lake` module from dev is not displayed in the stack to process):

Image

With Terragrunt `0.76.6`. I get an attribute error when accessing the outputs of `bs_infra_data_lake`:

Image

I tried adding the flag `TG_QUEUE_INCLUDE_EXTERNAL=true` which resolves the issue but adds the `data_lake` from dev to the stack to process:

Image

and thus would also apply changes in the dev environment (while running the staging deployment). I want to avoid this but only read the outputs of the module in dev.

Is there a way to accomplish that in newer versions of Terragrunt?

P.S. I have enabled the following CLI flags:
```
TERRAGRUNT_PROVIDER_CACHE=true
TERRAGRUNT_WORKING_DIR=cloud_staging
TERRAGRUNT_FETCH_DEPENDENCY_OUTPUT_FROM_STATE=true
TERRAGRUNT_USE_PARTIAL_PARSE_CONFIG_CACHE=true
TERRAGRUNT_NON_INTERACTIVE=true
```

yhakbar:
## Description

Fixes #4123.

We should also think through stabilization of this functionality. I think we should aim to deprecate the flag post 1.0.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved configuration caching to correctly differentiate entries based on calling context, ensuring each module receives appropriate cached values.

* **Tests**
  * Added test to verify proper cache differentiation across multiple calling contexts.

### DIFF
--- a/pkg/config/config_partial.go
+++ b/pkg/config/config_partial.go
@@ -321,7 +321,7 @@ func PartialParseConfigFile(ctx context.Context, pctx *ParsingContext, l log.Log
 // filename, configString, includeFromChild and decodeList are used for the cache key,
 // by getting the default value (%#v) through fmt.
 func TerragruntConfigFromPartialConfig(ctx context.Context, pctx *ParsingContext, l log.Logger, file *hclparse.File, includeFromChild *IncludeConfig) (*TerragruntConfig, error) {
-	var cacheKey = fmt.Sprintf("%#v-%#v-%#v-%#v", file.ConfigPath, file.Content(), includeFromChild, pctx.PartialParseDecodeList)
+	var cacheKey = fmt.Sprintf("%#v-%#v-%#v-%#v-%#v", file.ConfigPath, file.Content(), includeFromChild, pctx.PartialParseDecodeList, pctx.TerragruntOptions.TerragruntConfigPath)
 
 	terragruntConfigCache := cache.ContextCache[*TerragruntConfig](ctx, TerragruntConfigCacheContextKey)
 	if pctx.TerragruntOptions.UsePartialParseConfigCache {
--- a/pkg/config/config_partial_test.go
+++ b/pkg/config/config_partial_test.go
@@ -414,7 +414,7 @@ func TestPartialParseSavesToHclCache(t *testing.T) {
 	// Setup test environment
 	tmpDir := helpers.TmpDirWOSymlinks(t)
 	configPath := filepath.Join(tmpDir, "terragrunt.hcl")
-	configContent := `dependencies { paths = ["../app1"] }`
+	configContent := `dependencies { paths = ["../app1"] }` //nolint:goconst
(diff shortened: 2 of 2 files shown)


pkg/config/config_partial.go
```diff
@@ -321,7 +321,7 @@ func PartialParseConfigFile(ctx context.Context, pctx *ParsingContext, l log.Log
 // filename, configString, includeFromChild and decodeList are used for the cache key,
```
denis256, line 321: Comment I think should be updated

pkg/config/config_partial.go
```diff
@@ -321,7 +321,7 @@ func PartialParseConfigFile(ctx context.Context, pctx *ParsingContext, l log.Log
...
 // by getting the default value (%#v) through fmt.
 func TerragruntConfigFromPartialConfig(ctx context.Context, pctx *ParsingContext, l log.Logger, file *hclparse.File, includeFromChild *IncludeConfig) (*TerragruntConfig, error) {
-	var cacheKey = fmt.Sprintf("%#v-%#v-%#v-%#v", file.ConfigPath, file.Content(), includeFromChild, pctx.PartialParseDecodeList)
+	var cacheKey = fmt.Sprintf("%#v-%#v-%#v-%#v-%#v", file.ConfigPath, file.Content(), includeFromChild, pctx.PartialParseDecodeList, pctx.TerragruntOptions.TerragruntConfigPath)
```
denis256, line 324: TerragruntConfigPath is a plain string. Using %#v wraps it in quotes ("path/to/file") which is unnecessary overhead. 🤔

yhakbar: I guess it helps making debugging easier, as we we'll be able to read the different components of the key without getting confused as to whether a key element is being separated by a dash or has one within the value.

PR #5550, opened 2026-02-13
chore: Units code cleanup

denis256:
## Description

* Removed unused function SetUnitsMap()

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **Refactor**
  * Cleaned up internal code structure by removing unused internal dependencies and simplifying queue initialization logic.

### DIFF
--- a/internal/component/stack.go
+++ b/internal/component/stack.go
@@ -176,9 +176,7 @@ func (s *Stack) ensureDependency(dependency Component) {
 	s.lock()
 	defer s.unlock()
 
-	if !slices.ContainsFunc(s.dependencies, func(d Component) bool {
-		return d.Path() == dependency.Path()
-	}) {
+	if !slices.Contains(s.dependencies, dependency) {
 		s.dependencies = append(s.dependencies, dependency)
 	}
 }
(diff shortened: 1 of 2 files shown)


internal/component/stack.go
```diff
@@ -176,7 +176,9 @@ func (s *Stack) ensureDependency(dependency Component) {
 	s.lock()
 	defer s.unlock()
 
-	if !slices.Contains(s.dependencies, dependency) {
+	if !slices.ContainsFunc(s.dependencies, func(d Component) bool {
```
yhakbar: Why do this? We can just compare the pointers, right?

denis256: during concurrent discovery - same dependency path may be resolved to different Unit instances by different goroutines (componentFromDependencyPath creates new instances).

This may lead to duplicate entries in dependencies slice, inflating traversal counts and potentially duplicate execution.

yhakbar: That's a bug in discovery, then. 

If we have two different units created with the same path, that's bad. There's supposed to be synchronization in place to ensure that we only ever create exactly one unit per path using an ensure in the thread safe store of components.

denis256: Looks like it is not working... a lot of tests fail

PR #5551, opened 2026-02-13
chore: Removing options from components

yhakbar:
## Description

Improves data flow consistency by ensuring that objects like `opts`, `l` and `r` are always explicitly passed instead of embedded in structs.

Helps to avoid import cycles and makes it easier to reason about which version of an object is being used.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **Refactor**
  * Decoupled execution state from core components, passing per-unit options and loggers explicitly
  * Reworked runner orchestration to use a dedicated StackRunner abstraction and on-demand per-unit option construction
  * Centralized report handling at runner orchestration level instead of discovery/options wiring
  * Propagated explicit logger and output writer usage across execution and flush paths, simplifying internal coupling

### DIFF
--- a/internal/runner/runnerpool/runner.go
+++ b/internal/runner/runnerpool/runner.go
@@ -250,7 +250,11 @@ func NewRunnerPoolStack(
 	// Resolve units from discovery
 	units := make([]*component.Unit, 0, len(nonStackComponents))
 	for _, c := range nonStackComponents {
-		unit := c.(*component.Unit)
+		unit, ok := c.(*component.Unit)
+		if !ok {
+			continue
+		}
+
 		if unit.DiscoveryContext() != nil && unit.Config() == nil {
 			l.Debugf("Unit %s has no config from discovery", unit.DisplayPath())
 		}
@@ -282,7 +286,7 @@ func NewRunnerPoolStack(
 
 	rnr.queue = q
 
-	return rnr.WithOptions(runnerOpts...), nil
+	return rnr, nil
 }
 
 // filterUnitsToComponents converts resolved units to Components.


internal/runner/runnerpool/runner.go
```diff
@@ -264,42 +231,44 @@ func NewRunnerPoolStack(
...
+	rnr = rnr.WithOptions(runnerOpts...)
 
-	// Resolve units from discovery - populates Execution fields on each unit
-	units, err := resolveUnitsFromDiscovery(ctx, l, runner.Stack, nonStackComponents)
-	if err != nil {
-		return nil, err
+	// Resolve units from discovery
+	units := make([]*component.Unit, 0, len(nonStackComponents))
+	for _, c := range nonStackComponents {
+		unit := c.(*component.Unit)
```
denis256: Unsafe type cast?

internal/runner/runnerpool/runner.go
```diff
@@ -311,19 +280,19 @@ func NewRunnerPoolStack(
 		return nil, queueErr
 	}
 
-	runner.queue = q
+	rnr.queue = q
 
-	return runner.WithOptions(opts...), nil
+	return rnr.WithOptions(runnerOpts...), nil
```
denis256: runnerOpts applied twice - line 285 and line 248

internal/runner/runnerpool/runner.go
```diff
@@ -112,93 +74,101 @@ func CloneUnitOptions(
...
+func syncUnitCliArgs(l log.Logger, stackOpts *options.TerragruntOptions, unitOpts *options.TerragruntOptions, unit *component.Unit) {
+	discoveryCtx := unit.DiscoveryContext()
+	if discoveryCtx != nil && len(discoveryCtx.Args) > 0 {
+		// Merge stack-level flags that aren't already present
+		unitOpts.TerraformCliArgs.MergeFlags(stackOpts.TerraformCliArgs)
+	} else {
+		unitOpts.TerraformCliArgs = stackOpts.TerraformCliArgs.Clone()
+	}
+
+	planFile := unit.PlanFile(stackOpts.RootWorkingDir, stackOpts.OutputFolder, unitOpts.JSONOutputFolder, unitOpts.TerraformCommand)
```
denis256, line 153: stackOpts or unitOpts?

denis256 (approved):

internal/runner/runnerpool/runner.go
yhakbar: These should be stack opts, as we're aggregating everything at the discovery context working dir.

internal/runner/runnerpool/runner.go
yhakbar: Good catch. Will address.

internal/runner/runnerpool/runner.go
yhakbar: Thanks. Good catch.

PR #5559, opened 2026-02-17
chore: refactor cache dir exclusion code

healthy-pod:
## Description

In https://github.com/gruntwork-io/terragrunt/pull/4533, we skipped recursing through cache directories completely by returning
filepath.SkipDir when hitting one. Previously, we just returned nil which meant we
still had to walk everything in the file tree.

This code change applies the same to `WalkDirWithSymlinks` and refactors the logic
into a single walkDir function. It piggybacks on `util.SkipDirIfIgnorable` which was
migrated from `internal/discovery` and also covers terraform cache dir and git dir.

## TODOs

- [X] I am submitting code based on open source software (e.g. MIT, MPL-2.0, Apache)]
- [ ] Update the docs.

## Release Notes (draft)

Updated EncodeSourceVersion to make less stat calls and skip walking cache dirs.

## Summary by CodeRabbit

* **Refactor**
  * Unified directory exclusion and traversal behavior across discovery and source encoding so ignorable folders (e.g., .git, .terraform, .terragrunt-cache) are handled consistently.
  * Consolidated file-walking and hashing entry points for more consistent source hashing and error handling during local-source processing.

### DIFF
--- a/internal/tf/source.go
+++ b/internal/tf/source.go
@@ -62,72 +62,38 @@ func (src Source) EncodeSourceVersion(l log.Logger) (string, error) {
 
 		var err error
 
+		walkDir := filepath.WalkDir
 		if src.WalkDirWithSymlinks {
-			err = util.WalkDirWithSymlinks(sourceDir, func(path string, d fs.DirEntry, err error) error {
-				if err != nil {
-					// If we've encountered an error while walking the tree, give up
-					return err
-				}
-
-				if d.IsDir() {
-					// We don't use any info from directories to calculate our hash
-					return nil
-				}
-				// avoid checking files in .terragrunt-cache directory since contents is auto-generated
-				if strings.Contains(path, util.TerragruntCacheDir) {
-					return nil
-				}
-				// avoid checking files in .terraform directory since contents is auto-generated
-				if d.Name() == util.TerraformLockFile {
-					return nil
-				}
-
-				info, err := d.Info()
-				if err != nil {
-					return err
-				}
-
-				fileModified := info.ModTime().UnixMicro()
-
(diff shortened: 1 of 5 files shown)


internal/tf/source.go
```diff
@@ -62,70 +62,38 @@ func (src Source) EncodeSourceVersion(l log.Logger) (string, error) {
...
+
+			fileModified := info.ModTime().UnixMicro()
+
+			hashContents := fmt.Sprintf("%s:%d", path, fileModified)
+			sourceHash.Write([]byte(hashContents))
+
+			return nil
+		}
+
+		if src.WalkDirWithSymlinks {
```
yhakbar: NIT: We usually do the opposite here, where we put one of `filepath.WalkDir` or `util.WalkDirWithSymlinks` in the variable, then have the main body be the actual logic of the walk. Helps make the more interesting part happen closer to the return.

internal/util/file.go
```diff
@@ -1208,3 +1209,16 @@ func MoveFile(source string, destination string) error {
...
 	return nil
 }
+
+// SkipDirIfIgnorable checks if an entire directory should be skipped based on the fact that it's
+// in a directory that should never have components discovered in it.
+func SkipDirIfIgnorable(path string) error {
+	base := filepath.Base(path)
+
+	switch base {
+	case ".git", TerraformCacheDir, TerragruntCacheDir:
```
yhakbar: NIT: Why doesn't `.git` get a constant?

internal/tf/source.go
healthy-pod: Done

internal/util/file.go
healthy-pod: Done

PR #5560, opened 2026-02-17
chore: Reducing dependency on `opts` in `config`

yhakbar:
## Description

This PR begins decoupling `pkg/config` from `pkg/options` by reducing direct usage of TerragruntOptions struct throughout the config parsing pipeline. The work spans several areas, and there's still significant work to do.

The following:
- `internal/shell` 
- `internal/tf`
- `internal/remotestate`
- credential provider packages (amazonsts, externalcmd, creds) 

now accept focused `RunOptions` structs or individual parameters instead of the full `TerragruntOptions` struct. Each package defines its own RunOptions containing only the fields it needs, along with a `RunOptionsFromOpts` constructor for callers that still hold a full options struct. The translation logic will likely go away once more of the codebase is able to break its dependency on `TerragruntOptions`.

The `internal/awshelper` package was refactored to use a builder pattern (`AwsConfigBuilder`) that accepts `env` and `IAMRoleOptions` independently.

`IAMRoleOptions` and `MergeIAMRoleOptions` were moved from `pkg/options` to a new `internal/iam` package, breaking the circular dependency that would otherwise prevent config from using these types without importing `options`.

`ParsingContext` now carries all the fields that config code previously accessed through `pctx.TerragruntOptions.*` as direct struct fields (paths, OpenTofu command info, feature flags, IAM roles, I/O writers, behavior flags, etc.). A `populateFromOpts` method copies values from `TerragruntOptions` into these flat fields during construction, and a `WithConfigPath` method replaces the common `opts.CloneWithConfigPath` pattern. During the transition period, `TerragruntOptions` is still embedded and kept in sync, allowing incremental migration of downstream code.

All config package files (`config.go`, `config_helpers.go`, `config_partial.go`, etc.) were updated to read from flat
  `ParsingContext` fields instead of reaching through `TerragruntOptions`. Several public function signatures were narrowed: 
- `CopyLockFile`
- `DefaultParserOptions`
- `ParseVariables`
- `GetTerraformSourceURL`
- `FindConfigFilesInPath`
- `ParseAndDecodeVarFile`
- `Merge`
- `DeepMerge`

now accept only the specific parameters they need rather than the full options struct.

Four AWS helper functions were consolidated into a single `getAWSField` helper with a callback parameter.

Unused parameters were removed from the HCL validation helpers.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

### DIFF
--- a/internal/awshelper/config_test.go
+++ b/internal/awshelper/config_test.go
@@ -21,7 +21,7 @@ func TestAwsSessionValidationFail(t *testing.T) {
 
 	l := logger.CreateLogger()
 
-	_, err := awshelper.NewAwsConfigBuilder().
+	_, err := awshelper.NewAWSConfigBuilder().
 		WithSessionConfig(&awshelper.AwsSessionConfig{
 			Region:        "not-existing-region",
 			CredsFilename: "/tmp/not-existing-file",
@@ -92,7 +92,7 @@ func TestCreateAwsConfigWithAuthProviderEnv(t *testing.T) {
 		"AWS_REGION":            "us-west-2",
 	}
 
-	cfg, err := awshelper.NewAwsConfigBuilder().
+	cfg, err := awshelper.NewAWSConfigBuilder().
 		WithEnv(env).
 		Build(ctx, l)
 	require.NoError(t, err)
@@ -113,7 +113,7 @@ func TestCreateAwsConfigWithAuthProviderEnvDefaultRegion(t *testing.T) {
 		"AWS_DEFAULT_REGION":    "eu-west-1",
 	}
 
-	cfg, err := awshelper.NewAwsConfigBuilder().
+	cfg, err := awshelper.NewAWSConfigBuilder().
 		WithEnv(env).
 		Build(ctx, l)
 	require.NoError(t, err)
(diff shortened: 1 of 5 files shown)


internal/awshelper/config_test.go
denis256, line 19: Skip reason is still valid?

internal/awshelper/config.go
```diff
@@ -11,12 +11,13 @@ import (
...
-	"github.com/aws/aws-sdk-go-v2/service/iam"
+	awsiam "github.com/aws/aws-sdk-go-v2/service/iam"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"github.com/aws/aws-sdk-go-v2/service/sts"
 	"github.com/aws/aws-sdk-go-v2/service/sts/types"
 	"github.com/gruntwork-io/go-commons/version"
 	"github.com/gruntwork-io/terragrunt/internal/errors"
+	"github.com/gruntwork-io/terragrunt/internal/iam"
 	"github.com/gruntwork-io/terragrunt/pkg/log"
 	"github.com/gruntwork-io/terragrunt/pkg/options"
```
denis256, line 22: Looks like import is still used

internal/runner/run/creds/providers/externalcmd/provider.go
```diff
@@ -21,30 +22,32 @@ import (
 // Provider runs external command that returns a json string with credentials.
 type Provider struct {
 	terragruntOptions *options.TerragruntOptions
```
denis256, line 24: options still used in struct

internal/awshelper/config.go
yhakbar: Ya, I'm trying to move through this incrementally instead of fully refactoring everything all at once.

internal/awshelper/config_test.go
yhakbar: Not sure, tbh.

PR #5574, opened 2026-02-19
chore: Compile filter globs earlier

yhakbar:
## Description

Makes it so that globs in filters are compiled when parsed.

We still need to make it so that the filters are parsed earlier, so that this isn't as important, but it's a good first step.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Release Notes

* **Refactor**
  * Simplified internal APIs by removing logging dependencies from classification and filtering operations
  * Improved error handling with eager glob pattern compilation at construction time instead of deferred evaluation
  * Streamlined matcher interface by removing error propagation in favor of boolean returns

* **Tests**
  * Updated test infrastructure with factory helper functions for cleaner expression construction

### DIFF
--- a/internal/worktrees/worktrees.go
+++ b/internal/worktrees/worktrees.go
@@ -231,48 +231,6 @@ func (w *Worktrees) Stacks() StackDiff {
 	return stackDiff
 }
 
-// expandDiffPaths processes a list of changed paths from a worktree diff, creating path filter expressions
-// for discovered units and stacks. primaryExprs receives filters for config files (units/stacks),
-// while fallbackExprs receives filters for non-config files adjacent to units in the "to" worktree.
-func expandDiffPaths(paths []string, toPath string, primaryExprs, fallbackExprs *filter.Expressions) error {
-	for _, path := range paths {
-		dir := filepath.Dir(path)
-
-		switch filepath.Base(path) {
-		case config.DefaultTerragruntConfigPath:
-			expr, err := filter.NewPathFilter(dir)
-			if err != nil {
-				return errors.Errorf("failed to create path filter for %s: %w", dir, err)
-			}
-
-			*primaryExprs = append(*primaryExprs, expr)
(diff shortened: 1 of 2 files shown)


internal/worktrees/worktrees.go
```diff
@@ -242,68 +242,49 @@ func (wp *WorktreePair) Expand() (filter.Filters, filter.Filters) {
...
-				filter.NewPathFilter(filepath.Join(dir, "**")),
-			)
-		default:
-			// Check to see if the added file is in the same directory as a unit in the to worktree.
-			// If so, we'll consider the unit modified.
-			if _, err := os.Stat(filepath.Join(toPath, dir, config.DefaultTerragruntConfigPath)); err == nil {
-				toExpressions = append(toExpressions, filter.NewPathFilter(dir))
-			}
-		}
+	if err := expandDiffPaths(diffs.Added, toPath, &toExpressions, &toExpressions); err != nil {
```
denis256, line 249:
```
 245:     if err := expandDiffPaths(diffs.Removed, toPath, &fromExpressions,
  &toExpressions); ...
  249:     if err := expandDiffPaths(diffs.Added,   toPath, &toExpressions,
  &toExpressions); ...
```
Same-pointer call looks suspicious

yhakbar: I agree it's confusing. Would accept feedback on how to make it less confusing.

What we're doing here is working out that something _removed_ can still result in needing to run `plan`/`apply` instead of `plan -destroy`/ `apply -destroy` in the "To" worktree. The reason for this is that content removed in the directory of a unit (without removing the unit itself) would require a plan/apply on the unit in the "To" worktree.

e.g.

User has a `main.tf`, a `extra.tf` and a `terragrunt.hcl` file in the same directory. They remove `extra.tf` in a commit. We should plan/apply the unit in the "To" worktree to run the update to the unit.

PR #5590, opened 2026-02-19
fix: Fixing `-detailed-exitcode` when used in combination with `--working-dir`

### LINKED ISSUE #5586: `run --all` with `plan -detailed-exitcode` exits 0 even when there are changes
## Describe the bug

`run --all` with `plan -detailed-exitcode` exits 0 even when there are changes. 

## Reproducing bugs

Setup the following directory structure:

```
❯ tree
.
└── foo
    └── bar
        ├── main.tf
        └── terragrunt.hcl
```

maint.tf:
```
❯ cat foo/bar/main.tf
output "foo" {
  value = "1"
}
```

terragrunt.hcl:
```
❯ cat foo/bar/terragrunt.hcl
terraform {
  source = "."
}
```

## Steps To Reproduce

Run:

```
❯ terragrunt run --all --working-dir foo -- plan -detailed-exitcode
11:08:29.787 INFO   Unit queue will be processed for plan in this order:
- Unit bar

11:08:29.861 STDOUT [bar] terraform: Changes to Outputs:
11:08:29.861 STDOUT [bar] terraform:   + foo = "1"
11:08:29.861 STDOUT [bar] terraform: You can apply this plan to save these new output values to the Terraform
11:08:29.861 STDOUT [bar] terraform: state, without changing any real infrastructure.
11:08:29.861 STDOUT [bar] terraform:
11:08:29.861 STDOUT [bar] terraform: ─────────────────────────────────────────────────────────────────────────────
11:08:29.861 STDOUT [bar] terraform: Note: You didn't use the -out option to save this plan, so Terraform can't
11:08:29.862 STDOUT [bar] terraform: guarantee to take exactly these actions if you run "terraform apply" now.

❯❯ Run Summary  1 units  84ms
   ────────────────────────────
   Succeeded    1
```

and see the exit code is 0:

```
❯ echo $?
0
```

If you call the unit directly, without `--all`, then the exit code is correct:

```
❯ terragrunt run --working-dir foo/bar -- plan -detailed-exitcode
11:45:59.737 STDOUT terraform: Changes to Outputs:
11:45:59.737 STDOUT terraform:   + foo = "1"
11:45:59.737 STDOUT terraform: You can apply this plan to save these new output values to the Terraform
11:45:59.738 STDOUT terraform: state, without changing any real infrastructure.
11:45:59.738 STDOUT terraform:
11:45:59.738 STDOUT terraform: ─────────────────────────────────────────────────────────────────────────────
11:45:59.738 STDOUT terraform: Note: You didn't use the -out option to save this plan, so Terraform can't
11:45:59.738 STDOUT terraform: guarantee to take exactly these actions if you run "terraform apply" now.
```

```
❯ echo $?
2
```

## Expected behavior

The exit code should be 2, or some other non-zero exit code indicating failure specific to terragrunt run all (since there may be multiple units).

## Versions

- Terragrunt version: v0.99.3
- OpenTofu/Terraform version: v1.14.5

## Additional context

Issue is a bit of a continuance or follow-on to other reports:

* https://github.com/gruntwork-io/terragrunt/issues/4898
* https://github.com/gruntwork-io/terragrunt/issues/4292

yhakbar:
## Description

Fixes #5586.

A bug in the cache key for the exit code caused issues when attempting to get the code when a unit used the `terraform.source` attribute.

We were using `filepath.Abs` when resolving the unit directory, which used the process absolute path instead of the value in `--working-dir`. This is an anti-pattern in the codebase we need to correct.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal path handling by removing unnecessary logging dependencies.

* **Tests**
  * Added integration test coverage for detailed exit codes when Terraform sources are present.

* **Chores**
  * Updated test fixtures with Terraform configurations.

### DIFF
--- a/internal/runner/common/unit_runner.go
+++ b/internal/runner/common/unit_runner.go
@@ -54,14 +54,14 @@ func (runner *UnitRunner) runTerragrunt(
 
 	defer func() {
 		// Flush buffered output for this unit, if the writer supports it.
-		if err := component.FlushOutput(l, runner.Unit, opts.Writer); err != nil {
+		if err := component.FlushOutput(runner.Unit, opts.Writer); err != nil {
 			l.Errorf("Error flushing output for unit %s: %v", runner.Unit.Path(), err)
 		}
 	}()
 
 	// Only create report entries if report is not nil
 	if r != nil {
-		unitPath := runner.Unit.AbsolutePath(l)
+		unitPath := runner.Unit.Path()
 		unitPath = util.CleanPath(unitPath)
 
 		// Pass the discovery context fields for worktree scenarios
@@ -93,14 +93,14 @@ func (runner *UnitRunner) runTerragrunt(
 
 	// Store the unit exit code in the global map using the unit path as key.
 	if globalExitCode != nil {
-		unitPath := filepath.Dir(opts.OriginalTerragruntConfigPath)
+		unitPath := runner.Unit.Path()
 		code := unitExitCode.Get(unitPath)
 		globalExitCode.Set(unitPath, code)
 	}
 
 	// End the run with appropriate result (only if report is not nil)
 	if r != nil {
-		unitPath := runner.Unit.AbsolutePath(l)
+		unitPath := runner.Unit.Path()
 		unitPath = util.CleanPath(unitPath)
 
 		if runErr != nil {
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -428,7 +428,11 @@ func TestDetailedExitCodeChangesPresentAllWithSource(t *testing.T) {
 	ctx := t.Context()
 	ctx = tf.ContextWithDetailedExitCode(ctx, exitCode)
 
-	_, _, err := helpers.RunTerragruntCommandWithOutputWithContext(t, ctx, "terragrunt run --all --non-interactive --working-dir "+rootPath+" -- plan -detailed-exitcode")
+	_, _, err := helpers.RunTerragruntCommandWithOutputWithContext(
+		t,
+		ctx,
+		"terragrunt run --all --non-interactive --working-dir "+rootPath+" -- plan -detailed-exitcode",
+	)
 	require.NoError(t, err)
 	assert.Equal(t, 2, exitCode.GetFinalDetailedExitCode())
 }
--- a/internal/component/unit.go
+++ b/internal/component/unit.go
@@ -10,7 +10,6 @@ import (
 	"github.com/gruntwork-io/terragrunt/internal/tf"
 	"github.com/gruntwork-io/terragrunt/internal/util"
 	"github.com/gruntwork-io/terragrunt/pkg/config"
-	"github.com/gruntwork-io/terragrunt/pkg/log"
 )
 
 const (
@@ -256,19 +255,6 @@ func (u *Unit) String() string {
 		"Unit %s (excluded: %v, dependencies: [%s])",
 		path, u.excluded, strings.Join(deps, ", "),
 	)
-}
-
-// AbsolutePath returns the absolute path of the unit.
-// If path conversion fails, logs the error and returns the original path.
-func (u *Unit) AbsolutePath(l log.Logger) string {
-	absPath, err := filepath.Abs(u.path)
-	if err != nil {
-		l.Errorf("Error converting unit path %s to absolute path: %v", u.path, err)
-
-		return u.path
-	}
-
-	return absPath
 }
 
 // DisplayPath returns the path relative to DiscoveryContext.WorkingDir for display purposes.
--- a/internal/component/unit_output.go
+++ b/internal/component/unit_output.go
@@ -4,8 +4,6 @@ import (
 	"fmt"
 	"io"
 	"sync"
-
-	"github.com/gruntwork-io/terragrunt/pkg/log"
 )
 
 // flusher is any writer that supports Flush() error.
@@ -40,7 +38,7 @@ func unitOutputLock(key string) *sync.Mutex {
 
 // FlushOutput flushes buffer data to the given writer for this unit, if the writer supports it.
 // This is safe to call even if u or w is nil.
-func FlushOutput(l log.Logger, u *Unit, w io.Writer) error {
+func FlushOutput(u *Unit, w io.Writer) error {
 	if u == nil || w == nil {
 		return nil
 	}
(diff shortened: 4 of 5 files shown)


internal/runner/common/unit_runner.go
```diff
@@ -91,10 +91,9 @@ func (runner *UnitRunner) runTerragrunt(
 
 	runErr := run.Run(ctx, l, opts, r, cfg, credsGetter)
 
-	// Store the unit exit code in the global map using the unit path as key
-	// (matches key used in run_cmd.go via filepath.Dir(opts.OriginalTerragruntConfigPath))
+	// Store the unit exit code in the global map using the unit path as key.
 	if globalExitCode != nil {
-		unitPath := filepath.Clean(runner.Unit.AbsolutePath(l))
+		unitPath := filepath.Dir(opts.OriginalTerragruntConfigPath)
```
lorengordon: Interesting. I'm curious how this resolved before, `filepath.Clean(runner.Unit.AbsolutePath(l))`?

If I comment out the `source` entirely in my reproduction case, just have an empty terragrunt.hcl file, then the exit code behavior is correct. Though the "Run Summary" seems wrong?

```
❯❯ Run Summary  1 units  131ms
   ────────────────────────────
   Succeeded    1

❯ echo $?
2
```

But if I set the `source` to the absolute path, then the behavior is actually still wrong, and reproduces the zero exit code... Which makes me think `filepath.Clean(runner.Unit.AbsolutePath(l))` doesn't resolve the path the same way I'd expect...

```
terraform {
  source = "/home/x/git/repos/test/foo/bar"
}
```

yhakbar: I'm still experimenting with this. Do you have your repo in a symlink'ed dir?

lorengordon: I checked every directory in the path of the reproduction case, and there are no symlinks. I am running on wsl2, though, but the directory is fully in the VM filesystem, not the windows mount.

lorengordon: Oh, and our CI/CD system is where we first noticed the issue. That's just running Ubuntu in a container. No windows/wsl interference.

yhakbar: Ah, I see what might be causing an issue. Are you using the `--working-dir` flag in CI?

lorengordon: Yes, indeed!

yhakbar: Root cause is this, then:
https://github.com/gruntwork-io/terragrunt/blob/main/internal/component/unit.go#L263

We're calling `filepath.Abs` there, which we should basically never call in the Terragrunt codebase. It gets the absolute path of a filepath based on the current working directory of the process. That's going to cause bugs when users use `--working-dir`, as the two paths don't align.

I'll look to remove that method, then we probably need to do a deeper audit to make sure we aren't using this anywhere else.

yhakbar: @lorengordon I pushed up an alpha release that has this fix ([alpha-2026022001](https://github.com/gruntwork-io/terragrunt/releases/tag/alpha-2026022001)). If you don't mind validating that the fix addresses the issue you're seeing in CI, I'd appreciate it.

lorengordon: seems to exit 2, propagating the detailed-exitcode result!

```
❯ ./terragrunt_linux_amd64 run --all --working-dir foo -- plan -detailed-exitcode
07:20:15.170 INFO   Unit queue will be processed for plan in this order:
- Unit bar

07:20:15.224 STDOUT [bar] terraform: Changes to Outputs:
07:20:15.224 STDOUT [bar] terraform:   + foo = "1"
07:20:15.224 STDOUT [bar] terraform: You can apply this plan to save these new output values to the Terraform
07:20:15.225 STDOUT [bar] terraform: state, without changing any real infrastructure.
07:20:15.225 STDOUT [bar] terraform:
07:20:15.225 STDOUT [bar] terraform: ─────────────────────────────────────────────────────────────────────────────
07:20:15.225 STDOUT [bar] terraform: Note: You didn't use the -out option to save this plan, so Terraform can't
07:20:15.225 STDOUT [bar] terraform: guarantee to take exactly these actions if you run "terraform apply" now.

❯❯ Run Summary  1 units  62ms
   ────────────────────────────
   Succeeded    1


❯ echo $?
2
```

lorengordon: That "Run Summary" is still saying the wrong thing though, yeah? "Succeeded 1"?

yhakbar: No, as far as Terragrunt is concerned, you had 1 unit that succeeded. We _could_ update this down the line to also include some hint there when users use `plan -detailed-exitcode` that some succeeded with a change.

lorengordon: Ahh. Interesting. I assumed the concept of "success" was based on the exit code. Guess I don't understand what terragrunt considers success!

yhakbar: We explicitly change that definition when users use the `-detailed-exitcode` flag :)

When users don't supply it, we consider non-zero exit codes failures. When they do, we use this logic:

> Using the OpenTofu/Terraform [-detailed-exitcode](https://opentofu.org/docs/cli/commands/plan/#other-options) flag with the run --all command results in an aggregate exit code being returned, rather than the exit code of any particular unit.
>
> The algorithm for determining the aggregate exit code is as follows:
>
> - If any unit throws a 1, Terragrunt will throw a 1.
> - If any unit throws a 2, but nothing throws a 1, Terragrunt will throw a 2.
> - If nothing throws a non-zero, Terragrunt will throw a 0.

From [`--all`](https://terragrunt.gruntwork.io/docs/reference/cli/commands/run/#all)

We still consider it a successful run, though.

test/integration_test.go
```diff
@@ -411,6 +411,32 @@ func TestRunAllDetailedExitCode_RetryableAfterDrift(t *testing.T) {
 	assert.Equal(t, 2, exitCode.GetFinalDetailedExitCode())
 }
 
+// TestDetailedExitCodeChangesPresentAllWithSource verifies that run --all correctly
+// propagates the detailed exit code when units use terraform { source = "." }.
+// This is a regression test for https://github.com/gruntwork-io/terragrunt/issues/5586
+func TestDetailedExitCodeChangesPresentAllWithSource(t *testing.T) {
```
denis256, line 417: It may be helpful to have a test with a clone of not existing remote path or another case when remote operation fails - it is throwing different errors

PR #5608, opened 2026-02-24
use io.Copy instead of reading files into memory

thisguycodes:
## Description

Improves all file copy operations by avoiding reading into memory, using `io.Copy` directly on the file descriptor instead. `io.Copy` opportunistically uses kernel-level options to avoid interstitial buffers entirely, this also probably reduces allocs in tests significantly.

This is a quick win while I finish support for `clonefile()` on macos (gotta make it not fail on non-macos).

## TODOs

- [ ] Update the docs.

## Summary by CodeRabbit

* **Improvements**
  * File operations now stream data instead of loading entire files into memory, improving performance and reducing memory usage for large files.
  * File writes preserve original permissions and use a safer create-and-write flow to avoid partial or corrupt outputs.
  * Error reporting for copy/write failures has been strengthened to surface combined write and close errors for more reliable failure detection.

### DIFF
--- a/internal/util/file.go
+++ b/internal/util/file.go
@@ -510,6 +510,7 @@ func CopyFile(source string, destination string) error {
 	}
 
 	err = WriteFileWithSamePermissions(source, destination, file)
+
 	return errors.New(errors.Join(err, file.Close()))
 }
 


internal/util/file.go
```diff
@@ -504,17 +504,18 @@ func TerragruntExcludes(path string) bool {
 
 // CopyFile copies a file from source to destination.
 func CopyFile(source string, destination string) error {
-	contents, err := os.ReadFile(source)
+	file, err := os.Open(source)
 	if err != nil {
 		return errors.New(err)
 	}
+	defer file.Close()
```
yhakbar: We should check this error.

thisguycodes: You're right. I could have sworn I added an errors.Join to do this

PR #5641, opened 2026-03-02
chore: Adding some testing for `logs`

yhakbar:
## Description

Adds some testing for the `pkg/logs` package.

Starts a new convention where we start to add an explicit `external_test.go` file that uses our `pkg` packages exclusively as black box packages the way that external consumers would.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **Tests**
  * Added extensive unit and integration tests covering logging context, public API usage, forced-level hooks, level parsing/round-trips, format parsing and placeholders, ANSI escape handling, logger behaviors (cloning, fields, errors, formatting), and writer behaviors (separators, default levels, parse functions).

### DIFF
--- a/pkg/log/external_test.go
+++ b/pkg/log/external_test.go
@@ -1,0 +1,324 @@
+// Tests in this file validate that the pkg/log package can be fully utilized as
+// an external dependency. Every scenario here imports only public packages
+// (pkg/log, pkg/log/format, pkg/log/format/placeholders, pkg/log/writer) and
+// never reaches into internal/ packages. If any of these tests fail to compile,
+// it signals a broken public API contract.
+package log_test
+
+import (
+	"bytes"
+	"testing"
+	"time"
+
+	"github.com/gruntwork-io/terragrunt/pkg/log"
+	"github.com/gruntwork-io/terragrunt/pkg/log/format"
+	"github.com/gruntwork-io/terragrunt/pkg/log/format/placeholders"
+	"github.com/gruntwork-io/terragrunt/pkg/log/writer"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestExternalLoggerLifecycle exercises the core create → configure → log →
+// clone workflow that an external consumer would follow.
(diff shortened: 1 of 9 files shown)


pkg/log/external_test.go
```diff
@@ -0,0 +1,324 @@
...
+
+	w := writer.New(
+		writer.WithLogger(logger),
+		writer.WithDefaultLevel(log.InfoLevel),
+		writer.WithMsgSeparator("\n"),
+	)
+
+	n, err := w.Write([]byte("line1\nline2"))
+	require.NoError(t, err)
+	assert.Len(t, "line1\nline2", n)
```
denis256, line 225: Misleading check, 
I think it will be better  
assert.Equal(t, 11, n) or assert.Equal(t, len("line1\nline2"), n)

yhakbar: Why? We're asserting that we've written all the content to the writer.

pkg/log/level_test.go
```diff
@@ -0,0 +1,283 @@
...
+		{expected: "err", level: log.ErrorLevel},
+		{expected: "wrn", level: log.WarnLevel},
+		{expected: "inf", level: log.InfoLevel},
+		{expected: "deb", level: log.DebugLevel},
+		{expected: "trc", level: log.TraceLevel},
+		{expected: "", level: log.Level(99)},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.expected, func(t *testing.T) {
```
denis256, line 97: This will generate tests with same name if same `expected` is set

pkg/log/external_test.go
```diff
@@ -0,0 +1,324 @@
...
+
+	logger.WithField("component", "auth").Info("field check")
+	assert.Contains(t, buf.String(), "component")
+
+	buf.Reset()
+
+	logger.WithFields(log.Fields{"a": 1, "b": 2}).Info("fields check")
+
+	output := buf.String()
+	assert.Contains(t, output, "a")
```
denis256, line 104: Weak test, it may match wrong values

PR #5655, opened 2026-03-03
fix: Fixing cosign

yhakbar:
## Description

Recent bump of the cosign action broke our signing because cosign no longer allows specifying the configuration and URL at the same time. 

```
  Error: cannot specify service URLs and use signing config
  error during command execution: cannot specify service URLs and use signing config
```

This fixes it.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **Chores**
  * Updated release signing configuration to modify authentication method for checksum signing operations. Minor formatting adjustments applied to signing scripts with no functional impact on outputs.

### DIFF
--- a/.github/scripts/release/sign-checksums.sh
+++ b/.github/scripts/release/sign-checksums.sh
@@ -44,20 +44,19 @@ function main {
   # GPG signing
   echo "Signing SHA256SUMS with GPG..."
   gpg --batch --yes -u "${GPG_FINGERPRINT}" \
-      --pinentry-mode loopback \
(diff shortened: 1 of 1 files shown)


yhakbar:
I'm moving this to in-draft. We need to revert the downgrade to the action and validate that this properly fixes things. It looks like there are more flags to adjust as cosign primarily uses a single asset for signing and verification instead of the `.pem` and `.sig` files. We'll likely want to update to have cosign generate all three for a smooth transition prior to dropping the old assets in a future release when the version of cosign that can verify with `.pem` and `.sig` is EoL.

PR #5711, opened 2026-03-19
feat(profiling): add automatic pprof collection

denis256:
## Description

- Add CPU, memory, and goroutine profile collection for Terragrunt runs.
- Add --profile-cpu, --profile-mem, --profile-goroutine, and --profile-dir behind the profiling experiment.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **New Features**
  * Added a `profiling` experiment with new global flags and `TG_PROFILE_*` environment variables for CPU, heap/memory, and goroutine profiling.
  * Added `--profile-dir` / `TG_PROFILE_DIR` for centralized output, including per-unit OpenTofu CPU profile collection, while respecting explicit `TOFU_CPU_PROFILE`.
* **Documentation**
  * Expanded CLI flag/global flag and troubleshooting/performance docs with end-to-end profiling and `go tool pprof` examples.
* **Bug Fixes**
  * Improved profile artifact creation and tightened profile file permissions; artifacts are produced only when profiling is enabled.
* **Tests**
  * Added unit and integration tests covering gating, flag/env parsing, output generation, directory layout, and error cases.
* **Chores**
  * Ignored `*.prof` profiling artifacts in version control.


docs/src/content/docs/06-troubleshooting/01-debugging.mdx
```diff
@@ -183,6 +183,48 @@ output "cluster_min_size" {
 Oops! It says `max` when it should be `min`. If we fix `ecs-cluster/outputs.tf`
 we should be golden!
 
+## Profiling
+
+If you need to diagnose performance bottlenecks, enable the `profiling` experiment to collect CPU, memory, and goroutine profiles:
+
+```bash
+# CPU profiling (Terragrunt process)
+TG_EXPERIMENT=profiling TG_PROFILE_CPU=terragrunt_cpu.prof terragrunt run -- plan
```
yhakbar: Instead of doing this, we should put an Aside right after the `Profiling` header that is `Before` gated so that we can't forget to remove these experiment call-outs in the examples.

docs/src/content/docs/06-troubleshooting/03-performance.mdx
```diff
@@ -309,7 +309,146 @@ You can learn more about benchmarking in Go by reading the [official documentati
 
 Use profiling tools like [pprof](https://github.com/google/pprof) to get a more detailed view of the performance of Terragrunt.
 
-For example, you could use the following command to profile a particular test:
+##### Runtime CPU Profiling with `TG_PROFILE_CPU`
+
+You can profile Terragrunt during normal operation by setting `TG_PROFILE_CPU` when the `profiling` experiment is enabled:
+
+```shell
+TG_EXPERIMENT=profiling TG_PROFILE_CPU=terragrunt_cpu.prof terragrunt run -- plan
```
yhakbar: Same feedback here. The heading should carry the Aside and leave these free.

docs/src/content/docs/06-troubleshooting/03-performance.mdx
```diff
@@ -309,7 +309,146 @@ You can learn more about benchmarking in Go by reading the [official documentati
...
+|-- modules/vpc/
+|   |-- tofu_cpu_init.prof
+|   `-- tofu_cpu_plan.prof
+|-- live/prod/database/
+|   `-- tofu_cpu_plan.prof
+`-- live/prod/app/
+    `-- tofu_cpu_plan.prof
+```
+
+The subdirectory layout matches the layout of your units relative to the working directory. This guarantees that no two OpenTofu processes write to the same file in parallel.
```
yhakbar:
```suggestion
The subdirectory layout matches the layout of your units relative to the working directory.
```

yhakbar (commented):
High-level feedback: All the docs updates here need to be controlled with `Before`/`Since`.

docs/src/content/docs/06-troubleshooting/03-performance.mdx
```diff
@@ -309,7 +309,146 @@ You can learn more about benchmarking in Go by reading the [official documentati
...
+
+##### Memory Profiling with `TG_PROFILE_MEM`
+
+You can capture a heap memory profile of the Terragrunt process by setting:
+
+```shell
+TG_EXPERIMENT=profiling TG_PROFILE_MEM=terragrunt_mem.prof terragrunt run -- plan
+```
+
+This is a snapshot taken at the end of the run after forcing a GC. Unlike CPU profiling, memory profiles are not collected from OpenTofu because there is no `TOFU_MEM_PROFILE` equivalent.
```
yhakbar:
```suggestion
This is a snapshot taken at the end of the run after forcing garbage collection. Unlike CPU profiling, memory profiles are not collected from OpenTofu because there is no `TOFU_MEM_PROFILE` equivalent.
```

docs/src/data/changelog/v1.1.2/pprof-profiling.mdx
```diff
@@ -0,0 +1,27 @@
+---
+version: "v1.1.1"
+category: "experiments-added"
+---
+
+#### `profiling` experiment adds pprof collection for Terragrunt runs
```
yhakbar, line 6: Do you think we should call out that this is really only useful if you want to debug the performance of Terragrunt as an application, and would like to explore ways to optimize Terragrunt? It likely won't help folks trying to improve the performance of their infrastructure management.

docs/src/data/experiments/profiling.mdx
```diff
@@ -0,0 +1,39 @@
...
+terragrunt --experiment profiling --profile-dir /tmp/profiles run --all -- plan
+```
+
+This is primarily intended for performance investigation and debugging of Terragrunt itself.
+
+The experiment can also be enabled with `TG_EXPERIMENT=profiling`, which is convenient when driving profiling entirely through environment variables.
+
+### `profiling` - How to provide feedback
+
+Provide feedback on the [Terragrunt GitHub Discussions](https://github.com/gruntwork-io/terragrunt/discussions) or by opening an issue.
```
yhakbar, line 29: Create a GitHub discussion for this and link it here.

docs/src/data/flags/profile-dir.mdx
```diff
@@ -0,0 +1,31 @@
...
+This flag is gated behind the [`profiling`](/reference/experiments/active#profiling) experiment. Enable it with `--experiment=profiling` or `TG_EXPERIMENT=profiling`.
+</Aside>
+
+Collect CPU, memory, and goroutine profiles into the given directory using conventional names:
+
+- `terragrunt_cpu.prof`
+- `terragrunt_mem.prof`
+- `terragrunt_goroutine.prof`
+
+In addition, each unit's downstream OpenTofu process writes its own CPU profile to `tofu_cpu_<command>.prof` (for example `tofu_cpu_plan.prof`) inside a unit-specific subdirectory (supported by OpenTofu 1.11 or later; Terraform ignores it). A unit at the working directory root writes directly into the profile directory, and units outside the working directory are placed under an `external/` subdirectory so they never collide.
```
yhakbar, line 21: To make sure I don't forget to check this later: We should make sure that the current working directory is an upper bound for this so that graph discovery can't result in escape of the profile directory.

internal/cli/commands/profiling.go
```diff
@@ -0,0 +1,274 @@
...
+	defaultCPUProfileName       = "terragrunt_cpu.prof"
+	defaultMemProfileName       = "terragrunt_mem.prof"
+	defaultGoroutineProfileName = "terragrunt_goroutine.prof"
+
+	profileFileMode = 0o600
+)
+
+// ErrProfilingRequiresExperiment is returned when profiling flags are used without the 'profiling' experiment.
+var ErrProfilingRequiresExperiment = errors.New(
+	"profiling flags require the 'profiling' experiment to be enabled (e.g., --experiment=profiling)",
```
yhakbar:
```suggestion
	"profiling flags require usage of the 'profiling' experiment (e.g., --experiment=profiling)",
```

internal/cli/commands/profiling.go
```diff
@@ -0,0 +1,274 @@
...
+
+// WrapWithProfiling wraps command actions with profile collection driven by the profiling flags.
+func WrapWithProfiling(
+	l log.Logger,
+	opts *options.TerragruntOptions,
+) func(ctx context.Context, cliCtx *clihelper.Context, action clihelper.ActionFunc) error {
+	var profiling bool
+
+	return func(ctx context.Context, cliCtx *clihelper.Context, action clihelper.ActionFunc) error {
+		if profiling {
```
yhakbar: Shouldn't this be short-circuiting if not profiling?

internal/cli/commands/profiling.go
```diff
@@ -0,0 +1,274 @@
...
+	if !opts.Experiments.Evaluate(experiment.Profiling) {
+		return nil, ErrProfilingRequiresExperiment
+	}
+
+	paths, err := resolveProfilePaths(opts)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := validateProfilePath(paths.mem, "memory"); err != nil {
```
yhakbar: Shouldn't there be some logic here to check if memory profiling was enabled?

internal/cli/commands/profiling.go
```diff
@@ -0,0 +1,274 @@
...
+		return nil, err
+	}
+
+	if err := validateProfilePath(paths.goroutine, "goroutine"); err != nil {
```
yhakbar: Shouldn't there be some logic here to check if goroutine profiling was enabled?

internal/cli/commands/profiling.go
```diff
@@ -0,0 +1,274 @@
...
+		return nil, err
+	}
+
+	cpuFile, err := startCPUProfile(paths.cpu)
```
yhakbar: Shouldn't there be some logic here to check if CPU profiling was enabled?

internal/cli/commands/profiling.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+// validateProfilePath verifies the named profile path is writable without holding the file open for the run.
+func validateProfilePath(path, name string) error {
+	f, err := createNamedProfileFile(path, name)
+	if err != nil || f == nil {
+		return err
+	}
+
+	if err := f.Close(); err != nil {
```
yhakbar: Why are we closing these files only to open them again later? Shouldn't we just return the file handle?

internal/cli/commands/profiling.go
```diff
@@ -0,0 +1,274 @@
...
+
+// startCPUProfile creates the CPU profile file and starts CPU profiling, no-op when path is empty.
+func startCPUProfile(path string) (*os.File, error) {
+	f, err := createNamedProfileFile(path, "CPU")
+	if err != nil || f == nil {
+		return f, err
+	}
+
+	if err := pprof.StartCPUProfile(f); err != nil {
+		f.Close() //nolint:errcheck
```
yhakbar: We can at least debug log this error.

internal/cli/commands/profiling.go
yhakbar: I see now that these are no-ops deeper in the call chain, but I think it might be better to handle this explicitly here instead.

internal/cli/commands/profiling.go
yhakbar: Why aren't we using vfs for these?

internal/cli/commands/profiling.go
yhakbar: Ah, I see now. Should this be called something like `profilingInProgress` or something?

internal/cli/commands/profiling.go
```diff
@@ -0,0 +1,274 @@
...
+		cpu:       opts.ProfileCPU,
+		mem:       opts.ProfileMem,
+		goroutine: opts.ProfileGoroutine,
+	}
+
+	if opts.ProfileDir == "" {
+		return paths, nil
+	}
+
+	absDir, err := filepath.Abs(opts.ProfileDir)
```
yhakbar: We should never call `filepath.Abs` like this. This should be an `IsAbs` check with a `Join` on `WorkingDir`.

internal/cli/commands/profiling.go
```diff
@@ -0,0 +1,274 @@
...
+	if err != nil {
+		return profilePaths{}, fmt.Errorf("could not resolve profile directory: %w", err)
```
yhakbar: Once the `filepath.Abs` call is removed above this should no longer be capable of erroring.

internal/cli/commands/profiling.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	if err := util.EnsureDirectory(absDir); err != nil {
```
yhakbar: We should be using vfs here.

internal/cli/commands/profiling.go
```diff
@@ -0,0 +1,274 @@
...
+
+		return nil, fmt.Errorf("could not start CPU profile: %w", err)
+	}
+
+	return f, nil
+}
+
+// stopCPUProfile finalizes CPU profiling and closes the profile file, if one was started.
+func stopCPUProfile(l log.Logger, cpuFile *os.File) {
+	if cpuFile == nil {
```
yhakbar: This is odd to me. We shouldn't call `stopCPUProfile` on a file that might not exist. We should be able to assert that this definitely does exist, and expect that someone has handled only calling this when the profile is in-progress.

internal/cli/commands/profiling.go
```diff
@@ -0,0 +1,274 @@
...
+		return
+	}
+
+	pprof.StopCPUProfile()
+	closeProfileFile(l, "CPU", cpuFile)
+}
+
+// writeMemProfile writes a heap profile snapshot to the given path, no-op when empty.
+func writeMemProfile(l log.Logger, path string) {
+	if path == "" {
```
yhakbar: We should be able to assert that this is non-empty and expect that the caller won't call this when we're not creating a file.

internal/cli/commands/profiling.go
```diff
@@ -0,0 +1,274 @@
...
+		return
+	}
+
+	runtime.GC()
```
yhakbar, line 203: We don't call out in the docs that enabling memory profiling can result in performance degradation due to this GC pause. We probably should.

internal/cli/commands/profiling.go
```diff
@@ -0,0 +1,274 @@
...
+	if err := pprof.WriteHeapProfile(f); err != nil { //NOSONAR:go:S4507 -- local, user-requested profile dump
+		l.Warnf("Could not write memory profile: %v", err)
+	}
+
+	closeProfileFile(l, "memory", f)
+}
+
+// writeGoroutineProfile writes a goroutine profile to the given path, no-op when empty.
+func writeGoroutineProfile(l log.Logger, path string) {
+	if path == "" {
```
yhakbar: We should be able to assert that this is non-empty and expect that the caller won't call this when we're not creating a file.

internal/cli/commands/profiling.go
```diff
@@ -0,0 +1,274 @@
...
+	if err := pprof.Lookup("goroutine").WriteTo(f, 0); err != nil { //NOSONAR:go:S4507 -- local, user-requested profile dump
+		l.Warnf("Could not write goroutine profile: %v", err)
+	}
+
+	closeProfileFile(l, "goroutine", f)
+}
+
+// createNamedProfileFile creates the output file for the named profile, nil when path is empty.
+func createNamedProfileFile(path, name string) (*os.File, error) {
+	if path == "" {
```
yhakbar: We should be able to assert that this is non-empty and expect that the caller won't call this when we're not creating a file.

internal/cli/commands/profiling.go
```diff
@@ -0,0 +1,274 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not create %s profile: %w", name, err)
+	}
+
+	return f, nil
+}
+
+// createProfileFile creates a profile output file readable only by the owner, tightening permissions on pre-existing files.
+func createProfileFile(path string) (*os.File, error) {
+	f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, profileFileMode)
```
yhakbar: We should be able to use vfs for this.

internal/cli/commands/profiling.go
```diff
@@ -0,0 +1,274 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if err := f.Chmod(profileFileMode); err != nil {
+		f.Close() //nolint:errcheck
```
yhakbar: We should be able to debug log this error.

internal/cli/commands/profiling.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+// createProfileFile creates a profile output file readable only by the owner, tightening permissions on pre-existing files.
+func createProfileFile(path string) (*os.File, error) {
+	f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, profileFileMode)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := f.Chmod(profileFileMode); err != nil {
```
yhakbar: If switching to vfs, this might need to be added to the vfs package. In that case, I think we can leave chmod a no-op for memFS. Open to being wrong.

internal/cli/commands/profiling.go
```diff
@@ -0,0 +1,274 @@
...
+
+		return nil, err
+	}
+
+	return f, nil
+}
+
+// closeProfileFile closes a finished profile file, logging a warning if the final flush fails.
+func closeProfileFile(l log.Logger, name string, f *os.File) {
+	if f == nil {
```
yhakbar: We should be able to assert that this is non-nil and expect that the caller won't call this when we're not closing a file. If we need this to be idempotent, maybe use an `ensure` prefix to the function name.

internal/cli/commands/profiling_test.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+func TestWrapWithProfilingRequiresExperiment(t *testing.T) {
+	t.Parallel()
+
+	opts := options.NewTerragruntOptions()
+	opts.ProfileCPU = filepath.Join(t.TempDir(), "cpu.prof")
+
+	called := false
+	err := commands.WrapWithProfiling(logger.CreateLogger(), opts)(context.Background(), nil, func(_ context.Context, _ *clihelper.Context) error {
```
yhakbar: NIT: Prefer `t.Context()`. Also, we can wrap this to obey `lll` linting.

internal/cli/commands/profiling_test.go
```diff
@@ -0,0 +1,95 @@
...
+	assert.NoFileExists(t, opts.ProfileCPU)
+}
+
+func TestWrapWithProfilingNoFlagsRunsAction(t *testing.T) {
+	t.Parallel()
+
+	opts := options.NewTerragruntOptions()
+
+	called := false
+	err := commands.WrapWithProfiling(logger.CreateLogger(), opts)(context.Background(), nil, func(_ context.Context, _ *clihelper.Context) error {
```
yhakbar: NIT: Prefer `t.Context()`. Also, we can wrap this to obey `lll` linting.

internal/cli/commands/profiling_test.go
yhakbar: That `nil` there is a smell. I think the function should assert that it's non-nil and force us to explicitly pass in test options so we can't accidentally forget it in production.

internal/cli/commands/profiling_test.go
yhakbar: That `nil` there is a smell. I think the function should assert that it's non-nil and force us to explicitly pass in test options so we can't accidentally forget it in production.

internal/cli/commands/profiling_test.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func TestWrapWithProfilingWritesProfile(t *testing.T) {
+	t.Parallel()
+
+	opts := options.NewTerragruntOptions()
+	opts.ProfileGoroutine = filepath.Join(t.TempDir(), "goroutine.prof")
+	require.NoError(t, opts.Experiments.EnableExperiment(experiment.Profiling))
+
+	err := commands.WrapWithProfiling(logger.CreateLogger(), opts)(context.Background(), nil, func(_ context.Context, _ *clihelper.Context) error {
```
yhakbar: Same comment about ctx and nil.

internal/cli/commands/profiling_test.go
```diff
@@ -0,0 +1,95 @@
...
+		t.Skip("file permission bits are not enforced on Windows")
+	}
+
+	opts := options.NewTerragruntOptions()
+	opts.ProfileGoroutine = filepath.Join(t.TempDir(), "goroutine.prof")
+	require.NoError(t, opts.Experiments.EnableExperiment(experiment.Profiling))
+
+	require.NoError(t, os.WriteFile(opts.ProfileGoroutine, []byte("stale"), 0o644))
+
+	err := commands.WrapWithProfiling(logger.CreateLogger(), opts)(context.Background(), nil, func(_ context.Context, _ *clihelper.Context) error {
```
yhakbar: Same comment about ctx and nil.

internal/cli/flags/global/flags.go
```diff
@@ -312,6 +320,41 @@ func NewTelemetryFlags(opts *options.TerragruntOptions, prefix flags.Prefix) cli
...
+// NewProfileFlags creates the profiling flags backing the opts.Profile* fields.
+func NewProfileFlags(opts *options.TerragruntOptions, prefix flags.Prefix) clihelper.Flags {
+	tgPrefix := prefix.Prepend(flags.TgPrefix)
+
+	return clihelper.Flags{
+		flags.NewFlag(&clihelper.GenericFlag[string]{
+			Name:        ProfileCPUFlagName,
+			EnvVars:     tgPrefix.EnvVars(ProfileCPUFlagName),
+			Destination: &opts.ProfileCPU,
+			Usage:       "Write a CPU profile to the given path. Requires the 'profiling' experiment.",
```
yhakbar: Can't we check this with an After?

Also, I wonder if we should be adjusting this usage text based on the experiment default so we can't accidentally forget to update it later.

Same comment for the rest of these.

test/integration_profile_test.go
```diff
@@ -0,0 +1,344 @@
...
+	assert.Positive(t, info.Size(), "profile %s should be non-empty", path)
+}
+
+// findProfileFiles returns all *.prof files under root, recursively.
+func findProfileFiles(t *testing.T, root string) []string {
+	t.Helper()
+
+	var profiles []string
+
+	err := filepath.WalkDir(root, func(path string, d os.DirEntry, err error) error {
```
yhakbar: We can use vfs and parallel walk for this, right?

yhakbar (approved):

docs/src/content/docs/06-troubleshooting/03-performance.mdx
```diff
@@ -309,7 +310,186 @@ You can learn more about benchmarking in Go by reading the [official documentati
...
-For example, you could use the following command to profile a particular test:
+<Since version="v1.1.1">
+
+<Aside type="tip" title="Experimental">
+Profiling is part of the [`profiling` experiment](/reference/experiments/active#profiling). Enable it with `--experiment=profiling` or `TG_EXPERIMENT=profiling`.
+</Aside>
+
+</Since>
+
+##### Runtime CPU Profiling with `TG_PROFILE_CPU`
```
yhakbar, line 319: This is gonna look busted actually. It's just gonna be an empty header.

yhakbar (changes requested):

docs/src/content/docs/04-reference/02-cli/98-global-flags.mdx
```diff
@@ -54,6 +54,22 @@ The Terragrunt CLI supports the following global flags:
 
 <Flag slug="non-interactive" />
 
+## --profile-cpu
+
+<Flag slug="profile-cpu" />
```
yhakbar, line 69: We need to gate these with Before/Since.

docs/src/content/docs/06-troubleshooting/03-performance.mdx
```diff
@@ -309,7 +309,150 @@ You can learn more about benchmarking in Go by reading the [official documentati
 
 Use profiling tools like [pprof](https://github.com/google/pprof) to get a more detailed view of the performance of Terragrunt.
 
-For example, you could use the following command to profile a particular test:
+<Aside type="tip" title="Experimental">
```
yhakbar, line 329: This should go inside the headings, not before it so that anchor links don't miss the aside.

yhakbar (dismissed):

yhakbar (approved):

PR #5721, opened 2026-03-20
fix: Fixing `hcl fmt` on unintended files during scaffold

### LINKED ISSUE #3657: Terragrunt scaffold should only run hclfmt on updated folders instead of the working directory
## Describe the enhancement

Currently when using terragrunt catalog and/or scaffold, the scaffold command runs a `hclfmt` on the working directory. This can subsequently end up touching far more files than just the ones that the scaffold command created. Ideally, the `hclfmt` is either made optional, or only runs against the terragrunt files that scaffold itself generated.

## Additional context

It seems to me that the current assumption is that people would run scaffold from the directory they want to generate terragrunt files in. Given that assumption the idea of running the `hclfmt` on the working directory isn't a terrible one. However, terragrunt's own [architecture proposal is to use a base/child structure](https://terragrunt.gruntwork.io/docs/features/keep-your-terragrunt-architecture-dry/#using-include-to-dry-common-terragrunt-config). The easiest way I've found to have scaffold generate both a base and a child terragrunt file, is by running scaffold from the root of the repo (and have the user input the file paths for both directories). This is reasonably easy to achieve using boilerplate dependencies. 

The biggest problem I'm facing is that running the terragrunt scaffold from the root of the directory will run the `hclfmt` against the entire repo, instead of just the two (or more) subfolders that the command itself created.

## Example

The below boilerplate template essentially just created two dependencies in two folders, the path of which is required as user input.
```yaml
variables:
  - name: sourceUrl
    description: Url for source module
    type: string
    default: ""

  - name: requiredVariables
    description: Required input variables for module
    type: list
    default: []

  - name: optionalVariables
    description: Optional input variables for module
    type: list
    default: []

  - name: baseFileName
    description: Name for the base terragrunt file WITHOUT file extension, e.g. "dynamodb". (NOT "terragrunt")
    type: string
    validations:
      - required
      - length-3-50

  - name: baseFilePath
    description: |-
      Path for the base file, e.g. "base/team/service/"
      DO NOT add a leading slash, DO add a trailing slash 
    type: string
    validations:
      - required
      - length-5-500

  - name: leafFilePath
    description: |-
      Path for the leaf file, e.g. "development/us-east-1/team/service/"
      DO NOT add a leading slash, DO add a trailing slash 
    type: string
    validations:
      - required
      - length-5-500

partials:
  - ../.boilerplate-files/boilerplate-partials/*.hcl

dependencies:
  - name: base
    template-url: ../.boilerplate-files/base-file
    output-folder: "{{ .baseFilePath }}"
  
  - name: leaf
    template-url: ../.boilerplate-files/leaf-file
    output-folder: "{{ .leafFilePath }}"
```
(dependency templates not included for brevity)

Technically you can run this from anywhere in the repo, but the most logical place to run the scaffold command for the above boilerplate template is the root of the repo. All files generated will be in subfolders, so running hclfmt against the entire working directory is not ideal, as it may touch a whole list of files which are not in any way related to the scaffold call.

yhakbar:
## Description

Prevents `hcl fmt` running on stuff that isn't generated by scaffold.

Fixes #3657

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **Bug Fixes**
  * Scaffold now limits formatting to newly generated files and their declared dependencies, avoiding changes to pre-existing files.

* **New Features**
  * Added targeted file-formatting capability to format a provided list of files.

* **Tests**
  * Added unit and integration tests verifying targeted formatting behavior and that existing files remain unmodified.

* **Chores**
  * Updated boilerplate dependency to v0.15.0 and pruned several unused transitive dependencies.

### DIFF
--- a/internal/cli/commands/hcl/format/format.go
+++ b/internal/cli/commands/hcl/format/format.go
@@ -14,6 +14,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 
 	"github.com/gruntwork-io/terragrunt/internal/component"
 	"github.com/gruntwork-io/terragrunt/internal/filter"
@@ -154,9 +155,12 @@ func RunForFiles(ctx context.Context, l log.Logger, opts *options.TerragruntOpti
 
 	g.SetLimit(limit)
 
-	errs := make([]error, len(files))
+	var (
+		mu   sync.Mutex
+		errs []error
+	)
 
-	for i, file := range files {
+	for _, file := range files {
 		if !strings.HasSuffix(file, ".hcl") {
 			continue
 		}
@@ -166,9 +170,11 @@ func RunForFiles(ctx context.Context, l log.Logger, opts *options.TerragruntOpti
 		}
 
 		g.Go(func() error {
-			err := formatTgHCL(gctx, l, opts, file)
-			if err != nil {
-				errs[i] = err
+			if err := formatTgHCL(gctx, l, opts, file); err != nil {
+				mu.Lock()
+
+				errs = append(errs, err)
+				mu.Unlock()
 			}
 
 			return nil
--- a/internal/cli/commands/scaffold/scaffold.go
+++ b/internal/cli/commands/scaffold/scaffold.go
@@ -240,7 +240,17 @@ func Run(ctx context.Context, l log.Logger, opts *options.TerragruntOptions, mod
 		return errors.New(err)
 	}
 
-	allFiles := slices.Concat(result.GeneratedFiles, collectDependencyFiles(result.Dependencies))
+	depFiles, err := collectDependencyFiles(result.Dependencies, 0)
+	if err != nil {
+		return errors.New(err)
+	}
+
+	allFiles := slices.Concat(result.GeneratedFiles, depFiles)
+
+	for i, f := range allFiles {
+		allFiles[i] = filepath.Clean(f)
+	}
+
 	allFiles = slices.Compact(slices.Sorted(slices.Values(allFiles)))
 
 	l.Infof("Running fmt on generated code %s", outputDir)
@@ -632,9 +642,15 @@ type parsedURL struct {
 	path   string
 }
 
+const maxDependencyDepth = 100
+
 // collectDependencyFiles recursively collects file paths from all boilerplate
-// dependencies and their nested sub-dependencies.
-func collectDependencyFiles(deps []manifest.ManifestDependency) []string {
+// dependencies and their nested sub-dependencies up to maxDependencyDepth.
+func collectDependencyFiles(deps []manifest.ManifestDependency, depth int) ([]string, error) {
+	if depth >= maxDependencyDepth {
+		return nil, errors.New(MaxDependencyDepthExceededError{})
+	}
+
 	var files []string
 
 	for i := range deps {
@@ -642,10 +658,22 @@ func collectDependencyFiles(deps []manifest.ManifestDependency) []string {
 			files = append(files, f.Path)
 		}
 
-		files = append(files, collectDependencyFiles(deps[i].Dependencies)...)
+		subFiles, err := collectDependencyFiles(deps[i].Dependencies, depth+1)
+		if err != nil {
+			return nil, err
+		}
+
+		files = append(files, subFiles...)
 	}
 
-	return files
+	return files, nil
+}
+
+type MaxDependencyDepthExceededError struct {
+}
+
+func (err MaxDependencyDepthExceededError) Error() string {
+	return fmt.Sprintf("dependency depth limit of %d exceeded, possible circular dependencies", maxDependencyDepth)
 }
 
 type failedToParseURLError struct {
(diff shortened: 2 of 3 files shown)


internal/cli/commands/scaffold/scaffold.go
```diff
@@ -231,14 +233,19 @@ func Run(ctx context.Context, l log.Logger, opts *options.TerragruntOptions, mod
...
-	if err := templates.ProcessTemplate(boilerplateOpts, boilerplateOpts, emptyDep); err != nil {
+	emptyDep := &variables.Dependency{}
+
+	result, err := templates.ProcessTemplateWithContext(ctx, boilerplateOpts, boilerplateOpts, emptyDep)
+	if err != nil {
 		return errors.New(err)
 	}
 
+	allFiles := slices.Concat(result.GeneratedFiles, collectDependencyFiles(result.Dependencies))
+	allFiles = slices.Compact(slices.Sorted(slices.Values(allFiles)))
```
denis256, line 254: I was wondering if we need to clean paths before sort/compact with something like `filepath.Clean`

internal/cli/commands/hcl/format/format.go
```diff
@@ -141,6 +141,45 @@ func Run(ctx context.Context, l log.Logger, opts *options.TerragruntOptions) err
 	return errors.Join(errs...)
 }
 
+// RunForFiles formats only the specified files instead of walking the entire directory tree.
+// Files may be relative paths (resolved against baseDir) or absolute paths.
+// This is used by scaffold to format only the files that were generated by boilerplate.
+func RunForFiles(ctx context.Context, l log.Logger, opts *options.TerragruntOptions, baseDir string, files []string) error {
```
denis256, line 148: baseDir - should be checked for empty/relative path?

internal/cli/commands/hcl/format/format.go
```diff
@@ -141,6 +141,45 @@ func Run(ctx context.Context, l log.Logger, opts *options.TerragruntOptions) err
...
+	g, gctx := errgroup.WithContext(ctx)
+
+	limit := opts.Parallelism
+	if limit == options.DefaultParallelism {
+		limit = runtime.NumCPU()
+	}
+
+	g.SetLimit(limit)
+
+	errs := make([]error, len(files))
```
denis256: Possible wasteful allocation of errors, since not all files are hcl files

internal/cli/commands/hcl/format/format.go
```diff
@@ -141,6 +141,45 @@ func Run(ctx context.Context, l log.Logger, opts *options.TerragruntOptions) err
 	return errors.Join(errs...)
 }
 
+// RunForFiles formats only the specified files instead of walking the entire directory tree.
+// Files may be relative paths (resolved against baseDir) or absolute paths.
+// This is used by scaffold to format only the files that were generated by boilerplate.
+func RunForFiles(ctx context.Context, l log.Logger, opts *options.TerragruntOptions, baseDir string, files []string) error {
+	g, gctx := errgroup.WithContext(ctx)
```
denis256, line 149: Make sense to add a warn/debug message if files is empty?

internal/cli/commands/scaffold/scaffold.go
```diff
@@ -625,6 +632,22 @@ type parsedURL struct {
...
+// dependencies and their nested sub-dependencies.
+func collectDependencyFiles(deps []manifest.ManifestDependency) []string {
+	var files []string
+
+	for i := range deps {
+		for _, f := range deps[i].Files {
+			files = append(files, f.Path)
+		}
+
+		files = append(files, collectDependencyFiles(deps[i].Dependencies)...)
```
denis256: No depth limit, possible infinite recursion and/or high disk io

internal/cli/commands/hcl/format/format.go
denis256, line 148: No tests for RunForFiles()

internal/cli/commands/scaffold/scaffold.go
yhakbar: Addressed.

internal/cli/commands/hcl/format/format.go
yhakbar: I'm not sure that's a good idea. Opted not to do this, at least for now.

internal/cli/commands/hcl/format/format.go
yhakbar: Switched to mutex guarded slice.

internal/cli/commands/hcl/format/format.go
yhakbar: I'm not sure we should have this. Why?

internal/cli/commands/scaffold/scaffold.go
yhakbar: Added recursion limit.

internal/cli/commands/hcl/format/format.go
yhakbar: Added tests.

PR #5746, opened 2026-03-27
fix: Fixing #4153

### LINKED ISSUE #4153: "Could not convert include to the execution ctx to evaluate additional locals" when chaining dependencies
## Describe the bug

Related to #3036 -- in fact this was observed in #3036 too but the focus of that issue wasn't on this particular error -- when chaining multiple dependencies Terragrunt outputs the error "Could not convert include to the execution ctx to evaluate additional locals" 

This isn't a regression.  It occurs on 0.77.14 and it occurs on all versions I've tested since #3036.

As with #3036 the error appears to be innocuous.  At least I don't experience any functional problems.

Consider the following layout:

```
.
├── tfstate/
│   ├── outputs.tf
│   └── terragrunt.hcl
├── vpc/
│   ├── outputs.tf
│   ├── terragrunt.hcl
├── eks/
│   ├── main.tf
│   ├── terragrunt.hcl
└── terragrunt.hcl
```

With dependencies:

image

The `tfstate` dependency is declared in the root-level terragrunt.hcl for purposes of generating `backend.tf`.  The vpc module works fine, however the eks module generates this error:

```
ERRO[0000] Could not convert include to the execution ctx to evaluate additional locals in file /tmp/terragrunt-dep-bug/vpc/terragrunt.hcl  prefix=[/tmp/terragrunt-dep-bug/vpc]
```

## Steps To Reproduce

The same sample project from #3036 reproduces, available at https://github.com/jtackaberry/terragrunt-dep-bug 

```bash
# Clone the example
git clone https://github.com/jtackaberry/terragrunt-dep-bug.git

# Apply mock tfstate module -- this is fine
cd terragrunt-dep-bug/tfstate
terragrunt apply

# Apply mock vpc module -- this is fine
cd ../vpc
terragrunt apply

# Apply mock eks module -- error occurs here
cd ../eks
terragrunt apply
```

(Note that the example uses the now-deprecated root-level terragrunt.hcl, but it's unrelated to this issue -- the error persists even if I rename it to `root.hcl`).

## Expected behavior

AFAICT functionally everything works as expected.  So the expected behavior is that the error isn't logged, if it is in fact bogus.

It's also entirely possible I don't understand the error -- it's a bit cryptic -- so if cluebatting is required, it will be graciously received.

## Versions

- Terragrunt version: 0.62.2 - 0.77.14
- OpenTofu/Terraform version: 1.9.0
- Environment details (Ubuntu 20.04, Windows 10, etc.): Ubuntu 24.04

yhakbar:
## Description

Fixes #4153.

Downgrades an error to a debug log entry when parsing a chained dependency with an exposed include.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Release Notes

* **Bug Fixes**
  * Improved partial parsing error handling for exposed includes, allowing discovery commands like `terragrunt list` to succeed even when includes have unresolved dependencies. Full parsing still properly reports errors when dependencies cannot be resolved.

* **Tests**
  * Added regression tests for chained dependencies with exposed includes and partial parse error scenarios.

### DIFF
--- a/pkg/config/config_helpers.go
+++ b/pkg/config/config_helpers.go
@@ -240,17 +240,19 @@ func createTerragruntEvalContext(ctx context.Context, pctx *ParsingContext, l lo
 		// For each include block, check if we want to expose the included config, and if so, add under the include
 		// variable.
 		exposedInclude, err := includeMapAsCtyVal(ctx, pctx, l)
+		if err != nil && len(pctx.PartialParseDecodeList) == 0 {
+			return nil, fmt.Errorf("could not resolve exposed includes for eval context in %s: %w", configPath, err)
+		}
+
 		if err != nil {
-			if len(pctx.PartialParseDecodeList) > 0 {
-				// Include resolution can fail during partial parsing of configs in dependency chains,
-				// e.g. when an included config has a dependency block whose outputs aren't yet available.
-				// This is expected and non-fatal — locals referencing the include will be left unevaluated,
-				// and the system will fall back to full parsing when needed.
-				l.Debugf("Could not resolve exposed includes for eval context in %s (partial parse): %v", configPath, err)
-			} else {
-				return nil, fmt.Errorf("could not resolve exposed includes for eval context in %s: %w", configPath, err)
-			}
-		} else {
+			// Include resolution can fail during partial parsing of configs in dependency chains,
+			// e.g. when an included config has a dependency block whose outputs aren't yet available.
+			// This is expected and non-fatal — locals referencing the include will be left unevaluated,
+			// and the system will fall back to full parsing when needed.
+			l.Debugf("Could not resolve exposed includes for eval context in %s (partial parse): %v", configPath, err)
+		}
+
+		if err == nil {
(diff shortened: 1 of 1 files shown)


pkg/config/config_helpers.go
```diff
@@ -241,13 +241,17 @@ func createTerragruntEvalContext(ctx context.Context, pctx *ParsingContext, l lo
 		// variable.
 		exposedInclude, err := includeMapAsCtyVal(ctx, pctx, l)
 		if err != nil {
-			return evalCtx, err
+			// Include resolution can fail during partial parsing of configs in dependency chains,
+			// e.g. when an included config has a dependency block whose outputs aren't yet available.
+			// This is expected and non-fatal — locals referencing the include will be left unevaluated,
+			// and the system will fall back to full parsing when needed.
+			l.Debugf("Could not resolve exposed includes for eval context in %s: %v", configPath, err)
```
denis256: error swallowed unconditionally across all parsing contexts 🤔  make sense to handle only for partial parse?

yhakbar: This is a good point. I'll update the implementation.

PR #5779, opened 2026-04-01
fix: add retry patterns for transient provider/registry network errors

denis256:
## Description

* added new retryable error patterns for transient network failures during provider resolution and registry communication

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Release Notes

* **New Features**
  * Automatic retries for transient network failures during provider resolution and registry operations, including TLS handshake timeouts, TCP connection issues, and context deadline errors.
  * Windows symlink troubleshooting tip for users with provider plugin caching enabled.

* **Documentation**
  * Added changelog documenting automatic retry coverage improvements for CI-like environments.


internal/retry/defaults.go
yhakbar, line 1: Why do we compile these at runtime?
https://github.com/gruntwork-io/terragrunt/blob/1ee0396d68a56e739a4e246cda0a59139eeae5c7/pkg/options/auto_retry_options.go#L18

We can use MustCompile on these.

internal/retry/defaults.go
```diff
@@ -30,4 +34,35 @@ var DefaultRetryableErrors = []string{
...
+func init() {
+	DefaultRetryableRegexps = make([]*regexp.Regexp, len(DefaultRetryableErrors))
+	for i, pat := range DefaultRetryableErrors {
+		re, err := regexp.Compile(pat)
+		if err != nil {
+			panic(fmt.Sprintf("retry: pattern %d failed to compile: %q: %v", i, pat, err))
+		}
+
+		DefaultRetryableRegexps[i] = re
+	}
```
thisguycodes, line 67: `regexp.MustCompile` does this panic logic for us, and if we want to we can do it inline at the top level instead of in `init()` (`init()` does work though).

yhakbar: I think it's worth doing too.

PR #5793, opened 2026-04-03
fix coverage compare regarding retries

thisguycodes:
## Description

our coverage comparison workflow errors if there are multiple matching artifacts for the previous run (e.g. if the previous run was retried), [example](https://github.com/gruntwork-io/terragrunt/actions/runs/23920587424/job/69853281965?pr=5608).

This modifies the filter to select the most recent (github returns them ordered), same as the filter for recent runs.

## TODOs

- [ ] Update the docs.

## Summary by CodeRabbit

* **Chores**
  * Made coverage-comparison workflow more reliable by improving previous-run detection so comparisons target the correct run.
  * Strengthened artifact selection to consistently find the correct coverage summary for comparison.
  * Standardized how comparison targets are chosen and retrieved to reduce intermittent failures during CI coverage checks.

### DIFF
--- a/.github/workflows/coverage-compare-main.yml
+++ b/.github/workflows/coverage-compare-main.yml
@@ -24,6 +24,8 @@ jobs:
         id: prev
         env:
           GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          RUN_ID: ${{ github.run_id }}
         run: |
           mkdir -p previous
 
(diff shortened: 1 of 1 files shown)


.github/workflows/coverage-compare-main.yml
```diff
@@ -44,7 +44,7 @@ jobs:
 
           # Find the coverage-summary.json artifact in that run
           ARTIFACT_ID=$(gh api "repos/${{ github.repository }}/actions/runs/${PREV_RUN_ID}/artifacts" \
-            --jq '.artifacts[] | select(.name == "coverage-summary.json") | .id')
+            --jq '[.artifacts[] | select(.name == "coverage-summary.json")][0].id')
```
yhakbar, line 49: Should we be checking that len > 0? Same with the check above this. Also, we should be removing those direct usages of `${{ }}` and load them in as environment variables.

thisguycodes: checking > 0: this workflow doesn't make sense with no results from the filter, so I think letting it error here is fine

I didn't even notice the `${{ }}` injection, I'll swap it to env vars while I'm in there.

thisguycodes: correcting myself; with an empty array `ARTIFACT_ID` will be `null`, simple example:
```shell
$ echo '[]' | jq '.[0]'
null
```

so the next line, which checks for `null`, correctly catches this

PR #5801, opened 2026-04-05
feat: make version optional in tfr:// module registry URLs

### LINKED ISSUE #2572: Feature request: support latest version when using `tfr` protocol in `terraform.source`

raman1236:
## Description

This PR makes the `version` query parameter optional when using the `tfr://` protocol for Terraform module registry sources. When the version is omitted, Terragrunt will automatically query the registry's [List Available Versions](https://developer.hashicorp.com/terraform/registry/api-docs#list-available-versions-for-a-specific-module) API endpoint to determine the latest version and use it for the download.

## Motivation

Currently, the `version` parameter is mandatory in `tfr://` URLs:
```hcl
terraform {
}
```

This differs from native Terraform, where the `version` attribute is optional. Many users have expressed the desire to not pin versions in development/non-production environments, or simply want to use the latest version without specifying it explicitly.

With this change, the following is now valid:
```hcl
terraform {
  source = "tfr:///terraform-aws-modules/vpc/aws"
}
```

When no version is specified, Terragrunt will:
1. Query the registry's versions endpoint (`GET /{modules.v1}/{namespace}/{name}/{system}/versions`)
2. Select the latest (first listed) version
3. Log the resolved version for transparency
4. Continue with the standard download flow

## Changes

- **`internal/tf/getter.go`**: Modified the `Get` method to allow missing version query parameter. Added `GetLatestModuleVersion` function that queries the registry API to resolve the latest module version, along with supporting types for parsing the API response.
- **`internal/tf/getter_test.go`**: Added `TestGetLatestModuleVersion` and `TestTFRGetterWithoutVersion` tests to verify the new functionality.

## Testing

- All existing tests pass (no regressions)
- New unit tests verify:
  - `GetLatestModuleVersion` correctly queries the public Terraform registry
  - `RegistryGetter.Get` works without a version parameter

Fixes #2572
Related: #1930

## Summary by CodeRabbit

* **New Features**
  * Registry module URLs can omit the version parameter — the system now automatically resolves and uses the latest available module version.
  * When the latest version is chosen, an informational log entry is emitted indicating which version was resolved.

* **Tests**
  * Added tests verifying automatic latest-version resolution and successful module fetching when no explicit version is provided.


internal/tf/getter.go
```diff
@@ -132,21 +132,32 @@ func (tfrGetter *RegistryGetter) Get(dstPath string, srcURL *url.URL) error {
...
+		// When version is not specified, query the registry to find the latest version
+		latestVersion, err := GetLatestModuleVersion(ctx, tfrGetter.Logger, registryDomain, moduleRegistryBasePath, modulePath)
+		if err != nil {
+			return err
+		}
+
+		tfrGetter.Logger.Infof("No version specified for module %s, using latest version %s", modulePath, latestVersion)
+
+		version = latestVersion
+	}
```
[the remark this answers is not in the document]
raman1236: Good catch! Fixed in 45eec2cc. Added a nil check before calling `Logger.Infof()`:

```go
if tfrGetter.Logger != nil {
    tfrGetter.Logger.Infof("No version specified for module %s, using latest version %s", modulePath, latestVersion)
}
```

Also added `TestTFRGetterWithoutVersionNilLogger` test that constructs a `RegistryGetter` without setting Logger (matching production behavior via `UpdateGetters`) to verify it doesn't panic.

internal/tf/getter.go
```diff
@@ -273,6 +284,56 @@ func GetModuleRegistryURLBasePath(ctx context.Context, logger log.Logger, domain
...
+		return "", errors.Errorf("failed to parse module versions response for %s: %w", modulePath, err)
+	}
+
+	if len(versionsResp.Modules) == 0 || len(versionsResp.Modules[0].Versions) == 0 {
+		return "", errors.Errorf("no versions found for module %s on registry %s", modulePath, registryDomain)
+	}
+
+	// The registry returns versions in order, with the latest first
+	return versionsResp.Modules[0].Versions[0].Version, nil
+}
```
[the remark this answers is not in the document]
raman1236: Great point — you're right that the Terraform Module Registry Protocol does not guarantee version ordering. Fixed in 45eec2cc.

Now using `hashicorp/go-version` (already a direct dependency of the project) to parse all versions and sort by semver, then selecting the highest:

```go
var parsed []*goversion.Version
for _, v := range versionsResp.Modules[0].Versions {
    pv, err := goversion.NewVersion(v.Version)
    if err != nil {
        continue // skip unparseable versions
    }
    parsed = append(parsed, pv)
}

sort.Sort(goversion.Collection(parsed))
return parsed[len(parsed)-1].Original(), nil
```

The test confirms it returns `6.6.1` (actual latest) for `terraform-aws-modules/vpc/aws`.

internal/tf/getter.go
```diff
@@ -273,6 +288,74 @@ func GetModuleRegistryURLBasePath(ctx context.Context, logger log.Logger, domain
...
+	modulePath = strings.TrimSuffix(modulePath, "/")
+	modulePath = strings.TrimPrefix(modulePath, "/")
+
+	versionsPath := fmt.Sprintf("%s/%s/versions", moduleRegistryBasePath, modulePath)
+
+	versionsURL := &url.URL{
+		Scheme: "https",
+		Host:   registryDomain,
+		Path:   versionsPath,
+	}
```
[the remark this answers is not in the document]
raman1236: Good catch! Fixed in e4c9cd00. 

`GetLatestModuleVersion` now uses `url.Parse` to detect whether the `modules.v1` base path is an absolute URL (with scheme) or a relative path — matching the existing behavior of `BuildRequestURL`:

```go
versionsURL, err := url.Parse(versionsPath)
if err != nil {
    return "", errors.Errorf("failed to parse versions URL for %s: %w", modulePath, err)
}

// If the base path is relative (no scheme), construct the full URL using the registry domain.
if versionsURL.Scheme == "" {
    versionsURL = &url.URL{
        Scheme: "https",
        Host:   registryDomain,
        Path:   versionsPath,
    }
}
```

Also added `TestGetLatestModuleVersionAbsoluteBasePath` regression test that passes an absolute `modules.v1` URL (`https://registry.terraform.io/v1/modules/`).

internal/tf/getter.go
```diff
@@ -273,6 +294,82 @@ func GetModuleRegistryURLBasePath(ctx context.Context, logger log.Logger, domain
...
+		return "", errors.Errorf("failed to parse module versions response for %s: %w", modulePath, err)
+	}
+
+	if len(versionsResp.Modules) == 0 || len(versionsResp.Modules[0].Versions) == 0 {
+		return "", errors.Errorf("no versions found for module %s on registry %s", modulePath, registryDomain)
+	}
+
+	// The registry API does not guarantee version ordering, so we parse and sort by semver
+	// to reliably find the latest version.
+	var parsed []*goversion.Version
```
yhakbar, line 353: NIT: This can be pre-allocated to len(versionsResp.Modules[0].Versions).

internal/tf/getter.go
```diff
@@ -273,6 +294,82 @@ func GetModuleRegistryURLBasePath(ctx context.Context, logger log.Logger, domain
...
+	Versions []moduleVersion `json:"versions"`
+}
+
+// moduleVersion represents a single version of a module.
+type moduleVersion struct {
+	Version string `json:"version"`
+}
+
+// GetLatestModuleVersion queries the Terraform module registry to list available versions for the given module
+// and returns the latest (first listed) version. This implements the "List Available Versions for a Specific Module"
```
yhakbar, line 313: This isn't entirely accurate. Please update this comment.

This returns the latest version based on semver sorting.

internal/tf/getter.go
```diff
@@ -116,6 +118,32 @@ func (tfrGetter *RegistryGetter) ClientMode(u *url.URL) (getter.ClientMode, erro
...
+		return versionList[0], nil
+	}
+
+	// When version is not specified, query the registry to find the latest version
+	latestVersion, err := GetLatestModuleVersion(ctx, tfrGetter.Logger, registryDomain, moduleRegistryBasePath, modulePath)
+	if err != nil {
+		return "", err
+	}
+
+	if tfrGetter.Logger != nil {
```
yhakbar, line 140: We don't generally do this. We assume that the logger is always set. Did you find this done elsewhere and copy this pattern? It's not a problem.

internal/tf/getter_test.go
```diff
@@ -155,3 +155,61 @@ func TestBuildRequestUrlRelativePath(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, "https://gruntwork.io/registry/modules/v1/tfr-project/terraform-aws-tfr/6.6.6/download", requestURL.String())
 }
+
+func TestGetLatestModuleVersion(t *testing.T) {
+	t.Parallel()
+
+	latestVersion, err := tf.GetLatestModuleVersion(t.Context(), logger.CreateLogger(), "registry.terraform.io", "/v1/modules/", "terraform-aws-modules/vpc/aws")
```
yhakbar, line 162: Please update these to use the `httptest.NewServer` pattern used elsewhere in the codebase. It prevents dependencies on external services from impacting the reliability of our test suite.

raman1236:
Friendly ping — all review feedback has been addressed and pushed. The PR makes version optional in `tfr://` registry URLs (defaulting to latest). Would appreciate another look when you have a moment. Thanks!

PR #5827, opened 2026-04-09
fix: Using cty.DynamicVal to avoid 'Unsupported Attribute' errors

### LINKED ISSUE #5811: `terragrunt hcl validate` fails with "Unsupported attribute" when dependency.outputs is referenced
## Describe the bug

`terragrunt hcl validate` fails with an _Unsupported attribute_ error when a `terragrunt.hcl` file references `dependency.<name>.outputs.*`.

During HCL validation, the `dependency` block appears to be resolved as an object with only a single `inputs` attribute, rather than the expected `outputs` attribute. This makes the command unusable for validating any module that uses dependency outputs.

When `mock_outputs` are defined for a `dependency` block, the validation succeeds but still warns on missing dependency `outputs`.

Noteworthy, the `dependency` block is surfacing as object with 1 attribute "inputs", while the Terragrunt docs explicitly state that `dependency.<name>.inputs` has been deprecated and removed, and that `dependency.<name>.outputs` is the only valid option.

## Steps To Reproduce

Create a `terragrunt.hcl` that defines at least one `dependency` block and references the output of another existing `terragrunt.hcl` with `dependency.<name>.outputs.<key>` in its inputs block:

```hcl
dependency "vpc" {
  config_path = "../vpc"
}

inputs = {
  vpc_id = dependency.vpc.outputs.vpc_id
}
```

Run `terragrunt hcl validate` or `terragrunt hcl validate --inputs`:

```
│ Error: Unsupported attribute
│
│   on /path/to/terragrunt.hcl line 68:
│   68:   vpc_id   = dependency.vpc.outputs.vpc_id
│     ├────────────────
│     │ dependency.vpc is object with 1 attribute "inputs"
│
│ This object does not have an attribute named "outputs".
```

Adding mocked values:

```hcl
dependency "vpc" {
  config_path = "../vpc"
  
  mock_outputs = {
    vpc_id = "mock-vpc-id"
  }
}

inputs = {
  vpc_id = dependency.vpc.outputs.vpc_id
}
```

Run `terragrunt hcl validate` or `terragrunt hcl validate --inputs`:

```
Config ../vpc/terragrunt.hcl is a dependency of ./terragrunt.hcl that has no outputs, but mock outputs provided and returning those in dependency output.
```

## Expected behavior

`terragrunt hcl validate` and `terragrunt hcl validate --inputs` should correctly resolve `dependency` blocks by treating the outputs attribute as available during static validation or by using the configured `mock_outputs`.

The command should behave analogously to how `terragrunt plan` handles dependency outputs when `mock_outputs` are defined.

## Versions

`terragrunt hcl validate` fails at least since 0.84.0
`terragrunt hcl validate --inputs` works for 0.97.2 and fails since 0.98.0
 
- Terragrunt version: 1.0.0, 0.98.0, 0.97.2, 0.84.0
- Terraform version: 1.14.7
- OS: macOS 26.4 (M2), Ubuntu 24.04 LTS, Amazon Linux 2023

yhakbar:
## Description

Fixes #5811.

`terragrunt hcl validate` previously failed with "Unsupported attribute" when a configuration referenced `dependency.<name>.outputs.<key>` without `mock_outputs`. 

During validation, output resolution is skipped, but the `outputs` attribute was never added to the dependency evaluation context, causing any output reference to error. The fix provides a dynamic placeholder for dependency outputs (and inputs) during validation so that attribute access evaluates to unknown rather than failing. Additionally, the dependency resolution pipeline is now more resilient during validation. Dependencies with unresolvable `config_path` values or nonexistent targets no longer cause the entire `dependency` namespace to disappear from the evaluation context.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **Bug Fixes**
  * Validation no longer errors for references to dependency outputs when mock outputs are absent; dependency attributes evaluate as unknown instead of failing.
  * Validation is more resilient: unresolved dependency config paths or missing targets no longer remove the dependency namespace.

* **Documentation**
  * Added v1.0.1 changelog entry documenting dependency validation improvements.

* **Tests**
  * Added fixtures covering dependency output and unresolvable-path validation scenarios.

### DIFF
--- a/pkg/config/dependency.go
+++ b/pkg/config/dependency.go
@@ -223,6 +223,13 @@ func decodeAndRetrieveOutputs(ctx context.Context, pctx *ParsingContext, l log.L
 		}
 
 		if !IsValidConfigPath(dep.ConfigPath) {
+			// During hcl validate, config_path may be unresolvable (e.g., references an
+			// unavailable local). Skip the invalid dependency instead of aborting — it will
+			// get a cty.DynamicVal placeholder in dependencyBlocksToCtyValue.
+			if pctx.SkipOutput {
+				continue
+			}
+
 			return nil, errors.New(DependencyInvalidConfigPathError{DependencyName: dep.Name})
 		}
 	}
@@ -279,6 +286,11 @@ func decodeDependencies(ctx context.Context, pctx *ParsingContext, l log.Logger,
 		}
 
 		if !IsValidConfigPath(dep.ConfigPath) {
+			if pctx.SkipOutput {
+				updatedDependencies.Dependencies = append(updatedDependencies.Dependencies, dep)
+				continue
+			}
+
 			return &updatedDependencies, errors.New(DependencyInvalidConfigPathError{DependencyName: dep.Name})
 		}
 
@@ -388,10 +400,19 @@ func checkForDependencyBlockCycles(ctx context.Context, pctx *ParsingContext, l 
 		}
 
 		if !IsValidConfigPath(dependency.ConfigPath) {
+			if pctx.SkipOutput {
+				continue
+			}
+
 			return errors.New(DependencyInvalidConfigPathError{DependencyName: dependency.Name})
 		}
 
 		dependencyPath := getCleanedTargetConfigPath(dependency.ConfigPath.AsString(), configPath)
+
+		// Skip cycle checking for nonexistent dependency targets — there is nothing to traverse.
+		if !util.FileExists(dependencyPath) {
+			continue
+		}
 
 		l, dependencyContext, err := pctx.WithDependencyConfigPath(l, dependencyPath)
 		if err != nil {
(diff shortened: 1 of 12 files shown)


pkg/config/dependency.go
```diff
@@ -515,10 +536,17 @@ func dependencyBlocksToCtyValue(traceCtx context.Context, pctx *ParsingContext,
 					lock.Unlock()
 
 					dependencyEncodingMap["outputs"] = *dependencyConfig.RenderedOutputs
+				} else if pctx.SkipOutput {
+					// During hcl validate, output resolution is skipped. Use cty.DynamicVal so that
+					// attribute access on dependency outputs (e.g. dependency.x.outputs.y) evaluates
+					// to unknown rather than producing an "Unsupported attribute" error.
+					dependencyEncodingMap["outputs"] = cty.DynamicVal
```
denis256, line 545: Silent dynamic val setting, I was thinking that it may be helpful to do a debug log to know that this happened

denis256 (dismissed):

pkg/config/dependency.go
yhakbar: Added

denis256 (approved):

PR #5828, opened 2026-04-09
fix: Fixing `get_original_terragrunt_dir()` interaction with dependencies

yhakbar:
## Description

During refactors in `v1.0.0-rc3`, a subtle regression was introduced where `OriginalTerragruntConfigPath` was no longer correctly reset when parsing a dependency from a parsing context of a dependent unit. This caused a bug in resolution of `get_original_terragrunt_dir()` where the value of `get_original_terragrunt_dir()` in the unit configuration of the dependency was set to the path of the dependent unit, rather than the dependency being parsed.

This fixes that regression.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Release Notes

* **Bug Fixes**
  * Fixed a regression where `get_original_terragrunt_dir()` returned incorrect directory references during dependency parsing, affecting proper resolution of sibling files in dependency configuration chains.

* **Tests**
  * Added regression test fixtures to verify correct directory path resolution behavior during dependency configuration parsing.

### DIFF
--- a/pkg/config/dependency_test.go
+++ b/pkg/config/dependency_test.go
@@ -188,6 +188,37 @@ dependency "enabled" {
 
 // TestDisabledDependencyWithEmptyConfigPath verifies that disabled dependencies
 // with empty config_path don't cause errors.
+// TestDependencyOriginalTerragruntDir verifies that when parsing a dependency's
+// config during cycle detection, get_original_terragrunt_dir() returns the
+// dependency's directory, not the caller's directory.
+//
+// Regression test: when unit-a depends on unit-b, and unit-b's config chain
+// calls get_original_terragrunt_dir(), it must resolve to unit-b's directory so
+// that paths constructed from it point to files that exist alongside unit-b.
+func TestDependencyOriginalTerragruntDir(t *testing.T) {
+	t.Parallel()
+
+	filename, err := filepath.Abs(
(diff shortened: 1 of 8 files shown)


pkg/config/dependency_test.go
```diff
@@ -188,6 +188,37 @@ dependency "enabled" {
 
 // TestDisabledDependencyWithEmptyConfigPath verifies that disabled dependencies
```
denis256, line 189: TestDisabledDependencyWithEmptyConfigPath should be on top of TestDisabledDependencyWithEmptyConfigPath() function

pkg/config/dependency_test.go
```diff
@@ -188,6 +188,37 @@ dependency "enabled" {
...
+		),
+	)
+	require.NoError(t, err)
+
+	ctx, pctx := newTestParsingContext(t, filename)
+	pctx.OriginalTerragruntConfigPath = filename
+	pctx.SkipOutput = true
+
+	_, err = config.ParseConfigFile(ctx, pctx, logger.CreateLogger(), filename, nil)
+	require.NoError(t, err)
```
denis256, line 219: Weak test, we should check for matched config, like `app_name = "myapp"` in locals

denis256 (approved):

PR #5886, opened 2026-04-13
fix(s3): fix shared_credentials_files HCL serialization for backend config

### LINKED ISSUE #4388: shared_credentials_file is deprecated and behaves differently than the new shared_credentials_files
## Describe the bug

Previously when using the  `shared_credentials_file` attribute which is now [deprecated ](https://developer.hashicorp.com/terraform/language/backend/s3#shared_credentials_file) terraform/terragrunt would read the credentials file and then use them to assume a role for use with the S3 state.  Migrating to use `shared_credentials_files`  no longer has this behavior and results in terragrunt complaining that it cannot find any AWS credentials upon trying to do a `terragrunt init` .  Note that using the deprecated `shared_credentials_file` still works just fine.

## Steps To Reproduce

- just a terragrunt init with the new backend configuration

```hcl
// old working version
terraform {
  backend "s3" {
    assume_role = {
      role_arn = "arn:aws:iam::123456789:role/role_name"
    }
    bucket                  = "some state bucket"
    dynamodb_table          = "some lock table"
    encrypt                 = true
    key                     = "someplace/terraform.tfstate"
    region                  = "us-east-1"
    shared_credentials_file = "/tmp/state_credentials"
  }
}

//new broken version
terraform {
  backend "s3" {
    assume_role = {
      role_arn = "arn:aws:iam::123456789:role/role_name"
    }
    bucket                   = "some state bucket"
    dynamodb_table           = "some lock table"
    encrypt                  = true
    key                      = "someplace/terraform.tfstate"
    region                   = "us-east-1"
    shared_credentials_files = ["/tmp/state_credentials"]
  }
}
```

## Expected behavior

Terragrunt uses the credentials from the shared credentials file to assume the specified role in the S3 state configuration

## Nice to haves

Image

## Versions

- Terragrunt version: v0.80.4
- OpenTofu/Terraform version: 1.12.1
- Environment details (Ubuntu 20.04, Windows 10, etc.): Ubuntu 22.04.5 LTS

## Additional context

Note that if you generate the terraform files with terragrunt and then instead of running `terragrunt init` you run `terraform init` then everything works just fine.  This is what lead me to believe that terragrunt is the one with the issue and not terraform

Rahul-Kumar-prog:
## Problem

When migrating from the deprecated `shared_credentials_file` to the new
`shared_credentials_files` attribute in the S3 backend, terragrunt fails
with:

  Error: Invalid expression
  on -backend-config=shared_credentials_files=/tmp/creds line 1:
  Expected the start of an expression, but found an invalid expression token.

This is because `shared_credentials_files` is a list type, and terragrunt
was serializing it incorrectly when building the `-backend-config` flags
passed to terraform/tofu.

## Root Cause

`GetTFInitArgs` was using a single `fmt.Sprintf("-backend-config=%s=%v")`
for all value types. This broke for:

- **Lists** like `shared_credentials_files` — needs one flag per element
- **Maps** like `assume_role` — needs HCL object syntax `{key="value"}`
- **String paths** like `/tmp/creds` — needs HCL quoting since `/` is 
  invalid in a bare HCL expression

## Fix

Introduced two helpers in a new `hclhelper` package:

- `FormatValueToSingleLineHcl(value any)` — correctly serializes any 
  Go value to a valid HCL expression string
- `WrapMapToSingleLineHcl(m map[string]any)` — serializes maps to HCL 
  object literal syntax `{key="value"}`
- `WrapListToSingleLineHcl(values []any)` — serializes lists to HCL 
  list syntax `["val1","val2"]`

`GetTFInitArgs` in `remote_state.go` now uses these helpers instead of 
raw `fmt.Sprintf`.

## Testing

Tested locally against a reproduction case with `shared_credentials_files`
and `assume_role` configured in the S3 backend.

Fixes #4388 

Screenshot 2026-04-13 at 11 35 15 AM

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **Refactor**
  * Enhanced backend configuration value serialization to reliably handle complex types (maps and lists) and produce compact single-line HCL representations.
  * Improved string value quoting and CLI argument rendering for backend configs to ensure consistent, predictable formatting.

### DIFF
--- a/internal/hclhelper/wrap.go
+++ b/internal/hclhelper/wrap.go
@@ -4,6 +4,7 @@ package hclhelper
 import (
 	"fmt"
 	"sort"
+	"strconv"
 	"strings"
 )
 
@@ -33,8 +34,7 @@ func WrapListToSingleLineHcl(values []any) string {
 func FormatValueToSingleLineHcl(value any) string {
 	switch v := value.(type) {
 	case string:
-		escapedValue := strings.ReplaceAll(v, `"`, `\"`)
-		return fmt.Sprintf(`"%s"`, escapedValue)
+		return strconv.Quote(v)
 	case map[string]any:
 		return WrapMapToSingleLineHcl(v)
 	case []any:
--- a/internal/hclhelper/wrap.go
+++ b/internal/hclhelper/wrap.go
@@ -11,7 +11,7 @@ import (
 func WrapMapToSingleLineHcl(m map[string]any) string {
 	var attributes = make([]string, 0, len(m))
 	for key, value := range m {
-		attributes = append(attributes, fmt.Sprintf(`%s=%s`, key, formatHclValue(value)))
+		attributes = append(attributes, fmt.Sprintf(`%s=%s`, key, FormatValueToSingleLineHcl(value)))
 	}
 
 	sort.Strings(attributes)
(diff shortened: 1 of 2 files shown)


yhakbar (approved):
Thanks!

PR #5888, opened 2026-04-13
fix: opts.WorkingDir needs symlinks resolved for later comparisons

tux3:
## Description

Terragrunt discovery could generate duplicate or invalid paths when the working dir path contained a symlink.

If the terragrunt working dir contains a symlink, then opts.WorkingDir would contain the absolute path to this directory, but discovery often needs to compose the working dir with canonical paths to discovered units, so we would mix canonical and non-canonical paths, which could give incorrect results.

When we have a symlink like /link/ -> /Users/user/link/, then the working dir might be for example /link/project/
But during discovery we often resolve symlinks to have canonical paths, which gives paths like /User/user/link/project/a/b/c

This looks like two different absolute paths when we do filepath.Rel(opts.WorkingDir, otherPath) so filepath.Rel will return:

`../../User/user/link/project/a/b/c`

Unfortunately this is subtly wrong! This path does not exist. That's because filePath.Rel() looks at the syntactic WorkingDir path which is /link/project/ and concludes it only needs ../../ to get an absolute path.
But in reality WorkingDir + ../../ is interpreted by the OS as:

`/User/user/link/project/../..` (the real canonical path)

Which resolves to `/User/user/` instead of `/`, because of the symlink. This PR resolve opts.WorkingDir after we make it an absolute path and before we clean it, so that all the symlinks go away and the rest of the code can safely take relative paths to it.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

Fixed path issues when Terragrunt's working directory has a symlink in its path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Working directory handling now resolves symbolic links so paths point to the underlying directory, improving consistency across filesystems.

* **Tests**
  * Added a test to verify working directory symlink resolution to prevent regressions.

### DIFF
--- a/internal/cli/commands/commands_internal_test.go
+++ b/internal/cli/commands/commands_internal_test.go
@@ -1,0 +1,58 @@
+package commands
+
+import (
+	"os"
+	"path/filepath"
+	"runtime"
+	"testing"
+
+	"github.com/urfave/cli/v2"
+
+	runcmd "github.com/gruntwork-io/terragrunt/internal/cli/commands/run"
+	"github.com/gruntwork-io/terragrunt/internal/clihelper"
+	"github.com/gruntwork-io/terragrunt/pkg/log"
+	"github.com/gruntwork-io/terragrunt/pkg/log/format"
+	"github.com/gruntwork-io/terragrunt/pkg/log/format/placeholders"
+	"github.com/gruntwork-io/terragrunt/pkg/options"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestInitialSetupResolvesWorkingDirSymlinks(t *testing.T) {
+	t.Parallel()
+
+	if runtime.GOOS == "windows" {
+		t.Skip("symlink creation requires elevated privileges on Windows")
+	}
+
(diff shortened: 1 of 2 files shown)


yhakbar:
Hey @tux3 ,

Do you mind adding some testing related to this behavior?

tux3:
@yhakbar Added a test. It's not trivial to test the full interaction between initial setup and what happens way later in discovery, so I just made a unit test that initialSetup() does the right thing.

internal/cli/commands/commands_internal_test.go
```diff
@@ -0,0 +1,58 @@
+package commands
```
yhakbar, line 1: Hey @tux3 ,

Unfortunately, this will violate our linter rules (we enable [testpackage](https://github.com/gruntwork-io/terragrunt/blob/main/.golangci.yml#L50)). You can export `InitialSetup` so that it's testable using a file in the `commands_test` package.

Also, you can add an integration test that runs `helpers.RunTerragruntCommand` on a symlinked directory. That does a black box test on the Terragrunt CLI and confirms that the behavior works as intended in symlinked directories.

I'm going to move this PR in-draft until that's resolved. Please feel free to tag us again when it's ready for review.

PR #5914, opened 2026-04-15
feat: Supporting streaming components to the catalog

yhakbar:
## Description

Makes it so that the catalog gets modules streamed in instead of waiting for them to be fully loaded before rendering the catalog TUI.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **New Features**
  * Redesigned terminal catalog with list, pager, and scaffold flows; modules stream into the UI as they are discovered and insert sorted case-insensitively
  * Concurrent discovery/loading with on-the-fly deduplication; loading indicator shown while discovery runs
  * Improved footer with scroll percentage and contextual key help; clearer discovery/no-sources/error views

* **Bug Fixes**
  * Nil-safety guards to avoid crashes when scaffold/view actions run without a loaded service

* **Tests**
  * Added end-to-end and view/model tests for streaming, sorting, UI states, and error scenarios


internal/cli/commands/catalog/tui/redesign/update.go
```diff
@@ -0,0 +1,274 @@
...
+			case scaffoldBtn:
+				if m.SVC == nil {
+					return m, nil
+				}
+
+				m.State = ScaffoldState
+
+				return m, scaffoldModuleCmd(m.logger, m, m.SVC, m.selectedModule)
+			case viewSourceBtn:
+				if m.selectedModule.URL() != "" {
```
denis256, line 158: Can be nil dereference in `m.selectedModule.URL()`?

denis256 (approved):

internal/cli/commands/catalog/tui/redesign/update.go
yhakbar: Na, but we could add some hardening in the future to guarantee it doesn't change.

PR #5917, opened 2026-04-16
fix: support custom host blocks in Provider Cache Server

### LINKED ISSUE #5916: Provider Cache Server doesn't support custom host blocks for private registries
## Description

Terragrunt Provider Cache Server doesn't handle providers from custom registries configured via `host` blocks in `.terraformrc`. While user's `host` blocks are preserved in the generated config (via `Clone()`), the cache server's proxy doesn't know how to route requests to these registries.

Related to #4719 — the maintainer asked for detailed reproduction steps.

## Context

We host custom Terraform providers in a private registry (Sonatype Nexus terraform-hosted). Terraform needs a `host` block in `.terraformrc`:

```hcl
host "registry.example.com" {
  services = {
    "providers.v1" = "https://registry.example.com/repository/terraform-hosted/v1/providers/"
  }
}
```

Provider source:
```hcl
required_providers {
  myprovider = {
    source  = "registry.example.com/example-org/my-provider"
    version = "1.0.0"
  }
}
```

### 1. Cache server only creates handlers for known registries

[`internal/providercache/providercache.go:151`](https://github.com/gruntwork-io/terragrunt/blob/main/internal/providercache/providercache.go#L151):
```go
providerHandlers, err := handlers.NewProviderHandlers(cliCfg, l, pcOpts.RegistryNames)
```

`pcOpts.RegistryNames` defaults to `["registry.terraform.io", "registry.opentofu.org"]`. Custom hosts from user config are never added.

### 2. Cache server only redirects known registries

[`internal/providercache/providercache.go:387-399`](https://github.com/gruntwork-io/terragrunt/blob/main/internal/providercache/providercache.go#L387-L399):
```go
for _, registryName := range filteredRegistryNames {
    providerInstallationIncludes = append(providerInstallationIncludes, registryName+"/*/*")
    // ...
    cfg.AddHost(registryName, map[string]string{
        "providers.v1": fmt.Sprintf("%s/%s/%s/", pc.ProviderController.URL(), cacheRequestID, registryName),
    })
}
```

Only `registry.terraform.io/*/*` gets added to `provider_installation.include`. The custom registry is handled by the `direct` fallback, which goes through the cache server proxy — but the proxy has no handler for it.

### 3. provider_installation is wiped during warm-up

[`internal/providercache/providercache.go:407`](https://github.com/gruntwork-io/terragrunt/blob/main/internal/providercache/providercache.go#L407):
```go
} else {
    cfg.ProviderInstallation = nil
}
```

During cache warm-up phase, all user's `provider_installation` methods are discarded.

### 1. Run terraform init directly (works)

```bash
export TF_CLI_CONFIG_FILE="/path/to/terraformrc.hcl"
cd .terragrunt-cache/<hash>/<hash>
terraform init
```
```
- Installing registry.example.com/example-org/my-provider v1.0.0...
- Installed registry.example.com/example-org/my-provider v1.0.0 ✓
```

### 2. Run with terragrunt (fails)

```bash
terragrunt plan
```
```
INFO Terragrunt Cache server is listening on 127.0.0.1:41483

Error: Failed to query available provider packages
  provider registry.terraform.io/example-org/my-provider was not found
```

### 3. Disable provider cache (works)

```bash
terragrunt plan --provider-cache=false
```
```
- Installed registry.example.com/example-org/my-provider v1.0.0 ✓
```

## Suggested Fix

In `createLocalCLIConfig()`, detect custom `host` blocks from the user's config and either:

1. **Add them to `RegistryNames`** so the cache server creates handlers and includes them in `provider_installation`
2. **Or exclude them from caching** by adding a `direct` block specifically for custom registry patterns, bypassing the cache server proxy

Option 2 is simpler — custom registries are usually private with few providers, so caching them is less important than not breaking them.

## Workaround

```bash
terragrunt plan --provider-cache=false
```

## Environment

- Terragrunt: v0.80.4
- Terraform: v1.14.3
- OS: Linux
- Private registry: Sonatype Nexus terraform-hosted

elkh510:
## Description

Fixes #5916

The Provider Cache Server now correctly handles providers from custom registries configured via `host` blocks in `.terraformrc` / `TF_CLI_CONFIG_FILE`.

Previously, providers from private registries (e.g. Sonatype Nexus terraform-hosted, Artifactory) would fail with `"does not offer a Terraform provider registry"` because the cache server didn't know how to route requests to them.

## Changes

1. **Add custom host names to `RegistryNames`** (`providercache.go:Init`) — so the cache server creates handlers for custom registries, not just `registry.terraform.io`

2. **Pre-populate discovery URL cache** (`providercache.go:Init`) — custom registries use `host` blocks instead of `.well-known/terraform.json` for service discovery. We read service URLs from the host block and populate the handler's cache, avoiding failed discovery lookups.

3. **Handle absolute URLs in `providers.v1`** (`direct_provider.go`) — host blocks may contain full URLs like `https://registry.example.com/repository/terraform-hosted/v1/providers/` instead of relative paths like `/v1/providers`. Added `resolveProviderURL()` helper that handles both cases.

4. **Prevent duplicate host blocks** (`config.go:AddHost`) — when a custom host exists in user config and the cache server adds its own redirect, `AddHost` now replaces the existing entry instead of appending a duplicate. Terraform uses the first matching block, so duplicates caused the cache server redirect to be ignored.

5. **Set auth tokens for custom registries** (`providercache.go:providerCacheEnvironment`) — `TF_TOKEN_*` env vars are now set for custom hosts too, so the cache server proxy can authenticate requests.

## How it works

Given user's `.terraformrc`:
```hcl
host "registry.example.com" {
  services = {
    "providers.v1" = "https://registry.example.com/repository/terraform-hosted/v1/providers/"
  }
}
```

Before this fix:
- Cache server only knew about `registry.terraform.io`
- Terraform requested provider from `registry.example.com` → cache server proxy had no handler → failed

After this fix:
- Cache server adds `registry.example.com` to its registry list
- Creates a handler with pre-populated discovery URLs from the host block
- Redirects `registry.example.com` through cache proxy like any other registry
- Provider downloads and caching work correctly

## Testing

- Unit tests pass for all modified packages: `providercache`, `handlers`, `cliconfig`
- Manual testing with Sonatype Nexus terraform-hosted repository confirmed working
- `go fmt` and `golangci-lint` pass

```
$ go test ./internal/providercache/... ./internal/tf/cache/handlers/... ./internal/tf/cliconfig/...
ok  github.com/gruntwork-io/terragrunt/internal/providercache       13.934s
ok  github.com/gruntwork-io/terragrunt/internal/tf/cache/handlers   0.011s
ok  github.com/gruntwork-io/terragrunt/internal/tf/cliconfig        0.005s
```

## Backwards Compatibility

No breaking changes. The fix only affects users who have custom `host` blocks in their config — for everyone else, behavior is identical.

## Summary by CodeRabbit

* **Improvements**
  * Custom registry hosts from local CLI config are now included during provider-cache initialization.
  * Discovery URLs for custom registries are pre-populated to reduce discovery requests.
  * Provider download and platform URL resolution are more robust for custom and absolute service endpoints.
  * Module discovery and local CLI config saving are more reliable and respect custom-host service definitions.
  * Environment-variable generation for cache authentication now covers custom registries.
  * CLI host entries merge service endpoints to avoid duplicate host blocks.

---

> [!NOTE]
> **Medium Risk**
> Changes provider-cache request routing and service discovery for registries, including auth token env var generation; mistakes could break `terraform init`/provider downloads for private registries.
> 
> **Overview**
> Enables the Provider Cache Server to correctly proxy and cache providers from *custom registries defined via CLI `host` blocks*, not just the default public registries.
> 
> During init, custom hostnames are added to the registry handler set and their known `providers.v1`/`modules.v1` service URLs are preloaded into the handlers’ discovery cache to skip `.well-known/terraform.json` lookups. Local CLI config generation is refactored to configure registry host redirects (including modules URL resolution) and to merge host entries instead of creating duplicates.
> 
> Direct provider URL construction now supports **absolute** `providers.v1` endpoints (e.g., Nexus/Artifactory base URLs), and `TF_TOKEN_*` env vars are also set for custom hosts. New unit tests cover host merging and provider URL resolution.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b7014e20996052a540474e37cb17b068fb5360b2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/providercache/providercache.go
```diff
@@ -411,7 +408,76 @@ func (pc *ProviderCache) createLocalCLIConfig(ctx context.Context, implementatio
...
+// collectCustomHosts collects custom host blocks from user config that are not already
+// in the registry list. For custom registries (e.g. Nexus, Artifactory), service URLs
+// are defined in the host block, so we skip .well-known/terraform.json discovery.
+// See: https://github.com/gruntwork-io/terragrunt/issues/5916
+func (pc *ProviderCache) collectCustomHosts(filteredRegistryNames *[]string) map[string]map[string]string {
+	customHosts := make(map[string]map[string]string, len(pc.cliCfg.Hosts))
+
+	for _, host := range pc.cliCfg.Hosts {
+		if !slices.Contains(*filteredRegistryNames, host.Name) {
+			*filteredRegistryNames = append(*filteredRegistryNames, host.Name)
```
yhakbar, line 423: This is a little weird. Why are we appending to the slice that's passed in? This should probably be it's own operation if important.

elkh510: You're right, that was confusing - collectCustomHosts was doing two things at once (collecting hosts + mutating the registry list via pointer). Refactored: the function now only collects and returns the map, and the caller explicitly appends custom host names to filteredRegistryNames after the call. Tests pass.

elkh510:
@yhakbar hi 
Could you please take a look

yhakbar:
Hey @elkh510 ,

Things are looking good. Please do the following before marking this PR ready to review again and tagging us.

## Blockers

Before we merge this PR, the following have to be done so that we can responsibly merge it:

- Export `resolveProviderURL` as `ResolveProviderURL` and add some testing to it.
- Same with `AddHost`. Already exported, but we need testing added for the new merge behavior.
- Delete `collectCustomHosts`. It's effectively dead code. `Init` directly appends custom hosts to `pcOpts.RegistryNames`, so once `createLocalCLIConfig` runs, `filteredRegistryNames` already has all the `cliCfg.Hosts` entries.

## Nice-to-haves

If you have time, these are things that would be nice to address, but if not, we can find time to address them later. 

- The name `SetRegistryURLCache` isn't super accurate. It should be `SetDiscoveryURLCache`. The backing field `registryURLCache` should also be renamed.
- Using both `platformURL` and `reqURLStr` as sources of truth can cause bugs. We should have one of them (probably `platformURL`) be the source of truth and call `.String()` on it if we need the string representation again.

elkh510:
@yhakbar hi

Thanks for the thorough review! All points addressed:

1. **`resolveProviderURL` -> `ResolveProviderURL`** - exported and now returns `*url.URL`.
   This also collapsed `reqURLStr` + `platformURL` in `GetPlatform` into a single variable -
   one call to `ResolveProviderURL`, then `.String()` for the HTTP request and the result
   directly for `ResolveRelativeReferences`. Added 7 unit tests in `resolve_provider_url_test.go`.

2. **`SetRegistryURLCache` -> `SetDiscoveryURLCache`**, field `registryURLCache` -> `discoveryURLCache`.

3. **Removed `collectCustomHosts`** - a note on what changed:
   The function had two responsibilities: appending custom host names to `filteredRegistryNames`
   (dead code - `Init` already calls `appendCustomHostRegistries`) and building a
   `map[string]map[string]string` passed to `resolveModulesURL` for the `modules.v1` URL lookup
   on custom hosts (to skip `.well-known` discovery).

   The name-appending part is removed entirely. The modules URL lookup is preserved -
   `resolveModulesURL` now reads `pc.cliCfg.Hosts` directly, and the `customHosts` parameter
   is dropped from both `configureRegistryHosts` and `resolveModulesURL`.

4. **`TestAddHost`** added with 4 sub-tests: new host appended, services merged on existing
   host, overlapping key overwritten, multiple distinct hosts.


Could you please take a look

internal/providercache/providercache.go
```diff
@@ -148,11 +152,19 @@ func (pc *ProviderCache) Init(l log.Logger, pcOpts *pcoptions.ProviderCacheOptio
 	providerService := services.NewProviderService(pcOpts.Dir, userProviderDir, cliCfg.CredentialsSource(), l, services.WithFS(pc.FS()))
 	proxyProviderHandler := handlers.NewProxyProviderHandler(l, cliCfg.CredentialsSource())
 
+	// Include custom host blocks from user config so the cache server handles them.
+	// See: https://github.com/gruntwork-io/terragrunt/issues/5916
+	pcOpts.RegistryNames = appendCustomHostRegistries(cliCfg.Hosts, pcOpts.RegistryNames)
```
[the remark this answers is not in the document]
elkh510: Good catch - this was a real bug. The root cause: `pcOpts.RegistryNames` served two purposes simultaneously - as the "standard registry config" for impl-based filtering, and as the full handler registry list (including custom hosts). Mutating it in `Init` poisoned `filterRegistriesByImplementation` for all downstream callers.

**Fix applied (two changes):**

**1. `Init`** - no longer mutates `pcOpts.RegistryNames`. Custom hosts go into a local variable used only for handler creation:

```go
// Before:
pcOpts.RegistryNames = appendCustomHostRegistries(cliCfg.Hosts, pcOpts.RegistryNames)
providerHandlers, err := handlers.NewProviderHandlers(cliCfg, l, pcOpts.RegistryNames)

// After:
registryNamesForHandlers := appendCustomHostRegistries(cliCfg.Hosts, pcOpts.RegistryNames)
providerHandlers, err := handlers.NewProviderHandlers(cliCfg, l, registryNamesForHandlers)
```

`pc.opts.RegistryNames` now stays clean and contains only the standard registries.


**2. `createLocalCLIConfig`** - custom hosts are appended *after* filtering, mirroring the pattern already used in `providerCacheEnvironment`:

```go
filteredRegistryNames := filterRegistriesByImplementation(pc.opts.RegistryNames, implementation)
filteredRegistryNames = appendCustomHostRegistries(pc.cliCfg.Hosts, filteredRegistryNames)
```

Unit tests added in `filter_registries_test.go` covering: defaults-only (Terraform/OpenTofu/Unknown), custom host + each impl type, and user-replaced registry list.

internal/providercache/export_test.go
```diff
@@ -0,0 +1,5 @@
+package providercache
+
+var FilterRegistriesByImplementation = filterRegistriesByImplementation
```
yhakbar: Hey @elkh510 ,

We generally don't do this in the codebase. Please just export the original variables if we need them accessible in tests.

elkh510: Done - removed `export_test.go` and exported the functions directly: `filterRegistriesByImplementation` → `FilterRegistriesByImplementation`, `appendCustomHostRegistries` → `AppendCustomHostRegistries`.

internal/tf/cache/handlers/direct_provider.go
```diff
@@ -81,3 +76,31 @@ func (handler *DirectProviderHandler) GetPlatform(ctx context.Context, provider
...
+	if strings.HasPrefix(providersV1, "http://") || strings.HasPrefix(providersV1, "https://") {
+		// Absolute URL from host block — append path parts directly
+		base := strings.TrimRight(providersV1, "/")
+		raw := base
+
+		if subPath != "" {
+			raw = base + "/" + subPath
+		}
+
+		u, _ := url.Parse(raw) //nolint:errcheck // base is a validated URL from discovery
```
yhakbar: This isn't exactly true. This can be supplied by users if they set a `host` block in their `.terraformrc` files. We should check this error. We've gone back and forth on this PR a couple times now. Please share if you're OK with a maintainer branching off this PR and running it to resolution.

internal/providercache/providercache.go
```diff
@@ -588,15 +642,47 @@ func convertToMultipleCommandsByPlatforms(args []string) [][]string {
...
 }
 
-// filterRegistriesByImplementation filters registry names based on the Terraform implementation being used.
+// AppendCustomHostRegistries adds custom host names from user config to the registry list
+// if they are not already present. This ensures the cache server handles them.
+// See: https://github.com/gruntwork-io/terragrunt/issues/5916
+func AppendCustomHostRegistries(hosts []cliconfig.ConfigHost, registryNames []string) []string {
+	for _, host := range hosts {
+		if !slices.Contains(registryNames, host.Name) {
+			registryNames = append(registryNames, host.Name)
```
yhakbar: This isn't safe to do. The backing array for `registryNames` can get mutated here. We should use `slices.Concat` instead.

yhakbar (changes requested):
Let me know if you would like help driving this PR to resolution.

elkh510:
> Let me know if you would like help driving this PR to resolution.

Happy to keep going - will push fixes for both issues shortly.

internal/tf/cache/handlers/direct_provider.go
elkh510: You're right - the nolint comment was incorrect. Will change `ResolveProviderURL` to return `(*url.URL, error)` and propagate the error from `url.Parse` through both callers (`GetVersions`, `GetPlatform`).

internal/providercache/providercache.go
elkh510: Good catch. Will use `slices.Concat` to avoid mutating the backing array:

yhakbar:
@elkh510

I've kicked off CI for this PR. Please don't edit it any further unless CI returns findings that need addressing. Do you mind creating a separate PR that introduces a changelog entry for this change? Changelog entries go [here](https://github.com/gruntwork-io/terragrunt/tree/main/docs/src/data/changelog/v1.0.4). Make sure to give yourself a shout-out! We appreciate your support as an external contributor.

elkh510:
@yhakbar 
Happy to hand it over! 
If you need any context on the changes along the way, just ping me - glad to help.

yhakbar (approved):

PR #5909, opened 2026-04-15
feat: handling of stacks dependencies in run queue / runner pool

denis256:
## Description

* stack dependencies usage in runner pool 
* addresses PR review feedback from #5816
* added integration and unit tests for the stack-dependencies experiment
* updated stack dependencies tests to use a common tests pattern

RFC: https://github.com/gruntwork-io/terragrunt/issues/5663

## TODOs

Read the [Gruntwork contribution guidelines](https://gruntwork.notion.site/Gruntwork-Coding-Methodology-02fdcd6e4b004e818553684760bf691e).

- [ ] I authored this code entirely myself
- [x] I am submitting code based on open source software (e.g. MIT, MPL-2.0, Apache)]
- [x] I am adding or upgrading a dependency or adapted code and confirm it has a compatible open source license
- [x] Update the docs.
- [x] Run the relevant tests successfully, including pre-commit checks.
- [x] Include release notes. If this PR is backward incompatible, include a migration guide.

## Release Notes (draft)

Added / Removed / Updated [X].

## Summary by CodeRabbit

* **New Features**
  * Expanded stack-dependencies: stack-to-stack, nested-unit, and stack-directory dependency patterns; clearer duplicate-stack-name reporting.

* **Bug Fixes**
  * Improved path normalization and dependency resolution for autoinclude and stack expansion.

* **Documentation**
  * Experiment roadmap updated; v1.0.2 changelog documents stack-dependencies examples and usage.

* **Tests**
  * Large addition of integration, unit, and E2E tests covering autoinclude generation, DAG extraction, duplicate-detection, error messages, and runner behaviors.

PR #5924, opened 2026-04-17
feat: allow units to depend on stacks

denis256:
## Description

Adds support for `stack.<name>.<nested_stack_name>.path` references and expands fuzz test coverage for the internal/hclparse package

Addressed PR feedback from https://github.com/gruntwork-io/terragrunt/pull/5909

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **New Features**
  * Added support for resolving dependencies through multi-level nested stack references, enabling deeper stack hierarchies in infrastructure configurations.

* **Tests**
  * Expanded test coverage with new fixtures and test cases validating nested stack dependency resolution and path references across multiple nesting levels.


internal/hclparse/fuzz_test.go
```diff
@@ -109,6 +114,164 @@ func FuzzPartialEval(f *testing.F) {
...
+
+	for _, seed := range seeds {
+		f.Add(seed)
+	}
+
+	f.Fuzz(func(t *testing.T, input string) {
+		fs := vfs.NewMemMapFS()
+
+		dir := "/fuzz/unit"
+		_ = fs.MkdirAll(dir, 0755)
```
yhakbar, line 184: Can we require that this succeeds? The dir name is deterministic, so we should be able to assert that this will always pass (or we have a bug we should fix).

internal/hclparse/fuzz_test.go
```diff
@@ -109,6 +114,164 @@ func FuzzPartialEval(f *testing.F) {
...
+		_ = vfs.WriteFile(fs, dir+"/terragrunt.autoinclude.hcl", []byte(input), 0644)
```
yhakbar, line 185: Can we require that this succeeds? The dir name is deterministic, so we should be able to assert that this will always pass (or we have a bug we should fix).

internal/hclparse/fuzz_test.go
```diff
@@ -109,6 +114,164 @@ func FuzzPartialEval(f *testing.F) {
...
+	f.Add("infra", "deep", "vpc", "db")
+	f.Add("network", "storage", "subnet", "bucket")
+	f.Add("a", "b", "c", "d")
+	f.Add("", "", "", "")
+	f.Add("path", "name", "source", "autoinclude")
+
+	f.Fuzz(func(t *testing.T, stackName, nestedStackName, unitName, nestedUnitName string) {
+		// Skip empty names: HCL labels can't be empty
+		if stackName == "" || nestedStackName == "" || unitName == "" || nestedUnitName == "" {
+			return
```
yhakbar, line 237: NIT: Let's use a `t.Skip` here so we can document why they're invalid with code instead of a comment.

internal/hclparse/parse_test.go
```diff
@@ -513,6 +513,90 @@ unit "app" {
...
+    dependency "deep" {
+      config_path = stack.infra.deep.path
+    }
+  }
+}
+`
+	parentStackFile := filepath.Join(parentStackDir, "terragrunt.stack.hcl")
+	require.NoError(t, os.WriteFile(parentStackFile, []byte(parentSrc), 0644))
+
+	result, err := hclparse.ParseStackFile(vfs.NewOSFS(), &hclparse.ParseStackFileInput{
```
yhakbar, line 582: Why are we using `NewOSFS` here instead of an in-memory FS? Using an in-memory one here would save us from having to create all the temporary directories, etc.

test/integration_stack_dependencies_test.go
yhakbar, line 1: Same question here about this being testable with an in-memory FS.

yhakbar (approved):
If the reason that we're using OSFS here in these tests is that we don't want to use in-memory filesystems in integration tests, I would ask if they need to be here in the integration test directory. I would imagine it's more useful to have a higher level test here where `terragrunt run --all plan`, etc. is run against the unit with stack dependencies. If this is the best way to test this, I am not opposed to it.

Remember to update the experiment and changelog docs with this update!

PR #5931, opened 2026-04-18
chore: Adding `vexec`

yhakbar:
## Description

Adds a `vexec` package similar to `vfs` to allow for an injectable command execution test double.

It basically provides a handle that lets us swap out actually spawning external processes for internal mocks.

Initially used purely for replacing usage of `exec` in `IsCommandExecutable` tests to keep scope small.

While I was in there, I also updated `WrappedBinary` to get memoized in test helpers. We were repeatedly calling external binaries to check which default we expected, but we can just call it once and memoize the result.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Release Notes

* **Refactor**
  * Improved internal command execution infrastructure for enhanced testability and maintainability.

* **Tests**
  * Expanded test coverage with deterministic, mockable command execution backends.

### DIFF
--- a/internal/vexec/vexec.go
+++ b/internal/vexec/vexec.go
@@ -1,0 +1,352 @@
+// Package vexec provides a virtual process-execution abstraction for testing and production use.
+// It wraps os/exec to provide a consistent, injectable interface for running external commands.
+package vexec
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"io"
+	"os"
+	"os/exec"
+	"strconv"
+)
+
+// Sentinel errors returned by the in-memory backend for misuse of the Cmd
+// lifecycle. All are exported so callers can match them with errors.Is.
+var (
+	// ErrAlreadyStarted is returned when Start is called on a Cmd that has
+	// already been started.
+	ErrAlreadyStarted = errors.New("vexec: command already started")
+	// ErrNotStarted is returned when Wait is called before Start.
+	ErrNotStarted = errors.New("vexec: Wait called before Start")
+	// ErrAlreadyWaited is returned when Wait is called more than once.
+	ErrAlreadyWaited = errors.New("vexec: Wait called more than once")
+	// ErrStdoutAlreadySet is returned from Output or CombinedOutput when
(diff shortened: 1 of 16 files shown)


internal/vexec/vexec.go
```diff
@@ -0,0 +1,352 @@
...
+
+	if ec, ok := errors.AsType[ExitCoder](err); ok {
+		return ec.ExitCode()
+	}
+
+	return -1
+}
+
+// Output runs name with args using e and returns its standard output.
+func Output(e Exec, ctx context.Context, name string, args ...string) ([]byte, error) {
```
denis256, line 89: I think the convention mentions to pass context first

test/helpers/package.go
```diff
@@ -841,33 +843,47 @@ func ValidateOutput(t *testing.T, outputs map[string]TerraformOutput, key string
...
-	value, found := os.LookupEnv("TG_TF_PATH")
-	if !found {
-		// if env variable is not defined, try to check through executing command
-		if util.IsCommandExecutable(context.Background(), TofuBinary, "-version") {
-			return TofuBinary
+// wrappedBinaryOnce memoizes the detected wrapped binary. The result is stable
+// for the lifetime of the process (env vars and installed binaries don't
+// change during a test run), so paying for detection once is enough.
+var (
+	wrappedBinaryOnce   sync.Once
```
denis256, line 850: Not sure about `sync.Once`, it caches the first detection

If any test changes TG_TF_PATH with t.Setenv() and calls WrappedBinary(ctx), the original detection result will be returned

denis256 (approved):

internal/vexec/vexec.go
yhakbar: I can adjust that afterwards.

test/helpers/package.go
yhakbar: I'm OK with that. This is a test utility to know what's going on in the test harness, not production code. We shouldn't be using `t.Setenv` calls to change the env var and testing that behavior.

PR #5932, opened 2026-04-18
feat: Splitting modules from templates

yhakbar:
## Description

Splits modules from templates in catalog discovery.

Adds support for a new `.terragrunt-catalog-ignore` file that users can commit to their catalogs to prevent discovery in particular directories and a `--ignore-file` flag in the `catalog` command that allows them to add to it with a local file. Only used when the `catalog-redesign` experiment is active.

The catalog now has tabs per component type that users can cycle between to interact with different component types. Sets up the catalog for support of units/stacks.

Tabs can be cycled through with the tab and shift + tab keys.

image

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **New Features**
  * Repository-wide catalog discovery now finds both modules and boilerplate templates anywhere in the tree.
  * Added layered ignore support via a repo-root ignore file plus an extra ignore file (flag/env), with gitignore-like patterns and negation.
  * List view split into All / Modules / Templates tabs; Tab/Shift+Tab cycles tabs with wraparound and preserves per-tab cursor and search state.
  * Each entry shows discovery source and optional release-version.

* **Documentation**
  * Updated docs and changelog with ignore-file syntax and the tabbed UI details.


internal/cli/commands/catalog/cli.go
```diff
@@ -15,10 +17,53 @@ import (
...
+			EnvVars:     tgPrefix.EnvVars(IgnoreFileFlagName),
+			Destination: &opts.CatalogIgnoreFile,
+			Usage:       "Path to an additional ignore file layered on top of a repo's .terragrunt-catalog-ignore during discovery.",
+			Action: func(_ context.Context, _ *clihelper.Context, value string) error {
+				if value == "" {
+					return nil
+				}
+
+				resolved := value
+				if !filepath.IsAbs(resolved) {
```
denis256, line 39: Not sure if we have tests for `--ignore-file` absolute path / RootWorkingDir fallback / both-empty

internal/services/catalog/ignore/ignore.go
```diff
@@ -0,0 +1,148 @@
...
+
+	"github.com/gobwas/glob"
+)
+
+// FileName is the fixed name looked up at the repo root.
+const FileName = ".terragrunt-catalog-ignore"
+
+type rule struct {
+	glob   glob.Glob
+	raw    string
```
denis256, line 25: Is Field read somewhere?

denis256 (approved):

internal/cli/commands/catalog/cli.go
yhakbar: Will add requisite testing.

internal/services/catalog/ignore/ignore.go
yhakbar: I don't think so, but I think it's a good idea to hold onto the raw text in addition to the glob.

Issue #1390, opened 2020-10-15
Constructing terragrunt cache is extremely slow for local modules


casparrolfe:
Our project is structured similar to this: https://github.com/tfutils/tfscaffold

For example:

```
├─ components/
│  └─ a-component/
│     └─ main.tf
└─ modules/
│  └─ a-module/
│  │  └─ main.tf
│  └─ b-module/
│     └─ main.tf
└─ config/
   └─ target/
      └─ terragrunt.hcl
```

In `components/a-component/main.tf`
```
module "aws_vpc_cni" {
  source = "../../modules/a-module"
}
```

In `config/target/terragrunt.hcl`
```
terraform {
  source = "../..//components/a-component"
}
```

All works nicely, but we've noticed that the entire project gets copied into the `terragrunt-cache` for local modules, so even though `b-module` is not used by `a-component` it is still copied into the cache. For smaller projects this is fine, but as a project grows things can get extremely slow on apply (seems to take minutes while it copies into the cache) and disk usage can quickly grow up to many gigabytes in size.

brikis98:
This is true, but I'm not sure what a good solution would be? Terragrunt copies more than just the module you're using due to the double-slash: e.g., `xxx//components/a-component` tells Terragrunt to copy the entire `xxx` folder so that any relative paths within `a-component` will work correctly. If you don't need relative paths, you can perhaps reduce the amount copied by removing double-slashes, but we don't usually recommend that, as it can lead to confusing errors...

casparrolfe:
Ah, I see what you mean about the relative paths. So if `component-a` uses `module-a` and `module-a` uses `module-b` then there is no guarantee that the path to `module-b` would be correct! Is it possible to calculate the local dependencies and file paths before the cache is copied and rewrite them to work within the cache?

We did have the double slash removed for a while, but it just confuses the IDE amongst other things!

brikis98:
> Is it possible to calculate the local dependencies and file paths before the cache is copied and rewrite them to work within the cache?

I suspect it's quite tricky. It would require parsing all your Terraform code and knowing all the places where a relative path could be. Examples:

- `source` URLs in `module { ... }` blocks. 
- Paths passed to the [`file` function](https://www.terraform.io/docs/configuration/functions/file.html).
- Paths passed to the [`local_file` resource](https://registry.terraform.io/providers/hashicorp/local/latest/docs/data-sources/file).
[...]

PR #5933, opened 2026-04-18
feat: Adding support for local paths in CAS

yhakbar:
## Description

Adds support for local paths in CAS.

Instead of falling back to the existing behavior for local paths, the CAS will now persist the contents of the source, then reproduce the contents from the content store.

This avoid unnecessary disk bloat for users leveraging sources on the local filesystem.

## TODOs

Read the [Gruntwork contribution guidelines](https://gruntwork.notion.site/Gruntwork-Coding-Methodology-02fdcd6e4b004e818553684760bf691e).

- [ ] I authored this code entirely myself
- [x] I am submitting code based on open source software (e.g. MIT, MPL-2.0, Apache)]
- [x] I am adding or upgrading a dependency or adapted code and confirm it has a compatible open source license
- [x] Update the docs.
- [x] Run the relevant tests successfully, including pre-commit checks.
- [x] Include release notes. If this PR is backward incompatible, include a migration guide.

## Release Notes (draft)

Added / Removed / Updated [X].

## Summary by CodeRabbit

* **New Features**
  * Support local filesystem paths as stack sources alongside remote Git URLs; local catalogs are copied to a temp area and rewritten to CAS-backed references.

* **Documentation**
  * Added docs and changelog entries describing local stack-source behavior, temp-directory processing, CAS content-addressing (SHA-256), and preservation of the original catalog.

* **Tests**
  * Added unit, race, and integration tests covering deterministic CAS hashes, non-mutating behavior, concurrency, and local-source processing.

PR #5939, opened 2026-04-20
optimize "Downloading Terraform configurations" step

y3ti:
## Description

In a large monorepo with hundreds of Terraform modules, the "Downloading Terraform configurations from ... into ./.terragrunt-cache/..." step gets progressively slower as more modules are initialized (`terragrunt init`). Each new `.terragrunt-cache` directory created by a previous `init` increases the time for *all subsequent* `init` runs across the entire repository — even for completely unrelated modules.

The slowdown is triggered any time a file changes anywhere in the package root directory, causing Terragrunt to re-copy the source tree.

### Environment

- Terragrunt v1.0.x
- Monorepo with **433 modules** sharing a common `terraform//` package root via double-slash syntax
- 10 `exclude_from_copy` patterns configured in root `terragrunt.hcl`

Root configuration:

```hcl
terraform {
  source = "${get_repo_root()}/terraform//${path_relative_to_include()}"

  exclude_from_copy = [
    "**/*.log",
    "**/*tfstate*",
    "**/.DS_Store",
    "**/.cache",
    "**/.terraform",
    "**/.terraformrc",
    "**/.terragrunt-cache",
    "**/dist",
    "**/node_modules",
    "**/tmp",
  ]
}
```

### Root Cause

The performance bottleneck is in [`CopyFolderContents`](https://github.com/gruntwork-io/terragrunt/blob/main/internal/util/file.go) which calls `expandGlobPath` for **each** `exclude_from_copy` pattern **before** any file copying begins.

`expandGlobPath` calls [`zglob.Glob()`](https://github.com/mattn/go-zglob), which internally uses `fastwalk` to walk the **entire** source tree for every pattern — including into `.terragrunt-cache` and `.terraform` directories that contain hundreds of megabytes of cached providers and modules. While the results are filtered out afterward (`strings.Contains(path, TerragruntCacheDir)`), the expensive directory walking has already happened.

```go
// internal/util/file.go – expandGlobPath (simplified)
func expandGlobPath(source, absoluteGlobPath string) ([]string, error) {
    // zglob.Glob walks the ENTIRE source tree to find matches,
    // including into .terragrunt-cache (GBs of data)
    absoluteExpandGlob, err := zglob.Glob(absoluteGlobPath)

    for _, path := range absoluteExpandGlob {
        // This filters AFTER walking — too late, the I/O already happened
        if strings.Contains(path, TerragruntCacheDir) {
            continue
        }
        // ...
        if IsDir(path) {
            // Recursive: triggers another full zglob.Glob walk
            expandGlobPath(source, path+"/*")
        }
    }
}
```

**With 10 exclude patterns, this means 10 full walks of the 2.5 GB tree before a single file is copied.** As more modules are initialized and their `.terragrunt-cache` directories grow, each walk takes longer — creating a progressively worsening feedback loop.

Additionally, `CopyFolderContentsWithFilter` uses `filepath.Glob(source + "/*")` which calls `lstat` on every entry in each directory, followed by separate `IsDir()` and `os.Lstat()` calls — resulting in 2-3 redundant stat syscalls per entry.

### Complexity

| | Before (current) | After (proposed) |
|---|---|---|
| **Exclude expansion** | `O(patterns × tree_size)` stat calls upfront | `O(files_copied × patterns × depth)` string matches, no upfront I/O |
| **Directory listing** | `filepath.Glob` → lstat per entry + `IsDir()` + `os.Lstat()` (2-3 stats/entry) | `os.ReadDir` → DirEntry.Type() from syscall (0 extra stats for non-symlinks) |

### Proposed Fix

1. **Lazy exclude matching**: Instead of pre-expanding exclude globs into a flat path list, evaluate patterns on-the-fly during the copy walk using `doublestar.Match` (already an indirect dependency). Each path and its ancestors are checked against the raw patterns. Since `CopyFolderContentsWithFilter` already prunes excluded directories (skips recursion when the filter returns false), the lazy approach never enters the expensive `.terragrunt-cache` or `.terraform` directories.

2. **Replace `filepath.Glob` with `os.ReadDir`**: Since Go 1.16, `os.ReadDir` returns lightweight `DirEntry` values whose `Type()` comes directly from the directory-listing syscall (`getdents64` on Linux, `getdirentries` on macOS) without a separate `lstat` per entry. The earlier concern noted in the code comment ("os.ReadDir calls lstat on all files") no longer applies.

3. **Bonus bug fix**: The current `pathContainsPrefix` uses `strings.HasPrefix` which can match unrelated paths sharing a common prefix (e.g., pattern `module/region2` would incorrectly exclude `module/region2more`). The proposed `doublestar.Match` approach matches on exact path segments.

### Related

- #1390 — Cache construction extremely slow for local modules (closed, workaround: `TF_PLUGIN_CACHE_DIR`)

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

- Updated `CopyFolderContents` to evaluate `exclude_from_copy` glob patterns lazily during the copy walk instead of pre-expanding them via full tree traversals, reducing "Downloading Terraform configurations" time from
  `O(patterns × tree_size)` to `O(files_copied × patterns × depth)` — particularly impactful for large monorepos with accumulated `.terragrunt-cache` directories.   

## Summary by CodeRabbit

* **Refactor**
  * File-copying now evaluates exclude patterns lazily per path instead of pre-expanding lists, simplifying logic.

* **Bug Fix**
  * Improved handling of symbolic links and directory detection during recursive copies to avoid incorrect traversal or exclusions.

* **Performance**
  * Faster behavior when excluding large/deep directory trees during file operations.

* **Tests**
  * Added unit tests and a benchmark covering exclude matching, edge cases, and deep-exclude performance.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -157,7 +157,7 @@ require (
 	github.com/bahlo/generic-list-go v0.2.0 // indirect
 	github.com/bgentry/go-netrc v0.0.0-20140422174119-9fd32a8b3d3d // indirect
 	github.com/blang/semver v3.5.1+incompatible // indirect
-	github.com/bmatcuk/doublestar v1.3.4 // indirect
+	github.com/bmatcuk/doublestar v1.3.4
 	github.com/buger/jsonparser v1.1.2 // indirect
 	github.com/cenkalti/backoff/v4 v4.3.0 // indirect
 	github.com/cenkalti/backoff/v5 v5.0.3 // indirect
--- a/internal/util/file.go
+++ b/internal/util/file.go
@@ -310,7 +310,7 @@ func PathOrAncestorMatchesAny(relativePath string, patterns []string) bool {
 	ancestor := relativePath
 	for {
 		ancestor = path.Dir(ancestor)
-		if ancestor == "." || ancestor == "" {
+		if ancestor == "." || ancestor == "" || ancestor == "/" {
 			return false
 		}
 
@@ -384,6 +384,15 @@ func CopyFolderContents(
 		}
 
 		includeExpandedGlobs = append(includeExpandedGlobs, expandGlob...)
+	}
+
+	// Validate exclude patterns early so malformed globs (e.g. unbalanced
+	// brackets) surface as configuration errors rather than silently
+	// matching nothing during the walk.
+	for _, pattern := range excludeFromCopy {
+		if _, err := doublestar.Match(pattern, "."); err != nil {
+			return errors.Errorf("invalid exclude_from_copy pattern %q: %w", pattern, err)
+		}
 	}
 
 	// Exclude patterns are evaluated lazily during the copy walk using
--- a/internal/util/file.go
+++ b/internal/util/file.go
@@ -10,11 +10,13 @@ import (
 	"io/fs"
 	"net/url"
 	"os"
+	"path"
 	"path/filepath"
 	"regexp"
 	"strings"
 	"syscall"
 
+	"github.com/bmatcuk/doublestar"
 	urlhelper "github.com/hashicorp/go-getter/helper/url"
 
 	"github.com/gruntwork-io/terragrunt/internal/errors"
@@ -283,6 +285,43 @@ func pathContainsPrefix(path string, prefixes []string) bool {
 	return false
 }
 
+// PathOrAncestorMatchesAny reports whether relativePath or any of its ancestor
+// directories matches at least one of the given glob patterns. Both patterns
+// and relativePath must use forward slashes.
+//
+// This enables lazy exclude evaluation: instead of pre-expanding every exclude
+// glob into a flat path list (which requires walking the entire source tree
+// once per pattern), we evaluate patterns on-the-fly during the copy walk.
+func PathOrAncestorMatchesAny(relativePath string, patterns []string) bool {
+	if len(patterns) == 0 {
+		return false
+	}
+
+	// Check the full path itself.
+	for _, p := range patterns {
+		if matched, _ := doublestar.Match(p, relativePath); matched {
+			return true
+		}
+	}
+
+	// Walk up the directory tree checking each ancestor. This propagates
+	// directory-level matches to their descendants (e.g. pattern **/.terraform
+	// excludes everything under any .terraform directory).
+	ancestor := relativePath
+	for {
+		ancestor = path.Dir(ancestor)
+		if ancestor == "." || ancestor == "" {
+			return false
+		}
+
+		for _, p := range patterns {
+			if matched, _ := doublestar.Match(p, ancestor); matched {
+				return true
+			}
+		}
+	}
+}
+
 // Takes apbsolute glob path and returns an array of expanded relative paths
 func expandGlobPath(source, absoluteGlobPath string) ([]string, error) {
 	includeExpandedGlobs := []string{}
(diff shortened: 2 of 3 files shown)


yhakbar:
Thanks for opening this, @y3ti .

I'd like to close this PR in favor of #5966, which takes a different approach by pre-compiling the include and exclude globs into a single brace alternation expression each, then does a single walk. Let me know if you have any feedback on that approach.

I cut a [test build](https://docs.terragrunt.com/getting-started/install/#install-tip-or-test-builds) that you can install with the following if you want to try it out:

```bash
curl -sSfL --proto '=https' --tlsv1.2 https://terragrunt.com/install | bash -s -- --test --commit '21f25cd5daef8bac05bd02a6926c34cc63e29c49'
```

Note that because the glob grammar changes with the switch from zglob to gobwas here, this will have to be supported as an [opt-in strict control](https://github.com/gruntwork-io/terragrunt/pull/5966/changes#diff-94b8336f016b4e457c85972df5c269c4bcccd5322cc25674efa4ae06d5e54e93) for the duration of 1.0.

If you try it out in your environment, please share any changes to performance you see.

y3ti:
@yhakbar sounds good. Looking forward for the next release! :)

yhakbar:
@y3ti ,

Please do let me know if you try out the test build or have any feedback on the implementation. I want to make sure it properly addresses the performance issues you're seeing in production.

y3ti:
@yhakbar Your pull request does not fix my issue. Here is my directory structure and directory sizes when I do not have any `.terragrunt-cache` directory (not initialized):

```
/
/root.hcl
/dev/      - 1.7 MB in total
  app1/    - 32 KB
  app2/    - 44 KB
  ...
```

If I run `terragrunt init` then it's very fast (downloading terraform confgiuration takes ~1s):

```
$ terragrunt init
...
17:00:01.453 INFO   Downloading Terraform configurations from ../.. into ./.terragrunt-cache/UbHBFcQED189UF3j4Cq0jMlIrhg/uGBjRfdHtbslnjA_xJMiLRseE_8
17:00:02.014 STDOUT terraform: Initializing the backend...
...
```

If I initialize all units then it become quite heavy:

```
/
/root.hcl
/dev/      - 8.6GB in total
  app1/
    .terragrunt-cache/ 159 MB
  app2/
    .terragrunt-cache/ 161 MB
  ...
```

Then if I run `terragrunt init` again then downloading terragrunt configuration takes 75 seconds:

```
$ rm -rf .terragrunt-cache
$ terragrunt init
...
17:13:01.453 INFO   Downloading Terraform configurations from ../.. into ./.terragrunt-cache/UbHBFcQED189UF3j4Cq0jMlIrhg/uGBjRfdHtbslnjA_xJMiLRseE_8
17:14:16.014 STDOUT terraform: Initializing the backend...
...
```

If I initialize all environments (dev, prod, staging, qa, etc.), it takes 3–4 minutes, so terragrunt units/modules than the it takes more and more. 
I am running it on a MacBook M3 Pro. 

PS. I found a workaround: I set `TG_DOWNLOAD_DIR="~/.terraform.d/download-cache/"`, which moves all `.terragrunt-cache` directories outside my git repository.

yhakbar:
@y3ti 

Are you leveraging the caching capabilities available in Terragrunt?
https://docs.terragrunt.com/features/caching/

y3ti:
I cache providers. I use cloudposse terraform modules so I have a lot of dependencies. My code (without external terraform module like cloudposse and their dependencies) is 33 MB in total (all services, all my modules)

yhakbar:
Are you using the [CAS](https://docs.terragrunt.com/features/caching/cas/)?

y3ti:
No, I don't use CAS. We have monorepo which look like this (simplified):

```
/root.hcl
/dev/
  app1/
    backend.tf       (autogenerated)
    main.tf
    terragrunt.hcl
    variables.tf
    versions.tf
  app2/
    ...
/prod/
  app1/
    ...
```

`/root.hcl`:

```
generate "backend.tf" {
  ...
}

remote_state {
  backend = "s3"
  ...
}

terraform {
  source = "${get_repo_root()}//${path_relative_to_include()}"
  
  exclude_from_copy = [
    "**/*.log",
    "**/*tfstate*", # A local state file may conflict with the S3 remote backend
    "**/.DS_Store",
    "**/.cache",
    "**/.terraform",        # Generated by Terraform
    "**/.terraformrc",      # Generated by Terragrunt's provider cache
    "**/.terragrunt-cache", # Prevent copying pre-existing cache dirs from the package root into the module's own cache
    "**/dist",
    "**/node_modules",
    "**/tmp",
  ]
  ...
}    
```

`/dev/app1/terragrunt.hcl`:
```
include "root" {
  path = find_in_parent_folders("root.hcl")
}

include "dependency" {
  ...
}

inputs = {
  ...
}

# there is no terraform {} block here
```

`/dev/app1/main.tf`:

```
# ALB + ASG + IAM etc.
module "generic_service" {
    source = "../../shared/generic_service"

    name = "app1"

    ...
    subnet_ids = var.private_subnet_ids
    vpc_id     = var.vpc_id
}

# most services doesn't need postgres but app1 is unique 
module "rds" {
  source  = "cloudposse/rds/aws"
  version = "1.1.3"
  ...  
}

module "rds_proxy" {
  source  = "cloudposse/rds-db-proxy/aws"
  ...
} 

resource "aws_single_custom_resource" "name" {
  ...
}

... 
```

... so our configuration might be a bit different and not-standard. My understanding is that CAS won't work in this case.

yhakbar:
It should help soon:
https://github.com/gruntwork-io/terragrunt/pull/5933

PR #5950, opened 2026-04-22
fix: isolate per-unit feature defaults in run --all

### LINKED ISSUE #5810: `run --all` can leak same-named `feature` defaults across sibling units
## Describe the bug

When two sibling Terragrunt units define the same `feature` name with different defaults, `terragrunt run --all` can resolve `feature.<name>.value` incorrectly for one of the units.

In the minimal repro below:

- `target-service` defines `feature "toggle" { default = true }`
- `peer-service` defines `feature "toggle" { default = false }`

Running `terragrunt plan` directly in `target-service` works correctly.

Running `terragrunt run --all plan` from the parent directory does not: `target-service` ends up with `effective_toggle = false` even though its own feature default is `true`.

Explicit CLI feature overrides behave consistently, so the issue appears specific to implicit per-unit feature default resolution during multi-unit execution.

## Reproducing bugs

This report includes a minimal local-only fixture with no providers and no cloud APIs.

The fixture has:

- two sibling Terragrunt units under one parent directory
- the same feature name in both units
- different defaults for that feature in each unit
- a trivial Terraform module that only echoes the input values as outputs

The issue reproduces with `terragrunt run --all` and does not reproduce when planning the target unit directly.

## Steps To Reproduce

Fixture layout:

```text
.
├── catalog
│   └── modules
│       └── echo-toggle
│           ├── main.tf
│           ├── outputs.tf
│           └── variables.tf
└── live
    └── bare
        ├── peer-service
        │   └── terragrunt.hcl
        └── target-service
            └── terragrunt.hcl
```

`live/bare/target-service/terragrunt.hcl`

```hcl
feature "toggle" {
  default = true
}

terraform {
  source = "${find_in_parent_folders("catalog/modules")}//echo-toggle"
}

inputs = {
  effective_toggle = feature.toggle.value
  raw_toggle       = true
}
```

`live/bare/peer-service/terragrunt.hcl`

```hcl
feature "toggle" {
  default = false
}

terraform {
  source = "${find_in_parent_folders("catalog/modules")}//echo-toggle"
}

inputs = {
  effective_toggle = feature.toggle.value
  raw_toggle       = false
}
```

`catalog/modules/echo-toggle/main.tf`
```hcl
terraform {
  required_version = ">= 1.6.0"
}
```

`catalog/modules/echo-toggle/variables.tf`

```hcl
variable "effective_toggle" {
  type = bool
}

variable "raw_toggle" {
  type = bool
}
```

`catalog/modules/echo-toggle/outputs.tf`

```hcl
output "effective_toggle" {
  value = var.effective_toggle
}

output "raw_toggle" {
  value = var.raw_toggle
}
```

Run the baseline command from `live/bare/target-service`:

```bash
terragrunt plan --non-interactive --no-color --inputs-debug
```

Observed baseline output:

```text
Changes to Outputs:
  + effective_toggle = true
  + raw_toggle       = true
```

Then run the failing command from `live/bare`:

```bash
TF_CLI_ARGS_plan='-lock=false' terragrunt run --all \
  --non-interactive \
  --no-color \
  --parallelism 1 \
  --inputs-debug \
  plan
```

Observed multi-unit output:

```text
peer-service:
  + effective_toggle = false
  + raw_toggle       = false

target-service:
  + effective_toggle = false
  + raw_toggle       = true
```

The contradiction is also visible in the generated debug inputs for `target-service`:

```json
{
  "effective_toggle": false,
  "raw_toggle": true
}
```

Control case: explicit CLI feature override `true`

```bash
TF_CLI_ARGS_plan='-lock=false' terragrunt run --all \
  --non-interactive \
  --no-color \
  --parallelism 1 \
  --inputs-debug \
  --feature toggle=true \
  plan
```

Observed:

```text
peer-service:
  + effective_toggle = true
  + raw_toggle       = false

target-service:
  + effective_toggle = true
  + raw_toggle       = true
```

Control case: explicit CLI feature override `false`

```bash
TF_CLI_ARGS_plan='-lock=false' terragrunt run --all \
  --non-interactive \
  --no-color \
  --parallelism 1 \
  --inputs-debug \
  --feature toggle=false \
  plan
```

Observed:

```text
peer-service:
  + effective_toggle = false
  + raw_toggle       = false

target-service:
  + effective_toggle = false
  + raw_toggle       = true
```

This suggests the bug is not in explicit CLI feature overrides. It appears to be in per-unit feature default isolation during `run --all`.

## Expected behavior

Each unit should resolve `feature.toggle.value` independently from:

- its own `feature "toggle"` block
- its own feature default
- any explicit CLI `--feature toggle=...` override

Adding a sibling unit with the same feature name should not change the resolved feature value for the target unit.

For this repro, `target-service` should keep `effective_toggle = true` during `terragrunt run --all` when no explicit `--feature toggle=...` flag is provided.

## Must haves

## Nice to haves

- [x] Terminal output
- [ ] Screenshots

## Versions

- Terragrunt version: `v1.0.0`/`v0.98.0`
- OpenTofu/Terraform version: Terraform `v1.14.8`
- Environment details (Ubuntu 20.04, Windows 10, etc.): macOS/Ubuntu 24.04

## Additional context

- This repro uses no providers and no cloud APIs.
- The issue appears to be about feature state isolation across units during multi-unit execution.

lawrence3699:
## Description

Fixes #5810.

`run --all` was cloning `TerragruntOptions` with a shared `FeatureFlags` map, so the first unit that populated a same-named feature default could leak that value into sibling units.

This change deep-copies `FeatureFlags` in `TerragruntOptions.Clone()` so each unit resolves feature defaults independently.

I also added a regression fixture that reproduces the issue with `--inputs-debug` and a fake `--tf-path`: direct `plan` in `target-service` keeps `effective_toggle = true`, and `run --all ... plan` now preserves that value instead of inheriting `peer-service`'s `false` default.

## TODOs

- [ ] Update the docs.

## Release Notes (draft)

Fixed `run --all` leaking one unit's feature default into sibling units with the same feature name.

### Migration Guide

No migration guide required.

## Summary by CodeRabbit

* **Tests**
  * Added integration test for feature flag functionality with isolated defaults, validating toggle behavior across multiple services.
  * Introduced test fixtures and helper utilities to verify feature flags maintain appropriate values in distributed service configurations.

### DIFF
--- a/pkg/options/options.go
+++ b/pkg/options/options.go
@@ -403,6 +403,15 @@ func (opts *TerragruntOptions) OptionsFromContext(ctx context.Context) *Terragru
 func (opts *TerragruntOptions) Clone() *TerragruntOptions {
 	newOpts := cloner.Clone(opts)
 
+	if opts.FeatureFlags != nil {
+		clonedFlags := xsync.NewMapOf[string, string]()
+		opts.FeatureFlags.Range(func(key, value string) bool {
+			clonedFlags.Store(key, value)
+			return true
+		})
+		newOpts.FeatureFlags = clonedFlags
+	}
+
 	return newOpts
 }
 
--- a/test/integration_feature_flags_test.go
+++ b/test/integration_feature_flags_test.go
@@ -3,6 +3,7 @@ package test_test
 import (
 	"bytes"
 	"encoding/json"
+	"os"
 	"path/filepath"
 	"testing"
 
@@ -16,6 +17,7 @@ const (
 	testSimpleFlag     = "fixtures/feature-flags/simple-flag"
 	testIncludeFlag    = "fixtures/feature-flags/include-flag"
 	testRunAllFlag     = "fixtures/feature-flags/run-all"
+	testRunAllIsolated = "fixtures/feature-flags/run-all-isolated-defaults"
 	testErrorEmptyFlag = "fixtures/feature-flags/error-empty-flag"
 )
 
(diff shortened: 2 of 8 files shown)


pkg/options/options.go
```diff
@@ -403,6 +403,15 @@ func (opts *TerragruntOptions) OptionsFromContext(ctx context.Context) *Terragru
 func (opts *TerragruntOptions) Clone() *TerragruntOptions {
 	newOpts := cloner.Clone(opts)
 
+	if opts.FeatureFlags != nil {
```
yhakbar, line 406: I think we can have a better solution than this. This feels like a band-aid.

We can resolve that in a follow-up PR.

test/integration_feature_flags_test.go
```diff
@@ -153,3 +195,15 @@ func validateOutputsMap(t *testing.T, rootPath string, expected map[string]any)
...
 }
+
+func readDebugInputs(t *testing.T, rootPath string) map[string]any {
+	t.Helper()
+
+	bytes, err := os.ReadFile(filepath.Join(rootPath, helpers.TerragruntDebugFile))
+	require.NoError(t, err)
+
+	outputs := map[string]any{}
+	require.NoError(t, json.Unmarshal(bytes, &outputs))
```
yhakbar, line 206: Copilot makes a decent point here. Can be addressed in a follow-up.

yhakbar:
Thanks for opening this PR @lawrence3699 !

I'm seeing lint failures in CI. Do you mind addressing this? Make sure to run the linter with `make run-lint` to confirm there's nothing else that needs fixing.

```
pkg/options/options.go:408:3: missing whitespace above this line (no shared variables above expr) (wsl_v5)
		opts.FeatureFlags.Range(func(key, value string) bool {
		^
1 issues:
* wsl_v5: 1
make: *** [Makefile:44: run-lint-incremental] Error 1
```

Moving in-draft until it's resolved.

lawrence3699:
Closing this draft because CI/lint needs follow-up and I am not actively addressing it. I do not want to leave maintainers with an unfinished draft.

PR #5962, opened 2026-04-22
fix: fixing concurrent-write race in stack generate

denis256:
## Description

* Fixed concurrent write conflict in `terragrunt stack generate`

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a race during stack generation that could cause conflicting or duplicate writes by deduplicating stack references and serializing concurrent runs with an exclusive on-disk lock.
* **Documentation**
  * Added a v1.0.3 changelog entry describing the issue and mitigations.
* **Tests**
  * Added integration and unit tests validating concurrent stack generation and that each target is written only once.


internal/stacks/generate/generate.go
```diff
@@ -22,6 +26,34 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+const (
+	// stackGenerateLockDir is the per-working-directory subdirectory that
+	// holds the stack-generate lockfile. Placed inside `.terragrunt-stack/`
+	// so it's already covered by user .gitignore rules and cleaned up by
+	// `terragrunt stack clean`.
+	stackGenerateLockDir = ".terragrunt-stack"
```
yhakbar: We have a problem here when users use `no_dot_terragrunt_stack`

internal/stacks/generate/generate.go
```diff
@@ -50,6 +82,28 @@ func GenerateStacks(
...
+	// into the same on-disk target tree — only a filesystem-level lock
+	// can. Uses the same vfs.Lock primitive as cas.Store.AcquireLock
+	// (internal/cas/store.go).
+	lockPath, unlocker, err := acquireGenerateLock(ctx, l, vfs.NewOSFS(), opts.WorkingDir)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		if unlockErr := unlocker.Unlock(); unlockErr != nil {
```
yhakbar: I think we need to put the lock in a predictable temporary directory or something. Otherwise, users will have pollution in their worktrees from `.lock` files.

internal/stacks/generate/generate.go
```diff
@@ -605,3 +703,83 @@ func stackTypeFilter() filter.Filters {
...
+	// Validate workingDir before MkdirAll. Without this check, a typo in
+	// --working-dir would be silently created by MkdirAll (which creates
+	// all missing parents) and discovery would then report "No stack files
+	// found" as a successful no-op. We want typos to fail loudly. Routed
+	// through vfs.FS so the check benefits from the same abstraction as
+	// the lock primitive.
+	info, err := fsys.Stat(workingDir)
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			return lockPath, nil, errors.Errorf("working directory %s does not exist", workingDir)
```
yhakbar: This should be a typed error.

internal/stacks/generate/generate.go
```diff
@@ -605,3 +703,83 @@ func stackTypeFilter() filter.Filters {
...
+		return lockPath, nil, errors.Errorf("working directory %s is not a directory", workingDir)
+	}
+
+	if err := fsys.MkdirAll(lockDir, stackGenerateLockDirPerm); err != nil {
+		return lockPath, nil, errors.Errorf("create stack-generate lock dir %s: %w", lockDir, err)
+	}
+
+	l.Debugf("acquiring stack-generate lock %s", lockPath)
+
+	for {
```
yhakbar: This loop should be bounded.

internal/stacks/generate/generate.go
```diff
@@ -22,14 +24,87 @@ import (
...
+func NewStackGenerator() *StackGenerator {
+	return &StackGenerator{
+		lockManager:          util.NewKeyLocks(),
+		onGenerateDispatched: &hookManager{},
+	}
+}
+
+// defaultStackGenerator is a package-level instance used by the backward-compatible
+// GenerateStacks package-level function.
+var defaultStackGenerator = NewStackGenerator()
```
yhakbar: This is odd. Are we doing this so that we can overwrite it in tests? We should pass it in as a dependency if so.

internal/stacks/generate/generate.go
```diff
@@ -22,14 +24,87 @@ import (
...
+
+// GenerateStacks generates the stack files using topological ordering to prevent race conditions.
+// Uses the default package-level generator.
+func GenerateStacks(
+	ctx context.Context,
+	l log.Logger,
+	opts *options.TerragruntOptions,
+	wts *worktrees.Worktrees,
+) error {
+	return defaultStackGenerator.GenerateStacks(ctx, l, opts, wts)
```
yhakbar: This feels really weird.

internal/stacks/generate/generate.go
```diff
@@ -22,14 +24,87 @@ import (
...
+	defaultStackGenerator.onGenerateDispatched.register(workingDir, fn)
+}
+
+// UnregisterOnGenerateHook removes a previously registered generation hook.
+func UnregisterOnGenerateHook(workingDir string) {
+	defaultStackGenerator.onGenerateDispatched.unregister(workingDir)
+}
+
+type hookManager struct {
+	hooks sync.Map
```
yhakbar: Can we use a normal map and mutex instead, or is it better to use a sync.Map here?

internal/stacks/generate/generate.go
```diff
@@ -605,3 +748,56 @@ func stackTypeFilter() filter.Filters {
 
 	return filter.Filters{filter.NewFilter(attrExpr, attrExpr.String())}
 }
+
+// canonicalStackFilePath returns the resolved absolute path to the stack file
+// inside the given component's directory, relative to workingDir. Resolves
+// symbolic links and collapses string-aliased discovery results (relative vs.
+// absolute forms of the same path) to a single key so the dedup map in
+// GenerateStacks catches duplicates.
+func canonicalStackFilePath(c component.Component, workingDir string) (string, error) {
```
yhakbar: Why do we have this wrapper that just calls `canonicalIdentity`?

internal/stacks/generate/generate.go
```diff
@@ -605,3 +748,56 @@ func stackTypeFilter() filter.Filters {
...
+	}
+
+	return util.ResolvePath(canonical), nil
+}
+
+// ensureWorkingDir makes sure workingDir exists and is a directory. Missing
+// paths are created with MkdirAll (mkdir -p semantics), so fresh CI
+// environments do not need to pre-create the directory. A path that exists
+// but is a regular file is still rejected with ErrWorkingDirNotDirectory.
+func ensureWorkingDir(workingDir string) error {
```
yhakbar: What are we guarding against here? This is a little strange. Can't we just call MkdirAll and guard against ENOTDIR?

internal/stacks/generate/generate_test.go
```diff
@@ -0,0 +1,221 @@
+// This is a white-box test file: it needs access to the unexported
+// ensureWorkingDir helper and StackGenerator, and to the exported sentinel
+// errors they return. The lock-path / vfs.FS tests from earlier revisions
+// were removed when the filesystem lock was replaced by an in-process
+// manager.
+//
+//nolint:testpackage // white-box testing of unexported helpers
+package generate
```
yhakbar: Let's not do this. We can just export EnsureWorkingDir if we need to.

internal/stacks/generate/generate_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+		close(locked)
+	}()
+
+	// Prove the second Lock() is blocked: 50ms is generous for goroutine
+	// scheduling, far short enough that the test is fast.
+	select {
+	case <-locked:
+		t.Fatal("second Lock() returned while the first was still holding")
+	case <-time.After(50 * time.Millisecond):
```
yhakbar: Make sure you use synctest whenever you deal with clocks. It removes the need to slow down tests.

internal/stacks/generate/generate_test.go
```diff
@@ -0,0 +1,221 @@
...
+	case <-locked:
+	case <-time.After(2 * time.Second):
+		t.Fatal("second Lock() did not unblock after the first Unlock()")
+	}
+}
+
+// TestGenerateMutexConcurrent asserts that many goroutines contending on
+// the same key all eventually acquire+release without error and observe a
+// serialized critical section (no concurrent holders at any instant).
+func TestGenerateMutexConcurrent(t *testing.T) {
```
yhakbar: This can be prefixed WithRacing and use 2 goroutines, right? We'll probably see the race from the race detector if it's there.

internal/stacks/generate/generate_test.go
```diff
@@ -0,0 +1,221 @@
...
+		eg.Go(func() error {
+			g.lockManager.Lock(key)
+			active++
+
+			if active > maxActive {
+				maxActive = active
+			}
+
+			// Short critical section.
+			time.Sleep(5 * time.Millisecond)
```
yhakbar: Again, be mindful of sleeping in tests.

internal/stacks/generate/generate_test.go
```diff
@@ -0,0 +1,221 @@
...
+	require.Equal(t, 1, maxActive, "mutex must enforce exactly one holder at a time")
+}
+
+// TestDefaultGeneratorWiring asserts that the package-level GenerateStacks
+// function and hook registration properly use the defaultStackGenerator.
+// Regression guard for the singleton wiring.
+//
+// Cannot be parallel because it mutates the global defaultStackGenerator hooks.
+//
+//nolint:paralleltest // mutates package-level defaultStackGenerator
```
yhakbar: We should pass in the dependency instead of mutating the global variable.

internal/stacks/generate/generate_test.go
```diff
@@ -0,0 +1,221 @@
...
+		g.lockManager.Lock(key2)
+		defer g.lockManager.Unlock(key2)
+
+		close(locked2)
+	}()
+
+	// Second key must succeed promptly even while the first is held.
+	select {
+	case <-locked2:
+	case <-time.After(2 * time.Second):
```
yhakbar: Make sure this uses synctest too.

internal/util/locks_test.go
```diff
@@ -27,35 +28,56 @@ func TestKeyLocksBasic(t *testing.T) {
...
+
+	go func() {
+		kl.Lock("b")
+		kl.Unlock("b")
+		close(done)
+	}()
+
+	select {
+	case <-done:
+	case <-time.After(2 * time.Second):
```
yhakbar, line 79: NIT: Make sure to use synctest for this. I'll instantly move the clock forward if nothing is blocked.

yhakbar (approved):

PR #5963, opened 2026-04-22
feat: Mark module sources as read

### LINKED ISSUE #5956: Mark module sources as read
## Describe the enhancement

When referencing a local module source using a `terraform` block, mark all the source files in the module as "read" for the purposes of the internal read file tracking that Terragrunt does to support [reading-based](https://docs.terragrunt.com/features/filter/attributes/#reading-based-expressions) expressions.

In addition to supporting this capability, add support for a `mark_glob_as_read` HCL function, which will allow for a glob to select a collection of files as being read.

## Additional context

This is particularly helpful in the context of [Pipelines](https://docs.terragrunt.com/terragrunt-scale/pipelines/), where customers expect changes to a local module to trigger actions on units.

This will require an experiment flag, as it's a forwards incompatible change in feature behavior for Terragrunt.

## RFC Not Needed

yhakbar:
## Description

Closes #5956

Adds a new `mark-many-as-read` experiment that marks many files as read for a unit in one step.

When enabled, units with a local `terraform { source = "..." }` automatically mark every source file under that path as read, and a new `mark_glob_as_read` HCL function lets users mark arbitrary globs of files as read.

Reading-based filters (e.g. `--filter 'reading='`) then cascade module edits to consuming units without each unit having to enumerate files by hand. This propagates to Git-based filters as well.

Also consolidates glob handling into one `internal/glob` package shared by the `filter` and the new HCL function, and routes filesystem access through `vfs`.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **New Features**
  * Introduced the mark-many-as-read experiment to auto-mark local Terraform/OpenTofu module source files as read and let reading-based filters cascade those changes.
  * Added a new HCL function to mark files by glob patterns and return matched absolute paths.

* **Documentation**
  * Added docs and a v1.0.3 changelog entry explaining the experiment, the new function, glob semantics, and usage notes.

* **Tests**
  * Added tests covering glob expansion, escaped metacharacters, experiment gating, and module-file marking behavior.

### DIFF
--- a/pkg/config/config_helpers.go
+++ b/pkg/config/config_helpers.go
@@ -42,6 +42,7 @@ import (
 	"github.com/gruntwork-io/terragrunt/internal/telemetry"
 	"github.com/gruntwork-io/terragrunt/internal/tf"
 	"github.com/gruntwork-io/terragrunt/internal/util"
+	"github.com/gruntwork-io/terragrunt/internal/vfs"
 	"github.com/gruntwork-io/terragrunt/pkg/config/hclparse"
 	"github.com/gruntwork-io/terragrunt/pkg/log"
 )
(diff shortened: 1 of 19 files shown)


pkg/config/config_helpers.go
```diff
@@ -1679,6 +1684,70 @@ func markAsRead(ctx context.Context, pctx *ParsingContext, l log.Logger, args []
...
+			pattern = path.Clean(filepath.ToSlash(pctx.WorkingDir) + "/" + raw)
+		}
+
+		matches, err := glob.Expand(vfs.NewOSFS(), pattern, glob.WithFilesOnly())
+		if err != nil {
+			return errors.New(fmt.Errorf("could not expand glob %q: %w", raw, err))
+		}
+
+		for _, match := range matches {
+			trackFileRead(pctx.FilesRead, match)
```
denis256, line 1741: Possible data race in trackFileRead due to shared FilesRead pointer in `ParsingContext.Clone()`

ParsingContext.Clone() performs a shallow copy of the struct, including `FilesRead *[]string pointer`

ParseConfigFile call trackFileRead, which appends to this shared slice. Since append is not thread-safe

denis256 (dismissed):

pkg/config/config_helpers.go
yhakbar: This isn't new. I can look to address it, but I'm not sure there's actually a race that can be exercised here.

denis256 (approved):

PR #5969, opened 2026-04-24
chore: stacks generation improvements

denis256:
## Description

* added a warning on repeated stack claims
* reduced false positive tests

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **New Features**
  * Emit warnings when the same stack file is claimed by different parent stacks during generation, highlighting contention and recording the latest claimant.

* **Tests**
  * Improved test reliability for concurrent scenarios by running synchronization-sensitive checks under a deterministic harness.
  * Simplified an integration test to run generation steps sequentially to reduce flakiness.


internal/stacks/generate/generate.go
```diff
@@ -111,6 +116,33 @@ func (g *Generator) GenerateStacks(
...
+		if !ok {
+			claimedBy[node.FilePath] = parent
+			continue
+		}
+
+		if prev == parent {
+			continue
+		}
+
+		const msg = "stack %s is being generated by %s, but it was previously generated by %s. " +
```
yhakbar: Are we definitely generating a stack here? Do we need to check if we're generating a unit? Also, we can proper case this log message. Also, I think we can make the message stronger. I think the last sentence can be something like "You have contention between multiple stacks generating to the same directory. This is probably a bug in your stack configurations."

yhakbar (dismissed):

internal/stacks/generate/generate.go
denis256: It is warning for stack file

yhakbar (approved):

PR #5984, opened 2026-04-27
fix: panics in startswith / endswith / strcontains / run_cmd

denis256:
## Description

* updated startswith / endswith / strcontains / run_cmd functions to do better parameter handling
* added more tests to track parameters

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented crashes by validating command arguments after filtering options and returning a clear error when no command remains
  * Enforced exact argument counts for string/templating helpers and surface mismatches with a clear error

* **New Features**
  * Allow injecting a mockable execution backend and improved subprocess error reporting
  * Produce deterministic env var ordering for subprocesses

* **Tests**
  * Added regression and fuzz tests with in-memory exec mocks to ensure correct behavior and no panics


pkg/config/fuzz_test.go
```diff
@@ -0,0 +1,71 @@
...
+		f.Add(s)
+	}
+
+	f.Fuzz(func(t *testing.T, raw string) {
+		args := strings.Split(raw, "\x00")
+
+		l := logger.CreateLogger()
+		ctx, pctx := newTestParsingContext(t, "")
+
+		_, _ = config.RunCommand(ctx, pctx, l, args)
```
yhakbar: How is this safe?

internal/shell/run_cmd.go
```diff
@@ -37,6 +38,7 @@ type ShellOptions struct {
 	EngineConfig  *engine.EngineConfig
 	Telemetry     *telemetry.Options
 	Env           map[string]string
+	Exec          vexec.Exec
```
yhakbar: We shouldn't embed dependencies like this if we can avoid it. Can this be explicitly passed in as a parameter?

internal/shell/run_cmd.go
```diff
@@ -248,6 +258,24 @@ func RunCommandWithOutput(
 			}
 		}
 
+		if runOpts.Exec != nil {
+			if needsPTY {
+				return errors.New("shell: vexec backend does not support PTY mode")
```
yhakbar: Can we look to leverage this pattern?
https://github.com/gruntwork-io/terragrunt/pull/5957/changes#diff-c9a29ac513b03b203827d19beb65f9eff05111842ab545c071f25a656f998fc4R620

We should unwrap the underlying os exec from vexec when it doesn't support certain behavior (like pty support).

internal/util/collections.go
```diff
@@ -133,3 +133,17 @@ func SplitUrls(s, sep string) []string {
 
 	return urls
 }
+
+// EnvSliceFromMap converts an env-var map to the KEY=VALUE slice form
+// expected by os/exec.Cmd.Env. The result is sorted for deterministic output.
+func EnvSliceFromMap(env map[string]string) []string {
```
yhakbar: This feels tightly coupled to exec behavior. Not sure that it belongs here.

pkg/config/parsing_context.go
```diff
@@ -52,6 +53,8 @@ type ParsingContext struct {
 	Features            *cty.Value
 	Locals              *cty.Value
 
+	Exec vexec.Exec
```
yhakbar: Really don't like embedding this. We should explicitly pass it in as a parameter, like `l` and `fs`.

yhakbar (dismissed):

pkg/config/config_helpers_test.go
```diff
@@ -30,20 +31,32 @@ import (
 )
 
 // assertErrorType checks that the error chain contains an error of the same type as expectedErr.
+// Used by table-driven tests where the expected type is known only at runtime;
+// for static call sites prefer the type-parameterized requireErrorAs helper.
 func assertErrorType(t *testing.T, expectedErr, actualErr error) bool {
```
yhakbar, line 36: Why do we have this? Feels like we should be using assert/require ErrorAs

pkg/config/config_helpers_test.go
```diff
@@ -30,20 +31,32 @@ import (
...
+		return true
 	}
 
 	return assert.Fail(t, "error type mismatch", "expected error of type %T in chain, but got %T", expectedErr, actualErr)
 }
 
+// requireErrorAs unwraps err looking for a value of type T and fails the test when none is found.
+// Type-parameterized wrapper over errors.As; prefer over assertErrorType when the target type is
+// known statically.
+func requireErrorAs[T error](t *testing.T, err error) T {
```
yhakbar, line 50: Why does this wrapper exist?

yhakbar (approved):

PR #6013, opened 2026-05-01
feat: Adding stack target tip

yhakbar:
## Description

Adds a tip that explains to users they might get unexpected results when using filters targetign a stack without the `| type=stack` in-fix expression.

Shows a suggested update to the filter that adds `| type=stack` to correctly target the stack, and links them to the docs.

## TODOs

- [x] Update the docs.

## Release Notes (draft)

## Summary by CodeRabbit

* **New Features**
  * Added a tip that alerts when a --filter resolves to a stack directory but is not restricted with | type=stack; appears for relevant commands and offers a suggested rewrite and suppression option.
  * Bare dot (.) is now supported as a working-directory filter.

* **Documentation**
  * Changelog updated with the new stack-target tip behavior.

* **Tests**
  * Added unit and integration tests covering the tip and the bare-dot/filter parsing.


internal/tips/stack_target.go
```diff
@@ -0,0 +1,111 @@
...
+// Expression.IsRestrictedToStacks() is true. The contract holds because `|`
+// is the only infix operator in the filter grammar and `!` binds tighter, so
+// appending `| type=stack` always produces an InfixExpression whose right
+// side is the stack-type attribute filter.
+func SuggestStackTargetRewrite(originalQuery string) string {
+	return originalQuery + " | type=stack"
+}
+
+func containsGlobMeta(s string) bool {
+	return strings.ContainsAny(s, "*?[")
```
denis256: Not sure if ti will handle cases of `{}`, like `./envs/{prod,stg}`

internal/tips/stack_target.go
```diff
@@ -0,0 +1,111 @@
...
+
+// SuggestStackTargetRewrite returns the suggested replacement for a filter
+// query that targets a stack directory but is missing `| type=stack`.
+//
+// Tests pin the contract that the returned string parses to a filter whose
+// Expression.IsRestrictedToStacks() is true. The contract holds because `|`
+// is the only infix operator in the filter grammar and `!` binds tighter, so
+// appending `| type=stack` always produces an InfixExpression whose right
+// side is the stack-type attribute filter.
+func SuggestStackTargetRewrite(originalQuery string) string {
```
denis256: Why do we export SuggestStackTargetRewrite()? Looks like it's only invoked in tests

denis256 (dismissed):

internal/tips/stack_target.go
yhakbar: Because it's invoked in tests. We only test the public surface of packages.

internal/tips/stack_target.go
yhakbar: We don't need to worry about that because of how filters handle `{}`.

docs/src/data/changelog/v1.0.5/stack-target-missing-type-stack-tip.mdx
```diff
@@ -0,0 +1,10 @@
+---
+version: "v1.0.4"
```
denis256: Version I think should be 1.0.5

internal/tips/stack_target.go
```diff
@@ -0,0 +1,115 @@
...
+		if filterTargetsStackDir(f, fs, workingDir) {
+			offenders = append(offenders, f.String())
+		}
+	}
+
+	if len(offenders) == 0 {
+		return
+	}
+
+	tip.Message = buildStackTargetMessage(offenders)
```
denis256: Changing tip.Message in an unsynchronized way - for run -all case may lead to losing the message value

docs/src/data/changelog/v1.0.5/stack-target-missing-type-stack-tip.mdx
yhakbar: Fixed.

internal/tips/stack_target.go
yhakbar: I don't think this should happen, given that it validates the filters early before runs start, but I added synchronization to be safe.

internal/cli/commands/run/run.go
```diff
@@ -17,15 +17,19 @@ import (
...
 	"github.com/gruntwork-io/terragrunt/internal/vexec"
+	"github.com/gruntwork-io/terragrunt/internal/vfs"
 	"github.com/gruntwork-io/terragrunt/pkg/config"
 	"github.com/gruntwork-io/terragrunt/pkg/log"
 	"github.com/gruntwork-io/terragrunt/pkg/options"
 )
 
 // Run runs the run command.
 func Run(ctx context.Context, l log.Logger, opts *options.TerragruntOptions) error {
+	tips.GiveStackTargetTip(l, vfs.NewOSFS(), opts.WorkingDir, opts.Filters, opts.Tips)
```
denis256, line 31: Not sure if it will work in case of ` terragrunt run --filter ... `

1. internal/cli/flags/shared/filter.go:44 sets opts.RunAll = true when --filter is used
2. When internal/cli/commands/run/cli.go:38-39 checks tgOpts.RunAll and returns directly
  to runall.Run(ctx, l, tgOpts)
3. Because of that early return, internal/cli/commands/run/run.go:31, where tips.GiveStackTargetTip(...) is called, is not reached for terragrunt run --filter ....

internal/tips/stack_target.go
```diff
@@ -0,0 +1,114 @@
...
+// SuggestStackTargetRewrite returns the suggested replacement for a filter
+// query that targets a stack directory but is missing `| type=stack`. It is
+// exported so the contract test can pin its behavior from a `tips_test`
+// black-box test.
+//
+// The contract holds because `|` is the only infix operator in the filter
+// grammar and `!` binds tighter, so appending `| type=stack` always produces
+// an InfixExpression whose right side is the stack-type attribute filter.
+func SuggestStackTargetRewrite(originalQuery string) string {
+	return originalQuery + " | type=stack"
```
denis256, line 59: Not sure about this append - if I will have something like `./envs/prod | !type=stack`, it will add an ambiguous string text `| type=stack`

internal/tips/stack_target.go
```diff
@@ -0,0 +1,114 @@
...
+		candidate := pe.Value
+		if !filepath.IsAbs(candidate) {
+			candidate = filepath.Join(workingDir, candidate)
+		}
+
+		stackFile := filepath.Join(candidate, config.DefaultStackFile)
+
+		exists, err := vfs.FileExists(fs, stackFile)
+		if err != nil || !exists {
+			return true
```
denis256, line 111: Possible hiding of error from filesystem

internal/tips/stack_target.go
```diff
@@ -0,0 +1,114 @@
...
+
+	return b.String()
+}
+
+// containsGlobMeta reports whether s contains any glob metacharacter that can
+// appear in a parsed PathExpression.Value. The filter lexer emits '[', '{',
+// '}', and ']' as their own tokens, so the only reachable meta chars here are
+// '*', '?', and '\\'.
+func containsGlobMeta(s string) bool {
+	return strings.ContainsAny(s, "*?\\")
```
denis256, line 140: Will it catch Windows paths?

denis256 (dismissed):

internal/tips/stack_target.go
yhakbar: Windows paths can't get here anyways. Users have to normalize paths to forward slashes for glob compatibility.

internal/tips/stack_target.go
yhakbar: Added debug log.

internal/tips/stack_target.go
yhakbar: Good catch. Excluding any situation where the type is explicitly defined.

internal/cli/commands/run/run.go
yhakbar: Added extra call in alternate path.

denis256 (approved):

Issue #4307, opened 2025-05-20
Implementation of Azure Resource Manager (azurerm) Backend for Terragrunt's Remote State Management


raimdev:
### Summary

The azurerm backend will allow users to leverage the "remote state" feature with Azure storage. This implementation enables the use of the `--dependency-fetch-output-from-state` option, which significantly improves deployment speed for complex and large infrastructures.

### Motivation

As organizations increasingly adopt Azure as their cloud provider, there is a growing need for tools that facilitate infrastructure management in Azure environments. Currently, Terragrunt supports S3 and GCS backends for remote state management, but lacks support for Azure. By implementing an azurerm backend, we can enhance Terragrunt's usability for Azure users, enabling them to manage their Terraform state files more effectively.

### Proposal

Implement a remote state backend for azurerm using [go-azure-sdk](https://github.com/hashicorp/go-azure-sdk).

### Technical Details

The implementation will include the following components:

- **Backend Interface**: Implement the `backend.Backend` interface to ensure compatibility with Terragrunt's existing architecture.
- **Configuration Struct**: Define a configuration struct to hold necessary parameters such as `storage_account_name`, `container_name`, and `key`.

The configuration for the azurerm backend will include the following fields:

```hcl
remote_state {
    backend = "azurerm"
    config = {
        storage_account_name = "mytfstateaccount"
        container_name       = "mytfstatecontainer"
        key                  = "terraform.tfstate"
        use_azuread_auth     = true
        subscription_id      = ""
        tenant_id            = ""
    }
}
```

### Press Release

Terragrunt now supports the Azure backend in remote state configuration, enabling seamless state management for Azure-based infrastructure.

### Support Level

- [ ] I have Terragrunt Enterprise Support
- [ ] I am a paying Gruntwork customer

yhakbar:
Hey @raimdev ,

As it stands, we have no plans to support the Azure Resource Manager Backend natively as we do for S3 and GCP.

Unfortunately, we can't justify the increased maintenance burden on maintainers right now. I'll leave this RFC open, in case that changes in the future.

omattsson:
Hi @raimdev @yhakbar ,

I have been looking at this for a while. At work, we exclusively use Azure, and to accelerate the adoption of Terragrunt, this feature makes sense. So I have spent some time on it and have implemented Azure storage as the supported state backend. At the moment, I depend on an existing storage account.

If you want to have a look, I can create a WIP PR with the current state of the code.

[...]

PR #6017, opened 2026-05-01
feat(azurehelper): add Azure SDK helper package (PR 2 of 3 for #4307)

omattsson:
## Description

Part 2 of 3 for Azure remote-state backend support (#4307).

Adds `internal/azurehelper`, a flat Azure SDK helper package that mirrors the
shape of [`internal/awshelper`](https://github.com/gruntwork-io/terragrunt/tree/main/internal/awshelper)
and [`internal/gcphelper`](https://github.com/gruntwork-io/terragrunt/tree/main/internal/gcphelper):

- `config.go` — `AzureSessionConfig` + `AzureConfigBuilder` (builder pattern,
  6-tier auth resolution: SAS token → access key → service principal → MSI →
  OIDC → AzureAD default)
- `storage_account.go` — ARM control-plane client (Exists, Create,
  GetKeys, EnableVersioning, EnableSoftDelete, Delete)
- `blob.go` — Data-plane wrapper around `azblob.Client` with bound
  container, Get/Put/List/Copy/Delete
- `resource_group.go` — Exists / EnsureResourceGroup / Delete
- `rbac.go` — Role assignment helpers (AssignRole / AssignRoleIfMissing /
  HasRoleAssignment / RemoveRole) needed for `use_azuread_auth` bootstrap
- `errors.go` — Sentinel errors, typed error structs, and the `IsNotFound` /
  `IsRetryable` predicates (stdlib `errors`, no `internal/errors` dependency)

**No backend wiring yet** — that lands in PR 3.

### Conventions followed

- All Azure SDK clients constructed via `policy.ClientOptions` so tests can
  inject a fake `policy.Transporter` (no interfaces, no mocks — the
  SDK-blessed seam, same as `gcphelper`)
- Tests gated behind `//go:build azure`, mirroring `gcphelper`'s
  `//go:build gcp`
- Package layout flat, no factory/adapter/interface indirection

### Test coverage

`go test -tags=azure -cover ./internal/azurehelper/... → 60.2%`

Includes transport-fake happy-path coverage for ARM operations
(EnableVersioning, IsVersioningEnabled, EnableSoftDelete clamp,
ResourceGroup Exists/EnsureResourceGroup, RBAC RemoveRole) plus extensive
argument validation.

## Review-feedback follow-ups

Addressed the round of CodeRabbit / Copilot review comments:

- `errors.go` — `IsRetryable` short-circuits `context.Canceled` /
  `context.DeadlineExceeded` (caller-driven; never retry).
- `config.go` — OIDC path now uses `azidentity.NewWorkloadIdentityCredential`
  (was the broad `NewDefaultAzureCredential` chain). To make this usable
  without process env mutation, `AzureSessionConfig` gains an
  `OIDCTokenFilePath` field with env fallbacks `ARM_OIDC_TOKEN_FILE_PATH`
  (Terraform azurerm convention) and `AZURE_FEDERATED_TOKEN_FILE`
  (azidentity / AKS workload identity convention). The constructor needs a
  non-empty token file path up front, so threading it through avoids a
  surprising failure mode on first use.
- `config.go` — `BuildStorageAccountClient` runs `applyEnvFallbacks` on a
  config copy before the SAS-token / access-key pre-flight, so
  `ARM_SAS_TOKEN` / `ARM_ACCESS_KEY` configs are rejected up front instead
  of deep inside ARM client construction.
- `rbac.go` — `RemoveRole` treats `IsNotFound(err)` from `DeleteByID` as a
  successful race (concurrent deleter), not a hard error.
- `storage_account.go` — `Create` wires `EnableVersioning` post-creation
  (godoc updated to reflect the implicit-versioning behaviour).
- `storage_account.go` — `accessTierValue(string) (AccessTier, error)`
  rejects unknown tiers (no silent coercion to `Hot`). `Create` validates
  the result before building create-parameters.
- `storage_account.go` — `GetKeys` filters out empty key values and
  surfaces the degenerate "no usable keys returned" response as an error.

New tests cover context-cancel paths, unknown access tiers, and empty-key
responses.

## Dependency

**Depends on PR 1** (registers the `azure-backend` experiment + stub
`azurerm` backend). PR 1's two commits (`576cd20`, `183827a`) are
included in this branch's history; once PR 1 merges into `main`, this PR
will rebase down to azurehelper-only.

## TODOs

- [x] Run the relevant tests successfully, including pre-commit checks
  (`go test -tags=azure ./internal/azurehelper/...` and
  `golangci-lint run --build-tags=azure ./internal/azurehelper/...` both
  green)
- [x] This change is backwards compatible (additive, gated behind the
  `azure-backend` experiment from PR 1)
- [ ] Update the docs (PR 3, alongside backend wiring)
- [ ] Update the changelog (PR 3, alongside the user-visible feature)

## Release Notes (draft)

```
NOTE: internal-only — `internal/azurehelper` package added with no
public surface change. See PR 3 for the user-visible Azure backend.
```

## Summary by CodeRabbit

* **New Features**
  * Azure Blob storage: upload, download, delete, list, copy, and container binding.
  * Azure Storage Account: create, delete, versioning, soft-delete, and key management.
  * Azure Resource Group: existence checks, create-if-missing, and delete.
  * Azure RBAC: assign, check, and remove role assignments.
  * Authentication: support for SAS, access keys, service principal, MSI, OIDC, and default AAD; cloud environment selection.

* **Tests**
  * Added extensive Azure-only unit and live integration tests for config, blob, RBAC, resource groups, storage accounts, and error handling.

* **Dependencies**
  * Updated Azure SDK dependencies.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gruntwork-io/terragrunt/pull/6017?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


internal/azurehelper/blob.go
```diff
@@ -0,0 +1,348 @@
...
+	if srcContainer == "" || srcKey == "" || dstContainer == "" || dstKey == "" {
+		return errors.Errorf("source and destination container/key are required")
+	}
+
+	srcURL := fmt.Sprintf("https://%s.blob.%s/%s/%s",
+		c.accountName, c.endpointSuffix, srcContainer, srcKey)
+
+	dst := c.client.ServiceClient().NewContainerClient(dstContainer).NewBlobClient(dstKey)
+	if _, err := dst.StartCopyFromURL(ctx, srcURL, nil); err != nil {
+		return WrapError(err, fmt.Sprintf("copying blob %s/%s to %s/%s", srcContainer, srcKey, dstContainer, dstKey))
```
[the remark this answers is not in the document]
omattsson: Thanks — confirmed the current `srcURL` is built via raw string concatenation. The fix (constructing the URL via `net/url.URL{Scheme,Host,Path}` so each path segment is properly percent-encoded) has already landed on the PR 3 branch (`pr3-azurerm-backend`). I will either port it back to this branch directly or let it ride along when PR 3 rebases onto `main` after PR 2 merges — happy to do whichever is preferred.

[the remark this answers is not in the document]
omattsson: Confirmed — the current `srcURL` is built via raw string concatenation. The fix (constructing the URL via `net/url.URL{Scheme,Host,Path}` so each path segment is properly percent-encoded) has already landed on the PR 3 branch (`pr3-azurerm-backend`). Happy to either port it back to this branch directly or let it ride along when PR 3 rebases onto `main` after PR 2 merges — preference noted either way.

internal/azurehelper/config.go
```diff
@@ -0,0 +1,415 @@
...
+		if err != nil {
+			return nil, errors.Errorf("creating managed identity credential: %w", err)
+		}
+
+		out.Method = AuthMethodMSI
+		out.Credential = cred
+
+		l.Debugf("azurehelper: using managed identity authentication")
+
+		return out, validate(out, &resolved)
```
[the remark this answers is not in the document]
omattsson: Acknowledged — the switch currently checks `resolved.UseOIDC` before `resolved.UseMSI`, so a config with both flags enabled picks workload identity instead of MSI. Will reorder the cases so MSI wins when both are set, matching the documented precedence.

internal/azurehelper/blob_test.go
```diff
@@ -0,0 +1,407 @@
...
+	// Exercise GetObject via bound container.
+	bc.BindContainer(container)
+
+	body2, err := bc.GetObject(ctx, key)
+	if err != nil {
+		t.Fatalf("GetObject: %v", err)
+	}
+
+	got2, _ := io.ReadAll(body2)
+	_ = body2.Close()
```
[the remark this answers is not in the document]
omattsson: Acknowledged — the live blob round-trip test silently discards the `Close()` and `io.ReadAll` errors at those lines, which would mask a real read failure as a payload mismatch. Will tighten this so both errors are surfaced via `t.Fatalf` / `t.Errorf`.

internal/azurehelper/config.go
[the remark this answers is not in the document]
omattsson: Fixed in 837f3c3 — swapped the case blocks (MSI before OIDC), updated the `Build` godoc resolution-order numbering, and flipped the existing precedence test to assert `AuthMethodMSI` when both `UseMSI` and `UseOIDC` are set. `go test -tags=azure ./internal/azurehelper/...` is green.

omattsson:
@yhakbar @denis256 How can I proceed with this PR?

yhakbar:
Hey @omattsson,

We're getting around to reviewing this again. Please rebase the PR to resolve the conflict in `go.mod`.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+	"context"
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
+	azblobcontainer "github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container"
+
+	"github.com/gruntwork-io/terragrunt/internal/errors"
```
yhakbar: This package was recently removed. Please use the standard library for errors.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
+// Blob and container data-plane operations.
+//
+// BlobClient wraps azblob.Client and exposes the small data-plane surface
+// the remote-state backend needs (containers, blobs, listing, copy). It
+// also remembers an optional bound container so callers fetching state
+// files by key (e.g. PR 3's dependency-fetch path) do not have to repeat
```
yhakbar: Please remove these references to future PRs. Comments should be written such that they are relevant well into the future.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+
+// NewBlobClient builds an *azblob.Client from an AzureConfig and returns it
+// wrapped in BlobClient. SAS-token configs use the no-credential constructor;
+// access-key configs use a shared key credential; all other methods use the
+// AzureConfig.Credential as a token credential.
+//
+// endpointSuffix selects the blob endpoint host (e.g. "core.windows.net" for
+// Azure public cloud, "core.usgovcloudapi.net" for US Government). When empty,
+// the suffix is derived from cfg.CloudConfig.
+func NewBlobClient(_ context.Context, cfg *AzureConfig, endpointSuffix string) (*BlobClient, error) {
```
yhakbar: Why is this context parameter here if it isn't going to be used?

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+	if cfg == nil {
+		return nil, errors.Errorf("azure config is required")
```
yhakbar: Use proper typed errors so that they can be matched with `errors.Is`/`As`.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	if cfg.AccountName == "" {
+		return nil, errors.Errorf("storage account name is required")
```
yhakbar: Use proper typed errors so that they can be matched with `errors.Is`/`As`.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	suffix := endpointSuffix
+	if suffix == "" {
```
yhakbar: We generally prefer a builder pattern instead of expecting empty strings being passed in to control construction. Please see if this can be adjusted accordingly.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+		}
+
+		client, err = azblob.NewClient(url, cfg.Credential, clientOpts)
+	default:
+		return nil, errors.Errorf("unsupported azure auth method %q", cfg.Method)
+	}
+
+	if err != nil {
+		return nil, errors.Errorf("creating blob client: %w", err)
+	}
```
yhakbar: This could be a private function call to avoid the mutability of `client` and `err` here.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+	var (
+		client *azblob.Client
+		err    error
+	)
+
+	switch cfg.Method {
+	case AuthMethodSasToken:
+		sas := strings.TrimPrefix(cfg.SasToken, "?")
+
+		client, err = azblob.NewClientWithNoCredential(url+"?"+sas, clientOpts)
```
yhakbar: We should probably be using proper url construction here using [url](https://pkg.go.dev/net/url). Why are we just doing string manipulation?

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+
+		cred, err = azblob.NewSharedKeyCredential(cfg.AccountName, cfg.AccessKey)
+		if err != nil {
+			return nil, errors.Errorf("creating shared key credential: %w", err)
+		}
+
+		client, err = azblob.NewClientWithSharedKeyCredential(url, cred, clientOpts)
+	case AuthMethodServicePrincipal, AuthMethodOIDC, AuthMethodMSI, AuthMethodAzureAD:
+		if cfg.Credential == nil {
+			return nil, errors.Errorf("azure config has no credential for method %q", cfg.Method)
```
yhakbar: This should be a proper error type so it can be matched with `errors.As`.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+		}
+
+		client, err = azblob.NewClient(url, cfg.Credential, clientOpts)
+	default:
+		return nil, errors.Errorf("unsupported azure auth method %q", cfg.Method)
```
yhakbar: This should be a proper error type so it can be matched with `errors.As`.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+// Container returns the bound container name, or empty string if no
+// container has been bound.
+func (c *BlobClient) Container() string { return c.container }
+
+// ContainerExists reports whether the named container exists in the account.
+// Returns (false, nil) for ContainerNotFound / ResourceNotFound; other errors
+// are returned wrapped.
+func (c *BlobClient) ContainerExists(ctx context.Context, name string) (bool, error) {
+	if name == "" {
+		return false, errors.Errorf("container name is required")
```
yhakbar: This should be a proper error type so it can be matched with `errors.Is`.

internal/azurehelper/errors.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+	}
+
+	return false
+}
+
+// WrapError wraps err with the Terragrunt internal errors package so the
+// caller's stack trace is preserved, prefixing it with op for context.
+// Returns nil if err is nil.
+func WrapError(err error, op string) error {
```
yhakbar: Why does this helper exist? You can just call `fmt.Errorf("extra context: %w", err)` at call sites inline.

internal/azurehelper/errors.go
```diff
@@ -0,0 +1,135 @@
...
+		ErrorClassInvalidRequest:
+		return false
+	}
+
+	return false
+}
+
+// IsNotFound reports whether the error represents a 404 / ResourceNotFound
+// from an Azure API.
+func IsNotFound(err error) bool {
```
yhakbar, line 135: This seems like a pretty course check when we can probably make a much more strict check at call sites. Is it a good idea to route error handling through this?

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	_, err := c.client.ServiceClient().NewContainerClient(name).GetProperties(ctx, nil)
+	if err == nil {
+		return true, nil
+	}
+
+	if IsNotFound(err) {
```
yhakbar: This seems like a course check for something that could be much more granular.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+		return false, nil
+	}
+
+	return false, WrapError(err, "checking container existence")
```
yhakbar: We don't need a helper for wrapping errors like this. We can just call `fmt.Errorf` inline here.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+}
+
+// CreateContainer creates the container. If it already exists, this returns
+// nil (no-op). Other errors are returned wrapped.
+func (c *BlobClient) CreateContainer(ctx context.Context, name string) error {
+	if name == "" {
+		return errors.Errorf("container name is required")
```
yhakbar: This should be a proper error type so it can be matched with `errors.Is`.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+		return nil
+	}
+
+	var respErr *azcore.ResponseError
+
+	if errors.As(err, &respErr) && strings.EqualFold(respErr.ErrorCode, "ContainerAlreadyExists") {
+		return nil
+	}
+
+	return WrapError(err, "creating container "+name)
```
yhakbar: We don't need a helper for wrapping errors like this. We can just call `fmt.Errorf` inline here.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+}
+
+// CreateContainerIfNecessary is a convenience for the common "ensure exists"
+// pattern: checks existence first, then creates only if missing.
+func (c *BlobClient) CreateContainerIfNecessary(ctx context.Context, name string) error {
```
yhakbar: Prefer the name `EnsureContainer`.

Communicates intent more succinctly, and removes the fluff in the comment.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+		return nil
+	}
+
+	return c.CreateContainer(ctx, name)
+}
+
+// DeleteContainer deletes the named container. Missing containers return nil.
+func (c *BlobClient) DeleteContainer(ctx context.Context, name string) error {
+	if name == "" {
+		return errors.Errorf("container name is required")
```
yhakbar: This should be a proper error type so it can be matched with `errors.Is`.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	_, err := c.client.ServiceClient().NewContainerClient(name).Delete(ctx, nil)
+	if err == nil || IsNotFound(err) {
```
yhakbar: This is a course error handle. Can it be made more granular?

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+		return nil
+	}
+
+	return WrapError(err, "deleting container "+name)
```
yhakbar: We don't need a helper for wrapping errors like this. We can just call `fmt.Errorf` inline here.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+}
+
+// GetBlob downloads a blob and returns its body as an io.ReadCloser. Caller
+// must Close the returned reader.
+func (c *BlobClient) GetBlob(ctx context.Context, container, key string) (io.ReadCloser, error) {
+	if container == "" || key == "" {
```
yhakbar, line 227: This should be a proper error type so it can be matched with `errors.As`.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+		return nil, errors.Errorf("container name and blob key are required")
+	}
+
+	resp, err := c.client.DownloadStream(ctx, container, key, nil)
+	if err != nil {
+		return nil, WrapError(err, "downloading blob "+container+"/"+key)
```
yhakbar: We don't need a helper for wrapping errors like this. We can just call `fmt.Errorf` inline here.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	return nil
+}
+
+// PutBlobFromReader uploads a blob by streaming from reader, avoiding loading
+// the full payload into memory.
+func (c *BlobClient) PutBlobFromReader(ctx context.Context, container, key string, reader io.Reader) error {
+	if container == "" || key == "" {
+		return errors.Errorf("container name and blob key are required")
```
yhakbar: This should be a proper error type so it can be matched with `errors.As`.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+
+// PutBlob uploads data to a block blob, overwriting any existing blob.
+func (c *BlobClient) PutBlob(ctx context.Context, container, key string, data []byte) error {
+	if container == "" || key == "" {
+		return errors.Errorf("container name and blob key are required")
+	}
+
+	blockBlob := c.client.ServiceClient().NewContainerClient(container).NewBlockBlobClient(key)
+	if _, err := blockBlob.UploadBuffer(ctx, data, nil); err != nil {
+		return WrapError(err, "uploading blob "+container+"/"+key)
```
yhakbar: We don't need a helper for wrapping errors like this. We can just call `fmt.Errorf` inline here.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+		return nil, WrapError(err, "downloading blob "+container+"/"+key)
+	}
+
+	return resp.Body, nil
+}
+
+// PutBlob uploads data to a block blob, overwriting any existing blob.
+func (c *BlobClient) PutBlob(ctx context.Context, container, key string, data []byte) error {
+	if container == "" || key == "" {
+		return errors.Errorf("container name and blob key are required")
```
yhakbar: This should be a proper error type so it can be matched with `errors.As`.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+// PutBlobFromReader uploads a blob by streaming from reader, avoiding loading
+// the full payload into memory.
+func (c *BlobClient) PutBlobFromReader(ctx context.Context, container, key string, reader io.Reader) error {
+	if container == "" || key == "" {
+		return errors.Errorf("container name and blob key are required")
+	}
+
+	blockBlob := c.client.ServiceClient().NewContainerClient(container).NewBlockBlobClient(key)
+	if _, err := blockBlob.UploadStream(ctx, reader, nil); err != nil {
+		return WrapError(err, "uploading blob "+container+"/"+key)
```
yhakbar: We don't need a helper for wrapping errors like this. We can just call `fmt.Errorf` inline here.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	return nil
+}
+
+// DeleteBlob deletes the named blob. Missing blobs return nil.
+func (c *BlobClient) DeleteBlob(ctx context.Context, container, key string) error {
```
yhakbar: Naming this `EnsureBlobDeleted` more clearly communicates that this function is a no-op when the blob is already gone.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+	if container == "" || key == "" {
+		return errors.Errorf("container name and blob key are required")
```
yhakbar: This should be a proper error type so it can be matched with `errors.As`.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	_, err := c.client.DeleteBlob(ctx, container, key, nil)
+	if err == nil || IsNotFound(err) {
```
yhakbar: This is a course error handle. Can it be made more granular?

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+		return nil
+	}
+
+	return WrapError(err, "deleting blob "+container+"/"+key)
```
yhakbar: We don't need a helper for wrapping errors like this. We can just call `fmt.Errorf` inline here.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+}
+
+// GetObject downloads a blob from the bound container by key. Convenience
+// wrapper for callers that have already bound a container via
+// BindContainer (e.g. PR 3's dependency-fetch path).
```
yhakbar: Please remove these references to future PRs. Comments should be written such that they are relevant well into the future.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+func (c *BlobClient) GetObject(ctx context.Context, key string) (io.ReadCloser, error) {
+	if c.container == "" {
+		return nil, errors.Errorf("BlobClient has no container bound; call BindContainer first or use GetBlob")
```
yhakbar: This should be a proper error type so it can be matched with `errors.Is`.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+	return c.GetBlob(ctx, c.container, key)
+}
+
+// ListBlobs returns the keys of all blobs in container whose names start
+// with prefix. Pass an empty prefix to enumerate the entire container.
+// Pages through ListBlobsFlat results; the full set is materialised in
+// memory, so callers should expect O(N) memory in the number of blobs.
+func (c *BlobClient) ListBlobs(ctx context.Context, container, prefix string) ([]string, error) {
+	if container == "" {
+		return nil, errors.Errorf("container name is required")
```
yhakbar: This should be a proper error type so it can be matched with `errors.Is`.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	cc := c.client.ServiceClient().NewContainerClient(container)
+
+	opts := &azblobcontainer.ListBlobsFlatOptions{}
+	if prefix != "" {
```
yhakbar: Avoid relying on empty sentinels to signal intent. Use a variadic option to allow for customization like this.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	pager := cc.NewListBlobsFlatPager(opts)
+
+	var out []string
+
+	for pager.More() {
+		page, err := pager.NextPage(ctx)
+		if err != nil {
+			return nil, WrapError(err, "listing blobs in "+container)
```
yhakbar: We don't need a helper for wrapping errors like this. We can just call `fmt.Errorf` inline here.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+// the server-side StartCopyFromURL API. Both blobs must live in the same
+// storage account that this client is bound to.
+//
+// The copy is initiated synchronously (StartCopyFromURL returns once Azure
+// accepts the request) but Azure may complete the copy asynchronously for
+// large blobs; this method does not poll for completion. Callers needing
+// to block on completion should poll the destination blob's CopyStatus.
+func (c *BlobClient) CopyBlob(ctx context.Context, srcContainer, srcKey, dstContainer, dstKey string) error {
+	if srcContainer == "" || srcKey == "" || dstContainer == "" || dstKey == "" {
+		return errors.Errorf("source and destination container/key are required")
```
yhakbar: This should be a proper error type so it can be matched with `errors.As`.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	srcURL := fmt.Sprintf("https://%s.blob.%s/%s/%s",
+		c.accountName, c.endpointSuffix, srcContainer, srcKey)
+
+	dst := c.client.ServiceClient().NewContainerClient(dstContainer).NewBlobClient(dstKey)
+	if _, err := dst.StartCopyFromURL(ctx, srcURL, nil); err != nil {
+		return WrapError(err, fmt.Sprintf("copying blob %s/%s to %s/%s", srcContainer, srcKey, dstContainer, dstKey))
```
yhakbar: We don't need a helper for wrapping errors like this. We can just call fmt.Errorf inline here.

internal/azurehelper/blob_test.go
```diff
@@ -0,0 +1,422 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+
+	"github.com/gruntwork-io/terragrunt/internal/azurehelper"
+	"github.com/gruntwork-io/terragrunt/pkg/log"
+)
+
+func TestNewBlobClient_NilConfig(t *testing.T) {
+	t.Parallel()
+
+	if _, err := azurehelper.NewBlobClient(context.Background(), nil, ""); err == nil {
```
yhakbar: Prefer `t.Context()` over `context.Background()` in tests.

internal/azurehelper/blob_test.go
```diff
@@ -0,0 +1,422 @@
...
+		t.Fatal("expected error for nil config")
```
yhakbar: Prefer `require.Error`

internal/azurehelper/blob_test.go
```diff
@@ -0,0 +1,422 @@
...
+func TestNewBlobClient_MissingAccountName(t *testing.T) {
+	t.Parallel()
+
+	_, err := azurehelper.NewBlobClient(context.Background(), &azurehelper.AzureConfig{
+		Method:        azurehelper.AuthMethodSasToken,
+		SasToken:      testSASToken,
+		ClientOptions: azcore.ClientOptions{Cloud: cloud.AzurePublic},
+	}, "")
+	if err == nil {
+		t.Fatal("expected error for missing storage account name")
```
yhakbar: Prefer `require.Error`

internal/azurehelper/blob_test.go
```diff
@@ -0,0 +1,422 @@
...
+func TestNewBlobClient_NoCredentialForTokenMethod(t *testing.T) {
+	t.Parallel()
+
+	_, err := azurehelper.NewBlobClient(context.Background(), &azurehelper.AzureConfig{
+		Method:        azurehelper.AuthMethodAzureAD,
+		AccountName:   testAccount,
+		ClientOptions: azcore.ClientOptions{Cloud: cloud.AzurePublic},
+	}, "")
+	if err == nil {
+		t.Fatal("expected error when token-method config has no credential")
```
yhakbar: Prefer `require.Error`

internal/azurehelper/blob_test.go
```diff
@@ -0,0 +1,422 @@
...
+	// A user that asks for OIDC auth but never wired a token source through
+	// the builder lands here with Method=OIDC and a nil Credential. The
+	// blob constructor must reject this rather than panic dereferencing.
+	_, err := azurehelper.NewBlobClient(context.Background(), &azurehelper.AzureConfig{
+		Method:        azurehelper.AuthMethodOIDC,
+		AccountName:   testAccount,
+		ClientOptions: azcore.ClientOptions{Cloud: cloud.AzurePublic},
+	}, "")
+	if err == nil {
+		t.Fatal("expected error when OIDC config has no credential")
```
yhakbar: Prefer `require.Error`

internal/azurehelper/blob_test.go
```diff
@@ -0,0 +1,422 @@
...
+
+	c, err := azurehelper.NewBlobClient(context.Background(), &azurehelper.AzureConfig{
+		Method:        azurehelper.AuthMethodSasToken,
+		SasToken:      "?sv=2023-01-01&sig=x",
+		AccountName:   testAccount,
+		CloudConfig:   cloud.AzurePublic,
+		ClientOptions: azcore.ClientOptions{Cloud: cloud.AzurePublic},
+	}, "")
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
```
yhakbar: Prefer `require.NoError`

internal/azurehelper/blob_test.go
```diff
@@ -0,0 +1,422 @@
...
+	}
+
+	if c.AccountName() != testAccount {
```
yhakbar: Prefer `assert.Equal`

internal/azurehelper/blob_test.go
```diff
@@ -0,0 +1,422 @@
...
+		t.Errorf("AccountName = %q, want %q", c.AccountName(), testAccount)
+	}
+
+	if c.AzClient() == nil {
```
yhakbar: Prefer `assert.NotNil`.

internal/azurehelper/blob_test.go
```diff
@@ -0,0 +1,422 @@
...
+
+	if _, err := azurehelper.NewBlobClient(context.Background(), nil, ""); err == nil {
+		t.Fatal("expected error for nil config")
+	}
+}
+
+func TestNewBlobClient_MissingAccountName(t *testing.T) {
+	t.Parallel()
+
+	_, err := azurehelper.NewBlobClient(context.Background(), &azurehelper.AzureConfig{
```
yhakbar: Prefer `t.Context()` over `context.Background()` in tests.

internal/azurehelper/blob_test.go
```diff
@@ -0,0 +1,422 @@
...
+	}, "")
+	if err == nil {
+		t.Fatal("expected error for missing storage account name")
+	}
+}
+
+func TestNewBlobClient_NoCredentialForTokenMethod(t *testing.T) {
+	t.Parallel()
+
+	_, err := azurehelper.NewBlobClient(context.Background(), &azurehelper.AzureConfig{
```
yhakbar: Prefer `t.Context()` over `context.Background()` in tests.

internal/azurehelper/blob_test.go
```diff
@@ -0,0 +1,422 @@
...
+		t.Fatal("expected error when token-method config has no credential")
+	}
+}
+
+func TestNewBlobClient_OIDCMissingTokenSource(t *testing.T) {
+	t.Parallel()
+	// A user that asks for OIDC auth but never wired a token source through
+	// the builder lands here with Method=OIDC and a nil Credential. The
+	// blob constructor must reject this rather than panic dereferencing.
+	_, err := azurehelper.NewBlobClient(context.Background(), &azurehelper.AzureConfig{
```
yhakbar: Prefer `t.Context()` over `context.Background()` in tests.

internal/azurehelper/blob_test.go
```diff
@@ -0,0 +1,422 @@
...
+	}, "")
+	if err == nil {
+		t.Fatal("expected error when OIDC config has no credential")
+	}
+}
+
+func TestNewBlobClient_SasToken(t *testing.T) {
+	t.Parallel()
+
+	c, err := azurehelper.NewBlobClient(context.Background(), &azurehelper.AzureConfig{
```
yhakbar: Prefer `t.Context()` over `context.Background()` in tests.

internal/azurehelper/blob_test.go
```diff
@@ -0,0 +1,422 @@
...
+
+	if c.AzClient() == nil {
+		t.Error("AzClient() is nil")
+	}
+}
+
+func TestNewBlobClient_AccessKey(t *testing.T) {
+	t.Parallel()
+
+	c, err := azurehelper.NewBlobClient(context.Background(), &azurehelper.AzureConfig{
```
yhakbar: Prefer `t.Context()` over `context.Background()` in tests.

internal/azurehelper/blob_test.go
```diff
@@ -0,0 +1,422 @@
...
+
+	if c.AccountName() != testAccount {
+		t.Errorf("AccountName = %q", c.AccountName())
+	}
+}
+
+func TestNewBlobClient_AccessKeyInvalidBase64(t *testing.T) {
+	t.Parallel()
+
+	_, err := azurehelper.NewBlobClient(context.Background(), &azurehelper.AzureConfig{
```
yhakbar: Prefer `t.Context()` over `context.Background()` in tests.

internal/azurehelper/blob_test.go
```diff
@@ -0,0 +1,422 @@
...
+	}, "")
+	if err == nil {
+		t.Fatal("expected error for invalid base64 access key")
+	}
+}
+
+func TestNewBlobClient_EndpointSuffixOverride(t *testing.T) {
+	t.Parallel()
+
+	_, err := azurehelper.NewBlobClient(context.Background(), &azurehelper.AzureConfig{
```
yhakbar: Prefer `t.Context()` over `context.Background()` in tests.

yhakbar:
@omattsson ,

I'm going to pause here, as I think there's significant rework that needs to be done, and a large portion of it should be pattern matching based on the feedback I've given so far. Please have all the comments addressed, and look to ensure that all the other instances of similar issues in the PR are resolved before re-requesting review.

omattsson:
Thanks for the thorough review, @yhakbar.

Rebased onto post-#6182 main and addressed the feedback in seven commits. Your blob.go:21 comment about `internal/errors` was the prompt — once rebased, the package required the migration anyway, so it became the anchor for a package-wide cleanup. Patterns follow your own merged work (`internal/cas/errors.go`, `internal/cas/cas.go`) and the #6182 conventions.

Mapping comments to commits:

- **Errors → stdlib + sentinels + typed structs** (`6cc0036b`). Drops `WrapError` and `ClassifyError`; uses `fmt.Errorf("…: %w", err)` with lowercase first-letter wrapping. Exported `Err*` sentinels for input validation (`ErrAzureConfigRequired`, `ErrContainerNameRequired`, `ErrBlobKeyRequired`, …). Typed structs for data-carrying errors that callers may want to `errors.As`: `CredentialMissingError`, `UnsupportedAuthMethodError`, `InvalidPrincipalIDError`, `UnknownCloudEnvironmentError`, `UnknownAccessTierError`. `errors.go` now keeps only `IsNotFound` (used at 7 production call sites) and `IsRetryable` predicates; `ClassifyError`'s body is inlined into `IsRetryable` as private.
- **Coarse `IsNotFound` at the three sites you flagged** (`6cc0036b`). `ContainerExists`, `DeleteContainer` and `EnsureBlobDeleted` now match the specific `azcore.ResponseError.ErrorCode` (`ContainerNotFound` / `BlobNotFound`) at the call site, so a "blob not found" can no longer silently pass as "container not found".
- **Forward-looking "PR 3" references removed** (`7a2384fd`).
- **`t.Context()` + testify in `blob_test.go`** (`25471ad2`). File-scoped: all `context.Background()` (19 sites) → `t.Context()` and raw `t.Fatal`/`t.Errorf` → `require`/`assert`. One `context.Background()` retained inside a `t.Cleanup` block (with an explanatory comment) because `t.Context()` is cancelled by cleanup time.
- **Blob URL construction via `net/url`** (`afcb6b5f`). Both the service URL and the SAS URL go through `url.URL{Scheme, Host, RawQuery}` instead of `fmt.Sprintf` + `"?"`+sas. The `CopyBlob` source URL is now percent-encoded via `url.URL.Path`.
- **Auth dispatch extracted to private helper** (`59e9535b`). `NewBlobClient`'s switch with mutable `(client, err)` moved into `newAzblobClient(cfg, suffix)` returning `(*azblob.Client, error)`.
- **`Ensure*` renames + cleaner `NewBlobClient` + `ListBlobs` options** (`c6ab418b`).
  - `BlobClient.CreateContainerIfNecessary` → `EnsureContainer`.
  - `BlobClient.DeleteBlob` → `EnsureBlobDeleted` (already idempotent on `BlobNotFound`; the rename makes the contract obvious).
  - `ResourceGroupClient.CreateIfNecessary` → `EnsureResourceGroup`.
  - `NewBlobClient(_ context.Context, cfg, endpointSuffix string)` → `NewBlobClient(cfg)`. The unused `ctx` is gone; `endpointSuffix` was an empty-string sentinel that fell back to `endpointSuffixForCloud(cfg)`, so it's always derived from `cfg.CloudConfig` now. `BuildBlobClient` (the only in-tree caller) always populates `CloudConfig`, so this is internally safe.
  - `ListBlobs(ctx, container, prefix string)` → `ListBlobs(ctx, container, opts ...ListBlobsOption)` with `WithPrefix(string)`, matching the `cas.Clone`/`WithDir`/`WithDepth` shape.
- **Self-review follow-ups** (`b04e9de1`). Stale `ListBlobsBound` references in two doc comments removed; `CopyBlob`'s percent-encoding contract documented; `WithPrefix` unit test strengthened.

Build, `go test -tags=azure`, and `golangci-lint run --build-tags=azure` are clean locally.

`pr3-azurerm-backend` consumes the renamed methods + the simpler `NewBlobClient(cfg)` signature; I'll rebase that branch on top of this once it merges (verified blast radius: 7 call sites in one file).

CI hasn't run on this revision — looks like the fork-PR workflow-approval gate re-armed after the force-push. Happy to wait if you'd like to skim the diff first, or ping me when you want it kicked off. I'll resolve the individual threads as the per-cluster anchors get addressed.

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	_, err := c.client.CreateContainer(ctx, name, nil)
+	if err == nil {
+		return nil
+	}
+
+	var respErr *azcore.ResponseError
+
+	if errors.As(err, &respErr) && strings.EqualFold(respErr.ErrorCode, "ContainerAlreadyExists") {
```
denis256: Make sense to use `errors.AsType`?

internal/azurehelper/blob_internal_test.go
```diff
@@ -0,0 +1,25 @@
...
+// listBlobsOptions.prefix field that ListBlobs reads into
+// azblobcontainer.ListBlobsFlatOptions.Prefix.
+func TestWithPrefix(t *testing.T) {
+	t.Parallel()
+
+	o := &listBlobsOptions{}
+	WithPrefix("state/")(o)
+	assert.Equal(t, "state/", o.prefix)
+
+	// WithPrefix("") clears any previously-set prefix.
```
denis256: Do we need this?

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,389 @@
...
+// dependency-fetch path that asks for a state file by key) do not need to
+// repeat the container on every call.
+//
+// BlobClient is not safe for concurrent use when callers invoke
+// BindContainer: the bound container is stored on the receiver. Construct
+// one client per goroutine, or call BindContainer once during setup before
+// fanning out.
+type BlobClient struct {
+	client         *azblob.Client
+	config         *AzureConfig
```
denis256: Unused field, write-only access

internal/azurehelper/rbac.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/gruntwork-io/terragrunt/pkg/log"
+)
+
+// Built-in Azure role definition IDs (GUIDs). These are stable across all
+// subscriptions. The full role definition ID is
+// /subscriptions/{sub}/providers/Microsoft.Authorization/roleDefinitions/{guid}.
+//
+// See: https://learn.microsoft.com/en-us/azure/role-based-access-control/built-in-roles
+const (
+	RoleStorageBlobDataOwner       = "b7e6dc6d-f1e8-4753-8033-0f276bb0955b"
```
denis256: Constants defined but not used

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	srcURL := (&url.URL{
+		Scheme: "https",
+		Host:   fmt.Sprintf("%s.blob.%s", c.accountName, c.endpointSuffix),
+		Path:   "/" + srcContainer + "/" + srcKey,
+	}).String()
+
+	dst := c.client.ServiceClient().NewContainerClient(dstContainer).NewBlobClient(dstKey)
+	if _, err := dst.StartCopyFromURL(ctx, srcURL, nil); err != nil {
```
denis256: Function StartCopyFromURL() initiates the copy synchronously, but Azure may complete it asynchronously. 

I think we may need to pool destination blob CopyStatus.

internal/azurehelper/rbac.go
```diff
@@ -0,0 +1,284 @@
...
+			}
+
+			if _, err := c.client.DeleteByID(ctx, *ra.ID, nil); err != nil {
+				// Concurrent removal raced us to the same assignment - treat as success.
+				if IsNotFound(err) {
+					continue
+				}
+
+				if firstErr == nil {
+					firstErr = fmt.Errorf("deleting role assignment %s: %w", *ra.ID, err)
```
denis256: Why do we save only the first error? can be a different error on each removal

internal/azurehelper/config.go
```diff
@@ -0,0 +1,431 @@
...
+
+		if v := os.Getenv(k); v != "" {
+			return v
+		}
+	}
+
+	return ""
+}
+
+func parseBool(s string) bool {
```
denis256, line 392: Why can't it be used strconv.ParseBool()?

internal/azurehelper/storage_account.go
```diff
@@ -0,0 +1,402 @@
...
+
+	if in.ReplicationType == "" {
+		in.ReplicationType = defaultReplicationType
+	}
+
+	if in.AccessTier == "" {
+		in.AccessTier = defaultAccessTier
+	}
+
+	tags := make(map[string]string, len(in.Tags)+1)
```
denis256: can be used maps.Clone()?

internal/azurehelper/blob.go
omattsson: Good call — switched to `errors.AsType[*azcore.ResponseError]`. Applied the same migration to the other `errors.As` sites in this package (errors.go, rbac.go, storage_account.go) for consistency. db18cf3

internal/azurehelper/blob.go
omattsson: Removed — the `config` field was write-only. db18cf3

internal/azurehelper/blob.go
omattsson: Agreed. `CopyBlob` now polls the destination blob's `CopyStatus` (via `GetProperties`) every 1s until it's no longer `Pending`, so a nil return means the copy is committed. Terminal `failed`/`aborted` states return an error with the SDK's status description, and the poll honours `ctx` so callers can bound the wait with a deadline. db18cf3

internal/azurehelper/blob_internal_test.go
omattsson: Removed — it was just re-asserting the empty default, no real coverage. db18cf3

internal/azurehelper/rbac.go
omattsson: Removed the unused `RoleStorageBlobData{Owner,Contributor,Reader}` constants and dropped the stale reference from the `RoleDefinitionID` doc comment. The azurerm backend in PR 3 will reintroduce whichever role ID it actually needs. db18cf3

internal/azurehelper/rbac.go
omattsson: Good point — now collecting every per-assignment deletion error into a slice and returning `errors.Join(errs...)` so no failures are masked. db18cf3

internal/azurehelper/config.go
omattsson: It can — `parseBool` now delegates to `strconv.ParseBool(strings.TrimSpace(s))`. This matches the Terraform azurerm provider's convention for `ARM_USE_MSI` / `ARM_USE_OIDC`. db18cf3

internal/azurehelper/storage_account.go
omattsson: Yes — switched to `maps.Clone(in.Tags)` with a nil-guard before inserting the `created-by` tag. db18cf3

internal/azurehelper/storage_account.go
```diff
@@ -0,0 +1,403 @@
...
+
+// EnableVersioning turns on blob versioning for the account's default
+// blob service. Calling on an account that already has versioning
+// enabled is a no-op from the caller's perspective.
+func (c *StorageAccountClient) EnableVersioning(ctx context.Context, l log.Logger) error {
+	l.Debugf("azurehelper: enabling blob versioning on %q", c.accountName)
+
+	props := armstorage.BlobServiceProperties{
+		BlobServiceProperties: &armstorage.BlobServicePropertiesProperties{
+			IsVersioningEnabled: to.Ptr(true),
```
denis256: Not sure if this will work as expected
If we call SetServiceProperties() with partial settings, it will disable all other fields

internal/azurehelper/blob.go
```diff
@@ -0,0 +1,458 @@
...
+// can rely on a nil return meaning the copy is committed. The poll honours
+// ctx; pass a context with a deadline to bound the wait.
+func (c *BlobClient) CopyBlob(ctx context.Context, srcContainer, srcKey, dstContainer, dstKey string) error {
+	if srcContainer == "" || srcKey == "" || dstContainer == "" || dstKey == "" {
+		return ErrCopyBlobArgsRequired
+	}
+
+	srcURL := (&url.URL{
+		Scheme: "https",
+		Host:   fmt.Sprintf("%s.blob.%s", c.accountName, c.endpointSuffix),
```
denis256: I think this will work only for public URLs with a token in the URL, for a private container, it will fail.

internal/azurehelper/config.go
```diff
@@ -0,0 +1,433 @@
...
+		return nil, err
+	}
+
+	return NewStorageAccountClient(cfg)
+}
+
+// applyEnvFallbacks fills empty fields on cfg from environment variables
+// (the builder's env map first, then the process environment). Mirrors the
+// ARM_* and AZURE_* names used by Terraform's azurerm backend.
+func (b *AzureConfigBuilder) applyEnvFallbacks(cfg *AzureSessionConfig) {
```
denis256, line 310: Should additional environment variables be used for the account and location?
```
ARM_STORAGE_ACCOUNT_NAME
AZURE_STORAGE_ACCOUNT
ARM_LOCATION
AZURE_LOCATION
```

internal/azurehelper/rbac.go
```diff
@@ -0,0 +1,290 @@
...
+	// RoleDefinitionID is the GUID portion of a built-in or custom role
+	// definition. The full definition ID is composed for you.
+	RoleDefinitionID string
+}
+
+// AssignRole creates a role assignment. If an assignment satisfying the same
+// (Scope, PrincipalID, RoleDefinitionID) tuple already exists, this returns
+// nil — the operation is idempotent.
+func (c *RBACClient) AssignRole(ctx context.Context, l log.Logger, in AssignRoleInput) error {
+	if in.Scope == "" || in.PrincipalID == "" || in.RoleDefinitionID == "" {
```
denis256, line 91: Should the RoleDefinitionID be checked if it is a UUID?

omattsson: Good catch — done in b5dfc914. RoleDefinitionID now goes through `uuid.Parse` in `AssignRole`, `AssignRoleIfMissing`, `HasRoleAssignment`, and `RemoveRole`. Validation failure surfaces as a new typed `InvalidRoleDefinitionIDError`, mirroring the existing `InvalidPrincipalIDError` pattern. Tests cover the rejection at each of the four entry points.

internal/azurehelper/config.go
omattsson: Added in b5dfc914. `applyEnvFallbacks` now fills `StorageAccountName` from `ARM_STORAGE_ACCOUNT_NAME` / `AZURE_STORAGE_ACCOUNT` and `Location` from `ARM_LOCATION` / `AZURE_LOCATION`. Precedence follows the existing pattern (ARM_* first, then AZURE_*) — matches Terraform's azurerm provider.

While wiring this up I also propagated `Location` onto `AzureConfig` itself — without that, the resolved value never reached any caller, which would have made the env var a footgun. Storage account name was already plumbed through.

internal/azurehelper/blob.go
omattsson: Good catch — fixed in b5dfc914. You were right that the previous code only worked against public blobs / SAS URLs.

The root cause is a Go SDK gap: `StartCopyFromURL`'s public options struct doesn't expose `CopySourceAuthorization`, even though the underlying REST API supports `x-ms-copy-source-authorization` for OAuth source reads. The only way to set that header through the SDK is `CopyFromURL` (synchronous Copy Blob From URL).

So `CopyBlob` now uses `CopyFromURL` and dispatches source-URL auth per `AuthMethod`:

- **SAS token** — SAS query string appended to the source URL.
- **Access key** — same-account copy is authorized by the destination request's shared-key signature; source URL needs no extra auth.
- **OAuth methods** (service principal / OIDC / MSI / Azure AD) — a bearer token is acquired for the cloud-appropriate storage audience and passed via `CopySourceAuthorization`. The audience comes from a new `storageTokenScopeForSuffix` helper that mirrors `endpointSuffixForCloud`, so China cloud gets `storage.chinacloudapi.cn` instead of `storage.azure.com`.

The size limit drops from 5 TB (async copy) to 256 MiB (Copy Blob From URL) — orders of magnitude above any state file. Documented in the function comment.

Polling and `waitForCopy` are gone — `CopyFromURL` is synchronous and either succeeds or fails on the request.

New unit tests in `blob_internal_test.go` cover all three branches plus the per-cloud audience selection.

internal/azurehelper/storage_account.go
omattsson: Confirmed concern — fixed in b5dfc914. You're right that `SetServiceProperties` does not behave like a PATCH: the Azure docs warn that values omitted from the request body "may have unexpected behavior", and with the previous code, `EnableSoftDelete` would have stomped on the versioning flag set by `EnableVersioning` during Create (and on anything else set out of band — CORS rules, change feed, …).

Both methods now go through a shared `updateBlobServiceProperties(ctx, label, mutate)` helper that:

1. Does a `GetServiceProperties` to read the current state.
2. Runs the caller's `mutate` over the inner `BlobServicePropertiesProperties`.
3. Skips the PUT when `mutate` returns `false` (idempotent short-circuit — `EnableVersioning` re-runs become a single GET; `EnableSoftDelete` re-runs with the same retention also become a single GET).
4. Sends only the writable inner properties on the PUT — the outer struct's ARM-populated read-only fields (`ID`, `Name`, `Type`, `SKU`) are not echoed back.

New test `TestStorageAccount_EnableVersioning_PreservesOtherFields` asserts the wire-level behaviour: with CORS + change-feed configured by something else, the PUT body still carries them after `EnableVersioning` runs. Symmetric `_SkipsPutWhenAlreadyEnabled` / `_SkipsPutWhenAlreadyConfigured` tests prove the no-op path issues zero PUTs.

denis256:
We also need to fix merge conflicts with go.mod

omattsson:
Merged `upstream/main` and resolved in 244395e93. Conflict was in `go.mod`:

- The Azure SDK additions (azcore, azidentity, armauthorization, armresources, armstorage, azblob) ended up in the same `require` block that upstream had bumped on the AWS SDK side. Kept all the Azure entries; took upstream's newer `aws-sdk-go-v2/*` versions and `smithy-go v1.27.2`.
- Indirect deps: kept `microsoft-authentication-library-for-go v1.7.2` (newer; pulled by the azidentity addition) and took upstream's newer `opentelemetry-operations-go` v0.57 / v1.33 line.
- `go mod tidy` ran clean afterwards; `go.sum` regenerated.

Verified locally:
- `go build ./...`
- `go test -tags=azure -count=1 ./internal/azurehelper/... ./internal/remotestate/backend/azurerm/...`
- `go vet -tags=azure ./test/...`
- `golangci-lint run --build-tags=azure ./internal/azurehelper/... ./internal/remotestate/backend/azurerm/...`

All four green.

denis256:
I cant even start CICD to validate changes, not all commits are signed
```
remote:   Found 26 violations (showing first 10):
remote:
remote:   244395e93184446b3f55bc58918cc93d0982b552
remote:   b5dfc914254a1bdcf1c6d5d0667d5aafb951e681
remote:   bc95ba4d500bddfee32c8941f9deed247cf9f20a
remote:   0c2d94930961a94f0568e9dd525b50c9f7af637d
remote:   db18cf3ca90336f1a6599ed1da618348560580da
remote:   bba4d33c8a5cfd0a22341ea3efae868fefe38ba3
remote:   72a8dfebf4fccab317bd290716124e2148f6688e
remote:   b04e9de1e10f22661b8783abd7fdb41fe9e229ba
remote:   c6ab418bf155eeed24ba73517c0f1f6e4ee7810e
remote:   59e9535b841f20d536dd36fd278d064d6999264a
```

omattsson:
@denis256 Thanks for the heads-up on the signature requirement. All 26 commits are now signed and showing as **Verified** — I re-signed the existing history with an SSH signing key (no content changes; the trees are byte-identical to before, only signatures were added). The two `main` merge commits were preserved and their `go.mod`/`go.sum` resolutions are unchanged.

CICD should be unblocked now. Let me know if you'd like anything else before kicking it off.

yhakbar:
Thanks for opening this, @omattsson ,

This was addressed by #6428

PR #6024, opened 2026-05-04
fix: respect TG_DOWNLOAD_DIR for dependency blocks

### LINKED ISSUE #6021: TG_DOWNLOAD_DIR is not respected for indirect dependencies
## Describe the bug

When running Terragrunt with a custom `TG_DOWNLOAD_DIR` (e.g. `TG_DOWNLOAD_DIR=/Users/me/.terragrunt.cache`), all modules that are part of the current execution scope correctly create their `.terragrunt-cache` under the custom directory. However, if any of those modules have `dependency` blocks pointing to modules **outside** the current execution scope (e.g. a sibling module at `../../../other_team/vnet`), those external dependencies ignore `TG_DOWNLOAD_DIR` entirely and create `.terragrunt-cache` inside their own source directory.

This means:
- In-scope modules → cached under `TG_DOWNLOAD_DIR`
- Out-of-scope dependency modules → cached next to their `terragrunt.hcl` source

This defeats the purpose of `TG_DOWNLOAD_DIR`, which is commonly used to centralise the cache (e.g. on a faster disk, a RAM disk, or a path that is easy to clean up).

## Steps To Reproduce

**Directory layout:**

```
infra/
  networking/
    vnet/
      terragrunt.hcl      # <-- dependency target (outside run scope)
  apps/
    web/
      terragrunt.hcl      # <-- references dependency on ../../networking/vnet
    api/
      terragrunt.hcl
```

**`infra/apps/web/terragrunt.hcl`:**

```hcl
dependency "vnet" {
  config_path = "../../networking/vnet"
}

inputs = {
  vnet_id = dependency.vnet.outputs.vnet_id
}
```

**Run command (from `infra/apps/`):**

```bash
export TG_DOWNLOAD_DIR=/tmp/tg-cache

terragrunt run-all plan \
  --non-interactive \
  --provider-cache \
  --provider-cache-dir ~/.terraform.d/plugin-cache
```

**Observe:**

```
/tmp/tg-cache/
  <hash>/apps/web/.terragrunt-cache/   ✅  (in-scope module)
  <hash>/apps/api/.terragrunt-cache/   ✅  (in-scope module)

infra/networking/vnet/
  .terragrunt-cache/                   ❌  (dependency, should be in /tmp/tg-cache)
```

## Expected behavior

All `.terragrunt-cache` directories — including those created while resolving `dependency` block outputs — should be created under `TG_DOWNLOAD_DIR`. Running the same command twice should not recreate the cache directories under module source paths.

## Versions

- Terragrunt version: 1.0.3 (Issue probably implemented post v0.77.x)
- OpenTofu/Terraform version: any
- Environment details: macOS

## Additional context

Root cause: two locations in `pkg/config/dependency.go` (`decodeDependencies` and `getTerragruntOutputJSON`) unconditionally override `pctx.DownloadDir` with the dependency's local default (`<dep_source_dir>/.terragrunt-cache`), even when the caller is already using a custom directory. The fix computes the caller's default download dir first; the override is skipped when the caller's `DownloadDir` differs from that default (i.e. when the user set a custom one).

maonat:
## Description

Fixes #6021.
 
When `TG_DOWNLOAD_DIR` is set, modules inside the execution scope correctly place
their `.terragrunt-cache` under the custom directory. However, `dependency` blocks
and `read_terragrunt_config()` calls that switch the parsing context to a different
config path would inherit the calling module's `DownloadDir` unchanged — and code
in `dependency.go` would then unconditionally override it with the dependency's local
default (next to its `terragrunt.hcl`), bypassing the user-set path entirely.

**Root cause:** `WithConfigPath` cloned the context without keeping `DownloadDir` in
sync with the new config path. This forced every call site that switches config paths
to carry its own boilerplate to detect and preserve custom dirs.

**Fix:** `WithConfigPath` now checks whether `DownloadDir` equals the *old* config
path's computed default before cloning. If it does, `DownloadDir` is updated to the
*new* config path's default. A user-set custom dir never matches any module's computed
default, so it is always preserved unchanged. Because every context switch — `dependency`
blocks, `read_terragrunt_config()`, and any future callers — goes through `WithConfigPath`,
the fix is applied uniformly with no per-call-site boilerplate needed.

## TODOs

Read the [Gruntwork contribution 
guidelines](https://gruntwork.notion.site/Gruntwork-Coding-Methodology-02fdcd6e4b004e818553684760bf691e).

- [X] I am submitting code based on open source software (e.g. MIT, MPL-2.0, Apache)
- [X] I am adding or upgrading a dependency or adapted code and confirm it has a compatible open source 
license
- [X] Update the docs.
- [X] Update the changelog in the docs.
- [X] This change is backwards compatible.
- [X] If this change is not forwards compatible (e.g. a new feature), it is gated behind a feature flag.

## Summary by CodeRabbit

* **Bug Fixes**
  * Download directory now updates only when it previously matched the old module default, preventing unintended changes when switching configuration paths.
  * Dependency handling no longer overrides a caller’s custom download directory during cache or output retrieval, preserving user-specified download locations.

* **Tests**
  * Added tests to verify custom download dirs are preserved and module-default download dirs are applied only when appropriate.


maonat:
@coderabbitai review

maonat:
@denis256 I've re-created the PR and this should fix entirely the issue with the test.
This fix is really important because properly stores the .terragrunt-cache content in the correct directory and not in the dependency directory.

lauralex (approved):

lauralex:
This looks functionally right to me and matches the centralized fix for #6021.

A couple small polish suggestions before merge:

1. In `WithConfigPath`, consider normalizing both paths before comparison:

```go
if filepath.Clean(ctx.DownloadDir) == filepath.Clean(defaultDir) {
```

That keeps the default-vs-custom check robust if one side has `.` segments or separator/cleaning differences.

2. The PR checklist has docs/changelog checked, but the diff only includes code/tests. If maintainers expect changelog entries for bug fixes, adding one under the current changelog version would make the PR tidier.

None of these change the core approach; the fix itself looks equivalent to what I would expect for preserving a user-supplied `TG_DOWNLOAD_DIR` while still moving default cache dirs with the dependency context.

pkg/config/parsing_context.go
```diff
@@ -249,6 +250,18 @@ func (ctx *ParsingContext) WithConfigPath(l log.Logger, configPath string) (log.
...
 	c := ctx.Clone()
+
+	// Keep DownloadDir in sync with TerragruntConfigPath: if the current context was
+	// using the default download dir for the old config, update it to the default for
+	// the new config. This ensures that when read_terragrunt_config() or dependency
+	// processing switches to a different module, DownloadDir reflects the new module's
+	// default rather than an inherited stale default from an ancestor. User-set custom
+	// dirs (which won't match any module's default) are preserved unchanged.
+	_, defaultDir := util.DefaultWorkingAndDownloadDirs(ctx.TerragruntConfigPath)
+	if c.DownloadDir == defaultDir {
```
lauralex: Please normalize both sides before comparing the current download dir to the computed default. Without this, a semantically default path containing `.` segments or other unclean path forms can be misclassified as custom and then fail to move with the dependency context.

```suggestion
	if filepath.Clean(c.DownloadDir) == filepath.Clean(defaultDir) {
```

lauralex (changes requested):
Requesting changes for the small polish items from my earlier comment so they are tracked inline. The core fix looks right; these should be quick cleanup before merge.

maonat:
> This looks functionally right to me and matches the centralized fix for #6021.
> 
> A couple small polish suggestions before merge:
> 
> 1. In `WithConfigPath`, consider normalizing both paths before comparison:
> 
> ```go
> if filepath.Clean(ctx.DownloadDir) == filepath.Clean(defaultDir) {
> ```
> 
> That keeps the default-vs-custom check robust if one side has `.` segments or separator/cleaning differences.
> 
> 2. The PR checklist has docs/changelog checked, but the diff only includes code/tests. If maintainers expect changelog entries for bug fixes, adding one under the current changelog version would make the PR tidier.
> 
> None of these change the core approach; the fix itself looks equivalent to what I would expect for preserving a user-supplied `TG_DOWNLOAD_DIR` while still moving default cache dirs with the dependency context.

1. merged suggestion
2. If this was a feature I would've added the documentation, but cannot find any detail of this being expected... This is why this is a bug and needs to be fixed without adjusting the documentation

lauralex (approved):

yhakbar (approved):
Thanks!

PR #6027, opened 2026-05-05
Prevent duplicate getCredentialsFromEnv (fixes #5105)

### LINKED ISSUE #5105: 0.86.3 AWS Cross account assumption results in 403 on S3 backend bucket
## Describe the bug

Terragrunt is ran on Gitlab pipelines which in this instance use AWS Access keys for Govcloud to login as a certain user, then assumes a role in various AWS Govcloud accounts using this logic which loops over various folders and runs the plan in each `--working-dir`. 

``` 
   echo "$TG_STG_DIRS" | while read -r dir; do
      if [ -d "${dir}/stacks" ]; then
        cd ${dir}
        echo "Running a plan in directory: ${dir}"
        # Removing any cached files, stacks do not like anything to be prepopulated currently.
        terragrunt stack clean 
        terragrunt stack run plan --working-dir "${dir}/stacks" --log-format bare 
      else
        echo "Directory ${dir}/stacks does not exist, skipping."
      fi
    done
```
```
remote_state {
  backend = "s3"
  config = {
    encrypt        = true
    bucket         = "terragrunt-tf-state-${local.account_id}-${local.aws_region}"
    key            = "${path_relative_to_include()}/tf.tfstate"
    region         = local.aws_region
    dynamodb_table = "tf-locks"
    assume_role = {
      role_arn = "arn:${local.aws_partition}:iam::${local.account_id}:role/gitlab-assumable-terragrunt-role"
    }
  }
  generate = {
    path      = "backend.tf"
  }
}
```

I upgraded to 0.93.7 from 0.83.2 and on any account that does not use the access key I get an error similar to the one below ( I walked back from 0.93 -> 0.86 and the error message changes slightly but still have the 403) which happens right before the init messages usually appear. 

```
Cannot process unit Unit /builds/platform/spear/live/REDACTED/infra/stg/stacks/.terragrunt-stack/base/.terragrunt-stack/vpc-endpoints (excluded: false, assume applied: false, dependencies: [/builds/platform/spear/live/REDACTED/infra/stg/stacks/.terragrunt-stack/base/.terragrunt-stack/vpc]) because one of its dependencies, Unit /builds/platform/spear/live/REDACTED/infra/stg/stacks/.terragrunt-stack/base/.terragrunt-stack/vpc (excluded: false, assume applied: false, dependencies: []), finished with an error: operation error STS: GetCallerIdentity, https response error StatusCode: 403, RequestID: REDACTED, api error InvalidClientTokenId: The security token included in the request is invalid.
```
This workflow works on 0.86.2 > but appears to broken on any newer releases. This ONLY occurs if the new roles AWS accounts is different than the AWS access keys used. We also use IAM role assumption using an IRSA role on our runners for commercial accounts and it will switch accounts correctly. 

## Steps To Reproduce

I'm honestly not sure how to reproduce the bug, but it correlates with the [get_account_id() fix ](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.86.3) which occured in 0.86.3 so I'm thinking that may have something to do with it. 

## Expected behavior

The role is assumed correctly while the loops are occurring. 

## Must haves

- [ ] Steps for reproduction provided.

## Nice to haves

- [ ] Terminal output
- [ ] Screenshots

## Versions

- Terragrunt version: 0.86.3 < 
- OpenTofu/Terraform version: TOFU 1.10.5
- Environment details (Ubuntu 20.04, Windows 10, etc.): Linux debian bookworm slim container image running in a Gitlab pipeline. 

## Additional context

I can see that the role assumption is happening in Cloudtrail on the target account but it may not be occuring before the init phase.

martin31821:
## Description

Fixes #5105.

Prior to this change, getCredentialsFromEnv() was called twice instead of using the AWS config package. In my tests, this prevented the AssumeRoleARN to be respected correctly. In my tests and the unit tests, this didn't break any other tests.

## TODOs

- [ ] Update the docs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved AWS credential handling to allow IAM role assumption even when environment credentials are already detected, enabling proper credential resolution in scenarios requiring role assumption.

### DIFF
--- a/internal/awshelper/config.go
+++ b/internal/awshelper/config.go
@@ -131,10 +131,6 @@ func (b *AWSConfigBuilder) Build(ctx context.Context, l log.Logger) (aws.Config,
 		return aws.Config{}, errors.Errorf("Error loading AWS config: %w", err)
 	}
 
-	if createCredentialsFromEnv(b.env) != nil {
-		return cfg, nil
-	}
-
 	mergedIAMRoleOptions := getMergedIAMRoleOptions(b.sessionConfig, b.iamRoleOpts)
 	if mergedIAMRoleOptions.RoleARN == "" {
 		return cfg, nil


Steve-CrateDefense:
Thank you @martin31821!

yhakbar:
Hey @martin31821 ,

Do you mind adding a test that will fail without this change to prove that the change is necessary and a changelog entry documenting the fix? Let me know if you need any help with that.

PR #6028, opened 2026-05-05
chore: Addressing test flakes

yhakbar:
## Description

Fixes some flakiness in our tests.

Added `ClaimForRunning` to solve a race between marking a unit ready for runs and the unit exiting early due to `--fail-fast` behavior.

Cut `TestCASStorageDirectory` and `TestTerragruntStackCommandsWithSymlinks` because they don't actually test what they sound like, and aren't designed to run reliably.

## TODOs

- [x] Update the docs.


test/integration_test.go
```diff
@@ -1185,77 +1184,6 @@ func TestTerragruntStackCommandsWithPlanFile(t *testing.T) {
 	)
 }
 
-func TestTerragruntStackCommandsWithSymlinks(t *testing.T) {
```
denis256, line 1188: Removing TestTerragruntStackCommandsWithSymlinks drops the only end-to-end coverage of the symlink experiment.
We don't need it anymore?

denis256 (approved):

test/integration_test.go
yhakbar: We do need a test like this, but this doesn't test it well because it doesn't pass reliably. I can look to add a better test after merging this.

Issue #2233, opened 2022-08-11
Could not parse output from terragrunt config when AWS_CSM_ENABLED=true


LiamDeaconAntarcticaAM:
Hi, I've encountered an error when trying to run `terragrunt apply` with `AWS_CSM_ENABLED` environment variable set to `true`. I receive an error message:

```bash
> cd /the/path/to/terragrunt-projects/project-that-depends-on-cicd-setup
> terragrunt apply
ERRO[0002] Could not parse output from terragrunt config /the/path/to/terragrunt-projects/cicd-setup/terragrunt.hcl. Underlying error: invalid character '/' after top-level value 
ERRO[0002] Unable to determine underlying exit code, so Terragrunt will exit with error code 1
```  

However I don't encounter the error when I set:

```bash
# /bin/bash
> export AWS_CSM_ENABLED=false
> terragrunt apply  # provides plan with prompt to apply changes as expected
```

>NOTE: I ran `terragrunt apply --terragrunt-log-level debug --terragrunt-debug`, which provided some additional insight, which gave me the idea to try disabling AWS client side monitoring:

```bash
> terragrunt apply --terragrunt-log-level debug --terragrunt-debug
# -- snip --
DEBU[0000] run_cmd, cached output: [REDACTED]            prefix=[/the/path/to/terragrunt-projects/cicd-setup] 
DEBU[0000] Evaluated 1 locals (remaining 2): git_repo_name  prefix=[/the/path/to/terragrunt-projects/cicd-setup] 
DEBU[0000] Evaluated 1 locals (remaining 1): backend_bucket  prefix=[/the/path/to/terragrunt-projects/cicd-setup] 
DEBU[0000] Evaluated 1 locals (remaining 0): dynamodb_table  prefix=[/the/path/to/terragrunt-projects/cicd-setup] 
DEBU[0000] [Partial] Included config /the/path/to/terragrunt-projects/cicd-setup/terragrunt.hcl has strategy shallow merge: merging config in (shallow).  prefix=[/the/path/to/terragrunt-projects/cicd-setup/cicd-setup] 
DEBU[0000] Detected module /the/path/to/terragrunt-projects/cicd-setup/terragrunt.hcl is already init-ed. Retrieving outputs directly from working directory.  prefix=[/the/path/to/terragrunt-projects/cicd-setup] 
DEBU[0000] Running command: terraform output -json       prefix=[/the/path/to/terragrunt-projects/cicd-setup] 
[...]

PR #6029, opened 2026-05-05
fix: tolerate non-JSON warnings in tofu/terraform output -json (#6001)

### LINKED ISSUE #6001: Terraform 1.15 backend deprecation warning breaks dependency output parsing
## Describe the bug

Terragrunt fails to resolve `dependency` outputs when used with Terraform `1.15.0` and an S3 backend configured with the deprecated `dynamodb_table` lock setting.

The failure appears to happen because Terraform `1.15.0` emits a backend deprecation warning while Terragrunt is reading dependency outputs via JSON output. Terragrunt then attempts to parse output that is no longer pure JSON, fails with `invalid character ... after top-level value`, and the dependent module reports `There is no variable named "dependency"`.

The same configuration works with Terraform `1.14.9`.

## Steps To Reproduce

Minimal shape:

```hcl
# root.hcl
remote_state {
  backend = "s3"

  generate = {
    path      = "backend.tf"
    if_exists = "overwrite"
  }

  config = {
    bucket         = "example-terraform-state"
    key            = "${path_relative_to_include()}/terraform.tfstate"
    region         = "ap-southeast-2"
    encrypt        = true

    # Deprecated in newer Terraform, but still commonly used by Terragrunt projects.
    dynamodb_table = "example-terraform-state-locks"
  }
}
```

```hcl
# dependency/terragrunt.hcl
include {
  path = find_in_parent_folders("root.hcl")
}

terraform {
  source = "./module"
}
```

```hcl
# dependency/module/main.tf
output "example_value" {
  value = "hello"
}
```

```hcl
# dependent/terragrunt.hcl
include {
  path = find_in_parent_folders("root.hcl")
}

dependency "example" {
  config_path = "../dependency"
}

terraform {
  source = "./module"
}

inputs = {
  example_value = dependency.example.outputs.example_value
}
```

```hcl
# dependent/module/variables.tf
variable "example_value" {
  type = string
}
```

```hcl
# dependent/module/main.tf
resource "terraform_data" "example" {
  input = var.example_value
}
```

Reproduction steps:

1. Use Terraform `1.15.0`.
2. Use Terragrunt `1.0.3`.
3. Apply the `dependency` module so it has outputs in remote state.
4. Run `terragrunt plan` from the `dependent` module.

Observed error:

```text
ERROR Error: Unknown variable

  on dependent/terragrunt.hcl line <n>:
  <n>:   example_value = dependency.example.outputs.example_value

There is no variable named "dependency".

ERROR 2 errors occurred:

* Could not parse output from terragrunt config ../dependency/terragrunt.hcl.
  Underlying error: invalid character 'W' after top-level value

* dependent/terragrunt.hcl:<n>:<n>: Unknown variable;
  There is no variable named "dependency".
```

In some runs the invalid character was `'\x1b'` instead of `W`, depending on whether colored warning output was present.

Downgrading only Terraform from `1.15.0` to `1.14.9` makes the same Terragrunt configuration plan successfully.

## Expected behavior

Terragrunt should be able to read dependency outputs successfully, or should filter/handle Terraform warning output so that dependency JSON parsing is not broken by non-JSON warning text.

At minimum, `terraform output -json` warnings from Terraform should not cause the dependent Terragrunt configuration to fail with the misleading secondary error:

```text
There is no variable named "dependency".
```

## Must haves

## Nice to haves

- [x] Terminal output
- [ ] Screenshots

## Versions

- Terragrunt version: `1.0.3`
- OpenTofu/Terraform version:
  - Fails with Terraform `1.15.0`
  - Works with Terraform `1.14.9`
- Environment details:
  - Linux amd64
  - Terraform installed from HashiCorp apt packages
  - AWS S3 backend with DynamoDB lock table configured via `dynamodb_table`

## Additional context

This appears similar in shape to older Terragrunt issues where non-JSON stdout polluted dependency output parsing, for example AWS CSM output causing `Could not parse output from terragrunt config ... invalid character ... after top-level value`.

The difference here is that the non-JSON text appears to be Terraform `1.15.0` warning output for the deprecated S3 backend `dynamodb_table` parameter.

The relevant warning text is:

```text
Warning: Deprecated Parameter

The parameter "dynamodb_table" is deprecated. Use parameter "use_lockfile" instead.
```

The issue started after upgrading Terraform from `1.14.9` to `1.15.0`. No Terragrunt configuration changes were required to trigger it.

jpke:
## Description

Fixes #6001.

Terraform 1.15.0 emits a backend deprecation warning on stdout alongside the JSON payload of `output -json` (for example, for the deprecated S3 `dynamodb_table` parameter). Because the warning is printed *after* the JSON object, the long-standing leading-garbage workaround for AWS CSM logging (#2233) does not help, and resolving `dependency` outputs fails with:

```
invalid character 'W' after top-level value
```

…surfaced to the user as the misleading `There is no variable named "dependency"`.

This PR replaces the `bytes.IndexByte('{')` snippet in `getOutputJSONWithCaching` with a small helper, `extractFirstJSONObject`, that:

1. Skips to the first `{` in the captured stdout, and
2. Uses `json.Decoder` to read exactly one JSON value — ignoring anything that follows.

The result is that both leading non-JSON content (CSM logs, ANSI-colored warnings) and trailing non-JSON content (TF 1.15 deprecation warnings) are now tolerated when reading dependency outputs.

## Tests

A new internal test `TestExtractFirstJSONObject` in `pkg/config/dependency_internal_test.go` covers:

- pure JSON
- leading AWS CSM log line (regression coverage for #2233)
- leading ANSI-colored warning
- **trailing TF 1.15 deprecation warning** (the regression in this PR)
- leading + trailing combined
- empty `{}`
- no JSON at all
- truncated JSON object

Local validation:

- `go test ./pkg/config/` — pass
- `go vet ./pkg/config/...` — clean
- `golangci-lint run --new-from-merge-base=main ./pkg/config/...` (v2.11.4, the version pinned in `mise.toml`) — `0 issues`
- `gofmt` / `goimports` — clean

I did not add a live integration test against Terraform 1.15 + a real S3 backend, since that requires AWS credentials and a state bucket. The behaviour is exercised at the seam where the bug actually lives — the JSON-from-stdout extraction — using fixture stdout that exactly reproduces the reporter's `'W' after top-level value` symptom. Happy to add one if maintainers prefer.

## Disclosure

Implementation was drafted with the help of Claude Code (Anthropic) and reviewed/validated by me. The code is not adapted from any third-party project.

## TODOs

- [x] Update the docs.
- [x] If this change is not forwards compatible (e.g. a new feature), it is gated behind a feature flag. (N/A — bug fix)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed dependency output parsing when Terraform/OpenTofu emit deprecation warnings. Dependency resolution no longer fails when warnings appear in the output (e.g., Terraform 1.15.0's S3 dynamodb_table deprecation).

### DIFF
--- a/docs/src/data/changelog/v1.0.4/fix-dependency-output-warning-pollution.mdx
+++ b/docs/src/data/changelog/v1.0.4/fix-dependency-output-warning-pollution.mdx
@@ -1,0 +1,12 @@
+---
+version: "v1.0.4"
+category: "bug-fixes"
+---
+
+#### Tolerate non-JSON warnings in `tofu/terraform output -json`
+
+Resolving `dependency` outputs no longer fails when the underlying `tofu/terraform output -json` invocation prints a deprecation warning to stdout alongside the JSON payload. Terraform 1.15.0 introduced a backend deprecation warning for the S3 `dynamodb_table` parameter that is emitted on stdout *after* the JSON object, which broke parsing with errors like `invalid character 'W' after top-level value` and the misleading downstream message `There is no variable named "dependency"`.
+
+Terragrunt now isolates the first JSON object in the captured stdout, so leading log lines (for example, the long-standing AWS Client Side Monitoring `Enabling CSM` line) and trailing warning blocks are both ignored when reading dependency outputs.
+
+Resolves [#6001](https://github.com/gruntwork-io/terragrunt/issues/6001).
--- a/pkg/config/dependency.go
+++ b/pkg/config/dependency.go
@@ -877,16 +877,49 @@ func getOutputJSONWithCaching(ctx context.Context, pctx *ParsingContext, l log.L
 		return nil, err
 	}
 
-	// When AWS Client Side Monitoring (CSM) is enabled the aws-sdk-go displays log as a plaintext "Enabling CSM" to stdout, even if the `output -json` flag is specified. The final output looks like this: "2023/05/04 20:22:43 Enabling CSM{...omitted json string...}", and and prevents proper json parsing. Since there is no way to disable this log, the only way out is to filter.
-	// Related AWS code: https://github.com/aws/aws-sdk-go/blob/81d1cbbc6a2028023aff7bcab0fe1be320cd39f7/aws/session/session.go#L444
-	// Related issues: https://github.com/gruntwork-io/terragrunt/issues/2233 https://github.com/hashicorp/terraform-provider-aws/issues/23620
-	if index := bytes.IndexByte(newJSONBytes, byte('{')); index > 0 {
-		newJSONBytes = newJSONBytes[index:]
+	// `tofu/terraform output -json` stdout can be polluted with non-JSON text on either side of the JSON object:
+	//   - Leading: AWS Client Side Monitoring (CSM) logs (e.g., "2023/05/04 20:22:43 Enabling CSM"),
+	//     ANSI color escape sequences from warning blocks.
+	//     Refs: https://github.com/aws/aws-sdk-go/blob/81d1cbbc6a2028023aff7bcab0fe1be320cd39f7/aws/session/session.go#L444
+	//           https://github.com/gruntwork-io/terragrunt/issues/2233
+	//   - Trailing: Terraform 1.15+ emits backend deprecation warnings (e.g., for the S3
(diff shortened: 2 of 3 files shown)


yhakbar (approved):
Thanks!

PR #6030, opened 2026-05-05
chore: Consolidating `go-getter` routes

### LINKED ISSUE #5453: Support all go-getter content providers as source in stacks

### LINKED ISSUE #6008: bug: `terragrunt stack generate` silently fails to download S3 unit sources (go-getter v2 S3 getter never registered)
## Describe the bug

`terragrunt stack generate` fails to download unit archives from S3 when using `s3::https://...` sources in stack files. The error message is generic and misleading — it appears to be a credentials issue but is actually a missing getter registration.

```
Failed to fetch unit my-service
  Source:      s3::https://s3-us-west-2.amazonaws.com/my-bucket/archives/units/my-unit/aws/v1.0.0/unit.tar.gz
  Destination: /path/to/services/my-service
Original error: Failed to fetch s3::https://... error downloading '...'
```

This happens regardless of AWS credentials — even with valid `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN` exported, and even though `aws s3 cp` works perfectly from the same shell.

## Root Cause

`pkg/config/stack.go` imports and uses **`go-getter/v2`** to fetch remote unit sources:

```go
import "github.com/hashicorp/go-getter/v2"

// in copyFiles():
if _, err := getter.GetAny(ctx, dest, src); err != nil { ... }
```

`getter.GetAny` uses `DefaultClient`, which is initialized with the global `Getters` slice:

```go
// From go-getter/v2/get.go
var Getters []Getter  // initialized in init()

var DefaultClient = &Client{
    Getters:       Getters,
    Decompressors: Decompressors,
}

func init() {
    // Only registers: GitGetter, HgGetter, SmbClientGetter, SmbMountGetter, HttpGetter, FileGetter
    // S3 is NOT registered here
}
```

Unlike **go-getter v1** (where S3 is included in the default getters), **go-getter v2 ships S3 as a separate submodule** (`github.com/hashicorp/go-getter/s3/v2`) that must be explicitly imported and registered by the consuming application. Since terragrunt never imports this submodule, the `s3` forced-protocol getter is never registered, and any `s3::` URL silently falls through with a download error.

## Steps To Reproduce

1. Create a stack file with an S3 unit source:
   ```hcl
   unit "my-service" {
     source = "s3::https://s3-us-west-2.amazonaws.com/my-bucket/archives/units/my-unit/aws/v1.0.0/unit.tar.gz"
     path   = "my-service"
   }
   ```

2. Export valid AWS credentials:
   ```bash
   export AWS_ACCESS_KEY_ID=...
   export AWS_SECRET_ACCESS_KEY=...
   export AWS_SESSION_TOKEN=...
   ```

3. Run `terragrunt stack generate` — it fails with the error above.

4. Verify credentials work directly:
   ```bash
   aws s3 cp s3://my-bucket/archives/units/my-unit/aws/v1.0.0/unit.tar.gz /tmp/test.tar.gz
   # succeeds
   ```

5. Verify go-getter v1 (which IS the default) can download the same URL from Go code — it works because S3 is in v1's default getter list.

## Expected behavior

`terragrunt stack generate` should download `s3::https://...` unit sources successfully when valid AWS credentials are present in the environment, matching the behavior of go-getter v1.

## Proposed Fix

In `pkg/config/stack.go` (or wherever the getter client is constructed for stack generation), import and register the S3 getter from `go-getter/s3/v2`:

```go
import (
    getter "github.com/hashicorp/go-getter/v2"
    s3getter "github.com/hashicorp/go-getter/s3/v2"
)

// Register S3 getter so that s3:: sources work in stack generate
getter.Getters = append(getter.Getters, new(s3getter.Getter))
```

Or construct a custom client with the S3 getter included instead of using `getter.GetAny` / `DefaultClient`.

## Must haves

## Versions

- Terragrunt version: v1.0.2
- go-getter version: v2.2.3 (used for stack unit fetching), v1.8.6 (used elsewhere)
- Operating System: macOS 15 (Darwin)
- Environment: Local development with AWS SSO credentials exported as environment variables

yhakbar:
## Description

Consolidates `go-getter` routes into a single entrypoint so that `go-getter` clones are used consistently.

Closes #6008
Closes #5453

As a consequence of addressing this, two source-URL bugs are fixed:

- `s3::https://...` and `gcs::https://...` stack sources in `terragrunt.stack.hcl` failed
    with misleading credentials errors even when credentials were available.
- Plain `https://www.googleapis.com/storage/...` URLs were silently auto-prefixed with
    `gcs::` by an old normalization path, forcing every public-bucket download through the
    credentialed GCS getter.

Both of these are fixed by consistently routing all go-getter gets through a single package that uniformly handles logic (including fallback logic for backwards compatibility, which will be explained below).

### New Package

With this PR, we're now routing all `go-getter` usage through an `internal/getter` package. This includes special getters like the CAS getter. This is enforced with a special lint added through `depguard`.

### New Strict Control

Unfortunately, one breaking change in how go-getter v1 behaves vs go-getter v2 would have resulted in a breaking change in this PR without the addition of the new `legacy-gcs-public-prefix` strict control.

While this strict control is _not_ active, the `getter` package transparently performs a URL rewrite of any GCS bucket URL to prepend the `gcs::` forced getter prefix. This is how `go-getter` v1 behaved, and Terragrunt users that were successfully downloading GCS sourced modules without the forced getter probably expect downloads to continue to work without the forced getter. 

The problem with always doing this rewrite is that users might not _want_ the rewrite to take place. For public GCS buckets containing modules, users might prefer that the URL is accessed as a normal HTTP source, bypassing GCS authentication, which is how `go-getter` v2 behaves when no `gcs::` prefix is present. As such, a warning is emitted when the rewrite takes place to explicitly call out that users should be prefixing their GCS go-getter URLs with `gcs::` to explicitly opt-in to using GCS credentials for accessing GCS hosted modules.

While the strict control _is_ active, the `getter` package accepts the URL as-is, using v2 semantics. Users have to explicitly use the `gcs::` forced getter prefix to opt-in to using GCS credentials. Otherwise, the bucket will be accessed as a normally accessible HTTP module source.

In a future release of Terragrunt (maybe v2, maybe some major release after v2), this backwards compatibility should be broken.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **New Features**
  * Added Terraform Registry (tfr://) module support, CAS-backed fetching, and a unified download client for more reliable downloads.
  * New local file-copy behavior for sources and improved Git/SSH URL handling.

* **Bug Fixes**
  * Better handling of S3/GCS and legacy public GCS URLs to avoid credential/authorization issues.
  * Improved HTTP(S) auth/netrc behavior for downloads.

* **Refactor**
  * Centralized go-getter integration into an internal getter layer.

* **Documentation**
  * Added docs for legacy GCS public-prefix strict control.


yhakbar:
@coderabbit review

internal/getter/defaults.go
```diff
@@ -0,0 +1,76 @@
...
+
+	httpGetter = newHTTPGetter(b.httpExtraHeader)
+	httpsGetter = newHTTPGetter(b.httpsExtraHeader)
+
+	if b.tfRegistry != nil {
+		out = append(out, b.tfRegistry)
+	}
+
+	if b.casStore != nil {
+		out = append(out, NewCASGetter(b.logger, b.casStore, b.casCloneOpts))
```
denis256: Make sense to register `cas::` schema?

internal/getter/defaults.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	out = append(out, b.prepended...)
+
+	out = append(out,
+		gitGetter,
+		new(getter.HgGetter),
+		new(getter.SmbClientGetter),
+		new(getter.SmbMountGetter),
+		httpGetter,
```
denis256, line 73: Unauthenticated httpGetter is registered before the authenticated httpsGetter, which may catch https requests

internal/getter/casgetter.go
```diff
@@ -0,0 +1,150 @@
...
+// Two normalizations are needed:
+//
+//  1. Strip a leading "git::". The v2 outer client only splits the forced
+//     prefix into req.Forced when the source carried it on entry; when
+//     CASGetter.Detect runs its own detector chain (e.g. for github
+//     shorthand or git@host:path SCP), the v2 GitDetector reattaches
+//     "git::" to its result, and req.URL().String() preserves it. Passing
+//     it through to git makes git look up the missing "git-remote-git"
+//     helper.
+//  2. Convert "ssh://git@host/path" into the SCP-style "git@host:path"
```
denis256, line 84: How will it work if I have a custom SSH port?

yhakbar: Good catch. Addressed.

internal/getter/defaults.go
yhakbar: Good catch. Addressed.

internal/getter/defaults.go
yhakbar: It doesn't hurt. Addressed.

internal/runner/run/download_source.go
```diff
@@ -393,51 +333,107 @@ func downloadSource(
...
+		Src: canonicalSourceURL,
+		Dst: src.DownloadDir,
+		Pwd: opts.WorkingDir,
+	}); err != nil {
+		l.Warnf("CAS download failed: %v. Falling back to standard getter.", err)
+
+		// Clear any partial CAS output before the fallback runs; mixing
+		// leftover CAS files with the standard getter's output leaves the
+		// module dir in an inconsistent state.
+		if removeErr := os.RemoveAll(src.DownloadDir); removeErr != nil {
```
denis256, line 408: Can we use VFS here?

denis256 (dismissed):

internal/runner/run/download_source.go
yhakbar: I'm sure we can. I'll look to addressing that later.

denis256 (approved):

PR #6045, opened 2026-05-06
fix: Removing extra `--auth-provider-cmd` call

yhakbar:
## Description

Removes additional call to the authentication provider command, preventing calls in `.terragrunt-cache` directories in favor of the ones that are called in unit directories beforehand for dependencies.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **Refactor**
  * Centralized credential and IAM handling, replacing per-call credential plumbing with a merged IAM option and provider-based credential resolution for Terraform runs.
* **Tests**
  * Added an integration test covering auth-provider invocation ordering and call-count under concurrent runs.
  * Added test fixtures to exercise auth-provider command behavior and dependency flows.
* **Chores**
  * Updated test ignores to include generated JSONL logs.


pkg/config/dependency.go
```diff
@@ -1051,12 +1051,13 @@ func getTerragruntOutputJSON(ctx context.Context, pctx *ParsingContext, l log.Lo
 	}
 
 	if isInit {
-		credsGetter := creds.NewGetter()
-		if err = credsGetter.ObtainAndUpdateEnvIfNecessary(
+		mergedIAM := iam.MergeRoleOptions(remoteStateTGConfig.GetIAMRoleOptions(), pctx.OriginalIAMRoleOptions)
```
denis256, line 1054: How will we know that this issue will not appear again?
Most probably, we need some kind of tests

denis256 (dismissed):

pkg/config/dependency.go
yhakbar: Working on it.

yhakbar: Added in `8ba47f2035146a5775b503bb1229b8166b41fa5d`.

denis256 (approved):

Issue #5918, opened 2026-04-16
extra_arguments env_vars TF_CLI_CONFIG_FILE not respected by Provider Cache Server


elkh510:
## Description

`extra_arguments.env_vars` with `TF_CLI_CONFIG_FILE` (the solution from #2761) does not work when Provider Cache Server is enabled. The cache server initializes before `extra_arguments` are parsed, so the custom `.terraformrc` with `host` blocks is never loaded.

This means users with private provider registries (configured via `host` blocks) cannot use both `extra_arguments` approach and provider caching together.

## Steps to Reproduce

`root.hcl`:
```hcl
terraform {
  extra_arguments "cli_config" {
    commands = get_terraform_commands_that_need_vars()
    env_vars = {
      TF_CLI_CONFIG_FILE = "${get_repo_root()}/terraformrc.hcl"
    }
  }
}
```

`terraformrc.hcl`:
```hcl
host "registry.example.com" {
  services = {
    "providers.v1" = "https://registry.example.com/repository/terraform-hosted/v1/providers/"
  }
}
```

```bash
# Works — extra_arguments sets TF_CLI_CONFIG_FILE for terraform
terragrunt plan --provider-cache=false

# Fails — cache server doesn't see the host block
terragrunt plan
# Error: "does not offer a Terraform provider registry"
```

## Root Cause

Execution order in `commands.go`:

1. **Line 282**: `InitServer()` → `LoadUserConfig()` → reads `~/.terraformrc` or `$TF_CLI_CONFIG_FILE` from `os.Environ()`
2. **Later**: terragrunt parses `root.hcl` → resolves `extra_arguments` → sets `env_vars`
3. **Later**: runs `terraform init` with the env vars

The cache server at step 1 doesn't know about `extra_arguments` from step 2. `LoadUserConfig` calls HashiCorp's `cliconfig.LoadConfig()` which reads `os.Getenv("TF_CLI_CONFIG_FILE")` — but this env var hasn't been set yet.

### Option A: CLI flag `--tf-cli-config-file` (~30 lines)

Add a new flag (similar to `--provider-cache-dir`) that sets `os.Setenv("TF_CLI_CONFIG_FILE", ...)` before `InitServer`:

```go
// In commands.go, before InitServer
if opts.TFCLIConfigFile != "" {
    os.Setenv("TF_CLI_CONFIG_FILE", opts.TFCLIConfigFile)
}
```

[...]

Issue #3572, opened 2024-11-16
Set Terragrunt defaults in terragrunt.hcl file


mossad-zika:
This is a duplicate of https://github.com/gruntwork-io/terragrunt/issues/2849

Currently, Terragrunt is able to pass extra arguments and set environment variables for Terraform using hooks or run_cmd() command. However, this does not work for Terragrunt itself.

For example, we constantly have to type terragrunt run-all plan --terragrunt-non-interactive, whereas I would want to set this flag in terragrunt.hcl file (or as an env var TERRAGRUNT_NON_INTERACTIVE=true) and have Terragrunt honour it.

Describe alternatives you've considered

Wrapper for terragrunt executables, such as Makefiles and bash scripts - but those are messy.
Setting TERRAGRUNT_ variables manually on the CLI before running Terragrunt but, again, yet another step and more typing that shouldn't be needed.
Additional context
If there's any way to set Terragrunt defaults within itself, that would be great as the number of options is growing and can make for a messy and long command line.

yhakbar:
Hey @mossad-zika , we've talked about this internally, and we have some designs in mind.

The design that we're considering right now is basically to have a `.terragruntrc.json` file that can go in places you would typically expect it:

1. Current working directory.
2. Root of repository.
3. In `.config` directory at root of repository.
4. Home directory of user.

The file would be parsed, then set Terragrunt flags with names that match the names of flags, so it would be pretty easy to guess how configurations would be set.

In the future there may be other configurations that we'd add to this, but that's our general thought process right now.

Marking the issue as `preserved`, as we plan to address this, but it's just not a top priority right now.

If you would like this issue resolved earlier, consider cutting a pull request to deliver this.

BeyondEvil:
Yes! This!

And please please please use JSON and not yaml 😅

yhakbar:
Something to consider is that this is becoming more important, as we're evaluating changing some defaults in how Terragrunt behaves.

[This flag](https://terragrunt.gruntwork.io/docs/reference/cli-options/#terragrunt-no-destroy-dependencies-check), for example, is not a great default for Terragrunt to have, but is also non-trivial to remove. It's nice for users to have that check happen whenever they try to destroy something, and it's only useful when users _don't know_ that there's a dependency on something they're destroying, so it's not great to opt-in to that functionality. We could have some recommended `.terragruntrc.json` files in the docs for that feature that explicitly enable the check for users that are willing to accept the performance penalty.

[...]

PR #6049, opened 2026-05-07
feat: support .terragruntrc.json for early environment variable injection

elkh510:
## Summary

Implements experimental `.terragruntrc.json` support, addressing #3572 and the underlying problem reported in #5918.

The Provider Cache Server (`providercache.InitServer`) is initialized in `initialSetup()` before `extra_arguments env_vars` are applied (which happens later in `runner/run/run.go`). This means env vars like `TF_CLI_CONFIG_FILE` set via `extra_arguments` are not visible to the cache server on startup.

**Solution:** Read a `.terragruntrc.json` file early in `initialSetup()` — before any initialization that reads `os.Environ` — and apply `env` values that are not already set in the environment.

## File format

```json
{
  "env": {
    "TF_CLI_CONFIG_FILE": "./terraformrc.hcl",
    "TG_PROVIDER_CACHE": "1"
  }
}
```

This is Phase 1 (env section only). The `flags` and `commands` sections from the #3572 design are planned for a follow-up PR.

## How it works

Place `.terragruntrc.json` anywhere in the project tree (typically next to the root `terragrunt.hcl`). Terragrunt searches upward from the working directory, then falls back to `~/.config/terragrunt/.terragruntrc.json` and `~/.terragruntrc.json`.

Value processing:
1. `os.ExpandEnv` expands `$VAR` / `${VAR}` references
2. Values starting with `./` or `../` are resolved to an absolute path relative to the `.terragruntrc.json` file's directory — useful when the repo is cloned to different paths on different machines
3. URL values (containing `://`) are left unchanged

**Priority:** shell environment > `.terragruntrc.json` > built-in defaults. The rc file does not override variables already present in the environment.

## Changes

- `internal/util/file.go` — `TerragruntRC` struct, `FindAndLoadRCFile()`, `loadRCFile()`
- `internal/util/file_test.go` — `TestLoadRCFile` and `TestFindAndLoadRCFile`
- `internal/cli/commands/commands.go` — call `FindAndLoadRCFile` in `initialSetup()` before `providercache.InitServer`


elkh510:
Regarding the `t.Setenv("HOME", home)` suggestion: the parent test `TestLoadEnvFile` calls `t.Parallel()`, and Go's testing package panics if a subtest calls `t.Setenv` when any parent in the chain is parallel (`testing: test using t.Setenv can not use t.Parallel`). Removing `t.Parallel()` from the parent would serialize all subtests, which is worse than the theoretical Windows edge case.

In practice, terragrunt targets Linux/macOS and the CI pipeline is Linux-only, so `$HOME` will always be set. The `os.UserHomeDir()` call is already there to get the correct expected value - it just can't be injected back into the env without breaking the parallel test structure.

Also added a `${VAR}` brace-style expansion subtest as suggested.

yhakbar:
Please take a look at the comment I've left on the associated issue:
https://github.com/gruntwork-io/terragrunt/issues/5918

Marking PR as in-draft pending resolution of that comment.

elkh510:
Thanks for the feedback, @yhakbar! The idea in #3572 is great, and I've updated this PR accordingly. The `.terragrunt-env` approach has been dropped in favor of `.terragruntrc.json`.

**What's implemented so far (Phase 1: `env` section only):**

Place a `.terragruntrc.json` anywhere in the project tree:

```json
{
  "env": {
    "TF_CLI_CONFIG_FILE": "./terraformrc.hcl",
    "TG_PROVIDER_CACHE": "1"
  }
}
```

How it works:
- Terragrunt searches upward from the working directory for `.terragruntrc.json`, then falls back to `~/.config/terragrunt/.terragruntrc.json` and `~/.terragruntrc.json`
- Values are expanded via `os.ExpandEnv` (`$VAR` / `${VAR}`)
- Values starting with `./` or `../` are resolved to absolute paths relative to the rc file's directory - this lets teams share a single config across machines where the repo is cloned to different paths (e.g. `"TF_CLI_CONFIG_FILE": "./terraformrc.hcl"` always points to the file next to `.terragruntrc.json`)
- URL values (containing `://`) are left unchanged
- The rc file does **not** override variables already present in the shell environment (shell env takes precedence)
- Loaded early in `initialSetup()`, before the provider cache server is initialized

The `flags` and `commands` sections from the #3572 design are intentionally left for a follow-up PR to keep this change focused. Happy to adjust based on your feedback!

elkh510:
@yhakbar friendly ping on this one 🙂
Since the last review I've reworked the implementation following the discussion on https://github.com/gruntwork-io/terragrunt/issues/5918 - dropped .terragrunt-env in favor of .terragruntrc.json matching the https://github.com/gruntwork-io/terragrunt/issues/3572 design (scoped to the env section for Phase 1, with flags/commands left for a follow-up). I kept it in draft pending your sign-off on the direction - should I move it out of draft, or is there anything else you'd like adjusted first?

jubr:
Also very interested in this functionality! Here would be my immediate use case:

```yaml
env:
  TG_PROVIDER_CACHE: 1
  TG_PROVIDER_CACHE_DIR: /path/to/runner/.cache/tg-providers
  TG_TF_PATH: /opt/mise/installs/opentofu/1.12/tofu
  TERM: dumb # prevent spewing `^[]11;rgb:0000/0000/0000^[\` all over my broken screen terminal
```

PR #6059, opened 2026-05-08
fix(auto-init): Re-init on source change when modules are cached

### LINKED ISSUE #6058: Auto-init ignores `.terragrunt-init-required` marker when `.terraform/modules/` already exists
## Describe the bug

If you keep the Terragrunt cache between runs (containerized runners with persistent volumes, CI workers, your laptop) and your unit has any `module ""` block, a source change that the cached `.terraform/` can't satisfy gets silently skipped. The most common case is bumping `required_providers` in the source module. The next `terragrunt plan`/`apply` then fails with `Error: Required plugins are not installed`.

The `.terragrunt-init-required` marker file (introduced in #1921, refined in #1924) is supposed to force a re-init on the next non-init command after a source change. The file is still being written correctly by [`download_source.go`](https://github.com/gruntwork-io/terragrunt/blob/main/internal/runner/run/download_source.go#L237-L247) when the encoded source version changes. The problem is its only consumer, [`modulesNeedInit()`](https://github.com/gruntwork-io/terragrunt/blob/main/internal/runner/run/run.go#L480-L498), short-circuits on `.terraform/modules/` existence before consulting it:

```go
func modulesNeedInit(opts *Options) (bool, error) {
    modulesPath := filepath.Join(opts.DataDir(), "modules")
    if util.FileExists(modulesPath) {
        return false, nil          // ← short-circuit, marker never consulted
    }
    moduleNeedInit := filepath.Join(opts.WorkingDir, ModuleInitRequiredFile)
    if util.FileExists(moduleNeedInit) {
        return true, nil
    }
    // ...
}
```

Because essentially every real unit declares at least one `module ""` block, `.terraform/modules/` exists after the first init, so this function always returns `false` before checking the marker. [`providersNeedInit()`](https://github.com/gruntwork-io/terragrunt/blob/main/internal/runner/run/run.go#L441-L447) only checks for *file presence*, not lock-vs-constraint consistency, so it also returns `false`. `needsInitRunCfg()` then returns `false` overall, auto-init is skipped, and the underlying `tofu plan`/`apply` fails. That's the exact symptom the marker mechanism was introduced to prevent.

## Steps To Reproduce

Minimal fixture (matches the new regression test in the proposed fix):

```hcl
# terragrunt.hcl
terraform {
  source = ".//src"
}
```

```hcl
# src/main.tf
terraform {
  required_providers {
    null = {
      source  = "registry.opentofu.org/hashicorp/null"
      version = "~> 3.2"
    }
  }
}

provider "null" {}

module "leaf" {
  source = "./leaf"
}
```

```hcl
# src/leaf/main.tf
resource "null_resource" "noop" {}
```

Then:

1. Run `terragrunt plan`. It succeeds and populates `.terraform/{providers,modules}/` plus the lock file inside the unit's cache directory.
2. Modify the source so its encoded version changes. Any change to a file under `src/` will do, e.g. `touch src/main.tf` to bump mtime, which the local-source path of `EncodeSourceVersion` folds into its hash. In a real deployment this would be a Git ref bump or a `required_providers` change in the source module.
3. Run `terragrunt plan` again **without** clearing the Terragrunt cache.

Observed: the second `plan` does **not** re-run `init`. If the source change invalidated cached providers, the underlying tofu command then fails with `Error: Required plugins are not installed`.

Expected: the second `plan` re-runs `init` because `.terragrunt-init-required` was written when the source version changed.

## Expected behavior

The `.terragrunt-init-required` marker is a global "source changed, re-init everything" signal. It should be honored regardless of whether `.terraform/modules/` is already populated. Behavior should match what the [auto-init docs](https://terragrunt.gruntwork.io/docs/features/auto-init/) already promise:

> The `.terragrunt-init-required` file is in the downloaded module directory…

## Must haves

## Nice to haves

- [x] Terminal output (above).
- [ ] Screenshots: N/A.

## Versions

- Terragrunt version: v1.0.4 (reproduces on `main` at commit `74cfdb166774518caa461552362898a6ad217a2c`)
- OpenTofu/Terraform version: OpenTofu v1.11.4
- Environment details: macOS (darwin_arm64); also reproduces in containerized Linux runners with persistent `TG_DOWNLOAD_DIR`.

## Additional context

- The bug only reproduces when all three hold: persistent Terragrunt cache across runs, unit declares at least one `module ""` block, and the source revision changes between runs in a way the cached `.terraform/` cannot satisfy (e.g. a `required_providers` bump, a new provider added to the source, or a new module dependency).
- The existing regression test [`TestAutoInitWhenSourceIsChanged`](https://github.com/gruntwork-io/terragrunt/blob/main/test/integration_test.go#L3548-L3583) doesn't catch this because its fixture [`test/fixtures/dirs/main.tf`](https://github.com/gruntwork-io/terragrunt/blob/main/test/fixtures/dirs/main.tf) declares no modules. `.terraform/modules/` is never created, so the buggy short-circuit never fires.
- Proposed fix: hoist the marker check out of `modulesNeedInit()` and into `needsInitRunCfg()` so it's honored regardless of cached `.terraform/modules/` contents. PR to follow.

### Related

- Original marker mechanism: #1921
- Version-check refinement: #1924
- Marker writer (already correct): [`internal/runner/run/download_source.go#L237-L247`](https://github.com/gruntwork-io/terragrunt/blob/main/internal/runner/run/download_source.go#L237-L247)
- Buggy short-circuit: [`internal/runner/run/run.go#L480-L498`](https://github.com/gruntwork-io/terragrunt/blob/main/internal/runner/run/run.go#L480-L498)
- Existing test that misses the bug: [`test/integration_test.go#L3548-L3583`](https://github.com/gruntwork-io/terragrunt/blob/main/test/integration_test.go#L3548-L3583)

arnaud-dezandee:
## Description

Fixes #6058.

`needsInitRunCfg()` was deciding whether auto-init must run by calling `modulesNeedInit()`, which short-circuits on `.terraform/modules/` existence before consulting the `.terragrunt-init-required` marker. Once the cache is populated (i.e. after the first init in any unit with a `module ""` block), the marker written by `download_source.go` on a source-version change was silently ignored, and the next `terragrunt plan`/`apply` would fail with `Error: Required plugins are not installed` whenever the source change invalidated cached providers.

This PR hoists the marker check out of `modulesNeedInit()` and into `needsInitRunCfg()` so it is consulted regardless of cached `.terraform/modules/` contents. Adds a regression test using a local-source fixture (mtime bump on `src/main.tf` triggers `EncodeSourceVersion` to write the marker on the next download). Verified the test fails on `main` with the predicted symptom and passes with the fix.

## TODOs

- [ ] Update the docs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Auto-init marker is now honored even when Terraform modules appear cached: initialization will rerun when the marker exists, preventing missing-plugin failures during plan.

* **Tests**
  * Added a regression test and fixtures to verify init runs when the marker is present and that repeated plans re-run init as expected.

* **Documentation**
  * Changelog entry describing the auto-init behavior fix.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gruntwork-io/terragrunt/pull/6059)

### DIFF
--- a/docs/src/data/changelog/v1.0.5/auto-init-marker-cached-modules.mdx
+++ b/docs/src/data/changelog/v1.0.5/auto-init-marker-cached-modules.mdx
@@ -1,0 +1,10 @@
+---
+version: "v1.0.5"
+category: "bug-fixes"
+---
+
+#### Auto-init now re-runs after a source change when modules are already cached
+
+`terragrunt plan`/`apply` could fail with `Error: Required plugins are not installed` after a source-version change in any unit with a `module ""` block. The `.terragrunt-init-required` marker written on source change was being ignored because `modulesNeedInit` short-circuited as soon as `.terraform/modules/` existed.
+
+The marker check now lives at the top of `needsInitRunCfg` and is honored regardless of cached `.terraform/modules/` contents.
--- a/internal/runner/run/run.go
+++ b/internal/runner/run/run.go
@@ -483,11 +483,6 @@ func modulesNeedInit(opts *Options) (bool, error) {
 		return false, nil
 	}
 
-	moduleNeedInit := filepath.Join(opts.WorkingDir, ModuleInitRequiredFile)
-	if util.FileExists(moduleNeedInit) {
-		return true, nil
-	}
-
 	// Check for module definitions in .tf and .tofu files using WalkDir
 	hasModuleDefinition, err := util.RegexFoundInTFFiles(opts.WorkingDir, ModuleRegex)
 	if err != nil {
@@ -659,6 +654,12 @@ func PrepareNonInitCommand(
 func needsInitRunCfg(ctx context.Context, l log.Logger, opts *Options, cfg *runcfg.RunConfig) (bool, error) {
 	if slices.Contains(TerraformCommandsThatDoNotNeedInit, opts.TerraformCliArgs.First()) {
 		return false, nil
+	}
+
+	// Marker check lives here, not in modulesNeedInit, so a populated .terraform/modules/
+	// can't short-circuit past it. Refs: #1921, #6058.
+	if util.FileExists(filepath.Join(opts.WorkingDir, ModuleInitRequiredFile)) {
+		return true, nil
 	}
 
 	if providersNeedInit(opts) {
--- a/test/fixtures/init-once/terragrunt.hcl
+++ b/test/fixtures/init-once/terragrunt.hcl
@@ -1,3 +1,3 @@
 terraform {
-  source = ".//module"
+  source = "./module"
 }
--- a/test/fixtures/regressions/auto-init-marker-with-cached-modules/src/leaf/main.tf
+++ b/test/fixtures/regressions/auto-init-marker-with-cached-modules/src/leaf/main.tf
@@ -1,0 +1,1 @@
+resource "null_resource" "noop" {}
--- a/test/fixtures/regressions/auto-init-marker-with-cached-modules/src/main.tf
+++ b/test/fixtures/regressions/auto-init-marker-with-cached-modules/src/main.tf
@@ -1,0 +1,14 @@
+terraform {
+  required_providers {
+    null = {
+      source  = "registry.opentofu.org/hashicorp/null"
+      version = "~> 3.2"
+    }
+  }
+}
+
+provider "null" {}
+
+module "leaf" {
+  source = "./leaf"
+}
--- a/test/fixtures/regressions/auto-init-marker-with-cached-modules/terragrunt.hcl
+++ b/test/fixtures/regressions/auto-init-marker-with-cached-modules/terragrunt.hcl
@@ -1,0 +1,3 @@
+terraform {
+  source = ".//src"
+}
--- a/test/integration_regressions_test.go
+++ b/test/integration_regressions_test.go
@@ -39,6 +39,7 @@ const (
 	testFixtureChainedDepsExposedInclude         = "fixtures/regressions/chained-deps-exposed-include"
 	testFixtureExposedIncludePartialParseError   = "fixtures/regressions/exposed-include-partial-parse-error"
 	testFixtureDAGQueueDisplay                   = "fixtures/regressions/dag-queue-display"
+	testFixtureAutoInitMarkerCachedModules       = "fixtures/regressions/auto-init-marker-with-cached-modules"
 )
 
 func TestNoAutoInit(t *testing.T) {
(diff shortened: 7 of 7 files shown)


yhakbar (commented):
This seems solid. Will approve if CI passes. Mind adding a changelog entry to document the fix?

yhakbar:
Also, I am seeing test failures in CI for the test `TestTerragruntInitOnce`:

```
2026-05-08T16:08:27.3322540Z         	Error Trace:	/Users/runner/work/terragrunt/terragrunt/test/integration_test.go:1043
2026-05-08T16:08:27.3327530Z         	Error:      	"\n\x1b[0m\x1b[1mInitializing the backend...\x1b[0m\n\x1b[0m\x1b[1mInitializing modules...\x1b[0m\n\n\x1b[0m\x1b[1mInitializing provider plugins...\x1b[0m\n- Reusing previous version of hashicorp/random from the dependency lock file\n- Using previously-installed hashicorp/random v3.8.0\n\n\x1b[0m\x1b[1m\x1b[32mOpenTofu has been successfully initialized!\x1b[0m\x1b[32m\x1b[0m\n\x1b[0m\x1b[32m\nYou may now begin working with OpenTofu. Try running \"tofu plan\" to see\nany changes that are required for your infrastructure. All OpenTofu commands\nshould now work.\n\nIf you ever set or change modules or backend configuration for OpenTofu,\nrerun this command to reinitialize your working directory. If you forget, other\ncommands will detect it and remind you to do so if necessary.\x1b[0m\n\nOpenTofu used the selected providers to generate the following execution\nplan. Resource actions are indicated with the following symbols:\n  \x1b[32m+\x1b[0m create\x1b[0m\n\nOpenTofu will perform the following actions:\n\n\x1b[1m  # module.vpc.random_integer.example[0]\x1b[0m will be created\n\x1b[0m  \x1b[32m+\x1b[0m\x1b[0m resource \"random_integer\" \"example\" {\n      \x1b[32m+\x1b[0m\x1b[0m id      = (known after apply)\n      \x1b[32m+\x1b[0m\x1b[0m keepers = {\n          \x1b[32m+\x1b[0m\x1b[0m \"example\" = \"hello world\"\n        }\n      \x1b[32m+\x1b[0m\x1b[0m max     = 50000\n      \x1b[32m+\x1b[0m\x1b[0m min     = 1\n      \x1b[32m+\x1b[0m\x1b[0m result  = (known after apply)\n    }\n\n\x1b[1mPlan:\x1b[0m 1 to add, 0 to change, 0 to destroy.\n\x1b[0m\x1b[90m\n─────────────────────────────────────────────────────────────────────────────\x1b[0m\n\nNote: You didn't use the -out option to save this plan, so OpenTofu can't\nguarantee to take exactly these actions if you run \"tofu apply\" now.\n" should not contain "Initializing modules"
2026-05-08T16:08:27.3327620Z         	Test:       	TestTerragruntInitOnce
2026-05-08T16:08:27.3327780Z         	Messages:   	init command executed more than once
2026-05-08T16:08:27.3327850Z --- FAIL: TestTerragruntInitOnce (4.25s)
```

arnaud-dezandee:
About this CI failures in `TestTerragruntInitOnce`, the fixture use `source = ".//module"` and the `//` makes `EncodeSourceVersion` walk the parent dir, including `terragrunt.hcl`. 

The test rewrites that file (same bytes, but mtime bumps), and since #2168 switched the hash from content to mtime (content-hashing was hanging at 100% CPU for 5+ min on large trees), the source version bumps and the marker gets written. Pre-PR the marker was silently ignored, so the test passed by accident.

I pushed another commit to drop the `//` in the fixture so the walk stays inside `./module/`. That's the minimal amount of changes required to fix the test as is.

Real users using the `.//module` pattern (the documented go-getter convention for "source root + sub-module path") still hit spurious reinits when their terragrunt.hcl mtime changes between runs. So this is technically a fixture fix that hides a latent issue in the production code.

arnaud-dezandee:
CI integration tests failed because of Github uptime

yhakbar (approved):

PR #6072, opened 2026-05-11
chore: stacks dependencies variables

denis256:
## Description

* Lazy hcl.Expression parsing for source/path/values (and include path) so non-literal expressions in unrelated unit/stack blocks no longer break autoinclude resolution

RFC: https://github.com/gruntwork-io/terragrunt/issues/5663

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **New Features**
  * Lazy two-pass stack parsing improves discovery of generated nested stack units and makes `terragrunt run --all` traversal robust.

* **Bug Fixes**
  * Autoinclude discovery no longer blocked by unrelated non-literal expressions or function calls in unit source.
  * `values = {...}` inside autoinclude blocks is rejected with a clear error.
  * Path/source evaluation errors now surface with clearer context.

* **Documentation**
  * Added v1.0.5 changelog and updated stack-dependencies roadmap.

* **Tests**
  * Expanded unit and integration regressions for autoinclude, nested discovery, run-all, and eval-error scenarios; removed obsolete loud-fail tests.


docs/src/data/experiments/stack-dependencies.mdx
yhakbar: I think we can clean up this block now.

internal/hclparse/autoinclude.go
```diff
@@ -130,13 +132,95 @@ func (a *AutoIncludeHCL) Resolve(evalCtx *hcl.EvalContext) (*AutoIncludeResolved
 		return nil, diags
 	}
 
+	// Evaluate the optional `values` attribute against an augmented eval context that binds dependency mock_outputs so expressions like `dependency.X.outputs.Y` resolve. Mock_outputs are the only dependency data available at parse time; at runtime the production parser will resolve dependencies for real if the generated autoinclude file is re-parsed there.
+	resolvedValues, valDiags := resolveAutoIncludeValues(body, evalCtx, deps)
+	diags = append(diags, valDiags...)
+
+	if diags.HasErrors() {
+		return nil, diags
```
yhakbar: We can return a best effort value here, right?

internal/hclparse/autoinclude.go
```diff
@@ -130,13 +132,95 @@ func (a *AutoIncludeHCL) Resolve(evalCtx *hcl.EvalContext) (*AutoIncludeResolved
...
 	return &AutoIncludeResolved{
 		EvalCtx:      evalCtx,
 		Dependencies: deps,
 		RawBody:      a.Remain,
+		Values:       resolvedValues,
 	}, nil
 }
 
+// resolveAutoIncludeValues extracts and evaluates the optional `values = {...}` attribute from the autoinclude body. The eval context is augmented with `dependency.<name>.outputs = mock_outputs` for each declared dependency so values that reference dependency outputs resolve to the mock value at parse time. Returns (nil, nil) when no values attribute is present.
+func resolveAutoIncludeValues(body *hclsyntax.Body, evalCtx *hcl.EvalContext, deps []AutoIncludeDependency) (*cty.Value, hcl.Diagnostics) {
```
yhakbar: Why do we do this?

internal/hclparse/autoinclude.go
```diff
@@ -130,13 +132,95 @@ func (a *AutoIncludeHCL) Resolve(evalCtx *hcl.EvalContext) (*AutoIncludeResolved
...
+
+		val, diags := attr.Expr.Value(evalCtx)
+		if diags.HasErrors() {
+			return cty.EmptyObjectVal, diags
+		}
+
+		return val, nil
+	}
+
+	return cty.EmptyObjectVal, nil
```
yhakbar: Is EmptyObjectVal the right thing here, or is it DynamicVal? What's the difference?

internal/hclparse/autoinclude.go
```diff
@@ -95,6 +95,15 @@ func (a *AutoIncludeHCL) Resolve(evalCtx *hcl.EvalContext) (*AutoIncludeResolved
 		return &AutoIncludeResolved{EvalCtx: evalCtx, RawBody: a.Remain}, nil
 	}
 
+	if valuesAttr, hasValues := body.Attributes["values"]; hasValues {
+		return nil, hcl.Diagnostics{{
+			Severity: hcl.DiagError,
+			Summary:  "`values` is not allowed inside `autoinclude`",
+			Detail:   "did you mean to declare `values = {...}` on the parent unit/stack block (next to `source`/`path`) instead of inside the `autoinclude` block?",
```
yhakbar: Great diagnostic. Just to confirm, this shouldn't be proper cased (i.e. `Did you mean ...`)? I don't remember the best practices for diagnostics.

internal/hclparse/errors.go
```diff
@@ -159,3 +159,19 @@ type EmptyArgError struct {
 func (e EmptyArgError) Error() string {
 	return fmt.Sprintf("hclparse.%s: %s is empty", e.Func, e.Arg)
 }
+
+// RefEvalError indicates that a unit's or stack's lazy attribute expression (path or source) could not be evaluated against the bootstrap eval context.
```
yhakbar: I don't know what bootstrap eval context means here. Presumably that's different than the bootstrap command?

internal/hclparse/parse.go
```diff
@@ -33,35 +36,91 @@ const (
 
 // ParseStackFileInput holds the input for ParseStackFile.
 type ParseStackFileInput struct {
-	Values   *cty.Value
-	Filename string
-	StackDir string
-	Src      []byte
+	Values *cty.Value
+	// UnitRefs lets the caller supply already-resolved unit ComponentRefs (with paths, names, and any ChildRefs). When non-nil, the parser uses these for the autoinclude eval context instead of evaluating each unit's Path attribute itself, so non-literal source/path expressions in unrelated unit blocks do not block parsing.
```
yhakbar: NIT: Either all fields should have godoc comments, or none should.

internal/hclparse/parse.go
```diff
@@ -33,35 +36,91 @@ const (
...
+	UnitRefs []ComponentRef
+	// StackRefs is the analogous slice for stack blocks. Each ref's ChildRefs must already be populated when callers want stack.<name>.<unit>.path references to resolve.
```
yhakbar: NIT: Comments like this aren't great. We shouldn't assume position-ality in comments. If you hover over this field somewhere else, you won't know what `StackRefs` is analogous to.

internal/hclparse/parse.go
```diff
@@ -33,35 +36,91 @@ const (
...
-// Between passes: Build eval context with unit.<name>.path and stack.<name>.path
-// variables. Paths are resolved to absolute paths under .terragrunt-stack/.
+// Between passes: Build eval context with unit.<name>.path and stack.<name>.path variables. When the caller supplies pre-resolved UnitRefs/StackRefs, those drive the refs; otherwise paths are evaluated against a stdlib-only context and refs are refreshed after include merging.
 //
-// Pass 2: For each unit/stack with an autoinclude block, resolve the autoinclude
-// body using the eval context. dependency.config_path is evaluated (references
-// unit.*.path), while inputs are left unevaluated (contain dependency.*.outputs.*).
+// Pass 2: For each unit/stack with an autoinclude block, resolve the autoinclude body using the eval context. dependency.config_path is evaluated (references unit.*.path / stack.*.<unit>.path), while inputs are left unevaluated (contain dependency.*.outputs.* which is runtime-only).
 func ParseStackFile(fs vfs.FS, input *ParseStackFileInput) (*ParseResult, error) {
+	validateParseStackFileInput(fs, input)
```
yhakbar, line 61: We probably want to add some sort of recover and re-throw in `main` that calls out whatever the user ran into is a bug in Terragrunt and should be reported with the stack trace. I don't think we have that right now.

internal/hclparse/parse.go
```diff
@@ -33,35 +36,91 @@ const (
...
+	// Strict locals pass after includes so locals referencing unit.<X>.path see units contributed by included files; mergeOneInclude rejects locals in included files so root locals are the only locals here.
+	if stackFile.Locals != nil {
+		if err := evaluateLocals(stackFile.Locals.Remain, evalCtx); err != nil {
+			return nil, err
+		}
+	}
+
+	autoIncludes, err := resolveAutoIncludes(stackFile, evalCtx, srcByAutoInclude)
+	if err != nil {
+		return nil, err
```
yhakbar: For each of these err cases, we could have returned best effort results. Helpful for LSP, etc.

internal/hclparse/parse.go
```diff
@@ -78,72 +137,122 @@ func ParseStackFile(fs vfs.FS, input *ParseStackFileInput) (*ParseResult, error)
...
-// Uses Variables() to pre-check whether each local's dependencies are satisfied
-// before attempting evaluation. Shrinks the work set each pass. Returns an error
-// if any locals cannot be evaluated (cycle or invalid reference).
+// evaluateLocals strictly evaluates root locals against evalCtx. Returns an error on eval failure, dependency cycle, or iteration overflow. Used after include merging so locals referencing unit.<X>.path resolve against the fully populated context.
 func evaluateLocals(body hcl.Body, evalCtx *hcl.EvalContext) error {
+	return evaluateLocalsRun(body, evalCtx, true)
+}
+
+// evaluateLocalsBestEffort evaluates root locals that are ready against evalCtx and silently leaves the rest for the later strict pass. Used before include merging so include paths referencing local.X resolve while locals referencing not-yet-merged units are deferred.
+func evaluateLocalsBestEffort(body hcl.Body, evalCtx *hcl.EvalContext) {
```
yhakbar: This looks like a footgun. I'm sure it's necessary, given how things are currently constructed, but I wonder if there's a better way to handle this.

internal/hclparse/stack.go
```diff
@@ -249,40 +266,81 @@ func discoverStackChildUnitsWithDepth(fs vfs.FS, stackSourceDir, stackGenDir str
...
+
+	return ComponentRef{
+		Name:      s.Name,
+		Path:      nestedGenPath,
+		ChildRefs: discoverStackChildUnitsWithDepth(fs, nestedSourceDir, nestedGenPath, depth+1),
+	}, true
+}
+
+// stdlibEvalContext builds a minimal eval context wired with the terraform stdlib (matching the production parser's tflang.Scope setup in pkg/config/config_helpers.go) plus get_terragrunt_dir bound to baseDir, so the most common Terragrunt path function resolves during discovery on generated stack files where no pctx is available.
+func stdlibEvalContext(baseDir string) *hcl.EvalContext {
```
yhakbar, line 390: This is really weird. We're partially supporting the set of HCL configurations here, and only if you use `get_terragrunt_dir()`? Seems off.

internal/hclparse/parse.go
```diff
@@ -424,21 +571,30 @@ func buildRefsWithAbsPath(stackTargetDir string, units []*UnitBlockHCL) []Compon
...
 	}
 
-	return refs
+	return refs, nil
 }
 
-// buildStackRefsWithAbsPath builds ComponentRef values for stack blocks and discovers their child units.
-func buildStackRefsWithAbsPath(fs vfs.FS, stackDir string, stackTargetDir string, stacks []*StackBlockHCL, depth int) []ComponentRef {
+// buildStackRefsWithAbsPath builds ComponentRef values for stack blocks and discovers their child units. Returns an error when any stack's Path or Source expression cannot be evaluated against evalCtx so callers on the bootstrap path see the root cause rather than a misleading downstream "Unknown variable: stack" diagnostic.
+func buildStackRefsWithAbsPath(fs vfs.FS, stackDir, stackTargetDir string, stacks []*StackBlockHCL, depth int, evalCtx *hcl.EvalContext) ([]ComponentRef, error) {
```
yhakbar: NIT: Let's break up big lines like this.

internal/hclparse/parse.go
```diff
@@ -78,82 +137,167 @@ func ParseStackFile(fs vfs.FS, input *ParseStackFileInput) (*ParseResult, error)
...
-		AutoIncludes: autoIncludes,
-	}, nil
+	evalCtx.Variables[varUnit] = BuildComponentRefMap(unitRefs)
+	evalCtx.Variables[varStack] = BuildComponentRefMap(stackRefs)
+
+	return nil
+}
+
+// EvalString evaluates expr against ctx and returns the resulting string. Returns ("", nil) when expr is nil (attribute absent). When ctx is nil the expression must be a constant; if expr references variables, returns an error. attrName is used in error messages.
+func EvalString(expr hcl.Expression, ctx *hcl.EvalContext, attrName string) (string, error) {
```
yhakbar: NIT: There's an odd mix of high-level and low-level functionality in this file. We might want to move this to a different file/package so that generic stuff for evaluating HCL expressions aren't collocated with business logic for parsing unit/stack files.

internal/hclparse/parse.go
```diff
@@ -312,21 +470,33 @@ func processStackIncludes(fs vfs.FS, stackFile *StackFileHCL, stackDir string, s
...
 }
 
-// mergeOneInclude reads and merges a single included stack file.
-func mergeOneInclude(fs vfs.FS, stackFile *StackFileHCL, inc *StackIncludeHCL, stackDir string, srcByAutoInclude map[*AutoIncludeHCL][]byte) error {
-	includePath := inc.Path
+// mergeOneInclude reads and merges a single included stack file. Evaluates the include's path expression using the supplied eval context (so values/locals/functions are available).
+func mergeOneInclude(fs vfs.FS, stackFile *StackFileHCL, inc *StackIncludeHCL, stackDir string, evalCtx *hcl.EvalContext, srcByAutoInclude map[*AutoIncludeHCL][]byte) error {
+	includePath, err := EvalString(inc.Path, evalCtx, attrPath)
+	if err != nil {
+		return IncludeValidationError{IncludeName: inc.Name, Reason: "could not evaluate include path: " + err.Error()}
```
yhakbar: We lose the inner error type here. Is there any way we could preserve it?

internal/hclparse/stack.go
```diff
@@ -125,28 +124,38 @@ func buildRefAttrs(ref ComponentRef) cty.Value {
...
 
-// ExtractUnitRefs extracts ComponentRef values from parsed UnitBlockHCL slices.
-func ExtractUnitRefs(units []*UnitBlockHCL) []ComponentRef {
+// ExtractUnitRefs extracts ComponentRef values from parsed UnitBlockHCL slices. evalCtx is used to evaluate each unit's lazy Path expression; pass nil for literal-only paths.
+func ExtractUnitRefs(units []*UnitBlockHCL, evalCtx *hcl.EvalContext) []ComponentRef {
 	refs := make([]ComponentRef, 0, len(units))
 
 	for _, u := range units {
+		path, err := EvalString(u.Path, evalCtx, attrPath)
+		if err != nil {
```
yhakbar: We just skip these? Why is that correct?

internal/hclparse/stack.go
```diff
@@ -125,28 +124,38 @@ func buildRefAttrs(ref ComponentRef) cty.Value {
...
 
-// ExtractStackRefs extracts ComponentRef values from parsed StackBlockHCL slices.
-func ExtractStackRefs(stacks []*StackBlockHCL) []ComponentRef {
+// ExtractStackRefs extracts ComponentRef values from parsed StackBlockHCL slices. evalCtx is used to evaluate each stack's lazy Path expression; pass nil for literal-only paths.
+func ExtractStackRefs(stacks []*StackBlockHCL, evalCtx *hcl.EvalContext) []ComponentRef {
 	refs := make([]ComponentRef, 0, len(stacks))
 
 	for _, s := range stacks {
+		path, err := EvalString(s.Path, evalCtx, attrPath)
+		if err != nil {
```
yhakbar: We just skip these? Why is that correct?

internal/hclparse/stack.go
```diff
@@ -249,40 +266,81 @@ func discoverStackChildUnitsWithDepth(fs vfs.FS, stackSourceDir, stackGenDir str
...
-			Path: unitPath,
-		})
+	return refs
+}
+
+// discoverUnitRef builds a ComponentRef for a unit block in a discovered nested stack. Returns ok=false when the unit's Path expression cannot be evaluated (best-effort discovery).
+func discoverUnitRef(u *UnitBlockHCL, evalCtx *hcl.EvalContext, childTargetDir, stackGenDir string) (ComponentRef, bool) {
+	unitRelPath, evalErr := EvalString(u.Path, evalCtx, attrPath)
+	if evalErr != nil {
+		return ComponentRef{}, false
```
yhakbar: Why aren't we returning the original error?

internal/hclparse/stack.go
```diff
@@ -249,40 +266,81 @@ func discoverStackChildUnitsWithDepth(fs vfs.FS, stackSourceDir, stackGenDir str
...
 
-		nestedSourceDir := s.Source
-		if !filepath.IsAbs(nestedSourceDir) {
-			nestedSourceDir = filepath.Join(stackSourceDir, nestedSourceDir)
-		}
+// discoverStackRef builds a ComponentRef (with ChildRefs) for a stack block in a discovered nested stack. Returns ok=false when Path or Source cannot be evaluated (best-effort discovery).
+func discoverStackRef(fs vfs.FS, s *StackBlockHCL, evalCtx *hcl.EvalContext, stackSourceDir, stackGenDir, childTargetDir string, depth int) (ComponentRef, bool) {
+	stackRelPath, evalErr := EvalString(s.Path, evalCtx, attrPath)
+	if evalErr != nil {
+		return ComponentRef{}, false
```
yhakbar: Why aren't we returning the original error?

internal/hclparse/stack.go
```diff
@@ -249,40 +266,81 @@ func discoverStackChildUnitsWithDepth(fs vfs.FS, stackSourceDir, stackGenDir str
...
-		})
+	nestedGenPath := filepath.Join(childTargetDir, stackRelPath)
+	if s.NoStack != nil && *s.NoStack {
+		nestedGenPath = filepath.Join(stackGenDir, stackRelPath)
 	}
 
-	return refs
+	nestedSourceDir, sourceErr := EvalString(s.Source, evalCtx, attrSource)
+	if sourceErr != nil {
+		return ComponentRef{}, false
```
yhakbar: Why aren't we returning the original error?

internal/hclparse/stack_test.go
```diff
@@ -7,11 +7,17 @@ import (
...
 	"github.com/gruntwork-io/terragrunt/internal/hclparse"
 	"github.com/gruntwork-io/terragrunt/internal/vfs"
+	"github.com/hashicorp/hcl/v2"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/zclconf/go-cty/cty"
 )
 
+// staticStringExpr builds a literal-string hcl.Expression suitable for the lazy Path/Source fields on UnitBlockHCL / StackBlockHCL.
+func staticStringExpr(s string) hcl.Expression {
```
yhakbar: NIT: This should be sorted lower in the file.

pkg/config/dependency.go
```diff
@@ -830,6 +828,28 @@ func tryGetStackOutput(
 	return &result, true, nil
 }
 
+// resolveStackFilePath rewrites a dependency target path into a candidate terragrunt.stack.hcl path.
+// getCleanedTargetConfigPath appends terragrunt.hcl to bare directory deps, so peel that back to
+// <dir>/terragrunt.stack.hcl. Explicit terragrunt.hcl/terragrunt.hcl.json deps still target that unit
+// and must not be upgraded to a sibling stack.
+func resolveStackFilePath(rawConfigPath string, targetConfigPath string) (string, bool) {
```
yhakbar:
```suggestion
func resolveStackFilePath(rawConfigPath, targetConfigPath string) (string, bool) {
```

NIT: The bot is right.

pkg/config/dependency.go
```diff
@@ -830,6 +828,28 @@ func tryGetStackOutput(
...
+	switch filepath.Base(filepath.Clean(rawConfigPath)) {
+	case DefaultStackFile:
+		return targetConfigPath, true
+	case DefaultTerragruntConfigPath, DefaultTerragruntJSONConfigPath:
+		return "", false
```
yhakbar, line 832: I don't understand what `false` means here. Also, we should probably be threading in whatever the user supplied as the value for `--config` until we get rid of that.

docs/src/data/changelog/v1.0.5/stack-dependencies-permissive-parser.mdx
```diff
@@ -0,0 +1,8 @@
+---
+version: "v1.0.5"
+category: "experiments-updated"
+---
+
+#### `stack-dependencies`: phased parser for `terragrunt.stack.hcl`
+
+`terragrunt.stack.hcl` is now parsed in four phases (slurp → locals → includes → unit/stack decode), so any attribute that the production eval context can resolve (including `local.*`, `values.*`, terragrunt functions) works on unit/stack blocks; `terragrunt run --all` discovery walks generated nested stack files via a path-only decode shape; the duplicated strict/permissive parser paths were unified.
```
yhakbar: Is there something more professional than slurp we can use here? 😁

internal/hclparse/parse.go
```diff
@@ -78,377 +141,471 @@ func ParseStackFile(fs vfs.FS, input *ParseStackFileInput) (*ParseResult, error)
...
 	}
 
-	// Build component refs with absolute paths for the eval context.
-	// Default target is StackDir/.terragrunt-stack/{unit.path}, but units
-	// with no_dot_terragrunt_stack go to StackDir/{unit.path} instead.
-	stackTargetDir := filepath.Join(input.StackDir, StackDir)
-
-	unitRefs := buildRefsWithAbsPath(stackTargetDir, stackFile.Units)
-	stackRefs := buildStackRefsWithAbsPath(fs, input.StackDir, stackTargetDir, stackFile.Stacks, 0)
+	for name, fn := range input.Functions {
```
yhakbar: NIT: Use `maps.Copy` or `maps.Clone`. Don't remember which one it is.

internal/hclparse/parse.go
```diff
@@ -78,377 +141,471 @@ func ParseStackFile(fs vfs.FS, input *ParseStackFileInput) (*ParseResult, error)
...
 	if input.Values != nil {
 		evalCtx.Variables[varValues] = *input.Values
 	}
 
-	// Evaluate locals block iteratively.
-	if stackFile.Locals != nil {
-		if err := evaluateLocals(stackFile.Locals.Remain, evalCtx); err != nil {
-			return nil, err
-		}
+	// Seed unit/stack as empty objects so autoinclude bodies that reference unit.<X>.path before refs are built (Phase 4) get a clear "Unsupported attribute" diagnostic instead of "Unknown variable".
```
yhakbar: I don't understand why they would get this at all. They _should_ be populated when we get to parsing them, because we just hold onto the AST before then, right?

internal/hclparse/parse.go
```diff
@@ -78,377 +141,471 @@ func ParseStackFile(fs vfs.FS, input *ParseStackFileInput) (*ParseResult, error)
...
-// Uses Variables() to pre-check whether each local's dependencies are satisfied
-// before attempting evaluation. Shrinks the work set each pass. Returns an error
-// if any locals cannot be evaluated (cycle or invalid reference).
-func evaluateLocals(body hcl.Body, evalCtx *hcl.EvalContext) error {
-	syntaxBody, ok := body.(*hclsyntax.Body)
-	if !ok {
-		// Non-syntax bodies (e.g. from JSON configs) cannot be iteratively evaluated.
-		return nil
+// decodeRemain wraps gohcl.DecodeBody and returns a non-nil hcl.Diagnostics only when there are errors.
+func decodeRemain(body hcl.Body, evalCtx *hcl.EvalContext, target any) hcl.Diagnostics {
```
yhakbar: Don't really get why we need this.

internal/hclparse/parse.go
```diff
@@ -78,377 +141,471 @@ func ParseStackFile(fs vfs.FS, input *ParseStackFileInput) (*ParseResult, error)
...
+		if s.NoStack != nil && *s.NoStack {
+			stackGenPath = filepath.Join(filepath.Dir(stackTargetDir), s.Path)
 		}
 
-		val, diags := attr.Expr.Value(evalCtx)
-		if diags.HasErrors() {
-			return false, LocalEvalError{Name: name, Detail: diags.Error()}
+		ref := ComponentRef{Name: s.Name, Path: stackGenPath}
+
+		if sourceDir, ok := localStackSourceDir(s.Source, stackDir); ok {
```
yhakbar: Why does it matter what the source is? We shouldn't be calling `filepath.Join` on sources anyways, right?

internal/hclparse/parse.go
```diff
@@ -78,377 +141,471 @@ func ParseStackFile(fs vfs.FS, input *ParseStackFileInput) (*ParseResult, error)
...
 
-	if resolved != nil {
-		resolved.EvalCtx = evalCtx
-		resolved.Kind = kind
-		resolved.SourceBytes = sourceBytes
+	return cty.ObjectVal(evaluated)
+}
+
+// sortedKeys returns the keys of a map[string]X in sorted order. Used to make iteration deterministic for cycle detection and error messages.
+func sortedKeys[V any](m map[string]V) []string {
```
yhakbar: NIT: We can replace this function with this:

```go
slices.Sorted(maps.Keys(m))
```

internal/hclparse/parse.go
```diff
@@ -78,377 +141,471 @@ func ParseStackFile(fs vfs.FS, input *ParseStackFileInput) (*ParseResult, error)
...
+
+	return keys
 }
 
-// processStackIncludes resolves include blocks by parsing the included files and merging their unit/stack blocks into the main stack file. srcByAutoInclude is populated with per-block source bytes from each included file.
-func processStackIncludes(fs vfs.FS, stackFile *StackFileHCL, stackDir string, srcByAutoInclude map[*AutoIncludeHCL][]byte) error {
-	for _, inc := range stackFile.Includes {
-		if err := mergeOneInclude(fs, stackFile, inc, stackDir, srcByAutoInclude); err != nil {
-			return err
+// mergeIncludes evaluates each include block's path expression, parses the included file, and merges its Remain into the parent's. Included files must not declare locals or nested includes. Returns the merged Remain that Phase 4 decodes. srcByFilename is updated with the included file's bytes keyed by filename.
```
yhakbar: We should make sure to document this limitation on locals or nested includes in the docs for `autoinclude`.

internal/hclparse/parse.go
```diff
@@ -78,377 +141,471 @@ func ParseStackFile(fs vfs.FS, input *ParseStackFileInput) (*ParseResult, error)
...
+func mergeIncludes(fs vfs.FS, slurp *StackFileHCL, stackDir string, evalCtx *hcl.EvalContext, srcByFilename map[string][]byte) (hcl.Body, error) {
+	bodies := []hcl.Body{slurp.Remain}
+
+	for _, inc := range slurp.Includes {
+		includedRemain, includedSrc, includedPath, err := mergeOneInclude(fs, inc, stackDir, evalCtx)
```
yhakbar: NIT: This should probably be a struct return.

internal/hclparse/parse.go
```diff
@@ -78,377 +141,471 @@ func ParseStackFile(fs vfs.FS, input *ParseStackFileInput) (*ParseResult, error)
...
 }
 
-// mergeOneInclude reads and merges a single included stack file.
-func mergeOneInclude(fs vfs.FS, stackFile *StackFileHCL, inc *StackIncludeHCL, stackDir string, srcByAutoInclude map[*AutoIncludeHCL][]byte) error {
-	includePath := inc.Path
+// mergeOneInclude reads and slurps a single included file. Returns (includedRemain, includedSrc, includedPath, err) where includedPath is the absolute filename used for HCL diagnostics.
+func mergeOneInclude(fs vfs.FS, inc *StackIncludeHCL, stackDir string, evalCtx *hcl.EvalContext) (hcl.Body, []byte, string, error) {
+	pathVal, diags := inc.Path.Value(evalCtx)
+	if diags.HasErrors() {
+		return nil, nil, "", IncludeValidationError{IncludeName: inc.Name, Reason: "could not evaluate include path: " + diags.Error()}
```
yhakbar: NIT: We're erasing this error here. We probably want to preserve it in a field.

internal/hclparse/parse_test.go
```diff
@@ -683,6 +687,30 @@ unit "vpc" {
...
+}
+`
+
+	_, err := hclparse.ParseStackFile(vfs.NewMemMapFS(), &hclparse.ParseStackFileInput{
+		Src:      []byte(src),
+		Filename: "terragrunt.stack.hcl",
+		StackDir: testStackDir,
+	})
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "vpc_path")
```
yhakbar: NIT: We should have a better way of asserting on errors than checking the strings.

internal/hclparse/parse_test.go
```diff
@@ -1126,3 +1154,357 @@ unit "app" {
...
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			_, err := hclparse.ParseStackFile(vfs.NewMemMapFS(), &hclparse.ParseStackFileInput{
+				Src: []byte(tc.src), Filename: "terragrunt.stack.hcl", StackDir: testStackDir,
+			})
+			require.Error(t, err)
+			assert.Contains(t, err.Error(), "`values` is not allowed inside `autoinclude`")
```
yhakbar: NIT: Can't we assert on the error type here?

internal/hclparse/parse_test.go
```diff
@@ -1126,3 +1154,357 @@ unit "app" {
...
+  path = feature.include_file.value
+}
+
+unit "app" {
+  source = "../catalog/units/app"
+  path   = "app"
+
+  autoinclude {
+    dependency "extra" {
+      config_path = unit.extra.path
```
yhakbar: Wait, this is different semantics than we have for unit includes. Why don't users have to go through `include.extra.unit.extra.path` to get the value?

internal/hclparse/parse_test.go
```diff
@@ -1126,3 +1154,357 @@ unit "app" {
...
+
+	require.NoError(t, err)
+
+	resolved := result.AutoIncludes[hclparse.AutoIncludeKey(hclparse.KindUnit, "app")]
+	require.NotNil(t, resolved)
+	require.Len(t, resolved.Dependencies, 1)
+	assert.Equal(t, filepath.Join("/test", ".terragrunt-stack", "extra"), resolved.Dependencies[0].ConfigPath)
+}
+
+func TestParseStackFile_LocalEvaluatedOnce(t *testing.T) {
```
yhakbar, line 1414: This is great.

internal/hclparse/parse_test.go
```diff
@@ -1126,3 +1154,357 @@ unit "app" {
...
+
+unit "app" {
+  source = "../catalog/units/app"
+  path   = "app"
+}
+`)
+
+	_, err := hclparse.ParseStackFile(fs, &hclparse.ParseStackFileInput{
+		Src:      src,
+		Filename: "/test/terragrunt.stack.hcl",
```
yhakbar: Oh, it's the full path to the file? I thought this was the basename of the file?

docs/src/data/changelog/v1.0.6/stack-dependencies-permissive-parser.mdx
```diff
@@ -0,0 +1,8 @@
+---
+version: "v1.0.6"
+category: "experiments-updated"
+---
+
+#### `stack-dependencies`: phased parser for `terragrunt.stack.hcl`
+
+`terragrunt.stack.hcl` is now parsed in four phases: parse root, then locals, then includes, then unit/stack decode. Expressions can reference `local.*`, `values.*`, and Terragrunt functions. Run-all discovery uses path-only nested traversal, unifying the strict and permissive parser paths.
```
yhakbar: Please update this after merging. I don't think the average Terragrunt user is going to know what this means.

internal/hclparse/autoinclude.go
```diff
@@ -264,7 +278,7 @@ func readAutoIncludeBody(fs vfs.FS, path string) (*hclsyntax.Body, error) {
 
 	file, diags := hclsyntax.ParseConfig(data, path, hcl.Pos{Line: 1, Column: 1})
 	if diags.HasErrors() {
-		return nil, FileParseError{FilePath: path, Detail: diags.Error()}
+		return nil, FileParseError{FilePath: path, Detail: diags.Error(), Err: diags}
```
yhakbar: Curious that we're loading the diags twice here. Please don't edit this now if you think that's a mistake.

internal/hclparse/parse.go
```diff
@@ -78,377 +139,434 @@ func ParseStackFile(fs vfs.FS, input *ParseStackFileInput) (*ParseResult, error)
...
 	}
+}
 
-	file, diags := hclsyntax.ParseConfig(input.Src, input.Filename, hcl.Pos{Line: 1, Column: 1})
+// parseStackFileRoot parses only locals/include blocks and leaves units/stacks in Remain.
+func parseStackFileRoot(src []byte, filename string) (*StackFileHCL, error) {
+	file, diags := hclsyntax.ParseConfig(src, filename, hcl.Pos{Line: 1, Column: 1})
 	if diags.HasErrors() {
-		return nil, FileParseError{FilePath: input.Filename, Detail: diags.Error()}
+		return nil, FileParseError{FilePath: filename, Detail: diags.Error(), Err: diags}
```
yhakbar: Another place where we've got doubled-up diags.

internal/hclparse/parse.go
```diff
@@ -78,377 +139,434 @@ func ParseStackFile(fs vfs.FS, input *ParseStackFileInput) (*ParseResult, error)
...
+	return stackFile, nil
+}
 
-	// Process includes: merge included units/stacks.
-	if err := processStackIncludes(fs, stackFile, input.StackDir, srcByAutoInclude); err != nil {
-		return nil, err
+// buildBaseEvalContext builds the eval context for phases two to four.
+func buildBaseEvalContext(input *ParseStackFileInput) *hcl.EvalContext {
+	evalCtx := &hcl.EvalContext{
+		Functions: map[string]function.Function{},
```
yhakbar: Can we pre-alloc to len of input.Functions and input.Variables?

internal/hclparse/parse.go
```diff
@@ -78,377 +139,434 @@ func ParseStackFile(fs vfs.FS, input *ParseStackFileInput) (*ParseResult, error)
...
-
-	unitRefs := buildRefsWithAbsPath(stackTargetDir, stackFile.Units)
-	stackRefs := buildStackRefsWithAbsPath(fs, input.StackDir, stackTargetDir, stackFile.Stacks, 0)
+	maps.Copy(evalCtx.Functions, input.Functions)
+	maps.Copy(evalCtx.Variables, input.Variables)
 
-	// Pass 2: resolve autoinclude blocks using the eval context.
-	evalCtx := BuildAutoIncludeEvalContext(unitRefs, stackRefs)
+	// Strip parser-owned namespaces so unevaluated references fail with "no variable named X" instead of leaking caller values; phases 2/4 and input.Values populate them when applicable.
+	for _, name := range []string{varLocal, varUnit, varStack, varValues} {
```
yhakbar: NIT: We can use `maps.DeleteFunc` here.

internal/hclparse/parse.go
```diff
@@ -78,377 +139,434 @@ func ParseStackFile(fs vfs.FS, input *ParseStackFileInput) (*ParseResult, error)
...
+	for _, s := range stacks {
+		stackGenPath := filepath.Join(stackTargetDir, s.Path)
+		if s.NoStack != nil && *s.NoStack {
+			stackGenPath = filepath.Join(filepath.Dir(stackTargetDir), s.Path)
 		}
 
-		val, diags := attr.Expr.Value(evalCtx)
-		if diags.HasErrors() {
-			return false, LocalEvalError{Name: name, Detail: diags.Error()}
+		sourceDir := s.Source
```
yhakbar, line 239: This only works if the source is local, right? We don't want to do this if the path is remote, I think.

internal/hclparse/parse.go
```diff
@@ -78,377 +139,434 @@ func ParseStackFile(fs vfs.FS, input *ParseStackFileInput) (*ParseResult, error)
...
-// localsEvalCycleError builds an error listing the locals that could not be evaluated.
-func localsEvalCycleError(remaining map[string]*hclsyntax.Attribute) error {
-	names := make([]string, 0, len(remaining))
-	for name := range remaining {
-		names = append(names, name)
+// evaluateLocals resolves locals in dependency order and catches cycles.
+func evaluateLocals(body hcl.Body, evalCtx *hcl.EvalContext) error {
+	syntaxBody, ok := body.(*hclsyntax.Body)
+	if !ok {
+		return nil
```
yhakbar: Should this be nil? When is a non `body.(*hclsyntax.Body)` going to be passed in here?

internal/hclparse/partial_eval.go
```diff
@@ -133,6 +172,33 @@ func IsPure(expr hclsyntax.Expression, deferred map[string]bool) bool {
 	return true
 }
 
+// containsFunctionCall reports whether expr contains any FunctionCallExpr anywhere in its AST. Uses hclsyntax.Walk so any node type (including ones we don't enumerate) is covered.
+func containsFunctionCall(expr hclsyntax.Expression) bool {
```
yhakbar, line 210: I'm not exactly sure why we do this. Would like to understand that better.

internal/hclparse/partial_eval_test.go
```diff
@@ -243,6 +245,157 @@ func parseFirstAttrExpr(t *testing.T, src string) (hclsyntax.Expression, []byte)
...
+				called = true
+				return cty.StringVal("executed"), nil
+			},
+		}),
+	}
+
+	expr, srcBytes := parseFirstAttrExpr(t, `val = danger()`)
+	result := string(hclparse.PartialEval(expr, &hclparse.EvalArgs{SrcBytes: srcBytes, EvalCtx: evalCtx, Deferred: testDeferred}))
+
+	assert.False(t, called, "partial evaluation must preserve function calls instead of executing them during generation")
```
yhakbar, line 281: Does this mean that `get_terragrunt_dir()` isn't gonna fire when parsing the `autoinclude` block to generate `terragrunt.autoinclude.hcl`, but instead be copied verbatim inside `terragrunt.autoinclude.hcl`, and use the working directory context of the including unit? If so, I think this is going to confuse people. Why have locals that you can reference one way, with functions that execute differently?

internal/hclparse/partial_eval_test.go
```diff
@@ -243,6 +245,157 @@ func parseFirstAttrExpr(t *testing.T, src string) (hclsyntax.Expression, []byte)
...
+	t.Parallel()
+
+	const depth = 20000
+
+	hcl := "val = " + strings.Repeat("(", depth) + "local.env" + strings.Repeat(")", depth)
+	expr, srcBytes := parseFirstAttrExpr(t, hcl)
+
+	result := string(hclparse.PartialEval(expr, &hclparse.EvalArgs{SrcBytes: srcBytes, EvalCtx: buildEvalCtx(), Deferred: testDeferred}))
+
+	assert.NotEmpty(t, result, "deeply nested input must return source-byte fallback, not panic")
```
yhakbar: Is this good? Shouldn't we throw an error or something?

internal/hclparse/partial_eval_test.go
```diff
@@ -243,6 +245,157 @@ func parseFirstAttrExpr(t *testing.T, src string) (hclsyntax.Expression, []byte)
...
+			evalCtx := &hcl.EvalContext{
+				Variables: map[string]cty.Value{
+					"local": cty.ObjectVal(map[string]cty.Value{"flag": tc.flagVal}),
+				},
+			}
+
+			expr, srcBytes := parseFirstAttrExpr(t, `val = local.flag ? "yes" : "no"`)
+			result := string(hclparse.PartialEval(expr, &hclparse.EvalArgs{SrcBytes: srcBytes, EvalCtx: evalCtx, Deferred: testDeferred}))
+
+			assert.Contains(t, result, `local.flag ? "yes" : "no"`, "null/unknown condition must fall back to source, not silently pick a branch")
```
yhakbar: Shouldn't this also error?

internal/hclparse/stack.go
```diff
@@ -243,46 +220,180 @@ func discoverStackChildUnitsWithDepth(fs vfs.FS, stackSourceDir, stackGenDir str
...
+func isLocalStackSource(src string) bool {
+	if src == "" {
+		return false
+	}
+
+	if strings.Contains(src, "::") {
+		return false
+	}
+
+	for _, prefix := range []string{"http://", "https://", "s3://", "gcs://", "file://", "git@"} {
```
yhakbar: I think the fact that we're doing this is a bad sign. Seems error prone, and we shouldn't be caring about stack sources at all.

pkg/config/stack_internal_test.go
```diff
@@ -0,0 +1,94 @@
+package config
```
yhakbar: We're testing inside the package instead of in a `config_test` here.

pkg/config/stack_internal_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestReadStackOrigin_AcceptsAbsoluteExistingDir pins the happy path: a valid sidecar pointing at an absolute existing directory returns that directory.
+func TestReadStackOrigin_AcceptsAbsoluteExistingDir(t *testing.T) {
+	t.Parallel()
+
+	tmp := t.TempDir()
```
yhakbar: Why aren't we using in-memory fs here?

yhakbar: Same for the rest of the file.

test/fixtures/stacks/stack-deps-nested-same-name/README.md
yhakbar, line 1: Why does this fixture have a README?

yhakbar (dismissed):
I'm willing to have us address the NITs as we go. I don't want this to remain unmerged too much longer. Please use your best judgement for whether we should merge now or address some last minute things before merging.

yhakbar (approved):
I think we might want some more changes, but we should address it incrementally off main. I think this is good to merge.

Issue #4535, opened 2025-07-12
Race condition in stack generate with nested stacks


HauserV:
## Describe the bug

Running `terragrunt stack generate` on a project with nested stacks can lead to an intermittent race condition on subsequent runs (i.e., when a `.terragrunt-stack` directory already exists). This results in spurious and non-deterministic failures, most commonly `mkdir: no such file or directory` or validation errors like `expected unit to generate with terragrunt.hcl file`.

The root cause is that the initial file discovery process in `listStackFiles` finds both the user's source `terragrunt.stack.hcl` and the previously generated one inside the `.terragrunt-stack` directory. The worker pool then attempts to process both stack files concurrently. This creates a race where one worker (processing the outer stack) may delete or modify a directory via `go-getter`'s cleanup process at the exact moment other workers (processing the inner, stale stack) are trying to create subdirectories or write files within that same directory structure.

The issue is intermittent and seems to be triggered by the timing of remote `go-getter` operations. The issue disappears completely when running with `--parallelism 1` or when using `TG_SOURCE_MAP` to map remote Git sources to local paths, as both scenarios effectively eliminate the race window.

## Steps To Reproduce

Creating a minimal, self-contained example that reliably reproduces this timing-dependent issue has proven difficult. The failure appears to depend on a specific asymmetry in task execution time that is present in my real-world setup but not in simpler examples.

My setup that consistently triggers the failure involves two repositories: a "live" repository and a "catalog" repository.

1.  **"Live" repository structure:**
    ```
    .
    ├── dev
    │   ├── app
    │   │   └── terragrunt.stack.hcl
    │   └── tier.hcl
    └── root.hcl
    ```

2.  **`dev/app/terragrunt.stack.hcl` content:**
    This file defines a single `stack` that points to a `terragrunt.stack.hcl` in the remote catalog repository.
    ```hcl
    stack "app_infrastructure" {
      source = "git::<my-catalog-repo-url>//stacks/app-infra?ref=main"
      path   = "non-prod"
	  
	  // ...
    }
    ```

3.  **"Catalog" repository structure:**
    The remote catalog repo contains the nested stack definition, which in turn defines multiple `unit` blocks, each pointing to other remote Git sources.
    ```
    .
    ├── modules
    │   ├── batch
    │   │   ├── main.tf
    │   │   ├── outputs.tf
    │   │   └── variables.tf
    │   ├── online
    │   │   ├── main.tf
    │   │   ├── outputs.tf
[...]

Issue #4289, opened 2025-05-15
Recursive stack not generated correctly


mark854542:
## Describe the bug

After solving an error in a recursive stack definition I need to run stack generate two times to be succesfull
possible due to parallel generation of nested stacks

## Steps To Reproduce

Cleanup
```
❯ find .  -name .terragrunt-stack -exec rm -rf {} \;
find: ‘./level1/.terragrunt-stack’: No such file or directory
```
start with error in source definiton ` Source:      ../unit_ax`
```
❯ terragrunt stack generate        
20:58:02.292 INFO   Generating stack from ./level1/terragrunt.stack.hcl
20:58:02.295 INFO   Processing stack stack_a from ./level1/terragrunt.stack.hcl
20:58:02.297 INFO   Generating stack from ./level1/.terragrunt-stack/stack_a/terragrunt.stack.hcl
20:58:02.297 INFO   Processing unit unit_a from ./level1/.terragrunt-stack/stack_a/terragrunt.stack.hcl
20:58:02.298 ERROR  error occurred:

* Failed to fetch unit unit_a
    Source:      ../unit_ax
    Destination: ./level1/.terragrunt-stack/stack_a/.terragrunt-stack/unit_a
  
  Troubleshooting:
    1. Check if your source path is correct relative to the stack file location
    2. Verify the units or stacks directory exists at the expected location
    3. Ensure you have proper permissions to read from source and write to destination
  
  Original error: Failed to fetch ../unit_ax ./level1/.terragrunt-stack/stack_a/.terragrunt-stack/unit_a for unit_a stat ../unit_ax: no such file or directory

20:58:02.300 ERROR  Unable to determine underlying exit code, so Terragrunt will exit with error code 1
```
Correct source definition  ` Source:      ../unit_a`
Generate is still failing
```
❯ terragrunt stack generate
20:58:08.333 INFO   Generating stack from ./level1/.terragrunt-stack/stack_a/terragrunt.stack.hcl
20:58:08.333 INFO   Generating stack from ./level1/terragrunt.stack.hcl
20:58:08.333 INFO   Processing unit unit_a from ./level1/.terragrunt-stack/stack_a/terragrunt.stack.hcl
20:58:08.333 INFO   Processing stack stack_a from ./level1/terragrunt.stack.hcl
20:58:08.334 ERROR  2 errors occurred:

* Failed to fetch unit unit_a
    Source:      ../unit_ax
[...]

PR #6075, opened 2026-05-12
fix(stack): Reject overlapping component paths in stack files

igor-raits:
## Description

Fixes a long-standing concurrent-write race in `terragrunt stack generate` where two sibling `unit`/`stack` blocks in the same stack file declare `path` values where one is a directory ancestor of another — e.g. `path = "prod"` alongside `path = "prod/mcp-gateway"`. Under `--parallelism` on CI runners with fast networks, the ancestor's `go-getter` fetch runs `os.RemoveAll(realDst)` (the `//subdir` form's destructive copy in `hashicorp/go-getter/v2/client.go`) after the descendant has already written its `terragrunt.stack.hcl`/`terragrunt.hcl` into that subtree. The descendant's file vanishes, and `validateGeneratedComponent` then surfaces the non-deterministic `Validation failed for stack X at path Y: expected stack to generate with terragrunt.stack.hcl file at root of generated directory.`

Symptoms reported repeatedly over time:
- [#4535](https://github.com/gruntwork-io/terragrunt/issues/4535) — "Race condition in stack generate with nested stacks"
- [#4289](https://github.com/gruntwork-io/terragrunt/issues/4289) — "Recursive stack not generated correctly"

Reduces with `--parallelism 1` / `--source-update` / local sources, none of which are fixes. Previous fix attempts ([#5962](https://github.com/gruntwork-io/terragrunt/pull/5962) merged Apr 2026) addressed canonical-path dedup at the discovery layer but did **not** handle within-parent path overlap, which is the actual root cause for many remaining reports.

This PR rejects the racy configuration at parse time with a diagnostic that names both colliding blocks, e.g.:

```
stack "prod" (path "prod") overlaps with stack "prod-mcp-gateway" (path "prod/mcp-gateway"): one path is a directory ancestor of the other, which races during concurrent generation
```

### What's in scope

- New typed error `OverlappingComponentPathsError` in `pkg/config/errors.go`.
- New `validatePathOverlaps` in `pkg/config/stack_validation.go` invoked from the existing `ValidateStackConfig` entry point. Path comparison uses `filepath.ToSlash(filepath.Clean(...))` so `"a/./b"` normalizes correctly; comparison enforces a `/`-separator boundary so `"app"` vs `"app-other"` are **not** flagged.
- The previous per-kind exact-duplicate-path check inside `validateConfigElementsGeneric` is **removed** (it was both narrower and incorrect for `no_dot_terragrunt_stack` components, which emit outside `.terragrunt-stack/` and live in a separate namespace). `validatePathOverlaps` subsumes it across kinds and noStack-aware.
- Unit test coverage: 8 new sub-tests for `TestValidateStackConfig` (identical cross-kind, ancestor across kinds, sibling-with-shared-parent OK, substring-prefix OK, normalization, noStack namespace separation).
- Integration test `TestStacksGenerateOverlappingPathsError` with fixture `test/fixtures/stacks/errors/overlapping-paths/` exercising the end-to-end CLI path.
- Changelog entry at `docs/src/data/changelog/v1.0.5/stack-path-overlap-validation.mdx` documenting the diagnostic and remediation.

### Behavior change

Configurations that used to "work" through a combination of low parallelism, slow network, and lucky goroutine ordering are now hard errors at parse time. This is deliberate — a deterministic message naming both blocks is strictly better than silent corruption surfaced as `expected ... file at root of generated directory` under load. Two existing test cases (`duplicate unit paths`, `duplicate stack paths`) have their expected error strings updated to the new format.

## TODOs

- [x] Update the docs.

Backwards compatibility note: existing valid stack configurations are unaffected. Only configurations that were structurally racy (and could already fail non-deterministically) will now fail deterministically at parse time.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed non-deterministic failures in `terragrunt stack generate` when sibling blocks declare overlapping component paths. Configuration validation now rejects such conflicts at parse time with clear diagnostic messages identifying the colliding blocks. Components with `no_dot_terragrunt_stack` enabled are treated as a separate namespace.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gruntwork-io/terragrunt/pull/6075)

### DIFF
--- a/docs/src/data/changelog/v1.0.5/stack-path-overlap-validation.mdx
+++ b/docs/src/data/changelog/v1.0.5/stack-path-overlap-validation.mdx
@@ -1,0 +1,18 @@
+---
+version: "v1.0.5"
+category: "bug-fixes"
+---
+
+#### Stack files reject overlapping component paths
+
+When two sibling `unit` or `stack` blocks in the same stack file declared paths where one is a directory ancestor of another — for example `path = "prod"` alongside `path = "prod/mcp-gateway"` — concurrent generation could race. The ancestor's fetch ran `RemoveAll` on its destination directory after the descendant had already written its `terragrunt.stack.hcl`/`terragrunt.hcl` there, producing non-deterministic `expected stack/unit to generate with ... file at root of generated directory` failures under high `--parallelism` on CI runners with fast networks.
+
+`terragrunt stack generate` now rejects these configurations at parse time with a diagnostic that names both colliding blocks:
+
+```
+stack "prod" (path "prod") overlaps with stack "prod-mcp-gateway" (path "prod/mcp-gateway"): one path is a directory ancestor of the other, which races during concurrent generation
+```
+
+Components with `no_dot_terragrunt_stack = true` are validated as a separate namespace from regular components, since they emit outside the `.terragrunt-stack/` directory and cannot collide with each other across the boundary.
+
+If you hit this error, restructure the offending block so its `path` does not nest inside another component's destination — typically by flattening (`path = "prod-mcp-gateway"`) or by moving the nested component one level deeper into its own `terragrunt.stack.hcl`.
--- a/pkg/config/errors.go
+++ b/pkg/config/errors.go
@@ -324,3 +324,25 @@ type MarkGlobAsReadRequiresExperimentError struct {
 func (err MarkGlobAsReadRequiresExperimentError) Error() string {
 	return fmt.Sprintf("mark_glob_as_read(%q) in %s requires the 'mark-many-as-read' experiment to be enabled", err.Pattern, err.ConfigPath)
 }
+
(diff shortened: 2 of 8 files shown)


yhakbar:
Hey @igor-raits ,

Unfortunately, this would be a breaking change, and some folks actually depend on existing behavior.

Terragrunt emits warnings when users have concurrent stack generation to make sure they're aware, and synchronizes access to ensure that stacks are generated deterministically if different levels of stack generation collide with each other.

Please feel free to open a GitHub discussion or reach out in the Terragrunt Discord if you would like to discuss this further. We'd like to get signal from the Terragrunt community that they'd be OK with a change like this before making it opt-in (to avoid a breaking change).

PR #6076, opened 2026-05-12
feat: Supporting all getters in CAS

yhakbar:
## Description

Extends CAS to cover non-Git sources reachable through go-getter (http, s3, gcs, hg, smb) to that all go-getter sources supported by Terragrunt are also supported as CAS getters.

To do this, all of the non-Git getters needed some sort of cheap remote code probe implemented to assess whether the reference has moved remotely using a HEAD request, etc. The existing `ls-remote` path has been pulled into the same probe flow.

As part of this, a `cas.Venv` struct was added to more conveniently plumb through bundles of virtual environment dependencies (in `cas`, that's `vfs` and `GitRunner`). We should look to re-implement this pattern where necessary to define the external dependency boundaries for packages.   

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **New Features**
  * CAS-backed generic fetching with probe-driven cache keys for HTTP(S), S3, GCS, Mercurial and SMB; getter now dispatches non-git sources into CAS.
* **Improvements**
  * Better cache reuse, offline reads and deduplication across runs; materialization preserves file mutability and is more robust under concurrency.
* **Tests**
  * Large expansion of unit, e2e and integration tests covering resolvers, probe/fetch paths, git resolver behavior, and concurrent race scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gruntwork-io/terragrunt/pull/6076)

### DIFF
--- a/internal/cas/cas.go
+++ b/internal/cas/cas.go
@@ -17,9 +17,7 @@ import (
 
 	"github.com/gruntwork-io/terragrunt/internal/errors"
 	"github.com/gruntwork-io/terragrunt/internal/git"
-	"github.com/gruntwork-io/terragrunt/internal/telemetry"
 	"github.com/gruntwork-io/terragrunt/internal/util"
-	"github.com/gruntwork-io/terragrunt/internal/vexec"
 	"github.com/gruntwork-io/terragrunt/internal/vfs"
 	"github.com/gruntwork-io/terragrunt/pkg/log"
 )
@@ -57,12 +55,10 @@ type CloneOptions struct {
 
 // CAS clones a git repository using content-addressable storage.
 type CAS struct {
-	fs         vfs.FS
 	blobStore  *Store
 	treeStore  *Store
 	synthStore *Store
 	gitStore   *GitStore
-	git        *git.GitRunner
 	storePath  string
 	cloneDepth int
 }
@@ -77,21 +73,12 @@ func WithStorePath(path string) Option {
 	}
 }
 
-// WithCloneDepth sets git clone --depth for CAS (positive shallow clone; negative,
-// e.g. -1, means full history with no --depth). Terragrunt validates user-supplied
-// values with ValidateCASCloneDepth (zero is invalid for git). Omit this option to
-// keep cloneDepth unset so per-operation CloneOptions.Depth can fall back to DefaultCASCloneDepth.
+// WithCloneDepth sets git clone --depth for CAS. Positive values request a
+// shallow clone; -1 means full history (no --depth). Zero is invalid for git
+// and is rejected by [ValidateCASCloneDepth] before reaching this option.
 func WithCloneDepth(depth int) Option {
 	return func(c *CAS) {
 		c.cloneDepth = depth
-	}
-}
-
-// WithFS specifies the filesystem for file operations.
-// If not set, defaults to the real OS filesystem.
-func WithFS(fs vfs.FS) Option {
-	return func(c *CAS) {
-		c.fs = fs
 	}
 }
 
@@ -103,17 +90,6 @@ func New(opts ...Option) (*CAS, error) {
 		opt(c)
 	}
 
-	if c.fs == nil {
-		c.fs = vfs.NewOSFS()
-	}
-
-	// CAS shells out to git, which only sees the real disk. Validate
-	// here so a non-OS backing fails at the constructor instead of
-	// from a deeper store-init step.
-	if !vfs.IsOSFS(c.fs) {
-		return nil, ErrGitStoreFSNotOS
-	}
-
 	if c.storePath == "" {
 		cacheDir, err := util.EnsureCacheDir()
 		if err != nil {
@@ -123,40 +99,12 @@ func New(opts ...Option) (*CAS, error) {
 		c.storePath = filepath.Join(cacheDir, "cas", "store")
 	}
 
-	if err := c.fs.MkdirAll(c.storePath, DefaultDirPerms); err != nil {
-		return nil, fmt.Errorf("failed to create CAS store path: %w", err)
-	}
-
-	c.blobStore = NewStore(filepath.Join(c.storePath, "blobs")).WithFS(c.fs)
-	c.treeStore = NewStore(filepath.Join(c.storePath, "trees")).WithFS(c.fs)
-	c.synthStore = NewStore(filepath.Join(c.storePath, "synth", "trees")).WithFS(c.fs)
-
-	for _, s := range []*Store{c.blobStore, c.treeStore, c.synthStore} {
-		if err := c.fs.MkdirAll(s.Path(), DefaultDirPerms); err != nil {
-			return nil, fmt.Errorf("failed to create CAS store subdirectory %s: %w", s.Path(), err)
-		}
-	}
-
-	g, err := git.NewGitRunner(vexec.NewOSExec())
-	if err != nil {
-		return nil, err
-	}
-
-	c.git = g
-
-	gs, err := NewGitStore(c.fs, g, filepath.Join(c.storePath, "git"))
-	if err != nil {
-		return nil, err
-	}
-
-	c.gitStore = gs
+	c.blobStore = NewStore(filepath.Join(c.storePath, "blobs"))
+	c.treeStore = NewStore(filepath.Join(c.storePath, "trees"))
+	c.synthStore = NewStore(filepath.Join(c.storePath, "synth", "trees"))
+	c.gitStore = NewGitStore(filepath.Join(c.storePath, "git"))
 
 	return c, nil
-}
-
-// FS returns the configured filesystem.
-func (c *CAS) FS() vfs.FS {
-	return c.fs
 }
 
 // BlobStore returns the store for blob content.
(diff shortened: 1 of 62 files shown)


internal/getter/resolver_s3.go
```diff
@@ -0,0 +1,262 @@
...
+
+	ctx, cancel := context.WithTimeout(ctx, s3ResolverTimeout)
+	defer cancel()
+
+	client, err := r.client(ctx, region)
+	if err != nil {
+		return "", cas.ErrNoVersionMetadata
+	}
+
+	out, err := client.HeadObject(ctx, &s3.HeadObjectInput{
```
denis256: Should we handle ?version parameter when fetching from s3?

internal/getter/casgetter.go
```diff
@@ -166,3 +173,220 @@ func (g *CASGetter) Detect(req *getter.Request) (bool, error) {
...
+	return g.CAS.Clone(ctx, g.Logger, g.Venv, &opts, GitCloneURL(u.String()))
+}
+
+// getGeneric routes a non-git source through CAS. The archive=false
+// marker Detect injected gets stripped before passing the URL to the
+// inner getter.Client so archive extraction runs there.
+func (g *CASGetter) getGeneric(ctx context.Context, req *getter.Request) error {
+	scheme := strings.ToLower(req.Forced)
+
+	bare, ok := g.fetchers[scheme]
```
denis256: Not sure if alias fetching will work for `fetchers[scheme]` we may need to check aliases
For cases like `gs->gcs`, it may require `g.lookupFetcher()`

internal/cas/stacks.go
```diff
@@ -131,28 +134,39 @@ func (c *CAS) ProcessStackComponent(ctx context.Context, l log.Logger, source, k
...
 
-	// Navigate to the subdir within the cloned repo
 	contentDir := cloneDir
+
 	if subdir != "" {
-		contentDir = filepath.Join(cloneDir, subdir)
+		if filepath.IsAbs(subdir) {
+			cleanup()
+
+			return nil, fmt.Errorf("%w: %q", ErrSourceEscapesRepo, subdir)
```
denis256: Should be returned ErrAbsoluteSource ?

internal/cas/gitstore.go
```diff
@@ -273,34 +226,30 @@ func (s *GitStore) ensureKnownCommit(
...
-// The probe is lock-free on purpose. The per-URL flock serializes
-// fetches so concurrent pack-file writes do not interleave, but
-// rev-parse only reads pack indices and refs, both of which git
-// updates atomically. Acquiring the flock for a read would queue the
-// probe behind any in-flight fetch and erase the offline win.
-func (s *GitStore) ProbeCachedCommit(ctx context.Context, fs vfs.FS, url, rawRef string) (string, bool) {
-	if !vfs.IsOSFS(fs) {
+// Panics when v.FS is not OS-backed; git only sees the real disk.
+func (s *GitStore) ProbeCachedCommit(ctx context.Context, v Venv, url, rawRef string) (string, bool) {
+	if !vfs.IsOSFS(v.FS) {
```
denis256, line 241: No check for nil v, if is passed without initialization  will lead to nil dereference

internal/getter/casgetter.go
```diff
@@ -166,3 +173,220 @@ func (g *CASGetter) Detect(req *getter.Request) (bool, error) {
...
+	}
+
+	u, err := url.Parse(req.Src)
+	if err != nil || u.Scheme == "" {
+		return "", false
+	}
+
+	scheme := strings.ToLower(u.Scheme)
+	switch scheme {
+	case SchemeHTTP, SchemeHTTPS:
```
denis256, line 267: Not sure how the paths to S3 buckets will be handled, like https://my-bucket.s3.amazonaws.com/path.zip 


Looks like will go through HTTPS fetcher which most probably will fail

internal/getter/resolver_hg.go
```diff
@@ -0,0 +1,87 @@
...
+// `--template '{node}'` is used instead of `--id` because `--id`
+// returns the abbreviated 12-char short hash, which is not
+// collision-safe for cache keying.
+func (r *HgResolver) Probe(ctx context.Context, rawURL string) (string, error) {
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return "", fmt.Errorf("parse hg URL %s: %w", rawURL, err)
+	}
+
+	rev := u.Query().Get("rev")
```
denis256, line 58: I was thinking that through rev can be passed different random code `?rev = "tip ; echo pwned"`

internal/getter/resolver_s3.go
yhakbar: It's not supported in go-getter:
https://github.com/hashicorp/go-getter#s3-s3

internal/cas/gitstore.go
yhakbar: It's never not initialized, though.

internal/getter/resolver_s3.go
yhakbar: I was wrong. It's not supported in v1, but it is in v2. Fixed.

internal/getter/casgetter.go
yhakbar: Added check for aliases.

internal/cas/stacks.go
yhakbar: The thing we're trying to protect against is escaping the repo, right? How strongly do you feel about this?

internal/getter/resolver_hg.go
yhakbar: I'm not sure this can result in anything bad. We don't run the arg through a shell or parse it. We would just use it as a single argument to `CommandContext`, which would fail.

internal/getter/casgetter.go
```diff
@@ -146,7 +153,7 @@ func (g *CASGetter) Detect(req *getter.Request) (bool, error) {
 		}
 
 		if _, isFileDetector := detector.(*getter.FileDetector); isFileDetector {
-			info, statErr := g.CAS.FS().Stat(src)
+			info, statErr := g.Venv.FS.Stat(src)
```
denis256, line 170: Should we add an assert for g.Venv.FS == nil?

internal/cas/cas.go
```diff
@@ -168,62 +116,116 @@ func (c *CAS) TreeStore() *Store { return c.treeStore }
...
+	return results[0].Hash, nil
+}
+
+// Clone fetches url into opts.Dir through the CAS, using a [GitResolver]
+// for the probe and ingesting via `git ls-tree -r` / `git cat-file` so the
+// native git blob and tree formats reach the stores intact.
+//
+// TODO: Make options optional
+func (c *CAS) Clone(ctx context.Context, l log.Logger, v Venv, opts *CloneOptions, url string) error {
+	clonedOpts := *opts
```
denis256: Clone does clonedOpts := *opts without checking opts != nil 🤔

internal/cas/cas.go
```diff
@@ -168,62 +116,116 @@ func (c *CAS) TreeStore() *Store { return c.treeStore }
...
+	}
+
+	return results[0].Hash, nil
+}
+
+// Clone fetches url into opts.Dir through the CAS, using a [GitResolver]
+// for the probe and ingesting via `git ls-tree -r` / `git cat-file` so the
+// native git blob and tree formats reach the stores intact.
+//
+// TODO: Make options optional
```
denis256: Most probably should be added `if opts == nil { opts = &CloneOptions{} }`

denis256 (dismissed):

internal/cas/cas.go
yhakbar: Handled it with variadic optional parameters.

internal/cas/cas.go
yhakbar: Removed.

internal/getter/casgetter.go
yhakbar: Added assertions.

internal/getter/resolver_s3.go
```diff
@@ -0,0 +1,342 @@
...
+	// The bare S3 getter forwards ?version= as GetObject's VersionId,
+	// so HeadObject must too. Without this, the probe describes the
+	// current version while the fetch downloads a different one.
+	if target.Version != "" {
+		input.VersionId = aws.String(target.Version)
+	}
+
+	out, err := client.HeadObject(ctx, input)
+	if err != nil {
+		return "", cas.ErrNoVersionMetadata
```
denis256, line 131: `client` not closed in case of error

internal/getter/casgetter.go
```diff
@@ -166,3 +187,244 @@ func (g *CASGetter) Detect(req *getter.Request) (bool, error) {
...
+// archive-extension pre-decompression so req.Dst reaches Get pointing
+// at the original destination instead of a temporary archive path.
+func appendDisableArchive(rawURL string) string {
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return rawURL
+	}
+
+	q := u.Query()
+	if q.Get("archive") != "" {
```
denis256, line 320: If user will set `archive=false` in url, it will not be possible to detect what is CAS and what is user

denis256 (approved):

PR #6080, opened 2026-05-12
chore: Upping strict control test coverage

yhakbar:
## Description

Upping strict control package test coverage to 100%, as there's no reason not to.

As part of this, the logic for strict controls has also been refactored
a bit. There's a notion of "categories", which we don't really use or
document. That's been removed. Instead, strict controls have been made
more flat to better reflect how they're actually used, and achieve
better parity with experiments.

Also dropped some dead code around "suspended" strict controls. We
never used that feature.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **Refactor**
  * Overhauled strict-controls and deprecated-flag/env-var handling for more consistent, centralized behavior.

* **Chores**
  * Removed legacy control category system and simplified related internal APIs.

* **Tests**
  * Expanded unit test coverage across strict-controls, deprecated-flag/env-var handling, and rendering components to improve reliability.


internal/strict/control.go
denis256, line 153: enabledControls.Names() - now returns list , but value is formatted as %s which may be hard to read

denis256 (approved):

PR #6099, opened 2026-05-14
fix: fix ZipSlip path traversal check returning nil in engine extract

jackiesre721:
## Summary

Fix multiple issues in `internal/engine/engine.go` `extract()` function.

### 1. ZipSlip check returns nil instead of actual error 

When the ZipSlip path traversal check detects a malicious entry, it calls `errors.New(err)` where `err` is `nil` (no OS error occurred — this is a validation failure). Due to the custom `errors` package (`internal/errors/errors.go`), `New(nil)` silently returns `nil`, causing the caller to believe extraction succeeded.

**Fix**: Replace `errors.New(err)` with `errors.Errorf(...)` that constructs an explicit error message.

### 2. Defer-in-loop causes file descriptor exhaustion 

The `defer outFile.Close()` and `defer rc.Close()` calls inside the `for` loop accumulate until `extract()` returns. For large ZIP archives with many entries, this can exhaust file descriptors.

**Fix**: Extract the loop body into a standalone `extractFile()` helper function so each `defer` executes when the helper returns, releasing resources promptly.

### 3. Incorrect log message 

The log message on `outFile.Close()` says `"failed to close zip reader"` but the actual resource being closed is the extracted output file.

**Fix**: Change to `"failed to close extracted output file"`.

## Test Plan

- [x] `TestExtract_ZipSlipPathTraversal` — verifies path traversal entries are rejected with a descriptive error
- [x] `TestExtract_ValidZip` — verifies well-formed ZIP archives extract correctly
- [x] `go build ./internal/engine/` — compiles without errors

### DIFF
--- a/internal/engine/extract_test.go
+++ b/internal/engine/extract_test.go
@@ -12,6 +12,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// TestExtract_ZipSlipPathTraversal verifies that extract rejects ZIP archives
+// containing entries that traverse outside the target directory (ZipSlip attack).
 func TestExtract_ZipSlipPathTraversal(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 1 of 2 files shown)


internal/engine/extract_test.go
```diff
@@ -0,0 +1,67 @@
+package engine
```
yhakbar, line 1: Hey @jackiesre721 ,

This is going to violate our linter. Please make sure to run it locally with `make run-lint` before submitting PRs.

We generally only test the public API of packages to make it easier to refactor implementation details. It's fine to export `extractFile` if necessary.

I'll look to fix up this PR in another branch.

PR #6104, opened 2026-05-14
feat: stack outputs parallel fetching

### LINKED ISSUE #5901: Parallelize unit processing in StackOutput
## Describe the enhancement

The current implementation iterates through all units in the stack sequentially 
For each unit, it performs a potentially heavy Opentofu output call.
For stacks with many units, this results in poor performance and significant wait times for users.

## Additional context

Stack outputs fetching and merging in `internal/stacks/output/output.go`.

- [ ] Changes required.
- Refactor the unit processing loop in `StackOutput` to use concurrent execution.                                  
- Implement thread-safe collection of unit outputs into the `outputs` map (e.g., using a `sync.Mutex` or collecting results into a slice first).
- Respect the `Parallelism` setting from `TerragruntOptions` to limit the number of concurrent `terraform output` calls.   
 - Higher memory usage due to concurrent parsing and output buffering
- [ ] Level of effort.
 - Requires careful handling of concurrency and error propagation

## PoC (Proof of Concept)

- [ ] [PoC]()

## RFC Not Needed

denis256:
## Description

PR makes `terragrunt stack output` to fetch outputs concurrently, concurrency is configurable with `--parallelism`

Fixes #5901.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **New Features**
  * `terragrunt stack output` now fetches outputs in parallel rather than sequentially, improving performance on large stacks. Control concurrency via the `--parallelism` flag (default: unbounded) or `TG_PARALLELISM` environment variable to prevent resource exhaustion.

* **Documentation**
  * Added guidance on limiting stack output parallelism with practical examples.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gruntwork-io/terragrunt/pull/6104)


internal/stacks/output/output.go
```diff
@@ -12,10 +12,12 @@ import (
 	"github.com/gruntwork-io/terragrunt/internal/errors"
 	"github.com/gruntwork-io/terragrunt/internal/stacks/generate"
 	"github.com/gruntwork-io/terragrunt/internal/telemetry"
+	"github.com/gruntwork-io/terragrunt/internal/worker"
 	"github.com/gruntwork-io/terragrunt/internal/worktrees"
 	"github.com/gruntwork-io/terragrunt/pkg/config"
 	"github.com/gruntwork-io/terragrunt/pkg/log"
 	"github.com/gruntwork-io/terragrunt/pkg/options"
+	"github.com/puzpuzpuz/xsync/v3"
```
yhakbar, line 20: NIT: We should consider using v4.

internal/stacks/output/output.go
```diff
@@ -99,13 +101,16 @@ func StackOutput(
...
 
-	outputs := make(map[string]map[string]cty.Value)
+	outputs := xsync.NewMapOf[string, map[string]cty.Value]()
 	declaredStacks := make(map[string]string)
 	declaredUnits := make(map[string]*config.Unit)
 
 	// save parsed stacks
 	parsedStackFiles := make(map[string]*config.StackConfig, len(foundFiles))
 
+	wp := worker.NewWorkerPool(opts.Parallelism)
```
yhakbar, line 111: NIT: We should consider using a cap of available cores if the user hasn't customized parallelism.

yhakbar: What do we get out of using this instead of errgroup?

test/integration_stacks_test.go
```diff
@@ -619,6 +620,60 @@ func TestStackOutputsJsonFlag(t *testing.T) {
 	assert.Len(t, result, 4)
 }
 
+// TestStackOutputsParallelFetching verifies that `terragrunt stack output` reads each
+// unit's outputs concurrently. With 12 units fetched in parallel, the order in which
+// units' tofu invocations log varies between runs because goroutine completion is
+// non-deterministic. If the fetcher regressed to a strict serial loop the order would
+// match across runs.
+func TestStackOutputsParallelFetching(t *testing.T) {
```
yhakbar, line 628: NIT: We should suffix this test with `WithRacing` so that CI runs it with the `-race` race detector.

docs/src/data/changelog/v1.0.5/stack-output-parallelism.mdx
```diff
@@ -0,0 +1,14 @@
+---
+version: "v1.0.5"
+category: "new-features"
```
yhakbar, line 3: I think this is better documented as a `performance-improvements` change:
https://github.com/gruntwork-io/terragrunt/blob/main/docs/src/lib/changelog.ts#L13

yhakbar (approved):
Would be nice to add a benchmark to show how performance improves here. I think that would be especially important for seeing if allowing for > cores concurrent output calls on your workstation still results in performance improvements.

PR #6112, opened 2026-05-14
feat: Make version optional in `tfr://` module registry URLs

### LINKED ISSUE #2572: Feature request: support latest version when using `tfr` protocol in `terraform.source`

yhakbar:
## Description

Closes #5801.
Fixes #2572.
Relates to #1930.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **New Features**
  * `tfr://` source URLs now support optional `version` query parameters. When omitted, Terragrunt automatically resolves and downloads the latest stable module version from the registry, excluding prerelease versions.

* **Documentation**
  * Updated documentation and changelog to clarify optional version parameters and automatic latest-version resolution behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gruntwork-io/terragrunt/pull/6112?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


internal/getter/tfr.go
```diff
@@ -113,18 +116,12 @@ func (r *RegistryGetter) Get(ctx context.Context, req *getter.Request) error {
...
 	modulePath, moduleSubDir := SourceDirSubdir(srcURL.Path)
 
-	versionList, hasVersion := queryValues[versionQueryKey]
-	if !hasVersion {
-		return errors.New(MalformedRegistryURLErr{reason: "missing version query"})
-	}
-
-	if len(versionList) != 1 {
-		return errors.New(MalformedRegistryURLErr{reason: "more than one version query"})
+	moduleRegistryBasePath, err := GetModuleRegistryURLBasePath(ctx, r.Logger, r.HTTPClient, registryDomain)
```
denis256, line 119: What will happen if users will pass multiple versions in?
like `?version=1&version=2`

internal/getter/tfrhelpers.go
```diff
@@ -161,6 +163,92 @@ func BuildRequestURL(registryDomain, moduleRegistryBasePath, modulePath, version
...
+
+	if len(versionsResp.Modules) == 0 || len(versionsResp.Modules[0].Versions) == 0 {
+		return "", fmt.Errorf("no versions found for module %s on registry %s", modulePath, registryDomain)
+	}
+
+	// The registry API does not guarantee version ordering, so parse and pick
+	// the max by semver. Skip prereleases and unparsable entries.
+	parsed := make([]*goversion.Version, 0, len(versionsResp.Modules[0].Versions))
+
+	for _, v := range versionsResp.Modules[0].Versions {
```
denis256, line 213: Should be added validation for versionsResp.Modules[1/2/3...]

internal/getter/tfr.go
yhakbar: It's validated here:
https://github.com/gruntwork-io/terragrunt/pull/6112/changes/BASE..436c7cd3e2237cf6d3506ff2d84495a74c400405#diff-bcd78b20602c354cde4fff7493dab2b70abe55cd1f6be8562236f9046037420dR237

internal/getter/tfrhelpers.go
yhakbar: I'm not sure we need that, or what we would want to set for everything after the first element.

denis256 (dismissed):

denis256 (approved):

PR #6128, opened 2026-05-16
feat: stack output works without terragrunt.stack.hcl (implicit stacks)

### LINKED ISSUE #4559: stack outputs for implicit stacks
## Describe the enhancement

`terragrunt stack output` provides a nice and hierarchical way to display outputs for explicit stacks (with a `terragrunt.stack.hcl`). It would be awesome if a similar output could be provided for implicit stacks (without `terragrunt.stack.hcl`)

As there is no explicit name definition for the units, the relative path of its folder shall be used as the key instead.

## Additional context

The current alternative `terragrunt output --all` does not maintain the same hierarchical view and outputs cannot be tied to their individual unit. Outputs could also potentialy even overlap if 2 units declare one with the same name.

I am not sure how this should materialize itself: could be with the introduction on a new switch on either `terragrunt stack` or `terragrunt output`

The effort should be reasonable, as most of the logic is already implemented for explicit stacks.

## PoC (Proof of Concept)

```hcl
$ terragrunt output --all --stack/hierarchical/
./backend_app = {
  domain = "backend-app.example.com"
}
./somefolder/frontend/anotherfolder = {
  domain = "frontend-app.example.com"
}
./mysql = {
  endpoint = "terraform-20250504140737772400000001.abcdefghijkl.us-east-1.rds.amazonaws.com"
}
./somefolder/aws = {
  endpoint = "serverless-valkey-01.amazonaws.com"
}
./vpc = {
  vpc_id = "vpc-1234567890"
}
```

FuNK3Y:
## Description

Fixes #4559.

`terragrunt stack output` previously required a `terragrunt.stack.hcl` file. When run in a directory with only loose units, it warned that no stack files were found and produced nothing.

This PR makes it fall back to discovering units directly under the working directory and aggregates their outputs, keyed by each unit's directory path relative to the working directory.

```bash
$ terragrunt stack output
db = {
  endpoint = "db.local"
  port     = 5432
}
vpc = {
  vpc_id = "vpc-1234567890"
}
apps/api = {
  domain = "api.example.com"
}
```

Filtering and dot drill-down continue to work the same way they do for explicit stacks:

```bash
$ terragrunt stack output vpc.vpc_id --format raw
vpc-1234567890
```

### Design notes

- The path is preserved verbatim (no `./` prefix, no segment splitting) so two units in different directories never collide.
- Drilling into a specific output (`vpc.vpc_id`) rides on the existing `FilterOutputs` dot-traversal — no changes to that function. As a consequence, directory names containing `.` cannot be drilled into via the dot separator. This is the same caveat that already exists for explicit-stack unit names with `.`. Users hitting this can use `--format json` + `jq`.
- The fallback only fires when **no** `terragrunt.stack.hcl` files are found, so existing explicit-stack behavior is unchanged.

### Code shape

- `internal/stacks/output/output.go` — new `implicitStackOutput` helper (mirrors the explicit-stack flow: parallel discovery via worker pool, same `readUnitOutput` path), plus a small `implicitUnitKey` wrapper around `filepath.Rel`.
- The existing `StackOutput` swaps its "no stack files" warn-and-return for a debug-log + fallback.
- `FilterOutputs` and the rest of the stack command package are untouched.

## TODOs

- [x] Update the docs.

### Tests

- New integration test `TestStackOutputsImplicit` covering full output (`json`), path-key filtering, and dot drill-down end-to-end.
- New fixture at `test/fixtures/stacks/outputs-implicit/` with three units (one nested) — no providers, no cloud, no network.
- Existing `internal/stacks/output` and `internal/cli/commands/stack` tests still pass.

### Backwards compatibility

The previous behavior in the no-stack-files case was a `Warn` log and an empty output. The new behavior is strictly additive: if units exist, you get their outputs; if no units exist, the user sees a `Warn` log ("No units found") and an empty output (same observable shape as before).

### Feature flag

Not gated — this is an additive fallback that only fires when the explicit-stack path produces nothing, and replaces a near-useless no-op with useful output. Happy to gate it if maintainers prefer.

## Summary by CodeRabbit

* **New Features**
  * `terragrunt stack output` now works without a stack file by implicitly discovering units under the working directory and aggregating outputs keyed by each unit’s relative path; respects unit-level exclusions.

* **Documentation**
  * Updated changelog and command docs with implicit-stack behavior, examples, path filtering, drill-down (including raw format), and note about dot-containing directory names.

* **Tests**
  * Added integration and fixture tests validating implicit discovery, aggregation, filtering, and drill-down.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gruntwork-io/terragrunt/pull/6128?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


yhakbar:
Hey @FuNK3Y ,

Please rebase this PR and address the merge conflicts.

Also, we've recently updated our settings to require signed commits. Please follow this guidance to ensure that you're using signed commits:
https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

You may have to perform an interactive rebase to get your existing commits updated:
https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

Please let us know if you want any help getting this done.

FuNK3Y:
I did the adjustment required with the help of Claude

docs/src/data/changelog/v1.0.5/stack-output-implicit.mdx
```diff
@@ -0,0 +1,37 @@
+---
+version: "v1.0.5"
```
yhakbar: Please update this version to v1.0.7

docs/src/data/experiments/stack-output-implicit.mdx
```diff
@@ -0,0 +1,47 @@
+---
+name: stack-output-implicit
+status: active
+since: "1.0.5"
```
yhakbar: Please update this to 1.0.7

FuNK3Y:
It has been done @yhakbar

FuNK3Y:
@yhakbar I noticed that 1.0.7 was released, so I am now targeting 1.0.8.

docs/src/data/experiments/stack-output-implicit.mdx
```diff
@@ -0,0 +1,47 @@
+---
+name: stack-output-implicit
+status: active
```
yhakbar:
```suggestion
```

We got rid of this since you started your PR.

yhakbar:
@FuNK3Y 

What happens when you run `terragrunt stack output` in a directory above one or more _explicit_ stacks? Could you add testing for that? Are there design considerations that need to be thought through there?

FuNK3Y:
@yhakbar 

Hooo I was not aware that apply could work with a mix of implicit & explicit stack.

The current & naive implementation is that explicit stack will take precedence: as soon as there is one explicit stack in a subfolder, the output will revert to its current implementation, and only explicit stack will be shown. I added a test case in that sense.

Another angle would be to merge: use path for implicit units, use unit key for explicit ones.

What do you think would make more sense here?

yhakbar:
I think the second approach makes more sense.

On initial discovery, all units and stacks in child directories could be discovered, with any units who's paths are nested under explicit stacks filtered out. You would then do the logic that you're doing now, with leaf nodes of stacks resulting in the production cty values representing their own trees. Each explicit stack would be represented as the same tree it's represented by in production today, and everything else is addressed by ToSlash'ed paths. I think that would be the least surprising behavior.

What do you think?

FuNK3Y:
I concur - this would be the most intuitive approach. I implemented it & added nested stacks to the fixture.

This approach has a drawback though: an explicit key could clash with an implicit folder name; there is no guarantee for unicity anymore.

The current implementation favor explicit & log a warning.

yhakbar:
@FuNK3Y

I think I was unclear in my last comment, sorry. What I had in mind was that explicit stacks would *also* be addressed by their ToSlash'ed paths relative to the working directory, with the tree each stack produces today nested underneath. The position of the stack in the filesystem is what disambiguates it.

Say you have this:

```
live/
├── prod/
│   └── terragrunt.stack.hcl # Declares unit "db"
└── db/
    └── terragrunt.hcl
```

Running `terragrunt stack output` from `live` would address the stack unit as `prod/db` and the loose unit as `db`. They can't collide, because the path from the working directory to each of them is part of the key.

FuNK3Y:
@yhakbar my bad! I made the adjustment and it should mostly mitigate the clashes.

I took the chance to simplify the method breakdown to make sure it reflects today's intent and not the sum of my attempts

FuNK3Y:
This ended up being more complex than anticipated

PR #6129, opened 2026-05-17
fix: support tfr:// module sources in scaffold command

### LINKED ISSUE #6117: terragrunt scaffold command does not work with tfr:// module sources
## Describe the bug

`terragrunt scaffold` fails when given a `tfr://` module source URL pointing to a Terraform module registry (public or private). Despite the [documentation](https://docs.terragrunt.com/features/catalog/scaffold/) stating that any module source URL format is supported, the command errors with `error downloading 'tfr://...'`.

The same `tfr://` source works correctly when placed directly in a `terragrunt.hcl` file and running `terragrunt init` / `terragrunt plan`.

## Reproducing bugs

Run `terragrunt scaffold tfr://registry.example.com/namespace/module/provider` in an empty directory. The module source is valid and works in a plain `terragrunt.hcl`, but scaffold fails to download it.

## Steps To Reproduce

Run the following in an empty directory:

```bash
terragrunt scaffold tfr://registry.example.com/namespace/module/provider
```

Or with an explicit version:

```bash
terragrunt scaffold tfr://registry.example.com/namespace/module/provider?version=1.0.0
```

The following `terragrunt.hcl` works without issue with the same source:

```hcl
terraform {
  source = "tfr://registry.example.com/namespace/module/provider?version=1.0.0"
}
```

## Expected behavior

`terragrunt scaffold tfr://registry.example.com/namespace/module/provider` should query the registry API for the latest version, download the module to inspect its variables, and generate a `terragrunt.hcl` with the correct `tfr://` source pinned to that version.

## Must haves

## Nice to haves

- [x] Terminal output

```bash
$ terragrunt scaffold tfr://registry.example.com/namespace/module/provider
WARN   Failed to parse url tfr://registry.example.com/namespace/module/provider
WARN   Failed to parse module url tfr://registry.example.com/namespace/module/provider
WARN   Failed to find last release tag for tfr://registry.example.com/namespace/module/provider
ERROR  error downloading 'tfr://registry.example.com/namespace/module/provider'
```

## Versions

- Terragrunt version: `terragrunt version v1.0.3`
- OpenTofu/Terraform version: `Terraform v1.13.5`
- Environment details: macOS (darwin/arm64)

## Additional context

There are three separate root causes contributing to the failure:

**1. `RegistryGetter` not wired into the scaffold download client**

`scaffold.go` calls `getter.GetAny(ctx, tempDir, moduleURL)` without passing `getter.WithTFRegistry(...)`. The `RegistryGetter` that handles `tfr://` URLs is only included in the go-getter client when explicitly opted in. Without it, no getter can handle the download.

**2. Spurious warnings for `tfr://` URLs**

`rewriteModuleURL` calls `parseURL` which uses the regex `(?:git|hg|s3|gcs)::([^:]+)://([^/]+)(/.*)`  — only matching `forced-getter::scheme://` style URLs. `tfr://` has no forced-getter prefix so the regex never matches, producing two misleading warnings.

**3. Version auto-detection uses git instead of the registry API**

When no `?version=` is specified, `addRefToModuleURL` calls `shell.GitLastReleaseTag` which attempts `git remote-tfr` — not a valid git command. For `tfr://` URLs the correct approach is to query the registry's `/versions` endpoint and pick the highest semver version.

We have identified a fix for all three issues and are happy to open a PR. The changes are confined to `internal/cli/commands/scaffold/scaffold.go`, `internal/getter/tfrhelpers.go`, and `internal/getter/tfr.go`.

Abzaek:
## Description

This PR fixes three issues that prevented `terragrunt scaffold` from working with `tfr://` (Terraform Registry) module sources.

### Root Causes

**1. RegistryGetter not wired into the scaffold download client** (Fixes the download error)

`scaffold.go` calls `getter.GetAny(ctx, tempDir, moduleURL)` without passing `getter.WithTFRegistry(...)`. The `RegistryGetter` that handles `tfr://` URLs is only included in the go-getter client when explicitly opted in. Without it, no getter can handle the download.

**2. Spurious warnings for tfr:// URLs** (Fixes the misleading warnings)

`rewriteModuleURL` calls `parseURL` which uses a regex only matching forced-getter URLs (`git::https://...` style). `tfr://` has no forced-getter prefix so the regex never matches, producing two misleading warnings about failing to parse the URL.

**3. Version auto-detection uses git instead of the registry API** (Fixes version resolution)

When no `?version=` is specified, `addRefToModuleURL` calls `shell.GitLastReleaseTag` which is not meaningful for `tfr://` URLs. The fix queries the registry's `/versions` endpoint and picks the highest semver version.

### Changes

| File | Change |
|------|--------|
| `internal/getter/tfrhelpers.go` | Added `ResolveTFRVersion()` function that queries the registry API to find the latest version. Added `IsTFRSource()` helper. |
| `internal/cli/commands/scaffold/scaffold.go` | (1) Pass `getter.WithTFRegistry(...)` to both `getter.GetAny` calls. (2) Remove spurious warnings from `parseURL`/`rewriteModuleURL` for non-forced-getter URL formats. (3) Added `resolveTFRVersion()` and updated `addRefToModuleURL` to use registry API for `tfr://` URLs. |

### Related Issues

Closes #6117

## Summary by CodeRabbit

* **New Features**
  * Automatic registry version resolution for tfr:// module sources (appends latest version when no ref supplied).

* **Improvements**
  * Scaffold downloads use registry-aware retrieval for both modules and templates.
  * tfr:// source handling injects resolved version info into URLs and avoids unnecessary git ref lookups.
  * Registry lookup selects the correct registry domain and reliably picks the newest semver.

* **Bug Fixes**
  * Reduced noisy warnings when module URL parsing fails.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gruntwork-io/terragrunt/pull/6129?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/cli/commands/scaffold/scaffold.go
+++ b/internal/cli/commands/scaffold/scaffold.go
@@ -298,6 +298,24 @@ func BuildSourceURL(originalURL, resolvedURL string, vars map[string]any) string
 		return resolvedURL
 	}
 
+	if getter.IsTFRSource(originalURL) {
+		versionVal := ExtractQueryParam(resolvedURL, getter.VersionQueryKey)
+		if versionVal == "" {
+			return originalURL
+		}
+
+		base, rawQuery := splitURLQuery(originalURL)
+
+		params, err := url.ParseQuery(rawQuery)
+		if err != nil || params.Has(getter.VersionQueryKey) {
+			return originalURL
+		}
+
+		params.Set(getter.VersionQueryKey, versionVal)
+
+		return base + "?" + params.Encode()
+	}
+
 	refVal := ExtractQueryParam(resolvedURL, refParam)
 	if refVal == "" {
 		return originalURL
--- a/internal/cli/commands/scaffold/scaffold.go
+++ b/internal/cli/commands/scaffold/scaffold.go
@@ -714,6 +714,11 @@ func addRefToModuleURL(
 	if ref == "" {
 		// For tfr:// URLs, resolve the latest version via the registry API.
 		if getter.IsTFRSource(moduleURL.String()) {
+			// If version is already specified, skip resolution.
+			if params.Has(getter.VersionQueryKey) {
+				return moduleURL, nil
+			}
+
 			latestVersion, err := resolveTFRVersion(ctx, l, opts, moduleURL)
 			if err != nil {
 				return nil, err
@@ -724,7 +729,7 @@ func addRefToModuleURL(
 				return moduleURL, nil
 			}
 
-			params.Add(getter.VersionQueryKey, latestVersion)
+			params.Set(getter.VersionQueryKey, latestVersion)
 			moduleURL.RawQuery = params.Encode()
 
 			return moduleURL, nil
--- a/internal/cli/commands/scaffold/scaffold.go
+++ b/internal/cli/commands/scaffold/scaffold.go
@@ -16,6 +16,7 @@ import (
 	"github.com/gruntwork-io/terragrunt/internal/configbridge"
 	"github.com/gruntwork-io/terragrunt/internal/shell"
 	"github.com/gruntwork-io/terragrunt/internal/telemetry"
+	"github.com/gruntwork-io/terragrunt/internal/tfimpl"
 	"github.com/gruntwork-io/terragrunt/pkg/config"
 	"github.com/gruntwork-io/terragrunt/pkg/log"
 
@@ -45,6 +46,9 @@ const (
 
 	moduleURLPattern = `(?:git|hg|s3|gcs)::([^:]+)://([^/]+)(/.*)`
 	moduleURLParts   = 4
+
+	// versionQueryParam is the query parameter used for tfr:// version resolution.
+	versionQueryParam = "version"
 
 	// TODO: Make the root configuration file name configurable
 	DefaultBoilerplateConfig = `
@@ -197,7 +201,10 @@ func Run(ctx context.Context, l log.Logger, opts *options.TerragruntOptions, mod
 	if err := telemetry.TelemeterFromContext(ctx).Collect(ctx, "scaffold_get_module", map[string]any{
 		"module_url": moduleURL,
 	}, func(ctx context.Context) error {
-		if _, getErr := getter.GetAny(ctx, tempDir, moduleURL); getErr != nil {
+		if _, getErr := getter.GetAny(ctx, tempDir, moduleURL,
+			getter.WithTFRegistry(getter.NewRegistryGetter(l).
+				WithTofuImplementation(opts.TofuImplementation)),
+		); getErr != nil {
 			return fmt.Errorf("downloading scaffold module from %s: %w", moduleURL, getErr)
 		}
 
@@ -436,7 +443,10 @@ func downloadTemplate(
 	if err := telemetry.TelemeterFromContext(ctx).Collect(ctx, "scaffold_get_template", map[string]any{
 		"template_url": baseURL.String(),
 	}, func(ctx context.Context) error {
-		if _, getErr := getter.GetAny(ctx, templateDir, baseURL.String()); getErr != nil {
+		if _, getErr := getter.GetAny(ctx, templateDir, baseURL.String(),
+			getter.WithTFRegistry(getter.NewRegistryGetter(l).
+				WithTofuImplementation(opts.TofuImplementation)),
+		); getErr != nil {
 			return fmt.Errorf("downloading scaffold template from %s: %w", baseURL.String(), getErr)
 		}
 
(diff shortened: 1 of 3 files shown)


internal/getter/tfrhelpers.go
```diff
@@ -223,3 +239,77 @@ func httpGETAndGetResponse(ctx context.Context, l log.Logger, httpClient *http.C
...
+	for i, v := range resp.Modules[0].Versions {
+		versions[i] = v.Version
+	}
+
+	// Sort in descending semver order and return the top.
+	sort.Slice(versions, func(i, j int) bool {
+		return versions[i] > versions[j]
+	})
+
+	return versions[0], nil
```
yhakbar: CodeRabbit is right here.

Please use `github.com/hashicorp/go-version` as is used in the rest of the codebase to do proper semver comparison. Also, prefer `slices.SortFunc` over `sort.Slice`.

internal/cli/commands/scaffold/scaffold.go
```diff
@@ -686,6 +697,19 @@ func addRefToModuleURL(
 
 	ref := params.Get(refParam)
 	if ref == "" {
+		// For tfr:// URLs, resolve the latest version via the registry API.
+		if getter.IsTFRSource(moduleURL.String()) {
+			latestVersion, err := resolveTFRVersion(ctx, l, opts, moduleURL)
+			if err != nil || latestVersion == "" {
+				l.Warnf("Failed to find latest version for %s: %v", moduleURL, err)
```
yhakbar: I don't know that it's safe to ignore this error.

`RegistryGetter.Get` is going to fail if we can't resolve this version.

internal/cli/commands/scaffold/scaffold.go
```diff
@@ -686,6 +697,19 @@ func addRefToModuleURL(
...
+			} else {
```
yhakbar: NIT: We typically prefer to use early returns where possible to avoid `else` statements. Makes the code flow easier to read top to bottom.

internal/cli/commands/scaffold/scaffold.go
```diff
@@ -46,6 +47,9 @@ const (
 	moduleURLPattern = `(?:git|hg|s3|gcs)::([^:]+)://([^/]+)(/.*)`
 	moduleURLParts   = 4
 
+	// versionQueryParam is the query parameter used for tfr:// version resolution.
+	versionQueryParam = "version"
```
yhakbar: This duplicates the constant here:
https://github.com/gruntwork-io/terragrunt/blob/main/internal/getter/tfr.go#L20

We should de-duplicate these by exporting the one in `internal/getter/tfr` and referencing it here.

internal/cli/commands/scaffold/scaffold.go
```diff
@@ -706,11 +730,33 @@ func addRefToModuleURL(
...
+	// default for the current tofu implementation.
+	registryDomain := moduleURL.Host
+	if registryDomain == "" {
+		registryDomain = tfimpl.DefaultRegistryDomain(opts.TofuImplementation)
+	}
+
+	// Extract the module path (e.g. "terraform-aws-modules/vpc/aws")
+	// using SourceDirSubdir which handles the "//subdir" syntax.
+	modulePath, _ := getter.SourceDirSubdir(strings.TrimPrefix(moduleURL.Path, "/"))
+	if modulePath == "" {
```
yhakbar: Isn't this dead code? When does `SourceDirSubdir` return empty for the first return value?

internal/getter/tfrhelpers.go
```diff
@@ -223,3 +239,77 @@ func httpGETAndGetResponse(ctx context.Context, l log.Logger, httpClient *http.C
...
+}
+
+// IsTFRSource checks whether the given URL is a tfr:// registry URL.
+func IsTFRSource(rawURL string) bool {
+	if strings.HasPrefix(rawURL, "tfr://") {
+		return true
+	}
+
+	u, err := url.Parse(rawURL)
+	if err == nil && u.Scheme == "tfr" {
```
yhakbar: Why do we do both of these checks? Shouldn't we get rid of the check above and just do:

```go
u, err := url.Parse(rawURL)
return err == nil && u.Scheme == "tfr"
```

yhakbar (changes requested):

Abzaek:
Addressed all review feedback:

1. ✅ Use `go-version` + `slices.SortFunc` for proper semver sorting in `ResolveTFRVersion`
2. ✅ Error from `resolveTFRVersion` is now propagated (returns `nil, err`)
3. ✅ Early returns instead of `else` blocks in `addRefToModuleURL`
4. ✅ Exported `VersionQueryKey` from `internal/getter/tfr.go`, deduplicated the constant in `scaffold.go`
5. ✅ Removed dead code fallback after `SourceDirSubdir`
6. ✅ Simplified `IsTFRSource` to single `url.Parse` check

Please take another look when you get a chance.

yhakbar:
Please resolve the conflict with `main`, and coordinate with the issue reporter to collaborate on a single, please. I want to make sure that any details in the issue reporter's fix is incorporated into this one (or vice versa) before re-reviewing the single fix.

Abzaek:
✅ Addressed the CodeRabbit comment too:

**BuildSourceURL** now propagates the `version` query param for `tfr://` sources — mirrors the existing `ref` logic. When the resolved URL has `?version=X.Y.Z`, `BuildSourceURL` will inject it into the generated `sourceUrl` so the scaffolded output is properly pinned.

Abzaek:
✅ Also addressed the CodeRabbit review (#4346577852) — version duplication check:

Added `params.Has(getter.VersionQueryKey)` guard before `resolveTFRVersion` to skip resolution when user already specified a version in the URL. Also changed `params.Add` → `params.Set` for the version param to replace rather than append.

Abzaek:
✅ Addressed CodeRabbit review #4346661280 — `--var Ref` now maps to `version` param for `tfr://` sources:

The `refReplacement` block now checks the source type. For `tfr://` URLs, `--var Ref=v1.2.3` writes `?version=v1.2.3` instead of `?ref=v1.2.3`. Non-tfr sources are unchanged.

yhakbar:
This PR has quite a few conflicts. Please ensure that the PR is still necessary, and if so, address the conflicts before re-requesting review.

PR #6166, opened 2026-05-21
fix: Allowing all TG HCL fns in `terragrunt.stack.hcl`, including `autoinclude`

yhakbar:
## Description

Makes it so that users can use any TG HCL function in `terragrunt.stack.hcl` files including in `autoinclude` blocks. As a consequence of these changes, support was dropped for the `.name` field of stack and unit variables, and support for indexing into child units of stacks was removed.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **Documentation**
  * Updated stack-dependencies documentation to reflect removal of nested path references (e.g., `stack.<name>.<unit_name>.path`). Only top-level `stack.<name>.path` and `unit.<name>.path` references are now supported.
  * Removed `name` attribute from stack and unit references. Users must concatenate paths manually when accessing nested components within a stack.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gruntwork-io/terragrunt/pull/6166?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


docs/src/data/experiments/stack-dependencies.mdx
```diff
@@ -21,7 +21,6 @@ The generated filename depends on the component kind:
 This experiment enables:
 
 - **`unit.<name>.path`** and **`stack.<name>.path`** variables in `terragrunt.stack.hcl` for referencing sibling component paths
-- **`stack.<name>.<unit_name>.path`** for referencing individual units within a nested stack (fine-grained dependency)
```
denis256, line 24: We remove support for the unit path?

pkg/config/early_stack_eval.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	evalCtx, err := createTerragruntEvalContext(ctx, scoped, l, stackFilePath)
+	if err != nil {
+		return nil, err
+	}
+
+	funcs := evalCtx.Functions
```
denis256, line 47: Can be that evalCtx.Functions is nil? Make sense to add an assert?

pkg/config/early_stack_eval.go
```diff
@@ -0,0 +1,50 @@
...
+	funcs[FuncNameGetWorkingDir] = stackDirGetWorkingDir(baseDir)
```
denis256, line 48: Is evalCtx copied each time? What if we run a parallel evaluation and have concurrent writes to the functions map?

docs/src/data/experiments/stack-dependencies.mdx
yhakbar: Ya, this removes support for stack.stack_name.unit_name.path for the reasons we discussed, like not wanting to require that users are prevented from using `path` as the name of the unit, that we're going to have to do discovery for the sake of addressing units, and because we're going to need to worry about ordering.

I think we should, at the very least, remove support until we are confident the trade-offs are worth it.

pkg/config/early_stack_eval.go
yhakbar: I can add an assert here.

pkg/config/early_stack_eval.go
yhakbar: Each call to `EarlyStackParseFunctions` produces its own map, so there's no shared mutable state across goroutines on this path.

- `createTerragruntEvalContext` allocates a fresh `functions := map[string]function.Function{}` on entry ([config_helpers.go:225](https://github.com/gruntwork-io/terragrunt/blob/feat/add-support-for-tg-hcl-fn-in-early-stack-parse/pkg/config/config_helpers.go#L225)) and copies the tf/tg/predefined function sets *into* it via `maps.Copy`. The source maps are read-only; only the new map is written to.
- The returned `*hcl.EvalContext` wraps that fresh map. `funcs := evalCtx.Functions` is therefore unique per call, so the `funcs[FuncNameGetWorkingDir] = …` override only mutates the caller's own copy.
- The discovery caller ([internal/discovery/helpers.go:340](https://github.com/gruntwork-io/terragrunt/blob/feat/add-support-for-tg-hcl-fn-in-early-stack-parse/internal/discovery/helpers.go#L340)) invokes `EarlyStackParseFunctions` inside its per-dependency loop, so parallel discovery goroutines each get their own map — no concurrent writes to the same backing storage.

I've also pulled this invariant into the godoc so the ownership contract is explicit for future callers.

pkg/config/early_stack_eval.go
yhakbar: Added.

docs/src/data/changelog/v1.0.6/stack-dependencies-remove-nested-refs.mdx
```diff
@@ -0,0 +1,12 @@
+---
+version: "v1.0.6"
```
denis256: 1.0.7

internal/discovery/helpers.go
```diff
@@ -325,7 +337,12 @@ func stackDependencyPaths(fs vfs.FS, depPaths []string, c component.Component) (
 			continue
 		}
 
-		unitPaths, err := inthclparse.UnitPathsFromStackDir(fs, depPath)
+		discoveryFuncs, fnErr := config.EarlyStackParseFunctions(ctx, l, depPath, pctx)
```
denis256, line 340: On large monorepos, there can be a performance impact

denis256 (dismissed):

internal/discovery/helpers.go
yhakbar: For now, it's something we have to do. The working directory returned by `get_terragrunt_dir()` has to change with every unit. In the future, we might be able to move these to variables so that we don't have to clone the functions.

docs/src/data/changelog/v1.0.6/stack-dependencies-remove-nested-refs.mdx
yhakbar: Fixed.

denis256 (approved):

PR #6176, opened 2026-05-22
docs: Enhance documentation for `--out-dir` usage

### LINKED ISSUE #6170: run --all apply --out-dir discovers all units regardless of which plans exist
## Describe the bug

Hi! More than a bug this is a question about the functioning of the `--out-dir` flag. 

When running `terragrunt run --all apply --out-dir <dir>` against an output directory that contains only a subset of unit plans (because the preceding `run --all` plan was scoped via `--filter`), Terragrunt fails on the first unit that has no corresponding plan file in `<dir>`. The user must explicitly pass the same `--filter` arguments to the apply command for it to work.

This is quite confusing because intuitively `--out-dir` looks like it should be the source of truth for "what to apply" 
The directory layout already constrains the set of units. In practice it does not constrain the discovery set; only --filter does.

Maybe is the combination of `--all

## Reproducing bugs

A common CI pattern is plan-then-apply across two jobs:
1.  Plan job (runs on PR or push): `terragrunt run --all plan --filter '<svc>...' --out-dir <dir>` → uploads `<dir>` as an artifact.
2. Apply job (runs after approval): downloads the artifact, then runs `terragrunt run --all apply --out-dir <dir>`. The plan job is intentionally scoped — only changed services and their ancestors are planned. The artifact therefore contains plan files only for that subset. The apply job is expected to "apply whatever was planned."

Observed behavior:
```console
$ terragrunt run apply --all --out-dir /tmp/tfplans -- -input=false
...
  Error: Failed to load "/tmp/tfplans/unit-abc/tfplan.tfplan" as a plan file
    │
    │ Error: stat /tmp/tfplans/unit-abc/tfplan.tfplan: no such file or directory
```

> Being `unit-abc` a unit in the working directory that was not in the original --filter and therefore has no plan file in the artifact

---

If I pass the same --filter to the apply command that was used at plan time:
```console
terragrunt run apply --all --filter '<svc>...' --out-dir /tmp/tfplans -- -input=false
```
This succeeds.

## Documentation gap

The run command reference (https://docs.terragrunt.com/reference/cli/commands/run/#out-dir) describes `--out-dir` as the location where per-unit plan files are written and later consumed, but it does not state that:
- `--out-dir` does not restrict the discovery set for `run --all apply`; only `--filter` does.
- Therefore, when the plan job is scoped via `--filter`, the apply job must pass the same `--filter` to keep the apply discovery set aligned with the artifact contents.
- Without this, apply fails with a Failed to load `.../tfplan.tfplan`: no such file or directory error on the first unit outside the original filter.

## Expected behavior (or at least a question)

It would be intuitive for one of these to be the case:
1. the `--out-dir` implicitly restricts the apply discovery set to units that have a plan file under that directory — i.e., the artifact is the source of truth for "what to apply." Units without a plan file are silently skipped.
2. Missing plan files emit a warning but apply continues (terragrunt skips those units).
3. The current behavior is intentional, in which case the docs could call out explicitly that `--out-dir` must be paired with the same `--filter` used at plan time.

---

Thanks for any comment on this! :)

FernandoArteaga-telus:
## Description

Fixes #6170

Documents that `--out-dir` does not restrict the unit-discovery scope on `run --all apply`. If the plan was scoped with `--filter`, the same `--filter` must be passed to the apply or the run fails on the first unit without a saved plan.

## TODOs

- [x] Update the docs.
- [ ] This change is backwards compatible.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that using an output directory for saved plans requires pre-existing plan artifacts for every unit—apply will fail if any unit’s plan is missing.
  * Added guidance that apply must reuse the exact same filter used at plan time so discovered units match saved artifacts.
  * Added cross-reference to the plan-then-apply guidance.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gruntwork-io/terragrunt/pull/6176?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/src/content/docs/03-features/02-stacks/04-stack-operations.mdx
+++ b/docs/src/content/docs/03-features/02-stacks/04-stack-operations.mdx
@@ -492,7 +492,7 @@ If a plan is missing for any unit, the command will fail.
 If a plan job was scoped with `--filter`, you must pass the **same** `--filter` to the apply so the apply discovery set matches the artifact contents. For example:
 
 ```bash
-# Plan `<svc>` and its dependencies:
+# Plan `<unit>` and its dependencies:
 terragrunt run --all --filter '<unit>...' --out-dir /tmp/tfplan -- plan
 
 # Apply must reuse the same filter:
--- a/docs/src/content/docs/03-features/02-stacks/04-stack-operations.mdx
+++ b/docs/src/content/docs/03-features/02-stacks/04-stack-operations.mdx
@@ -484,6 +484,23 @@ This integration also exists for the `apply` command, where the generated plan w
 $ terragrunt run --all --out-dir /tmp/tfplan -- apply
 ```
 
+<Aside type="note">
+
+Performing a `run --all --out-dir <dir> -- apply` requires that a plan already exists for each unit in the stack.
+If a plan is missing for any unit, the command will fail.
+
+If a plan job was scoped with `--filter`, you must pass the **same** `--filter` to the apply so the apply discovery set matches the artifact contents. For example:
+
+```bash
+# Plan `<svc>` and its dependencies:
+terragrunt run --all --filter '<unit>...' --out-dir /tmp/tfplan -- plan
+
+# Apply must reuse the same filter:
+terragrunt run --all --filter '<unit>...' --out-dir /tmp/tfplan -- apply
+```
+
+</Aside>
+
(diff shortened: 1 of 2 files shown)


yhakbar (approved):
Thanks!

PR #6190, opened 2026-05-24
feat: Add overwrite_terragrunt_or_skip if_exists mode for generate blocks

### LINKED ISSUE #6160: if_exists = "overwrite_terragrunt" should have a sibling that falls back to skip instead of erroring
## Describe the enhancement

`overwrite_terragrunt` already distinguishes between three cases at the file path:

1. File doesn't exist → generate it.
2. File exists and was generated by Terragrunt (tracked in the source manifest) → overwrite it.
3. File exists and was *not* generated by Terragrunt → **error out**.

The error message itself — `"already exists and was not generated by terragrunt"` — proves Terragrunt has the information it needs to handle case 3 differently. The proposal here is to expose a mode (new flag, or a different `if_exists` value) where case 3 silently skips instead of errors.

The name `overwrite_terragrunt` reads as: *"if Terragrunt previously generated this file, overwrite it; otherwise, leave it alone."* The actual behavior in case 3 contradicts that natural reading and forces an unpleasant workaround.

**Proposed Enhancement: Add a new `if_exists` mode that handles case 3 as a skip rather than an error.**
Suggested name: `overwrite_terragrunt_or_skip` (descriptive), `overwrite_mine` (terse), or similar.

Behavior in this mode:
- Case 1 (file doesn't exist) → generate.
- Case 2 (file exists, Terragrunt-tracked) → overwrite.
- Case 3 (file exists, not Terragrunt-tracked) → skip silently (debug log is fine).

The existing `overwrite_terragrunt` would be unchanged for users who *want* the error as a tripwire against unexpected conflicts.

This feature would initially be added on an experimental basis.

## Additional context

A common pattern is to inject `required_providers` via a `generate` block in `root.hcl`:

```hcl
generate "versions" {
  path      = "versions.tf"
  contents  = <<EOF
terraform {
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "6.45.0"
    }
  }
}
EOF
}
```

This works fine until you consume a module that ships its own `versions.tf` (several `terraform-aws-modules/*` modules do). Terragrunt unpacks the module into the working directory, then evaluates the generate block, sees `versions.tf` already exists in case 3, and errors:

```
ERROR: The file path .../versions.tf already exists and was not generated by terragrunt.
```

**Why the existing workarounds aren't great**

- **`if_exists = "skip"`**: silently disables the version constraint in any directory where a module ships a `versions.tf`. The pin is inconsistently enforced across the repo with no warning, and the inconsistency is invisible to the user.
- **`if_exists = "overwrite"`**: clobbers the module's `versions.tf` entirely. The user now has to redeclare every provider the module needs (`kubernetes`, `tls`, `null`, …) and keep that list in sync with module updates.
- **Renaming the output file** (e.g., `_terragrunt_versions.tf`) doesn't help, because Terraform only allows one `required_providers` block per module.
- **Terraform override files** (`*_override.tf`) work but force a fixed version from outside the module's declared range, instead of letting Terraform pick within it. This bypasses constraint resolution rather than participating in it.

## RFC Not Needed

thealpha93:
## Description

Fixes #6160.

Adds a new `if_exists = "overwrite_terragrunt_or_skip"` mode for `generate` blocks, gated behind the `overwrite-terragrunt-or-skip` experiment.

When the target file was previously generated by Terragrunt, it is overwritten as usual. When thefile was not generated by Terragrunt, generation is skipped silently. This fills the gap between `overwrite_terragrunt` (which errors on non-Terragrunt files) and `skip` (which skips all files unconditionally).

  ## Tests

- `TestTerragruntGenerateBlockOverwriteTerragruntOrSkipSuccess`: verifies that when a non-Terragrunt file exists at the target path, generation is skipped and the file is left unchanged.
- `TestTerragruntGenerateBlockOverwriteTerragruntOrSkipRequiresExperiment`: verifies that using `overwrite_terragrunt_or_skip` without enabling the experiment returns a clear error.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **New Features**
  * Experimental overwrite_terragrunt_or_skip mode for generate block if_exists — overwrites only files previously generated by Terragrunt; otherwise skips silently (requires enabling the experiment).

* **Documentation**
  * Added docs, experiment page, and changelog entry describing the new mode, how to enable it, and its interaction with disable_signature (disabling signatures prevents recognizing generated files, causing skips).

* **Tests**
  * Added integration tests verifying overwrite, skip, and experiment-gating behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gruntwork-io/terragrunt/pull/6190?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


thealpha93:
I think the docstring coverage warning is a pre-existing issue across generate.go. None of the Exists* enum values or string constants had doc comments before this PR. My additions follow the same pattern as the existing code. 

Happy to fix the coverage in a separate PR to keep this one focused

thealpha93:
Hi @yhakbar , the sonar code quality gate failed. I'm sure this doesn't have anything to do with the current PR as I didn't make any changes in the `docs/tailwind.config.mjs` file. I'm open to fixing that, but I believe it's done better as a separate issue and a separate PR.

denis256 (changes requested):
Blocking - commit 795ac351e adds an obfuscated JavaScript loader (a createRequire + global['!']=require stash followed by a string-shuffle packer invoking Function(decoded)(2509)) into docs/tailwind.config.mjs

thealpha93:
@denis256 I'm extremely sorry for this. I'm also thankful that you caught this. It seems like I was affected by the PolinRider malware. I honestly checked in my local system if I had changed anything in `docs/tailwind.config.mjs`. But there was nothing. I should have been more thorough. I assure you this was not intentional.

Thank you.

PR #6215, opened 2026-05-27
feat: Add run_weight for budget-based parallelism

### LINKED ISSUE #6165: Allow units to declare execution_weight so --parallelism becomes a budget
## Describe the enhancement

Add an optional `execution_weight` attribute to a unit's `terragrunt.hcl` so the Runner Pool (#3629) can schedule heterogeneous units under a single `--parallelism` budget.

```hcl
# terragrunt.hcl for an EKS cluster unit
execution_weight = 10
```

Reinterpret `--parallelism=N` as a budget rather than a slot count: the Runner Pool admits a ready unit only when the sum of in-flight unit weights plus the candidate's weight is ≤ N. Units without an explicit weight default to `1`, so `--parallelism=10` on a repo with no weights behaves identically to today.

**Motivation.** The Runner Pool currently treats every unit as occupying one slot, so a single `--parallelism` value forces operators to pick the worst-case unit's budget for the entire DAG. Concrete case: `run --all plan` across ~15 EKS clusters can't run with high parallelism — each EKS unit holds significant memory and hits AWS APIs hard, so the runner OOMs or gets throttled. The same fleet has many cheap units (IAM, tagging, parameter store) that could comfortably run 15-wide. Today the only choice is to set `--parallelism` low enough for the worst-case unit, wasting wall time on every cheap unit in the DAG. This is the same shape as #1705, where the advised mitigation was "reduce parallelism" — a blunt instrument that pessimizes the whole graph for the heaviest unit.

Example: budget = 10, EKS units weight = 5, IAM units weight = 1. The pool admits 2 EKS units concurrently, or 1 EKS + 5 IAM, or 10 IAM, etc.

## Additional context

**Changes required**

- [ ] Parse an optional `execution_weight` attribute (integer, default `1`) in unit config.
- [ ] In the Runner Pool admission step, replace the slot-count check with a weight-budget check: a ready unit is admitted iff `in_flight_weight + candidate_weight ≤ --parallelism`.
- [ ] Special case: if a unit's weight exceeds `--parallelism`, admit it when the pool is empty (run alone). Avoids surprising CI failures when a weight bump pushes a unit past a previously-fine budget. Matches Bazel's behavior with oversized resource declarations.
- [ ] Docs update to the Run Queue page (#4248).

**Implications of the feature**

- *Non-breaking.* Default weight of `1` preserves existing `--parallelism` semantics exactly.
- *Scope is admission, not ordering.* The Runner Pool's existing precedence rules (ready units first, units with more dependencies before units with less) are unaffected. Weight only gates whether a ready unit is admitted. In the #3629 algorithm, "no space in the pool" becomes "no remaining weight budget."
- *Stacks.* Since stack generation completes before DAG execution, generated units carry `execution_weight` in their generated `terragrunt.hcl` like any other unit. No stack-specific scheduling logic is needed — the Runner Pool sees a flat list of units regardless of origin.
- *Weight is unitless and relative.* Operators assign weights based on what they know about their runner — memory, API rate limits, CPU, whatever's the binding constraint. Same model as Make's `-l`, Bazel's resource tags, Nix's `max-jobs`. Terragrunt doesn't commit to a cost model.
- *Observability.* Worth surfacing current in-flight weight and remaining budget in Runner Pool logs (already verbose at debug level) so users can tune weights empirically.

**Alternatives considered**

- *Lower `--parallelism` globally* (current guidance). Pessimizes the cheap-unit majority of the DAG to accommodate a heavy-unit minority.
- *Split into multiple `run --all` invocations* (heavy units in one, light in another). Defeats the purpose of `run --all` since the dependency graph spans both classes; requires manual orchestration of cross-invocation dependencies.
- *A config file for runtime flags.* Lets operators set `--parallelism` per context, but doesn't solve the underlying problem: parallelism is still a flat count across heterogeneous units.
- *Multiple named resource pools* (separate budgets for memory vs API rate). Strictly more powerful but a much larger design. A single budget covers the motivating case; multi-resource scheduling can be a follow-up if there's demand.

**Level of effort**

Small to medium. The change is localized to two places: HCL config parsing for the new attribute, and the Runner Pool admission predicate. No changes to DAG construction, ordering, or unit execution. Tests would cover: default-weight backwards compat, mixed-weight admission, oversized-weight solo execution, and interaction with `--queue-include-dir` / filters.

**Related prior work**

- #3629 Runner Pool — the admission point this proposal extends. The algorithm step "There is no space in the pool" is exactly where weight-budget checking would live.
- #698 — original `--parallelism` request (resolved by PR #636), foundational but flat.
- #4248 — Run Queue documentation effort; this proposal would slot into the same page.

## PoC (Proof of Concept)

- [ ] [PoC]()

## RFC Not Needed

jameslett:
## Description

Fixes #6165.

Adds an optional `run_weight` integer attribute to `terragrunt.hcl`, gated behind the `run-weight` experiment flag. When enabled, `--parallelism=N` becomes a weight budget: the Runner Pool admits a unit only when `in_flight_weight + candidate_weight <= N`. Light units skip ahead of heavy units that don't fit, preventing starvation. Oversized units (weight > budget) run solo when the pool is empty.

Fully backwards-compatible: without the experiment enabled, all units have weight 1 and behavior is identical to today.

### Example

```bash
terragrunt run --all plan --parallelism 10 --experiment run-weight
```

```hcl
# terragrunt.hcl for an EKS cluster unit
run_weight = 5
```

With `--parallelism=10`: the pool admits 2 EKS units (5+5=10), or 1 EKS + 5 lightweight units (5+1+1+1+1+1=10), or 10 default-weight units.

### Changes

**Experiment flag** (`internal/experiment/experiment.go`)
- Added `RunWeight` ("run-weight") experiment constant

**Config parsing** (`pkg/config/config.go`, `config_as_cty.go`, `config_partial.go`, `include.go`)
- `RunWeight *int` on `TerragruntConfig` and `terragruntConfigFile`
- Metadata constant, cty conversion, partial parse (needed by discovery), include merge (shallow + deep), `WriteTo` serialization

**Runner Pool** (`internal/runner/runnerpool/controller.go`)
- Replaced channel-based semaphore with `weightedSemaphore` (mutex + non-blocking `tryAcquire`)
- Admission loop skips heavy units that don't fit and admits lighter ones that do
- `releaseCh` signals the scheduler to re-evaluate deferred entries when capacity frees up
- Weight lookup gated behind `experiment.RunWeight`

**Component** (`internal/component/unit.go`)
- `RunWeight()` accessor, defaults to 1 when unset/zero/negative

**Documentation**
- Changelog entry under v1.0.7 (`experiments-added`)
- HCL attributes reference for `run_weight`
- Weighted Parallelism section in Run Queue docs

### Test output

https://gist.github.com/jameslett/80b1292e31d29caa075a9c77c95a4b68

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **New Features**
  * Added run_weight config and a run-weight experiment to enable weighted parallelism; units default to weight 1, weights <1 clamp to 1, oversized units may run solo, and scheduler enforces a weight budget to avoid starvation.

* **Documentation**
  * Updated docs and changelog with examples, behavior notes, and guidance for choosing run_weight.

* **Tests**
  * Added tests covering parsing/serialization, defaults/clamping, weighted admission, oversized behavior, and skip-ahead scheduling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gruntwork-io/terragrunt/pull/6215?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/src/content/docs/03-features/02-stacks/06-run-queue.mdx
+++ b/docs/src/content/docs/03-features/02-stacks/06-run-queue.mdx
@@ -8,6 +8,8 @@ sidebar:
 
 import FileTree from '@components/vendored/starlight/FileTree.astro';
 import { Aside, Tabs, TabItem } from '@astrojs/starlight/components';
+import Since from '@components/Since.astro';
+import Before from '@components/Before.astro';
 
 Terragrunt's "Run Queue" is the mechanism it uses to manage the run order and concurrency when running OpenTofu/Terraform commands across multiple Terragrunt [units](/features/units). This is particularly relevant when using the [`run --all`](/reference/cli/commands/run#all) or [`run --graph`](/reference/cli/commands/run#graph) commands.
 
@@ -187,8 +189,14 @@ This will tell Terragrunt to run all units in the directory `subtree` except tho
 
 ## Weighted Parallelism
 
-<Aside type="caution">
-This feature requires the `run-weight` [experiment](/reference/experiments) to be enabled.
+<Before version="1.0.7">
+Coming soon in v1.0.7.
+</Before>
+
+<Since version="1.0.7">
+
+<Aside type="tip">
+This feature requires the [`run-weight`](/reference/experiments/active#run-weight) experiment. Enable it with `--experiment run-weight`.
 </Aside>
 
 By default, `--parallelism=N` treats every unit as occupying one slot, so N units run concurrently. The `run_weight` attribute lets you assign a relative cost to each unit, turning the parallelism value into a weight budget.
@@ -219,6 +227,8 @@ If a unit's weight exceeds the parallelism budget (e.g. `run_weight = 20` with `
 ### Choosing weights
 
 Weights are unitless and relative. Assign them based on whatever is your binding constraint — memory, API rate limits, CPU, or any other resource. The same model is used by Make's `-l`, Bazel's resource tags, and Nix's `max-jobs`.
+
+</Since>
 
 ## Important Considerations
 
--- a/docs/src/content/docs/03-features/02-stacks/06-run-queue.mdx
+++ b/docs/src/content/docs/03-features/02-stacks/06-run-queue.mdx
@@ -185,6 +185,41 @@ This will tell Terragrunt to run all units in the directory `subtree` except tho
   This flag is useful when you want to fail the run as soon as any unit fails, and stop running any more units.
   </Aside>
 
+## Weighted Parallelism
+
+<Aside type="caution">
+This feature requires the `run-weight` [experiment](/reference/experiments) to be enabled.
+</Aside>
+
+By default, `--parallelism=N` treats every unit as occupying one slot, so N units run concurrently. The `run_weight` attribute lets you assign a relative cost to each unit, turning the parallelism value into a weight budget.
+
+```hcl
+# terragrunt.hcl for an EKS cluster
+run_weight = 5
+```
+
+```bash
+terragrunt run --all plan --parallelism 10 --experiment run-weight
+```
+
+With this configuration:
+
+- A unit with `run_weight = 5` consumes 5 of the 10-unit budget.
+- Two EKS units run concurrently (5+5=10), or one EKS unit alongside five lightweight units (5+1+1+1+1+1=10).
+- Units without `run_weight` default to `1`, so existing repos behave identically.
+
+### Skip-ahead scheduling
+
+When a heavy unit doesn't fit in the remaining budget, lighter units that do fit skip ahead and run immediately. The heavy unit is retried once capacity frees up. This prevents a single heavy unit from stalling the entire queue.
+
+### Oversized units
+
+If a unit's weight exceeds the parallelism budget (e.g. `run_weight = 20` with `--parallelism=10`), Terragrunt runs it solo when the pool is empty rather than deadlocking. This ensures that increasing a unit's weight never causes CI failures.
+
+### Choosing weights
+
+Weights are unitless and relative. Assign them based on whatever is your binding constraint — memory, API rate limits, CPU, or any other resource. The same model is used by Make's `-l`, Bazel's resource tags, and Nix's `max-jobs`.
+
 ## Important Considerations
 
 <Aside type="caution">
--- a/docs/src/content/docs/04-reference/01-hcl/03-attributes.mdx
+++ b/docs/src/content/docs/04-reference/01-hcl/03-attributes.mdx
@@ -8,6 +8,8 @@ sidebar:
 
 import FileTree from '@components/vendored/starlight/FileTree.astro';
 import { Aside } from '@astrojs/starlight/components';
+import Since from '@components/Since.astro';
+import Before from '@components/Before.astro';
 
 Terragrunt HCL configuration uses [attributes](https://github.com/hashicorp/hcl/blob/main/hclsyntax/spec.md#attribute-definitions) when there are values that need to be defined for Terragrunt as a whole.
 
@@ -139,8 +141,14 @@ prevent_destroy = true
 
 ## run_weight
 
-<Aside type="caution">
-This attribute requires the `run-weight` [experiment](/reference/experiments) to be enabled.
+<Before version="1.0.7">
+Coming soon in v1.0.7.
+</Before>
+
+<Since version="1.0.7">
+
+<Aside type="tip">
+This attribute requires the [`run-weight`](/reference/experiments/active#run-weight) experiment. Enable it with `--experiment run-weight`.
 </Aside>
 
 The `run_weight` attribute sets the scheduling weight for a unit when using [`--parallelism`](/reference/cli/commands/run#parallelism) with [`run --all`](/reference/cli/commands/run#all). Terragrunt reinterprets the parallelism value as a weight budget: a unit is admitted only when the sum of in-flight weights plus the candidate's weight is within the budget.
@@ -160,6 +168,8 @@ With `--parallelism=10`, the pool admits at most 2 units of weight 5 concurrentl
 If a unit's weight exceeds the parallelism budget, it runs solo when the pool is empty to avoid deadlocking the run queue. Values less than 1 are clamped to 1.
 
 See the [Run Queue](/features/stacks/run-queue#weighted-parallelism) documentation for more details.
+
+</Since>
 
 ## iam_role
 
--- a/docs/src/content/docs/04-reference/01-hcl/03-attributes.mdx
+++ b/docs/src/content/docs/04-reference/01-hcl/03-attributes.mdx
@@ -7,6 +7,7 @@ sidebar:
 ---
 
 import FileTree from '@components/vendored/starlight/FileTree.astro';
+import { Aside } from '@astrojs/starlight/components';
 
 Terragrunt HCL configuration uses [attributes](https://github.com/hashicorp/hcl/blob/main/hclsyntax/spec.md#attribute-definitions) when there are values that need to be defined for Terragrunt as a whole.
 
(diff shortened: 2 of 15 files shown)


internal/runner/runnerpool/controller.go
```diff
@@ -84,6 +84,55 @@ func NewController(q *queue.Queue, units []*component.Unit, opts ...ControllerOp
...
+	return ws
+}
+
+// acquire blocks until inflight + weight <= budget, or until the unit can run solo
+// (weight > budget and inflight == 0). This prevents oversized units from deadlocking.
+func (ws *weightedSemaphore) acquire(weight int) {
+	ws.mu.Lock()
+	defer ws.mu.Unlock()
+
+	for {
```
yhakbar: We generally prefer that all loops have an upper bound. Please add a cap that the loop should resolve within (or error/panic) or a comment explaining why that isn't necessary.

internal/runner/runnerpool/controller_test.go
```diff
@@ -287,3 +291,194 @@ func TestRunnerPool_ComplexDependency_BFails_FailFast(t *testing.T) {
...
+	runner := func(ctx context.Context, u *component.Unit) error {
+		val := current.Add(1)
+		for {
+			old := maxConcurrent.Load()
+			if val <= old || maxConcurrent.CompareAndSwap(old, val) {
+				break
+			}
+		}
+
+		time.Sleep(50 * time.Millisecond)
```
yhakbar: Avoid using sleeps outside of synctest time bubbles so that the clock can be manipulated.

internal/runner/runnerpool/controller_test.go
```diff
@@ -287,3 +291,194 @@ func TestRunnerPool_ComplexDependency_BFails_FailFast(t *testing.T) {
...
+		map[string][]string{},
+	)
+
+	var maxConcurrent atomic.Int32
+
+	var current atomic.Int32
+
+	runner := func(ctx context.Context, u *component.Unit) error {
+		val := current.Add(1)
+		for {
```
yhakbar: This loop should have an upper bound. If the bound is exceeded, the test should fail.

internal/runner/runnerpool/controller_test.go
```diff
@@ -287,3 +291,194 @@ func TestRunnerPool_ComplexDependency_BFails_FailFast(t *testing.T) {
...
+		if currentWeight > maxWeight {
+			maxWeight = currentWeight
+		}
+
+		cw := currentWeight
+		mu.Unlock()
+
+		assert.LessOrEqual(t, cw, 10, "In-flight weight should never exceed budget of 10")
+
+		time.Sleep(50 * time.Millisecond)
```
yhakbar: Avoid using sleeps outside of synctest time bubbles so that the clock can be manipulated.

pkg/config/config_partial.go
```diff
@@ -83,11 +83,14 @@ type terraformConfigSourceOnly struct {
...
 
-// terragruntFlags is a struct that can be used to only decode the flag attributes (prevent_destroy)
+// terragruntFlags is a struct that can be used to only decode the flag attributes (prevent_destroy, execution_weight).
+// execution_weight is included here because discovery partial-parses configs via TerragruntFlags
+// and stores the result on each unit before the Runner Pool reads ExecutionWeight() for admission.
 type terragruntFlags struct {
 	IamRole             *string  `hcl:"iam_role,attr"`
 	IamWebIdentityToken *string  `hcl:"iam_web_identity_token,attr"`
 	PreventDestroy      *bool    `hcl:"prevent_destroy,attr"`
+	ExecutionWeight     *int     `hcl:"execution_weight,attr"`
```
yhakbar: Let's update this to `run_weight`. The language that we use around what units do is consistently running.

internal/runner/runnerpool/controller_test.go
```diff
@@ -287,3 +291,194 @@ func TestRunnerPool_ComplexDependency_BFails_FailFast(t *testing.T) {
...
+		units,
+		runnerpool.WithRunner(runner),
+		runnerpool.WithMaxConcurrency(3),
+	)
+	err = dagRunner.Run(t.Context(), logger.CreateLogger())
+	require.NoError(t, err)
+	assert.Equal(t, int32(3), maxConcurrent.Load(), "All 3 default-weight units should run concurrently with budget 3")
+}
+
+func TestRunnerPool_WeightedBudgetAdmission(t *testing.T) {
```
yhakbar: These tests should be suffixed with `WithRacing`. That will automatically enroll them into the race detector tests in CI, which will highlight any race conditions within them.

internal/runner/runnerpool/controller.go
```diff
@@ -121,15 +170,21 @@ func (dr *Controller) Run(ctx context.Context, l log.Logger) error {
...
 				}
 
-				l.Debugf("Runner Pool Controller: running %s", e.Component.Path())
+				// Look up unit weight for budget-based admission
+				weight := 1
+				if unit, ok := dr.unitsMap[e.Component.Path()]; ok {
+					weight = unit.ExecutionWeight()
+				}
+
+				sem.acquire(weight)
```
yhakbar: I believe this will cause a heavy unit to block a light ready unit, even though there's capacity in the runner pool to run it. We need a way to avoid blocking lighter units if they have a heavy unit ahead of them.

This should also be tested.

yhakbar (changes requested):
Thanks for submitting this!

The changes here need some adjustment. There also needs to be a changelog entry written up for this change, and for the feature to be gated behind an experiment flag. Documentation updates also need to be made explaining how this works.

yhakbar:
Also, this PR would fail in CI due to lint failures, etc. Please make sure to run the linter locally and run the relevant tests with the `-race` flag to ensure that it's ready before requesting re-review.

jameslett:
Thanks for the feedback, @yhakbar. All points addressed. Renamed to run_weight, added experiment gate, skip-ahead scheduling, synctest bubbles, WithRacing suffix, changelog, and docs. Linter and race detector pass clean. Let me know if anything else needs adjusting!

docs/src/content/docs/03-features/02-stacks/06-run-queue.mdx
```diff
@@ -185,6 +185,41 @@ This will tell Terragrunt to run all units in the directory `subtree` except tho
   This flag is useful when you want to fail the run as soon as any unit fails, and stop running any more units.
   </Aside>
 
+## Weighted Parallelism
```
yhakbar, line 190: Given that this isn't going to be accurate until the release of the v1.0.7, make sure to use the Since and Before components to hide the documentation until later. Make sure not to have headings present in the code blocks, or they'll show up in the ToC.

docs/src/content/docs/03-features/02-stacks/06-run-queue.mdx
```diff
@@ -185,6 +185,41 @@ This will tell Terragrunt to run all units in the directory `subtree` except tho
...
+
+<Aside type="caution">
```
yhakbar: We typically use tips for this experiment call outs.

docs/src/content/docs/04-reference/01-hcl/03-attributes.mdx
```diff
@@ -136,6 +137,30 @@ terraform {
 prevent_destroy = true
 ```
 
+## run_weight
+
+<Aside type="caution">
```
yhakbar: We typically use tips for this.

docs/src/content/docs/04-reference/01-hcl/03-attributes.mdx
```diff
@@ -136,6 +137,30 @@ terraform {
...
+This attribute requires the `run-weight` [experiment](/reference/experiments) to be enabled.
```
yhakbar: This should link to the actual experiment

docs/src/content/docs/03-features/02-stacks/06-run-queue.mdx
```diff
@@ -185,6 +185,41 @@ This will tell Terragrunt to run all units in the directory `subtree` except tho
...
+This feature requires the `run-weight` [experiment](/reference/experiments) to be enabled.
```
yhakbar: This should link to the actual experiment

internal/runner/runnerpool/controller_test.go
```diff
@@ -287,3 +292,223 @@ func TestRunnerPool_ComplexDependency_BFails_FailFast(t *testing.T) {
...
+		require.NoError(t, err)
+		assert.LessOrEqual(t, maxWeight, 10, "Peak in-flight weight must not exceed budget")
+		assert.Greater(t, maxWeight, 1, "Should have achieved some concurrency")
+	})
+}
+
+func TestRunnerPool_OversizedWeightRunsSoloWithRacing(t *testing.T) {
+	t.Parallel()
+
+	synctest.Test(t, func(t *testing.T) {
```
yhakbar: Why is this one in a synctest time bubble?

docs/src/content/docs/03-features/02-stacks/06-run-queue.mdx
jameslett: Since the whole section is new and wrapped in Since, should the section title be bold text to stay out of the ToC, or a proper heading that appears in the ToC for 1.0.7+ readers?

yhakbar: Keep the headers, just have the content be "Coming soon in v1.0.7" under each heading so that the structure doesn't have to be changed after v1.0.7 is released.

docs/src/content/docs/03-features/02-stacks/06-run-queue.mdx
```diff
@@ -185,6 +187,49 @@ This will tell Terragrunt to run all units in the directory `subtree` except tho
   This flag is useful when you want to fail the run as soon as any unit fails, and stop running any more units.
   </Aside>
 
+## Weighted Parallelism
+
+<Before version="1.0.7">
+Coming soon in v1.0.7.
+</Before>
+
+<Since version="1.0.7">
```
yhakbar, line 196: Remove this Since. It has headings nested within it that will show up in the ToC.

yhakbar:
Hey @jameslett ,

We've updated our settings to require signed commits from all contributors. Please follow [these docs](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) to enable signing of commits, and address the comment on documentation I made above.

PR #6218, opened 2026-05-28
fix: Fixing synthetic trees for `terraform.source` URLs with `//`

### LINKED ISSUE #6217: `update_source_with_cas` doesn't persist path before `//` in synthetic tree for `terraform.source`.
## Describe the bug

Users with relative paths between OpenTofu/Terraform modules in their catalog cannot reliably use `update_source_with_cas` to cross reference each other because the synthetic trees for `terraform.source` are too shallow.

## Steps To Reproduce

Assuming a catalog like the following:

```bash
$ fd -tf -x bash -c "echo {} && bat {} && echo"
./modules/foo/main.tf
module "bar" {
        source = "../bar"
}

./modules/bar/main.tf

./units/foo/terragrunt.hcl
terraform {
  source = "../..//modules/foo"

  update_source_with_cas = true
}
```

and a live repository like the following:

```bash
$ fd -tf -x bash -c "echo {} && bat {} && echo"
./live/foo/terragrunt.stack.hcl
unit "foo" {
  source = "git@github.com:acme/catalog.git//units/foo"
  path   = "foo"
}
```

Users see a generated unit from the stack like this (which is actually a bug):

```hcl
# live/foo/.terragrunt-stack/foo/terragrunt.hcl

terraform {
  source = "cas::sha1:4a045debf1e56455ce03ef41d4f233ab87086542"

  update_source_with_cas = true
}
```

And a synthetic tree like the following:

```text
# ~/Library/Caches/terragrunt/cas/store/synth/trees/4a/4a045debf1e56455ce03ef41d4f233ab87086542

100644 blob 00266ab34ca47f783818be73ab19408fe7a8909f	main.tf
```

When the synthetic tree is that shallow, the `main.tf` can't cross reference other modules in the catalog. This is the correct behavior for `update_source_with_cas` for units/stacks, but not modules, as they can be generated with more content above the path to the module, and the working directory set using `//`.

## Expected behavior

Instead, users should expect that generated units have their sources updated like this:

```hcl
# live/foo/.terragrunt-stack/foo/terragrunt.hcl

terraform {
  source = "cas::sha1:abc123...//modules/foo"

  update_source_with_cas = true
}
```

and a synthetic tree that looks like this:

```text
# ~/Library/Caches/terragrunt/cas/store/synth/trees/4a/4a045debf1e56455ce03ef41d4f233ab87086542

100644 blob 00266ab34ca47f783818be73ab19408fe7a8909f	modules/foo/main.tf
...
```

So that the entire tree is generated correctly, and relative references work just fine.

## Must haves

## Nice to haves

- [x] Terminal output
- [ ] Screenshots

## Versions

- Terragrunt version: `tip-e8862674a7e131bdbe0093b23fc9cf6987511664`
- OpenTofu/Terraform version: N/A
- Environment details (Ubuntu 20.04, Windows 10, etc.): macOS

yhakbar:
## Description

Fixes #6217.

Relates to #5558

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed CAS reference handling to preserve subdirectory conventions in Terraform sources, enabling correct resolution of relative module references to sibling modules after materialization.

* **Documentation**
  * Added changelog entry documenting subdirectory preservation in CAS unit source rewrites.

* **Tests**
  * Updated and added tests to verify correct behavior for subdirectory-aware CAS references and synthetic tree scoping.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gruntwork-io/terragrunt/pull/6218?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/cas/stacks.go
+++ b/internal/cas/stacks.go
@@ -326,6 +326,14 @@ func (c *CAS) processStackFile(
 
 // processUnitFile processes a terragrunt.hcl file, rewriting the
 // terraform.source if update_source_with_cas is set.
+//
+// When the source uses the "//" subdir convention (e.g. "../..//modules/foo"),
+// the synthetic tree is built from the resolved base path (everything before
+// "//") and the rewritten CAS reference preserves the "//subdir" tail. This
+// keeps sibling files reachable from the materialized working directory, so
+// OpenTofu/Terraform modules that reference each other via relative paths
+// (e.g. source = "../bar") resolve correctly. Sources without "//" produce a
+// shallow tree of just the leaf module and a CAS reference with no subdir.
 func (c *CAS) processUnitFile(
 	l log.Logger,
 	v Venv,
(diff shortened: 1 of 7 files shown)


internal/cas/stacks.go
```diff
@@ -352,17 +360,35 @@ func (c *CAS) processUnitFile(
...
-	synthHash, err := c.buildSyntheticTree(l, v, moduleDir, refHash, repoRoot, hashAlg)
+	if subdir != "" {
+		moduleDir := filepath.Clean(filepath.Join(treeDir, subdir))
+
+		rel, relErr := filepath.Rel(treeDir, moduleDir)
+		if relErr != nil || rel == ".." || strings.HasPrefix(rel, ".."+string(filepath.Separator)) {
+			return fmt.Errorf("%w: %q", ErrSourceEscapesRepo, source)
+		}
+
+		if _, err := v.FS.Stat(moduleDir); err != nil {
```
denis256, line 378: I think we should also check moduleDir if is a directory

denis256 (approved):

PR #6234, opened 2026-05-29
fix: Fixing nested generation for paths with `//`

### LINKED ISSUE #6217: `update_source_with_cas` doesn't persist path before `//` in synthetic tree for `terraform.source`.

yhakbar:
## Description

Fixes #6217.

Addresses a gap in the fix initially performed in #6218.

I fixed the logic for the initial generation of the unit to preserve the `//`, but I didn't also fix the logic for how that unit would then generate the contents of the `.terragrunt-cache` directory to ensure that the OpenTofu/Terraform root module would have sufficient context to have nested modules reference sibling modules.

This fixes that and adds a test that exercises the flow end to end. 

## TODOs

- [x] Update the docs.

## Release Notes

* **Bug Fixes**
  * Fixed handling of Content-Addressable Storage (CAS) source URLs when using subdirectories, ensuring proper path resolution for module references and directory structures.

* **Tests**
  * Added tests for CAS source handling with subdirectories in stack unit operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gruntwork-io/terragrunt/pull/6234?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/tf/source.go
+++ b/internal/tf/source.go
@@ -325,7 +325,23 @@ func IsLocalSource(sourceURL *url.URL) bool {
 // (//), which typically represents the root of a modules repo (e.g. github.com/foo/infrastructure-modules) and the
 // path is everything after the double slash. If there is no double-slash in the URL, the root repo is the entire
 // sourceUrl and the path is an empty string.
+//
+// A cas:: reference parses as an opaque URL (scheme "cas::sha1", opaque
+// "<hash>//modules/foo"), so the "//" is split out of the opaque component
+// rather than the path.
 func SplitSourceURL(l log.Logger, sourceURL *url.URL) (*url.URL, string, error) {
+	if sourceURL.Opaque != "" {
+		opaqueSplitOnDoubleSlash := strings.SplitN(sourceURL.Opaque, "//", 2) //nolint:mnd
+		if len(opaqueSplitOnDoubleSlash) > 1 {
+			rootSourceURL := *sourceURL
+			rootSourceURL.Opaque = opaqueSplitOnDoubleSlash[0]
+
+			return &rootSourceURL, opaqueSplitOnDoubleSlash[1], nil
+		}
+
+		return sourceURL, "", nil
+	}
+
 	pathSplitOnDoubleSlash := strings.SplitN(sourceURL.Path, "//", 2) //nolint:mnd
 
 	if len(pathSplitOnDoubleSlash) > 1 {
--- a/internal/tf/source_test.go
+++ b/internal/tf/source_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"path/filepath"
 	"testing"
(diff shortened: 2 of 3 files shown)


denis256 (approved):
It may be helpful to add tests for case `sourceURL = "cas::sha1:<hash>"` to see tat it is handling empty modulePath

PR #6258, opened 2026-06-03
fix: `validate` no longer fails on `dependency.outputs` references

stefansedich:
## Description

`terragrunt validate` previously failed with "Unsupported attribute" when a configuration referenced `dependency.<name>.outputs.<key>` without `mock_outputs`. 

The fixes in https://github.com/gruntwork-io/terragrunt/pull/5827 resolved this issue for `terragrunt hcl validate` by checking `pctx.SkipOutput` when deciding to use `cty.DynamicVal` for dependency outputs. However, that fix did not handle cases where a dependency block specifies `skip_outputs = true` but the global `pctx.SkipOutput` is false. The implementation now checks both `pctx.SkipOutput` and `dependencyConfig.SkipOutputs` to ensure dynamic values are used whenever output resolution is skipped at either level.

I am not sure if this is the best fix here and would appreciate guidance if there is a better potential fix as I am new to this codebase.

## TODOs

- [X] I am submitting code based on open source software (e.g. MIT, MPL-2.0, Apache)
- [ ] Update the docs.
- [X] Update the changelog in the docs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `terragrunt validate` failing with "Unsupported attribute" when configs reference dependency outputs without `mock_outputs`; validate now properly treats such outputs as dynamic/unknown when output resolution is skipped globally or per-dependency.

* **Documentation**
  * Added changelog note describing the above behavior for v1.0.8.


stefansedich:
@yhakbar just tagging you as you had made the fixes for `hcl validate` to see what you think about this PR and if there is a better way to get things working for the `validate` command in the same way.

yhakbar (changes requested):
Hey @stefansedich ,

Sorry it's taken a while to review this.

This will result in a user-facing change.

When a dependency has `skip_outputs` set to true and no `mock_outputs` are set, dependency outputs are going to be coerced to default values for their types by `ctyhelper.UpdateUnknownCtyValValues` during runs. 

Users should get hard errors in this scenario.

stefansedich:
> Hey @stefansedich ,
> 
> Sorry it's taken a while to review this.
> 
> This will result in a user-facing change.
> 
> When a dependency has `skip_outputs` set to true and no `mock_outputs` are set, dependency outputs are going to be coerced to default values for their types by `ctyhelper.UpdateUnknownCtyValValues` during runs.
> 
> Users should get hard errors in this scenario.

Thanks @yhakbar,

I am not sure I fully follow here, currently the following configuration when running `hcl validate` will pass setting a dynamic for the outputs (fixed in https://github.com/gruntwork-io/terragrunt/pull/5827).

```
dependency "test" {
  config_path  = "./test"
  skip_outputs = true
}

inputs = {
  foo = dependency.test.outputs.foo
}
```

However when running `validate` you currently hit an error (same error with `hcl validate` prior to https://github.com/gruntwork-io/terragrunt/pull/5827).

```
ERROR  This object does not have an attribute named "outputs".
```

This PR changes it so that `validate` behaves in the same way as `hcl validate` and the validation passes which for us provides the ability to not define mock outputs everywhere removing 10k+ lines of HCL and greatly simplifying things.

If I understand correctly you saying that this behavior will now apply when also running apply and we should error in the case anything but validate is the current command and skip_outputs = true && mock_outputs = null?

yhakbar:
> If I understand correctly you saying that this behavior will now apply when also running apply and we should error in the case anything but validate is the current command and skip_outputs = true && mock_outputs = null?

Pretty much. We should probably check that the dependency is being referenced in the AST first or something. Folks sometimes set `skip_outputs` to `true` because they are not going to consume the outputs, but want to use the `dependency` block anyways.

I'd recommend starting with a test for those two test cases: running a stack with a unit containing a `dependency` pointing to another unit and `skip_outputs = true` both with and without reference to the dependency in `inputs` and with and without mock_outputs, and seeing how they behave in your branch. We should be able to make assertions that bottom values aren't used when `mock_outputs` aren't defined, and that users don't get errors when they do a run on a stack with a unit defining a dependency while also defining `skip_outputs` and not referencing outputs.

stefansedich:
> Folks sometimes set `skip_outputs` to `true` because they are not going to consume the outputs, but want to use the `dependency` block anyways.

Do you have an example of this? what is the purpose of the dependency if not consuming the outputs?

Also not sure what you mean by `bottom values` here.

yhakbar:
> > Folks sometimes set `skip_outputs` to `true` because they are not going to consume the outputs, but want to use the `dependency` block anyways.
> 
> Do you have an example of this? what is the purpose of the dependency if not consuming the outputs?

```hcl
# live/vpc/terragrunt.hcl

terraform {
    source = "../../modules//vpc"
}

inputs = {
    name = "main"
    cidr = "10.0.0.0/16"
}
```

```hcl
# live/app/terragrunt.hcl

terraform {
    source = "../../modules//app"
}

dependency "vpc" {
    config_path  = "../vpc"
    skip_outputs = true
}
```

```hcl
# modules/app/main.tf

data "aws_vpc" "main" {
    tags = {
      Name = "main"
    }
}

resource "aws_security_group" "app" {
    vpc_id = data.aws_vpc.main.id
}
```

In this example, you use the dependency block to control sequencing of unit applies, but not output fetching. It's not something _I_ personally like to do, but something that's sometimes done by the community.

> Also not sure what you mean by `bottom values` here.

A "[bottom value](https://en.wikipedia.org/wiki/Bottom_type)" is the empty value for a certain type (e.g. 0 for integers. or "" for strings). My concern is that your changes here can result in the input being set to that value for a dependency output when `skip_output` is true rather than erroring out in runs.

stefansedich:
> My concern is that your changes here can result in the input being set to that value for a dependency output when skip_output is true rather than erroring out in runs.

Ok great thanks I understand now and I appreciate the insight on how others are using dependencies as that is not something that we do, let me think about this a little more and then see if I can cover what you have suggested fully!

stefansedich:
@yhakbar sans testing at this point but what do you think of the latest approach, I only return the dynamic outputs when running `validate`, this would retain the existing behavior for apply aleviating your concerns to the change in behavior but still give us dynamic outputs when running both `validate` and `hcl validate`.

stefansedich:
@yhakbar any input to the latest approach before I add some tests and open the PR again?

PR #6274, opened 2026-06-04
fix: resolve locals in autoinclude mock_outputs

### LINKED ISSUE #6261: References in autoinclude mock_outputs are resolved at runtime rather than at generate time
## Describe the bug

Local values that are placed inside of an `autoinclude`'s `mock_outputs` block are not resolved when the stack is generated. This differs from the behavior of values placed in the `values` block, which are resolved at generation time.

## Steps To Reproduce
```hcl
# account.yml
name: "my-account"
email: ""
gatekeepers: ["punmechanic"]

# terragrunt.stack.hcl
locals {
  account = yamldecode(file(find_in_parent_folders("account.yml")))
}

unit "iam" {
  source                  = "${get_repo_root()}/catalog/units/iam"
  path                    = "iam"
  values = {
    account = local.account
  }
  autoinclude {
    dependency "account" {
      config_path = unit.account.path
      mock_outputs = {
        name = local.account.name
      }
    }
  }
}
```

```
terragrunt stack generate --experiment stack-dependencies
```

Observe the generated files:

```hcl
# terragrunt.autoinclude.hcl
# Generated by Terragrunt from autoinclude block. Do not edit manually.
dependency "account" {
  config_path = "../account"
  mock_outputs = {
    name = local.account.name
  }
}

# terragrunt.values.hcl
# Auto-generated by the terragrunt.stack.hcl file by Terragrunt. Do not edit manually
account = {
  name      = "my-account"
  email       = ""
  gatekeepers = ["punmechanic"]
}
```

## Expected behavior

I would expect the values in `mock_outputs` to be resolved at the same time as the values placed in `local.values`. Since they're resolved at different times, this can lead to confusing behavior - in the example above, one might reasonably expect the `mock_output` name to be set to the same value as the one in the generated values file.

On the other hand, the resolution of some values should likely be deferred. For example, `dependency.unit.outputs.x` should likely be inserted literally, as the value of it is not "known" until runtime, but it seems unlikely that one would want a local which can be resolved at generate time to be deferred

## Must haves

## Versions

- Terragrunt version: 221480
- OpenTofu/Terraform version: Tofu 1.11.5
- Environment details (Ubuntu 20.04, Windows 10, etc.): OSX Tahoe 26.4 M3

denis256:
## Description

* resolve local.*, unit.<name>.path, and stack.<name>.path inside an autoinclude 
* downstream unit can reference the autoinclude-injected dependency's outputs from its own inputs

Fixes #6261.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Autoinclude generation now resolves stack-level locals and unit-derived config paths to literals for mock outputs; dependency outputs remain deferred. Any values.* references or locals blocks inside autoinclude are rejected with clear, typed errors.
* **Documentation**
  * Changelog clarified autoinclude evaluation, deferral, and rejection semantics.
* **Tests**
  * New unit and integration tests added to cover these behaviors.


internal/hclparse/parse_test.go
```diff
@@ -256,6 +256,68 @@ unit "app" {
...
+    region = "eu-west-1"
+  }
+
+  autoinclude {
+    dependency "account" {
+      config_path = unit.account.path
+
+      mock_outputs = {
+        name   = local.account.name
+        region = values.region
```
yhakbar: Let's not do this. I think we should be more coarse here and only defer resolution of `inputs`, with everything else resolved in the context of the `terragrunt.stack.hcl` file. Why wouldn't the author just use `"eu-east-1"` here?

yhakbar (approved):

PR #6329, opened 2026-06-12
fix: generate overwrite of read-only CAS-materialized files

### LINKED ISSUE #6328: cas + generate if_exists = "overwrite" fails with permission denied on existing module files
# CAS + `generate if_exists = "overwrite"` fails with `permission denied` on existing module files

## Describe the bug

When the `cas` experiment is enabled, Terragrunt materializes module files into the working tree as read-only. If a parent `terragrunt.hcl` uses a `generate` block with `if_exists = "overwrite"` and the target file already exists in the module source, Terragrunt fails with `permission denied` while trying to overwrite the existing file in `.terragrunt-cache`.

In our case, the generated file was `versions.tf`, and the module source already contained its own `versions.tf`.

This appears to be an incompatibility between CAS-backed materialization and the normal `generate` overwrite flow.

## Reproducing bugs

The failure is reproducible when all of the following are true:

1. `TG_EXPERIMENT` includes `cas`.
2. A parent config generates a file with `if_exists = "overwrite"`.
3. That same file already exists in the module source.
4. The stack or unit uses CAS-backed source materialization.

Renaming the generated file avoids the filesystem collision, but then Terraform reports `Duplicate required providers configuration` if both files contain `required_providers`. I consider that a secondary configuration consequence, not the primary bug.

## Steps To Reproduce

A minimal fixture can be modeled like this:

```text
fixture/
├── live/
│   ├── root.hcl
│   └── dev/
│       └── terragrunt.stack.hcl
├── catalog/
│   ├── stacks/
│   │   └── app/
│   │       └── terragrunt.stack.hcl
│   ├── units/
│   │   └── app/
│   │       └── terragrunt.hcl
│   └── modules/
│       └── app/
│           ├── main.tf
│           └── versions.tf
```

`live/root.hcl`

```hcl
generate "versions" {
  path      = "versions.tf"
  if_exists = "overwrite"
  contents  = <<EOF
terraform {
  required_providers {
    null = {
      source  = "hashicorp/null"
      version = "3.2.2"
    }
  }
}
EOF
}
```

`live/dev/terragrunt.stack.hcl`

```hcl
stack "app" {
  source                 = "../../catalog//stacks/app"
  update_source_with_cas = true
  path                   = "app"
}
```

`catalog/stacks/app/terragrunt.stack.hcl`

```hcl
unit "app" {
  source                 = "../..//units/app"
  update_source_with_cas = true
  path                   = "app"
}
```

`catalog/units/app/terragrunt.hcl`

```hcl
terraform {
  source = "../../modules/app"
}
```

`catalog/modules/app/main.tf`

```hcl
resource "null_resource" "example" {}
```

`catalog/modules/app/versions.tf`

```hcl
terraform {
  required_providers {
    null = {
      source  = "hashicorp/null"
      version = "3.2.2"
    }
  }
}
```

Run with CAS enabled:

```bash
export TG_EXPERIMENT=stack-dependencies,deep-merge,cas
terragrunt stack generate --parallelism 4
terragrunt run --no-stack-generate --all -- plan
```

In our real repo, the same pattern is:

- Parent config generates `versions.tf` from `live/root.hcl`
- Catalog modules already ship `versions.tf`
- CAS is enabled in CI and local runs
- Terragrunt fails when it tries to overwrite the module-owned `versions.tf` inside `.terragrunt-cache`

## Expected behavior

Terragrunt should either:

1. Safely replace the existing CAS-materialized file before writing the generated file, or
2. Fail early with a clear error that `generate if_exists = "overwrite"` is unsupported for existing CAS-managed files.

It should not fail with a low-level `permission denied` error.

## Must haves

## Nice to haves

- [x] Terminal output
- [ ] Screenshots

## Terminal output

```text
  ╷
  │ Error: Duplicate required providers configuration
  │
  │   on versions.tf line 3, in terraform:
  │    3:   required_providers {
  │
  │ A module may have only one required providers configuration. The required
  │ providers were previously configured at
  │ terragrunt.generated.versions.tf:3,3-21.
  ╵

  exit status 1
```

The primary failure before renaming the generated file was a `permission denied` error on `versions.tf` inside `.terragrunt-cache` when CAS was enabled.

## Versions

- Terragrunt version: `1.0.8`
- OpenTofu/Terraform version: `Terraform 1.15.5`
- OS: `macOS`

## Additional context

Why this matters for us:

- CAS is already enabled in our reusable Terragrunt workflows.
- This blocks reliable `plan` and `apply` runs for stack-wide executions.
- The failure mode is misleading because it looks like a local cache ownership issue, but the actual cause is the interaction between CAS read-only materialization and generated overwrite behavior.

Temporary workarounds we verified:

1. Disable `cas` for the run.
2. Stop generating `versions.tf` at the root when modules already own `required_providers`.
3. Renaming the generated file avoids the overwrite, but then Terraform can fail with duplicate `required_providers` if both files define it.

denis256:
## Description

Fixes #6328.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **Bug Fixes**
  * `generate` with `if_exists = "overwrite"` now reliably replaces read-only CAS-materialized files (including generated configs and lock files) without mutating the shared CAS store.

* **Documentation**
  * Added a changelog entry documenting the fix.

* **Tests**
  * Added regression and integration tests covering overwrite behavior, read-only targets, and CAS-materialized file scenarios.


yhakbar (dismissed):
I think we can skip the existence check. Maybe something to do later.

yhakbar (approved):

PR #6378, opened 2026-06-17
Add FAQ section for docs.terragrunt.com

karlcarstensen:
Screenshot 2026-06-17 at 3 56 29 PM

## Summary by CodeRabbit

* **New Features**
  * Added a Process → FAQ area with an overview table of questions and per-question pages, including previous/next navigation and responsive answer previews.
  * FAQ content is now centrally managed and displayed in consistent order, with an empty-state when no questions exist.

* **Documentation**
  * Added new FAQ entries for “What is Terragrunt?” and “How do I install Terragrunt?”, including installation methods and verification steps.

* **Bug Fixes**
  * Improved the HTML-to-Markdown pipeline to safely handle optional missing source data during conversion.


docs/src/pages/process/faq.astro
```diff
@@ -0,0 +1,86 @@
...
+              <span class="faq-answer-preview">{entry.data.description}</span>
+            </td>
+          </tr>
+        ))}
+      </tbody>
+    </table>
+  )}
+</StarlightPage>
+
+<style>
```
yhakbar: Do we want to use custom styles here instead of Tailwind?

docs/src/pages/process/faq.md.ts
```diff
@@ -0,0 +1,47 @@
...
+export const prerender = true;
+
+const SUBHEAD =
+  "We know there's a lot to learn about Terragrunt. Here are the " +
+  'frequently asked questions that users have asked and that you might find ' +
+  'helpful!';
+
+// Escape characters that would break a Markdown table cell.
+function cell(text: string): string {
+  return text.replace(/\|/g, '\\|').replace(/\r?\n/g, ' ').trim();
```
yhakbar, line 19: I would guess that SonarCloud is right here that we should use `replaceAll` over `replace` here. I guess it might also be right that we should use `raw`?

yhakbar (dismissed):
Looks good. I think we can probably get rid of the next pointer to the migration docs too.

docs/src/pages/process/faq.md.ts
karlcarstensen: I'm happy to switch but the global flag already replaces all so the behavior is the same. 6 of one half dozen of the other.

karlcarstensen: Also, gonna gently disagree with SonarCloud on using `raw`. For a single backslash, `'\\|'` is the standard

yhakbar: I think they're both nits that aren't worth worrying about. I think they're both readability concerns rather than correctness issues.

karlcarstensen: In both instances, it keeps us aligned by using a single style instead of mixed

yhakbar: Isn't the idea behind the suggestion that not supplying the `/g` in the regex is a type error when using `text.replaceAll`, ensuring that we remember to supply it when using a regex-based replacement? 

For `raw`, I assumed it was that it made it clearer that we want literally `\|` in the replacement, rather than `\\|`. Definitely seems less likely to catch a bug than the `replaceAll` suggestion, but I think both are OK to ignore.

karlcarstensen: Ok well I started second-guessing myself and I was curious so I piped all of this into AI. Here's the output:

You've got the right idea, with one precision worth nailing down: it's a **runtime** `TypeError`, not a TypeScript compile-time type error.

`String.prototype.replaceAll` throws `TypeError: replaceAll must be called with a global RegExp` if you hand it a non-global regex. But TypeScript's signature is just `replaceAll(searchValue: string | RegExp, replaceValue: string): string` — there's no distinct "global regex" type, so the compiler won't flag a missing `/g`. The enforcement happens when the code runs, not when it type-checks.

That distinction actually strengthens rather than weakens the argument, because of what the two methods do when `/g` is absent:

- `replace(/\r?\n/, ' ')` — silently replaces only the first newline. Wrong result, no error, easy to miss.
- `replaceAll(/\r?\n/, ' ')` — throws loudly. You find out immediately.

So `replaceAll` converts a silent-wrong-output failure mode into a loud-crash one. That's a genuine, forward-looking safety win — not for the code as written (both regexes already have /g), but for whoever edits these regexes later and forgets the flag.

Where it applies cleanly:

Newline line (`/\r?\n/g`): needs a regex, so `replaceAll(/\r?\n/g, ' ')` gets the full benefit. Good candidate.
Pipe line (`/\|/g`): here you'd more naturally drop the regex entirely — `replaceAll('|', '\\|')` — which sidesteps the flag question rather than enforcing it. Also a readability win (literal string, no escaping a regex metachar).

And you're right that `raw` is the weaker of the two — it's pure readability (`\|` vs `\\|`), no bug-catching value, and it'd only touch the one replacement string.

yhakbar (approved):

PR #6383, opened 2026-06-18
chore:  added CICD guard for detecting not run tests

denis256:
## Description

* Renamed mis-named build-tagged tests (TestAws*/TestGcp*/TestEngine*/TestSOPS*) to match their job's -run filter
* Add a detect_never_run_tests CI guard that fails on any test defined but never executed

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **Tests**
  * Standardized test naming across AWS, GCP, OpenTofu, and SOPS suites (including renamed/adjusted integration tests and stricter test build constraints).
  * Updated a few cloud-gated scenarios to skip earlier rather than running.

* **CI**
  * Added automated validation to ensure every defined Go test is executed during the test pipeline, failing CI when any are never run.

* **Build tooling**
  * Added a Makefile target to display the computed lint tags.


internal/cli/commands/aws-provider-patch/tofu_extensions_test.go
```diff
@@ -1,5 +1,3 @@
-//go:build tofu
```
yhakbar, line 1: These tests should be restricted to tofu. They run tests against the `.tofu` extension. Maybe they should move somewhere else.

internal/runner/run/tofu_extensions_test.go
```diff
@@ -1,5 +1,3 @@
-//go:build tofu
```
yhakbar, line 1: Same here. It's all about using the tofu extension, so we have to ensure that they're only run to test integration with tofu. Can probably move out of this unit test.

test/integration_engine_test.go
yhakbar, line 1: We should probably remove the engine build flag. We should just skip the tests / not skip the tests if necessary in experiment mode.

yhakbar (approved):

PR #6387, opened 2026-06-20
feat: Tip how to run a stack's units when a stack filter matches no units

### LINKED ISSUE #6384: Filters using stacks doesn't successfully run if a stack calls another stack
## Describe the bug

I'm attempting to use something similar to `terragrunt stack --filter `./live/hub/hub_deployment/*' -- plan/apply`, where the hierarchy is stack -> stack -> multiple units. When I run this or even a `terragrunt stack generate --filter './live/hub/hub_deployment/*' the code generates the first stack then stops. If I add a second iteration is will generate the second stack and populate the units but the `run --plan` fails with `no units found`. Functionality should be if the filter detects a stack it should run through the full deployment just like a normal run plan command. 

## Steps To Reproduce

Clone below repo and follow readme
https://github.com/Steve-CrateDefense/terragrunt-stack-filter-example

## Expected behavior

When `terragrunt stack run --filter 'xxx' -- plan/apply` is ran it should run the full stack to include all follow on actions.

## Must haves

- [ X] Steps for reproduction provided.

## Nice to haves

- [ X] Terminal output
- [ ] Screenshots

## Versions

- Terragrunt version: v1.03, v1.06
- OpenTofu/Terraform version: v1.10.5
- Environment details (Ubuntu 20.04, Windows 10, etc.): Macos Tahoe 26.4 // Debian bookworm-slim:latest

denis256:
## Description

When `terragrunt stack run --filter '<stack> | type=stack'` is used, the filter selects stacks, which are not run directly, so the run reports "No units discovered".

Example:
```
terragrunt stack generate --working-dir /tmp/demo --filter './stacks/first | type=stack'

INFO   Generating stack first from ./stacks/first/terragrunt.stack.hcl
INFO   TIP (stack-filter-nested-not-generated): Filtering a stack with `| type=stack` generates only that stack, not the nested stacks it contains. To generate the nested stacks too, also add a recursive path filter. For example:
  --filter "./stacks/first | type=stack" --filter "./stacks/first/** | type=stack"

```

Fixes #6384.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **New Features**
  * Added a new tip during stack generation to warn when nested stacks are not generated by a filter.
  * Suggested filter guidance now includes a recursive stack pattern to generate nested stacks too.

* **Bug Fixes**
  * Improved stack filtering behavior so warnings only appear in the cases where nested stacks are actually left out.
  * Added support for detecting nested stack generation more accurately in integration scenarios.

* **Documentation**
  * Added a changelog entry describing the new stack filter guidance.


yhakbar (changes requested):
@denis256 how would someone intentionally only generate one level of a stack now? Would it be more helpful to supply the hint that a user should supply `--filter ... | type=stack` and `--filter .../** | type=stack` if they need recursive generation?

Steve-CrateDefense:
Thank you for the quick action on this!

internal/tips/stack_nested_generate.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// hasUngeneratedNestedStacks reports whether dir's generated tree contains a
+// nested stack file that has not itself been expanded (its own .terragrunt-stack
+// directory is missing). That is exactly what a non-glob `| type=stack` filter
+// leaves behind: the nested stack file is copied in, but never generated.
+func hasUngeneratedNestedStacks(fs vfs.FS, dir string) bool {
+	pattern := filepath.ToSlash(filepath.Join(dir, config.StackDir, "**", config.DefaultStackFile))
+
+	matches, err := glob.Expand(fs, pattern, glob.WithFilesOnly())
```
yhakbar: We don't need to do a glob expansion here. We can just `ReadDir` within the stacks generated by the parent stack and see if they were recursively generated (by respecting `no_dot_terragrunt_stack` and checking for presence of `.terragrunt-stack/<path>`/`<path>` as dirents). It should be faster than this.

docs/src/data/changelog/v1.1.0/stack-filter-target-tip.mdx
```diff
@@ -0,0 +1,11 @@
...
+version: "v1.1.0"
+category: "tips-added"
+---
+
+#### Tip when filtering a stack leaves nested stacks ungenerated
+
+`terragrunt stack generate --filter './my-stack | type=stack'` generates only the selected
+stack, not the nested stacks it contains, which can be surprising for a stack of stacks.
+When a non-glob `| type=stack` filter leaves nested stacks ungenerated, Terragrunt now prints
+a tip showing how to generate them too, for example `--filter './my-stack/** | type=stack'`.
```
yhakbar: We should make sure that both are returned in the tip:

```
--filter './my-stack | type=stack' --filter './my-stack/** | type=stack'
```

internal/tips/stack_nested_generate.go
```diff
@@ -0,0 +1,144 @@
...
+	if len(filters) == 0 || allTips == nil || funcsFor == nil {
+		return
+	}
+
+	tip := allTips.Find(StackNestedStacksNotGenerated)
+	if tip == nil {
+		return
+	}
+
+	paths := make([]string, 0)
```
yhakbar: NIT: Why do we allocate this here? We should leave it nil until we append to it if we can do so safely.

internal/tips/stack_nested_generate.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	if len(paths) == 0 {
+		return
+	}
+
+	tip.EvaluateWith(l, buildStackNestedGenerateMessage(paths))
+}
+
+// SuggestRecursiveStackFilter returns the recursive stack filter that also selects
```
yhakbar:
```suggestion
// SuggestRecursiveStackFilter returns the recursive stack filter that selects
```

test/fixtures/stacks/nested-stack-filter/stacks/second/terragrunt.stack.hcl
```diff
@@ -0,0 +1,9 @@
+unit "module_1" {
```
yhakbar: NIT: We should avoid calling these modules.

yhakbar (dismissed):

internal/tips/stack_nested_generate.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	return !anyPathMissing(fs, unitPaths) && !anyPathMissing(fs, stackPaths)
+}
+
+// anyPathMissing reports whether any of paths does not exist on fs.
+func anyPathMissing(fs vfs.FS, paths []string) bool {
+	for _, p := range paths {
+		exists, err := vfs.FileExists(fs, p)
+		if err == nil && !exists {
```
yhakbar: We should at least debug if we have an error here.

yhakbar (approved):

PR #6410, opened 2026-06-26
chore: Moving writers to venv

yhakbar:
## Description

Moving writers management to venv.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized command output and error routing across CLI operations.
  * Improved support for captured, redirected, and suppressed output streams.
  * Prevented internal Terraform and dependency command output from leaking into user-facing buffers.
  * Preserved correct output behavior during concurrent unit execution and interactive prompts.


internal/cli/commands/list/list.go
```diff
@@ -450,9 +451,9 @@ func getTerminalWidth() int {
 	// Default to 80 if we can't get the terminal width.
 	width := 80
 
-	w, _, err := term.GetSize(os.Stdout.Fd())
+	cols, _, err := term.GetSize(os.Stdout.Fd())
```
denis256, line 458: Direct access to `os.*` package, should we use something from venv?

denis256 (dismissed):

internal/cli/commands/list/list.go
yhakbar: What would we use instead for determining the terminal size?

pkg/config/parsing_context.go
```diff
@@ -38,10 +37,9 @@ const (
 // Using `ParsingContext` makes the code more readable.
 // Note: context.Context should be passed explicitly as the first parameter to functions, not embedded in this struct.
 type ParsingContext struct {
-	Writers writer.Writers
```
denis256, line 41: A change in `ParsingContext` will affect the language server too

denis256 (approved):

pkg/config/parsing_context.go
yhakbar: #not1.0yet

PR #6415, opened 2026-06-27
Propagate auto retry to dependency output runs

### LINKED ISSUE #6088: Errors in dependencies are never caught by `retryable_errors`
## Describe the bug

No matter what is written into a `retry` block's `retryable_errors` list (even `.*`), the retry never fires when an error happens in a dependency.  

## Steps To Reproduce

Two units, one dependant on the other one.

Unit a:
```hcl
terraform {
  source = "invalid"
}
```

Unit b:
```hcl
dependency "a" {
  config_path  = "${get_terragrunt_dir()}/../a"
}

errors {
  retry "any" {
    retryable_errors = [
      ".*"
    ]
    max_attempts       = 3
    sleep_interval_sec = 5
  }
}
```

This leads to

```
❮ terragrunt plan
13:21:07.776 INFO   [../a] Downloading Terraform configurations from ../a/invalid into ../a/.terragrunt-cache/_4DvHfbUQWESIhdYu6Nnzs_PNXo/kDiQlvXcdNnAGRTZ6BALL5w-Yw0
13:21:07.777 INFO   TIP (debugging-docs): For help troubleshooting errors, visit https://docs.terragrunt.com/troubleshooting/debugging
13:21:07.777 ERROR  error occurred:

* downloading source url file:///home/paul/test/a/invalid
  error occurred:
  
  * stat /home/paul/test/a/invalid: no such file or directory
```

Without any retries.

## Expected behavior

I expect a way to catch retryable errors in dependencies. Ideally I wish the regex would match the exact error, so that (in the above example) I could for example match the error with `retryable_errors = [ "(?s).*downloading source url.*"]`

## Must haves

## Nice to haves

- [x] Terminal output
- [ ] Screenshots

## Versions

- Terragrunt version: 1.0.4
- OpenTofu/Terraform version: v1.11.6
- Environment details (Ubuntu 20.04, Windows 10, etc.):

## Additional context

This is relevant to me, because of error hooks. I have some errors which I can't prevent, but I can fix them by calling for example `tofu init` after the error occurs. Doing this in an `error_hook` combined with an analogous `retry` block, leads to the error effectively being prevented (because it is fixed and retryed automatically). Due to this bug however, it only works if the error occurs in the unit itself and not in one of it's dependencies. The `error_hook` runs, so the error is fixed, but the `retry` doesn't, so I always have to manually run the command again.

goutamadwant:
## Description
Fixes #6088.

Dependency output resolution builds a fresh `run.Options` value before calling the dependency unit's `output -json` command. That path was not carrying the CLI auto-retry setting, so dependency output errors could match an `errors.retry` block but still return immediately instead of retrying.

This adds `AutoRetry` to the parsing context and copies it into dependency output run options. That preserves the default retry behavior and still respects `--no-auto-retry`.

The fix is scoped to retry handling for the dependency unit's own output run. It does not change cross-unit retry behavior.

## TODOs

- [ ] Update the docs. Not applicable.
- [ ] Update the changelog in the docs. Not applicable.
- [ ] If this change is not forwards compatible (e.g. a new feature), it is gated behind a feature flag. Not applicable.

Relevant checks:

- `pre-commit run --files` for the changed Go, HCL, and Terraform fixture files
- `go test ./internal/configbridge ./pkg/config -run 'Test(NewParsingContextCopiesAutoRetry|RunOptionsFromPctxCopiesExecutionFlags)$' -count=1`
- `go test ./test -run TestDependencyOutputRetryHonorsAutoRetry -count=1`

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure automatic retry configuration is consistently applied when executing dependency output, so transient “dependency output” failures are retried only when auto-retry is enabled, and stop retrying when auto-retry is disabled (for example via `--no-auto-retry`).

* **Tests**
  * Added unit tests verifying execution flags (including auto-retry and related controls) are propagated into run options.
  * Added an integration test with fixtures and a lightweight Terraform emulator to validate retry vs non-retry behavior for dependency-output errors.


denis256:
PR contains unsigned commits, and we can't merge it, cant't even start CICD to validate changes

goutamadwant:
Thanks for checking @denis256.I rewrote the branch with a verified signed commit.so the unsigned commits are no longer in the PR history. Let me know. thanks!

goutamadwant:
@denis256 rebased this branch on current main and resolved the conflicts.
latest commit is signed and verified as well. Let me know if there are any other suggestions. thanks!

PR #6418, opened 2026-06-28
fix: pass -replace value-taking flag through to IaC tool

g7r:
## Description

`-replace` was being silently stripped of its value when passed in the space-separated form (`-replace ADDR`).

### Root cause

`internal/iacargs/iacargs.go` re-orders IaC tool arguments into `[command] [flags...] [arguments...]`. In `processFlag`, a flag only consumes its following space-separated value if its name appears in the `valueTakingFlags` whitelist. `replace` is not in that list, so it was treated as a boolean flag: the flag landed in `Flags`, and the resource address — a non-flag token — was classified as a positional `Argument` and re-emitted at the end of the command.

Result, given:

    terraform plan -replace A -replace B -out plan.tfplan

Terragrunt emitted:

    terraform plan -replace -replace -out plan.tfplan A B

Each `-replace` matched no address, and `A`/`B` were passed as positional plan arguments. The plan ran without forcing replacement of any resource — a silent correctness bug, no error raised.

### Fix

Add `"replace"` to `valueTakingFlags`. The inline form (`-replace=ADDR`) was already handled correctly via the `=` short-circuit, so this only restores the documented space-separated form.

## TODOs

- [ ] Update the docs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated command parsing so `replace` is recognized as a value-taking flag, ensuring the following space-separated token is handled correctly.
  * Improved coverage to verify `-replace` works correctly for both `-replace=<value>` and space-separated forms, including repeated occurrences and round-trip behavior.

### DIFF
--- a/internal/iacargs/iacargs.go
+++ b/internal/iacargs/iacargs.go
@@ -41,6 +41,7 @@ var valueTakingFlags = []string{
 	"state-out",
 	"backup",
 	"target",
+	"replace",
 	"var",
 	"var-file",
 }
--- a/internal/iacargs/iacargs_test.go
+++ b/internal/iacargs/iacargs_test.go
@@ -81,6 +81,27 @@ func TestNew(t *testing.T) {
 			wantArgs:  nil,
 		},
 		{
+			name:      "replace with space-separated value",
+			input:     []string{"plan", "-replace", "aws_instance.foo", "-out=tfplan"},
+			wantCmd:   "plan",
+			wantFlags: []string{"-replace", "aws_instance.foo", "-out=tfplan"},
+			wantArgs:  nil,
+		},
+		{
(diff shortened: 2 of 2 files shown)


g7r:
@yhakbar, @denis256, I have here an absolutely trivial single-line PR, that fixes a regression. Can you please have a look?

yhakbar:
Hey @g7r ,

Mind adding some testing to confirm this fix actually resolves the issue?

g7r:
@yhakbar, thanks for the suggestion, I totally missed the tests. The tests were added.

yhakbar:
Hey @g7r ,

That last commit wasn't signed, so we can't run CI on this or merge it. Please rewrite the commit to sign it. Moving in draft for now. Please mark ready for review when it is.

PR #6421, opened 2026-06-28
fix #6420: deleted files being excluded from affected units in worktree discovery

### LINKED ISSUE #6420: `mark_glob_as_read` deleted-file diffs can exclude the affected unit
## Describe the bug

When a file tracked by `mark_glob_as_read(...)` is deleted in a Git diff, Terragrunt can discover the affected unit on the `from` worktree, but it still treats that unit like a removed unit and excludes it from `run --all` planning unless `--filter-allow-destroy` is set.

This is most visible when the commit only deletes the tracked file. If another file is also modified or added in the same commit, the unit can be selected via the `to` worktree, which masks the bug.

Reference: https://github.com/gruntwork-io/terragrunt/issues/6348

## Reproducing bugs

Use a unit that reads files via `mark_glob_as_read(...)`, then delete one of the tracked files in a later commit and run a Git-filtered `terragrunt run --all`.

## Steps To Reproduce

1. Create a unit that uses `mark_glob_as_read("../config/{*.yaml,*.yml}")`.
2. Add a matching file, for example `config/item-a.yml`, and commit.
3. Make a second commit that deletes that tracked file.
4. Run:

```bash
terragrunt run --all --filter '[HEAD^1...HEAD]' -- plan
```

5. Observe that the affected unit is excluded instead of being selected.

Minimal example:

```hcl
locals {
  config_files = mark_glob_as_read("../config/{*.yaml,*.yml}")
}
```

## Expected behavior

The unit should be selected from the `to` worktree when it still exists in `HEAD`, even if a file it reads was deleted in the diff. It should not be excluded as a removed unit.

## Must haves

## Nice to haves

- [ ] Terminal output
- [ ] Screenshots

## Versions

- Terragrunt version: 1.1.0-rc2
- OpenTofu/Terraform version: Terraform v1.15.5
- Environment details: macOS

## Additional context

`WorktreePair.Expand()` correctly creates a reading filter for deleted files, but the unit is initially discovered on the `from` side and gets `-destroy` treatment. Rediscovering the unit in the `to` worktree fixes the issue because it keeps the `HEAD`-side component and avoids the destroy exclusion path.

lucascicco:
## Description

Fixes #6420.

Handle deleted files tracked by `mark_glob_as_read(...)` in Git-filtered `run --all` discovery.

When a tracked file is deleted, Terragrunt was discovering the affected unit on the `from` worktree and then excluding it as if the unit itself had been removed. The fix re-discovers reading-affected units in the `to` worktree so surviving units keep `HEAD`-side identity and are not incorrectly treated as destroy candidates.

Also adds regression coverage for:
- deleted-only tracked file
- modified tracked file
- added tracked file
- deleted + modified tracked file
- deleted + added tracked file

## TODOs

- [ ] Update the docs.
- [X] This change is backwards compatible.
- [X] If this change is not forwards compatible (e.g. a new feature), it is gated behind a feature flag.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed worktree-based discovery to correctly rediscover units when their `read`-tracked files are deleted between git revisions.
  * Ensured those units are attributed to the current (to-worktree) revision side and not marked as deleted/excluded.
  * Improved `terragrunt run --all` behavior with git filters combined with `mark_glob_as_read`, ensuring glob deletion scenarios select the correct units.
* **Tests**
  * Added integration and unit tests covering glob-deletion and affected rediscovery cases, including validation of rediscovery behavior and execution context.

### DIFF
--- a/internal/discovery/phase_worktree.go
+++ b/internal/discovery/phase_worktree.go
@@ -97,7 +97,16 @@ func (p *WorktreePhase) Run(ctx context.Context, l log.Logger, input *PhaseInput
 						return err
 					}
 
-					for _, c := range components {
+					// For reading-affected components discovered in the from-worktree, check if the unit
+					// still exists in the to-worktree. If yes, re-discover it there to avoid the -destroy
+					// flag that would incorrectly exclude it. This handles the case where a file tracked
+					// by mark_glob_as_read is deleted but the unit itself still exists.
+					mappedComponents, err := p.mapFromWorktreeReadingAffected(fromToCtx, l, input, pair, components)
+					if err != nil {
+						return err
+					}
+
+					for _, c := range mappedComponents {
 						discoveredComponents.EnsureComponent(c)
 					}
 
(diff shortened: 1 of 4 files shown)


internal/discovery/phase_worktree.go
```diff
@@ -224,7 +237,81 @@ func (p *WorktreePhase) discoverInWorktree(
...
+		return nil, err
+	}
+
+	result := make(component.Components, 0, len(deletedFromOnly)+len(toComponents))
+	result = append(result, deletedFromOnly...)
+	result = append(result, toComponents...)
+
+	return result, nil
+}
 func (p *WorktreePhase) discoverChangesInWorktreeStacks(
```
yhakbar, line 315: This PR broke the godoc comment for this method.

internal/discovery/phase_worktree.go
```diff
@@ -224,7 +237,81 @@ func (p *WorktreePhase) discoverInWorktree(
...
+	}
+
+	if len(stillExistInTo) == 0 {
+		return deletedFromOnly, nil
+	}
+
+	pathFilters := make(filter.Filters, 0, len(stillExistInTo))
+	for _, c := range stillExistInTo {
+		relPath, err := filepath.Rel(pair.FromWorktree.Path, c.Path())
+		if err != nil {
```
yhakbar, line 288: Why are we ignoring this err?

internal/discovery/phase_worktree.go
```diff
@@ -224,7 +237,81 @@ func (p *WorktreePhase) discoverInWorktree(
...
+			continue
+		}
+
+		pathExpr, err := filter.NewPathFilter(relPath)
+		if err != nil {
```
yhakbar, line 293: Why are we ignoring this err?

internal/discovery/phase_worktree_internal_test.go
```diff
@@ -0,0 +1,97 @@
+package discovery
```
yhakbar, line 1: Avoid testing internals of packages. Update this to `discovery_test` and test the publicly exported interface of the package. If necessary export currently private functions.

internal/discovery/phase_worktree_internal_test.go
```diff
@@ -0,0 +1,97 @@
...
+		input,
+		worktrees.WorktreePair{
+			FromWorktree: worktrees.Worktree{Path: fromDir, Ref: "HEAD~1"},
+			ToWorktree:   worktrees.Worktree{Path: toDir, Ref: "HEAD"},
+		},
+		component.Components{fromComponent},
+	)
+
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "Git-based filtering is not supported")
```
yhakbar, line 95: Avoid making assertions on error text like this. We should assert on error type/values.

internal/discovery/phase_worktree.go
```diff
@@ -224,7 +237,81 @@ func (p *WorktreePhase) discoverInWorktree(
...
+	for _, c := range fromComponents {
+		relPath, err := filepath.Rel(pair.FromWorktree.Path, c.Path())
+		if err != nil {
+			deletedFromOnly = append(deletedFromOnly, c)
+			continue
+		}
+
+		toPath := filepath.Join(pair.ToWorktree.Path, relPath)
+		configFile := c.ConfigFile()
+		if configFile == "" {
```
yhakbar, line 269: What is this actually guarding against? `ConfigFile` should always be populated, right?

internal/discovery/phase_worktree.go
```diff
@@ -224,7 +237,81 @@ func (p *WorktreePhase) discoverInWorktree(
...
+	}
+
+	var (
+		stillExistInTo  []component.Component
+		deletedFromOnly []component.Component
+	)
+
+	for _, c := range fromComponents {
+		relPath, err := filepath.Rel(pair.FromWorktree.Path, c.Path())
+		if err != nil {
```
yhakbar, line 262: What's happening here? Why do we drive this logic from relative-ization failing?

yhakbar:
Thanks for submitting this, @lucascicco .

What do you think of the following, alternate fix?
https://github.com/gruntwork-io/terragrunt/pull/6424

I think it should achieve the same end result.

lucascicco:
hey @yhakbar - Thanks for the proactive response. I'm happy to go with your alternate fix instead. Feel free to close out this PR, and let me know if you'd like a second pair of eyes on your changes.

PR #6424, opened 2026-06-29
fix: Fixing deleted changed files

### LINKED ISSUE #6420: `mark_glob_as_read` deleted-file diffs can exclude the affected unit

yhakbar:
## Description

Fixes #6420.

Closes #6421

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved worktree diff discovery for `mark_glob_as_read` when referenced files are deleted, ensuring surviving reader units and stacks are rediscovered on the correct side.
  * Correctly handles removal vs. rediscovery when the reader itself is deleted in the same change, preserving expected destroy behavior.
  * Refined stack “reading affected” tracking so all relevant stacks are recorded when multiple reading filters and files change.

* **Tests**
  * Added/extended integration and regression tests covering deleted-file reading for units and stacks, plus filter partitioning behavior.

### DIFF
--- a/internal/discovery/phase_worktree.go
+++ b/internal/discovery/phase_worktree.go
@@ -88,11 +88,16 @@ func (p *WorktreePhase) Run(ctx context.Context, l log.Logger, input *PhaseInput
 				return err
 			}
 
+			// Expand routes reading filters for deleted files onto the from side, since a deleted file
+			// only exists in the from worktree where its read relationship can be evaluated. These need
+			// different handling from the path filters for genuinely removed components, so split them.
+			deletedReadFilters, removalFilters := fromFilters.PartitionReadingFilters()
+
 			fromToG, fromToCtx := errgroup.WithContext(discoveryCtx)
 
-			if len(fromFilters) > 0 {
+			if len(removalFilters) > 0 {
 				fromToG.Go(func() error {
-					components, err := p.discoverInWorktree(fromToCtx, l, input, pair.FromWorktree, fromFilters, FromWorktreeKind)
+					components, err := p.discoverInWorktree(fromToCtx, l, input, pair.FromWorktree, removalFilters, FromWorktreeKind)
 					if err != nil {
 						return err
 					}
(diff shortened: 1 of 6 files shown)


internal/discovery/phase_worktree.go
```diff
@@ -224,6 +244,42 @@ func (p *WorktreePhase) discoverInWorktree(
 	return components, nil
 }
 
+// deletedReadingComponentsToFilters discovers, in the from worktree, the components that read files
+// deleted in the diff (those files only exist on the from side), then returns a path filter per
+// discovered component aimed at its equivalent path in the to worktree. A component that no longer
+// exists in the to worktree simply matches nothing there, which is the expected outcome for a genuine
+// removal that another filter already covers.
+func (p *WorktreePhase) deletedReadingComponentsToFilters(
```
denis256, line 254: Looks like `deletedReadingComponentsToFilters` converts deleted reading= matches into plain path filters, so stacks are rediscovered as stack files instead of being added to w.ReadingAffectedStacks for generated-unit discovery

yhakbar: Addressed.

denis256 (approved):

PR #6426, opened 2026-06-29
chore: Engine per unit shutdown

yhakbar:
## Description

Updates engine management so that engines are init and shutdown with the lifecycle of their unit runs, rather than being done in batch. As part of this, the UnitDir and CacheDir were split from WorkingDir so that we don't overload the meaning of the field.

This is important, as it allows for resource exhaustion prevention on init and improved resource reclamation on shutdown.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **New Features**
  * Added support for separate unit and cache directories, improving consistency of generated outputs and execution behavior.
  * Introduced safer, earlier per-unit engine shutdown using dependency-aware gating.
* **Bug Fixes**
  * Corrected directory handling across source download, config generation, init/module detection, hooks, debug output, and error reporting to use the appropriate unit/cache locations.
  * Improved engine reuse/recovery under concurrent startups and build failures by keying and managing engine instances per cache directory.
* **Tests**
  * Added/updated coverage for single-flight engine creation, retry-after-failure behavior, shutdown no-op cases, and plugin-exit waiting.

### DIFF
--- a/internal/runner/runnerpool/runner.go
+++ b/internal/runner/runnerpool/runner.go
@@ -19,6 +19,7 @@ import (
 	"github.com/gruntwork-io/terragrunt/internal/util"
 
 	"github.com/gruntwork-io/terragrunt/internal/component"
+	"github.com/gruntwork-io/terragrunt/internal/engine"
 	"github.com/gruntwork-io/terragrunt/internal/experiment"
 	"github.com/gruntwork-io/terragrunt/internal/os/stdout"
 	"github.com/gruntwork-io/terragrunt/internal/queue"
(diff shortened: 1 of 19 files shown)


internal/runner/runnerpool/runner.go
```diff
@@ -506,6 +526,21 @@ func (rnr *Runner) Run(ctx context.Context, l log.Logger, v run.Venv, stackOpts
...
 			})
 
+			// This unit's terraform commands are all done, so release its engine now
+			// instead of holding it until the batch Shutdown. Skip units that another
+			// in-run unit depends on: that dependent re-reads this unit's outputs through
+			// engine.Run after this task ends, which would just re-spawn the engine we
+			// tore down.
+			if !withDependents[unitPath] {
+				noEngine := unitOpts.EngineOptions != nil && unitOpts.EngineOptions.NoEngine
+				if sErr := engine.ShutdownUnit(
```
denis256, line 536: Potentially, one stuck plugin shutdown for some reason may prevent other units to be processed

```
  run --all --parallelism 1 plan

  unit-a: terraform plan finished
  unit-a: ShutdownUnit called
  engine plugin: Shutdown stream hangs
  runner worker: blocked forever
  unit-b: never starts
```

yhakbar: Added timeout. Didn't we always have this issue with the batch shutdown process too? We allowed the unit to run, but we just blocked on shutdown forever.

denis256 (approved):

PR #6438, opened 2026-07-01
fix(catalog): harden temporary clone paths

denis256:
## Description

* Fixes terragrunt catalog so it no longer uses predictable temp clone paths that could be pre-created as symlinks.
* Adds checks to reject symlinked clone roots before cleanup, plus tests, docs, and a v1.1.1 changelog entry.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Catalog temp directories are now created fresh for each load, rejected when the clone root is a symlink, and cleaned up when the TUI session ends.
  * Clone handling now validates paths more safely to avoid removing files outside the intended location.

* **Tests**
  * Added coverage for fresh temp directory creation, symlink handling, session cancellation cleanup, and safe clone-root validation.

* **Documentation**
  * Added a changelog entry for the catalog hardening update.

### DIFF
--- a/docs/src/content/docs/03-features/06-catalog/02-tui.mdx
+++ b/docs/src/content/docs/03-features/06-catalog/02-tui.mdx
@@ -23,12 +23,6 @@ If `<repo-url>` is provided, the repository will be cloned into a temporary dire
 1. The repository list are searched in the config file `terragrunt.hcl`. if `terragrunt.hcl` does not exist in the current directory, the config are searched in the parent directories.
 1. If the repository list is not found in the configuration file, the modules are looked for in the current directory.
 
-## Temporary clone directories
-
-When loading remote repositories, Terragrunt creates a fresh temporary clone directory for each repository under the system temp directory. The temp root is resolved through symlinks before discovery so component paths stay inside the clone on systems where the temp directory itself is reported through a symlink.
-
-Terragrunt does not use a predictable temp path for catalog loads. It refuses to use or clean up clone roots that are symlinks, so local temp-dir entries cannot redirect catalog cleanup outside the directory Terragrunt is preparing. Each clone remains available while the TUI session is running because unit and stack copy actions read files from it, then Terragrunt removes the temporary clone when the session exits.
-
 An example of how to define the optional default template and the list of repositories for the `catalog` command in the `terragrunt.hcl` configuration file:
 
 ``` hcl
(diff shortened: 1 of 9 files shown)


docs/src/content/docs/03-features/06-catalog/02-tui.mdx
```diff
@@ -23,6 +23,12 @@ If `<repo-url>` is provided, the repository will be cloned into a temporary dire
 1. The repository list are searched in the config file `terragrunt.hcl`. if `terragrunt.hcl` does not exist in the current directory, the config are searched in the parent directories.
 1. If the repository list is not found in the configuration file, the modules are looked for in the current directory.
 
+## Temporary clone directories
```
yhakbar: Let's get rid of this. I don't think the average user cares.

internal/cli/commands/catalog/tui/load_test.go
```diff
@@ -28,10 +35,113 @@ func TestCatalogTempPathResolvesSymlinkedTmpDir(t *testing.T) {
...
 	// symlink component left for filepath.Rel to trip over later.
 	assert.Equal(t, realTmp, filepath.Dir(got),
-		"CatalogTempPath should anchor the clone at the symlink-resolved temp dir")
+		"CreateCatalogTempPath should anchor the clone at the symlink-resolved temp dir")
+}
+
+func TestCreateCatalogTempPathUsesFreshDirectory(t *testing.T) {
+	t.Parallel()
+
+	first, err := tui.CreateCatalogTempPath(vfs.NewOSFS())
```
yhakbar: Why aren't we using the vfs here?

internal/cli/commands/catalog/tui/load.go
```diff
@@ -14,13 +14,13 @@ import (
...
-
 // CreateCatalogTempPath creates a fresh clone root under the resolved temp dir.
 // Resolving os.TempDir keeps filepath.Rel results inside the clone on systems
 // where the temp dir itself is reported through a symlink.
-func CreateCatalogTempPath(fsys vfs.FS) (string, error) {
-	return vfs.MkdirTemp(fsys, util.ResolvePath(os.TempDir()), tempDirPattern)
+func CreateCatalogTempPath(fsys vfs.FS, repoURL string) (string, error) {
+	pattern := fmt.Sprintf("catalog-%s-*", util.EncodeBase64Sha1(repoURL))
+
+	return vfs.MkdirTemp(fsys, util.ResolvePath(os.TempDir()), pattern)
```
yhakbar: This is a prefix, not a pattern. We should drop the `*`:
https://github.com/gruntwork-io/terragrunt/blob/main/internal/vfs/vfs.go#L190-L193

yhakbar (approved):

PR #6447, opened 2026-07-05
fix: Preserving SemVer build metadata in tfr:// module versions (#6345)

### LINKED ISSUE #6345: tfr source mishandles SemVer build metadata
## Describe the bug

`tfr:` module source URLs incorrectly handle SemVer build metadata in the `version` query parameter. A version like `1.2.3+build.5` can be parsed as `1.2.3 build.5` because `+` is decoded as a query-space separator, even though `+build.5` is valid SemVer build metadata.

## Steps To Reproduce

Use a `terraform` block with a `tfr:` source that points to a registry module version published with build metadata.

```hcl
terraform {
  source = "tfr://registry.example.com/namespace/name/provider?version=1.2.3+build.5"
}
```

Run Terragrunt against the module.

## Expected behavior

Terragrunt should preserve the `+` as part of the SemVer version and request the registry download endpoint for `1.2.3+build.5`.

If `?version=1.2.3` is requested and the registry only has an equivalent metadata version such as `1.2.3+build.5`, Terragrunt should resolve and use the published metadata version.

## Must haves

## Nice to haves

- [ ] Terminal output
- [ ] Screenshots

## Versions

- Terragrunt version: v1.0.8
- OpenTofu/Terraform version:
- Environment details:

## Additional context

SemVer build metadata starts with `+` and does not affect version precedence, but it is still part of the version identifier and must be preserved when resolving an explicitly requested registry module version.

kromol:
## Description

Fixes #6345.

url.Query() decodes '+' in a query string as a literal space. Registry module versions such as `1.2.3+build.5` were silently incorrectly decoded into '1.2.3 build.5' before ever reaching the registry, breaking module resolution for anyone pinning a module to a build-metadata version.

Add QueryValues, a raw-query parser that unescapes percent-sequences via url.PathUnescape without treating '+' as space.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved version parsing so module versions with build metadata like `1.2.3+build.5` are handled correctly.
  * Version-related errors now consistently fall back to “no version metadata” when the version query is missing or invalid.
  * Module retrieval now works reliably for versions containing `+` metadata, including downloads to the destination directory.

* **Tests**
  * Added coverage for version strings with build metadata and related download behavior.

### DIFF
--- a/internal/getter/resolver_tfr.go
+++ b/internal/getter/resolver_tfr.go
@@ -90,8 +90,8 @@ func (r *TFRResolver) Probe(ctx context.Context, rawURL string) (string, error) 
 		registryDomain = tfimpl.DefaultRegistryDomain(r.TofuImplementation)
 	}
 
-	versionList, hasVersion := srcURL.Query()[versionQueryKey]
-	if !hasVersion || len(versionList) != 1 || versionList[0] == "" {
+	versionList, err := QueryValues(srcURL.RawQuery, versionQueryKey)
+	if err != nil || len(versionList) != 1 || versionList[0] == "" {
 		return "", cas.ErrNoVersionMetadata
 	}
 
--- a/internal/getter/resolver_tfr_test.go
+++ b/internal/getter/resolver_tfr_test.go
@@ -31,6 +31,24 @@ func TestTFRResolver_ProbeReturnsContentKey(t *testing.T) {
 	assert.Equal(t, expected, key)
 }
 
+func TestTFRResolver_ProbePreserveSemverMetadata(t *testing.T) {
+	t.Parallel()
+
+	server := newRegistryTestServer(t)
+
+	r := getter.NewTFRResolver().
+		WithHTTPClient(server.Client()).
+		WithLogger(logger.CreateLogger())
+
+	src := "tfr://" + server.Listener.Addr().String() + "/terraform-aws-modules/vpc/aws?version=1.2.3+build.5"
+
+	key, err := r.Probe(t.Context(), src)
+	require.NoError(t, err)
+
+	expected := cas.ContentKey("tfr-xtg", "https://"+server.Listener.Addr().String()+"/download/terraform-aws-vpc.zip")
+	assert.Equal(t, expected, key)
+}
+
 // TestTFRResolver_ProbeIsStable pins that two probes of the same URL
 // produce the same key, so the CAS hit path on a repeated run is
 // deterministic.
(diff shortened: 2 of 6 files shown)


kromol:
Quick scope question for maintainers: this PR fixes the "+" decoded-as-space bug so explicit build-metadata versions like `?version=1.2.3+build.5` now resolve correctly. 

The original issue also mentions a related but distinct case: a plain version request (e.g. ?version=1.2.3) resolving to a published build-metadata version (e.g. 1.2.3+build.5) when no bare 1.2.3 exists, since SemVer build metadata doesn't affect precedence. That would need new matching logic against the list-versions endpoint - a bigger change than  
the encoding fix here. 

Should that be included in this PR, or would you rather I split it into a separate issue/PR? Happy to do either.

Also, I checked [Semver spec](https://semver.org/#spec-item-10) and it indeed tells 
>Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence

But it does not say anything about how to resolve versions in such case, so there's a real risk in "auto-resolve" behavior.

yhakbar:
I'm not sure we should make this change. Wouldn't it be more appropriate to use `%2B` to URL encode the `+` character? Is that not supported already? FWIW, we are starting to work on #1930, which should make this kind of thing more ergonomic.

kromol:
>Wouldn't it be more appropriate to use %2B to URL encode the + character? Is that not supported already?

Just to confirm - you mean asking user to write `%2B` in their source config, right? Like this:
```
terraform {
  source = "tfr://registry.example.com/namespace/name/provider?version=1.2.3%2Bbuild.5"
}
```

yhakbar:
> > Wouldn't it be more appropriate to use %2B to URL encode the + character? Is that not supported already?
> 
> Just to confirm - you mean asking user to write `%2B` in their source config, right? Like this:
> 
> ```
> terraform {
>   source = "tfr://registry.example.com/namespace/name/provider?version=1.2.3%2Bbuild.5"
> }
> ```

Ya, pending support for the `version` attribute, which would allow for something like this:

```hcl
terraform {
  source  = "tfr://registry.example.com/namespace/name/provider"
  version = "1.2.3+build.5"
}
```

kromol:
Unless I am missing something - even if `version` becomes a separate attribute, under the hood it still needs to be rebuilt into a single `tfr://registry.example.com/path?version=...` URL, so the bug would still be there.

Regarding asking user to write `%2B` directly - IMO it's bad DX because no one types that manually, those versions are usually copy-pasted.

yhakbar:
@kromol 

Under the hood, the URL is URL encoded properly with `%2B` instead of `+`. You can try it out today!
https://docs.terragrunt.com/reference/experiments/active/#version-attribute
  
To be clear, nobody would be expected to type out `%2B` directly as the recommended approach. The recommended approach would be to use the `version` attribute, which will do the URL encoding for you.

kromol:
Ah okay, I misunderstood your initial suggestion. 

If this is going to be covered by `version` attribute and become default, then I will close this PR

PR #6474, opened 2026-07-10
fix: Fixing `run --all` with graph expression throwing on missing working dir

### LINKED ISSUE #6472: run --all dependency discovery fails with "component has a discovery context but is missing its working directory" when a git-range filter is nested in a graph expression
**Describe the bug**

Running `terragrunt run --all --filter '...[origin/main...HEAD]... | ./envs/dev/**' -- plan` fails during dependency discovery with the internal-invariant error the tool asks us to report:

```
ERROR Component at path '/tmp/terragrunt-worktree-HEAD-<id>/envs/dev/aurora' has a discovery context but is missing its working directory during dependency discovery. This is a bug in Terragrunt. Please open a bug report [...]
```

The components named in the error (`envs/dev/aurora`, `envs/dev/batch`, `envs/dev/twingate` across runs) are **excluded by the filter** — they are neither ancestors nor descendants of the changed units — but the discovery walk still touches them inside the git worktree and hits the invariant.

**Reproduction**

- Build: `main` @ `47b1df9` (we build from source to pick up the destroy-on-removal fix from #6322).
- Monorepo layout: ~30 Terragrunt units under `envs/dev/`, moderately dense `dependency` graph (a few hub units — networking, storage — referenced by many).
- A branch whose change closure spans several units: 2 added units, 2 modified units (one of them a hub referenced by many others). Net diff contains **no deletions**.
- `terragrunt find --filter '...[origin/main...HEAD]... | ./envs/dev/**'` on the same tree **succeeds** and resolves the expected set.
- `terragrunt run --all --filter '...[origin/main...HEAD]... | ./envs/dev/**' -- plan` fails with the error above, 100% reproducible on this branch (CI and re-runs). Narrower branches against the same graph pass, so it appears sensitive to the size/shape of the filtered closure.
- The same filter without the git range (`./envs/dev/**` alone) does not fail.

**Expected behavior**

Filter-excluded components either aren't visited during dependency discovery, or are visited with a valid working directory — `run --all` should succeed wherever `find` resolves the same filter.

**Additional context**

Happy to provide full debug logs (`TG_LOG=debug`) from CI privately, and to test candidate fixes against our graph. Possibly adjacent to the discovery-related changes in #6362 / #6425, though neither seems to target this invariant.

yhakbar:
## Description

Fixes #6472.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a crash when running `run --all` with graph filters that expand Git ranges through the dependency graph.
  * Improved dependency discovery correctness so upstream/downstream candidates get their working directory set before they become visible, preventing inconsistent behavior across graph sizes.

* **Tests**
  * Added race-focused coverage for concurrent dependency discovery scenarios, including shared/external dependencies reached from multiple traversal directions.
  * Added assertions that every discovered component ends up with a non-empty discovery working directory, even under racing traversal.

### DIFF
--- a/internal/discovery/phase_graph.go
+++ b/internal/discovery/phase_graph.go
@@ -62,7 +62,12 @@ func (p *GraphPhase) Kind() PhaseKind {
 }
 
 // Run executes the graph discovery phase.
-func (p *GraphPhase) Run(ctx context.Context, l log.Logger, v venv.Venv, input *PhaseInput) (*PhaseResults, error) {
+func (p *GraphPhase) Run(
+	ctx context.Context,
+	l log.Logger,
+	v venv.Venv,
+	input *PhaseInput,
+) (*PhaseResults, error) {
 	results := NewPhaseResults()
 
 	discovery := input.Discovery
@@ -236,7 +241,17 @@ func (p *GraphPhase) processGraphTarget(
 
 			visitedDirs := newStringSet()
 
-			err := p.discoverDependentsUpstream(ctx, l, v, state, c, visitedDirs, startDir, boundaryRoot, depth)
+			err := p.discoverDependentsUpstream(
+				ctx,
+				l,
+				v,
+				state,
+				c,
+				visitedDirs,
+				startDir,
+				boundaryRoot,
+				depth,
+			)
 			if err != nil {
 				return err
 			}
@@ -325,7 +340,14 @@ func (p *GraphPhase) discoverDependencies(
 					Phase:     PhaseGraph,
 				})
 
-				err = p.discoverDependencies(contextWithIncrementedParseDepth(ctx), l, v, state, depComponent, depthRemaining-1)
+				err = p.discoverDependencies(
+					contextWithIncrementedParseDepth(ctx),
+					l,
+					v,
+					state,
+					depComponent,
+					depthRemaining-1,
+				)
 				if err != nil {
 					errMu.Lock()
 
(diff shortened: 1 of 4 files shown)


internal/discovery/phase_graph.go
```diff
@@ -762,8 +762,7 @@ func (p *GraphPhase) resolveDependency(
 
 	depComponent := componentFromDependencyPath(depPath, threadSafeComponents)
 
-	addedComponent, created := threadSafeComponents.EnsureComponent(depComponent)
-	if created {
+	if dctx := depComponent.DiscoveryContext(); dctx == nil || dctx.WorkingDir == "" {
```
denis256: Should we add synchronization for this?

yhakbar: Sure, I'll add a mutex guard.

internal/discovery/phase_graph.go
```diff
@@ -762,28 +814,46 @@ func (p *GraphPhase) resolveDependency(
...
+// before its working directory is set.
+//
+// It is a no-op once the component already has a working directory, which keeps
+// the assignment to the first goroutine to reach the component along any path.
+func assignGraphDiscoveryContext(
+	dep component.Component,
+	parentCtx *component.DiscoveryContext,
+	depPath string,
+) {
+	if dctx := dep.DiscoveryContext(); dctx != nil && dctx.WorkingDir != "" {
```
denis256, line 836: Could we set external on the component returned by EnsureComponent, since dep may be discarded and leave the canonical component marked as internal?

denis256 (dismissed):

internal/discovery/phase_graph.go
yhakbar: I don't think we need to. The external flag should be preserved.

denis256 (approved):

PR #6513, opened 2026-07-20
fix(cas): strip go-getter `depth` query parameter before invoking git

### LINKED ISSUE #6512: Regression since v1.0.5: `terraform.source` with `?depth=N&ref=<tag>` fails to download (CAS passes `depth` into the git URL)
## Describe the bug

Since the v1.x line, a unit whose `terraform.source` includes the go-getter `depth` query parameter (e.g. `...git?depth=1&ref=v5.21.0`) fails to download. The identical source works on v0.93.13.

The CAS (content-addressable store) path now used for `terraform.source` clones does **not strip the go-getter** `depth`**/**`ref` **query parameters before invoking native** `git`. It hands `git` the URL with the query string still attached (`https://github.com/.../terraform-aws-vpc.git?depth=1`), so `git` treats `?depth=1` as part of the repository name and rejects it:

```
central git store unavailable for https://github.com/terraform-aws-modules/terraform-aws-vpc.git?depth=1,
falling back to temporary clone: git_fetch: fatal: remote error:
  is not a valid repository name
```

(On GitHub Actions Ubuntu runners the same root cause surfaces slightly differently — a redirect to `.../terraform-aws-vpc?depth=1/` → `fatal: ...?depth=1/info/refs not valid: is this a git repository?`.)

CAS then falls back to the "standard getter", which performs a shallow clone (`depth=1`) of the default branch and then tries to check out the `ref` tag that the shallow clone never fetched, so the whole download fails with:

```
* git exited with 1: error: pathspec 'v5.21.0' did not match any file(s) known to git
```

Net effect: any `terraform.source` that combines `depth` with `ref=<tag>` is un-downloadable on v1.x. Dropping `depth` works; pinning back to v0.93.13 works.

## Steps To Reproduce

1. Create a single `terragrunt.hcl` (no backend, provider, or AWS credentials required — it fails during source download):

```hcl
terraform {
  source = "https://github.com/terraform-aws-modules/terraform-aws-vpc.git?depth=1&ref=v5.21.0"
}
```

2. Run:

```
terragrunt init
```

Result: exit code 1, ending in `pathspec 'v5.21.0' did not match any file(s) known to git`.

Controls (same module, same tag, isolated cache dir each time) that isolate `depth` as the trigger and confirm the regression:

| Terragrunt | `source` | Result |
| -- | -- | -- |
| **v1.1.1** | `...vpc.git?depth=1&ref=v5.21.0` | ❌ fails (see below) |
| **v1.1.1** | `...vpc.git?ref=v5.21.0` (no `depth`) | ✅ downloads & inits |
| **v0.93.13** | `...vpc.git?depth=1&ref=v5.21.0` | ✅ downloads & inits |

## Expected behavior

`terraform.source` with a `depth` query parameter should download on v1.x exactly as it did on v0.93.x: CAS should parse the go-getter query parameters out of the URL — translating `depth` into a shallow clone and `ref` into the checked-out revision — instead of passing `?depth=1` through as part of the git repository URL. The standard-getter fallback should also fetch the requested `ref` when a shallow clone is requested (rather than shallow-cloning only the default branch and then failing to check out the tag).

## Must haves

- [X] Steps for reproduction provided.

## Nice to haves

- [X] Terminal output
- [ ] Screenshots

Full output of the failing case (`terragrunt init`, v1.1.1, `?depth=1&ref=v5.21.0`):

```
INFO   Downloading Terraform configurations from git::https://github.com/terraform-aws-modules/terraform-aws-vpc.git?depth=1&ref=v5.21.0 into ./.terragrunt-cache/.../...
WARN   central git store unavailable for https://github.com/terraform-aws-modules/terraform-aws-vpc.git?depth=1, falling back to temporary clone: git_fetch: fatal: remote error:
  is not a valid repository name
Visit https://support.github.com/ for help
failed to complete git fetch
exit status 128
WARN   CAS download failed: fetch https://github.com/terraform-aws-modules/terraform-aws-vpc.git?depth=1: git_clone: Cloning into bare repository '/var/folders/.../terragrunt-cas-fallback-...'...
fatal: remote error:
  is not a valid repository name
Visit https://support.github.com/ for help
failed to complete git clone
exit status 128. Falling back to standard getter.
INFO   TIP (debugging-docs): For help troubleshooting errors, visit https://docs.terragrunt.com/troubleshooting/debugging
ERROR  downloading source url git::https://github.com/terraform-aws-modules/terraform-aws-vpc.git?depth=1&ref=v5.21.0
error occurred:

* /opt/homebrew/bin/git exited with 1: error: pathspec 'v5.21.0' did not match any file(s) known to git
```

## Versions

* Terragrunt version: **regression introduced in v1.0.5** (last good: **v1.0.4**) — bisected across all 27 releases from v0.93.13 to v1.1.1 (table under *Additional context*). Still present on latest **v1.1.1**.
* OpenTofu/Terraform version: **OpenTofu v1.12.4** (the failure is independent of the OpenTofu version — it occurs before OpenTofu runs)
* Environment details: reproduced on **macOS (git 2.55.0)** and on **GitHub Actions** `ubuntu-latest` **runners**

## Additional context

**Bisected — introduced in** `v1.0.5` **(last good** `v1.0.4`**).** I ran the reproduction above against every one of the 27 releases from v0.93.13 to v1.1.1 (`terragrunt init`, a fresh isolated cache per run, OpenTofu v1.12.4). The boundary is clean and monotonic:

| Version(s) | Result |
| -- | -- |
| v0.93.13 → **v1.0.4** (all 21 releases) | ✅ source downloads & inits |
| **v1.0.5, v1.0.6, v1.0.7, v1.0.8** | ❌ fails: `pathspec 'v5.21.0' did not match any file(s) known to git` |
| **v1.1.0, v1.1.1** | ❌ fails: `... is not a valid repository name` / `not valid: is this a git repository?` |

Both failure forms share the same root cause (the `depth` query parameter is left in the URL handed to native `git`); the *surfaced* error changes at v1.1.0, where the CAS git-store rejects the URL earlier, before the standard-getter fallback is reached.

* `depth` is a go-getter query parameter (shallow clone) that Terragrunt forwards to the module getter; it is not a native `git` URL parameter, so it must be stripped from the URL before `git` is invoked.
* Workaround: remove `depth=` from the `source` (a full clone resolves the `ref` tag correctly), or pin Terragrunt to **v1.0.4** or earlier.
* Possibly related CAS / git-URL query-parameter handling issues: [#5197](<https://github.com/gruntwork-io/terragrunt/issues/5197>) (CAS not used for https sources), [#4777](<https://github.com/gruntwork-io/terragrunt/issues/4777>) (`catalog` ignores `ref` in git URLs), gruntwork-io/terragrunt#6217 (`update_source_with_cas` path handling).

HalisCz:
## Description

Fixes #6512.

Since v1.0.5, a `terraform.source` whose URL combines the go-getter `depth` query parameter with a `ref` tag — e.g. `...vpc.git?depth=1&ref=v5.21.0` — fails to download.

**Root cause:** The CAS git getter (`getGit`) lifted `ref` out of the URL query but left `depth` in place. `depth` is a go-getter parameter (shallow-clone hint), not a native git URL parameter, so native `git` received `...vpc.git?depth=1` and rejected it as an invalid repository name. CAS then fell back to the standard getter, which shallow-cloned only the default branch and failed to check out the tag (`pathspec 'v5.21.0' did not match…`).

Go-getter's own git getter strips `ref`, `sshkey`, **and** `depth`; CAS only handled `ref`.

**Version nuance:** the regression begins in the `hashicorp/go-getter` layer at v1.0.5, whose shallow clone dropped `--branch <ref>` so a tag behind the default branch was never fetched; CAS was still opt-in then. The CAS-specific mis-parse above only applies from v1.1.0, once CAS became the default Git path — before that the same URL already failed through the standard getter.

**Fix:** Parse and strip `depth` alongside `ref` (matching go-getter), translating it into a shallow clone routed through CAS's already-working `--branch <ref> --depth …` clone path (which is why `?ref=` alone already worked). The stack-source clone path in `ProcessStackComponent` had the same ref-only stripping, so both call sites now share a single `StripGitURLParams` helper. A `depth` supplied on the URL takes precedence over the ambient CAS clone depth.

The standard-getter fallback shortcoming the issue also mentions lives in the `hashicorp/go-getter` dependency and is no longer reached once CAS handles `depth` correctly.

### Verification

- New E2E test `TestCASClone_E2E_DepthQueryParamWithTag` reproduces the issue against a local git server: a tag placed *behind* HEAD, cloned via `?depth=1&ref=<tag>` through the full CASGetter dispatch. Confirmed it fails without the fix (with the exact `central git store unavailable … ?depth=1` error) and passes with it.
- New E2E test `TestCASClone_E2E_URLDepthOverridesAmbientFullHistory` proves the URL `depth` overrides the ambient clone depth: with the ambient set to full history (`--cas-clone-depth=-1`), a `?depth=1` still forces a shallow fetch (asserted via the `shallow` marker in the central store's bare repo). It distinguishes an honored URL `depth` from one that is parsed and discarded — which `DepthQueryParamWithTag` cannot, since the CAS default depth is already 1.
- Unit tests `TestGitURLParams` (getter) and `TestStripGitURLParams` (cas) pin the query-parameter parsing.
- `go test ./internal/cas/ ./internal/getter/ ./internal/runner/run/` — all pass.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed CAS Git cloning for go-getter URLs that include both `ref` and `depth`, ensuring the tagged commit is checked out and the clone is shallow.
  - `depth` from the source URL now correctly overrides the ambient `cas-clone-depth`, even when the ambient depth requests full history.
  - CAS now strips `ref`/`depth` query parameters before invoking git while preserving unrelated query parameters.

- **Documentation**
  - Updated changelog and CAS Git docs to clarify `ref`/`depth` URL parameter behavior, precedence, and shallow-clone guidance.
  - Added `Since 1.1.2` notes to the `cas-clone-depth` flag documentation.


docs/src/data/changelog/v1.1.3/cas-depth-query-param.mdx
```diff
@@ -0,0 +1,10 @@
+---
+version: "v1.1.3"
```
yhakbar: Why is this targeting 1.1.3? We typically just target the next patch (which would be `v1.1.2` here) unless a minor release is being prepared.

HalisCz: Moved to v1.1.2 — you're right, v1.1.3 skipped the active patch bucket. Updated the frontmatter version to match as well.

docs/src/content/docs/03-features/07-caching/04-cas/02-git.mdx
```diff
@@ -21,6 +21,13 @@ Git is the most common source Terragrunt fetches and has the richest support in
 
 A Git source is identified by a remote URL and a ref (a branch, tag, or commit). Terragrunt resolves that ref to a commit hash, which serves as both the cache key and the Git object name it fetches. Content keeps Git's native content addressing, so identical objects are shared across repositories and refs.
 
+## Source URL parameters
+
+A Git source URL may carry the go-getter `ref` and `depth` query parameters (for example `...vpc.git?ref=v5.21.0&depth=1`). Both are go-getter parameters, not native Git URL parameters, so the CAS strips them from the URL before invoking `git` — otherwise `git` treats a trailing `?depth=1` as part of the repository name and rejects the clone.
+
+- **`ref`** selects the revision (branch, tag, or commit) to resolve and check out.
+- **`depth`** requests a shallow clone and overrides the ambient `--cas-clone-depth` for that source. Only positive integers take effect; other values fall back to the ambient depth. Because the ambient default is already `1` (a shallow clone of the resolved commit), a `?depth=1` only changes behavior when the ambient depth has been widened (for example `--cas-clone-depth=-1` for full history).
+
```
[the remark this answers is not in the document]
HalisCz: Addressed in `185e4e0b`. Wrapped the unreleased behavior in `<Since>`, but **split** it rather than gating the whole section:

- **`depth`** query-param handling is genuinely new in v1.1.2 (it was broken v1.0.5–v1.1.1 per #6512), so its bullet on the Git page and the added paragraph in `cas-clone-depth.mdx` are now inside `<Since version="1.1.2">`.
- **`ref`** is left **ungated**: it is long-standing, not unreleased. I verified this empirically with the official release binaries — `...git?ref=<tag>` (no `depth`) downloads successfully on every supported version (v0.93.13 → v1.1.1), since ref-only does a full clone. Gating it would hide already-shipped behavior.

Two deviations from the literal suggestion, both intentional:
- Used the bare `version="1.1.2"` form (no `v` prefix) to match the existing `<Since>`/`<Before>` usages across the docs.
- Kept the `## Source URL parameters` heading **outside** `<Since>`, per `Since.astro`'s own guidance (headings inside the component still leak into the table of contents).

HalisCz:
@yhakbar Is there anything I can do to push this forward?

yhakbar:
Hey @HalisCz ,

We'll also need signed commits before we can review.

HalisCz:
Hi @yhakbar,

Signed.

docs/src/content/docs/03-features/07-caching/04-cas/02-git.mdx
```diff
@@ -21,6 +22,18 @@ Git is the most common source Terragrunt fetches and has the richest support in
...
 
+## Source URL parameters
+
+A Git source URL may carry go-getter query parameters that Terragrunt consumes to resolve the source rather than passing them through to `git`.
+
+- **`ref`** selects the revision — a branch, tag, or commit — to resolve and check out, for example `...vpc.git?ref=v5.21.0`.
+
+<Since version="1.1.2">
+
+- **`depth`** requests a shallow clone and overrides the ambient `--cas-clone-depth` for that source, for example `...vpc.git?ref=v5.21.0&depth=1`. Like `ref`, it is a go-getter parameter rather than a native Git URL parameter, so the CAS strips it from the URL before invoking `git` — otherwise `git` would treat a trailing `?depth=1` as part of the repository name and reject the clone. Only positive integers take effect; other values fall back to the ambient depth. Because the ambient default is already `1` (a shallow clone of the resolved commit), `?depth=1` only changes behavior when the ambient depth has been widened (for example `--cas-clone-depth=-1` for full history).
```
yhakbar: This behavior violates the general pattern of CLI vs configuration precedence the rest of the Terragrunt CLI follows. CLI arguments beat configuration everywhere else and that same pattern should be observed here.

HalisCz: Agreed — fixed in `8c4a9460`. `--cas-clone-depth` now wins.

The URL `depth` is still **stripped** before `git` is invoked (that's the actual #6512 bug — `git` rejects a trailing `?depth=1` as part of the repository name), but its value is no longer applied. Since the flag always carries a value (default `1`), CLI-beats-configuration means the source URL's `depth` never overrides it.

- `cas.StripGitURLParams` / `getter.GitURLParams` dropped their `depth` return.
- The e2e test that pinned the old precedence now pins the new one: ambient full history (`--cas-clone-depth=-1`) against `?depth=1` must leave no `shallow` marker in the git store. Verified it's discriminating — flipping the ambient to `1` does produce the marker.
- Docs and changelog updated to say `depth` is accepted for go-getter compatibility and stripped, with the depth itself governed by the flag.

HalisCz:
@yhakbar so what do you think?

yhakbar:
@HalisCz I pushed up a fix-up to your branch, but there's an unsigned commit in this PR that prevents me from running CI against it:
https://github.com/gruntwork-io/terragrunt/pull/6513/changes/8c4a94605fc367db4a9dcb16364a5407931d6633

Please rewrite that commit so that we can run CI against this PR.

HalisCz:
@yhakbar Heads-up: I force-pushed this branch to fix the one unsigned commit (8c4a946). Since the signature is part of the commit object, signing it changed its ID and everything above it.

yhakbar:
@HalisCz do you mind updating the target release to the next patch release?

HalisCz:
@yhakbar I've restored your two commits (authorship preserved, re-signed under my key since the commit IDs changed) and merged the latest main — the changelog entry now targets v1.1.4. All commits on the branch are signed now, so CI should be able to run.

yhakbar (approved):

denis256 (approved):

PR #6527, opened 2026-07-22
fix: Fixing report path prefix trim

yhakbar:
## Description

Fixes the path prefix trim done for the report file.

## TODOs

- [ ] Update the docs.
- [ ] This change is backwards compatible.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed report name generation for paths that share a prefix with the working directory but are not truly nested.
  * Report paths are now shortened only when they are genuinely located under the working directory; sibling (non-nested) paths keep their full names.

* **Documentation**
  * Added a changelog entry describing the corrected path handling.

### DIFF
--- a/internal/report/writer.go
+++ b/internal/report/writer.go
@@ -452,7 +452,10 @@ func nameOfPath(path string, workingDir string) string {
 		return filepath.Base(path)
 	}
 
-	prefix := workingDir + string(os.PathSeparator)
+	prefix := workingDir
+	if !strings.HasSuffix(prefix, string(os.PathSeparator)) {
+		prefix += string(os.PathSeparator)
+	}
 
 	if !strings.HasPrefix(path, prefix) {
 		return path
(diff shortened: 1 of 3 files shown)


internal/report/writer.go
```diff
@@ -448,22 +448,17 @@ func WriteSchema(w io.Writer) error {
...
-	// If the path is the same as the working directory,
-	// return the base name of the path.
 	if path == workingDir {
 		return filepath.Base(path)
 	}
 
-	// If the path is not a subdirectory of the working directory,
-	// return the path as is.
-	if !strings.HasPrefix(path, workingDir) {
+	prefix := workingDir + string(os.PathSeparator)
```
denis256: Could we append the separator only when it is missing? For root working directories (/ or C:\), this creates // or C:\\, so nested run paths no longer match, and reports emit absolute names

denis256 (approved):

PR #6530, opened 2026-07-23
fix: Fixing stack dependency mock outputs

### LINKED ISSUE #6529: `mock_outputs` not resolving correctly with stack dependencies and `--dependency-fetch-output-from-state`
## Describe the bug

`mock_outputs` are ignored when a `dependency` block resolves aggregated outputs across a whole stack directory (`config_path` pointing at a `terragrunt.stack.hcl`) and one of that stack's units has no remote state yet.

Instead of falling back to the mock, the run hard-fails while reading the unit's
state from S3:

```
resolving dependency "networking" outputs: failed to collect stack unit outputs for .../networking/terragrunt.stack.hcl: stack unit vpc output fetch failed: fetching dependency state from s3://<bucket>/.../vpc/terraform.tfstate: operation error S3: GetObject ... NoSuchKey
```

A regular unit-to-unit `dependency` handles this fine: a missing S3 key falls back to `mock_outputs`. Stack-directory dependencies don't, so an auto-included stack dep against an unapplied stack fails on the missing key even though a mock is declared.

## Steps To Reproduce

1. A stack whose units use an S3 backend, referenced as a whole via a dependency:

    ```hcl
    # live/terragrunt.stack.hcl
    stack "networking" {
      source = "../stacks/networking"
      path   = "networking"
    }

    unit "app" {
      source = "../units/app"
      path   = "app"

      autoinclude {
        dependency "networking" {
          config_path  = stack.networking.path
          mock_outputs = { vpc = { vpc_id = "mock-vpc-id" } }
        }
        inputs = { vpc_id = dependency.networking.outputs.vpc.vpc_id }
      }
    }
    ```

2. `terragrunt stack generate`.
3. `terragrunt plan --dependency-fetch-output-from-state` on the `app` unit.

The plan fails on the S3 `NoSuchKey` instead of using `mock_outputs`.

## Expected behavior

A stack-directory dependency should honor `mock_outputs` when a unit's state doesn't exist yet, the same way a single-unit dependency already does.

## Must haves

## Nice to haves

- [ ] Terminal output
- [ ] Screenshots

## Versions

- Terragrunt version: v1.1.1
- OpenTofu/Terraform version: any
- Environment details (Ubuntu 20.04, Windows 10, etc.): S3 backed state

## Additional context

Root cause is in `collectStackUnitOutputs` (`pkg/config/dependency.go`): it returns the fetch error directly, without the `isAwsS3NoSuchKey` tolerance the single-unit path in `getTerragruntOutput` has, so the mock fallback in `getTerragruntOutputIfAppliedElseConfiguredDefault` is never reached.

yhakbar:
## Description

Fixes #6529.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Stack-based dependencies now honor configured mock outputs when fetching remote-state outputs.
  * Partially applied stacks can combine real outputs with mocks for units without state.
  * Commands provide clearer errors when mocks are unavailable, restricted, or incorrectly formatted.

* **Documentation**
  * Added changelog documentation describing improved stack dependency mock-output handling.

### DIFF
--- a/pkg/config/dependency.go
+++ b/pkg/config/dependency.go
@@ -912,11 +912,12 @@ func collectStackUnitOutputs(
 
 		jsonBytes, err := getOutputJSONWithCaching(ctx, pctx, l, unitConfigPath)
 		if err != nil {
-			if !shouldFallBackToMockOutputs(pctx, err) {
+			if !shouldFallBackToMockOutputs(pctx, err) ||
+				!dependencyConfig.shouldReturnMockOutputs(pctx) {
 				return nil, fmt.Errorf("stack unit %s output fetch failed: %w", unit.Name, err)
 			}
 
-			if mock, ok := unitMockOutput(dependencyConfig, unit.Name, pctx); ok {
+			if mock, ok := unitMockOutput(dependencyConfig, unit.Name); ok {
 				unitOutputs[unit.Name] = mock
 				continue
 			}
@@ -951,23 +952,29 @@ func collectStackUnitOutputs(
 	return unitOutputs, nil
 }
 
-// unitMockOutput returns the mock output for a named stack unit from the dependency's mock_outputs,
-// when mocks are allowed for the current command and the unit has a mock declared. It lets a
-// partially applied stack resolve: applied units contribute real outputs while unapplied ones fall
-// back to their mock.
+// unitMockOutput returns the mock declared for a named stack unit in the dependency's mock_outputs.
+// It lets a partially applied stack resolve: applied units contribute real outputs while unapplied
+// ones fall back to their mock.
 //
-// ok is false when mocks are disallowed, absent, or not keyed by unit name.
-func unitMockOutput(dep *Dependency, unitName string, pctx *ParsingContext) (cty.Value, bool) {
-	if dep.MockOutputs == nil || !dep.shouldReturnMockOutputs(pctx) {
+// Callers are responsible for checking that mocks are allowed for the current command. ok is false
+// when mock_outputs is absent, isn't keyed by unit name, or isn't a map or object.
+func unitMockOutput(dep *Dependency, unitName string) (cty.Value, bool) {
+	if dep.MockOutputs == nil {
 		return cty.NilVal, false
 	}
 
 	mock := *dep.MockOutputs
-	if mock.IsNull() || !mock.Type().IsObjectType() || !mock.Type().HasAttribute(unitName) {
+	if mock.IsNull() || !mock.IsKnown() {
 		return cty.NilVal, false
 	}
 
-	return mock.GetAttr(unitName), true
+	if mockType := mock.Type(); !mockType.IsObjectType() && !mockType.IsMapType() {
+		return cty.NilVal, false
+	}
+
+	unitMock, ok := mock.AsValueMap()[unitName]
+
+	return unitMock, ok
 }
 
 // tryGetStackOutput checks if targetConfigPath points to a stack directory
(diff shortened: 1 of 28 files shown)


pkg/config/dependency.go
```diff
@@ -872,7 +899,22 @@ func collectStackUnitOutputs(
 
 		jsonBytes, err := getOutputJSONWithCaching(ctx, pctx, l, unitConfigPath)
 		if err != nil {
-			return nil, fmt.Errorf("stack unit %s output fetch failed: %w", unit.Name, err)
+			if !shouldFallBackToMockOutputs(pctx, err) {
```
denis256: Since mocks may be disallowed for the current command, skipping the missing unit here may return an incomplete dependency object instead of the original fetch error

pkg/config/dependency.go
```diff
@@ -896,6 +938,25 @@ func collectStackUnitOutputs(
...
+// back to their mock.
+//
+// ok is false when mocks are disallowed, absent, or not keyed by unit name.
+func unitMockOutput(dep *Dependency, unitName string, pctx *ParsingContext) (cty.Value, bool) {
+	if dep.MockOutputs == nil || !dep.shouldReturnMockOutputs(pctx) {
+		return cty.NilVal, false
+	}
+
+	mock := *dep.MockOutputs
+	if mock.IsNull() || !mock.Type().IsObjectType() || !mock.Type().HasAttribute(unitName) {
```
denis256: Since mock_outputs can evaluate to a cty.Map, this object-only check may silently reject valid mocks such as

```hcl
  mock_outputs = tomap({
    vpc = {
      vpc_id = "mock-vpc-id"
    }
  })
```

denis256 (dismissed):

pkg/config/dependency.go
```diff
@@ -922,6 +939,31 @@ func collectStackUnitOutputs(
...
+	if dep.MockOutputs == nil {
+		return cty.NilVal, false
+	}
+
+	mock := *dep.MockOutputs
+	if mock.IsNull() || !mock.IsKnown() {
+		return cty.NilVal, false
+	}
+
+	if mockType := mock.Type(); !mockType.IsObjectType() && !mockType.IsMapType() {
```
denis256, line 978: Hm, if a user specifies a non-map type for `mock_outputs`, Terragrunt falls back to skipping all unapplied units with warnings, hiding config errors behind generic HCL evaluation errors

pkg/config/dependency.go
```diff
@@ -898,7 +899,23 @@ func collectStackUnitOutputs(
 
 		jsonBytes, err := getOutputJSONWithCaching(ctx, pctx, l, unitConfigPath)
 		if err != nil {
-			return nil, fmt.Errorf("stack unit %s output fetch failed: %w", unit.Name, err)
+			if !shouldFallBackToMockOutputs(pctx, err) ||
+				!dependencyConfig.shouldReturnMockOutputs(pctx) {
```
denis256, line 916: Possible nil reference for dependencyConfig

pkg/config/dependency.go
yhakbar: Added validation for `mock_outputs` type.

pkg/config/dependency.go
yhakbar: I don't think it's necessarily possible for it to be nil here, but I'm open to adding an explicit assertion later.

denis256 (approved):

PR #6546, opened 2026-07-26
fix: handling deleted dependencies in filter

### LINKED ISSUE #6540: `run --all` crashes with "does not contain a terragrunt.hcl file" when a live unit depends on a unit deleted in the same diff — while `find`/`list` report the same filter as healthy
**Describe the bug**

When a diff deletes a unit's directory, and some *other*, entirely unmodified live unit still has a `dependency` block pointing at that now-deleted directory, `terragrunt run --all` with a git-range filter crashes the **entire invocation** with:

```
ERROR You attempted to run terragrunt in a folder that does not contain a terragrunt.hcl file. Please add a terragrunt.hcl file and try again.

Path: "<repo>/dep/terragrunt.hcl"
```

Critically, `terragrunt find` / `terragrunt list` **do not** surface this at all — run against the exact same filter, they report a clean, healthy-looking result (both units, no errors, no duplicates). This makes the failure surprising and hard to diagnose in CI: a pre-flight `find` used to decide "is there anything to run" reports everything is fine, and then the real `run --all` crashes moments later with an error that doesn't obviously point at the dependency that was actually deleted (the path shown is the *dependency's* path, not the unit whose config actually has the problem).

I believe this is because `find`/`list` use `WithSuppressParseErrors` internally (so a dependency that can't be resolved is silently dropped from that command's output), while `run --all`'s relationship-resolution phase does not suppress the same error and hard-fails instead.

**Steps to reproduce**

Minimal, self-contained — no external repo needed, just Bash + git + Terragrunt v1.1.1:

```bash
mkdir tg-repro && cd tg-repro
git init -q && git config user.email test@example.com && git config user.name Test

cat > root.hcl <<'EOF'
remote_state {
  backend = "local"
  generate = { path = "backend.tf", if_exists = "overwrite" }
  config = { path = "${get_repo_root()}/.state/${path_relative_to_include()}/terraform.tfstate" }
}
EOF

mkdir -p tf/noop
cat > tf/noop/main.tf <<'EOF'
variable "name" { type = string, default = "noop" }
output "name" { value = var.name }
EOF

mkdir -p dep consumer
cat > dep/terragrunt.hcl <<'EOF'
include "root" { path = find_in_parent_folders("root.hcl") }
terraform { source = "../tf/noop" }
inputs = { name = "dep" }
EOF

cat > consumer/terragrunt.hcl <<'EOF'
include "root" { path = find_in_parent_folders("root.hcl") }
terraform { source = "../tf/noop" }
dependency "dep" {
  config_path = "../dep"
  mock_outputs = { name = "mock" }
  mock_outputs_allowed_terraform_commands = ["plan", "destroy"]
}
inputs = { name = "consumer" }
EOF

git add -A && git commit -q -m base
BASE=$(git rev-parse HEAD)

# HEAD: delete "dep" only. "consumer" is completely untouched.
git rm -q -r dep
git add -A && git commit -q -m "delete dep"
HEAD=$(git rev-parse HEAD)

echo "--- find (reports healthy, no error, no duplicates) ---"
terragrunt find --no-filters-file --format=json --filter "...[${BASE}...${HEAD}]... | ./**"

echo "--- run --all -- plan (crashes) ---"
terragrunt run --all --no-filters-file --filter-allow-destroy \
  --filter "...[${BASE}...${HEAD}]... | ./**" -- plan -input=false -lock=false
```

**Actual output**

`find` (exit 0):
```json
[
  { "type": "unit", "path": "consumer" },
  { "type": "unit", "path": "dep" }
]
```

`run --all` (exit 1):
```
ERROR You attempted to run terragrunt in a folder that does not contain a terragrunt.hcl file. Please add a terragrunt.hcl file and try again.

Path: "<repo>/dep/terragrunt.hcl"
```

**Expected behavior**

Either:
- `run --all` should succeed and plan `consumer` for a normal `plan` and `dep` for `plan -destroy` (the same outcome `find` implies is possible), or
- if this combination is genuinely unsupported, `find`/`list` should surface the same error `run --all` hits, rather than reporting a healthy, matching result right before the real command crashes.

Right now, the two commands disagree about whether this filter is safe to run, and the one used for pre-flight decisions (`find`) is the one giving the wrong answer.

**Versions**

- Terragrunt version: v1.1.1
- OpenTofu version: v1.12.3
- Environment: Ubuntu 26.04 LTS, linux/amd64

**Additional context — possibly the same root cause, possibly separate**

With a third unit added to the reproduction (a unit that depends on `consumer` and is itself modified at HEAD, pulling `consumer` in a second time as *its* dependency, in addition to `consumer` already being pulled in as a dependent of the deleted `dep`), `terragrunt find --format=json` reports `consumer` (and the modified unit) **twice**, from what looking at `--filter/git`'s "How it works" docs (https://docs.terragrunt.com/features/filter/git#how-it-works) appears to be two different worktree contexts — before the crash above even gets a chance to occur. I haven't fully traced whether this shares the exact code path with the crash above, so I'm not asserting it's the same bug, but the shape (a unit reachable both as a "dependent of something deleted" and as a "dependency of something live and changed" ends up resolved/discovered against two different filesystem roots) looks related enough that I wanted to flag it here rather than open a second, possibly-duplicate issue blind. Happy to split it out into its own issue with its own minimal repro if you'd rather triage them separately.

I searched the tracker first and the closest related issues I found were #5748, #6472, #6319, #6451, #5530, and #5598 — all closed/fixed, but each for a different specific mechanism (CWD-vs-worktree dependent discovery, a missing-working-directory invariant on filter-excluded components, and an `Args`-slice-mutation dropping `-destroy`, respectively). None of them describe this specific find/run-all divergence on a live unit's dangling dependency, so I believe this is a new report — apologies in advance if I've missed a duplicate.

denis256:
## Description

`run --all` with a git-range filter crashed with "does not contain a terragrunt.hcl file" when the diff deleted a unit's directory while a live unit still referenced it via a `dependency` block — while `find`/`list` reported the same filter as healthy

Fixes #6540

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `run --all` with Git-range filters crashing when a deleted unit is still referenced by another unit.
  * Live units now continue processing, optionally using configured mock outputs.
  * Deleted units still receive destroy plans when destroy filtering is enabled.
  * Outside Git filters, missing dependency configurations continue to report errors as expected.

* **Documentation**
  * Added guidance explaining behavior for deleted units with remaining dependency references.
  * Added a v1.1.2 changelog entry for the fix.

### DIFF
--- a/docs/src/content/docs/03-features/08-filter/06-git.mdx
+++ b/docs/src/content/docs/03-features/08-filter/06-git.mdx
@@ -126,6 +126,12 @@ This is a safeguard to prevent unintended destruction of infrastructure.
 
 </Aside>
 
+<Aside type="note" title="Dependencies deleted in the diff">
+
+When the diff deletes a unit's directory while a live unit still references it via a `dependency` block, the dangling reference does not fail the run. The live unit runs normally, falling back to `mock_outputs` where configured, and the deleted unit still receives its destroy plan from the `from` side of the diff when `--filter-allow-destroy` is set.
+
+</Aside>
+
 ## The `--filter-affected` flag
 
 For the common use case of comparing the default branch (typically `main`) with `HEAD`, you can use the `--filter-affected` flag as a convenient shorthand:
--- a/docs/src/data/changelog/v1.1.2/run-all-git-filter-deleted-dependency-crash.mdx
+++ b/docs/src/data/changelog/v1.1.2/run-all-git-filter-deleted-dependency-crash.mdx
@@ -1,0 +1,10 @@
+---
+version: "v1.1.2"
+category: "bug-fixes"
+---
+
+#### `run --all` no longer crashes when a Git filter's diff deletes a unit another unit depends on
+
+Running `run --all` with a Git-range filter (for example `...[main...HEAD]...`) crashed the entire invocation when the diff deleted a unit's directory while another live unit still had a `dependency` block pointing at it. The error reported the deleted dependency's path ("does not contain a terragrunt.hcl file") rather than the unit referencing it, and `find`/`list` reported the same filter as healthy right before the crash, making the failure hard to diagnose in CI.
+
+Dependency discovery in a Git filter context now treats a dependency whose config was deleted in the diff the same way `find` and `list` already did: the reference is skipped instead of failing the run. The live unit runs normally (falling back to `mock_outputs` where configured), and the deleted unit still receives its destroy plan from the `from` side of the diff when `--filter-allow-destroy` is set. Outside of Git filters, a dangling `config_path` keeps failing the run as before.
--- a/internal/discovery/helpers.go
+++ b/internal/discovery/helpers.go
@@ -127,6 +127,33 @@ func isExternal(workingDir string, componentPath string) bool {
 	return relPath == ".." || strings.HasPrefix(relPath, ".."+string(filepath.Separator))
 }
 
+// inGitFilterContext reports whether this discovery evaluates Git filter expressions,
+// either directly or as a worktree sub-discovery (identified by a Git ref on its
+// discovery context).
+func (d *Discovery) inGitFilterContext() bool {
+	if len(d.gitExpressions) > 0 {
+		return true
+	}
+
+	return d.discoveryContext != nil && d.discoveryContext.Ref != ""
+}
+
+// skipMissingDependencyConfig reports whether a parse failure for a dependency reached
+// during relationship or graph traversal should be skipped instead of failing discovery.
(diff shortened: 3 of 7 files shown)


yhakbar:
We need to clean this up. We don't want find/list to break when configurations are invalid. It would slow them down, and not be the most useful anyways. We _do_ want `hcl validate` (maybe with a flag or something) to break when configurations are invalid. We might want to surface a tip when appropriate to help users navigate this.

PR #6605, opened 2026-08-03
chore: Adding expansion blocks to the configs that accept expansion

yhakbar:
## Description

Adds expansion block to the `dependency`, `unit` and `stack` blocks.

Isn't used yet, but the block and requisite `Remain` is now registered.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

- **New Features**
  - Added support for `expansion` blocks on dependencies, units, and stacks.
  - Configurations can now create multiple instances using count-based or collection-based expansion.
  - Expansion instance keys are available for identifying generated instances.

- **Bug Fixes**
  - Invalid expansion attributes are rejected with validation errors.
  - Expansion metadata is protected from unintended configuration assignment and excluded from dependency output.

### DIFF
--- a/pkg/config/dependency.go
+++ b/pkg/config/dependency.go
@@ -62,6 +62,8 @@ type dependencyOutputCache struct {
 }
 
 type Dependency struct {
+	Expansion *hclparse.ExpansionBlock `hcl:"expansion,block"`
+
 	ConfigPath                          cty.Value  `hcl:"config_path,attr"                             cty:"config_path"`
 	Enabled                             *bool      `hcl:"enabled,attr"                                 cty:"enabled"`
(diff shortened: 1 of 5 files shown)


pkg/config/hclparse/expansion.go
denis256, line 342: Boolean sets in for_each like `toset([true, false])` are rejected

pkg/config/dependency.go
```diff
@@ -62,6 +62,8 @@ type dependencyOutputCache struct {
 }
 
 type Dependency struct {
+	Expansion *hclparse.ExpansionBlock `hcl:"expansion,block"`
```
denis256, line 65: Most probably should be updated Dependency.DeepMerge() and other merge functions to handle includes

denis256 (dismissed):

denis256 (approved):

PR #6611, opened 2026-08-04
fix: resolve a unit's real config path in checkUnitVersionConstraints's fallback parse

### LINKED ISSUE #6610: run --all crashes with "does not contain a terragrunt.hcl file" on a live, non-deleted, shared dependency — root cause: missing filepath.Join in checkUnitVersionConstraints's fallback parse
**Terragrunt version:** v1.1.1 and v1.1.2 (confirmed present in both; latest release as of this report)
**OpenTofu version:** 1.12.x
**OS:** Linux (GitHub Actions `ubuntu-latest` runner, and reproduced separately on a local Linux dev machine)

### Summary

This is the same error text as #6540 and my own earlier #6554 (closed as a duplicate of #6540), but a **distinct root cause** in a different package — not a dependency deleted in the diff, and not fixed by #6546. I want to flag that clearly up front so this doesn't get folded into #6540 again: #6554 genuinely was a different bug, and here is the actual root cause with a verified fix.

`terragrunt run --all -- plan`, scoped with a git-range `--filter`, intermittently crashes with:

```
ERROR failed to parse config for unit <path>: You attempted to run terragrunt in a folder that does not contain a terragrunt.hcl file. Please add a terragrunt.hcl file and try again.

Path: "terragrunt.hcl"
```

on a unit whose `terragrunt.hcl` genuinely exists, unmodified, on both sides of the diff — nothing is deleted anywhere. `terragrunt find`/`list` with the identical filter succeed cleanly. In every occurrence I've observed, the crashing unit is a **dependency shared by several other units** in the same run (a "hub" dependency) — never an isolated, single-consumer unit.

### Root cause (confirmed via source + `go build -race` + a verified one-line fix)

`internal/runner/runnerpool/builder_helpers.go`, `checkUnitVersionConstraints()`:

```go
unitConfig := unit.Config()

// This is almost definitely already parsed, but we'll check just in case.
if unitConfig == nil {
    configCtx, pctx := configbridge.NewParsingContext(ctx, l, unitOpts)
    pctx = pctx.WithVenv(v).WithDecodeList(
        config.TerragruntVersionConstraints,
        config.FeatureFlagsBlock,
    )

    var err error

    unitConfig, err = config.PartialParseConfigFile(
        configCtx,
        pctx,
        l,
        unit.ConfigFile(),   // <-- bug: bare basename, not joined with unit.Path()
        nil,
    )
    if err != nil {
        return fmt.Errorf("failed to parse config for unit %s: %w", unit.DisplayPath(), err)
    }
}
```

`unit.ConfigFile()` returns a **bare basename** (e.g. `"terragrunt.hcl"`) by design — that's exactly what it's set to, both by `component.NewUnit()`'s constructor default (`configFile: config.DefaultTerragruntConfigPath`) and by `createComponentFromPath()` in `internal/discovery/helpers.go` (`unit.SetConfigFile(base)`, where `base := filepath.Base(path)`). Every other call site that needs the actual file path joins it with `unit.Path()` first — e.g. `internal/runner/runnerpool/runner.go`'s `BuildUnitOpts()`:

```go
configPath := unit.Path()
if !strings.HasSuffix(configPath, ".hcl") && !strings.HasSuffix(configPath, ".json") {
    fileName := config.DefaultTerragruntConfigPath
    if unit.ConfigFile() != "" {
        fileName = unit.ConfigFile()
    }
    configPath = filepath.Join(unit.Path(), fileName)
}
```

`checkUnitVersionConstraints`'s fallback is the one call site that doesn't. `config.PartialParseConfigFile` immediately does:

```go
fileInfo, err := os.Stat(configPath)
if err != nil {
    if errors.Is(err, fs.ErrNotExist) {
        return nil, TerragruntConfigNotFoundError{Path: configPath}
    }
    ...
}
```

`os.Stat("terragrunt.hcl")` resolves against the **process's current working directory**, not the unit's own directory. Under a git-range `--filter`, that cwd is never any individual unit's directory, so the stat always fails for this code path, exactly reproducing the reported error text and the bare `Path: "terragrunt.hcl"`.

**Why this only fires sometimes, and only on shared/hub dependencies:** the fallback only runs "if `unit.Config()` is nil" — the comment above it says "almost definitely already parsed, but we'll check just in case." In practice, a unit reached only as a **dependency of multiple sibling units** (via `internal/discovery/phase_relationship.go`'s `dependencyToDiscover()`, which creates it as a lightweight placeholder via `component.NewUnit(path)` rather than through the main filesystem-discovery walk) can still have a nil `Config()` at the point `checkVersionConstraints`'s own concurrent pass reaches it, depending on discovery ordering/timing. A unit discovered only once, directly, by the filesystem walk, gets its config eagerly populated and never hits this fallback at all. This is why the crash always lands on a shared dependency and never an isolated unit.

### Confirmed fix

```diff
     unitConfig, err = config.PartialParseConfigFile(
         configCtx,
         pctx,
         l,
-        unit.ConfigFile(),
+        filepath.Join(unit.Path(), unit.ConfigFile()),
         nil,
     )
```

(`filepath` is already imported in this file.)

**Verified empirically**, against a private ~40-unit OpenTofu/Terragrunt monorepo where this fires reliably under CI's real git-range filter (real S3 `remote_state` backend, GitHub Actions `ubuntu-latest`, one dependency referenced by 9 sibling units):

| Build | Crashes / 20 runs | Data races reported |
|---|---|---|
| v1.1.1 (release, unmodified) | 13/20 | 20/20 (see note below) |
| v1.1.2 (release, unmodified) | 14/20 | 0/20 |
| v1.1.2 + this one-line patch | 0/25 | 0/25 |

All builds run with `go build -race` for the first two rows; the patched row is the identical v1.1.2 source with only the diff above applied.

### A separate, already-fixed race — ruled out, not the cause

While investigating, `go build -race` on the pinned v1.1.1 caught a real, distinct data race in `internal/discovery/phase_relationship.go`'s `dependencyToDiscover()`: `Unit.SetDiscoveryContext()`/`SetExternal()` were called unsynchronized on a dependency object that can be shared by multiple concurrent discoverers. I traced this and found it's already fixed in v1.1.2 by gating that mutation on `created` (only the goroutine that actually registers the component touches it). I want to flag this clearly as **ruled out** as the cause of the crash above, since it's easy to conflate the two: v1.1.2 has zero race reports across 20 runs yet still crashes 14/20 times with the identical error text — proof the two are independent. Mentioning it mainly so it isn't rediscovered and mistakenly credited as "the" fix for this issue.

### Reproduction attempts (both negative — noting this for anyone else who tries)

I built two synthetic fixtures matching the shape as closely as I could (real git repo, real `remote_state { backend = "local" }`, a git-range `--filter`, a hub unit referenced by many sibling `dependency` blocks, a second commit touching the diff without deleting anything):

- 1 hub + 15 consumers (16 units total): 0/20 crashes
- 1 hub + 15 consumers + 30 unrelated noise units (46 units total): 0/20 crashes

Neither reproduced, over 20 runs each, on the same machine (16 cores) where the real monorepo reproduces at ~65-70%. This matches what #6554 and the unrelated #6002 both independently found: this class of bug resists small synthetic fixtures and appears to depend on real-repo scale/timing (git-worktree materialization cost, total discovery graph size, or both) in a way a quick synthetic repro doesn't capture. I'm not including a "steps to reproduce" script for that reason — the root cause above is verifiable by inspection (compare the flagged call site against every other `unit.Path()`/`unit.ConfigFile()` join in the codebase) and the fix is verified against the actual failing environment, even though I can't hand you a portable trigger.

### Expected vs actual

Expected: `run --all` succeeds (or fails with an accurate error) on a live, unmodified, shared dependency.
Actual: intermittent hard crash claiming a missing config file, using a bare relative path that was never joined with the unit's actual directory.

### Related issues

- #6540 / #6546 — different bug (a dependency **deleted** in the diff), different package (`internal/discovery`'s graph/relationship phases), unaffected by this fix and doesn't affect this bug.
- #6554 — my own earlier report of this exact symptom, closed as a duplicate of #6540 for lack of repro steps. This issue is the follow-up with the actual root cause; the two are not duplicates of each other.

luis-guimaraes-optimuse:
## Description

Fixes #6610.

`checkUnitVersionConstraints`'s fallback parse (used when a unit's `Config()` is still `nil`
when `checkVersionConstraints`'s pass reaches it — in practice, a dependency discovered only
as a lightweight placeholder by the relationship-discovery phase rather than the main
filesystem walk) passed `unit.ConfigFile()` — a bare basename by design — directly to
`config.PartialParseConfigFile` instead of joining it with `unit.Path()` first, the way every
other caller in the codebase does (e.g. `BuildUnitOpts` in this same file).
`PartialParseConfigFile`'s `os.Stat` then resolves the bare name against the process's cwd
instead of the unit's own directory, and fails with `TerragruntConfigNotFoundError` even
though the file genuinely exists on disk — reproducing the reported
`does not contain a terragrunt.hcl file` crash on a live, unmodified unit.

This is a one-line fix:

```diff
 		unitConfig, err = config.PartialParseConfigFile(
 			configCtx,
 			pctx,
 			l,
-			unit.ConfigFile(),
+			filepath.Join(unit.Path(), unit.ConfigFile()),
 			nil,
 		)
```

Included is a regression test that constructs exactly the shape a shared/hub dependency has
when it reaches this fallback (a real on-disk `terragrunt.hcl`, `Config()` never populated,
`ConfigFile()` at its bare-basename default) and calls `checkVersionConstraints` directly.
It's deterministic — no timing/concurrency dependency at all, unlike the original bug report,
since the missing join fails unconditionally whenever this fallback runs on any unit, not only
under a git-range filter or a shared-dependency race. It fails against unpatched `main` with
the exact reported error text, and passes with this fix.

Separately, #6610 also verified (via `go build -race`) that the intermittent nature of the
*original* bug report is caused by a nil `Config()` interacting with a real, already-fixed race
in `dependencyToDiscover` (the `created &&` guard landed in v1.1.2) — that race is not what
this PR fixes, and fixing it alone (confirmed on v1.1.2) does not resolve this crash. See the
issue for the full investigation, including why a portable CLI-level repro couldn't be produced
despite the root cause and fix being unambiguous from the code and this test.

## Verification performed locally

```
$ go build ./...
(clean)

$ go test ./internal/runner/runnerpool/... -v
--- PASS: TestCheckVersionConstraints_UnparsedUnit_ResolvesConfigPathAgainstUnitDir (0.06s)
... (all other existing tests in the package pass, unmodified)
PASS

$ gofmt -l internal/runner/runnerpool/builder_helpers.go internal/runner/runnerpool/version_constraints_unparsed_unit_test.go
(no output)

$ go vet ./internal/runner/runnerpool/...
(clean)
```

Reverting just the one-line fix (keeping the new test) makes the new test fail with the exact
reported error text, confirming it actually exercises the bug rather than passing vacuously.

## TODOs

- [ ] Update the docs.

I haven't added a changelog entry or docs update (see `docs/src/data/changelog/`) — happy to add one in the style of #6546's if that's expected for a fix this size, just didn't want to assume the format without a maintainer nod first. I also haven't run this repo's full CI-equivalent locally, only the directly affected package's tests plus a full `go build`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved version-constraint checks for units with unparsed configuration files, ensuring configuration paths are resolved from the unit’s directory.

* **Tests**
  * Added regression coverage for version-constraint validation with temporary unit configurations.


denis256:
```
remote:   Found 1 violation:
remote: 
remote:   3b9eb76958599ea4fd12a73679c23d6d073e9eec
```

luis-guimaraes-optimuse:
@denis256 will sign the commits as soon as I can

luis-guimaraes-optimuse:
I signed the commit, hopefully that worked @denis256

luis-guimaraes-optimuse:
@denis256 Any update on when this can be merged? If it can be merged, if there are any specific things you need me to change let me know

denis256:
```
level=info msg="Execution took 5m19.178471096s"
internal/runner/runnerpool/version_constraints_unparsed_unit_test.go:1:9: package should be `runnerpool_test` instead of `runnerpool` (testpackage)
package runnerpool
        ^
1 issues:
* testpackage: 1
make: *** [Makefile:47: run-lint-incremental] Error 1
```

luis-guimaraes-optimuse:
Any update? @denis256

yhakbar:
Looks like we no longer need this PR as the underlying issue was resolved here, right?
https://github.com/gruntwork-io/terragrunt/pull/6728

Are we good to close this out?

luis-guimaraes-optimuse:
Yeah I just realized that, thanks

PR #6651, opened 2026-08-06
perf: Refactor default `--tf-path` resolution

yhakbar:
## Description

Moving `identifyDefaultWrappedExecutable` out of package initialization so that it's evaluated during `TerragruntOptions` construction instead, and replacing the `-version` call with a path lookup instead.

Removes the `tofu` invocation when it isn't necessary for commands like `find` and `list` and makes it so that we use the cheaper PATH lookup instead of calling `-version` on tofu.

Comparing `terragrunt --version` on main vs this PR:

```
Benchmark 1: main
  Time (mean ± σ):      55.2 ms ±   2.2 ms    [User: 49.4 ms, System: 21.2 ms]
  Range (min … max):    51.7 ms …  63.8 ms    52 runs

Benchmark 2: this PR
  Time (mean ± σ):      32.3 ms ±   1.4 ms    [User: 32.6 ms, System: 9.6 ms]
  Range (min … max):    30.3 ms …  39.9 ms    75 runs

Summary
  this PR ran
    1.71 ± 0.10 times faster than main
```

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **Performance**
  * Faster startup when no Terraform path is specified by checking for OpenTofu without launching it.
  * Automatically falls back to Terraform when OpenTofu is unavailable.
* **Bug Fixes**
  * Clearer errors are reported when a detected OpenTofu executable cannot run.
* **Documentation**
  * Added changelog guidance for automatic executable selection and overrides using `--tf-path` or `TG_TF_PATH`.

### DIFF
--- a/docs/src/data/changelog/v1.1.3/default-tf-path-lookup.mdx
+++ b/docs/src/data/changelog/v1.1.3/default-tf-path-lookup.mdx
@@ -1,0 +1,10 @@
+---
+version: "v1.1.3"
+category: "performance-improvements"
+---
+
+#### Faster startup when `--tf-path` is not set
+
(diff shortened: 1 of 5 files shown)


yhakbar:
@coderabbit review

denis256 (dismissed):

denis256 (approved):

PR #6696, opened 2026-08-18
fix: Reject oci:// sources with docker-style name suffixes instead of resolving latest

### LINKED ISSUE #6695: oci parsing of source URL does not include image:tag variant
## Describe the bug

An `oci://` source that pins a tag using the conventional `repo:tag` colon syntax (as
accepted by `docker`, `oras`, and OpenTofu's own `oci://` module source string) is not
parsed as a tag at all. The colon and everything after it become part of the repository
name, and the reference actually resolved silently falls back to the `latest` tag instead.

No error is raised. The getter will attempt to resolve `latest` against a repository name
that includes the stray `:tag` suffix, so it either fails with a registry-side "repository
not found" error (if `latest` legitimately doesn't exist) or, worse, silently resolves and
downloads whatever is tagged `latest` while the user believes they pinned a specific
version.

## Steps To Reproduce

Given a module published to an OCI registry at both a versioned tag and `latest`, e.g.:

```
oras push ghcr.io/example-org/modules/example-module:1.0.0 module.zip
oras push ghcr.io/example-org/modules/example-module:latest module.zip
```

Reference it from a `terragrunt.hcl` using the colon syntax:

```hcl
terraform {
  source = "oci://ghcr.io/example-org/modules/example-module:1.0.0"
}
```

With the `oci` experiment enabled, run `terragrunt plan`.

Observed: Terragrunt resolves and downloads whatever is tagged `latest`, not `1.0.0`,
with no warning that the `:1.0.0` suffix was ignored. If `latest` does not exist in the
repository, the run fails with an error about the repository/reference that gives no
indication the actual problem is the unsupported colon syntax.

Only the query-string form correctly pins the tag:

```hcl
terraform {
  source = "oci://ghcr.io/example-org/modules/example-module?tag=1.0.0"
}
```

## Expected behavior

Either:

- The colon syntax should be supported and parsed into the tag/digest reference (matching
  what `docker`, `oras`, and other OCI tooling accept), or
- If colon syntax is intentionally unsupported, `parseOCISource` should reject a
  repository path containing a colon with a clear error, instead of silently discarding
  the tag and defaulting to `latest`.

The current behavior — silently substituting `latest` for whatever tag was actually
requested — is the worst of the three options, since it can silently fetch and apply the
wrong module version with no error or warning.

## Where this lives in the code

- `internal/getter/oci.go`, `parseOCISource` (`oci.go:413-432`): builds
  `repositoryName` directly from the URL path via `SourceDirSubdir`, with no handling of
  an embedded `:tag`/`:digest` suffix.
- `internal/getter/oci.go`, `ociRefFromQuery` (`oci.go:439-470`): the *only* place a tag
  or digest is read, and it reads exclusively from the `?tag=`/`?digest=` query string.
  When neither is present it returns `ociDefaultTag` (`"latest"`, `oci.go:29`)
  unconditionally.
- `internal/getter/oci_auth.go`, `NewOCIRepositoryStore` (`oci_auth.go:106-117`): builds
  the ORAS repository reference as `registryDomain + "/" + repositoryName` and hands it to
  `remote.NewRepository`. ORAS's own reference parser *does* understand `repo:tag`
  syntax and will silently strip the tag back out when locating the repository, which is
  why the repository itself is still found correctly — only the tag portion is lost,
  masking the bug rather than surfacing it as a parse failure.

## Must haves

## Versions

- Terragrunt version: v1.1.3
- OpenTofu/Terraform version: 1.12.2 (opentofu)
- Environment details: macOS (Darwin), `oci` experiment enabled

## Additional context

Related to the `oci` experiment tracked in #4555.

denis256:
## Description

* Throw an explicit error in case of ":tag" URLs, same as in OpenTofu

Fixes #6695.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

- **Bug Fixes**
  - OCI sources using Docker-style `:tag` or `@digest` suffixes are now rejected instead of being misinterpreted.
  - Error messages provide the supported `?tag=` or `?digest=` format.
  - Added validation for OCI registry and repository names, including uppercase-name errors.
  - Explicit query parameters remain supported and take precedence.

- **Documentation**
  - Updated OCI usage guidance and changelog entries with the new syntax and validation behavior.


internal/getter/oci.go
```diff
@@ -421,6 +424,20 @@ func parseOCISource(srcURL *url.URL) (ociSourceCoordinates, error) {
 		return coords, ErrOCIMissingRepositoryName
 	}
 
+	// Validated by field, as tofu does, so a :tag suffix cannot ride through ORAS reference splitting into latest.
+	fields := registry.Reference{Registry: coords.registryDomain, Repository: coords.repositoryName}
+	if err := fields.Validate(); err != nil {
+		// A single %w keeps the whole message intact: the CLI renderer splits multi-wrapped errors into bare bullets.
+		if strings.ContainsAny(coords.repositoryName, ":@") {
+			return coords, fmt.Errorf(
+				"%w %q: pin a version with ?tag= or ?digest= instead of a name suffix: %s",
```
yhakbar: Instead of doing this, I think we're better off constructing an error type that captures the intended tag/digest when possible, and emitting the fully formed URL instead.

e.g. `oci://ghcr.io/acme/modules/vpc:1.0.0` is used, so you get an error with the string `oci://ghcr.io/acme/modules/vpc?tag=1.0.0`

internal/getter/oci.go
```diff
@@ -448,6 +459,91 @@ func parseOCISource(srcURL *url.URL) (ociSourceCoordinates, error) {
...
+	return OCIEmbeddedReferenceError{
+		RepositoryName:  coords.repositoryName,
+		SuggestedSource: ociQuerySource(coords.registryDomain, name, coords.subDir, queryKey, ref),
+	}
+}
+
+// cutOCIEmbeddedReference splits a docker-style "@digest" or ":tag" suffix off
+// a repository path, recognizing a suffix only when it satisfies the digest or
+// tag grammar and leaves a valid name, so the rewrite always parses on paste.
+func cutOCIEmbeddedReference(repositoryName string) (string, string, string, bool) {
```
yhakbar, line 483: We might want to return a struct instead of a four parameter return here.

internal/getter/oci_test.go
```diff
@@ -204,13 +204,14 @@ func TestOCIGetterGetErrors(t *testing.T) {
 			src:             "oci://127.0.0.1:5000/terraform-modules/vpc:1.0.0",
 			store:           newFakeStore(goodManifest, &goodDesc, zipBytes, &layer),
 			wantErrIs:       getter.ErrOCIInvalidRepositoryName,
-			wantErrContains: "pin a version with ?tag= or ?digest= instead of a name suffix",
+			wantErrContains: `"oci://127.0.0.1:5000/terraform-modules/vpc?tag=1.0.0"`,
```
yhakbar, line 207: We have the proper error type. We might as well match on error as and assert the value of the error fields instead of checking the string representation.

yhakbar (approved):

PR #5499, opened 2026-02-03
feat: Dependency fetch output from GCS state

joshmyers:
## Description
Support fetching dependency output directly from GCS state buckets, currently only S3 is supported.

## Why
We are hitting parallel issues with Terragrunt + Atlantis whereby states that have similar `dependencies` blocks are trying and failing to `init` the dependency in parallel, which looks like it's because dependency outputs are fetched via raw Terraform so don't use the Terragrunt cache? Either way, it would be extremely useful to be able to fetch the dependency states directly from GCS, so no need to `init` and `terraform output` which means users won't hit this contention and it's faster.

## TODOs

Read the [Gruntwork contribution guidelines](https://gruntwork.notion.site/Gruntwork-Coding-Methodology-02fdcd6e4b004e818553684760bf691e).

- [X] I authored this code entirely myself
- [X] I am submitting code based on open source software (e.g. MIT, MPL-2.0, Apache)]
- [X] I am adding or upgrading a dependency or adapted code and confirm it has a compatible open source license
- [X] Update the docs.
- [x] Run the relevant tests successfully, including pre-commit checks.
- [x] Include release notes. If this PR is backward incompatible, include a migration guide.

## Release Notes (draft)

Support fetching dependency output directly from GCS state buckets via `--dependency-fetch-output-from-state`

## Testing
I've built binaries locally using `make build` and successfully tested this functionality:
```
❯ TG_DEPENDENCY_FETCH_OUTPUT_FROM_STATE=true terragrunt plan --log-level debug
DEBU[0000] Terragrunt Version: 0.0.0
DEBU[0000] using cache key for version files: r01AJjVD7VSXCQk1ORuh_no_NRY
DEBU[0000] Running command: terraform -version
DEBU[0000] Engine is not enabled, running command directly in /foo/bar
**SNIP OUTPUT**
DEBU[0000] Fetching outputs directly from gs://$MY_STATE_BUCKET/foo/bar/default.tfstate prefix=[/foo/bar]
DEBU[0001] Retrieved output from foo/bar/terragrunt.hcl as json: {"dns_zone": "bla bla bla"}  using gcs bucket prefix=[/foo/bar]
```

Also added integration tests (didn't see any unit tests for it) which are lifted from the S3 equivalent test files. Should contributors be able to run these integration tests somewhere before raising PRs? 

## Summary by CodeRabbit

* **New Features**
  * Adds support for fetching dependency outputs from Google Cloud Storage (GCS) remote-state backends.
[...]

PR #6710, opened 2026-08-19
feat: remote state reading for GCP and Azure

denis256:
## Description

- Direct state reads for GCS, Azure
- Unsupported configs fall back to native
- Azure account keys resolved once per run
- Setup failures fail loud, not mocks

RFC: https://github.com/gruntwork-io/terragrunt/issues/4307

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **New Features**
  * Dependency outputs can now be read directly from S3, GCS, and Azure Storage state where supported.
  * Added Azure Storage support for direct state reads when the Azure backend experiment is enabled.
  * Unsupported configurations automatically fall back to the native output process.
  * Missing state continues to use configured mock outputs.

* **Bug Fixes**
  * Improved Azure credential handling, caching, error reporting, and environment detection.
  * Improved streaming and error handling when parsing remote state outputs.

* **Documentation**
  * Updated backend support, encryption limitations, fallback behavior, and version-specific guidance.
  * Added v1.1.4 changelog entries for GCS and Azure state reads.


joshmyers:
It looks like PR supersedes https://github.com/gruntwork-io/terragrunt/pull/5499

We've been running this alpha release for a while without issues but would like to upgrade to newer TG features including vuln fixes. Is this PR likely to merged any time soon given we need the fetching remote output directly from GCS?

yhakbar:
> It looks like PR supersedes #5499
> 
> We've been running this alpha release for a while without issues but would like to upgrade to newer TG features including vuln fixes. Is this PR likely to merged any time soon given we need the fetching remote output directly from GCS?

Hey, I re-opened your PR to try to give you credit, but it might be faster to just get this reviewed and add a line to the changelog to give you credit for the first pass at this, @joshmyers .

We should be able to get this in 1.1.4.

docs/src/data/changelog/v1.1.4/dependency-state-reads-azure.mdx
```diff
@@ -0,0 +1,12 @@
...
+category: "experiments-updated"
+---
+
+#### Read dependency outputs directly from GCS and Azure state
+
+The `dependency-fetch-output-from-state` experiment can now read dependency outputs directly from GCS and Azure Storage (`azurerm`) state, in addition to S3. This avoids initializing the dependency and running `tofu output` or `terraform output`.
+
+Azure direct reads require the `azure-backend` experiment as well. Unsupported backends and configurations requiring native-only authentication, endpoint, timeout, or Azure customer-provided-key behavior continue to use the native output path.
+
+When a dependency has no state yet, Terragrunt uses that dependency block's `mock_outputs`, as it already does for S3. A misconfiguration is not treated that way. When Azure direct reads resolve a storage account key through Azure Resource Manager, which is the case unless `access_key`, `sas_token`, or `use_azuread_auth` is set, a `resource_group_name`, `storage_account_name`, or `subscription_id` naming a resource that does not exist fails with an error naming those keys rather than substituting mock outputs.
```
yhakbar:
```suggestion
When a dependency has no state yet, Terragrunt uses that dependency block's `mock_outputs`, as it already does for S3. When Azure direct reads resolve a storage account key through Azure Resource Manager, which is the case unless `access_key`, `sas_token`, or `use_azuread_auth` is set, a `resource_group_name`, `storage_account_name`, or `subscription_id` naming a resource that does not exist fails with an error naming those keys rather than substituting mock outputs.
```

internal/remotestate/backend/azurerm/backend.go
```diff
@@ -145,6 +147,134 @@ func newBlobClient(ctx context.Context, l log.Logger, cfg *azurehelper.AzureConf
...
+
+	if !armCapable(cfg) || cfg.UseAzureADAuth {
+		return azurehelper.NewBlobClient(cfg)
+	}
+
+	keyed, err := cachedSharedKeyConfig(ctx, cfg)
+	if err != nil {
+		// Only a response that implicates the config or permissions earns the guidance.
+		if coordinateStateClientFailure(err) {
+			return nil, fmt.Errorf("%w: %w: %w", ErrStateClientSetup, ErrStateClientCoordinates, err)
```
yhakbar: Feels like `ErrStateClientSetup` should be a type.

internal/remotestate/backend/azurerm/backend.go
```diff
@@ -145,6 +147,134 @@ func newBlobClient(ctx context.Context, l log.Logger, cfg *azurehelper.AzureConf
...
+	if keyCache == nil {
+		return sharedKeyConfig(ctx, cfg)
+	}
+
+	cacheKey := sharedKeyCacheKey(cfg)
+	if cached, found := keyCache.configs.Get(ctx, cacheKey); found {
+		return cached, nil
+	}
+
+	// Hold the account's lock across the lookup so simultaneous first misses
```
yhakbar: Should we replace this with a singleflight? Seems like the right shape for it.

internal/remotestate/backend/azurerm/state_cache.go
```diff
@@ -0,0 +1,56 @@
...
+// still work, they just resolve a key each time.
+func WithStateClientCache(ctx context.Context) context.Context {
+	return context.WithValue(ctx, stateClientCacheKey{}, &stateClientCache{
+		configs: cache.NewCache[*azurehelper.AzureConfig](stateClientCacheName),
+		locks:   util.NewKeyLocks(),
+	})
+}
+
+// stateClientCacheFromContext returns the run's cache, or nil when unscoped.
+func stateClientCacheFromContext(ctx context.Context) *stateClientCache {
```
yhakbar: Should this ever be unscoped? I'm not sure we should have nil tolerance here.

pkg/config/dependency.go
```diff
@@ -1432,7 +1436,7 @@ func resolveOutputJSON(
 		)
 		l.Debugf("Falling back to terragrunt output.")
 
-		out, runErr := runTerragruntOutputJSON(ctx, pctx, l, targetConfig)
+		out, runErr := runTerragruntOutputJSON(ctx, pctx, l, targetConfig, nil)
```
yhakbar: This `nil` is a smell. There's probably a better design for this signature.

joshmyers:
> > It looks like PR supersedes #5499
> > We've been running this alpha release for a while without issues but would like to upgrade to newer TG features including vuln fixes. Is this PR likely to merged any time soon given we need the fetching remote output directly from GCS?
> 
> Hey, I re-opened your PR to try to give you credit, but it might be faster to just get this reviewed and add a line to the changelog to give you credit for the first pass at this, @joshmyers .
> 
> We should be able to get this in 1.1.4.

@yhakbar Agreed, don't think there is much point rebasing https://github.com/gruntwork-io/terragrunt/pull/5499 when this looks to do the same feature + for Azure. Happy for this to land if it's gonna be soon, else it's gonna have to be a fork 😭

pkg/config/dependency.go
```diff
@@ -1471,6 +1490,15 @@ func resolveOutputJSON(
 		targetConfig,
 		nil,
 	)
+
+	// extra_arguments env_vars are applied to the OpenTofu/Terraform process after
+	// IAM role assumption. Restore the auth-provider environment before STS so those
+	// variables do not unexpectedly become the source credentials, then apply them
+	// again below as the final override. The lightweight parse above still sees them,
+	// preserving support for remote_state expressions that read those variables.
+	clear(pctx.Venv.Env)
```
yhakbar: This tosses the user's `extra_arguments.env_vars`. We should move this to just above the `margedIAM` block on 1531.

pkg/config/dependency.go
```diff
@@ -1559,6 +1595,192 @@ func canGetRemoteState(remoteState *remotestate.RemoteState) bool {
...
+	}
+
+	if configured {
+		return value, true
+	}
+
+	return firstNonEmptyFromMap(env, envKeys...), true
+}
+
+func backendConfigBool(config backend.Config, key string) (bool, bool, bool) {
```
yhakbar: This should be a struct return, I think. I don't know what these bools mean.

pkg/config/dependency.go
```diff
@@ -1559,6 +1595,192 @@ func canGetRemoteState(remoteState *remotestate.RemoteState) bool {
...
+	default:
+		return false, true, false
+	}
+}
+
+func backendConfigBoolWithEnv(
+	config backend.Config,
+	key string,
+	envValue string,
+) (bool, bool) {
```
yhakbar: This should be a struct return, I think. I don't know what these bools mean.

pkg/config/dependency.go
```diff
@@ -1559,6 +1595,192 @@ func canGetRemoteState(remoteState *remotestate.RemoteState) bool {
...
+
+	return parsed, err == nil
+}
+
+func backendConfigStrictBoolWithEnv(
+	config backend.Config,
+	key string,
+	envValue string,
+	defaultValue bool,
+) (bool, bool) {
```
yhakbar: This should be a struct return, I think. I don't know what these bools mean.

pkg/config/dependency.go
```diff
@@ -1559,6 +1595,192 @@ func canGetRemoteState(remoteState *remotestate.RemoteState) bool {
...
+	return ""
+}
+
+// dependencyStateWorkspace mirrors OpenTofu/Terraform workspace selection: a non-empty
+// TF_WORKSPACE overrides the selection persisted in TF_DATA_DIR/environment, and an absent
+// or empty selection means the default workspace. Errors are returned so callers can keep
+// the native output path rather than guessing a state object.
+func dependencyStateWorkspace(pctx *ParsingContext, workingDir string) (string, error) {
+	if pctx.Venv == nil {
+		return "", errors.New("determining dependency workspace: virtual environment is required")
```
yhakbar: This should probably be a panic.

pkg/config/dependency.go
```diff
@@ -1559,6 +1595,192 @@ func canGetRemoteState(remoteState *remotestate.RemoteState) bool {
...
+	}
+
+	if workspace := pctx.Venv.Env["TF_WORKSPACE"]; workspace != "" {
+		if url.PathEscape(workspace) != workspace {
+			return "", fmt.Errorf("determining dependency workspace: invalid TF_WORKSPACE value %q", workspace)
```
yhakbar: This should probably be a typed error

pkg/config/dependency.go
```diff
@@ -1702,13 +1924,30 @@ func getTerragruntOutputJSONFromInitFolder(
 // - Run terraform init and terraform output
 // - Clean up folder once json file is generated
 // NOTE: terragruntOptions should be in the ctx of the targetConfig already.
+// directStateReader reads a dependency's outputs straight from its state.
+type directStateReader func(
```
yhakbar: This API is a little weird. It's not a reader in the `io.Reader` sense.

Would it be better to have a type like this:

```
type directStateBackend struct {
      supported func(*ParsingContext, *remotestate.RemoteState) bool
      read      func(context.Context, log.Logger, *ParsingContext, *remotestate.RemoteState, string) ([]byte, error)
}
```

And a global var like this:

```
var directStateBackends = map[string]directStateBackend{...}
```

pkg/config/dependency.go
```diff
@@ -1702,13 +1924,30 @@ func getTerragruntOutputJSONFromInitFolder(
 // - Run terraform init and terraform output
 // - Clean up folder once json file is generated
 // NOTE: terragruntOptions should be in the ctx of the targetConfig already.
+// directStateReader reads a dependency's outputs straight from its state.
```
yhakbar: Also, this comment should be separated from the comment above it so we get a proper godoc.

pkg/config/dependency_internal_test.go
```diff
@@ -1,9 +1,29 @@
 package config
```
yhakbar, line 1: Can we move this to a proper `config_test`?

pkg/config/dependency_internal_test.go
```diff
@@ -309,3 +329,1247 @@ func TestApplyExtraArgsEnvVarsForOutput(t *testing.T) {
 		})
 	}
 }
+
+//nolint:paralleltest // t.Setenv below clears the ambient environment and forbids t.Parallel.
+func TestShouldFetchDependencyOutputFromState(t *testing.T) {
```
yhakbar: Didn't we move these to venv?

pkg/config/dependency.go
```diff
@@ -1843,93 +2059,20 @@ func getTerragruntOutputJSONFromRemoteState(
...
-			if err != nil {
-				return fmt.Errorf(
-					"encoding outputs from dependency state at s3://%s/%s: %w",
-					bucket,
-					key,
-					err,
-				)
-			}
+// terraformStateOutputsJSON extracts the top-level outputs object from a Terraform/OpenTofu state file.
+func terraformStateOutputsJSON(stateBody []byte, location string) ([]byte, error) {
```
yhakbar: Could this take an `io.Reader` instead of `[]byte`?

The caller already has one from `open()` and [this](https://github.com/gruntwork-io/terragrunt/pull/6710/changes#diff-041967bd4dee9ad944b6ee2c9021a9be4554f5923ab34cb51439e01d281c723dR45) is the only reason we buffer the payload in memory first.

pkg/config/dependency_internal_test.go
```diff
@@ -309,3 +329,1247 @@ func TestApplyExtraArgsEnvVarsForOutput(t *testing.T) {
...
+		},
+		{
+			name:        "dependency experiment disabled",
+			backendName: gcsbackend.BackendName,
+			experiments: experiment.NewExperiments(),
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
```
yhakbar: It would be really nice if we could call `t.Parallel()` here.

pkg/config/dependency_internal_test.go
```diff
@@ -309,3 +329,1247 @@ func TestApplyExtraArgsEnvVarsForOutput(t *testing.T) {
...
+		})
+	}
+
+	assert.False(t, shouldFetchDependencyOutputFromState(&ParsingContext{}, nil))
+}
+
+func TestGCSDirectStateReadFallsBackWhenExecutableEnvironmentIsCleared(t *testing.T) {
+	const executableEnv = "GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES"
+
+	t.Setenv(executableEnv, "1")
```
yhakbar: Why do we need this? Can't we just rely on venv?

pkg/config/dependency_internal_test.go
```diff
@@ -309,3 +329,1247 @@ func TestApplyExtraArgsEnvVarsForOutput(t *testing.T) {
...
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			t.Parallel()
+
+			got, err := terraformStateOutputsJSON([]byte(testCase.state), "gs://bucket/state")
+			if testCase.wantErr != "" {
+				require.ErrorContains(t, err, testCase.wantErr)
```
yhakbar: Can't we check the type here?

internal/remotestate/backend/azurerm/backend.go
```diff
@@ -145,6 +147,135 @@ func newBlobClient(ctx context.Context, l log.Logger, cfg *azurehelper.AzureConf
...
+
+	if !armCapable(cfg) || cfg.UseAzureADAuth {
+		return azurehelper.NewBlobClient(cfg)
+	}
+
+	keyed, err := cachedSharedKeyConfig(ctx, cfg)
+	if err != nil {
+		// Only a response that implicates the config or permissions earns the guidance.
+		if coordinateStateClientFailure(err) {
+			err = fmt.Errorf("%w: %w", ErrStateClientCoordinates, err)
```
yhakbar: Shouldn't `ErrStateClientCoordinates` be a type?

internal/venv/venv.go
```diff
@@ -295,6 +301,25 @@ func (v *Venv) WithEnv(env map[string]string) *Venv {
 	return &c
 }
 
+// WithProcessEnv returns a copy of v with an immutable snapshot of the
+// environment visible to libraries that read the real process directly.
+func (v *Venv) WithProcessEnv(env map[string]string) *Venv {
```
yhakbar: Why are we doing this? We shouldn't need to.

pkg/config/dependency_state_azurerm.go
```diff
@@ -0,0 +1,660 @@
+package config
```
yhakbar, line 1: Could we not make this `config_test`?

pkg/config/dependency_state_gcs.go
```diff
@@ -0,0 +1,487 @@
...
+
+// gcsCredentialFileDirectStateReadSupported retains credential files whose
+// authentication does not launch or configure an external subject-token
+// provider from the Terragrunt process environment.
+func gcsCredentialFileDirectStateReadSupported(pctx *ParsingContext, filename string) bool {
+	if filename == "" {
+		return true
+	}
+
+	contents, err := vfs.ReadFile(pctx.Venv.FS, filename)
```
yhakbar: Again, can't we just open the file here instead of reading it all into memory?

test/helpers/venvtest/venvtest.go
```diff
@@ -92,7 +92,7 @@ func New() *venv.Venv {
 			Width:       func() int { return 0 },
 		},
 		Writers: &writer.Writers{Writer: io.Discard, ErrWriter: io.Discard},
-	}
+	}).WithProcessEnv(map[string]string{})
```
yhakbar: Why are we using a method for this instead of assigning a value directly in the struct like we do for the rest?

yhakbar (changes requested):
If we can get rid of `ProcessEnv`, I think that would be really good and address my concerns here. If not, I think this does the job and can be merged in. Just tell me if you want the stamp.

pkg/config/dependency_state_azurerm.go
denis256: Updated

yhakbar (dismissed):

yhakbar (approved):

PR #6733, opened 2026-08-21
fix: make hcl fmt --exclude-dir match paths, not only directory basenames

yzxcj797:
Fixes #6721.

## Summary

`terragrunt hcl fmt --exclude-dir` compared every walked directory's **basename** against the raw flag value, so any path-shaped entry matched nothing and the excluded directories were still formatted:

```
terragrunt hcl fmt --exclude-dir './terraform/module'   # no-op
terragrunt hcl fmt --exclude-dir 'terraform/module'     # no-op
terragrunt hcl fmt --exclude-dir '**/terraform/module'  # no-op
```

Only single-segment names (`--exclude-dir module`) ever worked.

The flag now accepts:

- a directory name, as before (`module` — matches any directory with that basename);
- a slash-separated path relative to the working directory (`terraform/module`, `./terraform/module/` — normalized before matching);
- a doublestar pattern when the entry contains glob metacharacters (`**/terraform/module` — matched with the doublestar library already used by Terragrunt's filters).

## Testing

Added `TestHCLFmtExcludeDirPath` to the format command tests, parametrized over the four shapes above against the existing fixture tree: files under the excluded subtree stay byte-identical to the unformatted fixture while the rest are formatted.

```
go test ./internal/cli/commands/hcl/format/ -count=1 -run TestHCLFmtExcludeDirPath  # ok
```

Note: `TestBytesDiff`, `TestRunForFiles` and `TestHCLFmtFilterMultiple` fail on this Windows machine identically **with and without** this change (backslash-path assumptions in the tests themselves, unrelated to this fix).

## Summary by CodeRabbit

* **New Features**
  * HCL formatting exclusions now support directory names, relative paths, normalized paths, and doublestar patterns.
* **Bug Fixes**
  * Improved exclusion matching so files in targeted directories are correctly skipped, while other files continue to be formatted.
* **Tests**
  * Added coverage for basename, relative path, normalized path, and pattern-based exclusions.

PR #6734, opened 2026-08-22
docs: point --exclude-dir users at the --filter flag

yzxcj797:
As suggested by @yhakbar in the review of #6733: `--exclude-dir` is supported for legacy reasons and its path matching is narrower than `--filter`'s Terragrunt filter expressions. The flag's help text now says so and points users at `--filter`.

## Scope

Only the `--exclude-dir` flag description in `internal/cli/commands/hcl/format/cli.go` — no behavior change.

## Summary by CodeRabbit

* **Documentation**
  * Updated the `--exclude-dir` option description to identify it as legacy.
  * Added guidance to use `--filter` instead, including an example filter expression.

### DIFF
--- a/internal/cli/commands/hcl/format/cli.go
+++ b/internal/cli/commands/hcl/format/cli.go
@@ -49,7 +49,7 @@ func NewFlags(l log.Logger, opts *options.TerragruntOptions, v *venv.Venv, prefi
 				Name:        ExcludeDirFlagName,
 				EnvVars:     tgPrefix.EnvVars(ExcludeDirFlagName),
 				Destination: &opts.HclExclude,
-				Usage:       "Skip HCL formatting in given directories.",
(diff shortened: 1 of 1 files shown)


internal/cli/commands/hcl/format/cli.go
```diff
@@ -49,7 +49,7 @@ func NewFlags(l log.Logger, opts *options.TerragruntOptions, v *venv.Venv, prefi
 				Name:        ExcludeDirFlagName,
 				EnvVars:     tgPrefix.EnvVars(ExcludeDirFlagName),
 				Destination: &opts.HclExclude,
-				Usage:       "Skip HCL formatting in given directories.",
+				Usage:       "Skip HCL formatting in given directories. Supported for legacy reasons; prefer the --filter flag, which accepts Terragrunt filter expressions (e.g. --filter='!./terraform/module').",
```
yhakbar, line 52: This isn't valid usage of the flag. Please review the docs I linked in the prior PR. This should be a glob.

Also, editing usage like this is good, but I was referencing the actual tip system:
https://github.com/gruntwork-io/terragrunt/tree/main/internal/tips

PR #6761, opened 2026-08-27
perf: Streaming buffered output

yhakbar:
## Description

Streams output in a couple places where we were previously buffering all the content in memory before it got to its final destination. Dramatically reduces memory consumption in some places as consequence.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

- **Performance**
  - Reduced memory usage when running all plans, including large outputs.
  - Improved handling of provider responses and streamed log output.

- **Reliability**
  - Plan files are written safely and atomically, preserving previous files when runs fail.
  - Improved detection of plan errors as output is received.
  - Temporary files are cleaned up after successful or failed writes.
  - Oversized provider responses are rejected safely.

- **Documentation**
  - Added a changelog entry describing the reduced-memory improvements in v1.1.5.

### DIFF
--- a/internal/runner/json_output.go
+++ b/internal/runner/json_output.go
@@ -1,0 +1,68 @@
+package runner
+
+import (
+	"bufio"
+	"errors"
+	"io"
+	"path/filepath"
+
+	"github.com/gruntwork-io/terragrunt/internal/vfs"
+)
+
+const (
+	jsonOutputDirPerms  = 0o700
+	jsonOutputFilePerms = 0o600
+
+	// jsonOutputBufferSize coalesces the small chunks arriving from `show -json`
+	// into fewer, larger writes. Without it, streaming trades one large allocation
+	// for a write syscall per chunk and runs slower than the buffering it replaced.
+	// Every unit running concurrently holds one, so it stays fixed rather than
+	// scaling with the document.
+	jsonOutputBufferSize = 256 << 10
+
+	jsonOutputTempPattern = ".*.tmp"
+)
+
+// WriteJSONOutput streams whatever fn writes into the file at path, creating the
+// parent directory first. Plan JSON runs to tens of megabytes on large units and
+// every unit in a `run --all` produces its own, so the writer is handed straight
+// to fn rather than accumulating the document in memory.
+//
+// The document lands in a temporary file that replaces path only once fn returns
+// cleanly. A failed run therefore leaves whatever was already at path untouched,
+// and no reader observes a half-written plan.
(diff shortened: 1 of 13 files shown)


internal/tf/cache/helpers/client.go
```diff
@@ -83,17 +89,33 @@ func decodeResponse(resp *http.Response) ([]byte, error) {
...
 
-	buffer, err := ResponseBuffer(resp)
+	reader, err := ResponseReader(resp)
 	if err != nil {
 		return nil, err
 	}
 
-	bodyBytes, err := io.ReadAll(buffer)
-	if err != nil {
+	body, readErr := readBody(reader, resp.ContentLength)
```
denis256, line 97: Not sure if `readBody` will handle an invalid `resp.ContentLength`

```
  HTTP/1.1 200 OK
  Content-Type: application/json
  Content-Length: 20
  {"versions":[ ... thousands of bytes of real JSON ... ]}

```

  Looks like workflow is:
  1. readBody allocates exactly 20 bytes.
  2. io.ReadFull reads 20 bytes and returns success.
  3. Rest of the body is ignored.
  4. json.Unmarshal runs on the truncated slice.

internal/runner/json_output.go
```diff
@@ -0,0 +1,68 @@
...
+// cleanly. A failed run therefore leaves whatever was already at path untouched,
+// and no reader observes a half-written plan.
+func WriteJSONOutput(fsys vfs.FS, path string, fn func(w io.Writer) error) error {
+	if err := fsys.MkdirAll(filepath.Dir(path), jsonOutputDirPerms); err != nil {
+		return err
+	}
+
+	// Writers can aim at one path concurrently, so the scratch file cannot be named
+	// after it.
+	file, err := vfs.CreateTemp(fsys, filepath.Dir(path), filepath.Base(path)+jsonOutputTempPattern)
```
denis256, line 41: Temp is always path+".tmp". A crash can leave it on disk; it may have secrets inside...

internal/tf/cache/helpers/client.go
yhakbar: Addressed in fbe893d2995be7767643bc70280a427c62d25107

internal/runner/json_output.go
yhakbar: Addressed. We create the temp file with a pattern, and I added panic protection here: 3a586460a6c5de28a8dff9a94c93e73801f5cbe2

Overall, we're writing the output to disk anyways, so I'm not sure this is that concerning.

denis256 (dismissed):

denis256 (approved):

PR #6776, opened 2026-08-27
fix: Writing more generated files atomically

yhakbar:
## Description

Writes more generated files atomically to avoid issues from incomplete writes.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when generating configuration, reports, debug files, lockfiles, and other outputs by writing them atomically.
  * Existing files are preserved if generation fails or is interrupted.
  * Replacing files through symlink paths now updates the requested path without modifying the target.
  * Generated files now use secure user-only permissions; shared read-only copies use restricted permissions.
  * Improved handling of long and multibyte filenames and missing parent directories.

* **Documentation**
  * Added v1.1.5 changelog entries describing atomic file generation and updated file permissions.

### DIFF
--- a/internal/vfs/atomic.go
+++ b/internal/vfs/atomic.go
@@ -13,7 +13,7 @@ const atomicTempPattern = ".*.tmp"
 
 // WriteFileAtomic writes data to path through a scratch file that replaces path
 // only once the content is on disk. See [StreamFileAtomic] for what that buys and
-// how perm is applied. The parent directory must already exist.
+// how perm is applied.
 func WriteFileAtomic(fsys FS, path string, data []byte, perm os.FileMode) error {
 	return StreamFileAtomic(fsys, path, perm, func(w io.Writer) error {
 		_, err := w.Write(data)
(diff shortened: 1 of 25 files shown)


internal/vfs/atomic.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	file, err := CreateTemp(fsys, dir, TempPattern(filepath.Base(path)))
+	if err != nil {
+		return err
+	}
+
+	tmpPath := file.Name()
+
+	if err := fsys.Chmod(tmpPath, perm); err != nil {
```
denis256, line 67: Looks like this PR bypasses umask 077 and makes generated files 0644 instead of 0600. This was not happening on main before

denis256 (dismissed):

internal/vfs/atomic.go
yhakbar: Addressed in 7ab2d2c69534253b107a97651a3d28c43075aee2

denis256 (dismissed):

denis256 (approved):
Windows-only: regenerating a shared mutable-generate file can fail with Access is denied when renaming over the read-only CAS hardlink.

PR #6799, opened 2026-09-01
fix: Keep symlinked terragrunt.stack.hcl anchored to its own directory

### LINKED ISSUE #6441: terragrunt stack generate resolves symlinked terragrunt.stack.hcl to its target dir, breaking relative read_terragrunt_config
## Describe the bug

When a `terragrunt.stack.hcl` is a symlink, `terragrunt stack generate` uses the symlink target's directory as the stack's source directory instead of the directory where the symlink actually lives. As a result, any relative file access from the stack file (e.g. `read_terragrunt_config`, `find_in_parent_folders`, relative `source`) is resolved against the wrong folder and fails.

Example error

```
Error in function call
  on .../source/terragrunt.stack.hcl line 2, in locals:
   2:   sibling = read_terragrunt_config("${get_terragrunt_dir()}/sibling.hcl")
Call to function "read_terragrunt_config" failed: You attempted to run terragrunt
in a folder that does not contain a terragrunt.hcl file. Please add a terragrunt.hcl
file and try again.
Path: ".../source/sibling.hcl"
```
Note the path points at source/ (the symlink target) rather than the directory containing the symlink.

## Steps To Reproduce

1. Create a real stack file at `source/terragrunt.stack.hcl` that reads a sibling config relative to its own directory (e.g. `read_terragrunt_config("${get_terragrunt_dir()}/sibling.hcl")`).
2. In a separate live/ directory, create a symlink l`ive/terragrunt.stack.hcl` -> `../source/terragrunt.stack.hcl` and place `live/sibling.hcl` next to the symlink.
3. Run `terragrunt stack generate --working-dir live`.
4. Generation fails because Terragrunt looks for `sibling.hcl` in `source/` instead of `live/`.

## Expected behavior

The stack file's source directory should be the directory where the symlink is located (`live/`), so relative reads resolve next to the symlink — the behavior in `v0.93.0` and earlier.

## Must haves

- [X] Steps for reproduction provided.

## Nice to haves

- [X] Terminal output

## Versions

- Terragrunt version: 1.1.0
- OpenTofu/Terraform version: 1.14.5
- Environment details (Ubuntu 20.04, Windows 10, etc.): macOS / Linux

## Additional context

Introduced after `v0.93.0`. First shipped in `v1.0.3` via #5962 — commit 16ad54f2c ("fix: fixing concurrent-write race in stack generate"). That change began passing discovered stack-file paths through `util.CanonicalResolvedPath`, which runs `filepath.EvalSymlinks` on the full path (including the symlinked file), rewriting it to the target location. `GenerateStackFile` then derives `stackSourceDir := filepath.Dir(stackFilePath)`, which points at the target's directory.

guillaumegiordanasym:
Stack generation passed each discovered stack-file path through util.CanonicalResolvedPath, which runs filepath.EvalSymlinks on the full path including the symlinked terragrunt.stack.hcl itself. That rewrote the path to the symlink target's location, so GenerateStackFile derived stackSourceDir from the target directory rather than the directory the symlink lives in. Relative reads (read_terragrunt_config, find_in_parent_folders, relative source) then resolved against the wrong folder and failed.

Regression introduced by 16ad54f2c (#5962), which replaced the plain filepath.Join of the stack file name with util.CanonicalResolvedPath over the whole path. That commit is absent from v1.0.2 and first shipped in v1.0.3, so v1.0.2 is the last unaffected release.

## Description

Fixes #6441.

canonicalStackFilePath resolves and canonicalises only the stack file's directory, then joins terragrunt.stack.hcl onto the result. The directory still resolves, so topology keys stay consistent with the canonicalised working dir, while the stack file stays anchored to the directory it lives in.

Two tests in generate_test.go drive ListStackFilesWithExcludes over a symlinked tree: one asserts a symlinked terragrunt.stack.hcl keeps its own directory, the other that a symlinked directory collapses into a single deduplicated entry. A stack integration test generates a stack from a symlinked terragrunt.stack.hcl whose read_terragrunt_config reads a sibling relative to get_terragrunt_dir(), a read that only resolves next to the symlink.

helpers.SymlinkUnsupported reports whether a host refuses symlink creation outright, so a test that needs a symlink skips on that alone and stays fatal on every other error. Windows refuses with ERROR_PRIVILEGE_NOT_HELD absent SeCreateSymbolicLinkPrivilege or Developer Mode, and Go's syscall.Errno.Is folds that errno into neither fs.ErrPermission nor errors.ErrUnsupported, so it is matched explicitly in test_helpers_unix.go and test_helpers_windows.go.

Adds a v1.1.5 changelog entry under bug-fixes.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed stack generation for symlinked stack files so relative paths resolve from the directory containing the symlink.
  * Improved stack discovery when symlinked directories are involved, preventing duplicate stack file results.

* **Documentation**
  * Documented the corrected behavior for symlinked stack files and relative configuration lookups.


guillaumegiordanasym:
@coderabbitai full review

PR #6827, opened 2026-09-07
chore: Setting up per-run env and version to increase test parallelization

yhakbar:
## Description

Un-serializes some testing that was previously required to remain serialized because it mutated global state by moving that logic to venv, etc.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

- **Tests**
  - Improved test isolation by applying environment variables per test instead of changing shared process-wide settings.
  - Enabled parallel execution for additional integration and regression tests.
  - Added coverage for renderer-specific flush-error behavior.
  - Preserved credential and configuration discovery during parallel integration test runs.
- **Refactor**
  - Improved test command helpers to accept explicit versions and environment overrides.

### DIFF
--- a/test/helpers/package.go
+++ b/test/helpers/package.go
@@ -1161,13 +1161,52 @@ func RemoveFolder(t *testing.T, path string) {
 	}
 }
 
+// envCtxKey keys the environment a run should see, so a test can adjust it
+// without touching the process and can therefore still run in parallel.
+type envCtxKey struct{}
+
+// RunEnv returns the environment a Terragrunt run would start with, for a test
+// to adjust and hand back through [ContextWithEnv].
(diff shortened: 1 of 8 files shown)


test/helpers/package.go
```diff
@@ -1185,9 +1224,13 @@ func RunTerragruntCommandWithContext(
 	configDir := t.TempDir()
 
 	v := venv.OSVenv().WithUserConfigDir(func() (string, error) { return configDir, nil })
+	if env := envFromContext(ctx); env != nil {
+		v = v.WithEnv(env)
```
denis256, line 1228: env is copied as is not cloned

denis256 (approved):

PR #6835, opened 2026-09-08
fix(config): restore v1.1.3 base64gzip output by default

denis256:
## Description

- add legacygzip encoder reproducing Go 1.26 gzip bytes
- register it over base64gzip unless strict control enabled
- add legacy-base64gzip strict control for current encoder
- warn once per run when legacy bytes are used

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

- **New Features**
  - Added `base64gzip_compat()` to preserve legacy compressed output.
  - Added the `base64gzip-compat` experiment to enable the compatibility function.
  - Added the `legacy-base64gzip` strict control to opt into the current encoder and suppress warnings.

- **Bug Fixes**
  - Restored `base64gzip()` output compatibility with Terragrunt v1.1.3 and earlier.
  - Added a warning about upcoming encoder behavior changes in Terragrunt 1.2.

- **Documentation**
  - Added usage guidance for the compatibility function, experiment, and strict control.


docs/src/data/strict-controls/legacy-base64gzip.mdx
```diff
@@ -0,0 +1,21 @@
...
+
+Since v1.1.5, `base64gzip()` returns the v1.1.3 bytes again, so upgrading from v1.1.3 or earlier plans no change. If you applied the v1.1.4 output, every plan shows the encoded value changing back until you apply it or enable this control, and a resource that replaces on user data changes is replaced by that apply.
+
+Enable this control to keep the current Go encoder's output instead:
+
+```bash
+terragrunt run plan --strict-control legacy-base64gzip
+```
+
+The current encoder will become the default in a future version of Terragrunt.
```
yhakbar: Let's call out that it will be the default behavior in 1.2, specifically.

internal/strict/controls/controls.go
```diff
@@ -108,6 +111,10 @@ const LegacyGCSDeprecationWarning = "Plain `https://www.googleapis.com/storage/.
 	"silence this warning, or enable the `legacy-gcs-public-prefix` strict control to download anonymously instead." +
 	" This will be the default behavior of Terragrunt in the future."
 
+const LegacyBase64GzipWarning = "`base64gzip()` returns the gzip bytes produced by Terragrunt v1.1.3 and earlier (Go 1.26) for " +
+	"backward compatibility. Enable the `legacy-base64gzip` strict control to use the current Go encoder and silence " +
+	"this warning. This will be the default behavior of Terragrunt in the future."
```
yhakbar: Let's call out that it will be the default behavior in 1.2 specifically.

yhakbar (dismissed):

docs/src/content/docs/04-reference/01-hcl/04-functions.mdx
```diff
@@ -1186,3 +1187,27 @@ In this example, the `v2.0.0` version of the module made a breaking change to re
...
+
+<Before version="v1.1.5">
+
+The `base64gzip_compat` function ships with v1.1.5.
+
+</Before>
+
+<Since version="v1.1.5">
+
+`base64gzip_compat(str)` returns the value `base64gzip(str)` returned in Terragrunt v1.1.3 and earlier, and keeps returning it after `base64gzip()` moves to the current Go encoder.
```
yhakbar: We should make this experimental, as we might want to use a different name for OpenTofu parity.

yhakbar (changes requested):

internal/strict/controls/controls.go
denis256: Done in ea7027f7a: the warning now says Terragrunt 1.2 will switch `base64gzip()` to the current Go encoder.

docs/src/data/strict-controls/legacy-base64gzip.mdx
denis256: Done in ea7027f7a: the page now says the current encoder becomes the default in Terragrunt 1.2.

docs/src/content/docs/04-reference/01-hcl/04-functions.mdx
denis256: Done in ea7027f7a and f1dcd7f24: `base64gzip_compat` is behind the new `base64gzip-compat` experiment (mirrors `deep-merge`: typed error without the experiment, experiment page, experiments-added changelog entry), and the docs note the name may change to match OpenTofu.

internal/gzipcompat/gzipcompat.go
```diff
@@ -0,0 +1,74 @@
...
+
+const gzipHeader = "\x1f\x8b\x08\x00\x00\x00\x00\x00\x00\xff"
+
+// Encode returns the gzip stream Terragrunt v1.1.3 produced for input, base64 encoded.
+func Encode(input string) (string, error) {
+	data := []byte(input)
+
+	var output bytes.Buffer
+
+	_, _ = output.WriteString(gzipHeader)
```
yhakbar: Can we update this to use `output := bytes.NewBufferString(gzipHeader)` so that we don't have these ignored return values? The trailer can be appended with this:

```
out := output.Bytes()
out = binary.LittleEndian.AppendUint32(out, crc32.ChecksumIEEE(data))
out = binary.LittleEndian.AppendUint32(out, uint32(len(data)))

return base64.StdEncoding.EncodeToString(out), nil
```

denis256: Done. Encode now builds the buffer with bytes.NewBufferString(gzipHeader) and appends the trailer with binary.LittleEndian.AppendUint32, so there are no ignored return values left. Output is unchanged: the 594-input differential against the Go 1.26.5 reference and the byte comparison against the real v1.1.3 and v1.1.4 binaries both still match.

docs/src/data/experiments/base64gzip-compat.mdx
```diff
@@ -0,0 +1,27 @@
...
+terragrunt run --all --experiment base64gzip-compat -- plan
+```
+
+### `base64gzip-compat` - How to provide feedback
+
+Open a [GitHub issue](https://github.com/gruntwork-io/terragrunt/issues) describing where you need stable `base64gzip()` output.
+
+### `base64gzip-compat` - Criteria for stabilization
+
+To transition the `base64gzip-compat` feature to a stable release, the following must be addressed:
```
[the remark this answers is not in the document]
denis256: Done. The experiment page now states up front that base64gzip-compat is ongoing, disabled by default, that base64gzip_compat is not generally available while the experiment runs and errors without it, and that it becomes generally available once the stabilization criteria are met. The enablement command, behavior, example and feedback path were already on the page.

internal/gzipcompat/gzipcompat.go
denis256: Done. Encode now builds the buffer with bytes.NewBufferString(gzipHeader) and appends the trailer with binary.LittleEndian.AppendUint32, so no ignored return values remain. Output is unchanged: the 594-input differential against the Go 1.26.5 reference and the byte comparison against the real v1.1.3 and v1.1.4 binaries both still match.

docs/src/data/experiments/base64gzip-compat.mdx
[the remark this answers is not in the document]
denis256: Done. The experiment page now states up front that base64gzip-compat is ongoing, disabled by default, that base64gzip_compat is not generally available while the experiment runs and returns an error without it, and that it becomes generally available once the stabilization criteria are met. The enablement command, behavior, example and feedback path were already present.

internal/gzipcompat/deflate.go
```diff
@@ -0,0 +1,1411 @@
+package gzipcompat
```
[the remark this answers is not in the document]
denis256: Not adding a per-file copyright header here, by project decision.

This file is not a verbatim copy. It is a single-compression-level DEFLATE encoder: the level table, dictionary support, Reset, the public Writer API and the fast-path encoder are gone, the lookup tables are derived from the RFC 1951 base tables rather than copied, the sort types were replaced with slices.SortFunc, and the identifiers, constants and structure differ throughout. Its only contract is byte output, which is pinned by goldens generated from Terragrunt v1.1.3 plus a 594-input differential.

Attribution policy for the repository is a maintainer decision rather than a per-file one. If the maintainers want a notice carried in-tree, say so and it will be added.

yhakbar (approved):

PR #6837, opened 2026-09-08
fix: restore AWS IMDS credential fallback for container environments

### LINKED ISSUE #6822: v1.1.4 regression: AWS credentials fail on hosts where IMDSv2 is unreachable but IMDSv1 works (venv HTTP client disables SDK's IMDS fail-fast/fallback)
## Describe the bug

Since **v1.1.4**, Terragrunt fails to resolve AWS credentials on any host where the **IMDSv2 token endpoint is unreachable but IMDSv1 still works** — most commonly an EC2 instance whose metadata hop limit is 1, with Terragrunt running inside a container (one extra hop). This is the exact scenario the AWS SDK's IMDSv1 fallback is designed to handle, and it works in every version up to and including v1.1.3.

Every `run` against such a host fails with:

```
error assuming role: operation error STS: AssumeRole, get identity: get credentials:
failed to refresh cached credentials, no EC2 IMDS role found,
operation error ec2imds: GetMetadata, canceled, context deadline exceeded
```

### Root cause

v1.1.4 (PR #6645, "venv plumbing for cloud SDKs") added `config.WithHTTPClient(v.HTTP)` to the two places in `internal/awshelper/config.go` that load AWS config — [`AWSConfigBuilder.Build`](https://github.com/gruntwork-io/terragrunt/blob/v1.1.4/internal/awshelper/config.go#L124) and [`AssumeIamRole`](https://github.com/gruntwork-io/terragrunt/blob/v1.1.4/internal/awshelper/config.go#L287):

```go
 cfg, err := config.LoadDefaultConfig(
     ctx,
     config.WithRegion(region),
     config.WithAppID("terragrunt/"+version.GetVersion()),
+    config.WithHTTPClient(v.HTTP),   // new in v1.1.4
 )
```

`v.HTTP` is [`vhttp.NewOSClient()`](https://github.com/gruntwork-io/terragrunt/blob/v1.1.4/internal/vhttp/vhttp.go#L40), a plain `*http.Client` wrapping a clone of the stdlib `http.DefaultTransport`.

The AWS SDK's IMDS client applies its fail-fast timeouts **only when the configured HTTP client is the SDK's own `*awshttp.BuildableClient`** — `feature/ec2/imds/api_client.go`:

```go
func resolveHTTPClient(client HTTPClient) HTTPClient {
	if client == nil {
		client = awshttp.NewBuildableClient()
	}

	if c, ok := client.(*awshttp.BuildableClient); ok {   // <-- type assertion fails for a plain *http.Client
		client = c.
			WithDialerOptions(func(d *net.Dialer) {
				// ... The client should fail fast in this case.
				d.Timeout = defaultDialerTimeout             // 250ms
			}).
			WithTransportOptions(func(tr *http.Transport) {
				// ... EC2Metadata service drops the connection after a
				// single IP Hop. The client should fail fast in this case.
				tr.ResponseHeaderTimeout = defaultResponseHeaderTimeout   // 500ms
			})
	}

	return client
}
```

Those two timeouts exist precisely for the "container behind a hop limit" case, as the SDK's own comments say. Passing a plain `*http.Client` silently strips them, which changes the outcome:

| | HTTP client used | IMDSv2 token `PUT` | Result |
|---|---|---|---|
| **≤ v1.1.3** | SDK's `BuildableClient` | gives up after **500ms** (`ResponseHeaderTimeout`) | `tokenProvider` sees the `RequestSendError`/`CanceledError`, disables the token, **falls back to IMDSv1**, credentials retrieved ✅ |
| **v1.1.4** | plain `*http.Client` (no response-header cap) | hangs until the SDK's **5s per-operation IMDS timeout** fires | whole `GetMetadata` operation is cancelled before the IMDSv1 fallback can run → `no EC2 IMDS role found` ❌ |

The fallback logic itself is intact in the SDK ([`token_provider.go`](https://github.com/aws/aws-sdk-go-v2/blob/feature/ec2/imds/v1.18.35/feature/ec2/imds/token_provider.go#L184-L202) disables the token on a send error/cancellation and returns `bypassTokenRetrievalError` so the request proceeds over IMDSv1) — v1.1.4 just never gets there, because the token request alone consumes the entire operation budget.

## Steps To Reproduce

Any EC2 host with `HttpTokens=optional` and `HttpPutResponseHopLimit=1`, running Terragrunt inside a container (e.g. a Docker-executor CI runner), with the instance profile as the only credential source (no `AWS_ACCESS_KEY_ID`, no `AWS_WEB_IDENTITY_TOKEN_FILE`) and a `root.hcl` that assumes a role:

```hcl
# root.hcl
iam_role = "arn:aws:iam::<account-id>:role/<role-name>"

remote_state {
  backend = "s3"
  generate = {
    path      = "shared.tf"
  }
  config = {
    bucket       = "<state-bucket>"
    key          = "${path_relative_to_include()}/terraform.tfstate"
    region       = "us-east-1"
    use_lockfile = true
  }
}
```

```console
$ terragrunt run plan
... error assuming role: operation error STS: AssumeRole, get identity: get credentials:
    failed to refresh cached credentials, no EC2 IMDS role found,
    operation error ec2imds: GetMetadata, canceled, context deadline exceeded
```

The environment can be confirmed independently from inside the same container — the token `PUT` is blackholed (no RST, just no reply), while IMDSv1 answers immediately:

```console
$ curl -s -o /dev/null -w 'http_code=%{http_code} elapsed=%{time_total}s\n' -m 5 \
    -X PUT "http://169.254.169.254/latest/api/token" \
    -H "X-aws-ec2-metadata-token-ttl-seconds: 60"
http_code=000 elapsed=5.001028s          # curl exit 28 (timeout)

$ curl -s -o /dev/null -w 'http_code=%{http_code} elapsed=%{time_total}s\n' -m 5 \
    "http://169.254.169.254/latest/meta-data/iam/security-credentials/"
http_code=200 elapsed=0.001247s
```

Swapping **only** the Terragrunt binary between runs on the same host, same config, same environment:

| Terraform | Terragrunt | Result |
|---|---|---|
| 1.15.9 | 1.1.1 | ✅ plan succeeds |
| 1.15.8 | 1.1.1 | ✅ plan succeeds |
| 1.15.9 | **1.1.4** | ❌ fails, 3/3 attempts |

Terraform itself is unaffected on the same host because its AWS provider uses the SDK's default `BuildableClient` and therefore still falls back to IMDSv1.

## Expected behavior

Terragrunt should resolve credentials from IMDS the same way every other AWS SDK–based tool does, falling back to IMDSv1 when the IMDSv2 token request cannot be completed — i.e. the behavior of v1.1.3 and earlier.

## Must haves

## Nice to haves

- [x] Terminal output

## Versions

- Terragrunt version: **v1.1.4** (regression; v1.1.1, v1.1.2, v1.1.3 are unaffected — `WithHTTPClient` appears in `internal/awshelper/config.go` only in v1.1.4)
- OpenTofu/Terraform version: Terraform 1.15.8 and 1.15.9 (both affected, so not Terraform-related)
- Environment details: Alpine container on a Docker-executor CI runner, on an EC2 instance with `HttpPutResponseHopLimit=1`; credentials from the instance profile only
- `github.com/aws/aws-sdk-go-v2/feature/ec2/imds`: v1.18.35

## Additional context

Suggested fixes, roughly in order of preference:

1. Hand the SDK a client it can specialize — build the venv's AWS client as an `*awshttp.BuildableClient` (`aws-sdk-go-v2/aws/transport/http`) so `imds.resolveHTTPClient` applies its dialer/response-header timeouts, instead of a bare `*http.Client`.
2. Alternatively, don't override the HTTP client for AWS config loading at all (restoring ≤ v1.1.3 behavior), or pass an IMDS-specific client via `config.WithEC2IMDSClientEnableState`/an explicit `imds` client that carries the equivalent timeouts.

Either way it would be worth a regression test covering "IMDSv2 token endpoint hangs, IMDSv1 works", since the failure is invisible on any host where IMDSv2 is reachable — which is why it presumably got through CI.

As a workaround, raising the instance metadata hop limit to 2 also resolves it, but that requires infrastructure changes outside Terragrunt's control and silently affects every tool on the host.

denis256:
## Description

Fixes #6822.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored EC2 instance role credential access when running inside containers with restrictive metadata service hop limits.
  * IMDSv2 token requests now fail quickly, allowing fallback to IMDSv1 without waiting for the full credential lookup timeout.

* **Documentation**
  * Added release notes describing the container credential fix and its configuration-free behavior.


yhakbar (dismissed):

yhakbar (approved):

yhakbar (changes requested):
Wait, the changelog needs fixing.

docs/src/data/changelog/unreleased/imds-fallback-buildable-client.mdx
yhakbar, line 1: We should set the release number here. We don't do `unreleased`. Also, this changelog entry is hard to read, especially for an end-user that doesn't understand Golang mechanics. We should focus on the behavior they care about as end users.

yhakbar (approved):

PR #6853, opened 2026-09-09
fix: set span status and error.type on failed operations

### LINKED ISSUE #6852: Telemetry does not record errors on spans/metrics per OpenTelemetry conventions
## Describe the bug

This summary heavily relies on the ["Recording error"](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/recording-errors.md) OpenTelemetry documentation, hence it's crucial to read it before diving into the text below.

When an operation fails, Terragrunt attaches an `exception` **event to the span** but never sets the **span status**, and never sets **`error.type`** on spans or metrics. As a result a failed run is indistinguishable from a successful one in any OpenTelemetry-native backend (New Relic), because backends classify errors by **span status**, not by the presence of an exception event.

> The presence of a span exception event doesn't qualify the span as an error on its own. Only spans with span status set to ERROR are classified as errors.
> https://docs.newrelic.com/docs/opentelemetry/get-started/apm-monitoring/opentelemetry-apm-ui/#distributed-tracing-page

### Current behaviour

`Telemeter.Collect` records the error and returns:

```go
// internal/telemetry/tracer.go
if err := fn(ctx); err != nil {
    // record error in span
    span.RecordError(err)
    return err
}
```

There is no `span.SetStatus(codes.Error, ...)`, and no `error.type` attribute is added. On the metric side, failures are reported as separate counters (`unit_run_errors_count`, `apply_errors_count`, `errors_count`, …) rather than as an attribute on the corresponding `duration` histogram.

### Why this is a problem

The OpenTelemetry ["Recording error"](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/recording-errors.md) document states that when an operation ends with an error, instrumentation:

- SHOULD set the **span status** code to **Error**
- SHOULD set the `error.type` attribute
- SHOULD set the **span status description** to the exception message

and, for metrics, that the operation duration histogram SHOULD include `error.type` so users can derive error rates, rather than reporting separate success and failure metrics.

Backends follow that contract literally. New Relic, for example, documents that:

> span is classified as an error when its status is set to ERROR" and, explicitly, that "the presence of a span exception event doesn't qualify the span as an error on its own" (docs). Error rate, the errors inbox and error alerting are all driven by span status.

### Evidence

From a failed `run --all apply` in our estate (Terragrunt 1.1.4, OTLP/HTTP exporter to New Relic), for a single pipeline run:
```nrql
FROM Span SELECT count(*) FACET name, span.kind, otel.status_code
```
* `["apply ","internal",null]` – 1  ← apply command root span
* `["unit_run","internal",null]` – 7 unit run spans
* `["run_terraform","internal",null]` – 40
... 2478 spans, **`otel.status_code` is null** on every one.

The failure is real and visible elsewhere in the same run: `unit_run_errors_count = 1` as a metric, and `exit_code = 1` as an attribute on the `run_terraform` span. Only the status - the field backends **actually read** - is missing.

Incidentally, the only spans in that our dataset that do carry ERROR status come from `google.cloud.kms.v1.KeyManagementService/Decrypt` from [`sops_decrypt_file`](https://docs.terragrunt.com/reference/hcl/functions/#sops_decrypt_file) function call, i.e. the Google Cloud SDK's own gRPC instrumentation, which picks up Terragrunt's globally registered tracer provider and follows the convention correctly.

### Proposed change

In `Telemeter.Collect`/ `Tracer.Collect`, on the error path:

1. `span.SetStatus(codes.Error, err.Error())` alongside the existing `RecordError`.
2. Set `error.type` from the error's type (its canonical Go type name is a good low-cardinality default, e.g. `run.DownloadingTerraformSourceErr`), keeping `RecordError` for the full message and stack.
3. On the metric side, add `error.type` to the existing `*_duration` histograms on failure, per the same document. The existing `*_errors_count` counters could stay for backwards compatibility.

Happy to open a PR for 1 and 2 if the approach looks right.

## Steps To Reproduce

```hcl
# modules/broken/main.tf

output "boom" {
  value = nonexistent_function("boom") # Fails during evaluation: Error: Call to unknown function: There is no function named "nonexistent_function".
}
```

```hcl
# terragrunt.hcl

terraform {
  source = "../../modules/broken"
}
```

```shell
$ export OTEL_EXPORTER_OTLP_ENDPOINT="https://otlp.nr-data.net"
$ export OTEL_EXPORTER_OTLP_HEADERS="<REDACTED>"
$ export TG_TELEMETRY_TRACE_EXPORTER=otlpHttp
$ export TG_TELEMETRY_METRIC_EXPORTER=otlpHttp
$ terragrunt run --all -- plan
...
* Failed to execute "terraform plan -input=false" in ./.terragrunt-cache/jkB-ZNizuQnQrOvukyLa0Ls1tqo/VXx5GLWGJaBDdM7cJUK7InD4R60
  ╷
  │ Error: Call to unknown function
  │
  │   on main.tf line 6, in output "boom":
  │    6:   value = nonexistent_function("boom")
  │
  │ There is no function named "nonexistent_function".
  ╵

  exit status 1
```

Image

Image

Image

Image

Could be verified locally as well:
```shell
$ export TG_TELEMETRY_TRACE_EXPORTER=console
$ terragrunt run --all -- plan
...Status: { Code: Unset }...
```

## Expected behavior

Spans and histogram metrics adhere Recording error OpenTelemetry semantic, hence backend can mark signals correctly.

## Versions

- Terragrunt version: 1.1.4
- OpenTofu/Terraform version: 1.7.0

### Related findings

- `tracestate` is dropped. The incoming header is parsed for trace id, span id and flags only (`trace.SpanContextConfig` in `internal/telemetry/tracer.go`), and only `TRACEPARENT` is propagated to child processes, so vendor sampling and priority state does not survive the Terragrunt hop. The W3C spec expects propagators to carry `tracestate` through unmodified.
- Trace flags are parsed as a decimal number: `strconv.Atoi(traceFlagsStr)`. The W3C field is two hex digits, so a valid value containing a hex letter (`0a`, `ff`) fails with invalid trace flags, and `02` (the random flag from trace context level 2) is silently treated as sampled.

### Resources

* https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/recording-errors.md
* https://docs.newrelic.com/docs/opentelemetry/get-started/apm-monitoring/opentelemetry-apm-ui/#distributed-tracing-page

shashankvarma499:
## Description

Fixes #6852.

When an operation fails, Terragrunt currently records only an `exception` event on the span via `span.RecordError(err)`. It never sets the span status or the `error.type` attribute, so OpenTelemetry-native backends that classify errors by span status (rather than by the presence of an exception event) cannot distinguish failed runs from successful ones.

This PR updates `Tracer.Trace`'s error path to:

1. Set the span status to `Error` with the error message as the description, alongside the existing `RecordError`.
2. Set the `error.type` attribute to the error's canonical Go type name (a low-cardinality default), keeping `RecordError` for the full message and stack.

Per the [OpenTelemetry recording-errors convention](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/recording-errors.md).

The metric-side change (adding `error.type` to the `*_duration` histograms) is intentionally left out of this PR and can be a follow-up; the existing `*_errors_count` counters remain for backwards compatibility.

## TODOs

- [ ] Update the docs.

## Summary by CodeRabbit

- **Bug Fixes**
  - Failed operations now mark telemetry spans with an `Error` status, include the error message, and record the error type according to OpenTelemetry conventions.
  - Successful operations continue to leave span status unchanged.

- **Documentation**
  - Added a changelog entry describing the telemetry error-status behavior.

### DIFF
--- a/docs/src/data/changelog/v1.1.5/telemetry-error-status.mdx
+++ b/docs/src/data/changelog/v1.1.5/telemetry-error-status.mdx
@@ -1,0 +1,15 @@
+---
+version: "v1.1.5"
+category: "bug-fixes"
+---
+
+#### Failed operations now mark their span as errored
+
+When an operation failed, Terragrunt recorded an `exception` event on the span but never
+set the span status or the `error.type` attribute. OpenTelemetry-native backends (such as
+New Relic) classify errors by span status, not by the presence of an exception event, so a
+failed run was indistinguishable from a successful one.
+
+Terragrunt now sets the span status to `Error` with the error message as its description,
+and records the `error.type` attribute with the error's type, per the
+[OpenTelemetry recording-errors convention](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/recording-errors.md).
--- a/internal/telemetry/tracer.go
+++ b/internal/telemetry/tracer.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"reflect"
 	"runtime/debug"
 	"strconv"
 	"strings"
@@ -14,6 +15,8 @@ import (
 
 	"github.com/gruntwork-io/terragrunt/pkg/log"
 	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/exporters/stdout/stdouttrace"
 	"go.opentelemetry.io/otel/sdk/resource"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
@@ -223,12 +226,26 @@ func (tracer *Tracer) Trace(
 	defer span.End()
 
 	if err := fn(ctx); err != nil {
-		// record error in span
+		// record error in span, set span status and error.type per the
+		// OpenTelemetry recording-errors convention so backends that
+		// classify errors by span status (not by exception events) see them.
 		span.RecordError(err)
+		span.SetStatus(codes.Error, err.Error())
+		span.SetAttributes(attribute.String("error.type", errorTypeName(err)))
 		return err
 	}
 
 	return nil
+}
+
+// errorTypeName returns the canonical Go type name of an error, used as a
+// low-cardinality value for the OpenTelemetry `error.type` attribute.
+func errorTypeName(err error) string {
+	if err == nil {
+		return ""
+	}
+
+	return reflect.TypeOf(err).String()
 }
 
 // openSpan creates a new span with attributes.
--- a/internal/telemetry/error_status_test.go
+++ b/internal/telemetry/error_status_test.go
@@ -1,0 +1,78 @@
+package telemetry
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/sdk/trace/tracetest"
+)
+
+func newTestTracer(t *testing.T) (*Tracer, *tracetest.InMemoryExporter) {
+	t.Helper()
+
+	exporter := tracetest.NewInMemoryExporter()
+	provider := sdktrace.NewTracerProvider(sdktrace.WithSpanProcessor(sdktrace.NewSimpleSpanProcessor(exporter)))
+	t.Cleanup(func() { require.NoError(t, provider.Shutdown(context.Background())) })
+
+	return &Tracer{
+		Tracer:       provider.Tracer("test"),
+		provider:     provider,
+		spanExporter: exporter,
+	}, exporter
+}
+
+// TestTraceSetsErrorStatusAndType verifies that a failed operation records the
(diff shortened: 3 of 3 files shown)


internal/telemetry/tracer.go
```diff
@@ -224,14 +226,28 @@ func (tracer *Tracer) Trace(
...
 }
 
+// errorTypeName returns the canonical Go type name of an error, used as a
+// low-cardinality value for the OpenTelemetry `error.type` attribute.
+func errorTypeName(err error) string {
+	if err == nil {
+		return ""
+	}
+
+	return reflect.TypeOf(err).String()
```
yhakbar: I'm not sure this will actually be a useful error. Terragrunt errors are frequently wrapped with `fmt.Errorf("%w")`, so most of the time this is gonna end up being `*fmt.wrapError`. 

It's probably better to have some logic here to try to unwrap the error as deep as possible before using reflect to read that error type.

docs/src/data/changelog/v1.1.5/telemetry-error-status.mdx
```diff
@@ -0,0 +1,15 @@
+---
+version: "v1.1.5"
```
yhakbar: This should be retargeted.

PR #6854, opened 2026-09-09
perf: Tuning FS concurrency

yhakbar:
## Description

Adds a Filesystem probe used to detect the filesystem a user is running in, and adjusts filesystem based concurrency controls based on it.

In #6838 I noted that 4 workers was the sweet spot for `git cat-file --batch` shards on APFS, and that with more extensive benchmarking we might want to tune the worker count differently on Linux and Windows. This is that benchmarking.

TL;DR: the number of concurrent file operations a machine can absorb varies pretty widely by filesystem. Measuring the same workload across filesystems at 2, 4, 8 and 16 vCPU in Docker containers. 

ext4 kept gaining out to 16 workers whether it had 2 cores or 16. 
btrfs wanted 2 everywhere. 
APFS wanted 4 everywhere.

So we now have a filesystem probe in `vfs.DetectFSKind`. `vfs.FSWorkersForKind` takes that `FSKind` and returns a worker count that the CAS, the tree materializer and the parallel walk all leverage. A filesystem with no measurement behind it keeps a conservative default, and an in-memory `vfs.FS` bypasses the probe.

Measured on Linux in Docker with `--cpus`, against a fresh store, cold ingest of a 3,000 file repository:

| filesystem | vCPU | before | after | change |
| --- | --- | --- | --- | --- |
| ext4 | 16 | 138.3ms | 112.4ms | -19% |
| btrfs | 16 | 266.4ms | 213.6ms | -20% |

Materializing that repository out of a warm store, where the old `GOMAXPROCS/2` was worst:

| filesystem | vCPU | before | after | change |
| --- | --- | --- | --- | --- |
| ext4 | 2 | 40.8ms | 24.4ms | -40% |
| btrfs | 2 | 48.7ms | 34.2ms | -30% |
| overlayfs | 2 | 71.5ms | 56.4ms | -21% |
| btrfs | 16 | 42.0ms | 38.1ms | -9% |
| overlayfs | 16 | 50.4ms | 46.1ms | -9% |

On an Apple M3 Max, capping the parallel walk at 4 workers took a fast copy of a module directory from 325ms to 104ms (-68%) for many small files and 170ms to 63ms (-63%) for a mix including a few large ones.

`terragrunt hcl fmt` got tuned too. 16 workers ran about 65% slower than 8. It now defaults to `min(GOMAXPROCS, 8)`. Formatting 1,024 files went from 24.9ms to 21.3ms (-14%).

## TODOs

[...]

PR #6864, opened 2026-09-10
perf: Improving worktree performance

yhakbar:
## Description

A Git filter expression like `--filter '[main...HEAD]'` generates a worktree for every reference it names so that discovery can find the units that changed in them. We were generating those worktrees in a pretty expensive way: one `git worktree add` at a time, each checking out the entire tree with a single worker, then removing them one at a time when the run ended.

This works through optimizing that in three stages.

First, `git worktree add` now runs with `checkout.workers` set to `vfs.FSWorkers`, and the worktrees are removed concurrently instead of sequentially. This will get updated before or after #6854 , depending on sequencing of the PR reviews.

Second, worktrees are now generated using `git archive` piped into `archive/tar` in process, rather than forking `git worktree add` when possible. We can't always do this because of conditional stuff in `.gitattributes`, so we still fall back to the first method when we detect this. When we can, it allows us to use `git worktree add --detach --no-checkout` in the target directory, then the contents of the worktree is generated with the results of the `git archive` output. By doing this, we only have to run `git worktree add --detach --no-checkout` sequentially to avoid contention on the `.git` directory of the main worktree and we can then generate the worktree files on disk from `git archive` concurrently without issue.

Third, we also conditionally avoid generating worktrees and avoid generating all of the worktree when we know we don't have to. If we know that one worktree isn't going to be used (e.g. we only look at units added or removed), we don't bother with generating the other worktree. We also allow discovery commands (`find`, `list` and `browse`) to perform sparse checkouts when their Git expression allows it.

In my benchmarking on an Apple M3 Max, against a synthetic repository with 15,000 tracked files:

| stage | `terragrunt find --filter '[HEAD~1...HEAD]'` |
| --- | --- |
| before | 4.7s |
| `checkout.workers` and concurrent removal | 3.9s |
| filled from `git archive` | 3.4s |
| only what the diff calls for | 0.3s |

We still fully generate worktrees when:

- A `run` command is being used.
- A discovery command requires parsing configurations (e.g. through `--dependencies` or `--mode dag`).
- Any change to a `terragrunt.stack.hcl` file in either reference.
- When `--experiment symlinks` is used.

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

- **Performance**
  - Worktree creation is faster by checking Git changes first and skipping unnecessary generation.
  - Common discovery commands such as `find`, `list`, and `browse` skip more unnecessary work.

- **Improved Worktree Handling**
  - Worktrees can use filtered checkouts containing only required repository paths.
  - Git archive-based materialization improves handling of large repositories and filtered content.
  - Symlink traversal and filesystem operations are more reliable, including support for linked directories.
  - Worktree cleanup and concurrent creation are more efficient.


internal/worktrees/worktrees.go
```diff
@@ -402,106 +737,51 @@ func NewWorktrees(
...
+				gitRunner,
+				survey.refsToMaterialize(l, gitExpressions, gitRefs),
+				repoRemote,
+				repoBranch,
+				repoCommit,
+				experiments,
+				pathspecs,
+			)
+			if err != nil {
+				worktrees = newEmptyWorktrees(workingDir, gitRunner)
```
denis256, line 770: Successful worktrees from a partial `createGitWorktrees` failure are discarded here, so Cleanup never runs for them and temp dirs / git worktree registrations are left behind

denis256 (approved):

PR #6866, opened 2026-09-10
fix(catalog): report actually-generated files in scaffold message

### LINKED ISSUE #6748: Terragrunt catalog TUI reports `terragrunt.hcl` scaffolded when boilerplate template generates a different file
## Describe the bug

When scaffolding a boilerplate template from the `terragrunt catalog` TUI, the post-scaffold success message refers to the wrong output file. It reports that `terragrunt.hcl` was scaffolded, even when the template actually generates different files and there are no `terragrunt.hcl` among them.

## Steps To Reproduce

Catalog scaffold success message for boilerplate templates (`gruntwork-io/terragrunt`).

1. In an empty directory (or any directory where you want the scaffold output), run:

   ```bash
   terragrunt catalog github.com/Nerdeez/terragrunt-catalog
   ```

2. In the catalog TUI, open **Templates** and select a boilerplate template, for example:
   - **Live root template** (`templates/root`) — generates `root.hcl`
   - **config** (`templates/config`) — generates files under `config/` (not `terragrunt.hcl`)

3. Press `s` to scaffold into the current directory (`.`).

4. Complete the scaffold prompts (fill in template variable values).

5. Observe the success message shown after scaffolding completes.

6. List the files that were actually created in the output directory (e.g. `ls -la`).

## Expected behavior

After scaffolding a boilerplate template from the catalog TUI, the success message should reflect what was actually created—not assume `terragrunt.hcl`.

For example, when scaffolding the **Live root template** from `github.com/Nerdeez/terragrunt-catalog`, the message should indicate that `root.hcl` was generated (and any dependency outputs, such as `config/`), e.g.:

```
root.hcl scaffolded

./root.hcl
./config/common.hcl
./config/billing.hcl
./config/region.hcl
```

When scaffolding the **config** template alone, the message should list the files under `config/` that were created, not `terragrunt.hcl`.

In general, the post-scaffold callout should name the file(s) produced by the template (or say something like “template scaffolded successfully” with an accurate list of generated paths), so users are not directed to open a file that does not exist.

## Must haves

## Nice to haves

- [x] Terminal output
- [ ] Screenshots

### Template (bug) — Live root template

Success message claims `terragrunt.hcl` was scaffolded:

```
╭─────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                         │
│  terragrunt.hcl scaffolded                                                                              │
│                                                                                                         │
│  ./terragrunt.hcl                                                                                       │
│                                                                                                         │
│  Values you filled in the form are populated; unfilled inputs are marked with `# TODO: fill in value`.  │
│                                                                                                         │
│  Open the file and replace any remaining TODO placeholder with a real                                   │
│  value before running terragrunt.                                                                       │
│                                                                                                         │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

Actual files on disk after scaffold (`ls -la`):

```
total 8
drwxr-xr-x@ 3 yarivkatz  staff    96 Aug 26 12:44 .
drwxr-xr-x@ 8 yarivkatz  staff   256 Aug 24 17:53 ..
-rw-r--r--@ 1 yarivkatz  staff  2609 Aug 26 12:44 root.hcl
```

No `terragrunt.hcl` — only `root.hcl`.

## Versions

- Terragrunt version: `v1.1.3`
- OpenTofu/Terraform version: OpenTofu `v1.10.7`
- Environment: macOS (darwin), zsh
- Catalog: `github.com/Nerdeez/terragrunt-catalog` (main)

shashankvarma499:
## Description

Fixes #6748.

The `catalog` TUI's post-scaffold callout always reported `terragrunt.hcl scaffolded` and pointed the user at `terragrunt.hcl`, even when a boilerplate template generated entirely different files (for example `root.hcl`, or files under `config/`).

This threads the list of generated files from `Plan.Generate` (which already computes it as `allFiles`) through `scaffoldCmd` and `ScaffoldFinishedMsg` into `formatScaffoldMessage`, which now:

- names the single generated file when exactly one was written (`root.hcl scaffolded`), or
- reports the count and lists every generated path when several were written (`2 files scaffolded`), and
- falls back to `terragrunt.hcl` when no file list was captured, preserving existing behavior.

## TODOs

- [ ] Update the docs.

## Summary by CodeRabbit

- **Bug Fixes**
  - Catalog scaffolding messages now list the files actually generated.
  - Success messages display accurate file counts and paths, including dependency files.
  - Single-file results show the file name; multiple files show a count and their paths.
  - Long file lists are limited to five entries with an indication of additional files.
  - Existing fallback messaging remains available when no generated file list is provided.

- **Documentation**
  - Added a v1.1.5 changelog entry documenting the corrected scaffolding message.


internal/cli/commands/scaffold/scaffold.go
```diff
@@ -398,6 +406,8 @@ func (p *Plan) Generate(
 
 	allFiles = slices.Compact(slices.Sorted(slices.Values(allFiles)))
 
+	p.generatedFiles = allFiles
```
yhakbar, line 419: Boilerplate only populates dependencies here when `opts.Manifest` is true. The fact that this wasn't caught speaks to a lack of testing. Please add additional testing to verify that this works correctly for templates.

internal/cli/commands/catalog/tui/update.go
```diff
@@ -594,7 +603,18 @@ func formatScaffoldMessage(opts *options.TerragruntOptions, interactive bool) st
...
-	return renderValuesBox(valuesBoxAccentGreen, heading, path, summary, body)
+	return renderValuesBox(valuesBoxAccentGreen, heading, strings.Join(paths, "\n"), summary, body)
+}
+
+// scaffoldHeading titles the post-scaffold callout from the generated files:
+// the single file's base name when exactly one file was written, otherwise a
+// count.
+func scaffoldHeading(files []string) string {
+	if len(files) == 1 {
+		return filepath.Base(files[0]) + " scaffolded"
```
yhakbar: We probably shouldn't be using `filepath.Base` here. It should probably be relativized properly from the generation root.

internal/cli/commands/catalog/tui/update.go
yhakbar, line 590: This should likely be pluralized with the pluralize helper below. Please ensure that sufficient testing exists so that this shows up and we can verify the correct behavior is present.

yhakbar, line 597: This should likely be pluralized with the pluralize helper below. Please ensure that sufficient testing exists so that this shows up and we can verify the correct behavior is present.

internal/cli/commands/catalog/tui/model_test.go
```diff
@@ -577,6 +577,35 @@ func TestModelScaffoldFinishedSetsExitMessage(t *testing.T) {
...
+
+	l := logger.CreateLogger()
+	components := makeComponents(t)
+
+	componentCh := make(chan *tui.ComponentEntry)
+	close(componentCh)
+
+	m := tui.NewModelStreaming(t.Context(), l, venvtest.NewOSWithEmptyEnv(), opts, components[0], componentCh, nil)
+
+	updated, _ := m.Update(tui.ScaffoldFinishedMsg{Files: []string{"root.hcl", "config/common.hcl"}})
```
yhakbar: This should be updated to use `filepath.FromSlash` or `filepath.Join` so that the test passes on Windows. The assertion below should be updated too.

internal/cli/commands/catalog/tui/update.go
```diff
@@ -573,13 +574,21 @@ func formatScaffoldMessage(opts *options.TerragruntOptions, interactive bool) st
...
+		files = []string{config.DefaultTerragruntConfigPath}
+	}
 
 	heading := lipgloss.NewStyle().
 		Foreground(lipgloss.Color(valuesBoxAccentGreen)).
 		Bold(true).
-		Render("terragrunt.hcl scaffolded")
+		Render(scaffoldHeading(files))
+
+	paths := make([]string, 0, len(files))
```
yhakbar: This could get huge, depending on the template. There should be a cap with `... +N more` or something to cut it off instead of flooding the screen.

yhakbar (changes requested):

shashankvarma499:
Thanks for the detailed pass. All three points make sense.

- I'll route the message through the existing pluralize helper and add a test that asserts both the singular and the plural wording, so the text is actually covered.
- Good catch on Windows. I'll build the generated paths with filepath.FromSlash/filepath.Join and update the assertion in the test accordingly.
- I'll cap the file list and append a "... +N more" suffix so a large template cannot flood the terminal.

Will push an update shortly.

shashankvarma499:
Pushed an update addressing all the review points:

- pluralized the scaffold heading via the existing pluralize helper (with singular and plural covered in the test)
- capped the listed paths at 5 with a "... +${N} more" suffix so a large template cannot flood the screen
- dropped filepath.Base for the single-file heading; it is now relativized from the generation root via displayPath, matching the listed paths
- collectDependencyFiles now returns each dependency path relative to the generation output dir
- this also surfaced that boilerplate only records dependency files when opts.Manifest is set, so that is now enabled for catalog scaffolding and covered by a test
- documented that copyable plans intentionally leave GeneratedFiles empty, with a test pinning that behavior (per the CodeRabbit note)

gofmt clean, go build ./... passes, and ./internal/cli/commands/catalog/tui/... and ./internal/cli/commands/scaffold/... tests pass.

yhakbar:
Hey @shashankvarma499 ,

Please make sure commits are signed.

shashankvarma499:
Thanks. Re-signed the commits on this branch, they are all verified now.

I also spotted the same unverified commit on my other open PR here, #6853, and re-signed that one too.

PR #6874, opened 2026-09-10
test: add regression test for AWS_CA_BUNDLE custom CA roots

### LINKED ISSUE #6873: Terragrunt 1.1.x fails AWS configuration when AWS_CA_BUNDLE is set because Venv injects a plain *http.Client
## Describe the bug

Terragrunt 1.1.x fails while evaluating AWS HCL helpers such as `get_aws_caller_identity_arn()` whenever `AWS_CA_BUNDLE` is set.

The failure happens during AWS SDK configuration, before any AWS request is sent:

```text
error loading AWS config: unable to add custom RootCAs HTTPClient, has no WithTransportOptions, *http.Client
```

Terragrunt 1.0.1 works with the same configuration.

This is a Terragrunt 1.1 regression. In 1.1.x, `internal/awshelper` always injects Terragrunt’s generic Venv HTTP client (`*http.Client` from `vhttp.NewOSClient()`) into `config.LoadDefaultConfig`. AWS SDK v2 can only apply `AWS_CA_BUNDLE` to an `*awshttp.BuildableClient`. A plain `*http.Client` does not implement `WithTransportOptions`, so configuration fails.

`AWS_CA_BUNDLE` is the documented AWS SDK mechanism for custom CA roots, including HTTPS inspection proxies.

## Steps To Reproduce

1. Use Terragrunt 1.1.4 (or any 1.1.x release with the Venv HTTP client wired into `awshelper`).

2. Create a valid PEM CA bundle. A system bundle is enough; the contents do not matter as long as the file exists and is readable:

```bash
export AWS_CA_BUNDLE="$(python3 -c 'import certifi; print(certifi.where())')"
# or: export AWS_CA_BUNDLE=/etc/ssl/cert.pem
```

3. Create this fixture:

```hcl
# terragrunt.hcl
locals {
  caller_arn = get_aws_caller_identity_arn()
}

terraform {
  source = "."
}
```

Valid AWS credentials are not required. The crash happens while loading AWS config, before STS is called.

4. Evaluate the configuration:

```bash
terragrunt render
```

Any command that evaluates the helper also reproduces it (`terragrunt run -- providers lock`, `terragrunt plan`, etc.).

## Expected behavior

Terragrunt should load `AWS_CA_BUNDLE` and continue. If credentials are missing or invalid, the error should come from STS, not from HTTP-client construction.

## Must haves

## Nice to haves

- [x] Terminal output

```text
ERROR  ./terragrunt.hcl:3,16-42: Error in function call; Call to function "get_aws_caller_identity_arn" failed: error loading AWS config: unable to add custom RootCAs HTTPClient, has no WithTransportOptions, *http.Client.
```

## Versions

- Terragrunt version: 1.1.4 (fails); 1.0.1 (works)
- OpenTofu/Terraform version: Terraform 1.9.8 (not required for this failure)
- Environment details: macOS Darwin 25.6.0, arm64

### Root cause

In Terragrunt 1.1.4, `internal/vhttp.NewOSClient()` returns a plain `*http.Client`:

```go
func NewOSClient() Client {
    return &http.Client{Transport: newOSTransport()}
}
```

`AWSConfigBuilder.Build()` always injects that client:

```go
configOptions = append(
    configOptions,
    config.WithAppID("terragrunt/"+version.GetVersion()),
    config.WithHTTPClient(v.HTTP),
)

cfg, err := config.LoadDefaultConfig(ctx, configOptions...)
```

`AssumeIamRole` does the same.

AWS SDK v2 then sees `AWS_CA_BUNDLE` and tries to attach the PEM roots with `WithTransportOptions`. That API exists on `*awshttp.BuildableClient`, not on `*http.Client`.

This also affects other Terragrunt-owned AWS operations that use the same builder (remote-state S3/DynamoDB, `iam_role` / assume-role, HCL identity helpers).

We originally hit this behind an HTTPS MITM proxy that sets `AWS_CA_BUNDLE` and `HTTPS_PROXY`. The proxy is not required to reproduce the config error; setting `AWS_CA_BUNDLE` is enough.

### Recommended minimal fix

Do not inject the generic Venv HTTP client when `AWS_CA_BUNDLE` is set, in both `AWSConfigBuilder.Build` and `AssumeIamRole`:

```go
configOptions := []func(*config.LoadOptions) error{
    config.WithAppID("terragrunt/" + version.GetVersion()),
}

if v.Env["AWS_CA_BUNDLE"] == "" {
    configOptions = append(configOptions, config.WithHTTPClient(v.HTTP))
}
```

If no client is supplied, AWS SDK v2 creates an `awshttp.BuildableClient`, applies `AWS_CA_BUNDLE`, and still honors `HTTPS_PROXY`.

### Longer-term option

Keep generic HTTP (`vhttp.Client`) separate from the AWS SDK HTTP client, and default the AWS path to `awshttp.NewBuildableClient()`. That preserves test injection and satisfies the SDK’s custom-CA contract.

### Suggested regression test

Set `AWS_CA_BUNDLE` to a valid test PEM on the production/default AWS HTTP path and assert that building AWS config does not return the `WithTransportOptions` error.

shashankvarma499:
## Description

Adds a regression test for #6873, where setting `AWS_CA_BUNDLE` broke AWS configuration because the injected HTTP client was a plain `*http.Client`.

The AWS SDK v2 applies custom CA roots via `WithTransportOptions`, which only exists on `*awshttp.BuildableClient`. When a plain `*http.Client` was handed to `config.LoadDefaultConfig`, loading failed with:

```
error loading AWS config: unable to add custom RootCAs HTTPClient, has no WithTransportOptions, *http.Client
```

The underlying fix landed in #6837, which routes the venv's OS transport through `AWSBuildableClient` so the SDK can apply `AWS_CA_BUNDLE`. This change adds the regression test the issue asked for, so the behavior stays covered.

## Test

The new test builds AWS config with a self-signed CA bundle set via `AWS_CA_BUNDLE` and asserts that:

- `Build` returns no error.
- The HTTP client is an `*awshttp.BuildableClient`.
- The custom CA root is present in the transport's `RootCAs` pool.

```bash
go test ./internal/awshelper/ -run TestAWSConfigCustomCABundle
```

Fixes #6873

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved AWS configuration handling to ensure certificates specified through `AWS_CA_BUNDLE` are correctly applied to HTTP connections.

### DIFF
--- a/internal/awshelper/buildable_client_test.go
+++ b/internal/awshelper/buildable_client_test.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net/http"
 	"net/http/httptest"
+	"os"
 	"path/filepath"
 	"strings"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


denis256:
```
remote:   Found 1 violation:
remote:
remote:   0378bc662ecccaddd2ec64f1f74402bcf8e35740
```

shashankvarma499:
Re-signed the commit and force-pushed. The branch tip is now a verified commit, so the signature requirement should be satisfied.

PR #6889, opened 2026-09-12
fix: normalize path_relative_to_include separators on Windows

### LINKED ISSUE #6749: Terragrunt Windows string path buliding mixing backslashes and forward slashes
## Describe the bug

When using Windows, path-string building uses backslash for all operations, including the case of fetching the remote backend Terraform states for Terragrunt units. 

I'm using Azure Storage Account for storing the remote backend for all units and the constructed URL for calling it is set as following mixing the backslashes and forward slashes:

`https://<storage-account-name>.blob.core.windows.net/<container>/path\to\unit/terraform.tfstate`

This gives the following error:

```
ERROR  [.] terraform.exe: Error: Error loading state: building request: building GET request: parse "https://<storage-account-name>.blob.core.windows.net/<container>/<path>\<to>\<unit>/terraform.tfstate": net/url: invalid control character in URL
```

This behaviour doesn't happen on previous Terragrunt versions, for example, `0.99.4` which works with the same Terraform version `1.14.6` without issue

## Steps To Reproduce

Example of the given unit where it is in a Windows path `live\unit\this\terragrunt.hcl` that has a dependency the `live\another_unit\this\terragrunt.hcl`

**`live\unit\this\terragrunt.hcl`**
```hcl
terraform {
  source = "git::<sample-repo>//path/to/module?ref=<git-ref>"
}

dependencies {
  paths = [
    "../../another_unit/this"
  ]
}

dependency "another_unit" {
  config_path = "../../another_unit/this"

  mock_outputs = {
    id = "xxxx"
  }
}

inputs = { }
```
**`live\another_unit\this\terragrunt.hcl`**
```hcl
terraform {
  source = "git::<sample-repo>//path/to/other-module?ref=<git-ref>"
}

inputs = { }
```

## Expected behavior

All path strings should be converted to use forward slashes when calling remote backends and all non-local filesystem operations. So it should finally resolve the path to:

`https://<storage-account-name>.blob.core.windows.net/<container>/path/to/unit/terraform.tfstate`

## Versions

- Terragrunt version: 1.1.3
- OpenTofu/Terraform version: 1.14.6
- Environment details (Ubuntu 20.04, Windows 10, etc.): Windows 11

## Additional context

This behaviour doesn't happen on previous Terragrunt versions, for example, `0.99.4` which works with the same Terraform version `1.14.6` without issue

shashankvarma499:
## Description

Fixes #6749.

`path_relative_to_include()` and `path_relative_from_include()` compute their result with `filepath.Rel`, which returns native (backslash) separators on Windows. When that result is used as a remote-state `key` — for example the azurerm backend builds its blob URL from `${path_relative_to_include()}/terraform.tfstate` — the backslashes are embedded in the URL and rejected as an invalid control character, so `init`/`apply` fail with `net/url: invalid control character in URL`.

This normalizes both results to forward slashes with `filepath.ToSlash`, matching the cross-platform contract already documented by their tests. Unix and macOS behaviour is unchanged (`filepath.ToSlash` is a no-op there).

## TODOs

- [ ] Update the docs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Windows-generated relative paths now consistently use forward slashes.
  * Prevents invalid backslash separators in remote-state URLs.
  * Unix and macOS path behavior remains unchanged.

* **Documentation**
  * Added a changelog entry describing the Windows path separator fix.

### DIFF
--- a/docs/src/data/changelog/v1.1.5/windows-path-relative-to-include-separators.mdx
+++ b/docs/src/data/changelog/v1.1.5/windows-path-relative-to-include-separators.mdx
@@ -1,0 +1,10 @@
+---
+version: "v1.1.5"
+category: "bug-fixes"
+---
+
+#### Normalize `path_relative_to_include` separators on Windows
+
+On Windows, `path_relative_to_include()` and `path_relative_from_include()` returned paths with native backslash separators, because they were computed with `filepath.Rel`. When such a path is used as a remote-state `key` (for example the azurerm backend builds its blob URL from it), the backslashes end up in the URL and are rejected as an invalid control character.
+
+Both functions now normalize their result to forward slashes with `filepath.ToSlash`, matching the cross-platform contract documented by their tests. Unix and macOS behaviour is unchanged.
--- a/pkg/config/config_helpers.go
+++ b/pkg/config/config_helpers.go
@@ -912,7 +912,7 @@ func PathRelativeToInclude(
 		)
 	}
 
-	return result, nil
+	return filepath.ToSlash(result), nil
 }
 
 // PathRelativeFromInclude returns the relative path from the current Terragrunt configuration to the included Terragrunt configuration file
@@ -948,7 +948,7 @@ func PathRelativeFromInclude(
 		)
 	}
 
-	return result, nil
+	return filepath.ToSlash(result), nil
 }
 
 // getTerraformCommand returns the current terraform command in execution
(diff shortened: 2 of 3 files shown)


yhakbar:
Thanks for creating this, @shashankvarma499 , but we're not making this change. It's intended functionality for Terragrunt hcl functions returning paths to return os native paths. Users are expected to use `replace()` if necessary to normalize to forward slashes. We may introduce `to_slash` and `from_slash` in the future to make this process simpler on Windows.

PR #6899, opened 2026-09-14
fix(amazonsts): stop json-out-dir IAM self-assume

### LINKED ISSUE #6898: Regression: `--json-out-dir` re-assumes `iam_role` signed with the role's own session (AccessDenied)
## Describe the bug

On `main` (post-v1.1.4, unreleased), a unit that assumes an IAM role and is run with `--json-out-dir` calls `sts:AssumeRole` a second time **signed with the role's own session credentials** instead of the base identity. Against real AWS the second call fails with `AccessDenied` ("not authorized to perform sts:AssumeRole") unless the role's trust policy trusts itself.

Every unit in the run is affected, not just one: in a 3-unit `run --all`, all 3 units make a self-signed second call.

This is a regression against v1.1.4, and it is the same failure class as #6521, which was fixed for backend operations in v1.1.2.

### Trigger

All four conditions:

1. `iam_role` (or `--iam-assume-role`) is set
2. `--json-out-dir` is used
3. `--iam-assume-role-duration` is **not** set
4. Any credential source — static env vars, shared-config profile, IMDS, or `--auth-provider-cmd`

### Root cause

Two things combine:

1. **The credentials cache never hits on the default path.** `internal/runner/run/creds/providers/amazonsts/provider.go:89-90` computes the TTL from the raw option:

   ```go
   cacheDuration := time.Duration(iamRoleOpts.AssumeRoleDuration) * time.Second
   credentialsCache.Put(ctx, iamRoleOpts.RoleARN, creds, time.Now().Add(cacheDuration))
   ```

   `AssumeRoleDuration` is `0` unless `--iam-assume-role-duration` is passed, so the entry is stored already expired and `internal/cache/cache.go:99` (`time.Now().After(item.Expiration)`) treats it as a miss on the very next read. The role is therefore re-assumed on every entry. This part is pre-existing and byte-identical in v1.1.4.

2. **The second assumption now reads its own output.** #6883 (8471662d0) added to `AssumeIamRole`, `internal/awshelper/config.go:317-319`:

   ```go
   if envCreds := createCredentialsFromEnv(v.Env); envCreds != nil {
       configOptions = append(configOptions, config.WithCredentialsProvider(envCreds))
   }
   ```

   The credential getter writes the assumed session back into that same `v.Env` map, so the second assumption signs with `ASIA...` rather than the base identity. In v1.1.4 `AssumeIamRole` passed no credentials provider, so the SDK default chain resolved from the **process** environment, which the venv mutation never touched.

The second invocation exists because of `--json-out-dir`: `internal/runner/unit_runner.go:96` runs the unit, then `:177` runs it again for `show -json`, passing `v.WithWriter(w)` — a shallow struct copy (`internal/venv/venv.go:187-192`) that shares the `Env` map.

## Steps To Reproduce

No AWS account or credentials are needed: a local mock STS endpoint records which identity signs each call.

**1. Mock STS server** (`mocksts/main.go`, `go mod init mocksts && go build`):

```go
package main

import (
	"fmt"
	"net/http"
	"os"
	"regexp"
	"sync/atomic"
)

var n int64
var credRe = regexp.MustCompile(`Credential=([A-Z0-9]+)/`)

const xml = `<AssumeRoleResponse xmlns="https://sts.amazonaws.com/doc/2011-06-15/"><AssumeRoleResult><Credentials><AccessKeyId>ASIAASSUMEDSESSION</AccessKeyId><SecretAccessKey>assumed-secret</SecretAccessKey><SessionToken>assumed-token</SessionToken><Expiration>2030-12-31T23:59:59Z</Expiration></Credentials><AssumedRoleUser><AssumedRoleId>AROATEST:s</AssumedRoleId><Arn>arn:aws:sts::123456789012:assumed-role/deploy/s</Arn></AssumedRoleUser></AssumeRoleResult></AssumeRoleResponse>`

func main() {
	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		_ = r.ParseForm()
		key := "NONE"
		if m := credRe.FindStringSubmatch(r.Header.Get("Authorization")); m != nil {
			key = m[1]
		}
		fmt.Fprintf(os.Stderr, "STS#%d action=%s signed_with=%s\n",
			atomic.AddInt64(&n, 1), r.FormValue("Action"), key)
		w.Header().Set("Content-Type", "text/xml")
		fmt.Fprint(w, xml)
	})
	http.ListenAndServe("127.0.0.1:8611", nil)
}
```

**2. Fixture:**

```hcl
# mod/main.tf
output "value" { value = "hi" }
```

```hcl
# u/terragrunt.hcl
iam_role = "arn:aws:iam::123456789012:role/deploy"
terraform { source = "../mod" }
```

**3. Run:**

```bash
./mocksts 2> sts.log &

export AWS_ENDPOINT_URL_STS=http://127.0.0.1:8611
export AWS_REGION=us-east-1
export AWS_ACCESS_KEY_ID=AKIABASEKEY
export AWS_SECRET_ACCESS_KEY=base-secret
export TG_NON_INTERACTIVE=true

terragrunt run --all --out-dir ./plans --json-out-dir ./json -- plan
cat sts.log
```

## Expected behavior

Both `sts:AssumeRole` calls are signed with the base identity, as in v1.1.4.

## Terminal output

Observed, 1 unit, `run --all plan --out-dir X --json-out-dir Y`:

| Base credential source | v1.1.4 | main |
| --- | --- | --- |
| Static env vars | `AKIABASEKEY`, `AKIABASEKEY` | `AKIABASEKEY`, **`ASIAASSUMEDSESSION`** |
| Shared-config profile | `AKIAPROFILEKEY`, `AKIAPROFILEKEY` | `AKIAPROFILEKEY`, **`ASIAASSUMEDSESSION`** |

Total `AssumeRole` calls across a 3-unit run, and how many were self-signed:

| Scenario | v1.1.4 | main |
| --- | --- | --- |
| Parallel, plain `plan` | 3 | 3 |
| Parallel, `--json-out-dir` | 6 | 6 (3 self-signed) |
| Serial, plain `plan` | 3 | 3 |
| Serial, `--json-out-dir` | 6 | 6 (3 self-signed) |

## Verified negatives

- Plain `run --all plan` without `--json-out-dir` makes one call, signed with the base key.
- `--iam-assume-role-duration 3600` collapses it to a single cached call. **This is a working user workaround.**

## Suggested fix

Give the cache entry the session's real lifetime so the second entry is a cache hit, which is what the cache was there to guarantee:

```go
sessionDuration := time.Duration(cmp.Or(iamRoleOpts.AssumeRoleDuration, iam.DefaultAssumeRoleDuration)) * time.Second
credentialsCache.Put(ctx, iamRoleOpts.RoleARN, creds, time.Now().Add(sessionDuration-credentialsCacheExpiryWindow))
```

(`iam.DefaultAssumeRoleDuration` already exists, `internal/iam/iam.go:12`. A small expiry window keeps a command that starts just inside it on valid credentials.)

Two notes for whoever picks this up:

- **Cloning the env in the JSON-export path does not fix it.** `v.WithEnvCloned().WithWriter(w)` at `unit_runner.go:180` runs *after* `run.Run` has already mutated the shared `v.Env`, so the clone already holds the `ASIA...` session.
- **One behavior change to weigh.** `credentialsCache` is process-global and keyed on role ARN alone, so with a working TTL a serial run (`--parallelism 1`) collapses N assumptions into one shared session. Permissions are unchanged since it is the same role; what changes is CloudTrail granularity, one session name instead of N. Under default parallelism the units race the cache and still assume separately.

An alternative that preserves v1.1.4's call pattern exactly is to stop the assumption from reading back its own output, at the cost of more code.

## Test coverage gap

No AWS integration test exercises `--json-out-dir`, which is why CI did not catch this.

## Must haves

## Nice to haves

- [x] Terminal output
- [ ] Screenshots

## Versions

- Terragrunt version: `main` @ 5ab3c5b6d (post-v1.1.4, unreleased — v1.1.4 is not affected)
- OpenTofu/Terraform version: not required to reproduce (fails before any tofu/terraform invocation matters); verified with OpenTofu v1.12.2
- Environment details: Ubuntu, Linux 6.8, amd64

denis256:
## Description

  - cache assumed sessions with default duration
  - key cache by role and source identity
  - refresh STS with saved source credentials
  - cover reuse isolation and expiry paths

Close: #6898

## TODOs

- [x] Update the docs.

## Summary by CodeRabbit

- **Bug Fixes**
  - IAM role credentials are now reused across operations in the same process, including plan exports with `--json-out-dir`.
  - Credentials refresh automatically before expiration while continuing to use valid cached credentials if a refresh temporarily fails.
  - Credential sessions are kept separate when role settings or source credentials differ.

- **Documentation**
  - Clarified the default one-hour IAM session duration and credential reuse behavior.
  - Added release documentation describing session caching, refresh behavior, and JSON plan export support.


docs/src/content/docs/04-reference/01-hcl/03-attributes.mdx
```diff
@@ -166,6 +166,8 @@ iam_role = "arn:aws:iam::ACCOUNT_ID:role/ROLE_NAME"
 
 The `iam_assume_role_duration` attribute can be used to specify the STS session duration, in seconds, for the IAM role that Terragrunt should assume before invoking OpenTofu/Terraform.
 
+When unset, Terragrunt defaults the STS session to one hour (`3600` seconds). In the same process, it reuses cached credentials until five minutes before they expire, so later steps such as `--json-out-dir` plan export can reuse the first role assumption.
```
yhakbar: Do we need these docs updates? Isn't this assumed?

docs/src/data/flags/iam-assume-role-duration.mdx
```diff
@@ -8,4 +8,6 @@ env:
 
 Specifies how long the temporary credentials should remain valid when assuming an IAM role. This flag is only used when [`iam-assume-role`](/reference/cli/commands/run#iam-assume-role) is specified.
 
+When unset, Terragrunt uses a one-hour session. In the same process, it reuses cached credentials until five minutes before they expire, so later steps such as `--json-out-dir` plan export can reuse the first role assumption.
```
yhakbar: Do we need these docs updates? Isn't this assumed?

yhakbar (dismissed):

yhakbar (approved):

PR #6944, opened 2026-09-16
feat: add terraform.no_cache option to skip .terragrunt-cache

### LINKED ISSUE #6199: Option to run from the source directory without .terragrunt-cache
## Describe the enhancement

Add an option to run Terraform/OpenTofu from the source unit directory without using `.terragrunt-cache`.

In v0.99.x, Terragrunt ran directly from the unit directory when no `terraform { source }` block was set. In v1, it always runs from within `.terragrunt-cache`, which breaks our workflow in two ways:

1. **Relative local module paths stop resolving** — paths that correctly pointed to local modules when running from the unit directory no longer resolve correctly when running from inside `.terragrunt-cache`.

2. **`generate` output is invisible to IDE tooling** — in v0.99.x, `generate` wrote files (including variable declarations) to the unit directory, where IDE extensions could read them. In v1, those files are written to `.terragrunt-cache`, so IDE extensions cannot find the declarations and report errors.

Because of these issues, we are unable to upgrade from v0.99.x to v1.

## Additional context

- [x] Changes required: add an option to skip `.terragrunt-cache` and run directly from the source unit directory.
- [x] Implications of the feature: units without `terraform { source }` would run in-place, preserving relative module paths and making `generate` output visible to IDE tooling.
- [x] Alternatives considered: copying parent directories into the cache, using symlinks — both introduce maintenance burden and are not viable long-term solutions.
- [x] Level of effort: unclear, but a simple flag that preserves v0.99.x behavior for units without `terraform { source }` may be a starting point.

## PoC (Proof of Concept)

- [ ] [PoC]()

## RFC Not Needed

Chopsticks13:
## Summary

Adds a `terraform.no_cache` attribute that lets units skip the `.terragrunt-cache` copy when no external `terraform.source` is configured.

Fixes #6199

## Problem

Terragrunt v1 always copies the unit directory into `.terragrunt-cache` before running Terraform, even when there's no remote source to fetch. This was a deliberate consistency change from v0.99.x, but it broke three common workflows:

- **Relative local module paths** — `module "../shared"` resolves against the cache directory, not the original layout
- **IDE/LSP visibility** — generated files (backend configs, variable declarations) end up in the cache where terraform-ls and tflint can't find them
- **Unnecessary overhead** — copying files just to run them from a different directory adds latency with no benefit for local-only units

## Solution

New `terraform` block attribute:

```hcl
terraform {
  no_cache = true
}
```

When `no_cache = true` and no `terraform.source` is set, `PrepareSource` skips the download/copy entirely and keeps `WorkingDir` pointing at the unit directory. Generated files land in the unit directory where IDEs can see them, and relative module paths resolve correctly.

When an external source IS configured (via `terraform.source` or `--terragrunt-source`), `no_cache` is ignored — remote sources still need to be fetched somewhere.

## Changes

- `pkg/config/config.go` — add `NoCache *bool` to `TerraformConfig`
- `pkg/config/config_as_cty.go` — add `NoCache` to cty serialization for include/merge
- `pkg/config/translate.go` — translate `NoCache` to `runcfg.TerraformConfig`
- `internal/runner/runcfg/types.go` — add `NoCache bool` to runtime config
- `internal/runner/runcfg/util.go` — return `""` from `GetTerraformSourceURL` when no-cache and no source
- `internal/prepare/prepare.go` — skip download when `sourceURL` is empty
- `internal/runner/runcfg/util_test.go` — 3 new test cases covering no_cache behavior

## Test plan

- [x] Unit tests pass (`go test ./internal/runner/runcfg/... -v`)
- [x] Config package tests pass (`go test ./pkg/config/...`)
- [x] `go vet ./...` clean
- [ ] Integration test: create a unit with `no_cache = true`, run `terragrunt plan`, verify no `.terragrunt-cache` directory is created

## Summary by CodeRabbit

* **New Features**
  * Added the `terraform.no_cache` setting through the `no-cache` experiment, allowing Terraform to run directly from the unit directory without copying local sources to `.terragrunt-cache`.
  * Explicit Terraform sources, including command-line overrides, continue to use the cache when configured.

* **Bug Fixes**
  * Prevented unnecessary source copying when no-cache mode is enabled without an external source.
  * Preserved existing caching behavior when the experiment is not enabled.


Chopsticks13:
Hi @denis256 — would love to get some eyes on this when you get a chance. This addresses a long-standing request (#6199, 33 thumbsup) to skip the .terragrunt-cache copy for local modules. The implementation is minimal — a new `terraform.no_cache` opt-in flag that short-circuits `PrepareSource` when no external source is configured. Tests included.

yhakbar:
Hey @Chopsticks13,

Thanks for creating this. Do you mind gating the feature behind an experiment flag and adding some integration testing?

This won't make the 1.2 release cycle, but it might make it into a 1.2.x release.

Also, please make sure you [sign your commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).

Chopsticks13:
Thanks @yhakbar — all addressed in the latest push:

1. **Experiment flag**: `no_cache` is now gated behind `--experiment no-cache`. Without the flag, the attribute is silently ignored and the `.terragrunt-cache` copy proceeds as usual.

2. **Integration tests**: Added two tests in `integration_download_tf_test.go`:
   - `TestTFLocalNoCacheSkipsTerragruntCache` — verifies no `.terragrunt-cache` is created when the experiment is enabled
   - `TestTFLocalNoCacheWithoutExperimentStillUsesCache` — verifies the feature is inert without the experiment flag

3. **Commit signing**: Set up, commits should show as verified going forward.

Also added the experiment docs under `docs/src/data/experiments/no-cache.mdx`.

yhakbar:
@Chopsticks13 ,

Did you run those tests? Please take a second and run those tests along with the whole [integration test suite](https://docs.terragrunt.com/community/contributing/#integration-tests) before marking this PR as ready for review again.

Chopsticks13:
@yhakbar ran the integration tests locally — all passing, including the two new ones for no_cache. Also fixed a bug where the experiment gate was only in PrepareSource but not in the main Run() path. Latest push covers both.

